PR #1351, opened 2015-02-12
BUILD: add example apps into the build process

ChristianBeer:
- By default apps won't be built. In order to build the example apps use --enable-apps as option to ./configure. Not all apps are supported as some need further external libraries.
- Also fix a copy/paste error in the vboxmonitor Makefile

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -32,7 +32,7 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
 AC_CONFIG_SRCDIR(lib/shmem.cpp)
 
-AC_REVISION([$Revision$]) 
+AC_REVISION([$Revision$])
 REV=`echo '$Revision$' | awk "{print $2}"`
 RDATE=`date '+%Y.%m.%d'`
 if test -d .svn ; then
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -1308,6 +1322,7 @@ AC_CONFIG_FILES([
                  packages/solaris/CSW/boincmanager/Makefile
                  packages/solaris/CSW/boincmanager/pkginfo
                  packages/solaris/CSW/boincmanager/prototype
+                 samples/Makefile
```
ChristianBeer: none whitespace change

configure.ac
```diff
@@ -200,6 +210,9 @@ fi
 if test x$enable_server = xyes ; then
   configured_to_build="${configured_to_build} server"
 fi
+if test x$enable_apps = xyes ; then
+  configured_to_build="${configured_to_build} apps"
+fi
```
ChristianBeer: none whitespace change

configure.ac
```diff
@@ -121,7 +121,13 @@ AC_ARG_ENABLE(boinczip,
     [enable_boinczip=${enableval}],
     [enable_boinczip=no])
 
-AC_ARG_ENABLE(pkg-libs, 
+AC_ARG_ENABLE(apps,
+    AS_HELP_STRING([--enable-apps],
+                   [enable building the boinc example apps]),
+    [enable_apps=${enableval}],
+    [enable_apps=no])
+
```
ChristianBeer: none whitespace change

PR #1353, opened 2015-02-21
BUILD: enable manager build on travis-ci

ChristianBeer:
- pulls in wxwidgets 3.0.2 and builds it
- configures BOINC to use the downloaded library


ChristianBeer:
Building wxWidgets on travis-ci takes 23 minutes so we need to switch to the travis container-based-infrastructure where we can cache the wxWidgets build to reuse.

The script in this PR is usefull for this and will be integrated in #1354.

PR #1356, opened 2015-03-04
BoincAPI Windows static CRT leak

JasonGroothuis:
Handle and resource leakage corrected for main application timer thread
(first cut).  This covers the ordinary boinc_finish() run to completion
situation only. Medium speed XP 32 bit host estimated at 2-6 weeks
between needing a reboot due to low kernel resources.  One actual host
under observation exhibited the 'reboot to clear'  issue after 7 weeks,

The API will need further attention to what happens to the timer thread
handle and resources under various unusual exit conditions, perhaps
moving cleanup  to boinc_exit(), however since the timer thread calls
boinc_exit(), that's problematic. Cases of client shutdown, abort and
error exits, will need investigation, along with API review for the
dynamic C Runtime  with respect to handles.;  A Test Protocol should be
devised once applicable other areas are identified.

See MSDN documentation for _beignthreadex() and CreateThread() for
Visual Studio 2008/2010/2013, for details.

### DIFF
--- a/api/boinc_api.cpp
+++ b/api/boinc_api.cpp
@@ -190,6 +190,9 @@ static HANDLE hSharedMem;
 HANDLE worker_thread_handle;
     // used to suspend worker thread, and to measure its CPU time
 DWORD timer_thread_id;
+#if !(defined(_MT) && defined(_DLL))
+	HANDLE timer_thread_handle;
+#endif
 #else
 static volatile bool worker_thread_exit_flag = false;
 static volatile int worker_thread_exit_status;
@@ -697,7 +700,11 @@ int boinc_finish_message(int status, const char* msg, bool is_notice) {
     );
     finishing = true;
     boinc_sleep(2.0);   // let the timer thread send final messages
-    boinc_disable_timer_thread = true;     // then disable it
+	boinc_disable_timer_thread = true;     // then disable it
+	#if !(defined(_MT) && defined(_DLL)) // Static C-Runtime special care,
+		WaitForSingleObject( timer_thread_handle, INFINITE ); // Wait for the timer thread to exit, so we can free the handle
+		CloseHandle( timer_thread_handle );
+	#endif
 
     if (options.main_program) {
         FILE* f = fopen(BOINC_FINISH_CALLED_FILE, "w");
@@ -1240,7 +1247,13 @@ DWORD WINAPI timer_thread(void *) {
         if (!boinc_status.suspended) {
             nrunning_ticks++;
         }
+		#if !(defined(_MT) && defined(_DLL))
+			if (boinc_disable_timer_thread) break; // For static C-Runtime, we need to _endthreadex for resource cleanup
+		#endif
     }
+	#if !(defined(_MT) && defined(_DLL))  //Static C-Runtime
+		_endthreadex(0);
+	#endif
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


ChristianBeer:
This fails to build on Linux. Is this an XP only leak? I've never heard of users having to reboot with windows 7 because of low kernel ressources.

JasonGroothuis:
Hmm, this should be in _WIN32 blocks.  Thanks for the heads up.  I'll check where that is choking and update.
Vista onwards are indeed more tolerant of the leaks, Though still there of course, the memory management is more aggressive.

romw:
@JasonGroothuis What resource was being leaked?  Both process handles and thread handles are released when a process is terminated.  The last thing boinc_finish() does is terminate the application.

JasonGroothuis:
Very small amounts of static thread data blocks wrt to C - Runtime.  from https://support.microsoft.com/kb/104641?wa=wsignin1.0 
",Using CreateThread() in a program that uses the CRT (for example, links with LIBCMT.LIB) may cause a memory leak of about 70-80 bytes each time a thread is terminated. " .. The process termination approach is another more complex issue for another day, as it has its own issues with respect to multithreading and IO..that the boinc_critical_section() doesn't quite address for a number of application scenarios.  One symptom being blank/truncated stderr.txt.  In the interim, many of us use customised Boincapi for these reasons, and this relaltively trivial/specialised leak is a test of the procedure for getting hard fixes made.

romw:
Okay, but that memory is released when the process is terminated.

What leads us to the point that an XP machine starts reporting that the kernel has run out of a resource?  What resource does it run out of?

I don't see how this fixes the 7 weeks and you need to reboot the computer issue.

JasonGroothuis:
The MSDN documentation for CreateProcess() is vague on the issue:
"A thread in an executable that calls the C run-time library (CRT) should use the _beginthreadex and _endthreadex functions for thread management rather than CreateThread and ExitThread;"
https://msdn.microsoft.com/en-us/library/windows/desktop/ms682453%28v=vs.85%29.aspx

In practice (most noticeable on the XP scenario) it boils down to that the C Runtime is multithreaded, so allocates additional resources inside, that Windows API function won;t know about,  This  (small) cumulative leak is what manifests most  in the 2Gib kernel space under XP..  

As for the TerminateProcess() function used, it's asynchronous and initiates the cancellation/termination requests, though doesn't guarantee anything.  It is quite capable of corrupting DLL global data (among a number of other problems here), 
https://msdn.microsoft.com/en-us/library/windows/desktop/ms686714%28v=vs.85%29.aspx

romw:
> In practice (most noticeable on the XP scenario) it boils down to that the C Runtime is multithreaded,
> so allocates additional resources inside, that Windows API function won;t know about, This (small)
> cumulative leak is what manifests most in the 2Gib kernel space under XP.. 

Both _beginthreadex() and _endthreadex() end up calling CreateThread() and ExitThread() respectively.

As near as I can tell the memory leak that the first article is concerned about occurs if you call _beginthreadex() and terminate the thread with TerminateThread().  _beginthread()/_beginthreadex() makes a call to _calloc_crt to store internal information used by the CRT functions that require thread/process handles and the like. See: <VSInstallDir>/VC/crt/src/threadex.c

None of that is stored in the kernel address space.

As far as TerminateProcess() is concerned, it is more or less required until such time as the BOINC API becomes a little more symmetrical as far as the API contracts are concerned.  Science applications can create one or more threads for their use that the BOINC API doesn't know about.

JasonGroothuis:
Yes, correct on most points for simple threads that don;t call any CRT functions. The vagary in the MSDN text text can be read two ways... either the Process linking to the CRT, or the thread calling CRT functions.  It is this second case, as internal CRT calls in the thread can and do initiate helper threads, IO buffers, and crt reallocations.  

There's where we start to diverge from the trivial issue this addresses, and into the other problems.

Following onto the OS/Kernel/CRT/HelperThread issues, Leaving GPUs out for the moment, The missing part in the picture that applies to later Windows,  is the C-Runtime's inherent use of OS buffered IO, which is the default mode of operation for all of the underlying MFILE/MIOFILE implementation, and most definitely triggers portions in Kernel space (at minimum via drivers).  This delays IO for arbittrary OS determined periods from immediate to minutes (not just on XP, but even moreso on later implementation).  That OS delayed scheduling is driven by machine contention, which can be characteristically high on a crunching machine

Linking to the MS supplied COMMODE.OBJ goes some way to alleviating some of these delays (though bypasses desktop optimisations of course) , as does opening the streams in commit mode explicitly (which the api does not).

Those are not solutions, but workarounds/bandaids to the thread & exit handling strategies currently employed.  It's these set of workarounds (and others) that I need to address with particular projects, pending the need for reliable processing of tasks approaching or excedding 6 months CPU time.  Needless to say, truncated or missing stderrs and potential 7 week to clear issues then become forefront.

JasonGroothuis:
Closed for further work

PR #1364, opened 2015-06-10
Fix server_status.php after recent rewrite

tmannerm:
These bring back features from the old version and also fix showing color for daemon status


davidpanderson:
I merged the server_status.php changes, except the row coloring; thanks.
I wasn't seeing any problem w/ the row coloring.

PR #1367, opened 2015-07-07
vboxwrapper ncpus -> nthreads bug

marius311:
Typo where vboxwrapper was accepting `--ncpus` instead of `--nthreads` like it should.


marius311:
Actually.. it's not clear this is a bug. Looks like there's a move to deprecate `--nthreads` in favor APP_INIT_DATA.ncpus, but I'm not sure what the desired behavior for vboxwrapper should be. At the very least there's an internal inconsistency between the code and the usage description of vboxwrapper in the comment at the top of the file.

Issue #1128, opened 2015-02-04
Request: Ability to change the default priority of process


romw:
**Reported by DanielHendrycks on 15 Jun 41891870 06:57 UTC**
I request the ability to change the priority of a task's process. I want the setting, so when I get a new task, the task is automatically set to the specified priority level. Currently, the task's priority is ''low''. AFAICT, there is no preference to change it; instead, I must open the task manager and change the priority, manually, every time. Changing the priority should be an option because some want more crunching done quicker, without having to open the task manager every startup.

(F@H has this option. This shows it is plausibly a feature used by more than one person.)

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1155

AenBleidd:
Fixed

PR #1394, opened 2015-09-28
Fix issue #1128

AenBleidd:


romw:
David, can you review?

davidpanderson:
Thanks!  I shrank the code a bit and changed the numbering the low number = low priority (it's easier to explain that way).  I documented it here:
http://boinc.berkeley.edu/wiki/Client_configuration#Options

PR #1406, opened 2015-10-14
Web: rename recaptcha_get_html() because of a conflict with Drupal

ChristianBeer:
When also including an older recaptchalib.php this function gets redefined. Since it is only a wrapper function and only used within BOINC, renaming it is safe.

### DIFF
--- a/html/inc/account.inc
+++ b/html/inc/account.inc
@@ -78,7 +78,7 @@ function create_account_form($teamid, $next_url) {
     if ($publickey) {
         row2(
             "",
-            recaptcha_get_html($publickey)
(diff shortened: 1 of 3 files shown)


brevilo:
Thanks for adding the PR. Who is going (expected) to review and merge it? Is this process already established? I'm asking because getting this in is rather important to us.

brevilo:
:+1:

romw:
Historically, the project hasn't been big on process once somebody has commit access.  If you have commit access just make the changes.  If you are unsure about your changes send email to boinc_dev.  

If somebody has a problem with it, it should be sorted out in email or on boinc_dev.

PR #1418, opened 2015-10-22
Fix Coverity defects in server component

ChristianBeer:
This is for review by anyone and to make sure I don't break the build process.


JasonGroothuis:
The only real (general) caveats I've come across when scattergunning safe functions, that I've come across, aside from the obvious potential deadlocks not likely an issue with that code, are that you shift 'odd' behaviour reacting to external issues, to more local well defined faults.  I mean that's better because you can then isolate and handle the problems better as a robust subsystem, but that can and will expose flaws in systems it's connected to, and that can be frustrating at first.
I have seen the uninitialised  variables/members situations come up in not pretty ways, so only expect a reduction in weirdness there.

------------------------------------------------------------------------------------------------------
Jason Richard Groothuis 
bSc(compSci)

------------------------------------------------------------------------------------------------------


Date: Wed, 21 Oct 2015 23:58:22 -0700

ChristianBeer:
On 10/23/2015 12:37 AM, JasonGroothuis wrote:
> that can and will expose flaws in systems it's connected to, and that
> can be frustrating at first.
> I have seen the uninitialised variables/members situations come up in
> not pretty ways, so only expect a reduction in weirdness there.
That's exactly the reason. Even if it doesn't expose a flaw or a
security issue I still consider it a must have to eliminate the simple
coding flaws that can later creep up to be serious issues and are not so
easy to fix anymore. BOINC already has a lot of technical debt so I
thought it would be good to liquidate some of it.

When I reach out to people they always ask why they should trust BOINC
to be safe and secure software. With such a long defect list found by
static code analysis I'm not sure I can answer with full confidence anymore.

Regards
Christian

JasonGroothuis:
Thanks Christian.  I'm extremely encouraged by your approach to this, given that before the static analyses  assorted piecemeal efforts had resulted in little impact.   It's always daunting to see a possibly solid design fail in implementations, and vice versa, as well as a mixture of design an implementation problems.   For what it's worth I find the global api and infrastructure design component very solid, So I feel the  majority of implementation issues are just that.Apart from the static analysis, and questions about future flexibility,  the only  design level issues I tend to come across  are pretty abstract, in that there is a distinct 'control issue' at hand.  That's probably going to be pretty difficult to convey until such time as applications are granted some kindof rights as individual workers, but basically there is what I would consider murder going on.
---------------------------------------------------------------------------------------------------
Jason Richard Groothuis 
bSc(compSci)

------------------------------------------------------------------------------------------------------


Date: Thu, 22 Oct 2015 23:31:43 -0700

Issue #1423, opened 2015-10-31
When boinc gpu is paused and boinc is paused complete one context menu check is missing on Windows


informatorius:
This is how it looks when GPU is snoozed.
![boincsnoozegpu](https://cloud.githubusercontent.com/assets/6989214/10863147/fe0916f8-7fc3-11e5-94a4-8db1204536e1.png)

When boinc is snoozed complete additionaly to gpu snooze then GPU context menu check is missing.
![boincsnoozegpu complete](https://cloud.githubusercontent.com/assets/6989214/10863139/b9ef79c6-7fc3-11e5-860f-d18a30fe9bd3.png)

When complete snooze check is removed then gpu snooze check is shown again.
This looks inconsistent to me.

AenBleidd:
I suggest next scenario:
When "Snooze" is checked make "Snooze GPU" checked and disabled to avoid user enabling GPU only. I think it will be more understandable.

Rom Walton, David Anderson: What do you think?
I can fix this issue quite fast but I need your final decision.

Thanks

ChristianBeer:
It sounds very reasonable to make it clear that "Snooze" also includes "Snooze GPU" in the way @AenBleidd described it. @romw what do you think?

romw:
Sound fine to me.

AenBleidd:
Fixed in https://github.com/BOINC/boinc/pull/1425

informatorius:
Solved with Boinc 7.6.22

Issue #1421, opened 2015-10-27
When in boinc only GPU is paused then I wish a tray icon


informatorius:
When in boinc only GPU is paused then I wish a tray icon similar to the one wheb boinc is complete paused. e.g. the letter G could appear before the pause sign ||

AenBleidd:
I can fix this issue but I need an icon because unfortunately I'm very bad painter.

CharlieFenton:
On Oct 27, 2015, at 6:00 PM, Vitalii Koshura <notifications@github.com> wrote:
> I can fix this issue but I need an icon because unfortunately I'm very bad painter.

You will also need a Dock badge icon for the Mac.  See lines 395 - 459 in BOINCTaskBar.cpp.  Note that the badge is circular and must exactly fit with the mask at clientgui/res/macbadgemask.xpm.  The existing Mac Dock badges are clientgui/res/macsnoozebadge.xpm and clientgui/res/macdisconnectbadge.xpm.

informatorius:
We need someone who can decide the design of the icon. 
Then we can technically fit this in icon and xpm format.

PR #1425, opened 2015-11-01
Fix issue #1423

AenBleidd:


CharlieFenton:
Where is the new icon for the task bar? This will not build without the XPM file.  Also, you have not included the corresponding change and icon for the Mac Dock, as I explained in reply to your earlier posting <https://github.com/BOINC/boinc/issues/1421#issuecomment-151687660>.  

BOINC is cross-platform, so any changes to the UI must work for Macintosh and Linux as well as MS Windows.

AenBleidd:
@CharlieFenton, I'm sorry but this fix is not about the changing the icon. This one is fix only pop-up menu

CharlieFenton:
I apologize; I was confusing this with issue #1421.

However, I just tested this change and it also checks and disables the "Snooze GPU" menu item when CPU activity is set to "suspend" (RUN_MODE_NEVER), which I believe is not corrected nor what you intended.

AenBleidd:
@CharlieFenton, Hm, the main idea was to check and disable "Snooze GPU" when "Snooze CPU" is checked. Please correct me if I'm wrong and there is some another workflow which I don't know.
Thanks

ChristianBeer:
@AenBleidd can you please add a description to pull requests in the future? This would prevent such confusions. Referencing the issue is ok but please also write what you did or intended with the change.

AenBleidd:
@CharlieFenton, sure, I will. Thank you for advice.

CharlieFenton:
The logic here is a bit more complex than one might expect at first, due to the interaction with the Activity menu.

I have committed a fix for this problem, and added code to disable the Snooze GPU menu item in some additional cases where appropriate. 

When committing changes, please include in the commit message a description of the problem being solved or the purpose of the change, in addition to "Fix issue #xxxx" so that the commit log is self-explanatory without a need to go back to the original issue.  Also, please follow our convention of prefixing all commit messages with the BOINC component affected, such as "MGR:", "client", :server:". etc.  This will make it much easier to understand the change history.  Thank you.

PR #1426, opened 2015-11-02
Fix Issue 1423. Fix pop-up menu when CPU activity is set to "suspend"

AenBleidd:


CharlieFenton:
Hi Vitali,

I'm afraid your last (#1426) change still has problems.  If you snooze the GPU and then set the CPU Activity to Suspend, the check mark disappears from Snooze GPU but the menu item remains enabled.  Clicking on "Snooze GPU" does change the GPU status (as can be seen in the Activity menu) but the check mark does not appear next to "Snooze GPU" in the Task Bar menu until you Resume the CPU.

Please take a look at my latest commit to the GIT Head / Master branch to see how I fixed all of this.  It is important to test all changes very thoroughly before submitting a pull request, trying as many combinations of situations as possible.

Thank you!

AenBleidd:
@CharlieFenton, Thank you. I've compared your commit with mine and see the problems in my code. I understand now that there is much more functionality than I thought before. And this is the reason of bugs in my commit. Sorry for this. I'll try to make better commits in future.

PR #1431, opened 2015-11-07
Update BOINC-Manager.po "Boinc anhalten" => "Boinc pausieren"

informatorius:
Snooze should be translated to "pausieren" like in "Snooze GPU"=="GPU pausieren".
The current translation "anhalten" means "suspend" and is currently used when open Boinc manager and suspend boinc.
Suspend means stop until user resumes != Snooze means stop for 1 hour and auto resume.


romw:
Please use https://www.transifex.com/boinc/boinc/ to update translations.

Anything we commit through pull requests will just be overwritten by changes published in Transifex.

PR #1442, opened 2015-11-30
Web: use https urls everywhere

ChristianBeer:
Modern Browser complain if there are non-https elements in https websites. This makes sure that if the user visits via https all elements of the page are also retrieved using SSL. Furthermore it uses https urls in emails that are send to the user, if SECURE_URL_BASE is set in project/project.inc.
The only remaining places where URL_BASE is used is when it is used as master_url where it is important that it is in sync with what the scheduler uses.

### DIFF
--- a/html/inc/email.inc
+++ b/html/inc/email.inc
@@ -73,14 +73,14 @@ function send_auth_email($user) {
     $body = "This email was sent in response to a request on the ".PROJECT." web site.
 
 To log in to your ".PROJECT." account, visit:
-".URL_BASE."login_action.php?id=$user->id&t=$now&h=$x
+".secure_url_base()."login_action.php?id=$user->id&t=$now&h=$x
 (This link is valid for 1 day).
 After logging in, you can change your account's password or email address.
 ";
 
 $body .= "
 For further information and assistance with ".PROJECT.", visit
-".URL_BASE."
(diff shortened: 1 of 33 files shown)


davidpanderson:
Christian:
I'm not sure what you mean by "modern browsers complain...".
They complain if an HTTPS page contains non-HTTPS images.
They don't complain about non-HTTPS links.

Please review changes like this with me first.

-- D

ChristianBeer:
The rationale for using HTTPS links is that if a visitor uses SSL to visit the project in the first place, all links within the project should also be HTTPS because the visitor showed the intent to browse the project with encryption. If the visitor uses a HTTP url to visit in the first place, links are HTTP. Until now it was a mixture of both and while fixing the image problem, I also changed the remaining places where URL_BASE was used (except the bossa/bolt stuff) to make it consistent.

davidpanderson:
Christian:
The code used HTTPS for forms with passwords,
which is the only place where it matters.
Previously projects could use HTTPS for these forms, and HTTP for other pages.
With this change, if a project defines SECURE_URL_BASE,
then HTTPS is used for all pages, which is not necessarily what they want.

But regardless of this, the important point is that all major
changes need to be discussed with me.
If you're not sure if something is major, ask me.

-- D

PR #1447, opened 2015-12-07
Improve uotd

ChristianBeer:
UOTD selection now uses the DB layer and outputs the different stages of candidate selection so they end up in the logfile. Selecting a new UOTD can now be enforced (so it always happens at the same time or the current UOTD's profile is not approved by the project).

This will prevent the rare case where a project has not yet chosen a UOTD but there are profiles in the DB (a random profile would be shown).


JuhaSointusalo:
Does the new BoincProfile::delete() really work? It's calling
BoincDb::delete_aux() but the parameters don't look right.

And should it have the userid parameter anyway? It's an instance method so
you already have the id available.

ChristianBeer:
The profile table does not have an 'id'. The primary key is 'userid' because it is in a 1:1 relation with the user table.

JuhaSointusalo:
id or userid, either way the object should have enough data to allow it to
delete itself from database.

Most of the other delete() functions delete the object in question from the
database and don't take any parameters. Just thinking that BoincProfile
could follow the same pattern.

ChristianBeer:
Yes, of course. This should work too:
```
function delete() {
        $db = BoincDb::get();
        return $db->delete_aux("userid=$this->userid");
    }
```
You can't use $db->delete(...) because this has 'id=$this->id' hardwired.
I'm gona test and modify the branch later.

marius311:
I think this broke the UOTD for us. I just upgraded and am now getting this error:

```
January 13 2016 UTC: Starting
using default candidates_query
approved candidates for UOTD under UOTD_THRESHOLD
no new verified profile found, selecting old UOTD that was shown least recently
Database Error<br>
No screened users found
```

ChristianBeer:
I already pushed a fix for this: fd41262 if this is not working please let me know which MySQL version you are using.

marius311:
Hadn't seen that, thanks. That did indeed fix it.

PR #1449, opened 2015-12-09
Update filesys.cpp

DiablosOffens:
I tried to redirect the "slots" directory to another volume and came to the conclusion that it was not possible for some projects, because they use boinc_rename to move their results back to the project directory before they send it to the server. One of them was Virtual LHC@home. I compiled my own BOINC client for this, so that I can use my very fast but volatile SSD for such projects.
Now I want to share my findings and give you this simple patch. It don't have any negative side effects. There was only a flag missing which Windows needs if you want to move files or directories over volume boundaries.


davidpanderson:
looks good - thanks.

PR #1453, opened 2015-12-16
client (Linux): Reinstate XScreenSaver API-based idle detection

aggroskater:
Hello everyone,

I've been working on an idle detection bug with Debian:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=721298

After digging, it looks like idle detection with the XScreenSaver API was removed (bfae1032e5c1ac73f2d8d92f222293d8383a6cee), leaving idle detection up to checking the atime properties of files in /dev/input/\*. It seems that the X server doesn't update these files. Or at least, the atime is unaffected. Consequently, idle detection fails unless I happen to be working in a terminal (which does update the atime on the appropriate /dev/input/\* files).

The proposed pull request reinstates some of the code from the referenced commit (the DPMS code was not reinstated). It also modifies lib/prefs.{h,cpp}, as inclusion of the needed header for XSS ultimately pulls in other X-related headers where "Always" is also defined. Finally, I updated the configure.ac to set a "enable_xss" flag if the client is being compiled, and reinstated the other configure.ac checks in the referenced commit. These changes make a "git clone; ./_autosetup ; ./configure ; make ; ./client/./boinc_client" work as expected, with mouse movement and keyboard activity in non-terminals triggering boinc to suspend computation.

Let me know if you have any questions/comments. BOINC is a big project; I don't want to unintentionally break anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ addons:
     #- python-mysqldb
     - libfcgi-dev
     #- libcurl4-openssl-dev
-    #- libxss-dev
+    - libxss-dev
     - libnotify-dev
     - libxcb-util0-dev
     - libsqlite3-dev
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -41,6 +41,10 @@
 #include <cstring>
 #endif
 
+#if HAVE_XSS
+#include <X11/extensions/scrnsaver.h> //X-based idle detection
+#endif
+
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
@@ -1947,6 +1951,62 @@ bool interrupts_idle(time_t t) {
     }
     return last_irq < t;
 }
+
+#if HAVE_XSS
+// Ask the X server for user idle time (using XScreenSaver API)
+// Return true if the idle time exceeds idle_threshold.
+//
+bool xss_idle(long idle_threshold) {
+    static XScreenSaverInfo* xssInfo = NULL;
+    static Display* disp = NULL;
+    static bool error = false;
+        // some X call failed - always return not idle
+    
+    if (error) return false;
+
+    long idle_time = 0;
+    
+    if (disp == NULL) {
+        disp = XOpenDisplay(NULL);
+        // XOpenDisplay may return NULL if there is no running X
+        // or DISPLAY points to wrong/invalid display
+        //
+        if (disp == NULL) {
+            error = true;
+            //msg_printf(NULL, MSG_INFO, "XDisplay not found.");
+            return false;
+        }
+        int event_base_return, error_base_return;
+        xssInfo = XScreenSaverAllocInfo();
+        if (!XScreenSaverQueryExtension(
+            disp, &event_base_return, &error_base_return
+        )){
+            error = true;
+            //msg_printf(NULL, MSG_INFO, "XScreenSaverQueryExtension() failed.");
+            return false;
+        }
+    }
+
+    XScreenSaverQueryInfo(disp, DefaultRootWindow(disp), xssInfo);
+    idle_time = xssInfo->idle;
+
+    // convert from milliseconds to seconds
+    //
+    idle_time = idle_time / 1000;
+
+    //msg_printf(NULL, MSG_INFO, "XSS idle detection succeeded.");
+    //msg_printf(NULL, MSG_INFO, "idle threshold: %ld", idle_threshold);
+    //msg_printf(NULL, MSG_INFO, "idle_time: %ld", idle_time);
+
+    if ( idle_threshold < idle_time ) {
+        //msg_printf(NULL, MSG_INFO, "System is idle according to X.");
+    } else {
+        //msg_printf(NULL, MSG_INFO, "System is active according to X.");
+    }
+    return idle_threshold < idle_time;
+}
+#endif // HAVE_XSS
+
 #endif // LINUX_LIKE_SYSTEM
 
 bool HOST_INFO::users_idle(bool check_all_logins, double idle_time_to_run) {
@@ -1972,6 +2032,12 @@ bool HOST_INFO::users_idle(bool check_all_logins, double idle_time_to_run) {
         return false;
     }
 
+#if HAVE_XSS
+    if (!xss_idle((long)(idle_time_to_run * 60))) {
+        return false;
+    }
+#endif // HAVE_XSS
+
     // Lets at least check the dev entries which should be correct for
     // USB keyboards and mice.  If the linux kernel doc is correct it should
     // also work for bluetooth input devices as well.
@@ -1990,7 +2056,7 @@ bool HOST_INFO::users_idle(bool check_all_logins, double idle_time_to_run) {
     if (!device_idle(idle_time, "/dev/input/mice")) return false;
     if (!device_idle(idle_time, "/dev/kbd")) return false;
         // solaris
-#endif
+#endif // LINUX_LIKE_SYSTEM
     return true;
 }
 
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -47,7 +47,8 @@
 // lib/prefs.h definition in an enum.
 #undef Always
 #include <dirent.h> //for opening /tmp/.X11-unix/
-// (There is a DirScanner class in BOINC, but it doesn't do what we want)
+  // (There is a DirScanner class in BOINC, but it doesn't do what we want)
+#include "log_flags.h" // idle_detection_debug flag for verbose output
 #endif
 
 #include <cstdio>
@@ -1983,7 +1984,10 @@ const vector<string> X_display_values_initialize() {
   DIR *dp;
   struct dirent *dirp;
   if((dp = opendir(dir.c_str())) == NULL) {
-    msg_printf(NULL, MSG_INFO, "Error (%d) opening %s.", errno, dir.c_str());
+    if ( log_flags.idle_detection_debug ) {
+      msg_printf(NULL, MSG_INFO, 
+        "[idle_detection] Error (%d) opening %s.", errno, dir.c_str());
+    }
   }
 
   while ((dirp = readdir(dp)) != NULL) {
@@ -2010,6 +2014,12 @@ const vector<string> X_display_values_initialize() {
   // more than seven active, local Xservers. I'm sure they exist... somewhere.
   // But seven was the magic number for me).
   if ( display_values.size() == 0 ) {
+    if ( log_flags.idle_detection_debug ) {
+      msg_printf(NULL, MSG_INFO,
+        "[idle_detection] No DISPLAY values found in /tmp/.X11-unix/.");
+      msg_printf(NULL, MSG_INFO,
+        "[idle_detection] Using static DISPLAY list, :{0..6}.");
+    }
     display_values.push_back(":0");
     display_values.push_back(":1");
     display_values.push_back(":2");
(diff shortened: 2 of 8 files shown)


ChristianBeer:
FYI: The Travis VM does not have libxss installed so the test built ignored your code.
Can you uncomment the libxss-dev line in .travis.yml and add this commit to this PR? This will automatically trigger a new build.

aggroskater:
>Can you uncomment the libxss-dev line in .travis.yml and add this commit to this PR? 

Done and done.

ChristianBeer:
Thanks. Nice find btw.

JuhaSointusalo:
1. How does this behave when X in not available? Like in typical distro
repo install where client starts before X or is not able to connect to X.

2. You put the code inside #if LINUX_LIKE_SYSTEM . I think it would be
useful on FreeBSD and other unixes too.

(3. I would have simply undefined the Always macro instead of messing with
other files but maybe that's just me.)

CharlieFenton:
On Dec 16, 2015, at 2:33 PM, Juha Sointusalo <notifications@github.com> wrote:
> 2. You put the code inside #if LINUX_LIKE_SYSTEM . I think it would be
> useful on FreeBSD and other unixes too.

It would probably cause problems on the Macintosh.

Cheers,
--Charlie

aggroskater:
>1. How does this behave when X in not available? Like in typical distro repo install where client starts before X or is not able to connect to X.

Then this particular check would fail, returning "false" for determining whether the system is idle. Specifically, the initial call to connect to the X server (XOpenDisplay()) would fail, setting the static "error" variable to false, causing all future calls to xss_idle to also return false. In effect, BOINC would always think the machine was busy. If boinc-client was compiled with XSS support, then this would be a problem. If it wasn't, then this code wouldn't run.

This could be tweaked in a number of ways:

1. Don't make the function's variables (xssInfo, disp, error) static. This could at least handle the cases where the X server is started after the boinc-client. In theory, the X display would eventually be present and idle detection would function as normal. However, if the X server never appears, and the boinc-client is expecting it, then boinc would always see the machine as in use.

2. Change the return type of xss_idle to be more expressive than a simple bool. The "users_idle" function could then examine the return type to determine whether xss_idle has properly determined the idle state (according to X) or just errored out on some condition. If there's been an error, then the xss_idle check is disregarded.

> \2. You put the code inside #if LINUX_LIKE_SYSTEM . I think it would be useful on FreeBSD and other unixes too.

Truthfully, I wasn't quite sure whether BSD flavours would be included in LINUX_LIKE_SYSTEM's definition:

>#define LINUX_LIKE_SYSTEM (defined(__linux__) || defined(__GNU__) || defined(__GLIBC__)) && !defined(__HAIKU__)

Though of course, moving it outside of the LINUX_LIKE_SYSTEM guard section would still be fine. With regards to Macintosh problems, the "users_idle" function is actually defined twice. The first definition is in a guard for apple ("ifdef __APPLE__"). The definition this pull request edits is for the "else" version of users_idle intended to handle Linux/Unix? systems.

>(3. I would have simply undefined the Always macro instead of messing with other files but maybe that's just me.)

If that works, it'd certainly be cleaner than my "Always" -> "Always2" change. The definition, for me at least, is found in /usr/include/X11/X.h:

    /* Used in CreateWindow for backing-store hint */
    
    #define NotUseful               0
    #define WhenMapped              1
    #define Always                  2

[Milling around stackoverflow](https://stackoverflow.com/questions/1080635/others-library-define-naming-conflict), it seems like "#undef Always" right after the inclusion of \<X11/extensions/xscrnsaver.h\> should work. Locally, I've reverted my changes to lib/prefs.{h,cpp}, added the #undef right after the include in client/hostinfo_unix.cpp, and the client still compiles (and appears to function properly). I can push the work if preferred. It's cleaner than renaming variables in multiple files.

----------

Thanks for the comments!

JuhaSointusalo:
>> How does this behave when X in not available?

> Then this particular check would fail, returning "false" for determining whether the system is idle. In effect, BOINC would always think the machine was busy.

Ok. When a user installs BOINC using a GUI package manager, $DISPLAY and $XAUTHORITY will be set and when the install script starts the client it will inherit $DISPLAY and the idle detection will work just fine.

But, when the user reboots the machine the client will be started by init scripts. $DISPLAY and $XAUTHORITY are not available there so the idle detection doesn't work any more and the client will always think the machine is in use and stops processing tasks.

I'd like it the other way around, to run tasks always if the xss code fails. It's bad situation either way but at least it would be obvious to the user that this will need manual controlling.


> This could be tweaked in a number of ways:

I did some testing. You need to know the right display and need to have permission to access it. Getting the right display is easy, either you get it from $DISPLAY or just guess ":0" and you'd be right 99,9999% of time. Permission is the hard part. At least in Mint only the user who logged in can access the X server. Permissions can be given to other users with xhost but the user running xhost needs to already have permission to access X server. Even root doesn't have permission to access the server by default and can't run xhost without pulling off some really ugly tricks.

Since init scripts are run outside of X and as root the client installed from typical repo package would have no chance of working with xss.

I tried looking for a way to add the boinc user to the list of authorized users but didn't find anything good that would work with repo packages. Packages can install .conf fragments for X.Org server but there isn't any section to add authorized users. Some display managers have a similar way to add additional scripts to run when they start X server or set up X session but too many either didn't have or didn't document it clearly enough.

So, I would like to propose another method. Assuming different desktop environments have some agreement on a filesystem path where a package can drop a program to be autostarted when X session is started. Lets add a new program that opens the X server security for the boinc user and then notifies the client that an X server is available over here. xss code can then open the display it was notified of. Windows version has a boinctray.exe program that I believe has something to do with idle detection. Maybe that could be used as a starting point.

If you are interested in making those changes and get a green light I recommend to merge this in, in which ever form the powers that be decide, and make another pull request for the other changes. This way we'd have partially working idle detection while you continue the work.


A quick spin-off to GPU detection if none of you mind. Currently, as far as I have understood correctly, some GPU drivers need X to load the drivers and create device nodes and maybe some other reasons. With those drivers the client has a similar problem as the xss code, the client needs to be restarted from within X. With the new program the GPU detection could be changed to run every time the client is notified that there is a new X server available. I think this would solve most of the remaining GPU problems people have.


 >With regards to Macintosh problems, the "users_idle" function is actually defined twice. The first definition is in a guard for apple ("ifdef APPLE"). The definition this pull request edits is for the "else" version of users_idle intended to handle Linux/Unix? systems.

Also the code would still have HAVE_XSS around it. As long as that isn't defined in Xcode project there shouldn't be any problems.

>> (3. I would have simply undefined the Always macro instead of messing with other files but maybe that's just me.)

> I can push the work if preferred. It's cleaner than renaming variables in multiple files.

Well, I like pretty code so if you ask me...

LocutusOfBorg:
I have no decision poweer but +1 for me, and I would like to upload a patched version for Debian if possible
(This patch in this pull request is already on unstable)

aggroskater:
>But, when the user reboots the machine the client will be started by init scripts. $DISPLAY and $XAUTHORITY are not available there so the idle detection doesn't work any more and the client will always think the machine is in use and stops processing tasks.

>I'd like it the other way around, to run tasks always if the xss code fails. It's bad situation either way but at least it would be obvious to the user that this will need manual controlling.

Good points. I've updated the branch to reflect this. A message is sent to the event log that X-based idle detection is disabled if the calls fail.

>Well, I like pretty code so if you ask me...

And updated that as well.

>Permission is the hard part. At least in Mint only the user who logged in can access the X server. Permissions can be given to other users with xhost but the user running xhost needs to already have permission to access X server. Even root doesn't have permission to access the server by default and can't run xhost without pulling off some really ugly tricks.

Well crap. I hadn't even thought of that. I was just running my tweaked version locally as my own user, hence the connection to the X server worked. Indeed, running as the boinc user causes the xss_idle function to fail. Of course, running "xhost +SI:localuser:boinc" as my local user, and then running the boinc client as the "boinc" user works as expected. I believe Debian recently switched to having the X server run on a per-user basis, which might complicate a solution to this further from a packaging standpoint.

>I tried looking for a way to add the boinc user to the list of authorized users but didn't find anything good that would work with repo packages. Packages can install .conf fragments for X.Org server but there isn't any section to add authorized users. Some display managers have a similar way to add additional scripts to run when they start X server or set up X session but too many either didn't have or didn't document it clearly enough.

I've been looking around as well. I'm going to keep digging. In any case...

>Lets add a new program that opens the X server security for the boinc user and then notifies the client that an X server is available over here. xss code can then open the display it was notified of. 

I like that approach. The client could even just attempt to open the Xdisplay periodically, and not just have xss_idle permanently fail on the first problem with connecting to X.

romw:
Is this ready to be merged into master?

aggroskater:
Almost. I believe I've managed to tackle all of the issues at this point. I'm just setting up a log flag for debugging output if the user wants it. I don't want the output polluting the normal event log.

---------------------------------------------------------

Firstly, permitting access to DISPLAYs from the boinc user in a programmatic manner is possible. Add a suitable script to /etc/X11/Xsession.d/ for Debian-based systems (Fedora and others appear to use /etc/X11/xinit/xinitrc.d/, see below). According to "man 5 xession", this should work for "startx" and "xdm":

>/etc/X11/Xsession  is  a  Bourne  shell  (sh(1))  script  which is run when an X Window System session is begun by startx(1) or a display manager such as xdm(1).  (Some display managers only invoke Xsession when specifically directed to so by the user; see the documentation for your display manager to find out more.)

And [according to Debian at least](https://www.debian-administration.org/article/50/Running_applications_automatically_when_X_starts), gdm and kdm utilize this environment as well:

>Well the choosers gdm, kdm and xdm all support a standard startup system.
>These choosers use XSession to run programs either globally or per-user when you login.

All of this information is sourced from Debian (and is from the early 2000s) but simply dropping an appropriate file still works for me as we approach 2016, with gdm as my display manager under Debian testing. Subsequent logins as other users all have boinc as a permitted user. And targeting xdm, gdm, kdm, gets us most of the major display managers as well.

Examination of documentation for other distributions seems to indicate that they will source files from similar locations upon Xserver startup. 

  * Fedora: There's quite a bit of unrelated, but very similar, info about how Fedora has chosen display managers over the years, that tripped me up when trying to ascertain whether /etc/X11/Xsession.d/ existed or not:
    *  -> [1-ask.fedoraproject.org, still using /etc/X11/Xsession for display manager selection?](https://ask.fedoraproject.org/en/question/40425/xsession-handling-in-fedora/) 
    * -> [2-fedora devel mailing list, looks like no](http://www.spinics.net/lists/fedora-devel/msg192614.html) 
    * -> [3-roadmap since Fedora 18, definitely no, using systemd](https://fedoraproject.org/wiki/Features/DisplayManagerRework).
    * However, it appears that **/etc/X11/xinit/xinitrc.d/** is still the equivalent path on a Fedora-based system for global startup operations on Xservers. At least, [bug reports from Fedora 22](https://bugzilla.redhat.com/show_bug.cgi?id=1245755) still appear to be referencing it and indicating that it is in use in a similar manner to Debian's /etc/X11/Xsession.d/.
  * Arch: Appears [similar to Fedora](https://wiki.archlinux.org/index.php/Xinitrc) if their wiki is anything to go on.
  * Gentoo: Appears [similar to Fedora](http://www.gossamer-threads.com/lists/gentoo/desktop/190106). This mailing list is from 2009, but indicates towards the bottom that the same path is utilized as Fedora for the same purpose.
  * Ubuntu: Looks like [at least one user](https://askubuntu.com/questions/598628/etc-x11-xsession-lightdm) utilized the Debian /etc/X11/Xsession.d/ approach. But really, I'm operating on the assumption that most Debian derivatives will have this setup.

**Ultimately, this pull request doesn't break anything if the boinc-client can't access an (or any) Xserver.** In my examination, package managers should be able to implement appropriate "xhost" commands on Xserver startup, regardless of what display manager starts them (for the most part). If no Xserver is accessible, then boinc-client simply continues through other idle detection means.

---------------------------------------------------------

Secondly, I poured over X.org documentation, and after reading through "man Xserver", it looks like local Xservers will have sockets in a standard location, /tmp/.X11-unix/, with a standard naming scheme, Xn, where n is the DISPLAY number (from the man page's FILES section):

>        /tmp/.X11-unix/Xn             Unix domain socket for display number n

Consequently, we can interrogate Xservers as they come and go in a multi-seat/user environment.

---------------------------------------------------------

I'll push another commit with the log flag set up later. I spent last night reading over boinc's log_flag mechanism and I think I have it understood correctly. Might take me a day or so to make sure everything is working though.

aggroskater:
Added a log flag for use in cc_config.xml ("idle_detection_debug") that will output verbose logging information related to idle detection. I'd say this is ready for merge (and of course any other comments folks have).

davidpanderson:
Looks good!  Thanks very much.

PR #1456, opened 2015-12-20
Move duplicated zip / unzip code to a one place.

dimhotepus:
That's better to not duplicate it.

+ Check result of CreateDirectory and emit error to perror, if needed.


davidpanderson:
Thanks!

PR #1458, opened 2015-12-24
Cast as int instead of function

jkutner:
This prevents the error:

```
PHP Fatal error:  Call to undefined function int() in /tmp/boinc-project/bin/update_versions on line 375

      Fatal error: Call to undefined function int() in /tmp/boinc-project/bin/update_versions on line 375
```

As far as I can tell, `int()` is not a function. But I've seen this reported on Stackoverflow.com and some other places, so it must have been at one point?

There is an `intval()` function, which I can update this PR to use if desired.


davidpanderson:
Oops!  I didn't see the patch, and committed the same thing.  Thanks.

PR #1465, opened 2016-01-12
[1153] typo in file INSTALL

wmwong:
Fixed the typo in INSTALL when using the ./configure command. This caused an
error when trying to configure BOINC.


davidpanderson:
Thanks -- David

PR #1478, opened 2016-01-27
XSS Idle Detection Bugfix (XCloseDisplay)

aggroskater:
Previous merged PR: https://github.com/BOINC/boinc/pull/1453

Apologies. My initial code didn't close the connections to the X
displays after opening them. I had thought that falling out of scope
would close the connection automatically. However, thanks to a bug
report by Dan Merillat that I was able to replicate, I now know better:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=812835

This simple adjustment closes the connection to the display (if a
connection succeeded in the first place; if XOpenDisplay() failed, it
returns NULL and there is no connection). Testing on my end shows that
this resolves the issue of the boinc client piling up new connections to
X servers.


davidpanderson:
Thanks!

PR #1480, opened 2016-01-28
@advanced view: added connect-to-computer shortcut

thirschbuechler:
Added a connect-to-computer shortcut in the advanced view of boincmanager, as well as the corresponding menu entry.

Due to readabilty, i've chosen not to go for as little lines as possible (and put as little inside the conditional), as the `#ifdef` is done by the preprocessor anyway and not during runtime.

### DIFF
--- a/clientgui/AdvancedFrame.cpp
+++ b/clientgui/AdvancedFrame.cpp
@@ -341,7 +341,7 @@ bool CAdvancedFrame::CreateMenu() {
 
     menuFile->Append(
         ID_SELECTCOMPUTER, 
-        _("Select computer..."),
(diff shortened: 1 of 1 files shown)


CharlieFenton:
There are two problems with this patch:
[1] You set shortcuts [1] on (your added) line 739 without increasing the size of the array to 2 in the corresponding header file.
[2] There is no reason to add a new accelerator to the accelerator table since you set it in the menu creation on line 344.

thirschbuechler:
[2]: Wow. Embarrassingly overengineered, line 344 takes care of everything.
([1]: That one is a bit scary, i'm surprised my compiler didn't issue an error there.)

Thanks.

PR #1482, opened 2016-01-29
Add closing td tag on team members list

gchilders:
This simply adds a missing closing td tag on the team members list.


davidpanderson:
Thanks!

PR #1487, opened 2016-02-04
VBOX: fixed error code not being reported

marius311:
Prior to this if you dumped an error code in `completion_trigger_file` it was always reported to BOINC as exit status = 1. 

Should be straight forward but someone should double check this as I've neither compiled nor tested this fix.


romw:
Well, now that is embarrassing.

Issue #1187, opened 2015-02-04
boinc_client does not notice (or does not care) that "computer is in use"


romw:
**Reported by Bluefin Tuna on 3 Apr 42799116 09:47 UTC**
On Linux (Fedora 17 KDE spin), boinc_client (7.0.28 x86_64-pc-linux-gnu) is running as user "boinc", a user not used for anything else.

Settings as per "Computing Preferences (apply to all BOINC projects in which you participate)" are for Processor Usage:

```
Suspend work while computer is on battery power?    yes
Matters only for portable computers 	            yes
Suspend work while computer is in use?	            yes
Suspend GPU work while computer is in use?          yes
'In use' means mouse/keyboard activity in last	    5 minutes
Suspend work if no mouse/keyboard activity in last  --- minutes
Suspend work if CPU usage is above                  50%
Do work only between the hours of                   ---
Leave tasks in memory while suspended?              no
Switch between tasks every                          120 minutes
On multiprocessors, use at most	                    --- processors
On multiprocessors, use at most                     100% of the processors
Use at most                                         75% of CPU time 
```

The machine has 2 non-threaded CPUs. As per lscpu:

```
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
CPU(s):                2
On-line CPU(s) list:   0,1
Thread(s) per core:    1
Core(s) per socket:    2
Socket(s):             1
```

It also has a CUDA-capable graphics card, which is being used for BOINC projects.

Scenario:

- boinc runs while user is absent
- once the user activates the machine, boinc notices this and tunes down
  (the log shows nothing though, I think it should ... is there a "verbose"
  option?)
  ps then shows only the "boinc_client" running, not the tasks themselves
- however, after a quarter of an hour or so, the tasks start up again; computer interactivity slows noticeably. More amazing, there are actually 3 tasks running [2 x einstein@home + 1 x seti](e.g.) even though the machine has just 2 CPUs (maybe it's done differently on CUDA machines); here too logging the decision logic would be informative
- boinc knows the machine is in use though, the log says "Suspending network activity - computer is in use"
- to get the machine in hand, one has to "kill boinc"

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1215

romw:
**Commented by ChristianB on 12 Nov 42799609 05:40 UTC**
[...]

PR #1492, opened 2016-02-14
Change mouse activity detection on linux

formab:
Do nonblocking read on /dev/input/mice instead of checking file modification dates in /dev/input.

Related bugreports:
https://github.com/BOINC/boinc/issues/1187
https://bugs.launchpad.net/ubuntu/+source/boinc/+bug/1274456

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -24,6 +24,7 @@
 #include "cpp.h"
 #include "config.h"
 
+
 #if !defined(_WIN32) || defined(__CYGWIN32__)
 
 // Access to binary files in /proc filesystem doesn't work in the 64bit
@@ -55,6 +56,10 @@
 #include <cstdlib>
 #include <cstring>
 #include <ctime>
+
+
+#include "fcntl.h"
+
 
 #include <sys/param.h>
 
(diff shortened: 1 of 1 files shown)


romw:
This commit would remove checks against /dev/events and /dev/mouse.  How is this better?

romw:
Soory, my mistake.  /dev/input/events and /dev/input/mouse.

formab:
The file access times are not updated by mouse activity.
The /dev/input/mice is an aggregated input from /dev/input/mouse0 ... /dev/input/mouseN, /dev/input/event is a different interface for the same data.
https://www.kernel.org/doc/Documentation/input/input.txt

formab:
https://github.com/BOINC/boinc/pull/1453 seems to try to solve the same problem, but I think my commit is more in line with the original idea of checking the /dev/input. An works without X.

JuhaSointusalo:
What kind of system you have? This doesn't work everywhere.

Mint 17 = Ubuntu 14.04:

```
crw-r----- 1 root root 13, 64 tammi 30 20:37 event0
crw-r----- 1 root root 13, 65 tammi 30 20:37 event1
crw-r----- 1 root root 13, 66 tammi 30 20:37 event2
crw-r----- 1 root root 13, 67 tammi 30 20:37 event3
crw-r----- 1 root root 13, 63 tammi 30 20:37 mice
crw-r----- 1 root root 13, 32 tammi 30 20:37 mouse0
```

No chance of working.

Raspbian Wheezy:

```
crw-rw---T 1 root input 13, 64 tammi  1  1970 event0
crw-rw---T 1 root input 13, 63 tammi  1  1970 mice
```

Ok, here it could work, provided the user running BOINC is member of input group. The default user in Raspbian, pi, happens to be member of input group. So, if you run BOINC under that user account your code could work.

However, I installed BOINC from repository and the package created a user for BOINC. That user is not member of input. Also, that user should _not_ be added to input group. One of the event devices is your keyboard. If BOINC can read the device it would make it possible to turn BOINC into a keylogger.

I don't know why Raspbian has that kind of defaults. Maybe it makes hacking with hardware easier. I would expect other distros to choose more secure defaults.

formab:
I have ubuntu 15.10.

crw-rw----  1 root input 13, 63 febr  14 18:55 mice
crw-rw----  1 root input 13, 32 febr  14 18:55 mouse0

Yes I added boinc user to the input group, the installer can do it.
If you don’t allow boinc to access /dev/input/mice, you are just back where we started, it cant detect mouse activity, as the original code cant either. It is you choice.

JuhaSointusalo:
The thing with Linux idle detection is that there is no one method that works on every version of every distro. That's why the client has a ton of different methods. Presumably the old code does work on some version of some distro.

The security issue still stands. If the BOINC client is [compromised](https://googleonlinesecurity.blogspot.fi/2016/02/cve-2015-7547-glibc-getaddrinfo-stack.html) the only thing that prevents things from going even worse is the security your OS provides. Your code relies on punching a hole in the OS provided security.

For the record, I haven't commented on the security of the xss-idle code because I don't know X programming model well enough the tell off-hand what issues it may have. Allowing BOINC to access X server probably has similar issues as this code.

formab:
This code detects the activity on usb hid mouses, and call directly a kernel driver file. I would be surprised if it would work differently on different distros, maybe on different kernel versions.

The security part is independent of this code. If you opt on more security you don’t allow it access the mouse. But if you want to allow activity detection you add boinc user to the input group.

davidpanderson:
I'm closing this because there isn't consensus about security.  Is there a way to detect the presence of input without the ability to see keystrokes?

PR #1497, opened 2016-02-27
Fixed generate_svn_version.sh when building on detached head

whs:
If building from detached head (eg. checkout a tag), remote branch name detection will fail. In this case, use `origin` as hard coded default.

### DIFF
--- a/generate_svn_version.sh
+++ b/generate_svn_version.sh
@@ -13,7 +13,10 @@ elif [ -d .git ]; then
     GIT_LOG=`git log -n1 --pretty="format:%H"`
     HOST=`hostname`
     BRANCH=`git branch | sed -n 's/^\* *//p'`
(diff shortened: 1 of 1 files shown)


romw:
Thanks for your contribution.

----- Rom

ChristianBeer:
May I ask why you used a hard coded default? Wouldn't it be better to use the sha1 of the detached head?

whs:
I'm not sure how to code that, so I just went with the easiest way out.

ChristianBeer:
I looked at the script again (I only looked at the diff before). `origin` seems to be a sensible default for this specific operation since it is the default remote when doing a git clone. The only problem is if a build from detached head is done from a non-`origin` remote, the version.h information is misleading. Since those are mainly private builds I don't think it is a big problem. The sha1 in `GIT_REVISION` should still lead to the code used.

PR #1498, opened 2016-02-27
Add environment variable detection support to _autosetup

whs:
_autosetup claims that you can set `LIBTOOLIZE` to libtoolize's path, but this is not implemented. As homebrew use `glibtoolize` instead of `libtoolize`, this make it impossible to build on OS X.

This patch implements that support, and also changed mixed tab/space indentation to spaces.

### DIFF
--- a/_autosetup
+++ b/_autosetup
@@ -6,7 +6,7 @@
 ## 'echo -n' is not portable..
 case `echo "testing\c"; echo 1,2,3`,`echo -n testing; echo 1,2,3` in
   *c*,-n*) ECHO_N= ECHO_C='
-' ECHO_T='	' ;;
+' ECHO_T='      ' ;;
   *c*,*  ) ECHO_N=-n ECHO_C= ECHO_T= ;;
   *)       ECHO_N= ECHO_C='\c' ECHO_T= ;;
 esac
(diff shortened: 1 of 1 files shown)


CharlieFenton:
For instructions on the supported way to build BOINC on OS X, please see <http://boinc.berkeley.edu/trac/wiki/MacBuild>

whs:
The main idea of this patch is not to build on OS X, but to implement what is already mentioned in _autosetup: "If you have a newer version, set the environment variable 'LIBTOOLIZE' to its path" but is actually unimplemented.

romw:
Manatsawin, could you fix what TravisCI is complaining about?

whs:
Done :)

Seems that variable variables is bash-specific feature which dash doesn't implement. I only have tested this on bash-as-sh platforms.

romw:
Thanks again.

krytarowski:
This broke NetBSD. We don't ship with bash in `/bin/bash`

If we really need bash the most portable solution is `#!/usr/bin/env bash`

romw:
I've committed fix 27ae17e

krytarowski:
Thanks!

f8l:
In my opinion, depending on bash is not necessary.
@whs Please, test the patch at: https://wip.pkgsrc.org/cgi-bin/gitweb.cgi?p=pkgsrc-wip.git;a=blob_plain;f=boinc-git/patches/patch-__autosetup

Are `have_gmake` and `have_gm4` variables important (or used at all)?
If so, imagine someone with gmake at strange location runs _automake,
and then sets `MAKE` as instructed by the message.
This, I believe, short-circuits on `check_version make …` and skips setting `have_gmake`.
If the variables are not used, they should probably be removed,
as it would allow some simplifications.

By the way, why there are `echo >/dev/null` lines all over the place?

whs:
I'm not sure about the intention of this _autosetup script. I checked
autoreconf source and it doesn't seems to use gmake or gm4, only make and
m4. So, it seems that the checks in the script could possibly be outdated.

Is it possible that this script could be replaced by just autoreconf -i
without any checks, and the required versions moved to [the documentation](
https://boinc.berkeley.edu/trac/wiki/SoftwarePrereqsUnix)? Additionally, we
could check the exit status of autoreconf and direct users to the
documentation if it fails.

ChristianBeer:
The intention of _autosetup is to make sure that the correct tools are installed and can be used afterwards. Documentation is barely read by people trying to compile BOINC. Just telling people RTFM is also not the correct way. If we can find out what is wrong with the setup we should tell the user directly so he/she can fix it.

Did anyone try this on OSX with g{make|libtoolize|m4} tools?

whs:
Before submitting this PR I have tested this on OS X and Arch. Only glibtoolize needs to be prefixed, as libtool is another unrelated application.

Continued in #1500

PR #1499, opened 2016-02-28
remove spaces in pid_file

UweBeckert:
$d-pid_file has spaces in front of it and at the end, so the composed path is wrong.
Therefore trim it.


romw:
Thanks for the patch.

PR #1500, opened 2016-02-29
Use sh to detect applications in _autosetup

whs:
Continued from #1498 

This PR is made by applying @f8l's patch in the original PR. I also removed gmake detection as it is expected that the user should set MAKE to its path instead. The `echo >/dev/null` is also removed and replaced by inverting the if conditions instead.

From my testing, it seems that in my original PR user can specify application name in `$PATH` but this one an absolute path is required. As the error messages already specified that the variables should be set to its path I consider this behavior to be correct as well.

### DIFF
--- a/_autosetup
+++ b/_autosetup
@@ -1,18 +1,19 @@
-#!/usr/bin/env bash
+#!/bin/sh
 ## $Id$
 
 ## ---------- some portability checks/adjustments [stolen from configure] ----------
 ## 'echo -n' is not portable..
 case `echo "testing\c"; echo 1,2,3`,`echo -n testing; echo 1,2,3` in
   *c*,-n*) ECHO_N= ECHO_C='
-' ECHO_T='      ' ;;
+' ECHO_T='	' ;;
   *c*,*  ) ECHO_N=-n ECHO_C= ECHO_T= ;;
   *)       ECHO_N= ECHO_C='\c' ECHO_T= ;;
 esac
 ##----------
 
 ## ----------------------------------------------------------------------
-## Check that given command $1 has version >= $2.$3
+## Check that given command $1 has version >= $2
+## Full path $3 may be specified in order to skip command search.
 ## return 0 if ok, 1 too old or not found  (-> shell conventions).
 ## ----------------------------------------------------------------------
 check_version()
@@ -24,12 +25,7 @@ check_version()
     desired=`echo $2 | awk -F. '{print $1*100+$2}'`
     echo $ECHO_N "Checking version of '$1' >= $desired... $ECHO_C"
     name=$1
-    app_var_name=`echo $name | tr '[:lower:]-' '[:upper:]_'`
-    if [ ! -z ${!app_var_name+x} ]; then
-        name=${!app_var_name}
-    fi
-
-    fullpath=`type $name | awk '{ print $(NF) }'`;
+    fullpath=${3:-`command -v "$name"`};
     if [ -x "$fullpath" ]; then
         foundit=yes;
     fi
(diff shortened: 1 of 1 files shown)


romw:
Thanks for the submission

f8l:
@whs Can you tell me where it is stated that the user is supposed to set `MAKE` when `gmake` is available? The original code checked for `gmake` not without a reason. as the comment says:
> \#\# some sorry systems don't have proper GNU-make...

I would say that some systems don’t have `make`, so they ship GNU make instead (just like some ship `bash` instead of `sh`). Anyway, not checking for `gmake` broke the build on NetBSD (and probably some other systems which have `make`). I’ve made a [patch](https://wip.pkgsrc.org/cgi-bin/gitweb.cgi?p=pkgsrc-wip.git;a=blob_plain;f=boinc-git/patches/patch-__autosetup;h=374c202e7152ec0ef94c6da875e42fe4704c7ca3;hb=HEAD) to remedy this. I have to admit it’s not perfect, as it may check for `gmake` if `MAKE` doesn’t point to acceptable `make`, but I think it’s better to err on the side of inclusion, as it’s ultimately `configure`’s task to verify the environment. Nevertheless, improvements are welcome.

PS: @whs It may be a question of preference, but I advise against combining multiple unrelated changes into a single commit. In this one there are three: 1) getting rid of `bash`, 2) removing `gmake` detection, 3) getting rid of `echo > /dev/null`. If there were three separate commits, one could simply revert one of them if it introduces an undesirable change. Otherwise, one has to revert it manually by picking correct lines from a reverse patch in order not to lose other desirable changes.
Separate commits also improve readability. Take a look at #1498. Can you quickly tell which lines correspond to the change described in the commit message? (Well, you probably can, but not everyone is you. ;-)

Thank you for your time.

SETIguy:
I tend to agree.  The gmake check was there for a reason.  Not every unix
system is linux.

whs:
Previously you mentioned

>If so, imagine someone with gmake at strange location runs _automake,
>and then sets MAKE as instructed by the message.

I interpreted that that means if someone have gmake, they could just set MAKE and run _automake. I apologize if that is a misinterpretation

f8l:
It was a hypothetical situation described to illustrate how the #1498 version of the script may miss the documentation (if an error message can be regarded as one). It would be a problem if `have_gmake` was actually used, but you’ve checked that it is not (hence you removed it altogether).

Nevertheless, you are right that one could just set `MAKE` to `gmake`’s absolute path, but if I understood correctly, the script is meant to make it easier to cope with missing dependencies (my interpretation of @ChristianBeer’s words), and not to add some requirements (my supposition), which are not necessary. By ‘strange location’ I meant ‘outside of `$PATH`’, but I understand it could’ve been interpreted differently. Having GNU make as `gmake` when non-GNU `make` is available seems to be a common practice.

By the way, I’m not sure why (a new enough) GNU make is required. I’ve just successfully built using `make` supplied by NetBSD (with `gmake` deinstalled, and check in `_autosetup` disabled), which probably means the makefiles don’t depend on GNU-make-specific behaviour. Maybe they did in the past and the requirement survived… The reason why the check fails for `make` on my system is that it does not recognise the `--version` switch.

I’m starting to think that my patch fixed the build on NetBSD not because `gmake` was recognised, but because `make` could handle the build just fine and the patch just sidestepped the check in `_autosetup`. It is possible that my assumption
>The original code checked for gmake not without a reason.

is false. Your patch seems to be right (at least in NetBSD’s case. I’ll have to drop my change). In my opinion, it just reveals that `_autosetup` requires GNU make while it may be unnecessary. On the other hand, I know that some `make` versions are not compatible with each other, and requiring GNU make is probably one way of ensuring a compatible implementation (one which undersrands the dialect used in BOINC’s makefiles) is in place.

Are there any known versions of `make` which fail to build BOINC? Maybe there should be an option to skip the version check?

ChristianBeer:
Well, that brings us back to a question I asked months ago. What systems do we (BOINC) want to support? Do we really want to support systems that nobody uses or are so old there is no system around to test on? Do we really want to maintain software that is compatible with something that is not used nowadays?

In my opinion we should not and can not support and maintain anything we don't have the means to test on. Specify some reference systems that we can easily test on and drop anything that is older than that. This would also greatly simplify the search for a more modern build system or at least muck out the existing buildfile.

f8l:
I think dropping support for old systems is probably a good choice.
People who use them are usually knowledgeable enough to successfully
cope with missing dependencies and other problems which may arise.
Maybe it’s time to drop `_autosetup` entirely?

I can’t decide for you, but I’ll keep trying to build BOINC on NetBSD almost every day.
I still don’t know any project which supports that platform, but I intend to chenge that.
The esiest way to kickstart support for NetBSD would be
to send Linux binaries when *-*-netbsd version is requested,
as NetBSD can run Linux binaries in presence of the required libraries.

SETIguy:
There's a difference between "not supporting" and "making it impossible to
compile on anything except the four 'supported' operating systems with the
'supported' compiler".   Solaris 11.3 was released in October.  I'm sure
11.4 is under development.  I'm fairly sure its installed base is not
tiny.  It's just not a typical consumer OS.  But that account it isn't an
"old operating system".   It's not my job to tell BOINC Projects that they
can't support Solaris.

There's always been a drive among some BOINC developers to embrace every
platform specific extension and require newer OS versions and to not care
about anything but the platform they develop for.  Fine.  That's always
going to happen.  But there should be a rule: "Don't break things that
don't need to be broken.  Don't make extra work for other people."  Should
I start translating all the code to C++14 because it's better and newer and
not worry about people complaining when they can't compile it?

We can't even agree on what "supporting Windows means".  I would claim that
it means supporting command line MinGW library and applications builds.
Some would claim it means only VisualC++ 2010 and newer.

ChristianBeer:
I was mainly talking about the server related stuff not the Client. But the problem of testing is true for both. I agree we should not make it impossible to compile on more exotic systems. And seti@home as I presume has a lot of exotic systems attached. But it also makes developing more complicated if I can't use GNU extension on the server side just because I would have to provide a workaround or detection for some system the server is probably never installed on. I guess that Solaris nowadays supports things it didn't several years ago and some workarounds are not needed anymore and would make BOINC a little less error prone. Do you have a host OS breakdown from Seti? Where we can see how popular the other operating systems are?

I also agree with your "Don't break things for other people..." rule. The problem I have is that I don't know what other people are using and what other platforms I need to take care of. It would also be nice to know which C/C++ standard we should use.

I have seen no discussion about dropping Windows support or supporting specific compilers or not. It would be nice to have such a discussion and see what systems we really need to focus on. If it is not used anymore than it is not worth carrying around.

If the current projects don't care about that by not participating in such a discussion with the developers they really can't complain if support is dropped for a specific system only they use. Since the move to the Community governed model we also need to encourage projects to support the features they need for themselves by supplying specifications and requirements, so other developers can take that into account when changing things. Again, I can't check things I don't know about when I make changes to BOINC. If I break a thing I don't know about, I can't be blamed really.

PR #1504, opened 2016-03-13
Don't add example_app to project.xml unless requested

marius311:
Currently even if you don't give the `--test_app` option to `make_project`, you still get `example_app` in your `project.xml`, which means it goes in the database and is displayed under "Applications" on the website. This makes it so it only goes in `project.xml` if you actually requested it.


davidpanderson:
Thanks!

PR #1508, opened 2016-03-18
Wrapper improvements

ChristianBeer:
* add a feature to use the phrase "boinc_resolve(filename)" in the <command_line> option of job.xml this gets passed through boinc_resolve_filename(filename)
* fix macro substitution which was broken after a reshuffle of boinc_init_options()
* fix do_zip_outputs() by resolving the filename before checking for existence; it's possible it is a softlink
* add a feature to rename/move output files; use this when output files are large and softlinked so they are not copied within the project directory
* recognize the DEBUG flag and print debug output

### DIFF
--- a/samples/wrapper/wrapper.cpp
+++ b/samples/wrapper/wrapper.cpp
@@ -645,6 +645,32 @@ int TASK::run(int argct, char** argvt) {
         }
     }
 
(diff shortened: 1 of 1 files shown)


davidpanderson:
Please document this: http://boinc.berkeley.edu/trac/wiki/WrapperApp

PR #1509, opened 2016-03-19
Work around invalid XML in logs

skybon:
Ironically, should the user specify an invalid XML tag in configuration, BOINC will write a log message containing that tag unescaped and unclosed. Needless to say, it breaks XML parsing instantly. This patch auto-closes such tags in the log messages.

### DIFF
--- a/client/acct_mgr.cpp
+++ b/client/acct_mgr.cpp
@@ -409,7 +409,7 @@ int ACCT_MGR_OP::parse(FILE* f) {
         }
         if (log_flags.unparsed_xml) {
             msg_printf(NULL, MSG_INFO,
-                "[unparsed_xml] ACCT_MGR_OP::parse: unrecognized tag <%s>",
+                "[unparsed_xml] ACCT_MGR_OP::parse: unrecognized tag <%s/>",
                 xp.parsed_tag
             );
         }
--- a/client/log_flags.cpp
+++ b/client/log_flags.cpp
@@ -457,7 +457,7 @@ int CC_CONFIG::parse_options_client(XML_PARSER& xp) {
 
         msg_printf_notice(NULL, false,
             "http://boinc.berkeley.edu/manager_links.php?target=notice&controlid=config",
-            "%s: <%s>",
+            "%s: <%s/>",
             _("Unrecognized tag in cc_config.xml"),
             xp.parsed_tag
         );
@@ -512,7 +512,7 @@ int CC_CONFIG::parse_client(FILE* f) {
         if (xp.match_tag("log_flags/")) continue;
         msg_printf_notice(NULL, false,
             "http://boinc.berkeley.edu/manager_links.php?target=notice&controlid=config",
-            "%s: <%s>",
+            "%s: <%s/>",
             _("Unrecognized tag in cc_config.xml"),
             xp.parsed_tag
         );
--- a/lib/parse.cpp
+++ b/lib/parse.cpp
@@ -878,7 +878,7 @@ void XML_PARSER::skip_unexpected(
 
     if (verbose) {
         fprintf(stderr,
-            "%s: Unrecognized XML tag '<%s>' in %s; skipping\n",
+            "%s: Unrecognized XML tag '<%s/>' in %s; skipping\n",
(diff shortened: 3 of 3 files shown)


JuhaSointusalo:
Rather than changing individual messages it would be better to escape the message when it's written to XML. There may be other messages with less-than signs, with escaping you'd handle everything in just one place.

chausner:
Agree with JuhaSointusalo here. I am currently working on an alternative GUI application and am running into cases where BOINC responds with incorrect XML (e.g. message bodies are not escaped correctly). It is very ugly and fragile having to work around this.

ChristianBeer:
@skybon Some of the messages you changed are going into a plain textfile. Why are you parsing a plain textfile as XML? Closing them is only a bandaid to help your parser. I don't see a reason to escape those log messages that go into a plaintext file. The messages that are shown via the notices feature are a different kind but than `msg_printf_notice()` are the function that actually adds them to the notices.xml  needs to do the escaping and unescaping. Which is a godd idea.

@chausner Can you give some examples on what you mean by incorrect XML? It may be easier to fix than work around this. I think you mainly refer to the notices feature which needs to be fixed on Client and Server consistently.

chausner:
@ChristianBeer No, I was referring to the log messages whose bodies are not escaped properly when they contain special XML characters, for example, when setting the gui_rpc_debug flag. Another quirk exists with some of the polling routines, e.g. `<get_project_config>` returns malformed XML:
```XML
<boinc_gui_rpc_reply>
<project_config>
    <name>Einstein@Home</name>
...
```
I just got started with my app and have already run into those two issues. They are annoying to work around and I don't know how many more there are. Having read that this has apparently been a problem with the BOINC client for years, I propose to get this properly fixed now once and for all.

davidpanderson:
Seems to me the right thing to do is enclose messages in CDATA in GUI RPC replies.
There replies are nominally XML, so they should be parseable.

davidpanderson:
In the case of <get_project_config>, the <?xml header is coming from the project.  The client should trim this off before returning the project's reply in the RPC reply.

romw:
Thanks for the patch.

JuhaSointusalo:
@romw It's as if everybody (except maybe @skybon ) agreed that this patch is not the right way to handle the problem of un-escaped/encoded less-than signs appearing in GUI RPC replies. And yet you merge the patch?

skybon:
@JuhaSointusalo In all fairness, it was "everybody who spoke up", not just everybody. And in any case this patch does work around the problem for end users.

romw:
I am trying to catch up on some backlog.  I'll follow-up with a small change that will fix the problem for all regular parsers.  Using &gt; and &lt; in place of > and <.

If David wants to follow-up with the CDATA block after words that will work too.

davidpanderson:
I'll do the CDATA thing.
-- D

JuhaSointusalo:
> I'll follow-up with a small change that will fix the problem for all regular parsers. Using &gt; and &lt; in place of > and <.

Do you mean changing these messages? They go to the text logs as is, so &lt; and others aren't right there either. You really should have gone straight to the CDATA and skipped this.

@skybon Like I tried to say, you missed all the other messages with less-than signs. And I do wonder how adding an XML element in the middle of the message is helping. Now you need to add work-around in your parser/message handling for that.

davidpanderson:
I enclosed message bodies in CDATA,
and  eliminated the <?xml tag from the middle of get_project_config replies.
Let me know if any other XML errors in GUI RPC requests or replies.

ChristianBeer:
Since this is now obsolete can this be reverted please?

chausner:
> I enclosed message bodies in CDATA,
> and eliminated the <?xml tag from the middle of get_project_config replies.
> Let me know if any other XML errors in GUI RPC requests or replies.

Sorry for the late reply. I can confirm it's fixed for get_project_config but the issue is still there for lookup_account_poll and get_all_projects_list. Could someone please add the fix for these and check if there are any other RPC commands that pull XML directly from a server and fail to remove the XML declaration?

ChristianBeer:
@chausner Can you please open a separate issue for this?

PR #1514, opened 2016-03-24
manager fixes

JuhaSointusalo:
- fixes aborting and suspending tasks running on older clients 
- restores "run daemon" check


LocutusOfBorg:
Hi @JuhaSointusalo seems that you have to rebase your commits and fix the conflicts...

JuhaSointusalo:
@LocutusOfBorg Thanks, I hadn't noticed.

AenBleidd:
@JuhaSointusalo Could you please fix this pull request? Thanks

JuhaSointusalo:
@AenBleidd Thanks for reminding me. Should be fixed and ready to merge now.

AenBleidd:
@JuhaSointusalo Thank you!
@davidpanderson Could you please review this and merge? Thanks

davidpanderson:
I don't understand what this commit is trying to fix, or how; please explain.

Not sure why there's a new lookup_app_version() that doesn't take a platform arg.
A host could have 2 different app versions that have the same version, plan class etc.
but that differ in platform (e.g. Win x86 and Win x64)

JuhaSointusalo:
A new manager can't suspend, resume or abort tasks running on old client. The application column is also blank. I don't remember any more when the backward compatibility was broken, 7.6.20 something maybe. I just tested and the breakage is at least between 7.6.33 and 7.0.27.

Old clients don't include platform in \<result\>s and really old clients don't include plan_class in \<result\>s. The clients however do include everything in \<app_version\>s. That's why there needs to be different functions that look up the app_version with the information the client provided.

If there are multiple app_versions that differ only in platform the manager may indeed select wrong app_version. This is unavoidable because the client is not sending enough information and in any case it's not worse than what old client and old manager did together.

When the first manager that wasn't entirely backward compatible was released there was a few questions about it in forums but since then nothing. So I'm not sure how much backward compatibility between client and manager really matters to anyone.

AenBleidd:
@JuhaSointusalo Could you please remove indent changes from this PR and make it mergeable again? Thanks

JuhaSointusalo:
I'll sort it out once Charlie is done with the OSX changes.

AenBleidd:
@JuhaSointusalo Thank you

CharlieFenton:
On Mar 1, 2017, at 12:29 PM, Juha Sointusalo <notifications@github.com> wrote:
> I'll sort it out once Charlie is done with the OSX changes.


Thanks for waiting, Juha. I am done replacing deprecated APIs for OSX.

AenBleidd:
@JuhaSointusalo, could you please update this pull request? Thanks

Issue #1515, opened 2016-03-26
Compiler errors on openSUSE due to implicit conversion of strings


aaronpuchert:
The BOINC package for openSUSE is currently on version 7.2.42. When I tried to compile a more recent version (7.6.22), there were compiler errors of the sort

```
BOINCGUIApp.cpp:648:5: error: cannot convert ‘const wxString’ to ‘const char*’ in initialization [a few times...]
[...]
MainDocument.cpp:584:58: error: no matching function for call to ‘CBOINCBaseFrame::UpdateStatusText(const wxString&)’
             pFrame->UpdateStatusText(_("Starting client"));
                                                          ^
In file included from MainDocument.cpp:33:0:
BOINCBaseFrame.h:90:25: note: candidate: void CBOINCBaseFrame::UpdateStatusText(const wxChar*)
     inline void         UpdateStatusText( const wxChar* ){}
                         ^
BOINCBaseFrame.h:90:25: note:   no known conversion for argument 1 from ‘const wxString’ to ‘const wxChar* {aka const wchar_t*}’
MainDocument.cpp:586:72: error: no matching function for call to ‘CMainDocument::Connect(const wchar_t [10], int&, wxString&, int, int)’
                 Connect(wxT("localhost"), portNum, password, TRUE, TRUE);
                                                                        ^
In file included from MainDocument.cpp:32:0:
MainDocument.h:136:33: note: candidate: int CMainDocument::Connect(const wxChar*, int, const wxChar*, bool, bool)
     int                         Connect(
                                 ^
MainDocument.h:136:33: note:   no known conversion for argument 3 from ‘wxString’ to ‘const wxChar* {aka const wchar_t*}’
MainDocument.cpp:592:63: error: no matching function for call to ‘CBOINCBaseFrame::UpdateStatusText(const wxString&)’
             pFrame->UpdateStatusText(_("Connecting to client"));
                                                               ^
In file included from MainDocument.cpp:33:0:
BOINCBaseFrame.h:90:25: note: candidate: void CBOINCBaseFrame::UpdateStatusText(const wxChar*)
     inline void         UpdateStatusText( const wxChar* ){}
                         ^
BOINCBaseFrame.h:90:25: note:   no known conversion for argument 1 from ‘const wxString’ to ‘const wxChar* {aka const wchar_t*}’
[...]
ViewProjects.cpp:393:54: error: no matching function for call to ‘CAdvancedFrame::UpdateStatusText(const wxString&)’
[...]

PR #1524, opened 2016-04-12
Use wxString where necessary

aaronpuchert:
When building BOINC with `wxUSE_STL=1` there are compiler errors (#1515) because the implicit conversion to `const char*` and `const wxChar*` is deactivated. The solution was to replace `char*` or `wxChar*` by `wxString`. At some point in the stack, the `wxChar*` were implicitly converted to `wxString` again, so this seems like the obvious best way.

One advantage is that `wxString` owns the memory, so we know it always points to valid memory.

### DIFF
--- a/clientgui/BOINCBaseView.cpp
+++ b/clientgui/BOINCBaseView.cpp
@@ -913,11 +913,11 @@ void CBOINCBaseView::AppendColumn(int){
 }
 
 
-void CBOINCBaseView::append_to_status(wxString& existing, const wxChar* additional) {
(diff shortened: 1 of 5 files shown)


romw:
Thanks for the patch.

JuhaSointusalo:
@aaronpuchert About the command line option changes.

The old code had both options and switches. You changed everything to switches. Does the command line option parsing work correctly any more?

aaronpuchert:
Sorry, that was a mistake. Fixed by @romw in 8c11cad5b49952ec84309e1f5f9cd20fb41c82c0.

PR #1525, opened 2016-04-12
Prevent memory leak on reallocation failure

aaronpuchert:
If `realloc` fails, the old pointer is not freed. We should not overwrite the old pointer in this case, otherwise we have a memory leak.

We also have to free previously allocated memory in the case of reallocation failure.

### DIFF
--- a/lib/parse.cpp
+++ b/lib/parse.cpp
@@ -154,10 +154,11 @@ int copy_stream(FILE* in, FILE* out) {
 // append to a malloc'd string
 //
 int strcatdup(char*& p, char* buf) {
-    p = (char*)realloc(p, strlen(p) + strlen(buf)+1);
-    if (!p) {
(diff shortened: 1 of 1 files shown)


aaronpuchert:
@davidpanderson Would you have a look at this change?

This situation is unlikely to happen, but nevertheless it costs nothing to get it right.

ChristianBeer:
A comment above `strcatdup` that makes it clear that no `free(p)` is happening on failure and the caller has to free the pointer if `strcatdup` fails would be nice.

aaronpuchert:
Yes, that makes sense. I'll do that.

aaronpuchert:
The comment refers to the classic [levels of exception safety](https://en.wikipedia.org/wiki/Exception_safety). Is this sufficient?

PR #1531, opened 2016-04-26
lib: improve useless getaddrinfo message

jengelh:
Turn

	25-Apr-2016 17:10:31: getaddrinfo: No such file or directory

into

	25-Apr-2016 18:38:54: getaddrinfo: compute03:
	Name or service not known: No such file or directory

(From that, I was then able to conclude the hostname was missing
from /etc/hosts.)

### DIFF
--- a/lib/network.cpp
+++ b/lib/network.cpp
@@ -169,7 +169,7 @@ int resolve_hostname(const char* hostname, sockaddr_storage &ip_addr) {
     int retval = getaddrinfo(hostname, NULL, &hints, &res);
     if (retval) {
         char buf[256];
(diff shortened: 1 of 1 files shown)


ChristianBeer:
I find it more intuitive to have messages like this:
```25-Apr-2016 18:38:54: getaddrinfo(compute03) returned Name or service not known: No such file or directory```

In this case it would also be clever to get rid of the `perror()` call if `retval <> EAI_SYSTEM` because the error in `errno` is unrelated to the getaddrinfo call until `EAI_SYSTEM` is returned.

If you could change that in your PR it would be great. Thanks

ChristianBeer:
I submitted #1537 for discussion.

PR #1532, opened 2016-04-27
Fix #1530 (null pointer dereference)

### LINKED ISSUE #1530: boinc crashes when directory cannot be opened
```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff5e04dd9 in readdir64 () from /lib64/libc.so.6
Missing separate debuginfos, use: zypper install krb5-mini-debuginfo-1.12.1-30.1.x86_64 libX11-6-debuginfo-1.6.3-3.1.x86_64 libXau6-debuginfo-1.0.8-6.1.x86_64 libXext6-debuginfo-1.3.3-3.1.x86_64 libXss1-debuginfo-1.2.2-10.1.x86_64 libcom_err2-debuginfo-1.42.11-10.2.x86_64 libcurl4-debuginfo-7.37.0-7.1.x86_64 libgcc_s1-debuginfo-5.3.1+r233831-6.1.x86_64 libidn11-debuginfo-1.28-4.1.x86_64 libkeyutils1-debuginfo-1.5.9-4.3.x86_64 libldap-2_4-2-debuginfo-2.4.41-11.1.x86_64 libopenssl1_0_0-debuginfo-1.0.1i-12.1.x86_64 libpcre1-debuginfo-8.33-3.5.x86_64 libsasl2-3-debuginfo-2.1.26-8.1.x86_64 libselinux1-debuginfo-2.3-5.1.x86_64 libssh2-1-debuginfo-1.4.3-12.1.x86_64 libstdc++6-debuginfo-5.3.1+r233831-6.1.x86_64 libxcb1-debuginfo-1.11.1-1.1.x86_64 libz1-debuginfo-1.2.8-6.4.x86_64
(gdb) bt
#0  0x00007ffff5e04dd9 in readdir64 () from /lib64/libc.so.6
#1  0x00000000004a3ccb in X_display_values_initialize () at hostinfo_unix.cpp:1993
#2  0x00000000004a4589 in xss_idle (idle_threshold=idle_threshold@entry=60) at hostinfo_unix.cpp:2048
#3  0x00000000004a4b9f in HOST_INFO::users_idle (this=this@entry=0x70dea0 <gstate+16576>, 
    check_all_logins=<optimized out>, idle_time_to_run=<optimized out>) at hostinfo_unix.cpp:2173
#4  0x00000000004217fe in CLIENT_STATE::poll_slow_events (this=0x709de0 <gstate>) at client_state.cpp:930
#5  0x000000000047ff70 in boinc_main_loop () at main.cpp:375
#6  0x0000000000408049 in main (argc=1, argv=0x7fffffffea08) at main.cpp:542

(gdb) up ... up ... up
#1  0x00000000004a3ccb in X_display_values_initialize () at hostinfo_unix.cpp:1993
1993      while ((dirp = readdir(dp)) != NULL) {
(gdb) p dp
$1 = (DIR *) 0x0

1985      struct dirent *dirp;
1986      if((dp = opendir(dir.c_str())) == NULL) {
1987        if ( log_flags.idle_detection_debug ) {
1988          msg_printf(NULL, MSG_INFO, 
1989            "[idle_detection] Error (%d) opening %s.", errno, dir.c_str());
1990        }
1991      }
```
Great, so it checks for NULL but then passes it to readdir anyway :-/

aggroskater:


jengelh:
But if the directory does not even exist, there is no point in iterating over the `display_values` array - at which point we might just as well return from the function early on.

ChristianBeer:
On 29.04.2016 14:13, jengelh wrote:
>
> But if the directory does not even exist, there is no point in
> iterating over the |display_values| array - at which point we might
> just as well return from the function early on.
>

But it also may be possible that even if the directory does not exist,
an X server is running that we would not find if we return early.
Instead we would want to use the defaults and only if those don't work
we give up.

Regards
Christian

PR #1536, opened 2016-04-29
Added error codes

lfield:
Added three error codes for the VM use case.

### DIFF
--- a/html/inc/result.inc
+++ b/html/inc/result.inc
@@ -499,6 +499,9 @@ function result_error_mask_str($x){
 	case 202: return "EXIT_ABORTED_BY_PROJECT";
 	case 203: return "EXIT_ABORTED_VIA_GUI";
 	case 204: return "EXIT_UNKNOWN";
+	case 205: return "EXIT_INIT_FAILURE";
+	case 206: return "EXIT_NO_JOBS";
+	case 207: return "EXIT_JOB_FAILURE";
 	case -100: return "ERR_SELECT";
 	case -102: return "ERR_READ";
     case -103: return "ERR_WRITE";


ChristianBeer:
What do you mean with "VM use case"? At least one of the codes is already in use by vboxwrapper with another meaning. See https://github.com/BOINC/boinc/blob/master/lib/error_numbers.h#L41 you would need to start at 206 if you want to introduce new error codes for your own vboxwrapper.

lfield:
This is for use with the vboxwrapper. I will create a new pull request started at 206. Do I also need to update the header file?

ChristianBeer:
Yes, please. The header file is included by vboxwrapper anyway. Are you also planning to create a pull request for your vboxwrapper changes?

lfield:
I will do what ever is necessary if you give me some guidance. Do I need to do anything more than to update the following two files? 

https://github.com/BOINC/boinc/blob/master/lib/error_numbers.h
https://github.com/BOINC/boinc/blob/master/html/inc/result.inc

ChristianBeer:
Well, it depends on what you want to achieve. Changing error numbers by itself is useless if they are not used in the vboxwrapper code. Those files need to be changed in order to use the error codes and show some semi human readable text on the website. But vboxwrapper needs to report the error numbers for this to happen.

Btw: What does `EXIT_NO_JOBS` mean? Is this a job in the BOINC sense or in the CERN/vLHC sense?

lfield:
It is possible to return any arbitrary error code using completion trigger file feature in the vboxwrapper. The aim is show these codes as known errors in the Web portal. EXIT_NO_JOBS means that there are no more jobs to process if a single task processes multiple jobs by downloading from a parallel system.

ChristianBeer:
Ah I see. Then it would be nice to have both files changed and the commit message mentioning what project that was added for. I'm closing this so you can open a new PR.

lfield:
As results.inc does not currently have EXIT_OUT_OF_MEMORY 205, shall I add that too while I am updating these files?

ChristianBeer:
Yes please.

PR #1537, opened 2016-05-02
Lib: tweak error messages

ChristianBeer:
As described in PR #1531 the original error message does not provide enough information. This also fixes that no unrelated error message is shown.

### DIFF
--- a/lib/network.cpp
+++ b/lib/network.cpp
@@ -166,14 +166,14 @@ int resolve_hostname(const char* hostname, sockaddr_storage &ip_addr) {
     hints.ai_family = AF_INET;
     hints.ai_socktype = SOCK_STREAM;
     hints.ai_protocol = IPPROTO_TCP;
+    char buf[512];
+    snprintf(buf, sizeof(buf), "%s getaddrinfo(%s)", time_to_string(dtime()), hostname);
     int retval = getaddrinfo(hostname, NULL, &hints, &res);
     if (retval) {
-        char buf[512];
-        snprintf(buf, sizeof(buf), "%s: getaddrinfo(%s) returned %s", time_to_string(dtime()), hostname, gai_strerror(retval));
         if (retval == EAI_SYSTEM) {
             perror(buf);
         } else {
-            fprintf(stderr, "%s", buf);
+            fprintf(stderr, "%s: %s", buf, gai_strerror(retval));
         }
         return ERR_GETADDRINFO;
     }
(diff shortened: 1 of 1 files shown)


lib/network.cpp
```diff
@@ -223,7 +227,7 @@ int boinc_socket(int& fd, int protocol) {
     fd = (int)socket(protocol, SOCK_STREAM, 0);
     if (fd < 0) {
         char buf[256];
-        snprintf(buf, sizeof(buf), "%s: socket", time_to_string(dtime()));
+        snprintf(buf, sizeof(buf), "%s: socket failed", time_to_string(dtime()));
```
jengelh: If you change this line you might just as well print strerror(errno) too.

lib/network.cpp
```diff
@@ -168,9 +168,13 @@ int resolve_hostname(const char* hostname, sockaddr_storage &ip_addr) {
     hints.ai_protocol = IPPROTO_TCP;
     int retval = getaddrinfo(hostname, NULL, &hints, &res);
     if (retval) {
-        char buf[256];
-        snprintf(buf, sizeof(buf), "%s: getaddrinfo", time_to_string(dtime()));
-        perror(buf);
+        char buf[512];
+        snprintf(buf, sizeof(buf), "%s: getaddrinfo(%s) returned %s", time_to_string(dtime()), hostname, gai_strerror(retval));
```
jengelh: This could print "(date): getaddrinfo(host) returned The requested socket type is not supported". That looks a bit unfamiilar. Can't you just stick to the classic layout of "${function}: ${error cause}" rather than "${function} returned ${error cause}".

lib/network.cpp
ChristianBeer: This is done by `perror(buf)` which immediately prints `buf+strerror(errno)` to stderr. It seems the difference between `perror(buf)` and `fprintf(stderr, "%s: %s", buf, strerror(errno))` is that the former is thread-safe and the later not. I would rather stick with the thread-safe variant for now.

lib/network.cpp
ChristianBeer: I personally don't like the "chaining" using ':' like this: "${date}: ${function}: ${error cause}: {system error}". For me the ':' have different meanings. The first means "happend at", the second "failed because", the third "this also belongs to this error". In writing this I realize that there is a better solution (although it may not be thread-safe). I'll change the PR shortly.

PR #1538, opened 2016-05-02
Error codes

lfield:
Added three error codes for use with the completion trigger file mechanism in the vboxwrapper.

### DIFF
--- a/html/inc/result.inc
+++ b/html/inc/result.inc
@@ -499,6 +499,10 @@ function result_error_mask_str($x){
 	case 202: return "EXIT_ABORTED_BY_PROJECT";
 	case 203: return "EXIT_ABORTED_VIA_GUI";
 	case 204: return "EXIT_UNKNOWN";
+	case 205: return "EXIT_OUT_OF_MEMORY";
+	case 206: return "EXIT_INIT_FAILURE";	
+	case 207: return "EXIT_NO_SUB_TASKS";
+	case 208: return "EXIT_SUB_TASK_FAILURE";
 	case -100: return "ERR_SELECT";
 	case -102: return "ERR_READ";
     case -103: return "ERR_WRITE";
--- a/lib/error_numbers.h
+++ b/lib/error_numbers.h
@@ -39,6 +39,9 @@
 #define EXIT_ABORTED_VIA_GUI        203
 #define EXIT_UNKNOWN                204
 #define EXIT_OUT_OF_MEMORY          205
+#define EXIT_INIT_FAILURE           206
+#define EXIT_NO_SUB_TASKS           207
+#define EXIT_SUB_TASK_FAILURE       208
 
 // Function return values.
 // NOTE:  add new errors to the end of the list and don't change


ChristianBeer:
I know I am nit-picking but I think it is still not clear that the word "job" in the error description means something different than what BOINC understands as a job. In a year from now someone is going to wonder why those error codes where added or use the error codes in the BOINC code itself where the meaning of job is workunit which is different from what you said.

lfield:
Within a single Task, multiple computation jobs could be processed if another batch system is overlaid. We use the term job as it is distinct from Task and Work Unit and is well understood in this domain. 

https://en.wikipedia.org/wiki/High-throughput_computing
https://en.wikipedia.org/wiki/Batch_processing

ChristianBeer:
Yes, I know. The problem is that BOINC already has a different meaning of Job (a Workunit) and usually doesn't have another batch system overlaid. That is the special case for CERN VM applications where you run Jobs inside Tasks. But your Jobs are not to be mistaken for BOINC Jobs (Workunits). So I find it necessary to also have this distinction in the code.

lfield:
This is not necessarily restricted to VMs but could apply to any application which uses the late-binding  paradigm. We use the term job as it is distinct from Workunit, the term BOINC uses for a computational activity, so they will not be confused. If you would like to use a different term please make a suggestion.

ChristianBeer:
BOINC uses Job and Workunit synonymously, thus creating this specific problem. Like BOINC is using Result and Task synonymously. Except when in the wrapper context than a Task can have Subtasks, which are also called tasks, but have little to do with the actual Task send to the client. But it is clear from the context what is meant.

We don't need to rename the error descriptions. Just make it clear that they are for a specific context and this context is not the BOINC context of Jobs and Tasks.

marius311:
Just my 2 cents, adding these error codes makes no difference unless there's a corresponding change to vboxwrapper to actually return these errors, right? Wouldn't it make more sense for any specific project to just return whatever error code it wanted via completion_trigger_file, and then if it wanted to have it show a nice string description in the browser interface then just edit its own results.inc file? I don't totally see why it needs to be a BOINC-wide change.

ChristianBeer:
Yes, that is also possible. But this means the project needs to use error codes that are distinct from the BOINC ones and from real ones which opens another can of worms when upgrading.

I don't see a problem for including special application exit codes that can be of general use into BOINC and thus making it easier for a project to upgrade the BOINC code. There is precedence for that (look for `#ifdef SETIATHOME` or `#ifdef EINSTEIN_AT_HOME`)

lfield:
@ChristianBeer  For this discussion let's agree on using the term subtask. This abstract use case is for any task which uses late-binding to get subtasks and not necessary limited to containers (VM or otherwise).  In any case none of the error codes currently mention jobs or tasks. 

@Marius the vboxwrapper does provide a method via the completion trigger file to pass back error codes to the client. The important functionality that we need to implement is task back-off. In some cases we need to tell the client to go away and process tasks for another application/project, try again later. 

The error codes suggested are pretty generic; initialization failure, no sub-tasks, sub-task failure.

ChristianBeer:
Subtask sounds fine and would be consistent with what the wrapper does. It's also general enough to be included in the code.
Right now the Client does not react on those specific codes it just sends them along to the server. Are you planning on changing the Client implementation to increase backoff time for example? Right now a Client that is exclusively attached to your project and the app exits with no sub-tasks it will immediately contact the server to get a new task which would fail with the same reason possibly going into a vicious circle. You can probably achieve the same using the `temporary_exit_trigger_file` that wouldn't hit your server.

lfield:
Just to confirm the naming for the next pull request.
- 	EXIT_INIT_FAILURE
- 	EXIT_NO_SUB_TASKS
- 	EXIT_SUB_TASK_FAILURE
Any positive integer is classed as an error so the task will report an error and the server will back off. The error code can be seen in the Web portal and also the client should display this too.

ChristianBeer:
Looks good to me. Also just to be clear: the default BOINC server does not back off when a result with an error code is returned. It decreases the daily task counter that this host is allowed to get, it may take some tasks (and time) to get this to zero but in the meantime the host will get more Tasks from the Server (if available).

lfield:
Thanks. Which is the best forum to continue discussions? The message boards or the email lists

ChristianBeer:
That would be the boinc_dev mailinglist.

PR #1542, opened 2016-05-08
virtualbox icon should be const char*

smoe:
to avoid a warning when transforming into a string.

I remember to have contributed these kind of compile-time-optics patches a lot some long time ago :) The Debian build demons all publish their runs here:
https://buildd.debian.org/status/fetch.php?pkg=boinc&arch=amd64&ver=7.6.31%2Bdfsg-6&stamp=1460045578


aaronpuchert:
Technically, this doesn't conform with the [XPM standard](https://www.x.org/docs/XPM/xpm.pdf), see page 8. But that standard is not compatible with modern C/C++, so this is probably the way to go. Let's hope no one has a problem with that `const`.

PR #1547, opened 2016-05-20
Feature checkperm - checks permissions of apache user and presence of php modules

### LINKED ISSUE #1535: BOINC server/project does not check for permissions
Background:
I setup my own (test) BOINC project with the example app to help with the boinc-drupal integration. After I started the project, I went to connect myself with boinc-client.

Problem:  (puppet-eah-agent is the name of my host)

In the client stdoutdae.txt log file, the following occurred:
"Server can't open log file (../log_puppet-eah-agent/scheduler.log")

later on:
"Error reported by file upload server: can't open log file
'../log_puppet-eah-agent/file_upload_handler.log' (errno: 9)"

Solution:
The problem was that I didn't setup the permissions correctly in the log_puppet-eah-agent/ directory. Thus the scheduler and file upload handler daemons couldn't write to their respective files. Once I fixed
the permissions, everything worked.

Question/Issue:
I feel that it would be better/more illuminating if the BOINC project would report that there was a problem. Currently, the error message is in the boinc-client log. One might think, as I first did, that the problem was with the client setup, and not a problem with the daemons.

Another idea from Christian is another script that the BOINC admin runs, after make_project, which makes a series of checks on the BOINC project directory, checking for permissions, etc. This way the admin would be notified of any problems.

drshawnkwang:
I wrote a python script which is meant to be run after make_project. It checks whether the apache user can write to certain directories. It also checks whether certain php modules are installed.

If there are more directories to add or more php modules to check for, please let me know.

Fixes #1535

### DIFF
--- a/tools/check_project
+++ b/tools/check_project
@@ -1,0 +1,140 @@
+#!/usr/bin/env python
+
+# $Id$
+'''
+Checks project permissions, etc.
+
+This script is designed to run AFTER make_project is executed.
+'''
+
+import functools, argparse
+import grp, pwd
+import os, socket, sys
+
+#-----------------------------------------------------------
+# Functions
+def gethostname():
+   try:
+      return socket.getfqdn().split('.')[0]
+   except:
+      return 'localhost'
+#gethostname
+
+def check_user(user):
+   try:
+      mypwdb = pwd.getpwnam(user)
+   except KeyError as e:
+      print "ERROR: ", e
+      print "ERROR: Specified apache user (%s) not found, Aborting"%(user)
+      sys.exit(1)
+   #try
+# check_user
+
+def check_dir(dir):
+   if not os.path.isdir(dir):
+      print "ERREOR: project directory %s does not exst, Aborting"%(dir)
+      sys.exit(1)
+   #try
+   return dir
+#check_dir
+
+def determineapacheuser(user):
+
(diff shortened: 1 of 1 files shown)


ChristianBeer:
I wonder if we should check for mysqli explicitly since the mysql interface is deprecated since 5.5.0

drshawnkwang:
Obviously you know BOINC much much better than I. On line 160 of the script, there is a hardcoded python list of php modules. I added gd, mysql, and xml mostly based on my best guess on what is required. I'll add mysqli to that list.

If there are other modules you want added (or deleted) just let me know.

davidpanderson:
I don't see any references to check_project in the Wiki documentation.  Please add.

drshawnkwang:
Can do, but is the Trac wiki (http://boinc.berkeley.edu/trac/wiki/WikiStart) still the right place to add this type of documentation? Or is there a different Wiki site that should be made aware of?

davidpanderson:
This is where BOINC project-side features are documented.
-- D

drshawnkwang:
When I go to the Trac Wiki, I don't see a place I can create an account. Is there a link?

davidpanderson:
I'll need to add one manually.  Please send (via email) your desired user name and email addr.

PR #1558, opened 2016-06-11
Fix even more file permissions

GITNE:
* C/C++ files
* Drupal dummy readme file
* Xcode project file and *.nib resource files


romw:
Thanks.

PR #1561, opened 2016-06-17
Client: fix cpu detection for ARMv8

ChristianBeer:
Tested on an ODROID C2 which now returns "ARM" as p_vendor and "ODROID-C2" as p_model. Other kernels may report something like "AArch64 Processor rev 4 (aarch64)" as p_model.
Note: The aarch64 cpu reports the neon feature as "asimd".


ChristianBeer:
Please hold of with a merge for now. I try to increase the OS information too right now.

PR #1562, opened 2016-06-20
Client: improve CPU and OS detection on Linux

ChristianBeer:
- fix CPU detection for ARMv8: Tested on an ODROID C2 which now returns "ARM" as p_vendor and "ODROID-C2" as p_model. Other kernels may report something like "AArch64 Processor rev 4 (aarch64)" as p_model. Also added more technical info (if available) to p_model (at the end in square brackets) that resembles the Famliy/Model/Stepping info found for Intel/AMD CPUs.
- improve Linux OS detection: Gathers more information about the Linux OS using lsb_release, /etc/os-release or /etc/redhat-release (in that order) and adds that information to the os_name and os_version fields. This should work on most major distributions and will give a better overview on what Distribution is used by volunteers. Gathers information like this: 

```
os_name: Linux CentOS Linux
os_version:  CentOS Linux 7 (Core) [3.10.0-327.18.2.el7.x86_64]

os_name: Linux
os_version:  CentOS release 6.8 (Final) [2.6.32-642.1.1.el6.x86_64]

os_name: Linux Ubuntu
os_version:  Ubuntu 16.04 LTS [3.14.65-68]
product_name: ODROID-C2
p_model: ODROID-C2 [Impl 0x41 Variant 0x0 Part 0xd03 Rev 4][fp asimd crc32]

os_name: Linux Debian
os_version:  Debian GNU/Linux testing (stretch) [4.5.0-2-amd64]
```

Note that I added the new string in the front and moved the technical kernel info in square brackets at the end.

This is finished now. The result can be seen on my [Odroid C2](https://einsteinathome.org/host/12286420) which now shows some more information. I'm using the hardkernel kernel which unfortunately does not output a lot of useful information in `/proc/cpuinfo`.


JuhaSointusalo:
About the CPU part.

If p_model is reported as only "ODROID-C2" then anyone wanting to know what kind of processor the device has has to check ODROID's product page. "ARMv6/v7/v8" would give at least some kind of hint as to what the processor is. If you want the product name there you could make it something like "AArch64 Processor rev 4 (ODROID-C2)".

If I'm reading the code right then whichever of the "Processor", "CPU architecture" or "Hardware" comes first sets p_model. The order doesn't seem to be stable and differs from one kernel version to other. That means, update kernel and your processor changes. I don't think that's cool.

In the examples you show how product_name is filled from the "Hardware" string but I can't see that in the client code, only in the test code. Also, for some reason product_name is shown only for the owner of the device. Would be nice if it was made visible to others too.

Looking a bit more carefully at the /proc/cpuinfo there's "CPU implementer", "CPU architecture", "CPU variant", "CPU part" and "CPU revision". These are equivalent to x86's Family/Model/Stepping. These would make it easier to look up the exact CPU model instead of using some weirdo hardware name like "grouper".

The /proc/cpuinfo is very similar between ARM and AArch64. I think the code for them could be merged.

Putting it all together, how about like this, making it similar to x86 processors:

Raspberry Pi 1 B+:
`ARMv6-compatible processor rev 7 (v6l) [implementer 0x41 architecture 7 variant 0x0 part 0xb76 revision 7; BCM2708]`
ODROID-C2:
`AArch64 Processor rev 4 (aarch64) [implementer 0x41 architecture AArch64 variant 0x0 part 0xd03 revision 4; ODROID-C2]`

For reference, [this page](http://tessy.org/wiki/index.php?ARM) has a nice list of sample /proc/cpuinfos.

---
About the OS part.

I would like it better if distro and kernel parts were swapped. Like this:

`os_version: CentOS Linux 7 (Core) (3.10.0-327.18.2.el7.x86_64)`

That is displayed on the web pages and most of the time when looking at that information, like helping people, it's the distro and its version that is of interest so I would put that first. I have only needed to know the kernel version when I have had to deal with Intel GPU's on Haswell or later models.

The server would need changes in a few places if the parts were swapped. The kernel version is parsed for plan classes and there was one DB query that did something I don't know what with the kernel version.

Nitpick: The data from /etc/os-release is allowed to contain escaped characters which you don't handle.

ChristianBeer:
>If p_model is reported as only "ODROID-C2" then anyone wanting to know what kind of processor the device has has to check ODROID's product page. "ARMv6/v7/v8" would give at least some kind of hint as to what the processor is. If you want the product name there you could make it something like "AArch64 Processor rev 4 (ODROID-C2)".

>If I'm reading the code right then whichever of the "Processor", "CPU architecture" or "Hardware" comes first sets p_model. The order doesn't seem to be stable and differs from one kernel version to other. That means, update kernel and your processor changes. I don't think that's cool.

The idea indeed is that "Processor takes precedence over "CPU architecture" over "Hardware". In the examples I saw "Hardware" was always listed last but you are correct the order and availability of the info changes with each kernel but that is something we can't change. We always want to have the latest info maybe the new Kernel has more info than the last one. There could be an alternative fallback to "Hardware", a hard coded string that gets set as p_model if the kernel does not give us any information. Something like "unknown ARM 32bit processor" and "unknown ARM 64bit processor" and " (ODROID C2)" appended if "Hardware" is available. The idea of having the manufacturer in p_model is so that the user can look up the CPU info on the manufacturer webpage as a last resort because the kernel didn't give us enough information.

>In the examples you show how product_name is filled from the "Hardware" string but I can't see that in the client code, only in the test code. Also, for some reason product_name is shown only for the owner of the device. Would be nice if it was made visible to others too.

I checked again and indeed the commit that should have contained the product_name change does not contain it. I'll fix that later. My immediate goal is to gather the info that is available. Let's focus and what to show later but good point, I thought the product_name is visible publicly.

>Putting it all together, how about like this, making it similar to x86 processors:
>
>Raspberry Pi 1 B+:
>`ARMv6-compatible processor rev 7 (v6l) [implementer 0x41 architecture 7 variant 0x0 part 0xb76 revision 7; BCM2708]`
>ODROID-C2:
>`AArch64 Processor rev 4 (aarch64) [implementer 0x41 architecture AArch64 variant 0x0 part 0xd03 revision 4; ODROID-C2]`

Nice idea although I always think that the string now is too large to display and maybe the part between [ ] should be hidden by default and become part of a tooltip or so. Otherwise all the host tables and info pages look cluttered with technical strings. I'll gather this information too (if there is enough space in the string) and think about the display problem later.

>I would like it better if distro and kernel parts were swapped. Like this:
>`os_version: CentOS Linux 7 (Core) (3.10.0-327.18.2.el7.x86_64)`

>That is displayed on the web pages and most of the time when looking at that information, like helping people, it's the distro and its version that is of interest so I would put that first. I have only needed to know the kernel version when I have had to deal with Intel GPU's on Haswell or later models.

I would like that better too. Keeps the more technical things away from the user and any supporters can quickly see the real OS.

>The server would need changes in a few places if the parts were swapped. The kernel version is parsed for plan classes and there was one DB query that did something I don't know what with the kernel version.

For the same reason I just added the OS after the kernel version. Although the Android Client already adds the OS Name (e.g. Android N) to this field which can also be used in plan_classes already. So we just need to find this other DB query you mention and see if it can handle the changes.

>Nitpick: The data from /etc/os-release is allowed to contain escaped characters which you don't handle.

I didn't look into this much detail of the specs. I wonder if I could just call an unescape function after removing the quotes from left and right. That should take care of any escaped `"$", quotes, backslash, backtick`

Thanks for the review and the comment. I'm going to work on this again next week and improve it upon the discussed suggestions.

JuhaSointusalo:
> The idea indeed is that "Processor takes precedence over "CPU architecture" over "Hardware"...

Perhaps, before we discuss the implementation further, it would be best if you clarified a bit what information you want to get and if you would prefer some information more than other. There's at least CPU model, SoC model and board model.

Now that there are OpenCL capable GPUs bundled with ARM cores the SoC model might come handy sometimes. With Intel's processors one can tell from the CPU model what kind of GPU it includes but that doesn't work with ARM CPUs.

>>  `AArch64 Processor rev 4 (aarch64) [implementer 0x41 architecture AArch64 variant 0x0 part 0xd03 revision 4; ODROID-C2]`

> Nice idea although I always think that the string now is too large to display

I agree that it is rather large and there's much repetition. And the first part "AArch64 Processor rev 4 (aarch64)" is actually missing the CPU model.

To make it more useful it would be possible to decode the implementer, part and others into the actual CPU model name. Unlike Intel and AMD with their thousand different models there exists only a relatively small number of ARM compatible CPU cores. [ARM cores](https://en.wikipedia.org/wiki/List_of_ARM_microarchitectures)

And speaking of CPU models. The client puts "ARM" in p_vendor but not all ARM compatible CPUs are implemented by ARM Holdings.

More about the product_name. On Android GUI [sets it via RPC](https://github.com/BOINC/boinc/blob/b6508e0dc9d7096a34de93617196df14a9ec828b/client/gui_rpc_server_ops.cpp#L381), it's stored in client_state.xml and [read from there](https://github.com/BOINC/boinc/blob/master/client/client_state.cpp#L579) on client startup. The CPU info is gathered after the state file is read. I think your current implementation will end up repeating the hardware in product_name just before it's overwritten with info from GUI.

I believe product_name on Android is supposed to be the devices actual marketing name instead of some internal board name. If you are going to put the hardware to product_name on Android too I think it would be best to make it something like "marketing name (hardware)". Again keeping non-technical stuff first.

[Here's](https://www.androiddevice.info/) another site with even more cpuinfo examples. Such as [one](https://www.androiddevice.info/submission/36122/show) with no decoded info in cpuinfo, [one](https://www.androiddevice.info/submission/35088/show) with unknown hardware, [one](https://www.androiddevice.info/submission/34864/show) that's probably an ARM emulator.

> Although the Android Client already adds the OS Name (e.g. Android N) to this field which can also be used in plan_classes already. So we just need to find this other DB query you mention and see if it can handle the changes.

Unfortunately for Android the os_version format is "kernel version (Android version)". [os_version_num and android_version_num](https://github.com/BOINC/boinc/blob/56d59a778073dae2f4a7eef391da23e69d434194/sched/plan_class_spec.cpp#L39)

The DB query is in  [ops/failure_result_summary_by_host.php](https://github.com/BOINC/boinc/blob/8604e8e0bdda499694e1040d5f0b86fe5f5fe4e1/html/ops/failure_result_summary_by_host.php). I don't promise this query and plan classes are only places where the os_version is parsed. Check yourself too.

ChristianBeer:
Here is what I would like to have in the host fields of the DB:
This is mainly focused on Linux systems since there is a lack of information for those that does not apply to Windows, Mac or Android.
**p_vendor**: this nowadays represents the architecture of the CPU (currently Intel, AMD, ARM) NOT the manufacturer or implementer as in the case of ARM CPUs this would lead to problems with backward compatibility. I think this field is also used for homogeneous redundancy. It is also a nice way to get a census of available CPUs in a project per architecture.
**p_model**: this should represent the exact model of the CPU that is in a given host. This should be as accurate as possible as it can be used to target workunits to specific hosts (HR or others). This seems to be a problem on ARM based SoC as the info one gets through /proc/cpuinfo depends on the kernel used.  From the examples you posted this information is mainly in "Processor" or "model name" or "Processor name". But what if in neither of those there is a String that can be used to identify the CPU model? Again from the examples it seems to be common to use the "Hardware" field. This information is not very useful when p_model is used in plan_classes or HR but it would be better than nothing as it is now the case for most ARM64 CPUs that use a hardkernel.com kernel. One could also go for the "CPU implementer", "CPU architecture", "CPU variant", "CPU part" and "CPU revision" fields instead but those also can't be used in a plan_class easily and one has to look them up. Decoding them in the Client would mean we have to maintain a lookup table and release a new Client for every new ARM core. I don't like this idea.
**os_name**: This should be the name of the OS (without a version) including the Distribution name. Currently this is only "Linux" or "Android" without a way to further partition. Whereas for Windows this field also contains the high level version information. This field is mainly for troubleshooting issues. When looking up failures in the DB I want to see what Distribution is used or if this happens on this distribution only.
**os_version**: This should contain the most exact version number for the OS. For Linux based systems this includes the kernel version as well as the Distribution version. I would not change the order out of legacy reasons to not break queries that depend on the kernel version first. The extra information is also very useful when troubleshooting just to know if the system is running an experimental Debian that could explain a higher error rate then a stable Debian.
**product_name**: The Android Client already reports the product in this field. Since on Linux we get the same information from /proc/cpuinfo "Hardware" I thought it is a good idea to have this information also available. Whether this is the marketing name or internal board name I don't care. This can also be handy for a census of ARM based Linux hosts where a project might want to know if they should develop a more efficient app for a specific SoC board. I didn't change anything regarding how this value is gathered on Android so there it will always stay the marketing name.

rickybobpat:
Hi.  Tried the BOINC client yesterday on my Raspeberry Pi 3, and the client indicated unknown architecture.  Looking forward to this support!.

romw:
Is this ready to go?

ChristianBeer:
No, I still need to fix it.

ChristianBeer:
@BOINC This affects not only ARM 64bit systems but also all Linux systems as I add more information about the installed operating systems. The host.p_model and host.product_name changes are for AARCH64 only. I can't test if this also affects Android on AARCH64. Please review and comment as I want to give this to the Alpha testers asap.

JuhaSointusalo:
I'd have AArch32 and AArch64 share the same code but I'm not going to fight over it. (I swapped arm and aarch64 code, tested it on my RPi1 and it worked just fine.)
You need to initialise the char arrays before using them.
You have code to print architecture but you don't capture it anywhere.

Are you going to adjust the kernel version parsing in plan classes and elsewhere some other time?

Other than that, ok I guess.

ChristianBeer:
Another user also asked me if this applies to ARM 32bit. From the cpuinfo samples I saw it should work. I'm looking into this and also run some tests on a RPi1 but probably commit this later.
The char array's were a late night addition and I deleted the architecture capturing instead of moving it because of that too. It's fixed now.

I'll look into the plan_class parsing next. There should only be some minor changes so that the code does recognize the new square bracket content and parses the kernel version from there.

ChristianBeer:
I checked the current plan_class implementation which is using regular expressions which means a project using an os_version plan_class just needs to adjust the regex a little bit. I'll inform projects through the mailing list.

The other usage of `os_version` in [/html/ops/failure_result_summary_by_host.php#L44](https://github.com/BOINC/boinc/blob/8604e8e0bdda499694e1040d5f0b86fe5f5fe4e1/html/ops/failure_result_summary_by_host.php#L44) is more problematic but currently broken. It looks for a "-" within the first six characters of `os_version`. The current kernel version on my C2 is `3.14.77-81` and if I'm counting right the dash is number eight. So the query will fall back to showing the complete os_version currently even without my changes. So I'm going to extend the case either way to fix for longer kernel versions and to allow the query to output the kernel version if they are in square brackets.

AenBleidd:
@ChristianBeer, Sorry about disturbing you. Do you finish this pull request? There was no commits up to 3 months. I think this PR can fix several existing exiting issues.
Thanks
P.S. Sorry for the audacity

ChristianBeer:
@AenBleidd There was some discussion of adding the glibc version (when available) to `os_version` because that is a very relevant information for projects. I'll need to do some more testing for this but it shouldn't block this PR. I already distributed a Client version containing this PR and it is running for some time without problems.

@davidpanderson This should be reviewed and merged before the next release. There is another change that normalizes the `p_model` field (on all platforms) which I'm going to add when this is merged.

davidpanderson:
Please verify that these changes don't break the HR code:
https://github.com/BOINC/boinc/blob/master/sched/hr.cpp

Also, for the string manipulation code please have a version for std::string or char[], but not both.

ChristianBeer:
The HR code uses:
* `host.os_name` and looks if `Linux`, `Windows`, `Darwin`, `FreeBSD` or `Android` is a part of the string (case insensitive). If there is a Linux Distribution that uses one of these words as part of its name that could be a potential problem. The only one I could find was "Android-x86" (Remix OS). But since `strcasestr()` returns the first occurrence we are still save because `host.os_name` always starts with `Linux`. But this should only be the case if "Android-x86" or "Remix OS" use the wrong compile flags since they both should report as `Android` in `host.os_name`.
* `host.p_vendor` and looks for `Intel`, `AMD`, `Macintosh` or `ARM`. Actually this PR fixes the hack that the coarse HR class has (but the fine HR class has not) to look in `host.p_model` if `host.p_vendor` is empty on aarch64.
* `host.p_model` but only for non ARM CPUs. So this PR effectively allows to have better HR distinction on ARM in the future.

ChristianBeer:
@davidpanderson Please don't push changes to other peoples branches without asking them before. For example I don't know how you fixed the apparent merge conflict introduced by the unnecessary merge of #1751 since it is not documented in your commit message of the merge. In this case I was able to compare it with the introduced changes and I'm quite sure nothing is broken.

I fixed the duplicate function and resolved the merge conflict introduced by a788dba. From my point of view this is ready to be merged again.

davidpanderson:
hostinfo_unix.cpp is getting too big.
We should factor the Linux code into a separate file.

ChristianBeer:
> Huh? AFAIK I didn't change your branch.

The git log tells something different: 6733529d29062e505c6a14fb2a3d1a3bf17545bc
This commit shows that you merged branch `master` into branch `cpu_infos_aarch64` and pushed this to github.

brevilo:
Can someone confirm that this fixes #1597 and ?

ChristianBeer:
This PR was not intended to fix Android specific issues but it might still have an influence on #1597 which needs to be tested with a freshly build APK on Android 7.

brevilo:
Ok

sunk818:
Hi, is a configuration like below supported? It is Linux Ubuntu 16.0.4 LTS running on ARMv67 processors.

I was maybe hoping to compile my own BOINC client somehow.

Linux swc1paris01 4.4.127-mainline-rev1 #1 SMP Sun Apr 8 10:33:17 UTC 2018 armv7l armv7l armv7l GNU/Linux

[World Community Grid] This project doesn't support computers of type arm-unknown-linux-gnueabihf

Architecture:          armv7l
Byte Order:            Little Endian
CPU(s):                4
On-line CPU(s) list:   0-3
Thread(s) per core:    1
Core(s) per socket:    4
Socket(s):             1
Model name:            ARMv7 Processor rev 2 (v7l)
CPU max MHz:           1333.0000
CPU min MHz:           666.5000

TheAspens:
@sunk818 - World Community Grid does not provide research applications that run on Linux on ARM processors.  If you go to https://boinc.berkeley.edu/projects.php and look for projects that list the icon with the Penguin and ARM listed (this icon: https://boinc.berkeley.edu/images/linux_arm.png).

By the way - this is better discussed on the BOINC user forums here: https://boinc.berkeley.edu/dev/forum_index.php  You should be able to connect with other people there that have run the software on Raspberry Pi's for example.

Issue #1573, opened 2016-07-17
VBox proposal: specify boot order


marius311:
Currently the boot order that vboxwrapper sets up is 1) disk 2) dvd. I'm in a situation where I'd like to boot from an ISO but also use `enable_cache_disk`. Currently this is impossible since, if the cache disk is formatted, the VM will try to boot from it and hang since there's no boot sector (or whatever its called). 

I'd like to propose adding an option to `vbox_job.xml` called `boot_device` which will be either `disk` or `dvd` to specify which goes in the first boot slot. The default will be `disk` so as not to break backwards compatibility. I'm happy to code this up myself and submit a PR, I just want to get feedback if there is any first.

ChristianBeer:
This is a very good use-case. This could additionally be a command line option to vboxwrapper so it is a little bit more flexible.

PR #1575, opened 2016-07-19
VBOX: added boot_iso option

marius311:
Added an option `<boot_iso/>` to `vbox_job.xml` which puts the ISO first in the boot order in the case where you are attaching both a disk and an ISO. The default remains to boot from disk as before. Resolves https://github.com/BOINC/boinc/issues/1573. 

I did not add a command line option (I could though), since it seems like the options which have made it as command line options are ones which would need to be controlled on a per-job (rather than per-appversion) basis, and this doesn't seem to me to be one of them.


davidpanderson:
Please add here:
http://boinc.berkeley.edu/trac/wiki/VboxApps#TheVboxjobdescriptionfile
when you get a chance.

PR #1577, opened 2016-07-19
AndroidBOINC: Render time stamps according to current locale

GITNE:
Also, renders elapsed time of tasks with ```android.text.format.DateUtils.format()```


romw:
Thanks for the contribution

PR #1578, opened 2016-07-19
AndroidBOINC: Add basic Android TV support

GITNE:
The only thing I am unsure about is the [```android:screenOrientation```](https://developer.android.com/guide/topics/manifest/activity-element.html#screen) attribute on most of the activities. Why is portrait screen orientation enforced? I see no need for this. Please, can anyone explain why this is? Enforcing the portrait screen orientation on landscape-only devices, like Android TV and consoles, is really counter intuitive and causes problems in navigating the app. Hence, I would like to remove the portrait screen orientation enforcement from all activities, since the user or the system know best how to render layouts.


romw:
Thanks for the contribution

PR #1579, opened 2016-07-21
AndroidBOINC: Render numbers according to current locale

GITNE:
Although the implementation in this commit is functionally equivalent with the existing one, there is still one thing that needs further refinement. The [```projects_credits_header```](https://github.com/GITNE/BOINC/blob/localized_timestamps/android/BOINC/res/values/strings.xml#L226), [```projects_credits_host_header```](https://github.com/GITNE/BOINC/blob/localized_timestamps/android/BOINC/res/values/strings.xml#L227), and [```projects_credits_user_header```](https://github.com/GITNE/BOINC/blob/localized_timestamps/android/BOINC/res/values/strings.xml#L228) resource strings need to be reworked. ```projects_credits_header``` should be on a separate [```TextView```](https://developer.android.com/reference/android/widget/TextView.html), just like those in the task views, and then followed by either a ```projects_credits_host_header``` or ```projects_credits_user_header``` formatted string (instead of being hard coded). ```projects_credits_header``` should then be rendered bold per resource definition, not by content. Also, to assume that a space character ([```U+0020 SPACE```](https://en.wikipedia.org/wiki/Whitespace_character)) is the proper way to concatenate host and total credit strings for all locales is rather naive. I will push a fix to this change in the next commit.

On a side note; it is quite obvious that most of the Java code has been written by somebody who has spent most of his programming career writing C/C++ than Java code. :wink: A lot of things one would either not do in Java or do differently.


GITNE:
There you go, separate labels and content are much better! :happy: Please merge now.

Besides, I have realized that some translations are a total mess, like Slovenian and Hebrew. The Slovenian translation is a strange mix of mostly Russian and some Romanian. Hebrew is filled with Finnish. It should be best to clean those up, simply by removing them.

romw:
Thanks for the patch

PR #1583, opened 2016-07-24
AndroidBOINC: Android TV conformance and usability updates

GITNE:


ChristianBeer:
The Travis failure stems from f5c64abac07322fd644194a79d640c057e5fea9e which is now fixed in master and does not conflict with this PR.

romw:
Thanks for the patch

Issue #1569, opened 2016-07-11
Limit the number of tasks to 1 for a default install of the BOINC client


lfield:
By default the BOINC client will start one task per core. If this happens for our application, especially if there is 2GB per core or less, the machine can freeze. This is not the best first impression to give newcomers and so must be avoided. The decision on the amount of resources that a host can use for BOINC tasks and hence number of parallel tasks that the host can run, is a client side responsibility. If the client would like another task it will ask the server which may or may not assign one.

We  would like to ensure that by default only one task will be assigned and the volunteer has to perform a specific conscious action to increase this limit. Ideally this would be something that could be configurable by the BOINC GUI and the default would be 1 task.

AenBleidd:
Hello @lfield,

I think total memory limitation is better in this case because user can control the amount of memory he wants to be used by non-BOINC applications.

Best regards,
Vitalii

ChristianBeer:
There are already two options that control the number of in progress tasks per application/project.
One is a project setting: http://boinc.berkeley.edu/trac/wiki/ProjectOptions#Joblimitsadvanced
The other is a Client setting: http://boinc.berkeley.edu/wiki/client_configuration#Application_configuration
You would need to test which one has precedence if both are set. Developing a GUI for the Manager to manage an app_config.xml seems nice.

lfield:
After investigating the options, the memory limits do not work well for VM applications. The project settings will affect everyone and they can not be overridden by the volunteer. The client-side settings are good for the experienced volunteer but we need to place default values there for the novice. I think that if we could provide a default app_config.xml, this would solve our problem. We tried doing this by distributing it as a file for our app but it will overwrite changes made when we release a new application version. 

The fact that a GUI does exist, it should be possible to manage this file via the GUI.

GITNE:
[...]

PR #1594, opened 2016-07-29
Limiter

lfield:
Added a project preference to limit the number of tasks per host. The scheduler would need updating to understand the following XML tag.  #1569

`<limit_tasks>1</limit_tasks>`

### DIFF
--- a/html/project.sample/project_specific_prefs.inc
+++ b/html/project.sample/project_specific_prefs.inc
@@ -22,7 +22,9 @@ define('APP_SELECT_PREFS', true);
     // user can choose which apps to run
 define('NON_GRAPHICAL_PREF', false);
     // user can choose to run faster non-graphical app versions if available
-
+define ('LIMIT_TASKS_PREFS', true);
+    // user can limit numuber of tasks per host
+    
 // Project-specific prefs are represented in three ways:
 // - as an XML document (stored in the DB in this form)
 // - as a PHP structure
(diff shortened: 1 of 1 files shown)


ChristianBeer:
The comment says: `user can limit numuber of tasks per host` but the option is only a checkbox. So what is the limit here? What happens when I turn this off or on? I can already hear people ask for an input field where they can set their own limit.

Also doing it that way makes it complicated to introduce this preference into existing projects if the default should be to limit tasks. An absence of the tag would mean the same as `<limit_tasks>1</limit_tasks>`. Just imagine an existing user looking at the preference page where it says "Limit the number of tasks per host? No" but that can mean two things. Either the tag is missing from the project specific preferences (imposing the default to limit the tasks) or it is set to `<limit_tasks>0</limit_tasks>` (imposing no limit on tasks).

html/project.sample/project_specific_prefs.inc
```diff
@@ -57,6 +59,8 @@ define('APP_SELECT_DESC', tra('Run only the selected applications'));
 define('APP_SELECT_TOOLTIP', tra('Only get tasks for certain applications. Useful to focus on particular applications, or to exclude them.'));
 define('ACCEPT_ANY_DESC', tra('If no work for selected applications is available, accept work from other applications?'));
 define('NON_GRAPHICAL_DESC', tra("Use faster non-graphical applications if available?"));
+define("LIMIT_TASKS_DESC", tra("Limit the number of tasks per host?", "<br><span class=note>", "</span>"));
```
ChristianBeer: The two extra options are not needed since there is no `%1` and `%2` in the first argument. something like that would work tho:
`define("LIMIT_TASKS_DESC", tra("Limit the number of tasks per host? %1Description of this feature%2", "<br><span class=note>", "</span>"));`

davidpanderson:
I did a commit (fb5b9d9) that does this in a way that lets you choose the limits.

ChristianBeer:
Closing because it has conflicts.

Issue #1595, opened 2016-07-29
Remove or modify existing translation?


GITNE:
Hello guys!

I am working on a fix to the [```plurals```](https://developer.android.com/guide/topics/resources/string-resource.html#Plurals) issue of the [```notice_notification_single_header```](https://github.com/BOINC/boinc/blob/master/android/BOINC/res/values/strings.xml#L375) and [```notice_notification_multiple_header```](https://github.com/BOINC/boinc/blob/master/android/BOINC/res/values/strings.xml#L376) string resources. Now, I am wondering what would be better to do; to remove these or to just modify them to mimic the current (in some languages incorrect) implementation in languages I cannot provide proper translation for myself? Removing could notify or alert language coordinators because Transifex would lower its translation progress indicator. Language coordinators would then need to review the new or missing string resources. Whereas just modify the string resources to mimic the current implementation might not cause Transifex to send notification or alert messages to language coordinators. However, language coordinators may not always respond quickly enough or not at all, so that if the resource string is removed BOINC users of unsupported languages would be presented with the default (English) strings.

So, how should I proceed?

ChristianBeer:
Can you please give an example on what gets changed? I.e. how would the English strings look before and after your proposed changes?

What was done in the past is to change the English source string, compile a new template file for Transifex and inform the translators via a Transifex news item and the boinc_loc mailinglist that something big has changed with some instructions what to do.

GITNE:
@ChristianBeer Thank you for your timely response.

> Can you please give an example on what gets changed? I.e. how would the English strings look before and after your proposed changes?

No need for an example. It is a question about workflow. But, if really want to know, it is not a big change.

As you have suggested about past proceedings, it is probably best to remove the affected string resources in other languages and inform translators. I am going post a pull request with individual commits pertaining to languages I can handle and those I cannot, so you will be able to see what changed.

ChristianBeer:
Why do you want to tackle each language separately? All that needs changing is the English source string. This needs to be fed into Transifex where it will be added to every language and then automatically committed back into source as soon as a translator translates the new string.

So the workflow is:

1. change the template file here on github (@romw should know which file is for Android)
2. Transifex will pick up the change and change languages accordingly
3. tell translators that new strings are available and they should translate them
4. Transifex will commit the translations back to github

GITNE:
> change the template file here on github (@romw should know which file is for Android)

[...]

PR #1596, opened 2016-07-29
AndroidBOINC: Fix plurals issue in notification messages

GITNE:
See https://developer.android.com/guide/topics/resources/string-resource.html#Plurals for details.

As to #1595, in the next commit [```notice_notification_single_header```](https://github.com/BOINC/boinc/compare/master...GITNE:plurals#diff-f4e361a555c3a8b78cf2d29ffe035cb2L375) and [```notice_notification_multiple_header```](https://github.com/BOINC/boinc/compare/master...GITNE:plurals#diff-f4e361a555c3a8b78cf2d29ffe035cb2L376) would get removed in all other languages so that Transifex would inform translators about the new [```notice_notification```](https://github.com/BOINC/boinc/compare/master...GITNE:plurals#diff-f4e361a555c3a8b78cf2d29ffe035cb2R375) string resource. Translators will not have to bother about the obsolete string resources and will also not forget (or if nobody cares) to remove them.

### DIFF
--- a/android/BOINC/res/values-ca/strings.xml
+++ b/android/BOINC/res/values-ca/strings.xml
@@ -334,14 +334,10 @@
   <string name="about_name">BOINC</string>
   <string name="about_version">Versió</string>
   <string name="about_name_long">Berkeley Open Infrastructure for Network Computing</string>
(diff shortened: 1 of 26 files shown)


romw:
Thanks

PR #1601, opened 2016-08-02
AndroidBOINC: Fix value of quantity attribute to "other" than "many"

GITNE:


ChristianBeer:
This is already changed in master. Translations will be updated as soon as they get translated. I'll do the polish one in Transifex.

PR #1602, opened 2016-08-02
Proper rebase

GITNE:
This rebase should serve better than dbdb72032838af04f024681429c2e1796340048c and subsequent commits.

### DIFF
--- a/android/BOINC/assets/all_projects_list.xml
+++ b/android/BOINC/assets/all_projects_list.xml
@@ -1,6 +1,399 @@
 <?xml version="1.0" encoding="ISO-8859-1" ?>
 <projects>
     <project>
+        <name>theSkyNet POGS</name>
+        <url>http://pogs.theskynet.org/pogs/</url>
+        <general_area>Physical Science</general_area>
+        <specific_area>Astronomy</specific_area>
+        <description><![CDATA[TheSkyNet POGS is an astronomy project studying 16 different properties of galaxies, including brightness, mass, amount of dust and how fast stars are forming. We are using your processing power to conduct pixel-by-pixel calculations on multi-wavelength (ultraviolet, optical and near-infrared light) images of galaxies to produce an atlas that will help astronomers to better understand the distant universe.]]></description>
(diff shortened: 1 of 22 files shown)


ChristianBeer:
I don't see a problem with current master. What are you trying to fix?
I'll try to investigate where dbdb72032838af04f024681429c2e1796340048c comes from tomorrow. I don't have it locally and it looks odd here on github.

GITNE:
It is generally better to fix next to the initial commit than ```n``` commits after the fact. Actually, all commits since cfb4a139bf2660ec9d0fe94ec0bdd9b9b84c7bee can be discarded because they either break the build or are just *fiddling* around. We now know better how to deal with Transifex.

romw:
As a general rule, anything that rolls back the pointer to any branch is bad when dealing with a publically accessible repo.  Anybody who has downloaded any of the changes before the rollback would receive some ugly error messages the next time they pull updates.

Fixing the initial commit can only really work on your local repo.

GITNE:
A rebase is not a *rollback*. The rebase operation enables you to make changes to the commit history while explicitly not affecting other people.

What you seem to have in mind is either a ```reset``` or ```revert``` operation. Whereas a ```revert``` operation does not affect other people either because it creats a *negative* patch first and then commits it to ```HEAD```. New commits never affect other people by definition.

romw:
I'm going to close this PR as fa6a4ad already handles this.  The other language's XML files will be updated as they are translated in Transifex.

ChristianBeer:
Granted the rebase would look better and would be somewhat transparent to users but I don't like the idea of rewriting history. Yes, we made errors in the past and fiddled around. This is kind of the BOINC way :) and there are other examples already in commit history. We should rather focus on not making the same mistake again.

To also solve my confusion from earlier: It seems that github shows commits that are part of a pull request as belonging to the target repository. But those commits are not send via a `fetch` operation so I couldn't see dbdb72032838af04f024681429c2e1796340048c locally which confused me. I verified that the current master branch is sane (and was sane the whole time) and can be used again.

PR #1604, opened 2016-08-04
BOINC GAHP: print SVN_VERSION with --version CLA

bema-ligo:
The hardcoded version information written by boinc_gahp is not very helpful when I need to record which code version a particular binary was built from.

If there is a more suitable (i.e. GAHP protocol conformant) way to write SVN_VERSION to stdout I would be fine with that, too. If so, pleas implement this and reject this merge request.

### DIFF
--- a/samples/condor/Makefile
+++ b/samples/condor/Makefile
@@ -3,8 +3,8 @@ all: boinc_gahp
 clean:
 	rm boinc_gahp
 
-boinc_gahp: boinc_gahp.cpp ../../lib/remote_submit.h ../../lib/remote_submit.cpp
-	g++ -g -O0 -I../../lib \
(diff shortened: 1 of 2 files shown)


ChristianBeer:
I checked the [GAHP spec](https://research.cs.wisc.edu/htcondor/gahp/gahp_protocol.txt) about the VERSION command. This looks like this:
```
$GahpVersion: <major>.<minor>.<subminor> <build-month> <build-day-of-month> <build-year> <general-descrip> $
```
It seems we could add the SHA1 to the `<general-descrip>` part which should be "a string identifying a particular GAHP server implementation", which currently is `BOINC\\ GAHP`.

The advantage of adding the SHA1 to the VERSION command would be that this can then be retrieved via a Condor command directly.

PR #1608, opened 2016-08-09
Fixes from testing the job and cpu limiting functionality

lfield:
The parsing has been fixed and the max_jobs_per_rpc  and max_jobs_in_progress settings now use project_prefs.max_jobs for their values.


lfield:
David Anderson has applied these changes in a recent commit.

PR #1627, opened 2016-08-22
Drupal fix ignore user comment

drshawnkwang:
Drupal: Fixed bug where ignoring a user will not only hid his/her post, but all posts beneath it.

Fixed by added code to end of comment.tpl.php, adding two additional </div> close tags to match the two open <div> tags created above.

https://dev.gridrepublic.org/browse/DBOINCP-282


tristanolive:
Somehow managed to botch the merge... but these changes have been included manually.

PR #1629, opened 2016-08-23
Fix #1624 — Some header files not installed.

### LINKED ISSUE #1624: Some header files are not installed
#After doing a rather simple `./configure && make install` and trying to compile my very first worker program, I found out that multiple header files that become necessary after inclusion of `boinc_api.h` simply are not there in `/usr/local/include/boinc` with other files. After copying those files manually, the build succeeded, which makes me wonder whether I am missing something about the process.

## str_replace.h ##
First referenced by: `coproc.h` ← `hostinfo.h` ← `app_inc.h` ← `boinc_api.h`
Can be found at: `$SOURCEDIR/lib`.
This header file is deliberately put into `noinst_HEADERS` list inside `Makefile.am` with a note: “These are used internally and are not meant to be installed”.

## config.h ##
First referenced by: `str_replace.h` ← `coproc.h` ← `hostinfo.h` ← `app_inc.h` ← `boinc_api.h`
Can be found at: `$BUILDDIR`.
This file does not get installed regardless of whether the build occurs in-tree or out-of-tree.

## project_specific_defines.h ##
First referenced by: `config.h` ← `str_replace.h` ← `coproc.h` ← … ← `app_inc.h` ← `boinc_api.h`
Can be found at: `$SOURCEDIR`.
At first I thought this may be just a template for an application-specific definitions, but there are multiple projects mentioned there, so it looks more like shared file.

SamsonovAnton:
All header files referenced by *boinc_api.h* will be installed now, including *config.h*, *str_replace.h* and *project_specific_defines.h*.

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -133,6 +133,7 @@ pkginclude_HEADERS = \
     cert_sig.h \
     cl_boinc.h \
     common_defs.h \
+    $(top_builddir)/config.h \
     coproc.h \
     crypt.h \
     diagnostics.h \
@@ -152,9 +153,11 @@ pkginclude_HEADERS = \
     parse.h \
     prefs.h \
     procinfo.h \
+    $(top_srcdir)/project_specific_defines.h \
     proxy_info.h \
 	sched_msgs.h \
     stackwalker_imports.h \
+    str_replace.h \
     str_util.h \
     url.h \
     util.h \
@@ -168,7 +171,6 @@ noinst_HEADERS = \
     proc_control.h \
     procinfo.h \
     shmem.h \
-    str_replace.h \
     synch.h \
     unix_util.h
 


ChristianBeer:
I found this old commit 78f7610f6e766f8772b3922807050c18b23c03b3 which states that str_replace.h and config.h should not be a dependency of boinc_api.h maybe it's better to clean the dependency path so boinc_api.h does not depend on str_replace.h than installing the internal headers.

SamsonovAnton:
> Commit 78f7610f6e766f8772b3922807050c18b23c03b3 states that *str_replace.h* and *config.h* should not be a dependency of *boinc_api.h*. Maybe it's better to clean the dependency path, so *boinc_api.h* does not depend on *str_replace.h*, than installing the internal headers.

I may be missing some point here, but as far as I understand, the purpose of that patch was to refactor source and header files so they include needed headers explicitly rather than rely on some underlying (or, even worse, upper-level) header file to include those for them. That was indeed a proper correction.

However, that patch does not in any way eliminate the need for *str_replace.h*, as its is referenced by *coproc.h*, *parse.h* and *common_defs.h* that are exported as part of public API; the same holds true for *config.h*, which is referenced by *backend_lib.h*, *gui_rpc_client.h* and *str_replace.h*. Do those other headers really need to include *str_replace.h* and *config.h*? Can they be modified to not to depend on the files in question? That is pretty difficult topic for a bystander like me. Proposed patch just fixes the omitted functionality with minimal intervention.

ChristianBeer:
My Problem is that with this change we are exporting a header file that is still considered internal. It might be better to change coproc.h, parse.h and common_defs.h to not include internal headers. That's what I meant with "cleaning the dependency path". Making sure that header files that are exported do not depend on header files that are not exported. The commit I referenced should make clear what the intention of the developer is.
I'll see if I can find a way to clear this up without exporting the internal headers.

SamsonovAnton:
Trying to naively check whether dependencies are real, — by commenting `#include "•••.h"` out and attempting to compile a simple worker that does not import any string functionality from BOINC API:

* When `#include "str_replace.h"` is commented out…
 * in *common_defs.h* — success;
 * in *coproc.h* — error: `safe_strcpy()` is needed by inline function `void COPROCS::clear()` defined in *coproc.h*;
 * in *parse.h*, `str_replace()` is declared as external function without including *str_replace.h*.
* When `#include "config.h"` is commented out…
 * in *backend_lib.h* — false positive, it includes *sched_config.h* actually;
 * in *gui_rpc_client.h* — false positive, it includes *cc_config.h* actually;
 * in *str_replace.h* — success.

Checking references to `safe_strcpy()`:

* used by inline constructor of `DEVICE_STATUS` in *common_defs.h*, although just to assign an empty string;
* used by inline constructor of `COPROCS` defined in *coproc.h*, in addition to the aforementioned `COPROCS::clear()`.

According to *str_replace.h*, `safe_strcpy()` is just a syntactic sugar alias for `strlcpy()`, which may or may not be provided by *libc*, — that is what *config.h* is consulted for; without *config.h* it would simply divert to self-provided implementation, however I am not sure whether that one gets compiled when there is a prebuilt one.

Now, let us augment the worker to include string functionality from BOINC API. Because including *boinc_api.h* will trigger inclusion of other headers which will obscure the experiment, let us write a new non-worker program that includes *str_util.h* directly. Also, *config.h* and *str_replace.h* will be temporarily uninstalled during this test, along with *project_specific_defines.h*.

* [Documentation](http://boinc.berkeley.edu/trac/wiki/AppDev#Stringfunctions) is very scarce on this, but at least it points at *str_util.h* in the first place, — calling arbitrary functions listed there, like `is_valid_filename()`, `strip_whitespace()`, `time_to_string()` and `strcpy_overlap()`, was successful.

* But what about “Cross-platform support for `strlcpy()`, `strlcat()`, `strcasestr()` and `strcasecmp()` is in *lib/str_replace.cpp,h*” also mentioned in documentation? It obviously requires *str_replace.h* which in turn would require *config.h* for proper operation.

So, what can be concluded from this? I have no idea.

ChristianBeer:
I also arrived at a similar conclusion. The idea of `str_replace.h` is to provide declarations for functions that are missing from libc and that are actually implemented in str_util.cpp if needed. As long as you don't use functions declared in str_replace.h in exported headers we are save. But with the examples you already provided this is different. I also just finished reading through the discussion why not to include `strlcpy` and `strlcat` into libc where they make the point that they are unsafe because it truncates strings. But this is not a real problem as both return the final size which can be tested against the expected size. But this is beside the point here.

I think a better fix than exporting `str_replace.h` is to substitute the references to `safe_strcpy` with something portable so we don't need to include `str_replace.h` in `coproc.h` and `common_defs.h`. this should also fix the `config.h` issue. What do you think?

SamsonovAnton:
To tell the true, using `safe_strcpy()` in those parts of code seems overkill, as source strings are hardcoded literals of fixed length, so simple `strcpy()` would be enough.

By the way, there is a macro `strcpy2()` defined in *db_base.h* that calls `strlcpy()`. If that macro is removed, building a worker application still succeeds, so it is definitely not needed for worker development. It looks like this macro is used exclusively by *boinc_db.cpp*, — may be it would be better to move that macro there?

ChristianBeer:
Yes, using a simpler function for fixed length strings seems good. But reverting to `strcpy()` will only trigger warnings in code review tools like Coverity Scan. So we need a better solution that does not do that. My first guess would have been to use `snprintf()` but that has problems on Windows (see the comment in `str_replace.h`) so we are back at `strcpy()` or `strncpy()` where we have to ensure that the resulting array is null terminated and hope that the code review tool recognizes this.

And Yes, I also think we should move `strcpy2()` from the header to the implementation since `str_replace.h` is not even included in `db_base.h`.

PR #1635, opened 2016-09-01
Validator: implement "suspicious" results

ChristianBeer:
A validator now has the possibility to mark a single result as "suspicious" by making init_result() return VAL_RESULT_SUSPICIOUS. If this is the single quorum result of an adaptive replication, this will trigger another task to be generated for validation.


davidpanderson:
Please document here:
http://boinc.berkeley.edu/trac/wiki/ValidationSimple
and mention it here:
http://boinc.berkeley.edu/trac/wiki/AdaptiveReplication

ChristianBeer:
Documentation done.

PR #1637, opened 2016-09-02
remote submission: delete uploaded files in case of an error

bema-ligo:
Avoid leaving uploaded files lying around that the DB has no record of when these fail to be moved or inserted into the DB.

### DIFF
--- a/html/user/job_file.php
+++ b/html/user/job_file.php
@@ -162,30 +162,41 @@ function upload_files($r) {
     $delete_time = (int)$r->delete_time;
     $batch_id = (int)$r->batch_id;
     //print_r($_FILES);
-    $i = 0;
+    $i = -1;
+    $upload_error = "";
     foreach ($r->md5 as $f) {
+        $i++;
         $md5 = (string)$f;
         $name = "file_$i";
         $tmp_name = $_FILES[$name]['tmp_name'];
         if (!is_uploaded_file($tmp_name)) {
-            xml_error(-1, "$tmp_name is not an uploaded file");
+            $upload_error .= "$name:$md5 was not uploaded ";
+            continue;
         }
         $fname = job_file_name($md5);
         $path = dir_hier_path($fname, project_dir() . "/download", $fanout);
-        rename($tmp_name, $path);
+        if (!rename($tmp_name, $path)) {
+            $upload_error .= "could not move $tmp_name to $path ";
+            unlink($tmp_name);
+            continue;
+        }
         $now = time();
         $jf_id = BoincJobFile::insert(
             "(md5, create_time, delete_time) values ('$md5', $now, $delete_time)"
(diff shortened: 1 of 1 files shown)


html/user/job_file.php
```diff
@@ -162,30 +162,49 @@ function upload_files($r) {
...
+            continue;
         }
         if ($batch_id) {
             BoincBatchFileAssoc::insert(
                 "(batch_id, job_file_id) values ($batch_id, $jf_id)"
             );
         }
-        $i++;
+    }
+    if ($move_error) {
```
ChristianBeer: Shouldn't this be done inside the foreach loop? If multiple files have the same type of error only the last file will be reported because $*_error is overwritten each time and only reported outside the foreach loop.

bema-ligo: What exactly does "this" refer to?

The xml_error() actually terminates the function and thus the loop. If the xml_error() call is inside the loop, it is terminated after the first error, and other ("later") uploaded files aren't deleted and would stay in the temp directory forever. One could argue, though, that in case of one error _all_ files uploaded with the request need to be deleted immediately from the filesystem and the DB, but this would make it even more complicated.

Then again, though, the loop should probably go over all actually uploaded files in $_FILES rather than the list of md5s in the request, but that would mean a complete re-write of upload_files().

ChristianBeer: "this" refers to the three if statements on line 200ff. 

Use-Case: uploading two files (file1.zip, file2.zip) which both could not be moved to the download directory. Because of line 181 which is called once for each file, $move_error at the end (on line 200) only contains "could not move /tmp/file2.zip to download/*/file2.zip" thus hiding the error with file1.zip

A quick fix would be to initialize `$*_error` with an empty string. Concatenating the error messages and checking for `strlen($*_error)>0` in the if blocks at the end.

To catch all errors one could also only use one `$error_str` variable and concatenate the errors to it and call xml_error() only once reporting all the errors that happened.

bema-ligo:
Right, the numeric error value is always -1 anyway. Changed for now.

bema-ligo:
Right, the numeric error value is always -1 anyway. Changed for now.

FWIW none of these versions have actually been tested yet, not even ran through PHP for a syntax check. Please do that when you get the chance.

ChristianBeer:
Actually there are PHP tests performed on each pull request by scruinizer-ci. But lately they are not shown in the status here on github. There is in fact a syntax error here that made the check fail but that did not get shown. I'm investigating how to get the status back.

ChristianBeer:
I took the liberty to refactor the whole` upload_files()` logic and fix this problem too. This is now in #1640. The scrutinizer thing is still not resolved as this PR should not have passed the test. I'm working with scrutinizer support to get the check back into the github status.

PR #1642, opened 2016-09-07
Fixing Gridmanager fast looping issue

lfield:
This fixes an issue which causes the Gridmanager to log about 10K entries every second that look like:

08/05/16 23:10:38 [105816] GAHP[105817] -> 'S' '0'
08/05/16 23:10:38 [105816] GAHP[105817] <- 'RESULTS'
08/05/16 23:10:38 [105816] GAHP[105817] -> 'S' '0'
08/05/16 23:10:38 [105816] GAHP[105817] <- 'RESULTS'

### DIFF
--- a/samples/condor/boinc_gahp.cpp
+++ b/samples/condor/boinc_gahp.cpp
@@ -219,7 +219,6 @@ int process_input_files(SUBMIT_REQ& req, string& error_msg) {
     vector<string> upload_boinc_names, upload_paths;
     for (unsigned int i=0; i<absent_files.size(); i++) {
         int j = absent_files[i];
-        printf("file %d is absent\n", j);
         upload_boinc_names.push_back(boinc_names[j]);
         upload_paths.push_back(paths[j]);
     }


ChristianBeer:
I don't see the immediate connection between the BOINC_SUBMIT command and the fast looping of the grid manager because of this stray printf(). But we don't upload files this way so we don't have "absent" files thus we don't see the fast looping issue.

lfield:
The issue is that the print statement seems to make the gahp and Gridmanager go out of sync when it is asking for RESULTS..

lfield:
Please can this get merged soon so that it can be included in the next release of Condor.

ChristianBeer:
I can merge it as is but on a second thought wouldn't it be better to guard this `printf` with a `#ifdef DEBUG` like all the other debug outputs? There is no synchronization in debug mode so the issue should not return or did you already check that?
Looking for other unprotected `printf()` statements I found one on line 791 that should be `BPRINTF()` or it would not have the response_prefix added like all the other responses.

lfield:
If anything it is a debug statement but it is not needed.

PR #1645, opened 2016-09-09
Fix Windows build by including needed headers.

SamsonovAnton:
These are some minor fixes (all regarding *lib/str_replace.h* and *lib/str_util.h*) I had to apply when building *libboinc_staticcrt* and *libboincapi_staticcrt* for Windows while porting my application — there may be other fixes needed by other projects.


davidpanderson:
Thanks!

PR #1648, opened 2016-09-15
Drupal: convert preset attribute to tag

ChristianBeer:
The additional preset attribute in the venue tag is breaking preference parsing in the scheduler. We now use a tag instead of an attribute and clear any old attributes.
This is only a problem with generalprefs because they get propagated to other projects that do not have a scheduler which ignores the attribute. The Einstein@Home scheduler (only project to use the Drupal webcode right now) was patched to handle extra attributes so we don't need to strip the attribute from projectprefs.

### DIFF
--- a/drupal/sites/default/boinc/modules/boincwork/includes/boincwork.forms.inc
+++ b/drupal/sites/default/boinc/modules/boincwork/includes/boincwork.forms.inc
@@ -36,8 +36,15 @@ function boincwork_generalprefs_form(&$form_state, $venue, $prefs_preset = null,
     }
     
     // Determine if a preset is selected or if these are custom settings
-    if (!isset($prefs['@attributes']['preset'])) $prefs['@attributes']['preset'] = 'custom';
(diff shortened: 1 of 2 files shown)


brevilo:
Looks sane to me but Tristan should finally sign off on this. Thanks for adding the old/new conversion. I'll run the clean-up script nonetheless.

brevilo:
BTW, this is yet another example where BOINC's so-called XML parser shows that it's neither used everywhere not that it actually is a full XML parser as it can't even handle simple attributes.

ChristianBeer:
Referencing the original issue: https://dev.gridrepublic.org/browse/DBOINCP-297 where @drshawnkwang tested and posted a refined solution that I'm going to add to this PR.

drshawnkwang:
I will only add that I checkout-ed this branch with the latest changes, and verified that the preferences added to the database have format <venue name="xyz"> with no further attributes.

PR #1650, opened 2016-09-18
Locale DE: Fix spelling mistakes

m-ad:
I fixed some spelling mistakes in the German locale of the BOINC-Manager I stumbled upon, most of them had to do with capitalization ("sie" vs. "Sie").

(This is my first Github contribution, so I hope I did everything alright)


ChristianBeer:
Thanks for your contribution. Localizations are not committed directly to github but have to go through https://www.transifex.com/boinc/boinc/dashboard/, but the changes you made or not for nothing. There are several options you have now:
* create an account on transifex and upload your .po file to the BOINC-Manager resource
* change the strings directly in the transifex interface
* if both does not work for you just tell me so I can upload the changed po file

PR #1668, opened 2016-10-05
Drupal: default preference set is now 'standard'.

drshawnkwang:
The default preference set when a user visits the accounts preference page is (now) 'standard'.

Added form submit for BOINC: preferences preset admin interface. Admin may now Save configuration with the disk usage settings from config.xml. In this way, those XML settings will be loaded into the database as the default preferences presets. These presets are in turn used by drupal when the user visits the preferences page. The form on that page will be filled with the preferences from the drupal database.

If there are no preference preset loaded in the drupal database, then the standard preset will be filled in with the disk usage settings from config.xml. Once it is save into the drupal database, the admin must use the Save configuration with disk usage settings from config.xml button in order to reload the disk usage settings from config.xml again.

https://dev.gridrepublic.org/browse/DBOINCP-310


drshawnkwang:
There were some bugs in the PR code that I fixed with the latest commit.

PR #1669, opened 2016-10-06
Drupal: Fix workunit view for adaptive replication

drshawnkwang:
Suppress output if app is using adaptive replication

Fixed wrong variable name in view.
Added additional conditional logic using a) error_mask and b) config.xml `<dontsupressoutput>` setting to match BOINC code. This is to suppress output if the app uses adaptive replication.
(b) is obtained through a new field in the fiew which loads the setting from config.xml into a variable.

Changed workunit view template. Variables from view used to determine if task list is output.

https://dev.gridrepublic.org/browse/DBOINCP-255

### DIFF
--- a/drupal/sites/all/features/work_and_host_stats/work_and_host_stats.views_default.inc
+++ b/drupal/sites/all/features/work_and_host_stats/work_and_host_stats.views_default.inc
@@ -10631,6 +10631,43 @@ else {
       'field' => 'name',
       'relationship' => 'none',
     ),
+    'phpcode_8' => array(
+      'label' => 'dont suppress otuput',
+      'alter' => array(
+        'alter_text' => 0,
+        'text' => '',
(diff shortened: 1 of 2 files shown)


drupal/sites/all/features/work_and_host_stats/work_and_host_stats.views_default.inc
```diff
@@ -10631,6 +10631,43 @@ else {
       'field' => 'name',
       'relationship' => 'none',
     ),
+    'phpcode_8' => array(
+      'label' => 'dont suppress otuput',
```
drshawnkwang: Thanks, I'll correct it. That being said the label doesn't matter because it's a "hidden" field in a Drupal view, meaning the label is never shown to the user. The only people who would see it would be Drupal admins.

PR #1671, opened 2016-10-09
Fixed wrapper execution time error in android 

davidfcalle:
Android wrapper had an execution time error because de env variable $PWD was not found. Changing getenv("$PWD") for getcwd() fixes the issue.


davidpanderson:
Thanks.

Issue #1470, opened 2016-01-17
acct_mgr_request.xml - Not valid XML


Kati3e:
Run into another issue today with a user not being able to connect correctly to my account manager. After some debugging I have found the exact cause. The acct_mgr_request.xml is not valid XML due to a closing tag not being there. This issue was present on 2 machines, both running a different distro of Linux and versions of BOINC. (Ubuntu with 7.2.42 and ArchLinux with 7.6.22).

I've included a portion of the XML where the tag is missing.

I'm assuming this is an issue with World Community Grid, but I don't know exactly how this all works or who I should report this to, to get it fixed.

And shouldn't BOINC be validating XML before sending and receiving it? This seems like best practice to me.. Invalid XML could wreck havoc if devs aren't checking it before parsing.

Notice the missing ```</venue>``` tag for ```<venue name="work">```
This will fail any XML validation methods..
```
<global_preferences>

<source_project>http://www.worldcommunitygrid.org/</source_project>
    <source_scheduler>https://scheduler.worldcommunitygrid.org/boinc/wcg_cgi/fcgi</source_scheduler>

    <mod_time>1450950386</mod_time>
    <run_on_batteries>0</run_on_batteries>
    <run_if_user_active>1</run_if_user_active>
    <run_gpu_if_user_active>0</run_gpu_if_user_active>
    <idle_time_to_run>3</idle_time_to_run>
    <suspend_if_no_recent_input>0.0</suspend_if_no_recent_input>
    <suspend_cpu_usage>60.0</suspend_cpu_usage>
    <leave_apps_in_memory>0</leave_apps_in_memory>
    <cpu_scheduling_period_minutes>60</cpu_scheduling_period_minutes>
    <max_ncpus_pct>100.0</max_ncpus_pct>
    <cpu_usage_limit>100.0</cpu_usage_limit>
    <disk_max_used_gb>50.0</disk_max_used_gb>
    <disk_min_free_gb>1.0</disk_min_free_gb>
    <disk_max_used_pct>90.0</disk_max_used_pct>
    <disk_interval>60.0</disk_interval>
    <vm_max_used_pct>75.0</vm_max_used_pct>
    <ram_max_used_busy_pct>50.0</ram_max_used_busy_pct>
    <ram_max_used_idle_pct>90.0</ram_max_used_idle_pct>
    <work_buf_min_days>0.1</work_buf_min_days>
    <work_buf_additional_days>0.5</work_buf_additional_days>
[...]

PR #1676, opened 2016-10-14
Scheduler: fix global preference parsing

ChristianBeer:
The xp.element_contents() functions strips the linebreak before the closing tag. But the closing tag needs to be on it's own line because of the limitations of BOINC XML parsing.


ChristianBeer:
Related to #1470 and should fix the issue described if all projects involved apply this patch.

PR #1696, opened 2016-11-03
Drupal: show signatures in comments again

ChristianBeer:
Commit f64e644f inadvertently removed a crucial line that prevents signatures from being shown. I also got rid of the commented out comment that gave the impression that this function is not used.

### DIFF
--- a/drupal/sites/default/boinc/themes/boinc/template.php
+++ b/drupal/sites/default/boinc/themes/boinc/template.php
@@ -332,8 +332,10 @@ function boinc_preprocess_node_team_forum(&$vars, $hook) {
  * @param $hook
  *   The name of the template being rendered ("comment" in this case.)
  */
-///* -- Delete this line if you want to use this function
 function boinc_preprocess_comment(&$vars, $hook) {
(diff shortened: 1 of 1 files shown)


brevilo:
Adding @drshawnkwang as well as the author of the original commit.

tristanolive:
Good catch -- thanks Christian.

brevilo:
Just to be clear, this fixes the BOINC theme only, right?

drshawnkwang:
@brevilo , it will fix the issue on einstein. Einstein's theme is nearly all CSS.

tristanolive:
Einstein is a sub-theme of the BOINC theme, so changes to the BOINC template files are inherited by the Einstein theme. The exception to this is when an overriding template file is present in the Einstein theme, in which case that will supersede the inherited BOINC template (this is not common, as Shawn alludes).

tristanolive:
BTW, I missed a detail when merging this, so eed66cc has been added to correct that.

brevilo:
Ok, deployed both - works!

Issue #1519, opened 2016-04-02
MGR: Maximized window state is not restored on startup


chausner:
BOINC Manager saves and restores the window position and size correctly but does not restore the window state (maximized/normal). The state seems to be saved into a registry value "WindowMaximized" when closing the manager but it is discarded on startup.

Would be nice if this could be fixed, it has bugged me for years.

AenBleidd:
Fixed in PR #1701.

PR #1701, opened 2016-11-07
Fix issue #1519: MGR: Maximized window state is not restored on startup

AenBleidd:


AenBleidd:
Closed. New pull request opened #1711.

Issue #1570, opened 2016-07-11
Migrate to Gradle for build to enable Android Studio development


brierjon:
As mentioned in #1439 the BOINC for Android code still is configured for Eclipse which has limited support. Android Studio is where primary development for Android tools and resources are focused. 

This issue is created to track and discuss the migration to Android Studio from Eclipse. 

The main task requires migrating the build to Gradle which an export from eclipse and verification of the conversion to gradle for building is the largest barrier to converting from Eclipse to Android Studio.

Resources:
1. Migrating to gradle for building: https://developer.android.com/studio/build/index.html
2. Migrate from Eclipse to Android Studio: https://developer.android.com/studio/intro/migrate.html

Note: Eclipse support for Android Development tools ended Nov 2, 2016 - https://android-developers.blogspot.com/2016/11/support-ended-for-eclipse-android.html

brevilo:
@brierjon I presume this one could be closed now?

brierjon:
Yes, gradle build support and structure is present. Closing.

brevilo:
Thanks

Issue #1702, opened 2016-11-08
Unable to build ARM client


denravonska:
I'm trying to build the ARM Android client from master but get the following error:

```
cd android
./build_boinc_arm.sh

...
make[2]: Leaving directory '/home/marco/dev/boinc/lib'
Making all in client
make[2]: Entering directory '/home/marco/dev/boinc/client'
  CXX      libwhetneon_a-whetstone.o
In file included from cpu_benchmark.h:37:0,
                 from whetstone.cpp:37:
/home/marco/androidarm-tc/lib/gcc/arm-linux-androideabi/4.9.x/include/arm_neon.h:31:2: error: #error You must enable NEON instructions (e.g. -mfloat-abi=softfp -mfpu=neon) to use arm_neon.h
 #error You must enable NEON instructions (e.g. -mfloat-abi=softfp -mfpu=neon) to use arm_neon.h
  ^
Makefile:810: recipe for target 'libwhetneon_a-whetstone.o' failed
make[2]: *** [libwhetneon_a-whetstone.o] Error 1
```

I can see that some work on this was done in 7b2ca9e787a204f2a57f390bc7249bb7f9997fea but it seems to clash.

denravonska:
I did what the compiler suggested which is to change -mfpu=vfpv3-d16 to -mfpu=neon. It worked for me on my device but the test scope was very limited.

davidpanderson:
I can't help fix this, but I can tell you the idea: when we started Android, devices were about 50/50 between VFP and Neon for floating-point support.  We wanted the floating-point benchmark (whetstone) to use whichever was present.  So we needed the client to included versions of the benchmarks compiled for both types of device.  We did this by compiling the code into different C++ namespaces.

ChristianBeer:
I tracked it down to 19223fd86881c98e9d7b24932e177f232becdc10 where Rom added the flags. It turns out he missed the flag to switch the correct floating-point support for the benchmark. Instead of doing what the compiler says just add `-DANDROID_VFP` to lines 25 and 25 and 26 in  android/build_boinc_arm.sh

This will build you a VFP enabled client. If you want to have a NEON enabled Client you need to add `-DANDROID_NEON` and change to `-mfpu=neon` on the same lines. VFP is more general although all newer ARM CPUs (that are powerful enough for BOINC) do have NEON which seems to be faster.

@denravonska Can you please test the changes I described and create a pull request for them if you can build the android client this way?

denravonska:
[...]

PR #1703, opened 2016-11-08
Android studio support

denravonska:
I have imported the project into Android Studio, upped the target SDK and fixed the problems that arose. The ARM asset client doesn't compile due to #1702. Adding the fix proposed by the compiler allows you to compile, but I have limited ways of testing it which is why I didn't include it here.

The app starts and behaves correctly in Swedish.


ChristianBeer:
Thanks for looking into this. What else does the increased SDK version imply? Do we exclude older devices that way?

I just did a cursory inspection of your commits. The ones that stand out are related to the translation system. The way the BOINC translation system is set up is that you shouldn't edit the translations directly but update them on [transifex](https://www.transifex.com/boinc/boinc/dashboard/). Whenever a resource is 100% translated into a non-english language the file gets transfered into the github repository automatically, overwriting your changes in f2c3963a91. If you don't want to create an account there just tell me so I can manually copy the translations over.

The string you are adding in 3de9867aa was deliberately removed in 79aca0feb71 so if there is a reference to it this must be updated to use the new string.

denravonska:
We still target API 16 (Jelly Bean) but in order to build with newer features I had to bump the compile SDK to 23. Actually, I just did a quick search and the TV features requires API level 21, not 23, so it should've worked for me without changing this. I'll double check it tomorrow.

Regarding the automatic translation merge it will be a problem since the translations have been moved from `res/` to `app/src/main/res/` so I guess there will be a conflict there, right?

I wasn't aware about transifex. I'll add my translation modifications there instead. Thanks :)

ChristianBeer:
> Regarding the automatic translation merge it will be a problem since the translations have been moved from res/ to app/src/main/res/ so I guess there will be a conflict there, right?

Yes, that change needs to be coordinated with a corresponding change to the transifex-github bridge that we have running on a separate server.

brierjon:
This commit will allow us to close #1570

davidpanderson:
Should I go ahead and merge this?

denravonska:
I have been playing around with it when fiddling with TCP support and it seems to work both for building and debugging. I haven't tested signing a release but I guess the CI server can test that from this branch, right? The only issue will be the translations and the path change.

Also, someone should have a look at #1702 since it blocks the BOINC ARM build for Android.

ChristianBeer:
We should make sure this really works and built a Client ourselves before merging into master. We should also make a documented review of the code changes. I mean no offense to @denravonska but it's easy to sneak in a malicious change with this many changes and I would like to avoid that with code coming from a new contributor.

We should also discuss on how to proceed with the required changes to the translation system and what our options are to get it working again.

denravonska:
@brierjon True, fixed.

AenBleidd:
@denravonska Is this pull request finished and can be merged?
@brierjon, @ChristianBeer Could you please review this pull request? There was no commits during the last month related to this PR. 
Thanks

davidpanderson:
I merged this just to keep things moving.  Has the  Neon/VFP benchmark issue been fixed?

denravonska:
@davidpandersson I have tried the suggested fix and it does build, but I am unable to verify due to lack of hardware.

PR #1710, opened 2016-11-11
Fix issue #1709: Malformed XML in RPC replies for lookup_account_poll and get_all_projects_list

### LINKED ISSUE #1709: Malformed XML in RPC replies for lookup_account_poll and get_all_projects_list
As discussed in #1509, for some GUI RPC commands the BOINC daemon returns XML that comes directly from the BOINC server or a project server. Since this XML has an XML declaration it leads to malformed XML when the BOINC daemon wraps it in a GUI RPC reply:

```XML
<boinc_gui_rpc_reply>
<project_config>
    <name>Einstein@Home</name>
...
```

The issue has been fixed for get_project_config_poll in https://github.com/BOINC/boinc/commit/1fe02c9d4fe815a69096e5cfc8e8985aa2308890 but still needs to be addressed for lookup_account_poll and get_all_projects_list. Also, it would be good if someone familiar with the codebase could check that there are no other RPC commands where the issue remains.

chausner:
This fixes #1709 by removing the XML declaration from lookup_account_poll and get_all_projects_list replies as it is already done for get_project_config_poll. Also, any error messages are now retained in the responses.

### DIFF
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -804,7 +804,11 @@ void handle_lookup_account_poll(GUI_RPC_CONN& grc) {
             grc.lookup_account_op.error_num
         );
     } else {
-        grc.mfout.printf("%s", grc.lookup_account_op.reply.c_str());
+        const char *p = grc.lookup_account_op.reply.c_str();
+        const char *q = strstr(p, "<account_out"); 
+        if (!q) q = strstr(p, "<error");
+        if (!q) q = "<account_out/>\n"; 
+        grc.mfout.printf("%s", q); 
     }
 }
 
@@ -1072,7 +1076,9 @@ static void read_all_projects_list_file(GUI_RPC_CONN& grc) {
     int retval = read_file_string(ALL_PROJECTS_LIST_FILENAME, s);
     if (!retval) {
         strip_whitespace(s);
-        grc.mfout.printf("%s\n", s.c_str());
+        const char *q = strstr(s.c_str(), "<projects");
+        if (!q) q = "<projects/>";        
+        grc.mfout.printf("%s\n", q);
     }
 }
 


client/gui_rpc_server_ops.cpp
```diff
@@ -775,7 +775,8 @@ void handle_get_project_config_poll(GUI_RPC_CONN& grc) {
         );
     } else {
         const char *p = grc.get_project_config_op.reply.c_str();
-        const char *q = strstr(p, "<project_config");
+        const char *q = strstr(p, "<project_config>");
```
ChristianBeer: This will remove attribute support from this tag. If, for whatever reason, we decide to put an attribute to that tag it will not be recognized with this change.

client/gui_rpc_server_ops.cpp
```diff
@@ -775,7 +775,8 @@ void handle_get_project_config_poll(GUI_RPC_CONN& grc) {
...
+        if (!q) q = strstr(p, "<error>");
         if (!q) q = "<project_config/>\n";
```
ChristianBeer: This line is now dead code and is never reached. The output now also changes from `<project_config/>` to `<error>` which breaks code that expects the old behavior.

client/gui_rpc_server_ops.cpp
```diff
@@ -804,7 +805,11 @@ void handle_lookup_account_poll(GUI_RPC_CONN& grc) {
             grc.lookup_account_op.error_num
         );
     } else {
-        grc.mfout.printf("%s", grc.lookup_account_op.reply.c_str());
+        const char *p = grc.lookup_account_op.reply.c_str();
+        const char *q = strstr(p, "<account_out>"); 
+        if (!q) q = strstr(p, "<error>");
+        if (!q) q = "<account_out/>\n";
```
ChristianBeer: This is now also dead code and a change in behavior.

client/gui_rpc_server_ops.cpp
```diff
@@ -1072,7 +1077,9 @@ static void read_all_projects_list_file(GUI_RPC_CONN& grc) {
     int retval = read_file_string(ALL_PROJECTS_LIST_FILENAME, s);
     if (!retval) {
         strip_whitespace(s);
-        grc.mfout.printf("%s\n", s.c_str());
+        const char *q = strstr(s.c_str(), "<projects>");
+        if (!q) q = "<projects/>";        
+        grc.mfout.printf("%s\n", q);
```
ChristianBeer: This is more about the commit messages. I dislike a message like 'fix previous commit'. What does 'previous' mean in git? If I cherry-pick this commit 'previous' will get a complete new meaning. Please state why you changed what you changed in the commit. 
Please also prepend the first line with the component ("Client:"). This used to be an unwritten rule for every BOINC developer from the beginning but I just saw that even seasoned BOINC developers don't uphold this rule.

client/gui_rpc_server_ops.cpp
chausner: I don't see how there is dead code. Can you elaborate? The fact that in case of an error it no longer outputs `<project_config/>` but the error is actually a fix for a bug that was introduced when the XML declaration stripping was added in April. Before this change, the error message was output but due to the stripping it would return an empty `<project_config/>` with the error message getting lost.

client/gui_rpc_server_ops.cpp
chausner: See me comment above.

chausner:
Christian, I incorporated your suggestions. Please have a look again.

client/gui_rpc_server_ops.cpp
```diff
@@ -804,7 +804,11 @@ void handle_lookup_account_poll(GUI_RPC_CONN& grc) {
             grc.lookup_account_op.error_num
         );
     } else {
-        grc.mfout.printf("%s", grc.lookup_account_op.reply.c_str());
+        const char *p = grc.lookup_account_op.reply.c_str();
+        const char *q = strstr(p, "<account_out"); 
+        if (!q) q = strstr(p, "<error");
+        if (!q) q = "<account_out/>\n";
```
ChristianBeer: The logic now is to look for `<account_out` if that is present return the string starting from there. If it is not present look for `<error` and return the string from there (which can produce malformed XML because you do not check if the closing tag matches). If this is also not found return an empty `<account_out/>`. This `<error>` output is also not consistent with the other two functions you mentioned. Do the other two RPCs not report an error?

chausner: The error reporting is inconsistent. `get_project_config_poll` returns errors like this:
```
<project_config>
    <error_num>%d</error_num>
</project_config>
```
No special handling required here. `lookup_account_poll` returns
```
<account_out>
  <error_num>%d</error_num>
</account_out>
```
in case of a client-side error (e.g. GetHostByName) and
```
<error>
  <error_num>%d</error_num>
  <error_msg>%s</error_msg>
</error>
```
in case of a server-side error (e.g. DBNotFound). `get_all_projects_list` again does not have any error reporting at all and returns an empty response in the unlikely case that all_projects_list.xml cannot be read.

This is why there's special handling of `<error>` in `lookup_account_poll`.

ChristianBeer: That makes sense now.

AenBleidd:
@ChristianBeer Could you please check whether this pull request can be merged? Thanks

ChristianBeer:
I already reviewed and signed of. I'm not responsible for the Client so I won't merge this. @davidpanderson

davidpanderson:
Thanks.

PR #1711, opened 2016-11-12
Fix issue #1519: MGR: Maximized window state is not restored on startup

AenBleidd:


AenBleidd:
Opened new pull request instead of #1701.

davidpanderson:
Thanks!

PR #1712, opened 2016-11-16
Update translations

ChristianBeer:


ChristianBeer:
I'm going to solve this in a different way.

PR #1713, opened 2016-11-16
Update translation system

ChristianBeer:
Until now we used a transifex-github bridge to manage translations and templates. This proved to be unreliable lately and also needs a special setup to be maintained. This PR eliminates the need for a special bridging software and uses transifex tools directly with minimal maintenance required. Major changes include:
* website translations (project-generic) are now stored in one place where they are also used (no manual copying of files anymore)
* the transifex client is used to pull down translation resources that are 100% translated and put's them in the correct place
* the updatetrans.sh script does all the steps to update translations in the repository (supposed to run as a cronjob)

What updatetrans.sh can't do right now is update templates and upload them to transifex. That will be done by another script.


marius311:
A user with Polish language setting was noticing some errors that ended up being caused by the `$PROJECT` that appears [here](https://github.com/BOINC/boinc/blob/f1f80726154e1bfa62e7b9260cbaf211b9698677/html/languages/translations/pl.po#L5033) which led me to notice this PR deletes the Polish translation all together. 

I was just curious, 

1. What's the reason for this? Should I delete it off my project too? (like, does it contain some bad translations or something?) 

2. I really don't know anything about how the translations work, but how secure is? Could a malicious translator insert something worse than a `$PROJECT` in there? 

Thanks.

ChristianBeer:
The translation was removed because it was outdated and not complete. If you have the warning about `$PROJECT` on your side you can remove it to fix that, I'm not aware of any other ill sideeffects. But as I already told the reporter of this problem, someone needs to translate all strings of the resource so it is imported into github again.

The translations are not interpreted (the translator can't execute php code) but there are some instances where we try to substitute `%1` style variables in the translated strings with HTML code. And PHP tries to be extra clever and substitute anything that looks like a variable name with the content of this variable. The scope should be limited but I didn't look into this deeply.

Issue #1689, opened 2016-10-26
Feature request: 'Apply' button on Event Log options dialog.


RichardHaselgrove:
'Apply' would configure and activate the new set of choices (as 'OK' does now), but leave the dialog open for reuse. This would enhance usability during debugging:
- When you want to study the output from one of the verbose options, but not leave it running continuously.
- When you're not quite sure which option would be most useful, and want to try alternatives.
- If you activate the dialog with Ctrl+Shift+F while viewing the Event Log window (which works - nice touch), the default positioning obscures the very thing you want to concentrate on. Leaving it open would allow you to park it on a different part of the screen while you work.

AenBleidd:
Fixed in PR #1715

RichardHaselgrove:
Many thanks to AenBleidd for that. Tested in a private build, and works a treat.

ChristianBeer:
Is the button in the right order on Windows? Would be a shame if we have a different ordering as the user expects.

RichardHaselgrove:
'Apply' buttons are rare in user applications in Windows, but are common in the operating system dialogs known as 'property sheets'. I think the order (OK, Defaults, Cancel, Apply) conforms to the guidance in https://msdn.microsoft.com/en-us/library/windows/desktop/bb774538(v=vs.85).aspx

The only thing missing is that the 'Apply' button would normally be disabled when the dialog is first opened, and only become active when one of the checkboxes is toggled - but that would be a lot of extra work to implement, and IMHO hardly worth it.

AenBleidd:
@ChristianBeer, Yes, this is the default button order in Windows.
@RichardHaselgrove, it would be not so easy to implement such behavior of 'Apply' button. I think this is not a critical issue.

RichardHaselgrove:
"not a critical issue": Agreed

JuhaSointusalo:
Something like this in the constructor:

```
btnApply->Disable();
Bind(wxEVT_CHECKBOX, [btnApply] (wxCommandEvent&) {
    btnApply->Enable();
});
```

Converting it to pre C++11 if left as an exercise for the reader.

AenBleidd:
@JuhaSointusalo, thank you for the advice. I've made a new commit.
@RichardHaselgrove, please check the new behavior.
@davidpanderson, please review this.

RichardHaselgrove:
Just got back in, so only a very quick test - but seems to be working perfectly. Thanks to all who contributed.

[...]

PR #1715, opened 2016-11-20
Fix issue #1689: Feature request: 'Apply' button on Event Log options…

AenBleidd:
… dialog.


ChristianBeer:
Since we don't have a strict policy on commit messages I didn't comment on yours. But now you are using confusing messages that do not correspond to the content of the commit. Could you please choose the format that is very widely used throughout this whole repository?

The structure is:
```
Component: short description what was fixed

Longer description why this change is needed, including any reference to other issues
```
The empty line after the first line is mandatory!

In your case those two commits should look like:
```
Manager: add 'Apply' button to Event Log options dialog

As requested in issue #1689
```
```
Manager: only activate 'Apply' button when needed

Activates 'Apply' button when the checkboxes change their state or 'Defaults' button is pressed. See #1689
```

AenBleidd:
Closed. New PR #1718 created with proper messages.
@ChristianBeer Thank you for your help.

PR #1718, opened 2016-11-22
Manager: add 'Apply' button to Event Log options dialog

AenBleidd:
Activates 'Apply' button when the checkboxes change their state or 'Defaults' button is pressed. 
As requested in issue #1689.


AenBleidd:
Opened instead of #1715

davidpanderson:
Thanks.

PR #1720, opened 2016-11-25
Make project_dont_detach_when_done RPC work correctly

chausner:
The dont_detach_when_done RPC does not work as expected, it does not reset the flag. Looks like no one has noticed since the bug was introduced 5 years ago in c61103ac26bff4a40023dbc6a92781247e8801d8. This should fix it.


davidpanderson:
Thanks!

PR #1724, opened 2016-11-28
Locale: refactor template update process

ChristianBeer:
* Automatically build templates from source files and push them to transifex. Templates are automatically created in the right place, no more manual copying files.
* Use a consistent header for all template files
* Only Android and Drupal are still updated on their own because this needs interactive tools.


davidpanderson:
I didn't look at this closely, but I want to make sure we have the same understanding.  There should be two template builds:
1) Done by us, out of the BOINC source tree, covering  the Manager, the project web code (html/*), the BOINC web site (doc/*.php), and installers (we're currently doing Mac, not sure about Win).  These may as well get put into 1 template; there may be some overlap of strings.
2) Done by each project, of their project-specific content, which may be in PHP or HTML files, or possibly in CMS content.  They need to have a way to indicate which files to scan.

Whatever we do, we need to update http://boinc.berkeley.edu/trac/wiki/TranslationSystem accordingly.

ChristianBeer:
@davidpanderson I didn't change any functionality. I combined the several snippets of code that individually generated the BOINC templates into a single script which also makes sure the templates end up in the correct place. There are still the same templates as before (Manager, Client, Mac Setup, project generic website and BOINC website).
Projects can still use `html/ops/build_po.php` to build a project specific template that we can use as a resource on transifex. If any project wants to do this I'm willing to help work out a workflow. So far Einstein@Home is using transifex directly through Drupal and we are satisfied with how this is done.

Thanks for the link, I changed several over places but missed this one. Updating it right now.

PR #1731, opened 2016-12-06
Bootstrap

davidpanderson:


davidpanderson:
Bootstrap code has been live on SETI@home for a couple of weeks.  No outstanding issues.

JuhaSointusalo:
> No outstanding issues.

Not quite so. [pre] contents with very long lines still causes all text to run off-screen. The fix has been in your inbox for a few days now (plus some other suggestions).

marius311:
@davidpanderson I'm in the process of upgrading C@H to use this. One thing I notice is many elements had various classes removed by this commit, e.g. https://github.com/BOINC/boinc/commit/c5b948020d096e3b800a71e7bf7ce44b98760b9c#diff-11913bc966d97652546e747b86673dc8L215. This makes it more difficult to customize them with CSS, for example before I had, 

```css
td.threadline {
     white-space: nowrap;
     max-width: 450px;
     overflow: hidden;
     text-overflow: ellipsis;
}
```

to make the thread title end with ellipses after a certain point, now there's no easy way to do that. 

What are your thoughts on adding back some of these classes, so at least they're there? (even if they're not used by the default bootstrap.css)

marius311:
I can make a PR adding back what I need, then maybe you'd want to look it over and decide to add back other things so that there's some reason to things that were added back, as opposed to just being the things I needed at the moment.

marius311:
I do share the sentiment of wanting to keep only the bootstrap classes, but any idea how to do it here? I can definitely capture td:nth-child(2) or something like that for the Thread column, but can't figure out how to do it only on the forum_forum page...

davidpanderson:
I limited with width of the title column to 35% and it looks OK;
long thread titles take up multiple lines, as in:
https://boinc.berkeley.edu/test/forum_forum.php?id=1
Do you think this is OK?

Can you send me URLs on C@h of other things that don't look good?

marius311:
How did you limit just the title column? I couldn't find anything in the CSS of that link. 

On my end, I'm trying to avoid the line wrapping and instead extend long titles with ellipses, e.g.
image

The above is made possible because I added back the old class, I'm fine keeping my own BOINC branch to do so, although in general I try hard not to diverge too much from the main repo, hence why I'm suggesting a PR. Although maybe you have ideas how to do it without?

davidpanderson:
functions like row_heading_array() take an optional argument of an array of
additional <th> attributes, e.g.:

    row_heading_array(
        array(
            "",
            tra("Threads"),
            tra("Posts"),
            tra("Author"),
            tra("Views"),
            "<nobr>".tra("Last post")."</nobr>"
        ),
        array("", "width=35%", "", "", "", "")

    );

This could also be used to add a style attribute.

My concept these days is to minimize customizability (which creates
complexity, e.g. you have to document it)
and instead do things that are acceptable to everyone.

Speaking of which, are you still on schedule to move to Berkeley in the
fall?
Do you have a place to live?
I don't have any suggestions, but could inspect places if it would help.

-- D

marius311:
Oh I see now how you did it, thanks. In any case, this would still require a custom branch of the BOINC code (which I'm trying to avoid as much as possible) since I'd need a modified forum_forum.php to pass the array argument. I need to think more about how I think the best way to handle this is, I'll get back to you.

PR #1734, opened 2016-12-08
Locale: Update scripts and templates

ChristianBeer:
This is the current state of automatically generating localization templates. I didn't check if I added all the relevant files in `doc/` after the bootstrap conversion. I can add some more before pushing the templates to transifex.


davidpanderson:
Thanks.

Issue #1730, opened 2016-12-06
Please allow enabling "Report results immediately" for selected app or project


sirzooro:
I have just noticed that WUProp task was completed and waited for at least minute in "ready to report" state before BOINC Client sent it back to server and downloaded new one (this project is an NCI-one). I do not want to enable `<report_results_immediately>` globally on that machine, it would interfere with cron-scheduled updates for WCG. Instead I would like to do this in app_config.xml. Please add support for `<report_results_immediately>` in that file. 

You can also think about adding new configuration section there which will allow to specify options for all project apps.  It could look in this way:

```xml
<app_config>
    <options>
        <report_results_immediately>1</report_results_immediately>
    </options>
</app_config>
```

Such configuration may be be also useful for participation in CPU-based challenges if you crunch GPU tasks too - you could enable this for only for CPU tasks (or project), and leave GPU one as-is.

Ageless93:
Not sure we'd want it there in that place? Should the app_config.xml file only handle _scheduling parameters for specific applications or app versions_ as it does now? All communications - uploading, downloading and reporting, but also trickling in either direction - of tasks is a function for the core client, not something for the individual science applications to do. So if anything, this should be a function of the cc_config.xml file and not the app_config.xml file. 

(My personal opinion in this)

ChristianBeer:
Fixed in master.

PR #1736, opened 2016-12-18
Fix issue #1730: Please allow enabling "Report results immediately" for selected app or project 

entibasse:
As suggested by  sirzooro, this fix adds a new section `<options>` in app_config.xml file.
In section  `<options>`, this fix adds option `report_results_immediately`


davidpanderson:
Why is this needed? i.e. why not use the global report_results_immediately?

sirzooro:
@davidpanderson please check #1730, I put two use cases for this new option there.

davidpanderson:
I added this feature in a simpler way.

PR #1743, opened 2016-12-30
Fixed compilation using MinGW from Cygwin 32 and 64 bit

sirzooro:


lib/diagnostics_win.h
```diff
@@ -27,6 +27,7 @@ typedef LONG       NTSTATUS;
 typedef LONG       KPRIORITY;
 
 //MinGW-W64 defines this struct in its own header
+#define HAVE_CLIENT_ID 1
```
AenBleidd: I'm not sure that this macro is good because it can be the case when structs will be indefined.
Maybe we need something like this?
#if !defined(__MINGW32__) and !defined(HAVE_VM_COUNTERS)

lib/diagnostics_win.h
```diff
@@ -35,6 +36,7 @@ typedef struct _CLIENT_ID {
 #endif
 
 //MinGW-W64 defines this struct in its own header
+#define HAVE_VM_COUNTERS 1
```
AenBleidd: Same here:
#if !defined(__MINGW32__) and !defined(HAVE_VM_COUNTERS)

lib/diagnostics_win.h
```diff
@@ -67,6 +69,7 @@ typedef struct _VM_COUNTERS {
 #endif
 
 //MinGW-W64 defines this struct in its own header
+#define HAVE_SYSTEM_THREADS 1
```
AenBleidd: Same here:
#if !defined(MINGW32) and !defined(HAVE_SYSTEM_THREADS)

lib/diagnostics_win.h
```diff
@@ -112,6 +115,7 @@ typedef struct _SYSTEM_PROCESSES {
 #endif
 
 //MinGW-W64 defines this struct in its own header
+#define HAVE_THREAD_STATE 1
```
AenBleidd: Same here:
#if !defined(MINGW32) and !defined(HAVE_THREAD_STATE)

sirzooro:
Done. I also added one more change which could break MSVC build, it may have `set_*` functions in global namespace only instead of `std`.

lib/diagnostics_win.cpp
```diff
@@ -312,7 +312,7 @@ int diagnostics_update_thread_list() {
             // Enumerate the threads
             for(uiSystemIndex = 0; uiSystemIndex < pProcesses->ThreadCount; uiSystemIndex++) {
                 pThread = &pProcesses->Threads[uiSystemIndex];
-                pThreadEntry = diagnostics_find_thread_entry((DWORD)pThread->ClientId.UniqueThread);
+                pThreadEntry = diagnostics_find_thread_entry((DWORD)(uintptr_t)pThread->ClientId.UniqueThread);
```
AenBleidd: Double type cast looks bad. Why do we need it here?

lib/diagnostics_win.cpp
```diff
@@ -326,12 +326,12 @@ int diagnostics_update_thread_list() {
                     hThread = OpenThread(
                         THREAD_ALL_ACCESS,
                         FALSE,
-                        (DWORD)(pThread->ClientId.UniqueThread)
+                        (DWORD)(uintptr_t)(pThread->ClientId.UniqueThread)
```
AenBleidd: Same double type cast here.

lib/diagnostics_win.cpp
```diff
@@ -326,12 +326,12 @@ int diagnostics_update_thread_list() {
...
                     );
 
                     pThreadEntry = new BOINC_THREADLISTENTRY;
                     diagnostics_init_thread_entry(pThreadEntry);
-                    pThreadEntry->thread_id = (DWORD)(pThread->ClientId.UniqueThread);
+                    pThreadEntry->thread_id = (DWORD)(uintptr_t)(pThread->ClientId.UniqueThread);
```
AenBleidd: Same double type cast here.

PR #1762, opened 2017-01-21
Android: Fixed and reduced string comparisons

obsproth:
1. Replaced some String `==` comparisons with `equals`
This is just a simple fix.

2. Simplified power source prefs dialog code by adding `ID` field to `SelectionDialogOption`
Now IDs are used to identify power source options, instead of translated strings.


davidpanderson:
Thanks!

PR #1764, opened 2017-01-24
Always use config.h in str_replace.h

kyrsjo:
So that we always get the correct definition of HAVE_-flags such has HAVE_STRCASECMP.
Without this, builds with MSYS2/CYGWIN64 fail to link (you'll get a "double definition").

I still have some other problems with it, but at least this fixes some of them...


ChristianBeer:
fixed by #1790

Issue #1767, opened 2017-01-25
Check for new version


AenBleidd:
Add functionality to check for new version.
Show message about available version and download link.

AenBleidd:
Fixed in PR #1768 .

ChristianBeer:
This is already implemented but wording needs to be adjusted.

Issue #28, opened 2015-02-03
Check for new BOINC update


romw:
**Reported by KSMarksPsych on 13 Apr 37296509 21:20 UTC**
Posted by: elst93
Date: 8:12 PM 05-11-2005
Would it be possible to add a check for update in the Manager's Help menu?

Many applications have this feature these days. 

Migrated-From: http://boinc.berkeley.edu/trac/ticket/28

romw:
**Commented by Ageless on 9 Feb 38016773 12:53 UTC**
Added in 5.8 as an automatic function.

romw:
**Commented by Ageless on 7 Sep 38550783 00:26 UTC**
Could there be some kind of indicator for this in the GUI, not just in the message stream, where it will get lost or ignored? Perhaps with the icon in the tray changing colors, the way Windows Update does? 

Or could the new 6.4 or above versions auto-update themselves? [[br]]
Possibly download the new version when there's an internet connection and flag it available for install upon the next restart of BOINC, not unlike Firefox and Adobe Acrobat Reader do? Or is this still considered a security risk?

romw:
**Commented by Eric Myers on 17 Jul 38561377 17:20 UTC**
This feature needs careful thought about the user controls, and it HAS to be under user control for a wide range of use cases.

Some people won't want auto-update.  They need an easy way to turn it off.

Some people only want auto-update -- set it and forget it.   They need an easy way to turn it on.

Some people will want to do updates manually, but they will appreciate an easy GUI interface, for that -- easier than the initial download and install.   They will want some kind of indication that there is a new update (not just in the message stream), and/or a way they can take an action to check for new updates (or verify that they are up to date).

As already mentioned, this exists for other software in the form of Windows Update, FireFox and Acrobat.  Add Apple Software Update to the list.  We need to design the interface so that it works well  (Apple and Firefox do, IMHO, Acrobat does not).  Windows Update is widely familiar so perhaps worth emulating, at least in general?

There has to be user control over this functionality or the people who don't like however it's done will be annoyed.

romw:
**Commented by Nicolas on 14 Dec 38647846 23:06 UTC**
Something worth emulating from Windows Updates is the configuration options:
 * Download and install automatically.
 * Download and notify, but ask before installing.
 * Notify, but ask before downloading or installing.
 * Disable updates.

romw:
**Commented by romw on 25 Jan 38648356 00:53 UTC**
[...]

PR #1768, opened 2017-01-25
MGR: Client version check

AenBleidd:
Reused existing mechanism of version check.
Manager will check for new version during connection to the client.
Default message about new version will be shown in notices tab.
Added menu to check new version manually. In this case user will see the message with check result.
Fixed issue with new version response parse.

### DIFF
--- a/clientgui/MainDocument.cpp
+++ b/clientgui/MainDocument.cpp
@@ -164,6 +164,7 @@ void CNetworkConnection::Poll() {
             if (!retval) {
                 wxLogTrace(wxT("Function Status"), wxT("CNetworkConnection::Poll - Connection Success"));
                 SetStateSuccess(m_strNewComputerName, m_strNewComputerPassword);
+                m_pDocument->CheckForVersionUpdate();
             } else if (ERR_AUTHENTICATOR == retval) {
                 wxLogTrace(wxT("Function Status"), wxT("CNetworkConnection::Poll - RPC Authorization - ERR_AUTHENTICATOR"));
                 SetStateErrorAuthentication();
@@ -1278,6 +1279,21 @@ bool CMainDocument::IsUserAuthorized() {
     return true;
 }
 
+void CMainDocument::CheckForVersionUpdate() {
+    std::string version, url;
+    // It seems that this check will work if the client will check newer file version on start.
+    // Client will check this even the last check was more that 14 days ago and on start only.
+    // Client newer saves this information in its internal files 
+    // and never reread 'get_current_version.xml' between checks.
+    // In any other cases result will be empty.
+    rpc.get_newer_version(version, url);
+
+    if (!version.empty() && !url.empty()) {
+        char message[MAXPATHLEN];
+        snprintf(message, sizeof(message), "%s: %s", wxT("You can download it here"), url);
+        wxGetApp().SafeMessageBox(message, wxT("A new version of BOINC is available."));
+    }
+}
 
 int CMainDocument::CachedProjectStatusUpdate(bool bForce) {
     int     i = 0;
--- a/clientgui/MainDocument.cpp
+++ b/clientgui/MainDocument.cpp
@@ -1281,11 +1281,6 @@ bool CMainDocument::IsUserAuthorized() {
 
 void CMainDocument::CheckForVersionUpdate() {
     std::string version, url;
-    // It seems that this check will work if the client will check newer file version on start.
-    // Client will check this even the last check was more that 14 days ago and on start only.
-    // Client newer saves this information in its internal files 
-    // and never reread 'get_current_version.xml' between checks.
-    // In any other cases result will be empty.
     rpc.get_newer_version(version, url);
 
     if (!version.empty() && !url.empty()) {
--- a/clientgui/MainDocument.cpp
+++ b/clientgui/MainDocument.cpp
@@ -1282,12 +1282,6 @@ bool CMainDocument::IsUserAuthorized() {
 void CMainDocument::CheckForVersionUpdate() {
     std::string version, url;
     rpc.get_newer_version(version, url);
-
-    if (!version.empty() && !url.empty()) {
-        char message[MAXPATHLEN];
-        snprintf(message, sizeof(message), "%s: %s", wxT("You can download it here"), url);
-        wxGetApp().SafeMessageBox(message, wxT("A new version of BOINC is available."));
-    }
 }
 
 int CMainDocument::CachedProjectStatusUpdate(bool bForce) {
--- a/clientgui/MainDocument.cpp
+++ b/clientgui/MainDocument.cpp
@@ -1284,7 +1284,7 @@ void CMainDocument::CheckForVersionUpdate(bool showMessage) {
     std::string version, url;
     wxString message, title;
     title.Printf(_("Version Update"));
-    std::string applicationName = wxGetApp().GetSkinManager()->GetAdvanced()->GetApplicationName();
+    wxString applicationName = wxGetApp().GetSkinManager()->GetAdvanced()->GetApplicationName();
     if (IsConnected()) {
         rpc.get_newer_version(version, url);
 
@@ -1299,7 +1299,7 @@ void CMainDocument::CheckForVersionUpdate(bool showMessage) {
         }
     }
     else {
-        message.Printf("%s is not connected to the client", applicationName.c_str());
+        message.Printf("%s is not connected to the client", applicationName);
     }
     if (showMessage) {
         wxGetApp().SafeMessageBox(message, title);
--- a/clientgui/MainDocument.cpp
+++ b/clientgui/MainDocument.cpp
@@ -37,6 +37,7 @@
 #include "BOINCTaskBar.h"
 #include "DlgEventLog.h"
 #include "Events.h"
+#include "SkinManager.h"
 
 #ifndef _WIN32
 #include <sys/wait.h>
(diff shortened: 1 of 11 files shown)


AenBleidd:
Fixed issue #1767 .

AenBleidd:
@davidpanderson, @ChristianBeer, Could you please review this? Thanks

adamradocz:
It would be very nice an automatic update option as well which works like Windows Update. Download the update automatically and install it silently.
Is it possible?

AenBleidd:
@adamradocz, it is very dangerous to update app automatically. You can read more conversation here: #28

davidpanderson:
Not sure we want version check that often but OK.

AenBleidd:
@davidpanderson, I think check for new version after connect to client is not too often. It is the way many application do their checks for new version.

JuhaSointusalo:
I haven't tried this yet but reading through the changes I don't think this works quite as intended.

When user clicks the Check new version menu item the chain of calls goes through `handle_get_newer_version()` and ends up in `GET_CURRENT_VERSION_OP::do_rpc()` . Contrary to its name, `do_rpc()` doesn't actually do the RPC, it just queues it and returns. `handle_get_newer_version()` then doesn't wait for the version download to complete and returns the result of previous check.

In other words, user can be told that there is no new version and a couple of seconds later a notice pops up saying there is a new version. Bad.

The client may be busy with some other RPC, or BOINC server may be super slow, or the user's network could be down. So the version check may take a long time to complete. You can't block the client or GUI while waiting for reply so you'll need to make the operation asynchronous by adding a new GUI RPC command to poll the status of the version check, similar to, e.g.,  get_project_config / get_project_config_poll . A Please wait message box would be nice so that the user knows that stuff is happening.

By default the client limits simultaneous file transfers to 8. I don't know if downloading the version info is included but if it is you need to add some mechanism that overrides the limit when the user requested the version check. If the client is in the middle of downloading a new app version it can take hours before the transfer queue is empty.

The client has daily schedules and usage limits to control file transfers. The version info is not much to download but if it isn't too hard to implement I'd like for the automatic version check to obey the network preferences. This is mostly on the grounds that if the client is told to (not) do something then it should (not) do it.

clientgui/MainDocument.cpp
```diff
@@ -1278,6 +1280,31 @@ bool CMainDocument::IsUserAuthorized() {
...
+    title.Printf(_("Version Update"));
+    wxString applicationName = wxGetApp().GetSkinManager()->GetAdvanced()->GetApplicationName();
+    if (IsConnected()) {
+        rpc.get_newer_version(version, url);
+
+        if (!showMessage)
+            return;
+
+        if (!version.empty() && !url.empty()) {
+            message.Printf("%s: %s", _("A new version of BOINC is available for downloading here"), url);
```
JuhaSointusalo: Shouldn't that be the branded app name?

AenBleidd:
@JuhaSointusalo, thank you for your detailed review. I've tried to do minimum changes because of the compatibility with old clients. It was my main motivation, I have plans to continue my work on this case and implement it more accurate. I'll try to take into consideration all your suggestions and make version check more flexible and more stable.

ChristianBeer:
I think we also need to review the wording of messages. Some of them don't feel like proper English to me. As I'm also not a native English speaker I would like someone else to take a look at it.

CharlieFenton:
There is another issue with the version check. If a user has downloaded a "branded" (skinned) version of BOINC from World Community Grid, GridRepublic, etc., should BOINC be checking their web site for a new version with the same branding rather than Berkeley? If so, then it should also point them to that web site to download a new version. I believe that this is true of the existing version check in the Client as well as the manager's new version check.

AenBleidd:
@CharlieFenton, client uses link from xml so it will be default link for
'vanilla' BOINC and another for each 'branded' version.

Best regards,
Vitalii Koshura

2017-08-18 15:12 GMT+03:00 CharlieFenton <notifications@github.com>:

CharlieFenton:
I cannot find any place in the code where client_version_check_url in the cc_config.xml file is set to the branded version's web site, such as GridRepublic or World Community Grid. So it appears to me that the version check will always be checking for a new version of unbranded BOINC from boinc.berkeley.edu, even when the menu item says "Check for new GridRepublic version". In fact, I don't see any item in the branded skin.xml files with a URL for checking the current version, and the only place in the code which uses the organization_website is in the About dialog.

Can you or @davidpanderson explain to me how a version check is sent to GridRepublic, World Community Grid, etc.?

If I understand correctly, this is further complicated because WCG provides a branded BOINC installer for MS Windows but not for Macintosh, so they have Mac users download from boinc.berkeley.edu but Windows users download from WCG.

AenBleidd:
@CharlieFenton, I'm not sure that client_version_check_url should be
configured for branded version by the code. I think this option should be
configured by the installer. But this is my opinion only.

Thanks

Best regards,
Vitalii Koshura

2017-08-19 3:21 GMT+03:00 CharlieFenton <notifications@github.com>:

CharlieFenton:
Again, I am not certain, but I believe that, except for the names of the installer and Manager, all of the brand-specific customization is provided by the skin.xml file.

davidpanderson:
Do we want to continue supporting branded versions?

CharlieFenton:
I too have wondered whether we should continue supporting branded versions. I suppose we should check with folks at GridRepublic and WCG. I think those are the only two still active, though I'm not sure about the status of CharityEngine or ProgressThruProcessors.

TheAspens:
We (WCG) do want to continue our branded version.  Our branding in the past of consisted of the following:

1. Customized windows installer for things such as text shown during installation, the splash logo, inclusion of the project_init.xml to initiate attach to WCG, inclusion of cc_config.xml with a few different default options and the application in windows is seen as World Community Grid rather than BOINC.
2.  Skin customization
3.  Custom compiled binaries so that the icons displayed in windows render our logo.  Also, there is one extra panel during the "Attach to Project" that explains the different between a BOINC project (http://boinc.berkeley.edu/projects.php) and a WCG Project (https://www.worldcommunitygrid.org/research/viewAllProjects.do).

I have about 3-4 weeks of work of things to do and then I will be working on adopting a new version of the client (to go from 7.2 -> 7.10).    We will having one of the IBM teams do the security review and then going through our client release testing (so we hope to align with BOINC 7.10 release timeframe although we tend to be slower).

We are trying to figure out how to simplify the branded configuration vs the standard BOINC client.   What aspects of custom branding are you wanting to remove?

TheAspens:
> I cannot find any place in the code where client_version_check_url in the cc_config.xml file is set to the branded version's web site, such as GridRepublic or World Community Grid. So it appears to me that the version check will always be checking for a new version of unbranded BOINC from boinc.berkeley.edu, even when the menu item says "Check for new GridRepublic version". In fact, I don't see any item in the branded skin.xml files with a URL for checking the current version, and the only place in the code which uses the organization_website is in the About dialog.

WCG includes a version of the cc_config.xml in our customized installer which sets this value.  I'm attaching what is in our 7.2 client below
```
<?xml version="1.0"?>
<cc_config>
    <log_flags></log_flags>
    <options>
        <client_download_url>http://www.worldcommunitygrid.org/reg/ms/viewDownloadAgain.do</client_download_url>
        <client_version_check_url>http://www.worldcommunitygrid.org/download_all.php?xml=1</client_version_check_url>
        <network_test_url>http://www.ibm.com</network_test_url>
    </options>
</cc_config>
```

mblumberg:
Branded clients are mission-critical for us (GidRepublic, Charity Engine, Progress Thru Processors).

This functionality makes it possible to maintain a consistent user experience -- e.g. when a user sees an ad, comes to a website, downloads and runs and installer, etc, he or she gets constantly assured they are still in the same ecosystem, that the app they are installing is the one they meant to install, etc.

If any work is needed to update or maintain these features, we'd be glad to contribute.  (Kevin, if you want to chat about this, let me know.)

CharlieFenton:
Although the Mac Manager and screen saver still fully support branding, the Mac installer has gone through extensive changes since the last GridRepublic version (which was a very long time ago as far as I know.) The current Mac installer needs significant work to allow for branding of recent versions of BOINC.

mblumberg:
Thanks, Charlie.  Do you mean _{1}_ GR specifically faces a lot of work generating the new graphics and xml required by the new format, because our custom files are out of date?  Or _{2}_ Generally speaking, there's significant [technical] work remaining to get the Mac installer to the place where a skin could be added by anyone?  (Ie the mac version of the installer doesn't effectively support skins at this time?)

CharlieFenton:
Neither #1 nor #2. The scripts to generate the custom Mac installer for all branded builds are very much out of date and it will take significant work to create new scripts to generate branded installers for the Mac. Adding the skins is the easy part. The way the installer is built has changed considerably since the 7.0 series due, largely because of changes made by Apple to their development tools. 

And anyone can add skins to their copy of the Mac BOINC Manager after it is installed.

mblumberg:
Charlie, is there any up-to-date documentation of the process of building Mac installers?  -- E.g. something like the BOINC Wiki article [Building a Windows installer for the BOINC client](https://boinc.berkeley.edu/trac/wiki/AdminInstallerWin)?

CharlieFenton:
There  is a slightly out of date version of the build instructions at [http://boinc.berkeley.edu/trac/wiki/MacBuild]() which I will try to update soon. The latest version of this document (corresponding to any particular version of BOINC) can be found in the source tree at _mac_build/HowToBuildBOINC_XCode.rtf_. You can access the source code tree by following the instructions at [http://boinc.berkeley.edu/trac/wiki/SourceCodeGit]().

Please note that the issue is not with these instructions. Building a branded version will require making a special version of  the script _mac_installer/release_boinc.sh_ for each brand (GridRepublic, Charity Engine, ProgressThruProcesors). The very old and obsolete special scripts are still at _mac_installer/make_GridRepublic.sh_, _mac_installer/make_CharityEngine.sh_ and _mac_installer/make_ProgThruProc.sh_. Significant changes to these 3 scripts are needed. For example, they use PackageMaker, which has not been available for 5 years or so.

Issue #1777, opened 2017-02-01
World Community Grid keeps override BOINC prefs


AndCycle:
I recently discover the global_prefs.xml been overwrite by BAM and WCG by turn,
BOINC can't decide which one to use, as it just take whatever any project says about prefs.

here is the thread
https://boincstats.com/en/forum/9/11379,1

```
01-Feb-2017 08:31:54 [World Community Grid] Starting task MCM1_0130025_1109_0
01-Feb-2017 08:31:56 [World Community Grid] Started upload of MCM1_0130019_3555_1_r295879939_0
01-Feb-2017 08:32:03 [World Community Grid] Finished upload of MCM1_0130019_3555_1_r295879939_0
01-Feb-2017 08:48:10 [---] Contacting account manager at https://bam.boincstats.com/
01-Feb-2017 08:48:14 [---] Account manager: BAM! User: 34887, AndCycle
01-Feb-2017 08:48:14 [---] Account manager: BAM! Host: 531829
01-Feb-2017 08:48:14 [---] Account manager: Number of BAM! connections for this host: 20485
01-Feb-2017 08:48:14 [---] Account manager contact succeeded
01-Feb-2017 08:48:14 [---] General prefs: from http://bam.boincstats.com/ (last modified 30-Jan-2017 05:07:35)
01-Feb-2017 08:48:14 [---] Host location: none
01-Feb-2017 08:48:14 [---] General prefs: using your defaults
01-Feb-2017 08:48:14 [---] Preferences:
01-Feb-2017 08:48:14 [---]    max memory usage when active: 19115.83MB
01-Feb-2017 08:48:14 [---]    max memory usage when idle: 28673.74MB
01-Feb-2017 08:48:14 [---]    max disk usage: 100.00GB
01-Feb-2017 08:48:14 [---]    suspend work if non-BOINC CPU load exceeds 50%
01-Feb-2017 08:48:14 [---]    (to change preferences, visit a project web site or select Preferences in the Manager)
01-Feb-2017 08:48:17 [rosetta@home] Sending scheduler request: To report completed tasks.
01-Feb-2017 08:48:17 [rosetta@home] Reporting 2 completed tasks
01-Feb-2017 08:48:17 [rosetta@home] Not requesting tasks: don't need
01-Feb-2017 08:48:20 [rosetta@home] Scheduler request completed
01-Feb-2017 08:48:20 [World Community Grid] General prefs: from World Community Grid (last modified 29-Jan-2017 23:46:07)
[...]

PR #1778, opened 2017-02-01
client: only take global prefs from project when not attached to AM

AndCycle:
as the current scheduler take whatever global pref reply as is,
it should at least only take it if we didn't join any AM,

this is just a quick fix for #1777, 
there are room for further consideration about rewrite how we should take global pref.

### DIFF
--- a/client/cs_scheduler.cpp
+++ b/client/cs_scheduler.cpp
@@ -676,9 +676,10 @@ int CLIENT_STATE::handle_scheduler_reply(
     }
 
     // if the scheduler reply includes global preferences,
+    // and we currently don't use an account manager
     // insert extra elements, write to disk, and parse
     //
-    if (sr.global_prefs_xml) {
+    if (!gstate.acct_mgr_info.using_am() && sr.global_prefs_xml) {
         // skip this if we have host-specific prefs
         // and we're talking to an old scheduler
         //


client/cs_scheduler.cpp
```diff
@@ -676,9 +676,10 @@ int CLIENT_STATE::handle_scheduler_reply(
     }
 
     // if the scheduler reply includes global preferences,
+    // and we doesn't join any account manager,
```
ChristianBeer: small but important change needed: "and we currently don't use an account manager"

AndCycle:
@ChristianBeer thanks for the correction for my bad english.

AenBleidd:
I think this PR can be closed because @davidpanderson made similar changes.

ChristianBeer:
I just want to reference 0e4b412 which highlights a problem I didn't see with this PR. We can't simply ignore preferences if an AM is used we still need to check where those new preferences come from as the AM preferences are handled by the same function and we must allow new preferences from the AM to be stored.

AndCycle:
@ChristianBeer I though AM global pref is handled by `ACCT_MGR_OP::handle_reply` ?
do I get it wrong?
there are only two entry use `save_global_prefs`.

AndCycle:
just tested https://github.com/BOINC/boinc/commit/0e4b412d19f63856402ad1e18733d728a33aa101 ,
and no, it can't fix the symptoms.

AndCycle:
kind of mysterious because at a glance there is nothing wrong in 0e4b412 , why WCG global pref can slip through it.

AndCycle:
```
 !strcmp(global_prefs.source_project, gstate.acct_mgr_info.master_url)
```
this line doesn't make sense as it compare to it's current use one rather than data from sr

davidpanderson:
I committed a fix which takes into account:
- not all account managers provide prefs
- BAM! mixed http/https, which made things a bit tricky.

PR #1786, opened 2017-02-13
Client: normalize whitespace in p_model and p_vendor entries

ChristianBeer:
This was already done for Mac but not Linux and Windows. It collapses multiple consecutive blanks into one blank which is then send to the server.


davidpanderson:
Thanks!

PR #1789, opened 2017-02-14
Manager: Fix select computer dialog

### LINKED ISSUE #1306: Boinc manager can't connect to client using "localhost"
**Reported by saldsl on 10 Sep 44020772 12:41 UTC**
Tested on Fedora 19 x86_64 (with RPMs from repository), boinc version 7.2.33.
Without linking the gui_rpc_auth.cfg file to my home directory, when I select Advanced -> Select Computer and I type "localhost" as target and put the password Boinc Manager can't connect to client. If I type "127.0.0.1" it connects.

This happens only with 7.2.33, if I downgrade boinc to 7.0.65 "localhost" works, so I don't think to a firewall/hostnames configuration problem.

[https://bugzilla.redhat.com/show_bug.cgi?id=1048545]

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1335

ChristianBeer:
The hostname "localhost" was recognized as a special name which always prompted the Manager to try to read contents of gui_rpc_auth.cfg. The Manager used an empty password if the file is not readable but then can't connect because the Client actually uses a password. Any user supplied password was overwritten which made the dialog not work when using "localhost" but work when used with "127.0.0.1".
Now the dialog checks if the user has entered an empty password which prompts reading gui_rpc_auth.cfg and uses the user supplied password otherwise.

Fixes #1306

### DIFF
--- a/clientgui/AdvancedFrame.cpp
+++ b/clientgui/AdvancedFrame.cpp
@@ -1434,36 +1434,26 @@ void CAdvancedFrame::OnSelectComputer(wxCommandEvent& WXUNUSED(event)) {
     wxString            password = wxEmptyString;
     CMainDocument*      pDoc = wxGetApp().GetDocument();
     long                lRetVal = -1;
+    bool                bRetrievePasswordFromFile = FALSE;
 
     wxASSERT(pDoc);
     wxASSERT(wxDynamicCast(pDoc, CMainDocument));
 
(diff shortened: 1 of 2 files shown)


davidpanderson:
Thanks!

Germano0:
https://github.com/BOINC/boinc/issues/1306#issuecomment-286369822

PR #1790, opened 2017-02-16
Fix cross-compile from ubuntu with mingw

renyidong:
I was cross compiling from ubuntu xenial with mingw for windows.
A few error occurred because current code failed to recognize mingw.
There might be better ways(that I don't know) to fix it.

PR #1793, opened 2017-02-17
Build: try to build libraries with mingw on travis-ci

ChristianBeer:
Installs dependencies and runs the custom Makefile using the correct cross compiler.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,15 @@ addons:
     - libsqlite3-dev
     - libgtk2.0-dev
     - libwebkitgtk-dev
+    - mingw-w64
+    - binutils-mingw-w64-i686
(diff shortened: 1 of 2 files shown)


ChristianBeer:
The failed test is anticipated and will be fixed by #1790 which should be tested by restarting the travis-ci build for #1790 after this PR gets merged.

JuhaSointusalo:
I'm a bit surprised that there is a separate makefile for MinGW. I would have expected a normal Autotools build with configure and all that.

ChristianBeer:
@JuhaSointusalo I guess the answer is "historic reasons", cross compilation support was not that great some years ago. The good thing is we now have a testcase and someone can develop an autotools solution and test it immediately.

PR #1796, opened 2017-02-26
Check execinfo.h availability in configure.ac

ffontaine:
Currently, execinfo.h is included in lib/diagnostics.cpp if `__GLIBC__` is
defined. However, this does not work when cross-compiling boinc with
uclibc-ng. So, instead, check the presence of execinfo.h in configure.ac
and update lib/diagnostics.cpp accordingly.

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


davidpanderson:
Thanks!

PR #1797, opened 2017-02-26
Build CI: restructure dependency builds for linux

ChristianBeer:
This reduces the size of files to be cached between builds by the CI system.

DO NOT MERGE!


ChristianBeer:
New PR coming soon.

PR #1805, opened 2017-03-08
Added pagination and first level caching to per_app_list

nikolasgianna:
Implemented pagination and first level caching in per_app_list.php

### DIFF
--- a/html/user/per_app_list.php
+++ b/html/user/per_app_list.php
@@ -125,7 +125,7 @@ $offset = get_int('offset', true);
 if (!$offset) $offset=0;
 if ($offset % $items_per_page) $offset = 0;
 $x = $is_team?"teams":"participants";
-page_head("Top $x by application");
+page_head(tra("Top %1 by application", $x));
 $apps = BoincApp::enum("deprecated=0");
 if (!$appid) {
     $appid = $apps[0]->id;
@@ -168,7 +168,7 @@ if ($offset > 0) {
     echo "<a href=per_app_list.php?appid=$appid&is_team=$is_team&is_total=$is_total&offset=$new_offset>".tra("Previous %1", $items_per_page)."</a> &middot; ";
 }
 
-if ($n==$items_per_page) {
+if ($n == $items_per_page) {
     $new_offset = $offset + $items_per_page;
     echo "<a href=per_app_list.php?appid=$appid&is_team=$is_team&is_total=$is_total&offset=$new_offset>".tra("Next %1", $items_per_page)."</a>";
 }
--- a/html/user/per_app_list.php
+++ b/html/user/per_app_list.php
@@ -53,7 +53,7 @@ function col_title($is_team, $app, $appid, $is_total, $i) {
 function show_header($is_team, $apps, $appid, $is_total) {
     echo "<tr><th colspan=2>&nbsp;</th>";
     foreach ($apps as $app) {
-        echo "<th colspan=2>$app->name</th>\n";
+        echo "<th colspan=2 class=\"text-center\">$app->name</th>\n";
     }
     echo "</tr>";
 
@@ -62,7 +62,7 @@ function show_header($is_team, $apps, $appid, $is_total) {
     foreach ($apps as $app) {
         for ($i = 0; $i < 2; $i++) {
             $x = col_title($is_team, $app, $appid, $is_total, $i);
-            echo "<th>$x</th>\n";
+            echo "<th class=\"text-right\">$x</th>\n";
         }
     }
 
--- a/html/user/per_app_list.php
+++ b/html/user/per_app_list.php
@@ -26,7 +26,7 @@
 require_once("../inc/util.inc");
 require_once("../inc/team.inc");
 
-check_get_args(array("is_team", "appid", "is_total", "offset"));
+check_get_args(array("is_team", "appid", "is_total", "added_offset", "up", "offset"));
 
 define ('ITEM_LIMIT', 10000);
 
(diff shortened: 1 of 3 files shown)


AenBleidd:
@davidpanderson Please take a look at this PR. It seems that it is similar to the changes you've made in  ee3f8f1aa232a615f40c2dbee9047e2fb6551b83

davidpanderson:
Oops!  I didn't see this before my commit, which does essentially the same thing.

lfield:
Please note that this also implements caching which is also required.

html/user/per_app_list.php
```diff
@@ -96,34 +107,72 @@ function show_row($item, $apps, $is_team, $i) {
...
 
 $is_team = get_int('is_team', true);
 $appid = get_int('appid', true);
 $is_total = get_int('is_total', true);
+$items_per_page = 20;
+$offset = get_int('offset', true);
+if (!$offset) $offset=0;
+if ($offset % $items_per_page) $offset = 0;
+$x = $is_team?"teams":"participants";
+page_head("Top $x by application");
```
ChristianBeer: Can you make this translatable? Something like `tra("Top %1 by application", $x)` should work.

html/user/per_app_list.php
```diff
@@ -96,34 +107,72 @@ function show_row($item, $apps, $is_team, $i) {
...
+}
+
+end_table();
+
+if ($offset > 0) {
+    $new_offset = $offset - $items_per_page;
+    echo "<a href=per_app_list.php?appid=$appid&is_team=$is_team&is_total=$is_total&offset=$new_offset>".tra("Previous %1", $items_per_page)."</a> &middot; ";
+}
+
+if ($n==$items_per_page) {
```
ChristianBeer: Please put spaces around the `==` operator.

html/user/per_app_list.php
```diff
@@ -63,13 +67,20 @@ function show_header($is_team, $apps, $appid, $is_total) {
 // show a user or team, with their credit for each app
 //
 function show_row($item, $apps, $is_team, $i) {
```
ChristianBeer: There is still a DB lookup in this function which is not cached.

nikolasgianna: The reason first level caching was mentioned on the title of the pull request is because only the choice of items and the chosen app credit for each of them is cached. A DB lookup still has to be done for the credit of the other apps' credit and caching that too would be a possible future improvement.

ChristianBeer: I understand. But then the list will not be sorted. The sorting is done in the query that gets cached but the actual value is taken directly from the DB. This will lead to users that have a higher value listed after users with a lower value. This is perceived as a bug by the user. So I think we should cache everything right away in `get_top_items()`.

nikolasgianna: Ok, so I will look into implementing full caching. Are there any suggestions or ideas on how to best do so?

ChristianBeer: Rough sketch: Loop over the array returned by `get_top_items` and query credit and RAC per app and add this to `$data` (so it ends up in `$item` later). Then cache `$data`. Adjust the `foreach ($apps as $app) {` loop in `show_row` to not query the DB but look in `$item` for credit and RAC.

davidpanderson:
We didn't have the right indices on the credit_team and credit_user tables, so queries for high-score items were slow.  I added the indices.  With this change, caching may not be necessary.

nikogianna:
I added full caching and some checks in case there is some entry in the credit_user or credit_team table that doesn't correspond to an entry in the user or team table respectively.

nikolasgianna:
Implemented full caching and added checks in case an entry in the credit_team or credit_user tables doesn't correspond to an entry in the team or user tables respectively.

html/user/per_app_list.php
```diff
@@ -56,74 +60,319 @@ function show_header($is_team, $apps, $appid, $is_total) {
...
         }
-        echo "<td align=right>".format_credit($c->expavg)."</td><td align=right>".format_credit_large($c->total)."</td>\n";
-    }
-    echo "</tr>\n";
+
+       return $data;
+
+}
+
+function choose_users($x, $appid, $used_offset, $up) {
```
ChristianBeer: This and `choose_teams()` is compensating for the lack of a proper SQL JOIN clause implemented in `BoincCreditUser` for this use-case with several DB calls to eliminate old entries (mainly for the teams). Thus it would make more sense to me to write a custom `BoincCreditUser::get_list($where_clause, $order_clause, $limit)` function which does something like 
```
SELECT * from user u join user_credit cu on u.id=cu.userid where $where_clause order by $order_clause limit $limit
```
This will eliminate the need to check if a user or team still exists (because that is not part of the returned list) reducing the amount of queries per call.

html/user/per_app_list.php
```diff
@@ -56,74 +60,319 @@ function show_header($is_team, $apps, $appid, $is_total) {
...
+     global $store;
+     $x = 1;
+     $c = 1;
+   
+     //Store the sign of each team in the first column of an array
+     // 
+     foreach ($data as $item) {
+                $team = BoincTeam::lookup_id($item->teamid);
+                if (is_object($team)) {
+                   $sign = "<td>".team_links($team)."</td>";
```
ChristianBeer: Please don't store HTML in the cache. I know BOINC is not separating Logic and Presentation very good but HTML belongs into the `show_row()` function. The cache should only contain the raw values. This has the advantage that you don't need to invalidate the whole cache just because you tweaked the output a little bit.

In this occasion it seems ok to cache the output from `team_links()` (without the surrounding td) which is HTML just to avoid the Badge DB lookup when displaying a cached page. But for the other values I don't see the need to store anything but the raw values.

html/user/per_app_list.php
```diff
@@ -56,74 +60,319 @@ function show_header($is_team, $apps, $appid, $is_total) {
...
+     global $items_per_page;
+     $fetched_items = $items_per_page;
+     $check = 0;
+     $fetch = 1;
+  
+      //Do the same for the credit_user and user tables
+      //
+     if (!$up) {
+      while ($check != $fetch) {
+             $data = BoincCreditUser::enum("appid=$appid order by $x desc limit $used_offset,$fetched_items");
```
ChristianBeer: This looks like tab indentation. BOINCs Coding style for this part of the source code is to use 4 spaces per indentation level. Please adjust your editor accordingly.

nikolasgianna:
I considered the DB layer to be out of bounds trying to touch as little as possible and aiming at doing everything with the given functions. Since the suggestion is that I implement a new function I have tried to follow the pattern and have overridden a base class function that can be reused.

davidpanderson:
With the addition of the indices on credit_user and credit_team,
performance may be OK without caching or joins.

html/user/per_app_list.php
```diff
@@ -26,104 +26,243 @@
...
+
 // return a column title (Average or Total),
 // hyperlinked if this is not the current sort column
 //
-function col_title($is_team, $app, $appid, $is_total, $i) {
-    $x = $i?"Total":"Average";
-    if ($app->id == $appid && ($is_total?$i:!$i)) {
+function col_title($is_team, $app, $appid, $is_total, $i)
+{
+    $x = $i ? "Total" : "Average";
```
ChristianBeer: If this is shown to the user you need to enclose it with `tra()`

html/user/per_app_list.php
```diff
@@ -26,104 +26,243 @@
...
+    if ($app->id == $appid && ($is_total ? $i : !$i)) {
         return $x;
     } else {
-        return "<a href=per_app_list.php?appid=$app->id&is_team=$is_team&is_total=$i>$x</a>";
+        return "<a href=per_app_list_dev.php?appid=$app->id&is_team=$is_team&is_total=$i>$x</a>";
```
ChristianBeer: leftover from testing?

html/user/per_app_list.php
```diff
@@ -26,104 +26,243 @@
...
+
+function get_top_items($is_team, $appid, $is_total, $offset)
+{
+    global $items_per_page;
+    $x = $is_total ? "total" : "expavg";
+    
     if ($is_team) {
-        $items = BoincCreditTeam::enum("appid=$appid order by $x desc");
+        
+        $data  = BoincCreditTeam::show_list("appid=$appid", $x, $offset . ", " . $items_per_page);
```
ChristianBeer: I personally associate functions that are called `show_` with ones that `echo` stuff and functions that return data are usually called `get_`. So I find this a bit misleading.

html/user/per_app_list.php
```diff
@@ -26,104 +26,243 @@
...
+$items_per_page = 20;
+$offset         = get_int('offset', true);
+
+if (!$offset)
+    $offset = 0;
+if ($offset % $items_per_page)
+    $offset = 0;
+
+$x = $is_team ? "teams" : "participants";
+page_head(tra("Top %1 by application", $x));
```
ChristianBeer: To give the translator a chance to translate this better I guess there should be two complete strings without the `%1`. like this:
```
$title= $is_team ? tra("Top teams by application") : tra("Top participants by application");
page_head($title);
```

html/user/per_app_list.php
```diff
@@ -26,104 +26,243 @@
...
 }
 
-$is_team = get_int('is_team', true);
-$appid = get_int('appid', true);
-$is_total = get_int('is_total', true);
+end_table();
+
+if ($offset > 0) {
+    $new_offset = $offset - $items_per_page;
+    echo "<a href=per_app_list_dev.php?appid=$appid&is_team=$is_team&is_total=$is_total&offset=$new_offset>" . tra("Previous %1", $items_per_page) . "</a> &middot; ";
```
ChristianBeer: another left over from development. Also 5 lines down too.

html/user/per_app_list.php
```diff
@@ -26,104 +26,243 @@
 require_once("../inc/util.inc");
 require_once("../inc/team.inc");
 
+check_get_args(array(
+    "is_team",
+    "appid",
+    "is_total",
+    "added_offset",
+    "up",
```
ChristianBeer: We don't need `added_offset` and `up` anymore.

html/user/per_app_list.php
```diff
@@ -26,104 +26,243 @@
...
+{
     echo "<tr><th colspan=2>&nbsp;</th>";
     foreach ($apps as $app) {
         echo "<th colspan=2>$app->name</th>\n";
     }
     echo "</tr>";
-
+    
     echo "<tr>";
     echo "<th>Rank</th><th>Name</th>\n";
```
ChristianBeer: Some missing `tra()`'s not spotted earlier.

nikolasgianna:
Using microtime() I am getting 650 microsecs on average for building the table from JOIN queries and 750 microsecs on average without the JOIN . Interestingly, using first level caching only, the table takes 80 millisecs on average to build and without any caching at all it takes 100 millisecs. These are all quick measurements just to get a rough idea of the times.

ChristianBeer:
The timings are not that promising but we recognized that it is better to have one slower join query than lots of faster select queries. At one point the amount of fast queries becomes the bigger problem which bogs down the whole database.

nikolasgianna:
Hope I have not missed anything.

ChristianBeer:
In f041c5873aee0c0e59b518962a8db69c1f61b658 you introduced tabs again where there were spaces before.

nikolasgianna:
Sorry for that.

davidpanderson:
alter table credit_user
    add index cu_total(appid, total),
    add index cu_avg(appid, expavg);

alter table credit_team
    add index ct_total(appid, total),
    add index ct_avg(appid, expavg);

Please add these indices and check the performance.

nikolasgianna:
The indices have been added. The performance looks good.

ChristianBeer:
@davidpanderson I would suggest to revert ee3f8f1aa232a615f40c2dbee9047e2fb6551b83 completely and re-implement the other changes (inc/bootstrap.inc) in a separate commit.

@nikolasgianna Can you add the CSS class changes from lines 49 and 58 of ee3f8f1aa232a615f40c2dbee9047e2fb6551b83 to your commit? That would completely substitute the relevant portions of the commit.

ChristianBeer:
Thanks for your patience.

nikolasgianna:
Thank you for the feedback.

PR #1819, opened 2017-03-17
Build: restructure dependency builds for Linux

ChristianBeer:
This reduces the size of files to be cached between builds by the CI system. The scripts are flexible enough so they can also be used by developers locally.


.travis.yml
```diff
@@ -80,5 +80,5 @@ script:
 - if [[ "${BOINC_TYPE}" == "server" ]]; then ( ./configure --disable-client --disable-manager && make ) fi
 - if [[ "${BOINC_TYPE}" == "client" ]]; then ( ./configure --disable-server --disable-manager && make ) fi
 - if [[ "${BOINC_TYPE}" == "apps" ]]; then ( ./configure --enable-apps --disable-server --disable-client --disable-manager && make ) fi
-- if [[ "${BOINC_TYPE}" == "manager" ]]; then ( ./build/getWxWidgets.sh && ./configure --disable-server --disable-client --with-wxdir=./build/wxWidgets-3.0.2/buildgtk && make ) fi
 - if [[ "${BOINC_TYPE}" == "libs-mingw" ]]; then ( cd lib && CC=/usr/bin/x86_64-w64-mingw32-gcc; CXX=/usr/bin/x86_64-w64-mingw32-g++; make -f Makefile.mingw ) fi
+- if [[ "${BOINC_TYPE}" == "manager" ]]; then ( ./3rdParty/buildLinuxDependencies.sh && ./configure --disable-server --disable-client --with-wxdir=./3rdParty/buildCache/linux && make ) fi
```
ChristianBeer: This does not work. configure can't find wxWidgets.
```
checking for ./3rdParty/buildCache/linux/wx-config... no
checking if wxWidgets works... no
```
Find out why. It worked fine locally. Additionally there seems to be a bug in configure since it should fail when building the manager and no wxWidgets was found not silently drop the manager build.

Issue #1555, opened 2016-06-08
"Show graphics" option grayed out when gui_rpc_auth.cfg is set with password


AquaL1te:
When a password is set in /var/lib/boinc/gui_rpc_auth.cfg and I login to the BOINC Manager with host name 127.0.0.1 and the password set in gui_rpc_auth.cfg, the "Show graphics" option is grayed out in the Tasks tab. The "Show graphics option is working when no password is set. Are there extra configurations needed to use a password and have the option to see the graphics?

# User/group associations
```
# id keesj
uid=1000(keesj) gid=1000(keesj) groups=1000(keesj),36(kvm),986(libvirt),978(boinc),135(mock),1001(docker)
```

```
# id boinc
uid=984(boinc) gid=978(boinc) groups=978(boinc)
```

# System info
```
# fpaste --sysinfo --printonly
=== fpaste 0.3.8.1 System Information (fpaste --sysinfo) ===
* OS Release (lsb_release -ds):
     "Fedora release 23 (Twenty Three)"

* Kernel (uname -r ; cat /proc/cmdline):
     4.5.5-201.fc23.x86_64
     BOOT_IMAGE=/vmlinuz-4.5.5-201.fc23.x86_64 root=/dev/mapper/fedora_defiant-root ro rd.lvm.lv=fedora_defiant/root rd.lvm.lv=fedora_defiant/swap rhgb quiet pcie_aspm=force LANG=en_US.UTF-8

* Desktop(s) Running (failed: "ps -eo comm= | grep -E
         '(gnome-session|startkde|startactive|xfce.?-session|fluxbox|blackbox|hackedbox|ratpoison|enlightenment|icewm-session|od-session|wmaker|wmx|openbox-lxde|openbox-gnome-session|openbox-kde-session|mwm|e16|fvwm|xmonad|sugar-session|mate-session|lxqt-session|cinnamon)'
         "):
     N/A

* Desktop(s) Installed (ls -m /usr/share/xsessions/ | sed 's/\.desktop//g' ):
     gnome-classic, gnome

* SELinux Status (sestatus):
     SELinux status:                 enabled
     SELinuxfs mount:                /sys/fs/selinux
     SELinux root directory:         /etc/selinux
     Loaded policy name:             targeted
     Current mode:                   enforcing
     Mode from config file:          enforcing
     Policy MLS status:              enabled
     Policy deny_unknown status:     allowed
     Max kernel policy version:      30

[...]

PR #1820, opened 2017-03-17
Manager/Lib: fix 'New Manager window' function on Linux

### LINKED ISSUE #1177: launch another boinc mgr -> execv: Permission denied
**Reported by toralf on 16 Jul 42651934 12:46 UTC**
Under a stable Gentoo Linux I got that error message in .xsession-errors when I tried to to that action.

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1205

ChristianBeer:
This was not working because the manager was not detecting its own executable name and path so it couldn't start a new instance of itself. Windows and Mac use different codepaths so it worked there.
The new library function can be extended for Windows and Mac to avoid code duplication.

Fixes #1177

### DIFF
--- a/lib/util.cpp
+++ b/lib/util.cpp
@@ -625,3 +625,20 @@ double rand_normal() {
     cached = true;
     return z*cos(PI2*u2);
 }
+
+// determines the real path and filename of the current process
+// not the current working directory
+//
+int get_real_executable_path(char* path, size_t max_len) {
+#ifdef HAVE__PROC_SELF_EXE
+    int ret = readlink("/proc/self/exe", path, max_len);
+    if ( ret >= 0) {
+        path[ret] = '\0'; // readlink does not null terminate
+        return 0;
+    } else {
+        perror("readlink");
+        return ERR_PROC_PARSE;
+    }
+#endif
+    return ERR_NOT_IMPLEMENTED;
+}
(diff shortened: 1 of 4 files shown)


ChristianBeer:
Nice regression. I kind of saw it coming when I wrote the code but didn't test in the end. Good that we have Travis for that. I'm going to fix that later.

ChristianBeer:
Fixed and ready to be merged.

ChristianBeer:
We might want to hold of on merging this. There is a regression on Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=858187 which already has this fix as part of 7.6.33+dfsg-10

ChristianBeer:
Might also fix #1555 (referenced so the status of this PR is visible in the issue)

lib/util.cpp
```diff
@@ -625,3 +625,24 @@ double rand_normal() {
...
 }
+
+// determines the real path and filename of the current process
+// not the current working directory
+//
+int get_real_executable_path(char* path, size_t max_len) {
+#ifdef HAVE__PROC_SELF_EXE
+    int ret = readlink("/proc/self/exe", path, max_len);
+    if ( ret >= 0) {
+        path[ret] = '\0'; // readlink does not null terminate
```
JuhaSointusalo: In the somewhat unlikely event that the link is `max_len` bytes long or longer this write goes one past the end of the buffer. Better have `max_len-1` with `readlink`.

davidpanderson:
Thanks.

PR #1821, opened 2017-03-17
Build: restructure build process on Mac to make it more flexible and allow CI support

ChristianBeer:
This allows to build BOINC on a Mac CI system as well as on a developers computer. See script headers for documentation.
The short version is:
* run ./3rdParty/buildMacDependencies.sh to compile all dependency libraries into 3rdParty/buildCache/mac (can be configured with --cache_dir)
* run ./mac_build/buildMacBOINC-CI.sh to compile all BOINC components using the build cache (--cache_dir can be used here too)

There are options for building debug versions but they are not very well tested. Currently you can't selectively build single targets using this.

We currently build every BOINC component in one go on Mac. This is kind of intended as the build capacity for OSX on Travis CI is capped at 128 builds at the same time right now and extra builds will only add to the already full queue.


ChristianBeer:
The last remaining work (besides merging master back in and fixing the conflict) is to add a `-q|--quiet` or `-v|--verbose` option to the build scripts. This would replace the `make 1>/dev/null` and `make clean 1>/dev/null`redirections in the several build scripts. We don't need the stdout from the makefiles in the automated build since the log size is limited there but a developer may want to see this locally.

CharlieFenton:
@Christian: shouldn't there be a "shift" statement after every case in the "case $key in" blocks of your commit aaec0f9? Without that, I believe that the "while" statement will never exit if there are any arguments other than -prefix or --prefix.

ChristianBeer:
@CharlieFenton No there is a general `shift` after the end of the `case` block that makes sure the `while` loop terminates. The extra `shift` within the `case $key in` blocks is only needed if the argument is followed by a path (like `--prefix` is ) because there are two `shifts` required to get to the next argument.

CharlieFenton:
I have reviewed your changes and approve of them. As far as I am concerned, they are ready to be merged into GIT master.

PR #1825, opened 2017-03-21
Drupal: Fixed bug where font-awesome icons were not appearing correctly on some translated pages.

drshawnkwang:
Changed template for team forum topic to use different variables. Now template uses $results, not topics/rows variable. The former has numerical values, which won't be translated into another language.

https://dev.gridrepublic.org/browse/DBOINCP-360

### DIFF
--- a/drupal/sites/default/boinc/themes/boinc/templates/views-view-table--boinc-team-forum-topics.tpl.php
+++ b/drupal/sites/default/boinc/themes/boinc/templates/views-view-table--boinc-team-forum-topics.tpl.php
@@ -60,7 +60,7 @@
       if ($topic_index == $topic_count) {
         $row_class .= ' last';
       }
-      if ($topic->timestamp OR $topic->new_comments) {
+      if (!empty($topic->timestamp) OR $topic->new_comments) {
         $row_class .= ' updated';
       }
     ?>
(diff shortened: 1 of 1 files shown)


drupal/sites/default/boinc/themes/boinc/templates/views-view-table--boinc-team-forum-topics.tpl.php
```diff
@@ -51,29 +50,29 @@
...
-      elseif (!$first_non_sticky AND !($topic->sticky=="True")) {
+      elseif (!$first_non_sticky AND !($result[$id]->node_boincteam_forum_node_sticky)) {
         $row_class .= ' first-non-sticky';
         $first_non_sticky = TRUE;
       }
       if ($topic_index == $topic_count) {
         $row_class .= ' last';
       }
-      if ($topic->timestamp=="new" OR $topic->new_comments) {
+      if ($topic->timestamp OR $topic->new_comments) {
```
ChristianBeer: Why are you now checking if `topic->timestamp != 0` when before it was a string comparison? I would expect from the name and the previous comparison that this contains an actual timestamp if the topic is not new.

drshawnkwang: It's a hacky solution to the problem, `$topic->timestamp` is a string which is either blank (empty) or ==`new`. The problem is that `new` is translated into a language, example `neu`, by drupal. Therefore a string comparison won't work.

Thus my solution is to simply check the existence of `$topic->timestamp`. If the topic is not new, then the field should be empty.

ChristianBeer: As I expected. The name is misleading. I would suggest to use `empty()` or `isset()` in this case to avoid potential E_NOTICE warnings when the variable is not set. If it is guaranteed to always contain an empty string "" or "new" then use `!empty($topic->timestamp)`

PR #1827, opened 2017-03-24
Remote job submission: first draft of protocol spec file for BOINC-GAHP

ChristianBeer:
Based on http://research.cs.wisc.edu/htcondor/gahp/gce_gahp_protocol.txt version 0.1, http://boinc.berkeley.edu/trac/wiki/CondorBoinc version 27 and https://github.com/BOINC/boinc/blob/f640f89bb266db26c09ee2bf6f677dd0df6610ea/samples/condor/boinc_gahp.cpp

### DIFF
--- a/samples/condor/boinc_gahp_protocol.txt
+++ b/samples/condor/boinc_gahp_protocol.txt
@@ -1,0 +1,916 @@
+
+
+
+                           GRID ASCII HELPER PROTOCOL
+                                   for BOINC
+
+                                  FIRST DRAFT
+
+
+
+1.    INTRODUCTION
+
+    The object of the Grid ASCII Helper Protocol (GAHP) is to allow the
+    use of the client library or package of a grid or cloud service via
+    a simple ASCII-based protocol. A process which implements GAHP is
+    referred to as a GAHP server. GAHP is designed to handle both
+    synchronous (blocking) and asynchronous (non-blocking) calls.
+
+    This GAHP specification focuses on the Berkeley Open Infrastructure
+    for Networked Computing (BOINC) framework. The objective is to
+    provide computing resources from a BOINC-based volunteer computing project
+    to an HTCondor-based submit system.
+
+1.1   WHY GAHP?
+
+    Although most grid and cloud services provide client libraries or
+    modules in various languages that can be incorporated directly into
+    applications that wish to use those services, there are several
+    distinct advantages to using an independent GAHP server process
+    instead. For instance, parts of the native API may provide only
+    synchronous/blocking calls. Users who require a
+    non-blocking/asynchronous interface must make use of multiple
+    threads. Even if the native module is thread-safe, thread safety
+    requires that _all_ modules linked into the process be both
+    re-entrant and in agreement upon a threading model.
+    This may be a significant barrier when trying to integrate the
+    service client module into legacy non-thread-safe code, or when
+    attempting to link with commercial libraries which either have no
+    support for threads or define their own threading model. But because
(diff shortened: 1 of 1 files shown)


samples/condor/boinc_gahp_protocol.txt
```diff
@@ -0,0 +1,916 @@
...
+    3.5.2 GAHP COMMANDS SPECIFIC TO BOINC
+
+    The following commands are specific to the BOINC-GAHP.
+
+    -----------------------------------------------
+
+    BOINC_SELECT_PROJECT
+
+    Select a project to be used for subsequent commands. This is a synchronous
+    command even when in asynchronous mode!
```
ChristianBeer: We should explicitly state that this is not required as the BOINC-GAHP server may have a default.

bema-ligo: The BOINC-GAHP server will only have a default if read for the config file, which I think was implemented for easier testing. Either we should completely specify the config file (name, location, contents) or be completely silent about it in the specification. IMHO the config file is part of the GAHP daemon implementation, not of the protocol. Conformant to the protocol not having specified a BOINC project should result in an error, even if a config file is present.

ChristianBeer: Right now there is no sequence requirement specified in section 3.4 so `BOINC_SELECT_PROJECT` is optional and other commands should fail when no config file is present. The current implementation is using an uninitialized char array if no config file is present and no `BOINC_SELECT_PROJECT` command was received. The values are not further checked in the remote submit API and directly used.

If we make it mandatory then there is no need for a config file. If we allow `NULL` as `project_url` and `authenticator` to signal the server to use a config file then we need to add the `F` return value so the server can signal to the client that there is no config file thus still no project selected.

I don't think we need to specify the config file in the protocol just that the server may implement a way to have a default project already configured that eliminates an initial `BOINC_SELECT_PROJECT` because all users on the condor site should use the same BOINC account but you don't want to give them the authenticator.

bema-ligo: >  The current implementation is using an uninitialized char array if no config file is present and no BOINC_SELECT_PROJECT command was received.
This is a bug (in the implementation, not in the protocol) and should be fixed. However, *how* it should be fixed should be specified in the protocol. If the config file is an officially supported option, it should be fully specified in the protocol.

ChristianBeer: I don't see the config file as part of the protocol it is rather an implementation detail. The protocol should state if the BOINC-GAHP server must have a default project selected or if `BOINC_SELECT_PROJECT` must be called at least once when initiating the connection (making it mandatory). If the former the implementation can choose how to do that not the protocol.

@lfield What are your thoughts on this?

lfield: It depends. If the individual calls don't have the option of specifying the project URL as a parameter then BOINC_SELECT_PROJECT would be required. In any call, if it is a valid call is should be documented.

lfield:
Please can this pull request been merged. As it is only a document it will not break any code and the document does describe the current implementation. Any future changes to the implementation should be accompanied by an update of the specification.

TheAspens:
@ChristianBeer @bema-ligo @lfield - Is everyone ok if I merge this?

ChristianBeer:
This is ready to merge from my perspective as the protocol is fully documented as it is currently implemented on the BOINC side.

lfield:
Yes, please merge.

PR #1830, opened 2017-03-29
[#1702][android]: remove explicit FPU flags

brevilo:
* Those explicit FPU flags prevent NEON builds
* The whetstone benchmarks (VFP and NEON) are correctly built without those

### DIFF
--- a/android/build_boinc_arm.sh
+++ b/android/build_boinc_arm.sh
@@ -22,8 +22,8 @@ export PATH="$PATH:$TCBINARIES:$TCINCLUDES/bin"
 export CC=arm-linux-androideabi-gcc
 export CXX=arm-linux-androideabi-g++
 export LD=arm-linux-androideabi-ld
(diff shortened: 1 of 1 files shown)


brevilo:
> Although this was not enforced consistently in the past it is still the prevailing style

That's what I figured, the current "standard" is neither defined, nor followed by everyone. It also lacks crucial information like this PR/issue references that I added. If it's fine with you I'm not going to change the existing commits/PRs but I'll keep that in mind for future commits, ok?

Having said that, please establish a formal (documented) standard that makes use of tagging (like I did) and adding auto-references to GitHub PRs/issues. This would also facilitate creating automated changelogs.

ChristianBeer:
> If it's fine with you I'm not going to change the existing commits/PRs but I'll keep that in mind for future commits, ok?

Yes of course. They are a nice example how to structure the message so it is also machine readable which might come handy in the future.

davidpanderson:
Thanks!

PR #1834, opened 2017-03-29
Drupal: fix i18n add missing strings

drshawnkwang:
Changed links in menubar to use `url()` function so language code is properly inserted into URL.
Added a number strings to the boinc textgroup, with context details.

### DIFF
--- a/drupal/sites/all/features/boinc_standard/boinc_standard.features.menu_links.inc
+++ b/drupal/sites/all/features/boinc_standard/boinc_standard.features.menu_links.inc
@@ -82,11 +82,10 @@ function boinc_standard_menu_default_menu_links() {
   );
   // Translatables
   // Included for use with string extractors like potx.
-  t('BOINC');
-  t('Credentials');
-  t('Dashboard');
-  t('Home');
-
+  bts('Home', array(), NULL, 'boinc:menu-link');
+  bts('Credentials', array(), NULL, 'boinc:menu-link');
+  bts('Dashboard', array(), NULL, 'boinc:menu-link');
+  bts('BOINC', array(), NULL, 'boinc:footer-link');
 
   return $menu_links;
 }
--- a/drupal/sites/all/features/boinc_standard/boinc_standard.features.menu_links.inc
+++ b/drupal/sites/all/features/boinc_standard/boinc_standard.features.menu_links.inc
@@ -81,7 +81,7 @@ function boinc_standard_menu_default_menu_links() {
     'parent_path' => 'dashboard',
   );
   // Translatables
-  // Included for use with string extractors like potx.
+  // Use bts() function to translate additional strings.
   bts('Home', array(), NULL, 'boinc:menu-link');
   bts('Credentials', array(), NULL, 'boinc:menu-link');
   bts('Dashboard', array(), NULL, 'boinc:menu-link');
--- a/drupal/sites/all/features/stats_charts/stats_charts.panels_default.inc
+++ b/drupal/sites/all/features/stats_charts/stats_charts.panels_default.inc
@@ -38,7 +38,7 @@ function stats_charts_default_panels_mini() {
     $pane->access = array();
     $pane->configuration = array(
       'admin_title' => 'Project stats overview',
-      'title' => 'Einstein@Home Progress',
+      'title' => bts('Einstein@Home Progress', array(), NULL, 'boinc:frontpage'),
       'body' => '<?php echo boincstats_get_project_stats_overview(); ?>',
       'format' => '3',
       'substitute' => TRUE,
--- a/drupal/sites/all/features/stats_charts/stats_charts.panels_default.inc
+++ b/drupal/sites/all/features/stats_charts/stats_charts.panels_default.inc
@@ -38,7 +38,7 @@ function stats_charts_default_panels_mini() {
     $pane->access = array();
     $pane->configuration = array(
       'admin_title' => 'Project stats overview',
-      'title' => variable_get('site_name', '') . " " . bts("Progress", array(), NULL, 'boinc:frontpage'),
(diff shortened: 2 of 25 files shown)


drupal/sites/default/boinc/modules/boinctranslate/includes/other-boinc-translation-strings.txt
```diff
@@ -32,3 +32,11 @@ Report|boinc:forum
 Cancel report|boinc:forum
 Reset flags|boinc:forum
 Ignore user|boinc:forum
+All tasks|boinc:workunit-and-host-lists
+Time reported or deadline|boinc:workunit-and-host-lists
+Run time (sec)|boinc:task-deatils
+CPU time (sec)|boinc:task-deatils
```
ChristianBeer: typo in above 2 lines

drupal/sites/all/features/stats_charts/stats_charts.panels_default.inc
```diff
@@ -38,7 +38,7 @@ function stats_charts_default_panels_mini() {
     $pane->access = array();
     $pane->configuration = array(
       'admin_title' => 'Project stats overview',
-      'title' => 'Einstein@Home Progress',
+      'title' => bts('Einstein@Home Progress', array(), NULL, 'boinc:frontpage'),
```
drshawnkwang: You are correct, this string should be in the 'project' textgroup. However, what do you mean by "this is also in the wrong place it should be in the project specific part not in the generic part"? 

From my reading, you are saying "this": the stats_charts panel Feature that generates our front page chart, is project-specific and should go into a different piece of code, not the generic Boinc-Drupal code. (?)

ChristianBeer: "this" means the string that explicitly mentions the project. When another project is using the Drupal code this is the first thing they need to change. I don't know if it is possible to make the project name configurable in this file so that the string would be "@project-name Progress" and then it is ok to be part of the "boinc" textgroup.

drupal/sites/all/features/stats_charts/stats_charts.panels_default.inc
```diff
@@ -38,7 +38,7 @@ function stats_charts_default_panels_mini() {
     $pane->access = array();
     $pane->configuration = array(
       'admin_title' => 'Project stats overview',
-      'title' => 'Einstein@Home Progress',
+      'title' => variable_get('site_name', '') . " " . bts("Progress", array(), NULL, 'boinc:frontpage'),
```
ChristianBeer: This wouldn't allow for translations where the order of the words need to change because it is translated as "Progress Einstein@Home". Better use a substitution instead of the concatenation so the translator can change the order of the two words.

drupal/sites/all/features/boinc_standard/boinc_standard.features.menu_links.inc
```diff
@@ -82,11 +82,10 @@ function boinc_standard_menu_default_menu_links() {
   );
   // Translatables
   // Included for use with string extractors like potx.
```
ChristianBeer: This comment (also present in lots of other places) is inaccurate. As we found out the `t()` or `bts()` function needs to be called at least once for a non-english language so that the string gets added to the Drupal Translation Interface (DTI). This means that the project that is providing the canonical translation templates for Drupal needs to visit each webpage on its site to ensure that every string is added to the DTI and thus is part of the exported template. The comment suggests that some static string extractor is used which is not the case for the "boinc" and "project" textgroups currently. For projects importing translations this is not a problem as those get added to the DTI automatically.

drshawnkwang: I suspect this comment is automatically generated by the Drupal Features module/feature, which adds text wrapped with `t()` to the bottom of Feature's code & files. Do you recommend removing them as I find them?

drupal/sites/all/features/user_profiles/user_profiles.features.content.inc
```diff
@@ -705,13 +705,13 @@ Zimbabwe|Zimbabwe',
...
-  t('Your opinions about this project');
-  t('Your personal background');
-  t('Zip code');
+  bts('Avatar', array(), NULL, 'boinc:account-profile-edit');
+  bts('Country', array(), NULL, 'boinc:account-profile-edit');
+  bts('Profile image', array(), NULL, 'boinc:account-profile-edit');
+  bts('Website URL', array(), NULL, 'boinc:account-profile-edit');
+  bts('Your opinions about this project', array(), NULL, 'boinc:account-profile-edit');
+  bts('Your personal background', array(), NULL, 'boinc:account-profile-edit');
+  bts('Postal code', array(), NULL, 'boinc:account-profile-edit');
```
ChristianBeer: This looks inconsistent. The labels 'Your personal background' and 'Your opinions about this project' are wrapped with `bts()` here and and in lines 63 and 493 above while 'Postal code' is only wrapped in `bts()` here put not on line 698. I don't know whats the right way but it should be consistent I guess.

drupal/sites/all/features/boinc_standard/boinc_standard.features.menu_links.inc
ChristianBeer: I think it would be best to remove them since they are inaccurate now. The comment suggests that you could remove the following code because it is only needed for potx extraction which is wrong. It is needed for normal operation now that we use `bts()` and context information.

drupal/sites/all/features/user_profiles/user_profiles.features.content.inc
drshawnkwang: Thanks for catching this, I have added bts() to line 698 in the next commit.

PR #1836, opened 2017-03-30
Drupal: Remove project specific name from Feature user_profiles. 

drshawnkwang:
Replaced with site_name variable.

Found as part of https://dev.gridrepublic.org/browse/DBOINCP-317
Similar change requested in PR #1834.

### DIFF
--- a/drupal/sites/all/features/user_profiles/user_profiles.features.content.inc
+++ b/drupal/sites/all/features/user_profiles/user_profiles.features.content.inc
@@ -492,9 +492,9 @@ Zimbabwe|Zimbabwe',
       'default_value_php' => NULL,
       'label' => 'Your opinions about this project',
       'weight' => '4',
-      'description' => 'Tell us your thoughts about Einstein@Home
+      'description' => 'Tell us your thoughts about ' . variable_get('site_name', '') . '
 
-   1. Why do you run Einstein@Home?
+   1. Why do you run '. variable_get('site_name', '') . '?
    2. What are your views about the project?
    3. Any suggestions? ',
       'type' => 'text_textarea',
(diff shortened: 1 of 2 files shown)


drupal/sites/all/features/user_profiles/user_profiles.features.content.inc
```diff
@@ -492,9 +492,9 @@ Zimbabwe|Zimbabwe',
       'default_value_php' => NULL,
       'label' => 'Your opinions about this project',
       'weight' => '4',
-      'description' => 'Tell us your thoughts about Einstein@Home
+      'description' => 'Tell us your thoughts about ' . variable_get('site_name', '') . '
```
ChristianBeer: Same comment as for #1834. In translations the order of words changes based on the language grammar, please use a substitution instead of a concatenation.

Also it seems that these strings do not get translated right now but that might be fixed by #1834. It's shown on account/profile/edit.

drupal/sites/all/features/user_profiles/user_profiles.features.content.inc
```diff
@@ -492,9 +492,9 @@ Zimbabwe|Zimbabwe',
...
 
-   1. Why do you run Einstein@Home?
+   1. Why do you run '. variable_get('site_name', '') . '?
```
ChristianBeer: Same as above.

drupal/sites/all/features/user_profiles/user_profiles.features.content.inc
drshawnkwang: Somehow I missed that this string should be translated. As a result I will close this PR. And then add `bts()` to these strings. Then I will add those changes as part of #1834. Thanks for finding this.

drshawnkwang:
Closing PR, making changes part of #1834, as strings need to be translated in addition to having project-specific text removed.

PR #1839, opened 2017-04-02
Client: Fix missed 'else' keyword.

AenBleidd:
From PVS Studio:
V646
Consider inspecting the application's logic. It's possible that 'else' keyword is missing.
https://www.viva64.com/en/w/V646/print/


davidpanderson:
Thanks!

PR #1842, opened 2017-04-03
Make server report in_progress only when at least one job sent

nikolasgianna:
The BOINC server currently reports in_progress as soon as a job is submitted even if it has not been sent to a volunteer (when queried using the query_batch2 function). Now the function looks in the BoincResult table for the sent_time. If it is zero we have a submitted but unsent job and we can report that.


davidpanderson:
I committed something equivalent to this.
Note: the first 2 items in the PR ($r->batch) are incorrect, I think.

PR #1843, opened 2017-04-03
Fix inconsistency between uploaded files and queried ones in boinc_gahp

nikolasgianna:
When uploaded the files just get the md5 as their name but are then submitted with the "jf_" in front of the name. Thus the batch submission fails due to non existent input file.


lfield:
Is there any chance that this pull request could be merged soon?

PR #1849, opened 2017-04-05
Manager: fix uninitialized class members in constructor

AenBleidd:
From PVS Studio:
V730
Not all members of a class are initialized inside the constructor.
Consider inspecting: m_bCancelInProgress.
Consider inspecting: m_bUseDefaultPassword.
https://www.viva64.com/en/w/V730/print/


davidpanderson:
Thanks!

PR #1851, opened 2017-04-06
Manager: show master_url in Statistics tab when project_name is unknown

### LINKED ISSUE #636: [PATCH] Until <project_name> is known, Manager's Statistics tab displays an empty string
**Reported by Pepo on 1 Aug 38375190 12:26 UTC**
''(This is posibly a duplicate of, an extension of or just another side-effect of events described in ticket:578, #417 and #633 and related changesets.)''

While attaching a new project to host, its scheduler is for any reason down and does not return project's name, e.g.:

``` xml
<scheduler_reply>
    <message priority="low">Server error: can't attach shared memory</message>
    <request_delay>3600</request_delay>
    <project_is_down/>
</scheduler_reply>
```

In Manager's Statistics tab, an empty string is displayed instead of the project's name in the project list at the right side (''"All projects"'' and ''"A. p. (sum)"'' Mode view), a higher minigraph without title is displayed in the graphs area (''"All projects"'').

Project URL should be displayed instead (as it is in the Projects, Messages and Disk tabs).

Migrated-From: http://boinc.berkeley.edu/trac/ticket/659

ChristianBeer:
Original patch from Pepo: http://boinc.berkeley.edu/trac/ticket/659

Fixes #636

### DIFF
--- a/clientgui/ViewStatistics.cpp
+++ b/clientgui/ViewStatistics.cpp
@@ -408,7 +408,11 @@ void CPaintStatistics::DrawProjectHead(wxDC &dc, PROJECT* project1, const wxStri
 	wxCoord y0 = 0;
 
 	if (project1) {
-	    head_name = wxString(_("Project")) + wxT(": ") + wxString(project1->project_name.c_str(), wxConvUTF8);
+		if (project1->project_name.length() > 0) {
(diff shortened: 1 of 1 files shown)


davidpanderson:
This situation, AFAIK, can never occur.

ChristianBeer:
That's right, I couldn't reproduce now that I tested it. It seems that the introduction of the get_project_config.php RPC solved the particular problem originally reported.

PR #1854, opened 2017-04-06
Manager: display zero free space in disk tab

### LINKED ISSUE #714: [PATCH] BOINC Mgr does not display zero free space in disk tab
**Reported by Pepo on 24 Jun 38731388 15:06 UTC**
The Changeset [15194] implemented displaying of the division of free space into available/unavailable to BOINC (because people asking for help were unable to figure out, why does BOINC not get free disk space assigned or how much space it has available).

Unfortunately if BOINC gets no available space, this information is skipped and not displayed at all. In such case BOINC should inform about this fact too.

(The attached patch is untested.)

Migrated-From: http://boinc.berkeley.edu/trac/ticket/737

ChristianBeer:
Also show the slice in the pie diagram if there is no more space available for BOINC.

Original patch by Pepo: http://boinc.berkeley.edu/trac/ticket/737

Fixes #714


AenBleidd:
Does anybody test this?

PR #1855, opened 2017-04-06
Build: install locale files in out-of-tree builds

### LINKED ISSUE #720: Unix "make install" doesn't install gettext files correctly
**Reported by Nicolas on 21 Mar 38747925 14:40 UTC**
The custom install rules in `locale/client/Makefile.am` are incorrect.
1. It's trying to copy the `.po` files. Those are source code; it should install the `.mo` files.
2. It doesn't work with out-of-source builds. The makefile copies the files from the current directory, which isn't necessarily the source directory, where the files actually are. The ''only'' file in `build_dir/locale/client` is the Makefile.

To reproduce 2., do

```
mkdir build;
cd build;
../configure --prefix=~/local/boinc-test; #or wherever...
cd locale/client;
make install; # this looks in build/locale/client/<lang> instead of locale/client/<lang>
```

Migrated-From: http://boinc.berkeley.edu/trac/ticket/743

ChristianBeer:
Fixes #720

### DIFF
--- a/locale/Makefile.am
+++ b/locale/Makefile.am
@@ -2,7 +2,7 @@
 
 include $(top_srcdir)/Makefile.incl
 
-datadir = $(prefix)/share/locale 
+datadir = $(prefix)/share/locale
 
 locale_dirs = \
 	ar \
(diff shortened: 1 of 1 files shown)


AenBleidd:
Does anybody test this?

PR #1860, opened 2017-04-10
Give choice between COM and VboxManage interface

nikolasgianna:
Give the option to Windows users to choose between the COM interface and the VboxManage interface using a preprocessor flag (COM_OFF). Change the way logs are parsed to get the process ID and to get the guest vm log entries. Add some more detailed logging.

### DIFF
--- a/samples/vboxwrapper/vbox_common.cpp
+++ b/samples/vboxwrapper/vbox_common.cpp
@@ -149,6 +149,7 @@ int VBOX_BASE::run(bool do_restore_snapshot) {
 
     retval = is_registered();
     if (ERR_TIMEOUT == retval) {
+		vboxlog_msg("Error: Timeout");
 
         return VBOXWRAPPER_ERR_RECOVERABLE;
 
@@ -156,17 +157,26 @@ int VBOX_BASE::run(bool do_restore_snapshot) {
 
         if (is_vm_machine_configuration_available()) {
             retval = register_vm();
-            if (retval) return retval;
+			if (retval){
+				vboxlog_msg("Could not register");
+				return retval;
+			}
         } else {
             if (is_disk_image_registered()) {
                 // Handle the case where a previous instance of the same projects VM
                 // was already initialized for the current slot directory but aborted
                 // while the task was suspended and unloaded from memory.
                 retval = deregister_stale_vm();
-                if (retval) return retval;
+				if (retval){
+					vboxlog_msg("Could not deregister stale VM");
+					return retval;
+				}
             }
             retval = create_vm();
-            if (retval) return retval;
+			if (retval){
+				vboxlog_msg("Could not create VM");
+				return retval;
+			}
         }
 
     }
@@ -183,15 +193,22 @@ int VBOX_BASE::run(bool do_restore_snapshot) {
     // Check to see if the VM is already in a running state, if so, poweroff.
     poll(false);
     if (online) {
+		vboxlog_msg("VM was running");
         retval = poweroff();
-        if (retval) return ERR_NOT_EXITED;
+		if (retval){
+			vboxlog_msg("Could not stop running VM");
+			return ERR_NOT_EXITED;
+		}
     }
 
     // If our last checkpoint time is greater than 0, restore from the previously
     // saved snapshot
     if (do_restore_snapshot) {
         retval = restore_snapshot();
-        if (retval) return retval;
+		if (retval){
+			vboxlog_msg("Could not restore from snapshot");
+			return retval;
+		}
     }
 
     // Has BOINC signaled that we should quit?
@@ -205,7 +222,10 @@ int VBOX_BASE::run(bool do_restore_snapshot) {
 
     // Start the VM
     retval = start();
-    if (retval) return retval;
+	if (retval){
+		vboxlog_msg("Could not start ");
+		return retval;
+	}
 
     return 0;
 }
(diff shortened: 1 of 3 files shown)


davidpanderson:
Please change code format to match existing code; e.g. 4-space indent, curly brackets on same line

JuhaSointusalo:
In the future please format commit messages like so:

```
component: short one line summary
<blank line>
longer description of what was changed and why
```

Short summary followed by a blank line is the format many Git based tools expect. The summary line lets other people quickly decide if they are interested in your commits. If you are making multiple changes document all of them in the commit message or, better yet, split the changes in multiple commits.

Unless your change is trivial, like fixing a simple typo, the why part is important in commit message. It helps future programmers understand why the changes were made and to decide if they are worth keeping. Right now I can't tell why you would want to have these changes made.

I wasn't paying much attention to the early days of vboxwrapper but I believe at first vboxwrapper used VBoxManage interface even on Windows and later on switched to COM interface. I think the VBoxManage interface wasn't reliable enough or something like that. And now you want to go back to the old thing, why?

As for process ID and guest log entries. As far as I can see these are already working just fine so why change them?

nikolasgianna:
The default interface for Windows users remains the COM one. The difference is that now a user compiling the vboxwrapper is given a choice between the two interfaces. The use of the COM interface presented some problems and since the VboxManage interface is already implemented for Windows it is now merely a matter of choice. 
    The change of the process ID and guest log entries is made because on Windows the parsing was not done correctly and especially for the process ID, BOINC could not acquire control of the VM instance.

marius311:
> The change of the process ID and guest log entries is made because on Windows the parsing was not done correctly and especially for the process ID, BOINC could not acquire control of the VM instance.

@nikolasgianna (I'm the C@H admin where we use vboxwrapper extensively so I'm quite interested in this PR) Can you say more about this piece?  Is this something that has always been a problem? Is there a Github Issue about it yet? Do you have an example what the error logs from this error would look? I'm usually on the lookout for common errors from our jobs but sometimes its quite hard to figure out what is what.

lfield:
In general we have been seeing a high failure rate of tasks for applications that are using VirtualBox. While many of these may be related issues with the setup e.g. hardware acceleration not enabled, the code does not necessarily fail gracefully. We are digging into the errors and either trying to handle them or ensure a clear error message is provided. Due to the use of the API, we are getting different errors for Windows. Our preference is that the behaviour is the same across all platforms and this pull request allows the interface to be chosen. We are currently using this to build the Windows version so that it uses the same interface as Linux and Mac and are digging into the failed tasks.

ChristianBeer:
The COM interface was implemented after using the VboxManage interface for some time on RNA World. We saw a high error rate (especially on Windows) with this constant spawning of new processes to check the status of the current VM. On Linux the VboxManage calls did not produce so much errors that's why Rom saw no reason to also implement the COM API for Linux. I'm looking forward to your tests.

lfield:
This pull request does not necessarily address any issues nor is there any suggestion on which method is better. It is just an option so that both approaches can be used.

marius311:
Yea, I agree being able to optionally turn off COM could be useful, I might well try it for us. My question was just that there seems to be a second partially separate piece to this PR which involves parsing the log, and I was wondering if I should look for this change to solve some errors I may already be seeing?

lfield:
Maybe. It was discovered while testing this option which is why it is part of this pull request.

davidpanderson:
Rom, can you please review this?

Note: please to formatting changes in a separate PR, or at least a separate commit.
Makes it hard to see what's important.

ChristianBeer:
@davidpanderson you can always add "?w=1" to the URL and github will not show whitespace changes for the whole PR. This is the same as doing a local `git diff -w`.

JuhaSointusalo:
> COM_OFF

Fair enough.

> guest log entries

The old code uses `get_vm_log()` to read the last 16 KB of` VBox.log`. `get_vm_log()` uses `read_file_malloc()` to read the file which in turn reads it in binary mode. That means lines end with `"\r\n"`. `get_vm_log()` removes `"\r"` from the log and then tries to fix the log so that it doesn't begin with a partial line. The test on [line 526](https://github.com/BOINC/boinc/blob/master/samples/vboxwrapper/vbox_common.cpp#L526) is broken. If removing `"\r"` from the log shrunk it by more than 115 bytes then the log is left starting with a partial line.

`dump_vmguestlog_entries()` isn't prepared to handle partial lines. The first line may contain `"Guest Log: "` but the timestamp at the beginning of the line may be incomplete. In that case the `sscanf()` on [line 285](https://github.com/BOINC/boinc/blob/master/samples/vboxwrapper/vbox_common.cpp#L285) fails and `current_timestamp` is set to/left as garbage. If that garbage happens to look like a timestamp in future then that garbage is copied to `vm_log_timestamp` and since the timestamp is in future `is_timestamp_newer()` test for the following lines fail and no more guest log entries are copied to vboxwrapper's `stderr`.

Additional problem with `read_file_malloc()` is that it first tries to open the file in exclusive mode. Since client uses `read_file_malloc()` to read `stderr.txt` that ensures that the file is complete, nobody is writing to it any more. With `VBox.log` that is however not desired. VirtualBox has the log file open all the time and trying to open the file in exclusive mode always fails with `ERROR_SHARING_VIOLATION`. That results in five seconds delay every time vboxwrapper tries to read the log file.

The new code avoids those problems by always reading the entire file directly. Although I would add a check that` sscanf()` did succeed just to be safe.

Even though the new code isn't using `get_vm_log()` I think it should be fixed. It's still used elsewhere.

JuhaSointusalo:
> process id

The old code goes though `read_file_malloc()` and has the same five seconds delay as reading guest log entries has. As far as I can tell that is the only problem it has.

The new code avoids the delay by reading the file directly but has a change I would call a bug. Whereas the old code returned with `ERR_NOT_FOUND` if it couldn't find the process ID, the new code returns with `ERR_NOT_FOUND` only if the log file format doesn't match what the code expects. If the new code can't find process ID or the log file doesn't exist at all it returns with `BOINC_SUCCESS`.

There are some cosmetic issues as well:
The code first tries to find a line with `"Process ID"` and then within that line `"Process ID: "`. IMHO that kind of two stage search doesn't add any value.
The code opens the log in binary mode. `read_file_malloc()` needs to do that because it `fseek()`s the file but it shouldn't be necessary here.
The code searches for the end of line even though it's already processing the log file line by line. Just use `line.substr(start, string::npos)`.

----

As for the real bug here that made you do all this work, it's not in vboxwrapper but in BOINC lib in `file_size()`. The `CreateFile()` call on [line 354](https://github.com/BOINC/boinc/blob/master/lib/filesys.cpp#L354) uses wrong sharing mode. VirtualBox has the log file open for writing but `file_size()` allows only reading and the call fails with `ERROR_SHARING_VIOLATION`. The correct sharing mode is `dwShareMode = FILE_SHARE_READ | FILE_SHARE_WRITE`. Also, [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363858(v=vs.85).aspx) says that one can use `dwDesiredAccess = 0` if the intention is to only query file metadata like its size.

This bug is present in vboxwrapper version 26198 but not the older ones. In 26198 it breaks only getting guest log entries but not getting process ID because the COM interface uses different method to get the process ID.

This bug affects client as well. The client checks slot directory sizes with different function but for output files in project directory it uses `file_size()`. Science app may keep output files open for writing and with this bug the client can't tell how large they are thus making it possible for tasks to exceed their allowed disk usage without client noticing.

This is why I asked for more information. So that whoever reviews your PR can check that you didn't just work around a bug but really fixed it.

romw:
I'll be able to look at this sometime this week.  Hopefully sooner rather than later.

Sorry about the delay.

PR #1861, opened 2017-04-11
Fix android api

davidekim:
I removed the signal handlers to help with debugging and added David Anderson's file based messaging as the simple/better alternative to help with debugging. @davidpanderson, I didn't add the other changes in https://github.com/BOINC/boinc/tree/android_api_dpa since I didn't have time to thoroughly test them yet.  Is there anyone who can test these changes in parallel with their application? Thanks!

### DIFF
--- a/api/boinc_api.cpp
+++ b/api/boinc_api.cpp
@@ -108,14 +108,11 @@
 
 using std::vector;
 
-//#define DEBUG_BOINC_API
+#define VERBOSE
+    // enable a bunch of fprintfs to stderr
 
-#ifdef DEBUG_BOINC_API
-static volatile bool abort_called = false;
-static volatile bool quit_called = false;
-static volatile bool resume_called = false;
-static volatile bool suspend_called = false;
-#endif
+#define MSGS_FROM_FILE
+    // get messages from a file "msgs.txt" instead of shared mem
 
 #ifdef __APPLE__
 #include "mac_backtrace.h"
@@ -328,7 +325,7 @@ static inline void release_mutex() {
 pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
 static void init_mutex() {}
 static inline void acquire_mutex() {
-#ifdef DEBUG_BOINC_API
+#ifdef VERBOSE
     char buf[256];
     fprintf(stderr, "%s acquiring mutex\n",
         boinc_msg_prefix(buf, sizeof(buf))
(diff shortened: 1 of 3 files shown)


davidekim:
@davidpanderson, I just noticed you committed your changes but when I tested them, I was still getting seg faults.  And the file based message needs to copy msg_buf to buf for it to work. I've added these changes to this pull request and I've tested this with our app.  Let me know if you would like me to test additional updates you might make related to this in the future.  Thanks!

brevilo:
FYI, the current merge conflict flagged by GitHub is due to the fact that David A.'s own [reimplementation](https://github.com/BOINC/boinc/commits/android_api_dpa) of David K.'s changes made it into master prematurely, via a (fast-forward) merge on 9. April 2017 between 6:40:43 and 7:37:27 UTC. The conflicts should be resolved by reverting the commits concerned on master. David A.'s already aware of this.

tl;dr: the conflict isn't an issue of the PR itself, it doesn't need to be changed for the time being.

davidekim:
@davidpanderson  I'm not sure if setitimer or calling SIGALRM so frequently is causing the segfaults. Is it necessary to send SIGALRM to the worker thread so often (every .1 sec)?  My original updates called SIGALRM when I thought was only necessary:

1. in exit_from_timer_thread to tell the worker thread to exit.
2. in handle_process_control_msg to tell the worker thread to suspend/resume/exit
3. in timer_handler in the options.check_heartbeat section but I'm not sure if this is necessary.

Also, making sure in_critical_section is never less than 0 prevented incorrect !in_critical_section logic.

davidpanderson:
I changed boinc_api.cpp (in master) to
- send signals to the worker thread only when necessary, as David Kim describes above
- in the signal handler, make sure we're the worker thread.
  Maybe this was necessary after all.
- Added an optional #define ANDROID so that you can test the Android signal/thread logic on Linux;
  it should work on Linux too.

David K., can you please test this new version?

davidekim:
Will do, thanks!

DK

brevilo:
> I changed boinc_api.cpp (in master) to

Guys, please don't develop in (and break) master. This affects numbers of other devs. Also, as long the old broken **and** premature commits stay in master this PR can never be merged easily - they'll conflict.

Mistakes happen (premature merge) but please don't manifest them, fix them. I simple `git revert` would do that.

brevilo:
Unfortunately this PR was mostly discussed via email and developed in `master`. Good luck stringing together the various bits and pieces should it ever become necessary. Sigh.

PR #1863, opened 2017-04-11
Build: use default sysconfdir instead of /etc

### LINKED ISSUE #1227: Installation makefile loftily wants to install stuff to /etc ... Nope!
**Reported by Bluefin Tuna on 5 Nov 43381410 16:45 UTC**
Don't know how to solve this, but the installation makefile from the source distro thinks it has license to install to /etc.

DESTDIR in this case is unset (would anything else make sense? Looking at the code below, where DESTDIR is partially missing, no) and sysconfdir is "/etc".

```
install-exec-hook:
        chmod +x boinc-client
        $(INSTALL) -d $(DESTDIR)$(sysconfdir)/init.d
        $(INSTALL) -b $(srcdir)/boinc-client $(DESTDIR)$(sysconfdir)/init.d/boinc-client
        if [ -d /etc/sysconfig ] ; then \
          $(INSTALL) -d $(DESTDIR)$(sysconfdir)/sysconfig ; \
          $(INSTALL) $(srcdir)/boinc-client.conf $(DESTDIR)$(sysconfdir)/sysconfig/boinc-client ; \
        elif [ -d /etc/default ] ; then \
          $(INSTALL) -d $(DESTDIR)$(sysconfdir)/default ; \
          $(INSTALL) $(srcdir)/boinc-client.conf $(DESTDIR)$(sysconfdir)/default/boinc-client ; \
        else \
          $(INSTALL) -d $(DESTDIR)$(sysconfdir) ; \
          $(INSTALL) $(srcdir)/boinc-client.conf $(DESTDIR)$(sysconfdir)/boinc-client.conf ; \
        fi
```

The above is not even being run with root permission in my case (instead it is run by user 'boinc'), so simply fails and must be commented out.

Anyway, things are no longer handled that way in Fedora 17. Startup/shutdown operations are now done by via systemd, not init.

For example, httpd is started via a symlink 

/etc/systemd/system/multi-user.target.wants/httpd.service

pointing to

/usr/lib/systemd/system/httpd.service

Does the boinc client drop its privileges from root to some specified user? Not sure about that.

Anyway, this needs an overhaul.

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1255

ChristianBeer:
When using --prefix this override tries to install files from client/scripts to /etc instead of ${prefix}/etc. This is against what the user wants and leads to an error message as normally only root can install files to /etc. There is no need to be backwards compatible in this place as installs are usually done via repository packages.
This still uses /etc to determine where the config file should be copied to which is ok because the user most likely wants to use the Client on the machine it was built on.

Note: The startup scripts and config files in client/scripts are outdated and might not work on recent Linux distributions. Newer ones can be found in the Debian/Fedora packages.

fixes #1227

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -538,16 +538,6 @@ if test "x${libdir}" = "xNONE" -o -z "${libdir}" ; then
   bindir="${exec_prefix}/lib"
 fi
 
-dnl if --sysconfdir was not specified, default ${sysconfdir} to '/etc/'.
-dnl this is done to preserve the behaviour of previous BOINC versions
-dnl which always installed to /etc/ in client/scripts/. Now we install
-dnl to ${sysconfdir}/.
-dnl the default value for ${sysconfdir} set by automake is '${prefix}/etc',
-dnl but we also test for an empty string
-if test "${sysconfdir}" = '${prefix}/etc' -o "x${sysconfdir}" = "x"; then
-  sysconfdir=/etc
-fi
-
 AC_DEFINE_UNQUOTED(BOINC_PREFIX,"`eval echo ${prefix}`",[Directory tree where BOINC will be installed])
 AC_DEFINE_UNQUOTED(BOINC_EXECPREFIX,"`eval echo ${bindir}`",[Directory where BOINC executables will be installed])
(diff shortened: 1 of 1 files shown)


AenBleidd:
Does anybody test this?

Issue #1864, opened 2017-04-11
The BOINC manager should display the build SHA1 in its About dialog


brevilo:
Showing the arbitrary build version isn't enough to know exactly the source code version used for a build. Show the abbreviated SHA1 of the HEAD used for building the app in the dialog as well.

ChristianBeer:
This is needed to better distinguish development versions without creating a new minor release just for a limited test.

brevilo:
The Android app can be changed in a pretty easy way. I'm going to open a PR for it soon.

Alternatively the client itself could have the SHA1 compiled in and a new RPC (?) added such that the manager can query it.  This assumes that manager and client are compiled from the same source though.

Vulpine05:
I'm interested in working on this feature.  I have a few of questions:

1. In theory, the Client and Manager could have different builds.  Should both be displayed in the About dialogue?
2. Should the SHA1 be displayed for a pre-release version only, or for a release version as well?
3. How many characters for the SHA1 should be displayed?  I would assume 7, similar to what Github displays for commits.

AenBleidd:
@Vulpine05,
1. Only the Manager. The client reports its version to the console, so technically you can add it to that console message.
2. For every build and every version
3. For Android we do 10 characters, so I'd follow the same approach here as well: https://github.com/BOINC/boinc/actions/runs/17310796416/job/49144675017#step:10:102

Vulpine05:
Thank you, @AenBleidd.  I've been able to get something working for Windows.  I have a few more follow-up questions.

First, I think it makes sense to have a dirty tag.  It can look like this:

<img width="577" height="265" alt="Image" src="https://github.com/user-attachments/assets/44206f72-d55d-4092-80ac-91001016f0e4" />

Let me know if you have any comments.

Second, I currently am having Visual Studio run pre-compiler commands to obtain and store the hash, and it is writing to a new file.  However, I am thinking it would be cleaner to have hashes stored in version.h in the root directory.  That way there aren't separate files for one line (one file for the manager, another for the client when I get to that).  However, I want to see what your thoughts are on sharing that same file.  I don't want to break anything with the versioning script.  I would think I could search and overwrite the definition for the hash.

AenBleidd:
@Vulpine05.
1. Why did you put 'dirty' there?
[...]

PR #1866, opened 2017-04-13
Android: use dynamic versioning instead of hard-coded values

brevilo:
* Use commit date as version code
* Derive version name from release tag and add commit SHA1
* No more manual updates required, just proper release tagging!
* Resolves https://github.com/BOINC/boinc/issues/1864 (for Android)

FYI, the required tag format is as follows:
`client_release/x.y/x.y.z[suffix]` (suffix is optional but must be separated with `-` or `_`)

### DIFF
--- a/android/BOINC/app/build.gradle
+++ b/android/BOINC/app/build.gradle
@@ -1,4 +1,44 @@
+plugins {
+    id 'org.ajoberstar.grgit' version '1.7.1'
+    id 'com.palantir.git-version' version '0.7.1'
+}
+
 apply plugin: 'com.android.application'
+
+import org.ajoberstar.grgit.Grgit
+
+// Use commit date as version code
+def buildVersionCode() {
+    def repo = Grgit.open()
+    def head = repo.head()
+
+    // Sanity check across git plugins
+    assert head.getAbbreviatedId(10) == versionDetails().gitHash : "Internal error: SHA1 mismatch!"
+
+    return head.time
+}
+
+// Derive version name from release tag and add commit SHA1
+def buildVersionName() {
+    def pattern = /client_release\/\d+\.\d+\/(?<major>\d+)\.(?<minor>\d+)\.(?<revision>\d+)(?<suffix>[-_].*)/
+    def message = 'Invalid release tag encountered:'
+
+    def tag = versionDetails().lastTag
+    def match = (tag =~ pattern)
+
+    // Sanity checks for tag format
+    assert match.hasGroup() : "$message $tag"
+    assert 1 == match.size() : "$message $tag"
+    assert 5 == match[0].size() : "$message $tag"
+
+    def major = match.group('major')
+    def minor = match.group('minor')
+    def revision = match.group('revision')
+    def suffix = match.group('suffix')
+    def commit = versionDetails().gitHash
+
+    return "${major}.${minor}.${revision}${suffix} (${commit})"
+}
 
 android {
     compileSdkVersion 23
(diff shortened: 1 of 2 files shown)


android/BOINC/app/build.gradle
```diff
@@ -1,5 +1,45 @@
...
+
+    return head.time
+}
+
+// Derive version name from release tag and add commit SHA1
+def buildVersionName() {
+    def pattern = /client_release\/\d+\.\d+\/(?<major>\d+)\.(?<minor>\d+)\.(?<revision>\d+)(?<suffix>[-_].*)/
+    def message = 'Invalid release tag encountered:'
+
+    def tag = versionDetails().lastTag
```
ChristianBeer: In current master this produces: `vboxwrapper/26198-132-g247e6b096b` for a development build which does not match the pattern and thus leaves the variables undefined. There should be a sensible default for `${major}`, `${minor}`, ... in the returned string below that indicates that this is a development version.

android/BOINC/app/build.gradle
```diff
@@ -1,5 +1,45 @@
...
+
+// Use commit date as version code
+def buildVersionCode() {
+    def repo = Grgit.open()
+    def head = repo.head()
+
+    // Sanity check across git plugins
+    assert head.getAbbreviatedId(10) == versionDetails().gitHash : "Internal error: SHA1 mismatch!"
+
+    return head.time
```
ChristianBeer: This is fine until 07/18/2036 where the maximum allowed value for `versionCode` (2,100,000,000) is reached. https://developer.android.com/studio/publish/versioning.html#appversioning

brevilo:
@davidpanderson please don't merge PRs that are still in discussion. Christian's comments are valid points and my PR should have been improved before merging. That's the whole point of proposing a PR and getting it reviewed. Otherwise I could just push the stuff to `master` myself.

android/BOINC/app/build.gradle
brevilo: Right. How about subtracting the past by, say,  the "Columbus factor" `1492000000` (2017-04-12 14:26:40)? This gives plenty of headroom. Note that the next release would have to be based on a commit not older than 2017-04-12 14:27:53 in this case.

android/BOINC/app/build.gradle
brevilo: This isn't really possible unfortunately since there's no reliable tagging in that case. How about displaying `DEVELOPENT` as version, followed by `(SHA1)` (as per this PR)?

brevilo: Also, if there's a commit offset relative to the latest release tag found, I also would like to flag it as a development version. Do you agree? That would mean **any** official release build **must** be **directly** based on a release tag.

ChristianBeer: Yes, if there is a proper tag and no offset display the version number like `7.6.33 (6f6c8b7eb5)`. If there is an offset or the tag pattern does not match, display `DEVELOPMENT (981b0409d3)`

android/BOINC/app/build.gradle
ChristianBeer: My statement was just meant as information for someone else stumbling about the `versionCode` limitation. I think it is fine and we don't need a factor. If this is still a problem 19 years from now let the future BOINC developers worry about that. :smiley:

brevilo: Ok.

android/BOINC/app/build.gradle
brevilo: ACK

PR #1867, opened 2017-04-13
Android: update copyright details

brevilo:
It's 2017 by now 😄


ChristianBeer:
This really begs the question why is the year part of the translatable string? This means that every translator needs to change the string after this gets updated on transifex. Wouldn't it be better to use a substitution here and keep the years in code?

brevilo:
I agree with you Christian but the PR got merged in the meantime so I can't improve on it anyway. A new PR is needed...

PR #1873, opened 2017-04-19
Android: improved auto-versioning

brevilo:
* Current version doesn't support development well enough
* Gradle seemingly doesn't support distinguishing build variants (debug/release) so I can't enforce proper release tagging
* Thus any non-matching tag will flag a build as DEVELOPMENT and a Gradle warning is displayed
* Also: release builds must be tagged directly (no commit offset)
* Also: release builds are not allowed to use a dirty working tree

This is a follow-up to PR #1866 that got merged prematurely.


davidpanderson:
Thanks!

PR #1878, opened 2017-04-22
SCHED_BATCH -> SCHED_IDLE

smoe:
It was a BOINC RasPi forum entry that made me revisit chrt and relearn about scheduler policies+priorities. BOINC executes the tasks (if !highpriority) under the SCHED_BATCH policy. This also let me unearth again that SCHED_IDLE is what in the very early UNIX days was reached by renicing to the highest values but today SCHED_IDLE is less of a priority than SCHED_BATCH with nicelevel 19.  I just executed the "chrt -i ..." manually and it seems just fine. Graphics still work. Communication with the BOINC servers is performed with a higher priority outside the task, which lets me think that we should expect any problem. So I suggest to have SCHED_IDLE the default for compute jobs.

Best,
Steffen

### DIFF
--- a/client/app_start.cpp
+++ b/client/app_start.cpp
@@ -1090,11 +1090,11 @@ int ACTIVE_TASK::start(bool test) {
                 }
             }
 #endif
-#if HAVE_SCHED_SETSCHEDULER && defined(SCHED_BATCH) && defined (__linux__)
+#if HAVE_SCHED_SETSCHEDULER && defined(SCHED_IDLE) && defined (__linux__)
             if (!high_priority) {
                 struct sched_param sp;
                 sp.sched_priority = 0;
-                if (sched_setscheduler(0, SCHED_BATCH, &sp)) {
+                if (sched_setscheduler(0, SCHED_IDLE, &sp)) {
                     perror("sched_setscheduler");
                 }
             }


Ferroin:
Just a bit of background here.  SCHED_BATCH on Linux essentially equates to adding 0.5 to the nicelevel.  A SCHED_BATCH task will be lower priority than stuff of the same nice level, but higher than stuff of a lower nicelevel.  On the other hand, SCHED_IDLE actually changes how the scheduler looks at the task, not only reducing its runtime impact on the system, but actually reducing the scheduling overhead  regarding it as well.  It is not, however, a true idle scheduler, as SCHED_IDLE tasks will still run on occasion when there are other tasks to schedule.

smoe:
Thank you for the confimatory extra info. I presume an occasional wake-up helps to avod dead locks when the idle'd task is not granted the opportunity to free a resource.

lfield:
From the code perspective this seems fine. The question is does anyone disagree about changing the priority. From the man page (`man sched `).

>SCHED_OTHER: Default Linux time-sharing scheduling
>SCHED_OTHER is the standard Linux time-sharing scheduler that is intended for all threads that do not require the special real-time mechanisms.

>SCHED_BATCH: Scheduling batch processes
>SCHED_BATCH is similar to SCHED_OTHER in that it schedules the thread according to its dynamic priority (based on the nice value).  The difference is that  this  policy  will  cause  the  scheduler to always assume that the thread is CPU-intensive. Consequently, the scheduler will apply a small scheduling penalty with respect to wakeup behavior, so that this thread is mildly disfavored in scheduling decisions.

>SCHED_IDLE: Scheduling very low priority jobs
> The process nice value has no influence for this policy. This policy is intended for running jobs at extremely low priority (lower even than a +19 nice value with the SCHED_OTHER or SCHED_BATCH policies).

Issue #1882, opened 2017-04-24
Replace autotools build system


ChristianBeer:
The current autotools centered build system is very fragile and hard to maintain (if you are not a autotools expert). It is also Unix/Linux centered and only half heartedly supports cross compilation (e.g. mingw).

I propose to replace it gradually with a [cmake](https://cmake.org/) based build system which offers multi-platform and multi-compiler support and an easier language to write build configuration files. It can for example build Visual Studio project files as well as regular Makefiles as well as control files for other build tools.

I already did some initial testing and can create config.h to replace autoheader and have some simple rules to build basic libraries. The challenging part is backwards compatibility (what special checks are still relevant and what can be omitted) and creating/using static libraries for release builds which I have not yet explored.

AenBleidd:
@ChristianBeer Agreed. I have some experience creating cmake scripts so I can try to help you with this.

JamesMolson:
I am working on this right now and should have an early (i.e. there might be some missing features) pull request ready for people to view within the next few days.

talregev:
I want to start a process to convert boinc compilation to CMake.
@ChristianBeer @JamesMolson Do you want to join the efforts?

BryanQuigley:
I've been wrestling with autotools repeatedly.  Any chance to help with this?

talregev:
@BryanQuigley What help do you need? 
Also we compile boinc with autotools inside the ci. You can see an example how to compile in your system.

AenBleidd:
I do agree that autotools centered build system is very complex, however it's proved to work for two decades already.
Yes. it's used at the moment on *nix systems only (including Android) but it works fine and stable.
Moving it to cmake will require a huge effort and a log of resources (and I even not sure how to test it afterwards on all the systems and configuration we currently support running/building BOINC on).
Whine in general this sounds interesting, this is not something we are gonna work.
Maybe one day we will reconsider this, but until then I don't see any reason keeping this ticket open.

PR #1886, opened 2017-04-25
Build Windows: Add AppVeyor support for Windows CI builds.

AenBleidd:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,8 +4,8 @@ matrix:
   fast_finish: true
 
 platform:
-  - Win32
-#  - x64
+#  - Win32
+  - x64
 
 configuration:
   - Debug
@@ -17,17 +17,16 @@ environment:
       solution_name: win_build\boinc_vs2013.sln
       depends_git_path: http://boinc.berkeley.edu/git/boinc_depends_win_vs2013.git
       depends_path: C:\projects\boinc_depends_win_vs2013
-    - TOOLCHAIN_VERSION: 12.0
-      solution_name: win_build\boinc_vs2013_ibm.sln
-      depends_git_path: http://boinc.berkeley.edu/git/boinc_depends_win_vs2013.git
-      depends_path: C:\projects\boinc_depends_win_vs2013
+
+cache:
+  - C:\projects\boinc_depends_win_vs2013
 
 before_build:
-  - git clone -q --branch=master %depends_git_path% %depends_path%
+  - call win_build\load_dependencies.bat %depends_git_path% %depends_path% %platform% %configuration%
   - if %TOOLCHAIN_VERSION%==12.0 call "%VS120COMNTOOLS%\vsvars32.bat"
 
 build_script:
-  - devenv %solution_name% /Build
+  - devenv %solution_name% /Build "%configuration%|%platform%"
 
 build:
   parallel: true
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,35 @@
+image: Visual Studio 2013
+
+matrix:
+  fast_finish: true
+
+platform:
+  - Win32
+#  - x64
+
+configuration:
+  - Debug
+#  - Release
+
+environment:
+  matrix:
+    - TOOLCHAIN_VERSION: 12.0
+      solution_name: win_build\boinc_vs2013.sln
+      depends_git_path: http://boinc.berkeley.edu/git/boinc_depends_win_vs2013.git
+      depends_path: C:\projects\boinc_depends_win_vs2013
+    - TOOLCHAIN_VERSION: 12.0
+      solution_name: win_build\boinc_vs2013_ibm.sln
+      depends_git_path: http://boinc.berkeley.edu/git/boinc_depends_win_vs2013.git
+      depends_path: C:\projects\boinc_depends_win_vs2013
+
+before_build:
+  - git clone -q --branch=master %depends_git_path% %depends_path%
+  - if %TOOLCHAIN_VERSION%==12.0 call "%VS120COMNTOOLS%\vsvars32.bat"
+
+build_script:
+  - devenv %solution_name% /Build
+
+build:
+  parallel: true
+
+test: off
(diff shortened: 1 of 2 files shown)


AenBleidd:
@davidpanderson, @ChristianBeer 
Please review this and merge when it will be decided to add this service support only.
Also @ChristianBeer I'd want to ask you to add a badge to README.md when this PR will be merged because I can't do this for this repository.
Thanks

ChristianBeer:
Thanks for looking into this. I'll look into ways to integrate AppVeyor into github once the PMC made a decision.

Things that already caught my eye:
- The difference between building "Vanilla" and "IBM" should be the branding only. There should be no functional differences just different filenames of the executables and changed wording in dialogs. I don't think we need to build both, Vanilla is enough. I don't know if it is possible to get rid of the separate IBM solution file and create that on the fly from the Vanilla solution so we don't need to maintain two solution files side by side. That would be nice to have in the future.
- You chose `Win32` as `platform` but does that really build the 32bit version of Client and Manager? On Linux we are only building the 64bit version as this is the one used the most. Can you clarify this part? I would want the Windows CI to also build the 64bit executables only.
- The way you pull in the dependencies will cause an issue with the berkeley server. Right now you clone the complete dependency repo which is >1GB in total but you only need the latest version of the dependencies. This causes a traffic spike every time an AppVeyor build is happening. We should find a way to cache the dependencies that are really needed in some other way that at least limits the traffic to the berkeley server. AppVeyor offers a similar caching feature as Travis CI: https://www.appveyor.com/docs/build-cache/ and is limited to 1GB for a free account. Can you try to identify the files from the dependency repo that are really needed to build and check the size of those? If this is within the 1GB limit we just need to find a way to get that into the cache which could be as simple as zipping it up and downloading it inside the AppVeyor build.

AenBleidd:
1 - I'll check the differences between these solutions. If there is no functional changes then we can switch off CI build for it. If we want to get rid of separate solution then this is a nice task in scope of #1882 where we can add similar projects inside one solution automatically.
2 - Yes, Win32 configuration really builds 32-bit binaries. No problem to switch it to x64
3 - The biggest part is wxWidgets binaries (~650 MB for x64). Other are quite small so if we will chose x64 build it will be less than 1GB.

I'll take a look at this today and will make corresponding changes.
Thanks

AenBleidd:
I made changes. Now only one solution is in build queue. Windows pre-built binaries are too big so I had to use dirty hack with removing the unnecessary files after the first git clone. But this leads to an impossibility to include other configurations (e,g. Release) and platforms (e.g. , Win32) to build queue. My script currently supports only 1 configuration and 1 platform at the same time. The main problem is that we cannot include all pre-build binaries because they are too big (~1.4 GB).
@ChristianBeer Could you please review this? Thanks

ChristianBeer:
The wxWidgets library for release should be smaller than for Debug. Since we don't use the build from CI systems right now you can change to a release configuration if that makes the dependencies even smaller. I prefer to have at least one configuration and one platform that can be used to detect build breaks on CI systems. If that is working we can figure out better ways to handle the dependencies. Priority should be to get a simple system working and then improve on that.

Btw: Is the dependency path `C:\projects\boinc_depends_win_vs2013` hardcoded in the solution file? I think that should be changed sooner than later in a separate PR.

AenBleidd:
@ChristianBeer Yes, this solution is works now. I made some calculations and I can say that we can have even two combinations of dependencies from next list:
1 - Win32 Debug
2 - Win32 Release
3 - x64 Debug
4 - x64 Release.
But this will require additional changes to load_dependencies.bat file. These changes will be small and can be easily made.
Yes, dependency path is hardcoded  now as (..\boinc_depends_win_vs2013) for VS 2013 solution and as (..\boinc_depends_win_vs2010) for VS 2010 solution.
This can be changed also but not in the scope of this task I guess.
As I mentioned before such changes could be done in better way in scope with #1882.

ChristianBeer:
Closing and reopening to trigger a new AppVeyor build.

ChristianBeer:
Second test with proper organization set up

ChristianBeer:
The build succeeded but on a quick glance I saw some warnings about a potential flawed build because of linking issues. @AenBleidd you might want to check if the output is as expected. I'm not sure how the caching works as I don't see an option in the appveyor admin interface but the build by itself was rather quick.

AenBleidd:
@ChristianBeer 
1 - Output is correct, These warning during linking are because of renaming of output files. I'll take a look at it.
2 - Cache in AppVeyor works next: the cache some folder and then before next build this folder will be filled with the content from previous build. The problem is that they doesn't do any check. That is why I've added a temporary file with revision name (SHA1 actually) and I make a check whether there was some changes with third-party dependencies. Also as I see from the logs, first build was canceled that is why the second build didn't use cache. As I mentioned previously I want to put third-party pre-build dependencies somewhere else but not in git repo on Berkeley servers. Some cloud (or ftp) is better I think.

JuhaSointusalo:
FWIW, I'm currently making sure that BOINC builds on MSYS2+MinGW. After I'm done with that I intend to add VS2017 solution using vcpkg for dependencies (unless Rom suddenly emerges from somewhere).

AenBleidd:
@Juha, VS2017 will be great but currently there was some information that
there will be some incompatibility win WinXP if BOINC will be built using
VS greater than 2010.
In any case new solution will be easier to add to existing AppVeyor
workflow.

Thanks

Best regards,
Vitalii Koshura

2017-08-09 18:02 GMT+03:00 Juha Sointusalo <notifications@github.com>:

JuhaSointusalo:
Yep. I will test XP compatibility. In any case we could use VS2017
solution. Older versions can be had only by signing up to Dev Essentials
(which is free but still) and using vcpkg means most people won't be
needing dependencies from Berkeley's server.

AenBleidd:
Yes, this will be great.

Thanks

Best regards,
Vitalii Koshura

2017-08-09 19:01 GMT+03:00 Juha Sointusalo <notifications@github.com>:

PR #1892, opened 2017-04-30
vboxwrapper: Remove unused variables

AenBleidd:
From PVS Studio:
V808
'vm_machine_uuid' object of 'CComBSTR' type was created but was not utilized.
'clean_mediums' object of 'vector' type was created but was not utilized.
https://www.viva64.com/en/w/V808/print


davidpanderson:
Thanks!

PR #1893, opened 2017-04-30
Move 'strlen' function outside of the loop

AenBleidd:
From PVS Studio:
V814
Decreased performance. The 'strlen' function was called multiple times inside the body of a loop.
https://www.viva64.com/en/w/V814/print

Affected components:
api
client
lib
samples

### DIFF
--- a/api/boinc_api.cpp
+++ b/api/boinc_api.cpp
@@ -1089,13 +1089,14 @@ static void handle_upload_file_status() {
     char path[MAXPATHLEN], buf[256], log_name[256], *p, log_buf[256];
     std::string filename;
     int status;
+    const size_t prefixLen = strlen(UPLOAD_FILE_STATUS_PREFIX);
 
     relative_to_absolute("", path);
     DirScanner dirscan(path);
     while (dirscan.scan(filename)) {
         strlcpy(buf, filename.c_str(), sizeof(buf));
         if (strstr(buf, UPLOAD_FILE_STATUS_PREFIX) != buf) continue;
-        strlcpy(log_name, buf+strlen(UPLOAD_FILE_STATUS_PREFIX), sizeof(log_name));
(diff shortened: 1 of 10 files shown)


davidpanderson:
I'm inclined to not do this.  It makes the code slightly harder to read, and it's the sort of thing that an optimizing compiler takes care of.
Also, unrelated, let's stick to lower_case type var names.

AenBleidd:
Ok. As you wish.

adamradocz:
Putting the length inquiry outside the loop is a common best practice, I don't think it makes the code harder to read. Are you sure it's handled by the compiler?

davidpanderson:
Of course it makes it harder to read.  There's a new variable, you have to go find where it is, then you have to go back to where you were.

AenBleidd:
Ok. From one POV compiler will probably do this for us (only disassembling can tell us whether compiler did this or not). From other POV this doesn't mean that the code should be written non-optimal way. 
I do not agree that the code became harder for understanding: almost all modern IDE highlight selected variable and help navigate to its definition. Also names are descriptive (in this case). So this is a personal taste of everyone. Personally I prefer to write optimal code as much as possible. But I can not do such decisions. These changes are not important so if not - then not. I'll focus on some another and more important changes.

ChristianBeer:
I'll sponsor this PR as I find it very useful. Not all of the occurrences in the diff can be optimized by the compiler and since the variable is declared in the vicinity of the usage it is not harder to read. As @AenBleidd said modern IDEs can highlight variable names and even if not the variable name should be descriptive enough to get the meaning without a big code search.

What needs further changes is the names as the current BOINC Code style says: no camel case (outside the Manager code).

I can't reopen this PR so please open a new one with adjusted variable names.

AenBleidd:
Ok. I'll do this today

AenBleidd:
Created new #1895 instead of this one.

PR #1899, opened 2017-05-02
fix_issu-1898-android-localization

### LINKED ISSUE #1898: Wrong localization settings on Android
Hi,

When I try to build an APK in release mode, I get the following error:
```
/home/vagrant/BOINC/android/BOINC/app/src/main/res/values-ro/strings.xml
Error:(351) Error: "social_invite_content_url" is translated here but not found in default locale

```
The default local has the following values:
```
<string name="social_invite_content_url_google">https://play.google.com/store/apps/details?id=edu.berkeley.boinc</string>
    <string name="social_invite_content_url_amazon">http://www.amazon.com/gp/mas/dl/android?p=edu.berkeley.boinc</string>
```

adamradocz:
The Android BOINC doesn't build in release mode.
Fix #1898


AenBleidd:
I guess this should be done through Transifex service: https://www.transifex.com/projects/p/boinc/

adamradocz:
It wasn't build so, if that's the case, something isn't working.

ChristianBeer:
This needs to be fixed on transifex by translating the android resource for each of the languages and then importing the translations into the repository. The xml files will only get overwritten the next time I import translations. Please see http://boinc.berkeley.edu/trac/wiki/TranslationSystem on how the system works.

PR #1902, opened 2017-05-04
Drupal: Add context details to BOINC translation strings

drshawnkwang:
Added context/location details to all `bts()` function calls.

https://dev.gridrepublic.org/browse/DBOINCP-336

### DIFF
--- a/drupal/sites/default/boinc/modules/boincwork/boincwork.module
+++ b/drupal/sites/default/boinc/modules/boincwork/boincwork.module
@@ -365,7 +365,7 @@ function generalprefs_page($action = null, $venue = null, $advanced = FALSE) {
       $path = 'account/prefs/computing/edit';
       $venues = array(
         "{$path}/generic" => bts('Generic', array(), NULL, 'boinc:account-preferences-location'),
-        "{$path}/home" => bts('Home', array(), NULL, 'boinc:account-preferences-location'),
+        "{$path}/home" => bts('Home', array(), NULL, 'boinc:account-preferences-location:-1:ignoreoverwrite'),
         "{$path}/school" => bts('School', array(), NULL, 'boinc:account-preferences-location'),
         "{$path}/work" => bts('Work', array(), NULL, 'boinc:account-preferences-location')
       );
@@ -537,7 +537,7 @@ function projectprefs_page($action = null, $venue = null) {
       $path = 'account/prefs/project/edit';
       $venues = array(
         "{$path}/generic" => bts('Generic', array(), NULL, 'boinc:account-preferences-location'),
-        "{$path}/home" => bts('Home', array(), NULL, 'boinc:account-preferences-location'),
+        "{$path}/home" => bts('Home', array(), NULL, 'boinc:account-preferences-location:-1:ignoreoverwrite'),
         "{$path}/school" => bts('School', array(), NULL, 'boinc:account-preferences-location'),
         "{$path}/work" => bts('Work', array(), NULL, 'boinc:account-preferences-location')
       );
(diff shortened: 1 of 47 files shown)


drupal/sites/default/boinc/modules/boincwork/boincwork.module
```diff
@@ -635,7 +635,7 @@ function privacyprefs_page($action = null) {
     // Edit preferences link
     $output .= '<tr>';
     $output .= '<td></td>';
-    $output .= '<td>' . l(bts('Edit privacy preferences', array('@project' => PROJECT)), "account/prefs/privacy/edit") . '</td>';
+    $output .= '<td>' . l(bts('Edit privacy preferences', array('@project' => PROJECT)), "account/prefs/privacy/edit", NULL, 'boinc:account-preferences-privacy') . '</td>';
```
tristanolive: Context string was added to the l() function in this one instead of to bts().

drupal/sites/default/boinc/modules/boincwork/includes/boincwork.forms.inc
```diff
@@ -1178,27 +1178,27 @@ function boincwork_projectprefs_form(&$form_state, $venue) {
...
     global $base_path;
     $form['prefs']['form control tabs']['#value'] .= '<li class="tab">' . 
-      l(bts('Clear'), "account/prefs/project/clear/{$venue}",
+      l(bts('Clear', array(), NULL, 'boinc:form-clear'), "account/prefs/project/clear/{$venue}",
         array(
           'query' => 'destination=' . urlencode(drupal_get_path_alias('account/prefs/project/combined')),
           'attributes' => array(
             'onclick' => 'return confirm(\'' . bts('This will remove all of your settings from the @name preference set. Are you sure?',
-              array('@name' => $venue)) . '\')'
+              array('@name' => $venue), array(), NULL, 'boinc:account-preferences') . '\')'
```
tristanolive: Extra array() passed to bts().

drupal/sites/default/boinc/themes/boinc/templates/forum-list.tpl.php
```diff
@@ -110,13 +110,13 @@
...
+    <?php print bts('Please be responsible in what you write and do not create posts which are offensive or insulting. Offensive posts or threads may be deleted by forum moderators without warning or discussion. Do not respond to offensive postings. Click on the "report" button at the bottom of the post to call it to the attention of the moderators.', array(), NULL, 'boinc:forum-fine-print'); ?>
   </p>
   <p>
     <?php print bts('We also ask that you keep all discussion on the message boards related to @project or BOINC with the small exception of the Science message board where you are free to discuss anything relevant to the underlying science. Participants interested in broader discussions should post to unofficial forums for @project.',
-      array('@project' => variable_get('site_name', bts('this project')))); ?>
+      array('@project' => variable_get('site_name', bts('this project'))), NULL, 'boinc:forum-fine-print'); ?>
```
tristanolive: Does nested bts() also need context?

drshawnkwang: In this case, what we should do is find all instances of `variable_get('site_name', XYZ)`. Currently XYZ is very inconsistent. Sometimes its a string, sometimes its blank, sometimes it uses `bts()`, etc.

I propose we change XYZ to `'Drupal-BOINC'`, without `bts()` or `t()`. The idea being that as an admin you should **really** set the `site_name` variable. And that translators should not have to translate this default string that appears in case the admin does not `site_name`, because it should never show up.

PR #1932, opened 2017-06-05
Drupal: Added patch to Views contrib module, 6.x-2.x-boinc-3-dev

drshawnkwang:
to eliminate E_STRICT PHP notices.

https://dev.gridrepublic.org/browse/DBOINCP-380

Drupal issue: https://www.drupal.org/node/893128

### DIFF
--- a/drupal/sites/default/boinc/modules/contrib/views/views.info
+++ b/drupal/sites/default/boinc/modules/contrib/views/views.info
@@ -4,7 +4,7 @@ package = Views
 core = 6.x
 
 ; Information added by Drupal.org packaging script on 2015-02-11
-version = "6.x-2.18-boinc-2-dev"
+version = "6.x-2.18-boinc-3-dev"
 core = "6.x"
 project = "views"
 datestamp = "1494530273"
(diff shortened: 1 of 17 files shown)


drupal/sites/default/boinc/modules/contrib/views/views.info
```diff
@@ -4,7 +4,7 @@ package = Views
 core = 6.x
 
 ; Information added by Drupal.org packaging script on 2015-02-11
-version = "6.x-2.18-boinc-2-dev"
+version = "6.x-2.18-boinc-3-dev"
 core = "6.x"
 project = "views"
 datestamp = "1494530273"
```
tristanolive: Minor adjustment to make: The datestamp in the module .info file should be set to the current UNIX timestamp. The utility of doing this is unclear, since available updates are no longer being flagged as they were prior to Drupal 6 end-of-life, but it may still prove valuable with LTS to know quickly which patches we have already reviewed and which are new.

drshawnkwang:
I completely forgot about this. Thanks for the tip Tristan.

PR #1942, opened 2017-06-15
remove ghost mediums from the media registry

nikolasgianna:
After noticing erratic behavior when a machine was deleted (VDIs sometimes remaining in the media registry) and reproducing the problem with VBoxManage commands it was noticed that in order to replicate what is normally done when deleting a machine via VirtualBox Manager (right click on the machine and remove) all that was needed was the command:
VBoxManage -q unregistervm <vm_name> --delete
That way VirtualBox took care of deleting everything. So the commands following this were removed.

### DIFF
--- a/samples/vboxwrapper/vbox_vboxmanage.cpp
+++ b/samples/vboxwrapper/vbox_vboxmanage.cpp
@@ -706,14 +706,6 @@ int VBOX_VM::deregister_vm(bool delete_media) {
         vbm_popen(command, output, "network throttle group (remove)", false, false);
     }
 
-    // Delete its storage controller(s)
-    //
-    vboxlog_msg("Removing storage controller(s) from VM.");
-    command  = "storagectl \"" + vm_name + "\" ";
-    command += "--name \"Hard Disk Controller\" ";
-    command += "--remove ";
-
-    vbm_popen(command, output, "deregister storage controller (fixed disk)", false, false);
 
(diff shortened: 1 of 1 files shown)


ChristianBeer:
May I suggest to squash the two commits and put the explanation for deletion given as comments in the first commit as the commit message of the new single commit? the explanation in the PR still remains but the commit message is more easily accessible should someone in the future stumble upon the commit itself.

romw:
Nikolas,

What versions of VirtualBox was this tested against?

All,

How many product generations do we want to support of VirtualBox?  I think originally I was trying to support all the way back to 4.0.x, in theory.

davidpanderson:
There's no harm in leaving in 4.x support.

My understanding is that projects using Docker require 5.x.  If we reach the point where all projects require 5.x, we can remove 4.x-specific code.

nikolasgianna:
The commits were squashed. As for the versions it was tested against, they were 5.1.14 and 5.1.22 for Windows and 5.1.22 for Linux. Was it a requirement for previous versions to also remove the vdi separately?

lfield:
All VritualBox version less than 5.1 are no longer supported. In LHC@home the minimum required version is 5.0 due to problems with version 4.x. In addition as Windows 10 has issues with some 5.0 versions we will shortly require 5.1 as a minimum. We would recommend that all other projects using VirtualBox do the same.

PR #1943, opened 2017-06-18
Adding user->id to be return with an e-mail lookup

iamtheghostlove:
For 3rd parties e-mail verification is the only way to verify users. This allows third parties to match a users id by his e-mail address, and thus securely identify their ownership of that account. 

Using this information we can access show_user.php?userid=id and be certain of an e-mails credit ownership.

### DIFF
--- a/html/user/lookup_account.php
+++ b/html/user/lookup_account.php
@@ -63,6 +63,7 @@ if (LDAP_HOST && $ldap_auth) {
     if (!$passwd_hash) {
         echo "<account_out>\n";
         echo "   <success/>\n";
+        echo "<id>$user->id</id>\n";
         echo "</account_out>\n";
         exit();
     }
--- a/html/user/lookup_account.php
+++ b/html/user/lookup_account.php
@@ -63,7 +63,7 @@ if (LDAP_HOST && $ldap_auth) {
     if (!$passwd_hash) {
         echo "<account_out>\n";
         echo "   <success/>\n";
-        echo "<id>$user->id</id>\n";
+        echo "   <id>$user->id</id>\n";
         echo "</account_out>\n";
         exit();
     }


html/user/lookup_account.php
```diff
@@ -63,6 +63,7 @@
     if (!$passwd_hash) {
         echo "<account_out>\n";
         echo "   <success/>\n";
+        echo "<id>$user->id</id>\n";
```
ChristianBeer: Please also indent the new response line as was done to the existing line above.

iamtheghostlove:
Of course, my mistake. Thanks for the speedy review.

JuhaSointusalo:
I don't like this. BOINC allows its users to be anonymous and the way this is implemented partially breaches that anonymity.

While this change may be intended to allow verifying users signing up to third party services in reality this allows anyone to check any email address and then, with the user id, see the person's activity on the project.

@iamtheghostlove In boinc_dev as an alternative you suggested looking up credits by email address. Please notice that that would still allow linking email addresses to user ids.

davidpanderson:
I agree w/ Juha that lookup by email addr should return only existence/nonexistence of the account
(which can be found by other means, i.e. trying to create the account).

Getting anything more should require a password hash.

iamtheghostlove:
I don't think that it is secure for me to ask my users for their authentication strings since it exposes my ability to change their project preferences. Is this your suggestion for accessing the WebRPC? If it is, I can work with that, but I believe it is less secure then this alternative.

Is it correct to say your concern is more with the rest of the information exposed by the show_user request as opposed to the actual credit information? I'm unsure why viewing a users credits earned exposes them to anything malicious. Is the credit information something you explicitly want hidden?

If there was another call such as (e-mail -> only return credits earned, no additional information) would this be more acceptable? 

Or perhaps a hash could be provided by the API that references their credit info, as long as it was separate from the authentication hash used to modify their project preferences. The users could supply me with their "3rd Party Hash" to access a limited amount of information.

Another option is allowing the users to specify within their project options that they would like to expose their credit information to third parties. 

Lastly there is a 4th option, where I could supply my users with a hash to put into their user info such as country, or url. Then if they give me their ID, and I can make sure they own the account by checking this field to the hash I have provided. This however seems like a convoluted work around, that would need to be unique to each 3rd party. In this same manner I could have them post to the forums with this hash, and look up their most recent forum post, but I do not want to clog the forums with such posts.

Please advise on what is best.

Thank you for your help.

davidpanderson:
If the RPC returns the user ID, then you can use that to view the user page, which includes signup time, list of computers, list of other projects, team, profile, etc.

Being able to get this info based on email address will be viewed as an invasion of privacy.

If we allow credit lookup based on email address, someone could match the credit numbers to an account ID (using the top-user list or XML export) and do the same thing.

iamtheghostlove:
Okay, I understand, so you would not like an e-mail address to connect to anymore information what-so-ever. I can see why some users are concerned about this.

How do you feel about these other options I've suggested?

1. Allowing users to check an option in their preferences 'Expose ID to third parties' to show e-mail->id if they have decided they are not concerned with this.

2. Providing a hash in their profile section, which they can provide to third parties to verify their ownership of the account. This is the same as the authentication hash but is read-only with the API.

ChristianBeer:
I don't think this PR is the correct place to discuss the problem this PR should address. Please open an issue about your problem. Which is if I understand correctly that you want to verify that a userid and email-address provided by a user belong together and that user has control over the account. That is basically what the Gridcoin community also wants (I don't know if you are a part of this) and they have already pondered about possible solutions but so far didn't come forward with a viable one.

PR #1946, opened 2017-06-20
Modifications that have been included in the version that is currently used by LHC@home.

nikolasgianna:
This is the version being used from LHC@home currently.

### DIFF
--- a/samples/vboxwrapper/vbox_common.cpp
+++ b/samples/vboxwrapper/vbox_common.cpp
@@ -123,7 +123,6 @@ VBOX_BASE::VBOX_BASE() : VBOX_JOB() {
     vboxsvc_pid = 0;
 
     log_pointer = 0;
-    state = "poweredoff";
 
 #ifdef _WIN32
     vm_pid_handle = 0;
@@ -417,6 +416,7 @@ string VBOX_BASE::read_vm_log(){
     size_t line_end;
     size_t line_start;
     string msg;
+    static string state = "poweredoff";
     string virtualbox_vm_log;
     virtualbox_vm_log = vm_master_name + "/Logs/VBox.log";
 
--- a/samples/vboxwrapper/vbox_common.cpp
+++ b/samples/vboxwrapper/vbox_common.cpp
@@ -123,7 +123,7 @@ VBOX_BASE::VBOX_BASE() : VBOX_JOB() {
     vboxsvc_pid = 0;
 
     log_pointer = 0;
-    state = "PoweredOff";
+    state = "poweredoff";
 
 #ifdef _WIN32
     vm_pid_handle = 0;
--- a/samples/vboxwrapper/vbox_common.cpp
+++ b/samples/vboxwrapper/vbox_common.cpp
@@ -25,6 +25,7 @@
 #endif
 
 #else
+#include <algorithm>
 #include <sys/wait.h>
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -439,6 +440,8 @@ string VBOX_BASE::read_vm_log(){
                 sanitize_format(msg);
 
                 state = msg.c_str();
+
+		std::transform(state.begin(), state.end(), state.begin(), ::tolower);
             }
 
             line_pos = line.find("Guest Log:");
--- a/samples/vboxwrapper/vbox_common.cpp
+++ b/samples/vboxwrapper/vbox_common.cpp
@@ -161,7 +161,9 @@ int VBOX_BASE::run(bool do_restore_snapshot) {
 
         if (is_vm_machine_configuration_available()) {
             retval = register_vm();
+          
             if (retval){
+
                 vboxlog_msg("Could not register");
                 return retval;
             }
@@ -171,18 +173,21 @@ int VBOX_BASE::run(bool do_restore_snapshot) {
                 // was already initialized for the current slot directory but aborted
                 // while the task was suspended and unloaded from memory.
                 retval = deregister_stale_vm();
+              
                 if (retval){
+
                     vboxlog_msg("Could not deregister stale VM");
                     return retval;
                 }
             }
             retval = create_vm();
+
             if (retval){
+
                 vboxlog_msg("Could not create VM");
                 return retval;
             }
         }
-
     }
 
     // The user has requested that we exit after registering the VM, so return an
@@ -199,7 +204,9 @@ int VBOX_BASE::run(bool do_restore_snapshot) {
     if (online) {
         vboxlog_msg("VM was running");
         retval = poweroff();
+
         if (retval){
+
             vboxlog_msg("Could not stop running VM");
             return ERR_NOT_EXITED;
         }
@@ -209,7 +216,9 @@ int VBOX_BASE::run(bool do_restore_snapshot) {
     // saved snapshot
     if (do_restore_snapshot) {
         retval = restore_snapshot();
+
         if (retval){
+
             vboxlog_msg("Could not restore from snapshot");
             return retval;
         }
@@ -217,7 +226,7 @@ int VBOX_BASE::run(bool do_restore_snapshot) {
 
     // Has BOINC signaled that we should quit?
     // Try to prevent starting the VM in an environment where we might be terminated any
-    // second.  This can happen if BOINC has been told to shutdown or the volunteer has 
+    // second.  This can happen if BOINC has been told to shutdown or the volunteer has
     // told BOINC to switch to a different project.
     //
     if (boinc_status.no_heartbeat || boinc_status.quit_request) {
@@ -226,7 +235,9 @@ int VBOX_BASE::run(bool do_restore_snapshot) {
 
     // Start the VM
     retval = start();
+
     if (retval){
+
         vboxlog_msg("Could not start ");
         return retval;
     }
@@ -408,6 +419,7 @@ string VBOX_BASE::read_vm_log(){
     string msg;
     string virtualbox_vm_log;
     virtualbox_vm_log = vm_master_name + "/Logs/VBox.log";
+
     string console = "Console: Machine state changed to \'";
 
     std::ifstream  src(virtualbox_vm_log.c_str(), std::ios::binary);
@@ -443,6 +455,7 @@ string VBOX_BASE::read_vm_log(){
             }
             log_pointer = src.tellg();
         }
+
         return state;
     }
     else return "Error in parsing the log file";
@@ -1021,6 +1034,7 @@ int VBOX_BASE::launch_vboxvm() {
 
     // Execute command
     if (!CreateProcess(
+
                 NULL, 
                 cmdline,
                 NULL,
@@ -1031,7 +1045,9 @@ int VBOX_BASE::launch_vboxvm() {
                 NULL,
                 &si,
                 &pi
+
                 )) {
+
         vboxlog_msg(
                 "Status Report: Launching virtualbox.exe/vboxheadless.exe failed!."
                 );
@@ -1041,7 +1057,7 @@ int VBOX_BASE::launch_vboxvm() {
                 GetLastError()
                 );
         goto CLEANUP;
-    } 
+    }
 
     while(1) {
         GetExitCodeProcess(pi.hProcess, &ulExitCode);
@@ -1162,7 +1178,7 @@ int VBOX_BASE::vbm_popen(string& command, string& output, const char* item, bool
             // lock is held by a previous instance of vboxmanage whos instance data hasn't been
             // cleaned up within vboxsvc yet.
             //
-            // Error Code: VBOX_E_INVALID_OBJECT_STATE (0x80bb0007) 
+            // Error Code: VBOX_E_INVALID_OBJECT_STATE (0x80bb0007)
             //
             if (VBOX_E_INVALID_OBJECT_STATE == (unsigned int)retval) {
                 if (retry_notes.find("Another VirtualBox management") == string::npos) {
@@ -1183,7 +1199,7 @@ int VBOX_BASE::vbm_popen(string& command, string& output, const char* item, bool
             // Experiments performed by jujube suggest changing the sleep interval to an exponential
             // style backoff would increase our chances of success.
             //
-            // Error Code: CO_E_SERVER_EXEC_FAILURE (0x80080005) 
+            // Error Code: CO_E_SERVER_EXEC_FAILURE (0x80080005)
             //
             if (CO_E_SERVER_EXEC_FAILURE == (unsigned int)retval) {
                 if (retry_notes.find("Unable to communicate with VirtualBox") == string::npos) {
@@ -1196,6 +1212,7 @@ int VBOX_BASE::vbm_popen(string& command, string& output, const char* item, bool
             }
 
             // Retry?
+
             if (!retry_failures && 
                     (VBOX_E_INVALID_OBJECT_STATE != (unsigned int)retval) && 
                     (CO_E_SERVER_EXEC_FAILURE != (unsigned int)retval)
@@ -1287,7 +1304,9 @@ int VBOX_BASE::vbm_popen_raw(
 
     // Execute command
     if (!CreateProcess(
+
                 NULL, 
+
                 (LPTSTR)command.c_str(),
                 NULL,
                 NULL,
@@ -1298,6 +1317,7 @@ int VBOX_BASE::vbm_popen_raw(
                 &si,
                 &pi
                 )) {
+
         vboxlog_msg("CreateProcess failed! (%d).", GetLastError());
         goto CLEANUP;
     }
@@ -1374,7 +1394,7 @@ CLEANUP:
 
     // Execute command
     fp = popen(modified_command.c_str(), "r");
-    if (fp == NULL){
+    if (fp == NULL) {
         vboxlog_msg("vbm_popen popen failed! (%d).", errno);
         retval = ERR_FOPEN;
     } else {
--- a/samples/vboxwrapper/vbox_common.cpp
+++ b/samples/vboxwrapper/vbox_common.cpp
@@ -375,10 +375,6 @@ string VBOX_BASE::get_error(int num){
 		    More information can be found here: https://en.wikipedia.org/wiki/X86_virtualization\n \
 		    Error Code: ERR_CPU_VM_EXTENSIONS_DISABLED\n",
 		
-		"   NOTE: VirtualBox has reported an improperly configured virtual machine. It was configured to require\n \
-		    hardware acceleration for virtual machines, but your processor does not support the required feature.\n \
-		    Please report this issue to the project so that it can be addresssed.\n",
-		
 		"   NOTE: VirtualBox hypervisor reports that another hypervisor has locked the hardware acceleration\n \
 		    for virtual machines feature in exclusive mode.\n",
 		
@@ -390,14 +386,14 @@ string VBOX_BASE::get_error(int num){
 	
 		    "VM environment needed to be cleaned up.",
        
-		    "Forign VM Hypervisor locked hardware acceleration features.",
+		    "Foreign VM Hypervisor locked hardware acceleration features.",
 
 		    "VM Hypervisor was unable to allocate enough memory to start VM.",
 
 	     	"VM Hypervisor failed to enter an online state in a timely fashion."
 	};
 
-	std::vector<std::string> v(args, args + 14);
+	std::vector<std::string> v(args, args + 13);
 	
 	return v[num];
 }
--- a/samples/vboxwrapper/vbox_common.cpp
+++ b/samples/vboxwrapper/vbox_common.cpp
@@ -120,6 +120,10 @@ VBOX_BASE::VBOX_BASE() : VBOX_JOB() {
     headless = true;
     vm_pid = 0;
     vboxsvc_pid = 0;
+
+    log_pointer = 0;
+    state = "PoweredOff";
+
 #ifdef _WIN32
     vm_pid_handle = 0;
     vboxsvc_pid_handle = 0;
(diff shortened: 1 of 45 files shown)


marius311:
Is it possible to get a general breakdown of what exactly these changes are and why they were made? I'm curious if any of these are in response to any of the vboxwrapper issues that we run into at Cosmology@Home.

lfield:
Over the past few months we have invested considerable effort to chase all the issues that we have been seeing with the vboxwrapper. We have arrived at a point where we are now confident the issues we saw have been solved or avoided. A phone conference may be the best forum to discuss all the details. The main difference is that the Windows version follows the same approached as for Linux and Mac. The other main change is that the polling function has been changed to avoid hammering on the showvminfo command every second and it now extracts the relevant information from the log file. This changes from a pull model to a push model by using the log file as a one-way message bus.

romw:
@nikolasgianna I went ahead and merged the 'ghost vdi' pull request and it seemed to have an adverse reaction with this pull request.

Could you go ahead and update this pull request with the de-conflicted state.  I'll go ahead and complete the pull request after that.

Two things I did notice that you might want to do while you are in there making changes are:
1. Provide a #define symbol as a parameter to pVM->get_error().  Straight-up numbers just seems wrong to me.
2. Changes to the startup banner for vboxwrapper give a false impression.  The BOINC client version that is running the instance of vboxwrapper is actually in the init_data.xml at runtime.  the #define's used in that context describe what major/minor version of the build tree used to build vboxwrapper, hence why it was originally included as part of vboxwrapper's build number.  If the intent is to provide the client version, you'll need to add a log message someplace after 'boinc_get_init_data(aid);' to actually detect which version of the BOINC client is running vboxwrapper.

samples/vboxwrapper/vbox_common.cpp
```diff
@@ -262,51 +283,206 @@ void VBOX_BASE::dump_hypervisor_logs(bool include_error_logs) {
...
+	virtualbox_vm_log = vm_master_name + "/Logs/VBox.log";
+	
+
+	std::ifstream  src(virtualbox_vm_log.c_str(), std::ios::binary);
+	    
+	if ((src.is_open()) && (src.seekg(log_pointer))) {
+
+            while (std::getline(src, line)) {
+
+                line_start = line.find("Console:");
```
JuhaSointusalo: In the one VBox.log that I have at hand I have this line in it:

    97:40:20.193248 Console::powerDown(): A request to power off the VM has been issued (mMachineState=Saving, InUninit=0)

How does the rest of the vboxwrapper behave if this happens to be the last line with `Console:` in it and `state` then becomes garbage?

samples/vboxwrapper/vbox_common.cpp
```diff
@@ -262,51 +283,206 @@ void VBOX_BASE::dump_hypervisor_logs(bool include_error_logs) {
...
+                    msg = line.substr(line_start, line_end - line_start);
+                    sanitize_format(msg);
+                    state = msg.c_str();
+                }
+                line_pos = line.find("Guest Log:");
+                if (line_pos != string::npos) {
+					
+                    msg = line.substr(line_pos, line.size() - line_pos);
+                    sanitize_format(msg);
+                    vboxlog_msg(msg.c_str());
```
JuhaSointusalo: It's not entirely clear to me if it's required to open the file in binary mode in order to use `tellg` and `seekg` on Windows. Assuming it is, you need to remove `'\r'` from the end of the message. There appears to be `sanitize_output` that does it.

samples/vboxwrapper/vbox_common.cpp
```diff
@@ -262,51 +283,206 @@ void VBOX_BASE::dump_hypervisor_logs(bool include_error_logs) {
...
+                }
+                log_pointer = src.tellg();
+           }
+           return state.c_str();
```
JuhaSointusalo: `c_str` is unnecessary here as you are returning an `std::string`.

samples/vboxwrapper/vbox_common.cpp
```diff
@@ -262,51 +283,206 @@ void VBOX_BASE::dump_hypervisor_logs(bool include_error_logs) {
...
+    string line;
+    size_t line_pos;
+    VBOX_TIMESTAMP current_timestamp;
+    string msg;
+	string virtualbox_vm_log;
+	virtualbox_vm_log = vm_master_name + "/Logs/VBox.log";
+
+	if (boinc_file_exists(virtualbox_vm_log.c_str())) {
+
+		std::ifstream  src(virtualbox_vm_log.c_str(), std::ios::binary);
```
JuhaSointusalo: No need to open the file in binary mode here.

samples/vboxwrapper/vbox_common.cpp
```diff
@@ -262,51 +283,206 @@ void VBOX_BASE::dump_hypervisor_logs(bool include_error_logs) {
...
+		while (std::getline(src, line))
+		{
+			line_pos = line.find("Guest Log:");
+			if (line_pos != string::npos) {
+				sscanf(
```
JuhaSointusalo: I would check the return value of `sscanf`. Otherwise you risk setting `vm_log_timestamp` to garbage.

samples/vboxwrapper/vbox_common.h
```diff
@@ -18,8 +18,8 @@
 
 // Provide cross-platform interfaces for making changes to VirtualBox
 
-#ifndef BOINC_VBOX_COMMON_H
-#define BOINC_VBOX_COMMON_H
+#ifndef _VBOX_COMMON_H_
+#define _VBOX_COMMON_H_
```
JuhaSointusalo: Identifiers that begin with underscore and uppercase letter are reserved for implementation. The code in master was just corrected.

samples/vboxwrapper/vbox_vboxmanage.cpp
```diff
@@ -1721,41 +1698,53 @@ int VBOX_VM::get_vm_network_bytes_received(double& received) {
...
+
+		std::ifstream  src(virtualbox_vm_log.c_str(), std::ios::binary);
+		while (std::getline(src, line))
+		{
+
+			found = line.find(comp);
+			if (found != string::npos){
+
+				pid_start = found;//line.find("Process ID: ");
+				if (pid_start == string::npos) {
```
JuhaSointusalo: This test is always false.

samples/vboxwrapper/vbox_vboxmanage.cpp
```diff
@@ -1721,41 +1698,53 @@ int VBOX_VM::get_vm_network_bytes_received(double& received) {
...
+					vboxlog_msg("Something wrong with the process ID finding\n %s ", line.c_str());
+					return ERR_NOT_FOUND;
+				}
+				pid_start += strlen("Process ID: ");
+				pid_end = line.find("\n", pid_start);
```
JuhaSointusalo: `pid_end` is unused.

samples/vboxwrapper/vboxwrapper.cpp
```diff
@@ -735,17 +728,47 @@ int main(int argc, char** argv) {
...
+		random_checkpoint_factor = drand() * 600;
+
+		vboxlog_msg(
+			"Feature: Checkpoint interval offset (%d seconds)",
+			(int)random_checkpoint_factor
+			);
+
+		// Display trickle value if specified
+		//
+		if (trickle_period > 0.0) {
```
JuhaSointusalo: Why showing trickle period depends on whether automatic checkpoints are enabled?

samples/vboxwrapper/vbox_common.cpp
```diff
@@ -262,51 +283,206 @@ void VBOX_BASE::dump_hypervisor_logs(bool include_error_logs) {
...
+	std::ifstream  src(virtualbox_vm_log.c_str(), std::ios::binary);
+	    
+	if ((src.is_open()) && (src.seekg(log_pointer))) {
+
+            while (std::getline(src, line)) {
+
+                line_start = line.find("Console:");
+		if (line_start != string::npos) {
+					
+                    line_start += 35;
```
JuhaSointusalo: Please don't use magic numbers.

samples/vboxwrapper/vboxwrapper.cpp
```diff
@@ -767,103 +784,49 @@ int main(int argc, char** argv) {
...
         const char*  temp_reason = "";
 
         if (VBOXWRAPPER_ERR_RECOVERABLE == retval) {
-            error_reason =
-                "    BOINC will be notified that it needs to clean up the environment.\n"
-                "    This is a temporary problem and so this job will be rescheduled for another time.\n";
+			error_reason = pVM->get_error(ENV_UNCLEAN);
             unrecoverable_error = false;
-            temp_reason = "VM environment needed to be cleaned up.";
+            temp_reason = pVM->get_error(VM_ENV).c_str();
```
JuhaSointusalo: Here, what you get from `pVM->get_error` is temporary object which is deleted after the statement. The pointer you get from `c_str` must not outlive the temporary object. You made this mistake multiple times. I'm flagging only this one but correct all of them.

nikolasgianna:
Changed as asked. Please mention anything else you spot.

samples/vboxwrapper/vbox_vboxmanage.cpp
```diff
@@ -58,821 +58,816 @@ using std::string;
...
+        //   Accessible:           yes
+        //   Logical size:         10240 MBytes
+        //   Current size on disk: 0 MBytes
+        //   Type:                 normal (base)
+        //   Storage format:       VDI
+        //   Format variant:       dynamic default
+        //   In use by VMs:        test2 (UUID: 000ab2be-1254-4c6a-9fdc-1536a478f601)
+        //   Location:             C:\Users\romw\VirtualBox VMs\test2\test2.vdi
+        //
+        if (enable_isocontextualization && enable_isocontextualization) {
```
JuhaSointusalo: Testing the same flag twice. It should probably be `if (enable_isocontextualization && enable_cache_disk)` since the code is checking for cache disk and to match [the code in the COM interface](https://github.com/BOINC/boinc/blob/master/samples/vboxwrapper/vbox_mscom_impl.cpp#L1171). Although, the code seem confused:

```
if (enable_isocontextualization && ???) {
    blah
} else {
    if (...) {
        blah
    } else {
        if (enable_isocontextualization) {
            if (enable_cache_disk) {
```

It seems one can't have cache disk without ISO contextualization so the last one or two tests are always false. I'm not sure how it's supposed to work. @romw comment?

JuhaSointusalo:
I think you had used tabs to indent instead of spaces earlier. That's why asked to fix indentation. The changeset can be made readable by adding `?w=1` at the end of the URL: https://github.com/BOINC/boinc/pull/1946/files?w=1 . Sorry about that. I'll try to be phrase myself more carefully next time.

nikolasgianna:
Is there anything else I should add or modify to this pull request?

lfield:
Are there any objections to this pull request? This has been running in production on LHC@home for over 3 months now.

samples/vboxwrapper/vbox_vboxmanage.cpp
ChristianBeer: This code changed a bit since Juhas comment but still needs to be cleaned because it now tests the same flag twice.

samples/vboxwrapper/vbox_vboxmanage.cpp
```diff
@@ -58,821 +58,816 @@ using std::string;
...
+            } else {
+                // Did the user delete the VM in VirtualBox and not the medium?  If so,
+                // just remove the medium.
+                command  = "closemedium disk \"" + virtual_machine_slot_directory + "/" + image_filename + "\" ";
+                vbm_popen(command, output, "remove virtual disk", false, false);
+                if (enable_floppyio) {
+                    command  = "closemedium floppy \"" + virtual_machine_slot_directory + "/" + floppy_image_filename + "\" ";
+                    vbm_popen(command, output, "remove virtual floppy disk", false, false);
+                }
+                if (enable_isocontextualization) {
```
ChristianBeer: This is code that can never be reached. The logic is:
```
if (enable_isocontextualization && enable_isocontextualization) {
    blah
} else {
    if (...) {
        blah
    } else {
        if (enable_isocontextualization) {
```

samples/vboxwrapper/vbox_vboxmanage.cpp
```diff
@@ -58,821 +58,816 @@ using std::string;
...
-    if (vbm_popen(command, output, "VM state", false, false, 45, false) == 0) {
-        vmstate_start = output.find("VMState=\"");
-        if (vmstate_start != string::npos) {
-            vmstate_start += 9;
-            vmstate_end = output.find("\"", vmstate_start);
-            vmstate = output.substr(vmstate_start, vmstate_end - vmstate_start);
+        //
+        // What state is the VM in?
+        //
+        vmstate =read_vm_log();
```
ChristianBeer: This reads the status directly from the VirtualBox log. The subsequent string comparisons are using capitalized names now and where using lower case names before. It would be better to lower case the string extracted from the log and compare with lowercase names to compensate for past or future changes in VirtualBox.

samples/vboxwrapper/vbox_vboxmanage.cpp
```diff
@@ -883,73 +878,91 @@ int VBOX_VM::poll(bool log_state) {
...
+            }
+            else if (vmstate == "LiveSnapshotting") {
                 online = true;
                 saving = false;
                 restoring = false;
                 suspended = false;
                 crashed = false;
-            } else if (vmstate == "deletingsnapshotlive") {
+            }
+            else if (vmstate == "DeletingsnapshotOnline") {
```
ChristianBeer: The old and new string differ in more than lower/uppercase shouldn't we be backwards compatible with older VirtualBox version? If not the wrapper should have a minimum requirement that is checked at startup. The same is true for some of the following state names.

samples/vboxwrapper/vbox_vboxmanage.cpp
nikolasgianna: The change is because VirtualBox writes the states in VBox.log with this capitalisation. Previously, lowercase was used because VBox uses only lowercase letters for the showvminfo command output (VMstate field). 

The procedure followed by VirtualBox to translate the MachineState:

https://www.virtualbox.org/browser/vbox/trunk/src/VBox/Frontends/VBoxManage/VBoxManageInfo.cpp#L135

samples/vboxwrapper/vbox_vboxmanage.cpp
nikolasgianna: If we go with this approach we would not depend on the showvminfo command and thus as long as VirtualBox does not change it's MachineState names we would not need to change these strings.

samples/vboxwrapper/vbox_vboxmanage.cpp
nikolasgianna: Should I go ahead and delete completely this piece (the never reached if) or should I copy the functionality someplace else. I am asking because it was not written by me and I am not sure what should be done.

ChristianBeer: I'm also not sure what needs to be done. I remember that isocontextualization was added for CERN so if it still works for you after removing and you see no need for the deleted functionality I'm fine with it.

samples/vboxwrapper/vbox_vboxmanage.cpp
ChristianBeer: My point is that both versions (showvminfo and VBox.log output) should work here. Something like:
```
 vmstate =strtolower( read_vm_log() );
...
else if (vmstate == "deletingsnapshotonline" || vmstate="deletingsnapshotlive") {
...
}
```
This way we can reuse this part of the code for the old showvminfo output as well as the new VBox.log output

nikolasgianna:
I left the iso removing in and it seems to work fine. Also put in support for both showvminfo and vm log options.

lfield:
Just for info. LHC@home does not need the iso functionality any more. I had an exchange with Rom sometime ago where he asked me about this as he wanted to drop support for it.

samples/vboxwrapper/vbox_common.cpp
```diff
@@ -253,34 +269,201 @@ void VBOX_BASE::dump_hypervisor_logs(bool include_error_logs) {
...
+            if (line_start != string::npos) {
+
+                line_start += strlen(console.c_str());
+                line_end = line.find("\'", line_start);
+
+                msg = line.substr(line_start, line_end - line_start);
+
+                sanitize_format(msg);
+
+                state = msg.c_str();
```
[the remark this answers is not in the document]
nikolasgianna: It is declared in vbox_common.h

samples/vboxwrapper/vbox_common.h
```diff
@@ -97,161 +110,181 @@ struct VBOX_TIMESTAMP {
...
+        // whether we were instructed to only register the VM.
+        // useful for debugging VMs.
+        bool register_only;
+        // for optional remote desktop; dynamically assigned
+        int rd_host_port;
+        bool headless;
+
+
+        std::streamoff log_pointer;
+        std::string state;
```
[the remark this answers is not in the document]
nikolasgianna: The state is changed only if new lines are added to the log. State holds the last read state even if no new lines are added to the log.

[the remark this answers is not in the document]
JuhaSointusalo: Apart from initialisation, `state` is accessed only in `VBOX_BASE::read_vm_log`. Since it's used as a temporary variable I don't see the need to have it declared a class member variable.

State string is already held in `static string vmstate_old` in `VBOX_VM::poll`. If you want to have the last read state string available outside of `VBOX_VM::poll` you could a bit of refactoring. I don't think it's useful to have two variables for the same thing.

nikolasgianna:
I went with turning 'state' into a static variable with a default value of "poweredoff". That way it doesn't have to be outside the function but will still hold the last read state. Thanks for the suggestion.

samples/vboxwrapper/vboxwrapper.cpp
```diff
@@ -906,95 +871,37 @@ int main(int argc, char** argv) {
...
-            }
+        pVM->report_clean(unrecoverable_error, skip_cleanup, do_dump_hypervisor_logs,
+                retval, error_reason, pVM->vm_pid, temp_delay, temp_reason,
+                current_cpu_time, last_checkpoint_cpu_time, fraction_done,
+                bytes_sent, bytes_received);
 
-            if (do_dump_hypervisor_logs) {
-                pVM->dump_hypervisor_logs(true);
-            }
+        if (unrecoverable_error) checkpoint.update(elapsed_time, current_cpu_time);
```
JuhaSointusalo: This code needs to go before `report_clean` because `report_clean` doesn't return.

JuhaSointusalo:
@nikolasgianna @lfield Two questions:

1. Have you tested how your changes work if the VM process or VirtualBox service crashes? That is, are crashes logged in such a way that you can detect them?
2. Have you tested that the COM interface still works even with your changes, that there is no accidental breakage?

nikolasgianna:
1. In case the VM simply crashes or exits unexpectedly there is the heartbeat mechanism to protect against a rogue VM that's not under our control. After 20 minutes it will kick in and clean up.
2. As far as I tested and with the latest changes the COM inteface functions exactly like before (except for the poll function).

samples/vboxwrapper/vbox_common.cpp
```diff
@@ -190,11 +200,12 @@ int VBOX_BASE::run(bool do_restore_snapshot) {
     vm_name = vm_master_name;
 
     // Check to see if the VM is already in a running state, if so, poweroff.
-    poll(false);
     if (online) {
```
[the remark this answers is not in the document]
nikolasgianna: Simply putting the poll back doesn't do the trick because in this point we have no VM log to parse so it always fails. Had to bring back the old poll function (poll2) as you suggested and use it at this point.

samples/vboxwrapper/vbox_vboxmanage.cpp
```diff
@@ -966,944 +972,940 @@ int VBOX_VM::poll(bool log_state) {
...
+        string command;
+        string output;
+        int timeout = 0;
 
-    boinc_get_init_data_p(&aid);
+        boinc_get_init_data_p(&aid);
 
 
-    vboxlog_msg("Starting VM using VboxManage interface. (%s, slot#%d)", vm_name.c_str(), aid.slot);
+        log_pointer = 0;
```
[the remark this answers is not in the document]
nikolasgianna: Kept the initialisation inside start() so that it's done every time a VM physically starts.

JacobWKlein:
I just wanted to point out a screenshot of the negative performance impact I noticed, comparing 26200 vs 26197. It might be attributed to changes in this PR. Not sure, but I assume it is, based on timing.

To summarize the screenshot:
- BOINC was crunching 4 RNA Tasks for 5450 minutes CPU Time. All using COM, based on Logs. Using VirtualBox v5.1.38.
- vboxwrapper 26200 used 40.2 additional minutes CPU Time, for a running task, during that time.
- vboxwrapper 26197 used 1.2 additional minutes CPU Time, for a running task, during that time.
- So, performance of 26200 is 33 times worse.

boinc vboxwrapper 26200 vs 26197

lfield:
This is an interesting observation. I would suggest opening an issue report for this.

nikolasgianna:
Since the COM interface is being used and the only change is the poll function, a simple test would be to build the vboxwrapper using the original poll function instead of the one which reads virtualbox's logs. That way, all other things equal, the poll function would definitely be at fault of using too much CPU time.

JacobWKlein:
Per your request, I created Issue 2636:
[https://github.com/BOINC/boinc/issues/2636](https://github.com/BOINC/boinc/issues/2636)

PR #1954, opened 2017-06-23
Web: fix sample project_specific prefs

ChristianBeer:
The allow_non_preferred_apps tag was renamed in 671446ced1 for the scheduler but the sample project_specific prefs were not updated.

Existing projects need to update the file on their own if they use a scheduler that was build after 671446ced1 and rename the old tag in the DB. See html/ops/fix_project_prefs.php to do that.

### DIFF
--- a/html/ops/fix_project_prefs.php
+++ b/html/ops/fix_project_prefs.php
@@ -1,0 +1,218 @@
+#!/usr/bin/env php
+<?php
+// This file is part of BOINC.
+// http://boinc.berkeley.edu
+// Copyright (C) 2017 University of California
+//
+// BOINC is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


davidpanderson:
Let's do it the other way around and change the scheduler code.  No need for DB repair.  I'll do this soon.

ChristianBeer:
The supplied script html/ops/fix_project_prefs.php does more than just rename the allow_non_preferred_apps tag, does your comment mean this script is also rejected as a whole or just the part of renaming the tag in the PHP code?

davidpanderson:
Just renaming allow_non_preferred_apps.
Changing it was a mistake on my part.

AenBleidd:
@ChristianBeer, I found this stale branch. Are you ok to delete it?

PR #1956, opened 2017-06-26
Lib: fix escape_url() for UTF8 encoded strings

### LINKED ISSUE #1951: Lib: escape_url() does not correctly escape UTF8 strings
The [escape_url()](https://github.com/BOINC/boinc/blob/e712c7af9ac5b4f9c16d691e656788a2112cefe3/lib/url.cpp#L161) function does not correctly escape UTF8 characters (umlauts) which in case of the [create_account RPC](https://github.com/BOINC/boinc/blob/e712c7af9ac5b4f9c16d691e656788a2112cefe3/client/acct_setup.cpp#L154) leads to an invalid username error and no account is created.

```
before: aeiou äëïöü áéíóú àèìòù
after: aeiou%20%-#%+$%-#%++%-#%+/%-#%,&%-#%,,%20%-#%+!%-#%+)%-#%+-%-#%,#%-#%,*%20%-#%*0%-#%+(%-#%+,%-#%,"%-#%,)
correct: aeiou%20%C3%A4%C3%AB%C3%AF%C3%B6%C3%BC%20%C3%A1%C3%A9%C3%AD%C3%B3%C3%BA%20%C3%A0%C3%A8%C3%AC%C3%B2%C3%B9
```
Can be tested here: https://www.w3schools.com/tags/ref_urlencode.asp

I see two solutions:

1. fix the existing function to properly recognize UTF8 which in my experience only makes the function more complex and we still might not catch every case
1. find a library that does this kind of stuff and has proven itself to cover all the cases

I also think that this should be part of the next 7.6 maintenance release.

ChristianBeer:
The default modifier for char with GCC is signed but for character encoding we always need to use unsigned as the code tables only have positive values.

fixes #1951

### DIFF
--- a/lib/url.cpp
+++ b/lib/url.cpp
@@ -162,7 +162,8 @@ void escape_url(const char *in, char*out, int out_size) {
     char buf[256];
     int x, y;
     for (x=0, y=0; in[x] && (y<out_size-3); ++x) {
-        if (isalnum(in[x])) {
+        unsigned char c = (unsigned char)in[x];
+        if (c < 0x80 && isalnum(c)) {
             out[y] = in[x];
             ++y;
         } else {
--- a/lib/url.cpp
+++ b/lib/url.cpp
@@ -117,9 +117,9 @@ static char x2c(char *what) {
 
 void c2x(char *what) {
     char buf[3];
-    char num = atoi(what);
-    char d1 = num / 16;
-    char d2 = num % 16;
+    unsigned char num = atoi(what);
+    unsigned char d1 = num / 16;
+    unsigned char d2 = num % 16;
     int abase1, abase2;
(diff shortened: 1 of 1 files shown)


lfield:
Is there any reason why this cannot be merged? It seems like a trivial change.

AenBleidd:
@lfield, as far as I remember @davidpanderson decided to implement this is a different way in the commit ff7633a65def104b81a82b839f98fbe4b383f42f

lfield:
@AenBleidd Thanks. I see that the function expects an unsigned char. 

https://github.com/BOINC/boinc/blob/master/lib/url.cpp#L142

I will close this PR.

AenBleidd:
@ChristianBeer, I found this stale branch. Are you ok to delete it?

PR #1968, opened 2017-07-11
db_purge: use zlib

bema-ligo:
The current db_purge pipes its output through external programs for compression, which leads to bad error handling with possible data loss (e.g. if the disk runs full) and is also risky (what's the current PATH?). Add an option (here: compression type) to use zlib to produce gz files without an external program / pipe.

### DIFF
--- a/sched/Makefile.am
+++ b/sched/Makefile.am
@@ -262,7 +262,7 @@ db_dump_SOURCES = db_dump.cpp
 db_dump_LDADD = $(SERVERLIBS)
 
 db_purge_SOURCES = db_purge.cpp
-db_purge_LDADD = $(SERVERLIBS)
+db_purge_LDADD = $(SERVERLIBS) -lz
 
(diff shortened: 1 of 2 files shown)


ChristianBeer:
I investigated why the Travis build failed.
The `gzbuffer()` function is only available in zlib 1.2.3.5 and newer but the Travis CI image we use has 1.2.3.4 which comes with Ubuntu 12.04. Travis is currently in the process to upgrade build images to Ubuntu 14.04 LTS (which has zlib 1.2.8) but they are still missing packages that we need.

bema-ligo:
Building db_purge should not fail because of a too old zlib. Either we should make zlib 1.2.3.5 a requirement and add a configure check for this version. or we should make this feature work without setting the zlib buffer size, by splitting the text to write into portions of 8k (the default buffer size).

ChristianBeer:
This was reimplemented without `gzBuffer` and some other XML related changes. We'll provide a new PR.

bema-ligo:
#2005

PR #1978, opened 2017-07-17
client: add task name and tweak postponed message

JuhaSointusalo:
Makes it easier to see which task had problems when reading logs afterwards. Also include back off time in notice and drop fractional seconds.


davidpanderson:
Thanks!

PR #1979, opened 2017-07-17
client/lib: don't flush stdout and stderr in main loop

JuhaSointusalo:
The client flushes stdout and stderr at every iteration of the main loop. On Windows, stderr is opened in commit mode and this results in a write to NTFS $Log every time the stream is flushed even if nothing was written to the stream since last flush. These unnecessary writes totals to several hundred megabytes per day. Some users are concerned that this shortens the lifespan of their hardware.

Fix this by removing the flushes from main loop. stderr is unbuffered so it never needed the flush. After freopen() in diagnostics_init() stdout is fully buffered. Change stdout to line buffered mode so that log messages are visible in log files immediately.

MSVCRT doesn't support line buffered streams. It treats them as fully buffered. Emulate line buffered stream by flushing stdout in logging functions when compiling with Visual C++.

### DIFF
--- a/client/client_msgs.cpp
+++ b/client/client_msgs.cpp
@@ -125,6 +125,11 @@ void show_message(
     // print message to the console
     printf("%s", evt_message);
 
+#ifdef _MSC_VER
+    // MSVCRT doesn't support line buffered streams
+    fflush(stdout);
+#endif
+
     // print message to the debugger view port
     diagnostics_trace_to_debugger(evt_message);  
 }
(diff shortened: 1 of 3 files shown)


CharlieFenton:
Will removing the flush() calls cause any issues on platforms other than Windows?

JuhaSointusalo:
It shouldn't. stderr is unbuffered so fflush(stderr) is nothing more than
an expensive no-op.

For stdout, it would only matter if there is some rogue code that outputs
directly to stdout and the output doesn't end in line break. The visibility
of this output would be delayed until the next log message or until stdout's
buffer is filled. I didn't find any such rogue code. There is some code
that printfs error messages but those ended in line break.

I did consider setting stdout to unbuffered mode but decided it would be
unnecessary. Client's logging function ends all messages in line break so
unbuffered mode would have just resulted in unnecessary system calls.

PR #1980, opened 2017-07-17
lib: fix out of bounds write in get_real_executable_path

JuhaSointusalo:
If the link were max_len bytes long or longer, the null terminator would have been written one past the end of path.

### DIFF
--- a/lib/util.cpp
+++ b/lib/util.cpp
@@ -631,7 +631,7 @@ double rand_normal() {
 //
 #ifdef HAVE__PROC_SELF_EXE
 int get_real_executable_path(char* path, size_t max_len) {
-    int ret = readlink("/proc/self/exe", path, max_len);
+    int ret = readlink("/proc/self/exe", path, max_len - 1);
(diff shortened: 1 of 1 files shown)


JuhaSointusalo:
Before I make pull requests I always test the original code that it really has the bug I think it has.

It turns out that it is actually not possible to trigger the out of bounds write with the code in BOINC as it is currently. `readlink` operates on paths up to `PATH_MAX` and at least on Linux `PATH_MAX` includes the null terminator. So, if the path is `PATH_MAX-1` bytes long excluding null terminator `readlink` succeeds and we have space for the null terminator. However, if the path is `PATH_MAX` bytes long excluding null terminator `readlink` fails with `ENAMETOOLONG`.

At the moment the only caller of `get_real_executable_path` uses `PATH_MAX` sized buffer which is large enough so there is no out of bounds write.

But bug is bug and since BOINC has not been too careful with buffer sizes in the past better fix the code.

PR #1981, opened 2017-07-18
Drupal: Add i18n translatable chunks for Transifex

drshawnkwang:
A new input filter is defined in the boinctranslate module. This input filter allows an editor to split the node content into chunks. Each chunk is hashed for translation. When one chunk is altered, the remainder of the page, and their translations, are not affected.

https://dev.gridrepublic.org/browse/DBOINCP-234


drshawnkwang:
boinctranslate.form.inc is new code, no previous code was copied from anywhere. 

boinctranslate.helper.inc contains functions that were in boinctranslate.module. But has one new function `function _boinctranslate_filter_tips_long()`, which is simply text output.

If you like, I can cancel this PR, and try using git rebase to split the code changes into two separate PRs: one to move code into boinctranslate.helper.inc, and second that contains the input filter functionality.

tristanolive:
Considering that no developers will be affected by the blip in the history (and this is by no means the first such blip), I'm ok with just knowing what you shared. It really isn't worth the effort of splitting the commit just for a bit of filter tips output...

PR #1982, opened 2017-07-18
Drupal: Fixed a few bugs and added features to task table.

drshawnkwang:
Fixed bug in table where CPU time was not being loaded correctly.
Removed WHERE deprecated=0 flag from queries.
Added application version and platform to table.
Added sort to Sent column.

https://dev.gridrepublic.org/browse/DBOINCP-387
https://dev.gridrepublic.org/browse/DBOINCP-388

### DIFF
--- a/drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
+++ b/drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
@@ -1640,15 +1640,15 @@ function boincwork_tasktable($category = 0, $queryid = 1, $tselect = NULL, $app_
     ),
     array(
       'data' => bts('Run time', array(), NULL, 'boinc:task-table'),
-      'field'=> 'elapsed_time',
+      'field' => 'elapsed_time',
     ),
     array(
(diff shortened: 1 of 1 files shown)


drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
```diff
@@ -1640,15 +1640,15 @@ function boincwork_tasktable($category = 0, $queryid = 1, $tselect = NULL, $app_
     ),
     array(
       'data' => bts('Run time', array(), NULL, 'boinc:task-table'),
-      'field'=>'elapsed_time',
+      'field'=> 'elapsed_time',
```
ChristianBeer: Please use a consistent style for blanks around `=>`. In other occasions within this file there is a blank before and after the operator. Applies also to lines 1647 and 1651 below.

drshawnkwang: Thanks for noticing this.The subsequent commit should fix this.

Issue #1991, opened 2017-07-26
Docker virtual interface IP blocked by StopForumSpam


marius311:
When running your server inside Docker via boinc-server-docker and testing locally on your machine, `$_SERVER['REMOTE_ADDR']` will generally be `172.17.0.1` which is the address Docker gives to the virtual network interface it creates, 

```
$ ifconfig
docker0   Link encap:Ethernet  HWaddr   
          inet addr:172.17.0.1  Bcast:0.0.0.0  Mask:255.255.0.0
          inet6 addr: fe80::42:1ff:fe9f:b21/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:33430 errors:0 dropped:0 overruns:0 frame:0
          TX packets:41980 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:137755626 (137.7 MB)  TX bytes:359891556 (359.8 MB)
```

This address appears blocked by StopForumSpam https://api.stopforumspam.org/api?ip=172.17.0.1 and so in this scenario you will not be able to create a user account. 

Disabling StopForumSpam isn't a great solution because there's no automated way to do it when testing locally but then not do it when running the server in production. 

I propose we just whitelist this IP explicitly in https://github.com/BOINC/boinc/blob/20d07be2b8ee5c59e585bd95e82f7d99b25f693b/html/inc/email.inc#L98

Thoughts on this solution or other suggestions? I can submit a PR but wanted feedback first.

ChristianBeer:
Using `$_SERVER['REMOTE_ADDR']` for this seems to be safe as it can't be spoofed (to my knowledge). I made a quick test and it seems `192.168.0.0/24` is blocked while `10.0.0.0/8` is not blocked but both are also private ranges. If we trust `$_SERVER['REMOTE_ADDR']` to always contain the real thing then we should whitelist all three private ranges for this. There is also really no need to test against SFS for those addresses.

The only downside is for projects which use a proxy frontend and a private network to reach the webserver. In this case every traffic on the webserver looks like it is originating from the proxy which most likely has a private IP which makes testing the IP useless. So maybe the check should be done before sending the SFS query and remove the IP if it is private and only check the email address.

[...]

PR #1992, opened 2017-07-26
skip SFS check for obviously private IP addresses

marius311:
Fixes https://github.com/BOINC/boinc/issues/1991


davidpanderson:
Thanks!

PR #1996, opened 2017-07-26
Drupal: enable translate node-by-node

drshawnkwang:
Adds feature that allows admin/editor to control whether or not a node is translated on a node-by-node basis.

A new fieldgroup is added to the Page content type. Within is a single field: field_boinctranslate_transifex. This is a simple checkbox. If the field is checked (TRUE), the default, then the node is processed when editted for translation. And the content is filtered using the boinctranslate filter, i.e., the viewed content is translated.

If the checkbox is FALSE, then the node will not be processed using the input filter, meaning the content will not be translated.

Additionally, removed content-type selection from input filter. Now the only way to have a node translated it to have the option 'Export for translation' selected, which is only available in the Page content type. However, it is enabled by default.

https://dev.gridrepublic.org/browse/DBOINCP-394


drshawnkwang:
Closing this PR for now. We want to fix a problem with input filters first. Then this PR will be re-opened.

PR #1997, opened 2017-07-28
Drupal: Fixed bug where anonymous users, who are not logged in, could …

drshawnkwang:
…access and edit the account/profile paths.

Changed access of account/profile, account/profile/view, and account/profile/edit to user_is_logged_in.

https://dev.gridrepublic.org/browse/DBOINCP-398


tristanolive:
Good find!

PR #1999, opened 2017-07-31
Drupal: Fix boinc_standard Feature

drshawnkwang:
Added multiple filter format variables to boinc_standard Feature via strongarm. This sets the input filter settings  from the code.

Changed weights of input filters to use positive integers.

https://dev.gridrepublic.org/browse/DBOINCP-396


tristanolive:
Input filter / Features integration is buggy... thanks for working out these fixes!

PR #2005, opened 2017-08-03
db_purge: use zlib

bema-ligo:
The current `db_purge` pipes its output through external programs for compression, which leads to bad error handling and possible data loss (e.g. if the disk runs full). Add a compress option to use zlib to produce gz files without relying on an external program / pipe.

### DIFF
--- a/sched/Makefile.am
+++ b/sched/Makefile.am
@@ -263,7 +263,7 @@ db_dump_SOURCES = db_dump.cpp
 db_dump_LDADD = $(SERVERLIBS)
 
 db_purge_SOURCES = db_purge.cpp
-db_purge_LDADD = $(SERVERLIBS)
+db_purge_LDADD = $(SERVERLIBS) -lz
 
 trickle_credit_SOURCES = trickle_credit.cpp trickle_handler.cpp
 trickle_credit_LDADD = $(SERVERLIBS)
--- a/sched/db_purge.cpp
+++ b/sched/db_purge.cpp
@@ -39,6 +39,8 @@
 #include <string>
 #include <time.h>
 #include <errno.h>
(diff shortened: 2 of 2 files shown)


AenBleidd:
I think this pull request should be splitted into two: first about xml and second about zlib

bema-ligo:
@AenBleidd 
> I think this pull request should be splitted into two: first about xml and second about zlib

Which of the two PRs should the third commit belong to? Logically it belongs to the XML one, but technically it changes only db_purge and requires the zlib commit.

AenBleidd:
23f0f6e578631811beb705e91bf3f620313fef9b and ad4cd5ce92011393f914556fa9f5e6ac17ec5afa are related to xml only so they can be moved to a separate pull request as far a I see.

bema-ligo:
@AenBleidd moved 23f0f6e and ad4cd5c to #2006

ChristianBeer:
Sorry for creating the conflicts by merging #2021 before this. I didn't see that there was a slight overlap. Can you please rebase or merge master again?

bema-ligo:
@ChristianBeer Not easily, i.e. automatically. This will require more work, and probably re-building the branch, losing the references. I'll try to keep the original SHA1s in the commit messages, though.

bema-ligo:
The conflicting functionality was split out in #2031. Original SHA1s of this branch were kept in "cherry-picked from" comments.

bema-ligo:
From a recent occurrence on Einstein@Home I can confirm that the current db_purge will inevitably lead to data loss when using compression and the disk runs full, and that this PR definitely fixes it. So I strongly recommend to merge at least this one.

davidpanderson:
Thanks!

PR #2007, opened 2017-08-03
Drupal: Option to translate nodes (pages) individually

drshawnkwang:
A new fieldgroup containing one field, which is a boolean allowing the admin/editor to choose whether or not to have this node (page) be translated. Allows translation on a node-by-node basis. Default is TRUE: node is translated.

https://dev.gridrepublic.org/browse/DBOINCP-394


drshawnkwang:
Other PRs which were prerequisites have been merged. Re-opening PR.

Issue #1998, opened 2017-07-29
Addition of cryptocurrency address/account fields to profile?


grctest:
Yo,

I've added ~35 cryptocurrency fields to the profile which doesn't interfere with established SQL tables however does inflate the profile/gz stat extracts, how plausible would it be to merge this into master?

Would it be more likely to be accepted if it was hidden away within its own page rather than visible within the profile? Or if a seperate stats extract (xml.gz) file would be more appropriate?

The justification for the introduction of cryptocurrency fields is that it may help enable an absolute proof of CPID ownership when registering an account for earning rewards on Gridcoin (with the team req removed) and/or on future BOINC cryptocurrencies.

Relevant repo: https://github.com/grctest/project-rain-site
Location of changes: https://github.com/grctest/project-rain-site/tree/master/ProjectRain_Docker/images/makeproject/preCompileReplace
Raised this issue within the '[For the betterment of BOINC](https://boinc.berkeley.edu/dev/forum_thread.php?id=11736&postid=79781)' thread.

Best regards,
CM.

grctest:
Old screenshots (don't have the latest version running at the very moment, sorry):
![image](https://user-images.githubusercontent.com/15861527/28950359-3dfb71f6-78bb-11e7-8f5f-284e57b93ecb.png)

---

The following screenshots show the old project-rain version, the pull request version auto-hides the fields when blank (so as to not immediately spam the profile view with crypto links).

## User profile (self-view)

![image](https://user-images.githubusercontent.com/15861527/28950386-7c143c34-78bb-11e7-9d4d-580962b1f157.png)

## Browsing another user's profile

![image](https://user-images.githubusercontent.com/15861527/28950481-201481ae-78bc-11e7-8c06-030a443af80f.png)

## Example user xml 

```
<?xml version="1.0" encoding="ISO-8859-1" ?>
<user>
    <id>2</id>
    <cpid></cpid>
    <create_time>1497573052</create_time>
    <name>derp</name>
    <country>United Kingdom</country>
    <bitshares>123</bitshares>
    <steem></steem>
    <gridcoin></gridcoin>
    <ethereum></ethereum>
    <ethereum_classic></ethereum_classic>    
    <golem></golem>
    <nxt></nxt>
    <ardor></ardor>
    <hyperledger_sawtooth_lake></hyperledger_sawtooth_lake>
    <hyperledger_fabric></hyperledger_fabric>
    <hyperledger_misc></hyperledger_misc>
[...]

PR #2009, opened 2017-08-04
Project-Rain Merge

grctest:
I have successfully implemented 'project-rain' functionality into the BOINC web server, tested with the boinc-server docker.

I have added cryptocurrency fields to the BOINC stat extracts (user.xml.gz), configured within the user's profile.

I do notice that there are new changes since development (late may), I'll look into updating my version in the near future.

Relevant issue: https://github.com/BOINC/boinc/issues/1998


davidpanderson:
Our policy is to not have code in the BOINC distribution that's directly related to virtual currencies.

grctest:
What about a single custom field which doesn't mention virtual currencies?

Where is this virtual currency policy documented? Is this the stance of the entire PMC?

davidpanderson:
The policy is the consensus of the PMC.

I'm not sure a new DB field would accomplish without accompanying code.  Projects can manually add such a field.  Might be better to use a new table.

grctest:
> The policy is the consensus of the PMC.

Can you direct me to discussions and or minutes related to the PMC's consensus on this topic please? As far as I'm aware the PMC isn't actively governing in a public/transparent manner (other than this enforced undocumented 'policy').

>  Might be better to use a new table.

These fields were added as a new table, rather than integrated within existing tables as you previously instructed (can't locate the old post but I previously attempted to include it within the main tables which caused errors) : https://github.com/grctest/boinc/blob/07c9c3dacd51d0c1fc88170ba0aac142d9eaa2ee/db/schema.sql#L338

---

Might be better to dump the fields to a separate xml gz than user.xml.gz given that it would be massively inflated.

Cheers

ChristianBeer:
This PR needed to be rejected either way because it contains unrelated changes and reverts changes done to master. Two commits that it reverts are: 90f8a1d4b4280ead1 and 20d07be2b8ee5c59e58

PR #1950, opened 2017-06-21
Manager: Use wxHTMLWindow in task property window

AenBleidd:
As requested in #1935 to add the possibility to select any text in this window.

Issue #1935, opened 2017-06-07
Cannot select text in task property window


Germano0:
In BOINC Manager 7.6.22 Linux, it is not possible to select text in working unit property/details window

ChristianBeer:
I'm not sure what you mean. I guess you want to copy the information shown in the task properties window to somewhere else and because you can't select the textboxes this is only possible via a screenprint at the moment. If this is th correct interpretation I'm going to file this as a feature request, if not please elaborate.

Germano0:
zomg I don't know what I had in my mind that day. I meant teXt not teSt. I am going to edit the first message

AenBleidd:
I looked into this and can tell that this is not trivial task.
Maybe we can make 'Copy' button that will copy all text in this window as a workaround?
Will this be acceptable?

Germano0:
Speaking about my needs, I can say that it is acceptable to have a copy button that copy all text in the window

ChristianBeer:
I renamed the title to the actual name. A workunit is a name for something on the server. A Client/Manager will only get tasks which are slightly different things conceptually.

AenBleidd:
@ChristianBeer So, is my solution is ok? If yes - I can implement this in two days. If no - I'll try to find another way to make text selectable.

ChristianBeer:
This is not an urgent issue as we are not able to ship it (no new release in sight). Better to find a solution that makes the textfields selectable by the user. If this is not possible we could still go with the copy button which to me does not seem to be trivial as you need to anticipate additions to this dialog in the future. So ideally it should automatically gather the data displayed and "export" as simple text into the clipboard.

AenBleidd:
All text fields are added to the container so the faster way with button solution is go through all text fields and copy them as simple text into the clipboard.
To make static text fields selectable is hard because:
1 - static text control does not accept user interaction (select/copy...)
2 - It is not easy to use another control and keep current alignment (actually wxtextctrl doesn't have 'autosize' command). I'll try to play with this on weekend, maybe I'll find some solution.

AenBleidd:
Results:
1 - static text control does not accept user interaction (select/copy...)
2 - wxTextCtrl cannot be used to display all information in one control because there is no possibility to use different text styles (e.g. bold and normal). wxTE_RICH & wxTE_RICH2 flags are supported in Win32 only.
[...]

PR #2013, opened 2017-08-06
Manager: Fix Properties Window

AenBleidd:
Fix text styles (font, size) and set backgroung to window default.

### DIFF
--- a/clientgui/DlgItemProperties.cpp
+++ b/clientgui/DlgItemProperties.cpp
@@ -59,20 +59,11 @@ CDlgItemProperties::CDlgItemProperties(wxWindow* parent) :
     
     m_bSizer1 = new wxBoxSizer( wxVERTICAL );
     
-    m_scrolledWindow = new wxScrolledWindow( this, wxID_ANY, wxDefaultPosition, wxDefaultSize, wxHSCROLL|wxVSCROLL );
-    m_scrolledWindow->SetScrollRate( 5, 5 );
-    wxBoxSizer* m_bSizer2;
-    m_bSizer2 = new wxBoxSizer( wxVERTICAL );
-    
     const long style = wxBORDER_NONE;
-    m_txtInformation = new wxHtmlWindow( m_scrolledWindow, wxID_ANY, wxDefaultPosition, wxDefaultSize, style, "" );
+    m_txtInformation = wxWebView::New( this, wxID_ANY, wxWebViewDefaultURLStr, wxDefaultPosition, wxDefaultSize, wxWebViewBackendDefault, style );
+    m_txtInformation->EnableContextMenu( false );
 
-    m_bSizer2->Add( m_txtInformation, 1, wxEXPAND, 5 );
-
-    m_scrolledWindow->SetSizer( m_bSizer2 );
-    m_scrolledWindow->Layout();
-    m_bSizer2->Fit( m_scrolledWindow );
-    m_bSizer1->Add( m_scrolledWindow, 1, wxEXPAND | wxALL, 5 );
+    m_bSizer1->Add( m_txtInformation, 1, wxEXPAND | wxALL, 5 );
     
     m_btnClose = new wxButton( this, wxID_OK, _("&Close"), wxDefaultPosition, wxDefaultSize, 0 );
     m_btnClose->SetDefault(); 
@@ -357,7 +348,6 @@ void CDlgItemProperties::renderInfos(PROJECT* project_in) {
         addProperty(_("Last scheduler reply"), dt.Format());
     }
     renderInfos();
-    m_scrolledWindow->FitInside();
 }
 
 // show task properties
@@ -445,7 +435,6 @@ void CDlgItemProperties::renderInfos(RESULT* result) {
         addProperty(_("Executable"), wxString(avp->exec_filename, wxConvUTF8));
     }
     renderInfos();
-    m_scrolledWindow->FitInside();
 }
 
 //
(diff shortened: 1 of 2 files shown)


AenBleidd:
Fixes are made by request in #1935 after #1950 
I have no possibility to test these changes on Linux and MacOS.
@JuhaSointusalo could you please test this?

CharlieFenton:
This fixes the background color on Macintosh, but copy (command-C on the mac) no longer works. One way to fix this is to add a "Copy selected" button where clicking on the button invokes just this one line of code:
     m_txtInformation->Copy();

For an example of how to add this button to the dialog, see the implementation of the "Copy selected" button in DlgEventLog.cpp and DlgEventLog.h.

AenBleidd:
Ok. I this case I think it will be better to add two buttons 'Copy all' and 'Copy selected' to make it looks like Event Dialog.

JuhaSointusalo:
Tested on Windows 10 and Mint 17 (Cinnamon). Looks good, thanks. Ctrl-C worked on both as well as Ctrl-A. Although on Windows I needed to click on the WebView area to make Ctrl-A work. That's probably not a big issue.

It's interesting that Cmd-C doesn't work on Mac but does on Linux. Both are using WebKit so you'd think they would behave the same. Anyway, how about Ctrl-C handler? [Here's how wxHtmlWindows does it](https://github.com/wxWidgets/wxWidgets/blob/master/src/html/htmlwin.cpp#L1567).

Presumably you switched to wxWebView to get cells to not word wrap? That may be a small problem. At the moment Debian's wxWidgets package uses an old, unmaintained, more security holes than Swiss cheese version of WebKit which they want to get rid of. They could switch to a newer WebKit but then they need to switch to GTK3 backend for wxWidgets but they have been dragging their feet about it because then they could run into problems in GTK3 backend. So, don't be surprised if you need to do this a third time :/

davidpanderson:
Looks good on Win7

JuhaSointusalo:
???

Didn't you see Charlie's message saying Cmd-C doesn't work on Mac? Copying the text in the dialog was the reason why this and the previous changes were made in the first place and if it doesn't work then what is the point of merging this PR now?

AenBleidd:
>
> It's interesting that Cmd-C doesn't work on Mac but does on Linux.

Looks like known issue: https://forums.wxwidgets.org/viewtopic.php?t=43163
Unfortunately, still without answer. I want to make some investigations
through wxWidgets code to find an issue. The problem is that I have no Mac
to test and debug it.

They could switch to a newer WebKit but then they need to switch to GTK3
> backend for wxWidgets but they have been dragging their feet about it
> because then they could run into problems in GTK3 backend.

I hope this will fix some known problems. In ane case there should be the
moment in future when we will use new version of wxWidgets.

Anyway, how about Ctrl-C handler?

I can try to do this but again: I have no Mac to test it even during
development.

 So, don't be surprised if you need to do this a third time :/

I can do this as much as it need. I just want my work to be useful.

Didn't you see Charlie's message saying Cmd-C doesn't work on Mac? Copying
> the text in the dialog was the reason why this and the previous changes
> were made in the first place and if it doesn't work then what is the point
> of merging this PR now?


While this is not merged to release brunch it is not good but ok: in any
case I can try to handle this issue during this week.

Thanks

Best regards,
Vitalii Koshura

2017-08-06 23:28 GMT+03:00 David Anderson <notifications@github.com>:

CharlieFenton:
For what it's worth, I tried calling EnableContextMenu(true). This recognized a right click with the mouse (or control-click) to bring up the contextual menu, which included "Copy". Although this did work, it is not how users expect to perform a copy and so would be hidden functionality.

I also tried adding an accelerator for command-C (as in DlgSelectComputer.cpp line 179), but it still did not recognize the keystroke. 

A command-C handler might work, but you would probably also need a command-A handler (for select all.) 

I feel that a "Copy selected" button is easier and better. Since wxWebView::Copy() copies only the current selection and wxWebView does not offer a method to copy everything, a "Copy all" button would need to first call wxWebView::SelectAll(). I don't know if SelectAll() highlights all the text, which would look strange to the user. You could call wxWebView::ClearSelection() after; however that would undo any selection that the user had made, but perhaps that is acceptable.

AenBleidd:
@CharlieFenton, Thank you for this investigation. I wanted to do the same
before implementing buttons. Also there will be an issue with "Copy
selected" button:
there is no 'OnSelect' event in wxWebView component so for now I see no
possibility to enable/disable this button in case of user selection. I'll
try to play around this but I'm not sure that I'll success with this
without an event.

> however that would undo any selection that the user had made, but perhaps
> that is acceptable.

It is possible (I guess) to save user selection and the restore it. I'll
play with this functionality too.

Best regards,
Vitalii Koshura

2017-08-07 2:42 GMT+03:00 CharlieFenton <notifications@github.com>:

CharlieFenton:
> there is no 'OnSelect' event in wxWebView component so for now I see no
> possibility to enable/disable this button in case of user selection

You could use EVT_MOUSE_EVENTS(func) or EVT_LEFT_UP(func): and test whether there is a selection after a mouse up. And if you add a "Select all" button, that of course would also enable the button.

To test whether or not any text is selected in your OnMouseUp event handler, you could use either this:
virtual bool wxWebView::HasSelection() which eturns true if there is a current selection.
or this :
virtual wxString wxWebView::GetSelectedText() cons which returns the currently selected text, if any. I believe it would return an empty string if there is no selection. You can then test whether the returned wxString is empty using:
bool wxString::IsEmpty() const

> It is possible (I guess) to save user selection and the restore it. I'll
> play with this functionality too.

Actually, I was mistaken earlier; this is not necessary. The "Copy all" button can simply call the following without changing the selection:
virtual wxString GetPageText() const which returns the text of the entire page and copy / write that to the clipboard.

AenBleidd:
@CharlieFenton, thank you for the advice.

Best regards,
Vitalii Koshura

2017-08-07 10:52 GMT+03:00 CharlieFenton <notifications@github.com>:

JuhaSointusalo:
I think implementing those buttons would be more complicated than implementing key press handler. I think the attempt with an accelerator failed because wxID_COPY just isn't hooked up to anything in wxWebView. [common](https://github.com/wxWidgets/wxWidgets/blob/b99d28bb4657c2533df717a5ddb1cfc19cabe92a/src/common/webview.cpp) [osx](https://github.com/wxWidgets/wxWidgets/blob/4312de64c746f6614a1f3285370b901cb92d6e9b/src/osx/webview_webkit.mm)

I read somewhere, maybe WekKit blog, some time ago that different WebKit ports differ considerably from each other. Handling/not handling ctrl-c probably comes from there.

Even if you implement the buttons you'd still need to implement ctrl-c because that's what everyone expects. Since ctrl-c is already handled on Windows you could implement it as ctrl-b or some other unused combination and change it to ctrl-c just before submitting a PR.

On Windows, some dialogs or messageboxes lets one to copy their content by pressing ctrl-c without selecting text. Now that Charlie discovered GetPageText() that would be a nice addition.

AenBleidd:
@Juha, thank you for the additional information. I'll try this and let you
know.

Thanks

Best regards,
Vitalii Koshura

2017-08-07 18:00 GMT+03:00 Juha Sointusalo <notifications@github.com>:

CharlieFenton:
I implemented the Copy Selected button in my tests, and sent my rough "proof of concept code" to Vitali. It is actually quite easy to do. 

Command-C for copy is not implemented for the Event Log in the BOINC Manager. For some reason, pressing command-C in the Event Log closes that window, which is odd. Command-C also cancels / closes the Add Project, Use Account Manager and Other Options dialogs, but not for others. I have no idea why it does this, A quick look at the code doesn't any any reason for it, so it may be in wxWidgets.

So implementing command-C for the Properties dialog would be inconsistent with the rest of BOINC Manager on the Mac, unless we also implement it for other dialogs

Cheers,
--Charlie

CharlieFenton:
I tried control-C on BOINC Manager under Windows 7 (which is generally considered the equivalent of command-C on the Mac.) As with the Mac, it does not copy selected lines from the Event Log. I found that alt-C on Windows 7 cancels / closes the Event Log and the Add Project and Use Account Manager dialogs, the same as command-C does on the Mac. (But it does not cancel / close the Other Options dialog like command-C does on the Mac.)

To clarify, these keystrokes close the Event log window only if it is the active (foreground) window.

So I feel that adding Copy Selected and Copy All buttons to the Properties dialog is more consistent with how the Event log works than implementing a command-C / control-C handler. (note that, usually,  wxWidgets automatically replaces control-C implementations with command-C on builds for the Macintosh.

JuhaSointusalo:
The ones that can be closed with alt-c have c as mnemonic. FWIW, I couldn't quickly find any window in Windows or Visual Studio that had any mnemonic set for Cancel or Close buttons or even OK for that matter.

Anyway, if Mac needs the copy button so be it.

PR #2015, opened 2017-08-08
Drupal: Fixed two bugs in task table.

drshawnkwang:
1. BOINC state_num() function was changing the object. Using clone $result to pass a copy of the object instead.
2. Anonymous platforms were not correctly supported. Now two LEFT JOINs and a new pretty_application_version() function should handle anonymous platforms correctly.

https://dev.gridrepublic.org/browse/DBOINCP-387

### DIFF
--- a/drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
+++ b/drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
@@ -1849,33 +1849,27 @@ function delete_col(&$array, $offset) {
  * Parameters
  * @param int $appverid
  *  app_version_id field, may be negative- anonymous platforms.
- * @param int $versnum
+ * @param int $vernum
  *  version_num field, application version number, may be NULL
  * @param int $plan_class
  *  plan_class field, applcation plan class, may be NULL
- * @param string $platform
+ * @param string $plfm
  *  platform name, may be NULL
  */
 function pretty_application_version($appverid, $vernum, $plan_class, $plfm) {
   switch ($appverid) {
     case ANON_PLATFORM_UNKNOWN:
       return "Anonymous platform";
-      break;
     case ANON_PLATFORM_CPU:
       return "Anonymous platform CPU";
-      break;
     case ANON_PLATFORM_NVIDIA:
       return "Anonymous platform NVIDIA GPU";
-      break;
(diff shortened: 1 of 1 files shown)


drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
```diff
@@ -1829,9 +1831,43 @@ function delete_col(&$array, $offset) {
...
+    case ANON_PLATFORM_ATI:
+      return "Anonymous platform ATI GPU";
+      break;
+    case ANON_PLATFORM_INTEL:
+      return "Anonymous platform Intel GPU";
+      break;
+    case 0:
+      return "---";
+      break;
+    default:
```
tristanolive: Oliver, do you approve the changes as they are or should negative values be handled first? I'm assuming the latter, though you did mark this request as approved.

brevilo: What exactly do you mean by first? The current switch-case/if-else tree is unambiguous for all defined `ANON*` values, 0 and *other* negative values. If you mean by "first" that the `if` should precede the `switch`, then no, that would in fact be wrong because the `ANON*` values are negative as well and thus need to be checked first.

tl;dr: I approve the change as it is

tristanolive: Looking again, I see the concern in your original comment was addressed by 414fc98. I thought it was still an open question that you had and was waiting for a response (hence my confusion over your approval).

tristanolive:
Not to drag this PR out... I recently learned that Scrutinizer is not in an enforcing mode, so it is checking code but not ever giving a failing status. This particular PR, it claims, has 10 problems with it, mostly quite minor. I don't think we necessarily need to get that count to zero, but it may be worth reviewing the issues that were found. In this case, incorrect documentation should certainly be addressed. Unused "break" statements... not so much.

brevilo:
Good point! We should definitely make Scrutinizer a part of our review procedure. I mean what's the point of it if we don't use it.

That said, can its mode be changed for subtrees? IOW, can we enforce it for Drupal only?

drshawnkwang:
In the latest commit I've fixed some of the issues Scrutinizer found.

PR #2020, opened 2017-08-09
back end: add feature for assigning WUs to a particular version num

davidpanderson:
Allow projects to specify that WUs should be processed by particular version nums.
Add --app_version_num arg to create_work
Set app.min_version to lowest version with outstanding jobs
  (can used xadd to do this)


lfield:
Please could you let us know what tests were performed before the merge.

davidpanderson:
- create workunit with an app_version_num
- verify that the WU is dispatched using that version num
- verify that other WUs are dispatched using the latest version

PR #2025, opened 2017-08-11
Drupal: Changed how application list is obtained for project preferences.

drshawnkwang:
Application list is now obtained from the database. The project-specific preferences XML upload/configuration are overrides for the applications list.
Modified projectpref.xsd file, <apps> tag is now <app_overrides>.

https://dev.gridrepublic.org/browse/DBOINC-87


drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
```diff
@@ -203,6 +203,21 @@ function boincwork_get_project_specific_config() {
...
+  // upload is an override.
+  if (!array_key_exists('app_overrides', $xml['project_specific_preferences'])) {
+    $simpleapps = array(
+      'title' => array(
+        '@value' => 'Run only the selected applications',
+        '@position' => 1,
+      ),
+      '@position' => 0,
+    );
+    $xml['project_specific_preferences']['app_overrides'] = $simpleapps;
```
brevilo: This `$xml` is used for validation, I'm going to update the schema (XSD) and make `app_overrides` mandatory, ok?

drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
```diff
@@ -566,14 +615,18 @@ function boincwork_generate_prefs_element(&$form, $type, $elements, $user_prefs
...
         $checked = TRUE;
-        if (isset($app['@attributes']['selected']) AND
-            $app['@attributes']['selected'] == 'false') {
-          $checked = FALSE;
-        }
+        // If user_applications is not defined, it means there are no
+        // app_ids set in the user's project preferences payload
+        // XML. Thus all Web form values should be checked.
+
+        //if (isset($app['@attributes']['selected']) AND
```
brevilo: How about this snippet? There shouldn't be any disabled code in production.

drupal/sites/default/boinc/modules/boincwork/includes/projectprefs.xsd
```diff
@@ -134,10 +134,10 @@
 
     <!-- grouping types -->
 
-    <xs:element name="apps">
+    <xs:element name="app_overrides">
         <xs:complexType>
             <xs:sequence>
-                <xs:element name="title" type="xs:string" minOccurs="1" maxOccurs="1" />
+                <xs:element name="title" type="xs:string" minOccurs="0" maxOccurs="1" />
```
brevilo: Hm, we could think about removing the `title` entirely. Now that it gets a [fixed value](https://github.com/BOINC/boinc/pull/2025/files#diff-ea2a15c2242d752fb293111403e676afR214) anyway there's really a point to define it anymore, let alone override it. Also, this is a string that should match the upstream value to facilitate translation which is another reason to remove the override for good. Thoughts?

drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
drshawnkwang: Commented-out code removed.

drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
```diff
@@ -531,33 +555,58 @@ function boincwork_generate_prefs_element(&$form, $type, $elements, $user_prefs
...
+      $myapp_ids = $user_prefs['apps_selected']['app_id'];
+    }
+    else if (is_array($user_prefs['app_id'])) {
+      $myapp_ids = $user_prefs['app_id'];
+    }
+    // Fix the array in the case where there is exactly one
+    // app_id. myapp_ids should be an associative array where the
+    // values are themselves an array. In effect the app_id array
+    // needs to be pushed down onre more level in the tree.
+    if (is_array($myapp_ids) AND !is_array($myapp_ids[1])) {
```
tristanolive: $myapp_ids is not defined if neither of the above `if ... elseif`  conditions are met. This should be initialized to NULL at least to limit the possibility of bugs / unintended behavior popping up later.

drupal/sites/default/boinc/modules/boincwork/includes/boincwork.forms.inc
```diff
@@ -1248,10 +1247,12 @@ function boincwork_add_project_specific_prefs(&$form, $prefs) {
   ksort($ordered_array);
   $primed_array = array_merge($ordered_array, $unordered_array);
   $xml = array('project_specific_preferences' => $primed_array);
-    
   foreach ($xml['project_specific_preferences'] as $wrapped_element) {
     $type = key($wrapped_element);
-    $element= reset($wrapped_element);
+    $element = reset($wrapped_element);
+    if ($type=='app_overrides') {
```
tristanolive: Need spacing around "==", if style points count.

drupal/sites/default/boinc/modules/boincwork/includes/projectprefs.xsd
tristanolive: It sounds like there's no reason to keep it around. It will live on in our commit memory...

drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
drshawnkwang: This should be fixed with the latest commit 3fead8d.

drupal/sites/default/boinc/modules/boincwork/includes/boincwork.helpers.inc
```diff
@@ -203,6 +203,17 @@ function boincwork_get_project_specific_config() {
   
   // Convert XML to array for validation
   $xml = load_configuration($raw_config_data);
+  // By default the returned array should have an 'app_overrides'
+  // element. Even if the admin did not specificy an <apps> container
```
ChristianBeer: Did you mean `Even if the admin did not specificy an <app_overrides> container`?

drshawnkwang: Fixed. Thanks for catching this.

drshawnkwang:
> What I didn't get from the comments and the code is whether the resulting payload contains the `apps_selected` container or not.

The `<apps_selected>` container is saved to the payload XML in the BOINC database, `user` table, `project_prefs` field. The code that does this starts at line 356 of boincwork.helpers.inc. This block of code sets part of an array with a the key `apps_selected`. This array is then saved using `boincwork_save_prefs()`. That function in turn calls `save_configuraiton()` which uses `array_to_xml()` which converts the aforementioned array into the payload XML.

https://github.com/BOINC/boinc/pull/2025/files/3fead8d9a53f140ac3e7e60fcc22106fc6c1186d#diff-ea2a15c2242d752fb293111403e676afR356

PR #2027, opened 2017-08-12
Manager: use "client" instead of "daemon"

davidpanderson:
use uniform terminology

### DIFF
--- a/clientgui/BOINCGUIApp.cpp
+++ b/clientgui/BOINCGUIApp.cpp
@@ -599,7 +599,7 @@ void CBOINCGUIApp::OnInitCmdLine(wxCmdLineParser &parser) {
 #if (defined(__WXMAC__) && defined(_DEBUG))
     parser.AddLongOption("NSDocumentRevisionsDebugMode", _("Not used: workaround for bug in XCode 4.2"));
 #endif
-    parser.AddSwitch("nd", "no-daemon", _("Not run the daemon"));
+    parser.AddSwitch("nd", "no-daemon", _("Don't run the client"));
 }
 
 
--- a/clientgui/BOINCBaseFrame.cpp
+++ b/clientgui/BOINCBaseFrame.cpp
@@ -642,7 +642,7 @@ void CBOINCBaseFrame::ShowDaemonStartFailedAlert() {
     );
 #else
     strDialogMessage.Printf(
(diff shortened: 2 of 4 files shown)


AenBleidd:
doc/versions.inc file should not be in this pull request

CharlieFenton:
Commit d73d512 missed changing "daemon" to "client" in these places:

* on line 629 of BOINCBaseFrame.cpp:
    strDialogTitle.Printf(
        _("%s - Daemon Start Failed"),
        pSkinAdvanced->GetApplicationName().c_str()
    );

* On lines 462 and 464 of client_state.cpp:
#ifdef _WIN32
        msg_printf(NULL, MSG_INFO, "Running as a daemon (GPU computing disabled)");
#else
        msg_printf(NULL, MSG_INFO, "Running as a daemon");
#endif

clientgui/BOINCGUIApp.cpp
```diff
@@ -599,7 +599,7 @@ void CBOINCGUIApp::OnInitCmdLine(wxCmdLineParser &parser) {
 #if (defined(__WXMAC__) && defined(_DEBUG))
     parser.AddLongOption("NSDocumentRevisionsDebugMode", _("Not used: workaround for bug in XCode 4.2"));
 #endif
-    parser.AddSwitch("nd", "no-daemon", _("Not run the daemon"));
+    parser.AddSwitch("nd", "no-daemon", _("Don't run the client"));
```
JuhaSointusalo: You could change the name of the command line switch too.

Short options are typically one character after hyphen. wxWidgets' command line parser seems to handle `-nd` without confusing it with `-n` or `-d` but I'm not sure if that's by design or by accident. I'd remove the short option.

`--no-daemon` is with hyphen but the rest of Manager and client uses underscore. Please change that.

CharlieFenton:
Changing the command line option by eliminating or modifying either the short or the long form will break any scripts that users may have which invoke that option. It is fine to add a version with the underscore, but the old ones should remain for backward compatibility.

JuhaSointusalo:
We have been beta testing for quite a while now so it's a bit late, but
anyone running beta release should know that anything new in it can change
or even be removed. --no-daemon is not in 7.6.33 so IMHO it can be changed.

PR #2030, opened 2017-08-14
Manager: Properties Window fix

AenBleidd:
'Copy all' and 'Copy Selected' buttons added.

### DIFF
--- a/clientgui/DlgItemProperties.cpp
+++ b/clientgui/DlgItemProperties.cpp
@@ -30,6 +30,8 @@
 IMPLEMENT_DYNAMIC_CLASS(CDlgItemProperties, wxDialog)
 
 BEGIN_EVENT_TABLE(CDlgItemProperties, wxDialog)
+    EVT_BUTTON(ID_COPYSELECTED, CDlgItemProperties::OnCopySelected)
+    EVT_BUTTON(ID_COPYALL, CDlgItemProperties::OnCopyAll)
 
 END_EVENT_TABLE()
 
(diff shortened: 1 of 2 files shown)


AenBleidd:
I found no suitable way to track events in wxWebView. It has its own mouse event handler. So in this commit 'Copy Selected' button is enabled always. I don't like this behavior but unfortunately have no ideas how to fix this.  This solution is not perfect but is better than nothing.

CharlieFenton:
This solution is ugly and inefficient, but one possibility might be to set up a wxTimer to call the following periodically (perhaps twice per second) while the dialog is open:
m_pCopySelectedButton->Enable(m_txtInformation->HasSelection());

AenBleidd:
@CharlieFenton, thank you for the idea. I'll do this today.

Best regards,
Vitalii

Sent via Android

14 авг. 2017 г. 12:04 пользователь "CharlieFenton" <notifications@github.com>
написал:

This solution is ugly and inefficient, but one possibility might be to set
up a wxTimer to call the following periodically (perhaps twice per second)
while the dialog is open:
m_pCopySelectedButton->Enable(m_txtInformation->HasSelection());

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/BOINC/boinc/pull/2030#issuecomment-322136750>, or mute
the thread

JuhaSointusalo:
What kind of approaches did you try to handle mouse events? wxWebView inherits wxEvtHandler so setting a handler that is implemented in CDlgItemProperties with wxWebView::bind<>() is one option.

Or is the problem that the window system sends the events directly to the native control and wxWidgets never sees them?

AenBleidd:
@JuhaSointusalo 
> What kind of approaches did you try to handle mouse events? wxWebView inherits wxEvtHandler so setting a handler that is implemented in CDlgItemProperties with wxWebView::bind<>() is one option.

I tried this but in a little different way. I'll try this again today.
Thank you

clientgui/DlgItemProperties.cpp
```diff
@@ -85,6 +96,10 @@ CDlgItemProperties::CDlgItemProperties(wxWindow* parent) :
         break;
     }
 
+#ifdef __WXMAC__
+     m_txtInformation->SetEditable(true);
```
JuhaSointusalo: Is this right? Was it supposed to be EnableContextMenu(true)?

AenBleidd: I'm not sure about this code. I have no possibility to test this on Mac.
Actually I got this piece of code from @CharlieFenton.
@CharlieFenton could you please comment whether this line is necessary or not? 
Thanks

JuhaSointusalo:
Oh:
```
    m_txtInformation = wxWebView::New( this, wxID_ANY, wxWebViewDefaultURLStr, wxDefaultPosition, wxDefaultSize, wxWebViewBackendDefault, style );
    m_txtInformation->EnableContextMenu( false );
```
I thought context menu didn't work on Mac automatically because Macs just want to be different but you actually disable context menu?

AenBleidd:
Yes, I disabled manually context menu because it contains many unnecessary
functions related to html page.

Best regards,
Vitalii

Sent via Android

14 авг. 2017 г. 19:22 пользователь "Juha Sointusalo" <
notifications@github.com> написал:

JuhaSointusalo:
>
> Yes, I disabled manually context menu because it contains many unnecessary
> functions related to html page.
>

You are right, it wouldn't work here.

CharlieFenton:
SetEditable(true) is not needed. I have removed it in commit 07e55b9. Thank you for bringing this to my attention.

AenBleidd:
@JuhaSointusalo I tried different bindings. All mouse and keyboard events seem to be handled by native control (I check on Windows only with IE as native control). Due to the implementation of the wxWebView component no events are catched from native control. I'll try to play a little bit more with this.

JuhaSointusalo:
I tried it too earlier but like you didn't get it to work. It's puzzling
because when I look at the Manager with Spy++ I see a nice hierarchy of
controls and when I look at wxWidgets source it looks like this should be
supported.

I'll try it again after I have VS2017 solution ready. It'll be easier when
I have wxWidgets source available to VS.

PR #2036, opened 2017-08-14
Client, Win: append .exe to service process program name if missing

davidpanderson:


AenBleidd:
@davidpanderson Maybe it's better and safer to use std::string instead of char array?

davidpanderson:
Yes, but in practice 1023 is plenty.

When I write new code I use std::string, but I'm not sure it's worth converting
all the char arrays, as long as we use size-aware functions like strlcat().

AenBleidd:
OK

PR #2038, opened 2017-08-15
client: fix report_result_error buffer overflow

AndCycle:
this is just a quick fix, I need someone with insight to review is this a proper fix,

tl;dr
CLIENT_STATE::report_result_error made an assumption that err_msg  won't get too large,
unfortunately, it will get really large due to this line

https://github.com/BOINC/boinc/blob/08c5b3b276b079c1433ed8f1e569314d9df88168/client/client_state.cpp#L1570

full detail and related log
http://boinc.berkeley.edu/dev/forum_thread.php?id=11772


AenBleidd:
@AndCycle  Could you please make pull request without merge commit from master?

AndCycle:
@AenBleidd my fault, sorry

davidpanderson:
I'll fix this in a way that doesn't use fixed-size buffers.

PR #2042, opened 2017-08-15
client: eliminate possible buffer overflow in reporting result errors

davidpanderson:
A result with a lot of failed uploads could overflow a 4K buffer.
Change report_result_error() so you just pass it the error message,
rather than va_args nonsense.


AenBleidd:
@davidpanderson Maybe it will be better to use 'snprintf' with buffer overflow check instead of 'sprintf' without such it?

PR #2047, opened 2017-08-18
client: more startup messages, hopefully find source of delay

davidpanderson:


CharlieFenton:
Since the purpose of these extra messages is to debug a specific problem, it might be better to display them only if an appropriate log flag is set (though I don't know which flag that would be), or at least compile them conditionally so they do not remain in the client forever.

JuhaSointusalo:
When you have a problem and need a shotgun approach to debug it, make a branch to contain the debug code. Make builds from the branch and hand them to testers. Add more commits there while you try to nail down the problem.

Once you know what the problem is and how to fix it make a **new** branch and add **only** the changes that is needed to fix the problem to it. Then make a pull request from the new branch.

Debug code like this has no place in master. BOINC already has enough code in it that looks like it's junk left over from some debugging session years ago.

davidpanderson:
This is not one-time debugging code.  We'll always want to see the startup tasks.

RichardHaselgrove:
Community, David, community.

Remember that you asked the community for help.

RichardHaselgrove:
Who exactly is going to test this debug code?

If you're going to test it yourself, make a private build from local sources - doesn't even need a branch.

If you're going to ask the community to test, we need at least a v7.8.2 tag so our private builds match, and a v7.8.2 alpha build so community members without build capability can also test.

davidpanderson:
??  It will get tested in 7.8 if we decide to back-port it there.
Otherwise it will get tested in 7.10.
That's how we do testing of client changes.

RichardHaselgrove:
Further research reported on the boinc_dev mailing list (though rapidly getting swamped by unrelated messages) suggests that commit 37b64313fed14b97e1a400ea9c606dbeac50b08f (or its parent) might be implicated.

JuhaSointusalo:
The code in master is:

    if (cc_config.lower_client_priority) {
        set_client_priority();
    }

and set_client_priority is:

#ifdef _WIN32
    STP stp = (STP) GetProcAddress(GetModuleHandle(_T("kernel32.dll")),
"SetThreadPriority");
    if (!stp) return;
    if (stp(GetCurrentThread(), THREAD_MODE_BACKGROUND_BEGIN)) {
        msg_printf(NULL, MSG_INFO, "Running at background priority");
    } else {
        msg_printf(NULL, MSG_INFO, "Failed to set background priority");
    }
#endif

The default for cc_config.lower_client_priority is false and there is no
message about setting priority so that code is not run.

RichardHaselgrove:
The delay seemed to occur in check_file_existence, so it's the changes in cs_files.cpp we need to look at, not the changes in client_state.cpp

JuhaSointusalo:
You are building form 7.8 branch right? Could you try master. file_size was
broken and requested too little sharing. Although then you should be seeing
messages about files not found.

RichardHaselgrove:
I've got a clone based on master, so for the moment I've been testing on an export at the 7.8.1 tag level, rather than a branch (I'm trying to replicate the problems seen on the public 7.8.1 alpha build).

I can certainly replicate the process directly on Master (probably in the morning now, it's getting late here), but I was thinking that David's plan was to get v7.8 prepared for full release, before we start testing for v7.10. But if we report carefully which code-base we're testing at each stage, we should be able to avoid getting too confused.

JuhaSointusalo:
I suggested building from master just see if the problem is already fixed.

RichardHaselgrove:
OK, I'll check that for sure. But dinner has just come out of the oven... 

    On Saturday, 19 August 2017, 22:11, Juha Sointusalo <notifications@github.com> wrote:
 

 I suggested building from master just see if the problem is already fixed.

Ageless93:
I've just built a debug BOINC from master, but cannot test the delay until tomorrow. For me, it only happens after coming out of hibernation, after midday.

RichardHaselgrove:
OK, I have two builds to compare: one with a fresh pull from master (hence including the files in this PR), and one from the v7.8.1 tag with extra msg_printf below the ones that David added.

With v7.8.1, I get

19/08/2017 23:12:52 |  | Checking presence of project files
19/08/2017 23:13:15 |  | Finished checking project files

from this code

    msg_printf(NULL, MSG_INFO, "Checking presence of project files");
	check_file_existence();
    msg_printf(NULL, MSG_INFO, "Finished checking project files");

which clearly localises the problem.

With the build from master, I get

19-Aug-2017 23:05:40 [---] Checking presence of project files
19-Aug-2017 23:05:40 [---] Finished checking project files

so the problem has gone away.

The question becomes, what do we need to backport from master to the v7.8.(x) tag to get the non-delay behaviour, without introducing other unwanted changes?

check_file_existence() lives in cs_files.cpp, and according to github that file hasn't been touched for two years. Maybe it's in a dependency?

Anyway, that's it for tonight - I'll watch for emails, but no more building.

Ageless93:
I can confirm that the client built from master does not have the long delay. I added a lot of debug flags and start-up time is just 2 seconds. 

`log flags: file_xfer, sched_ops, task, benchmark_debug, checkpoint_debug,  cpu_sched, file_xfer_debug, gui_rpc_debug, notice_debug, poll_debug, proxy_debug, sched_op_debug, slot_debug, state_debug, statefile_debug, task_debug, unparsed_xml`
[start-log-master7.8.7z](https://www.dropbox.com/s/b76ca33ybb3gtax/start-log-master7.8.7z?dl=0)

RichardHaselgrove:
Similar result here. I've made three builds this morning: one from master, one from the 7.8 branch, and one from the 7.8.1 tag. I've added all of David's debug prints, plus one of my own: the client reports every file that it checks, whether or not an error is found.

You can see that I have 918 project files on this machine. Master checks all of them in under a second (as I would expect on a fast machine like this): branch and tag plod through them at maybe 50 per second.

[master_startup_log.zip](https://github.com/BOINC/boinc/files/1237037/master_startup_log.zip)
[branch_78_startup_log.zip](https://github.com/BOINC/boinc/files/1237038/branch_78_startup_log.zip)
[tag_781_startup_log.zip](https://github.com/BOINC/boinc/files/1237039/tag_781_startup_log.zip)

RichardHaselgrove:
I've found and fixed the cause of the startup delay in v7.8 - writing report to mailing lists next.

[fixed_tag_781_startup_log.zip](https://github.com/BOINC/boinc/files/1238117/fixed_tag_781_startup_log.zip)

PR #2057, opened 2017-08-21
Fixed displaying the menu "Computing"

progger:
NO_COMPUTING defined always after commit https://github.com/BOINC/boinc/commit/e9244a784420e4305ac1f74a842b7d1305bd41da


davidpanderson:
Thanks!

PR #2058, opened 2017-08-22
Fixed touch md5 file path

progger:
Wrong path for touch in https://github.com/BOINC/boinc/commit/b38bf1d9838edad00fc376b2730acba8f4617a30

### DIFF
--- a/tools/stage_file
+++ b/tools/stage_file
@@ -66,7 +66,6 @@ function stage_file($path) {
     if (!file_exists($path)) error_exit("no such file: $path\n");
     $file = basename($path);
     $dl_path = dir_hier_path($file, $download_dir, $fanout);
-    $dl_md5_path = "$dl_path.md5";
 
     if ($verbose) {
         echo "staging $file to $dl_path\n";
--- a/tools/stage_file
+++ b/tools/stage_file
@@ -90,7 +90,7 @@ function stage_file($path) {
             if (!$ret) error_exit("    copy failed\n");
             if ($verbose) echo "    moved file\n";
         }
-        touch("$path.md5");
+        touch("$dl_path.md5");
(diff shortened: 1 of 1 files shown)


lfield:
I have two comments on this. The first is that it should be $dl_md5_path otherwise remove the variable as it is not used. 
https://github.com/BOINC/boinc/blob/master/tools/stage_file#L72
The second is why is this file touched in the first place? It has no content. What is the purpose?

davidpanderson:
The touch() is needed in the (unlikely) case where there's a correct .md5 file in the dir hierarchy,
but no data file.
This fix is correct and the $dl_md5_path can be removed (or use it as the arg to touch())

The code needs comments to explain what it's doing, and to point out that the logic in process_input_template.cpp expects the .md5 file to have a later mod time than the data file.
I'll do this later.

TheAspens:
The [code path if the file doesn't exist](https://github.com/BOINC/boinc/blob/aae742d8893a614e7688dfbdfa43305f5741fc9f/html/inc/dir_hier.inc#L105) will overwrite an existing md5 file if it exists or create a new one if it doesn't.

TheAspens:
Just re-read this bit:

> the logic in process_input_template.cpp expects the .md5 file to have a later mod time than the data file.

If the .md5 file needs a later timestamp than the file itself, then I agree that the touch is required otherwise the .md5 file is a few milliseconds younger than the data file.

davidpanderson:
Another problem case is when both files exist, and the .md5 is older.
The current code - with or without this fix - doesn't handle this correctly.
Not an error, just a redundant MD5 calculation

lfield:
ok, I have removed the unused variable. Any objection to merging?

PR #2059, opened 2017-08-22
web: fix beta app version handling in apps.php page

bema-ligo:
In apps.php treat beta app versions independently of non-beta app versions, even if these have the same plan class. This is consistent with what the (array) scheduler does (try 'beta' apps first (with highest version number per plan class), then 'non-beta' (dito)).


lfield:
This looks good to me.

PR #2060, opened 2017-08-22
improved file_upload_handler

ChristianBeer:
This is the file_upload_handler code which is running on Einstein@Home for the last year.
- fix `get_file_size()` to be read-only, it does not need to check if the existing file is writable
- fix `copy_socket_to_file()` to make completely uploaded files read-only and add a check if a subsequent upload happens (data will be discarded)
- reopen logfile on user defined signal so log rotation works properly when running as FCGI
- optional variety postfix for upload_dir and logfile when running as FCGI (we use this to split file uploads of small (O(100kB)) and large (O(3MB)) files to be stored in separate file systems for better speed and efficiency

### DIFF
--- a/sched/file_upload_handler.cpp
+++ b/sched/file_upload_handler.cpp
@@ -60,6 +60,7 @@ using std::string;
 #define ERR_PERMANENT   false
 
 char this_filename[256];
+string variety = "";
 double start_time();
 
 inline static const char* get_remote_addr() {
@@ -639,14 +640,15 @@ void boinc_catch_signal(int signal_num) {
 }
 
 void boinc_reopen_logfile(int signal_num) {
+    char log_name[MAXPATHLEN];
     char log_path[MAXPATHLEN];
 
     // only handle SIGUSR1 here
     if (signal_num != SIGUSR2) {
         boinc_catch_signal(signal_num);
     }
-
-    if (get_log_path(log_path, "file_upload_handler.log") == ERR_MKDIR) {
+    sprintf(log_name, "file_upload_handler%s.log", variety.c_str());
+    if (get_log_path(log_path, log_name) == ERR_MKDIR) {
         fprintf(stderr, "Can't create log directory '%s'  (errno: %d)\n", log_path, errno);
     }
 #ifndef _USING_FCGI_
@@ -701,7 +703,8 @@ void usage(char *name) {
         "Usage: %s [OPTION]...\n\n"
         "Options:\n"
         "  [ -h | --help ]        Show this help text.\n"
-        "  [ -v | --version ]     Show version information.\n",
+        "  [ -v | --version ]     Show version information.\n"
+        "  [ -u V | --variety V]  Use V to construct logfile name and upload_dir from config.xml (FCGI only)\n",
         name
     );
 }
@@ -721,6 +724,10 @@ int main(int argc, char *argv[]) {
         } else if(option == "-h" || option == "--help") {
             usage(argv[0]);
             exit(0);
+#ifdef _USING_FCGI_
+        } else if(option == "-u" || option == "--variety") {
+            variety = "_" + string(argv[++c]);
+#endif
         } else if (option.length()){
             fprintf(stderr, "unknown command line argument: %s\n\n", argv[c]);
             usage(argv[0]);
@@ -739,6 +746,12 @@ int main(int argc, char *argv[]) {
             "can't parse config file"
         );
         exit(1);
+    }
+
+    // check if --variety was specified and add it to config.upload_dir
+    if (!variety.empty()) {
+        log_messages.printf(MSG_NORMAL, "Using variety: %s\n", variety.c_str());
+        strcat(config.upload_dir, variety.c_str());
     }
 
     log_messages.pid = getpid();
@@ -768,6 +781,7 @@ int main(int argc, char *argv[]) {
     }
 
 #ifdef _USING_FCGI_
+    log_messages.flush();
     while(FCGI_Accept() >= 0) {
         counter++;
         //fprintf(stderr, "file_upload_handler (FCGI): counter: %d\n", counter);
--- a/sched/file_upload_handler.cpp
+++ b/sched/file_upload_handler.cpp
@@ -114,7 +114,7 @@ double bytes_left=-1;
 int accept_empty_file(char* name, char* path) {
     int fd = open(path,
         O_WRONLY|O_CREAT,
-        S_IRUSR|S_IWUSR|S_IRGRP|S_IWGRP|S_IROTH
+        config.fuh_set_initial_permission
     );
     if (fd<0) {
         return return_error(ERR_TRANSIENT,
@@ -169,7 +169,7 @@ int copy_socket_to_file(FILE* in, char* name, char* path, double offset, double 
             // coverity[toctou]
             fd = open(path,
                 O_WRONLY|O_CREAT,
-                S_IRUSR|S_IWUSR|S_IRGRP|S_IWGRP|S_IROTH
+                config.fuh_set_initial_permission
             );
             if (fd<0) {
                 if (errno == EACCES) {
@@ -290,10 +290,15 @@ int copy_socket_to_file(FILE* in, char* name, char* path, double offset, double 
 
         bytes_left -= n;
     }
-    // upload complete; make the file read-only so it does not get overwriten by a subsequent upload.
+    // upload complete; set new file permissions if configured
     //
-    if (fchmod(fd, S_IRUSR|S_IRGRP|S_IROTH)) {
-        log_messages.printf(MSG_CRITICAL, "can't make file %s read only: %s\n", path, strerror(errno));
+    if (config.fuh_set_completed_permission >= 0) {
+        if (fchmod(fd, config.fuh_set_completed_permission)) {
+            log_messages.printf(MSG_CRITICAL, "can't set %03o permissions on %s: %s\n",
+                config.fuh_set_completed_permission,
+                path,
+                strerror(errno));
+        }
     }
     close(fd);
     return return_success(0);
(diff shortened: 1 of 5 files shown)


TheAspens:
Would it be possible to change this code:

`if (fchmod(fd, S_IRUSR|S_IRGRP|S_IROTH)) {
    log_messages.printf(MSG_CRITICAL, "can't make file %s read only: %s\n", path, strerror(errno));
}`

so that we don't assume everyone wants 444 for the permissions.  Could this be modified so that this explicitly setting of the permissions to 444 instead it would only set it if a value in the server configuration and the value set would be used?  (for example it could 440 in the config and and that would set the file when done uploading to user read, group read, world no access).

I'm continuing to look at the code.

sched/file_upload_handler.cpp
```diff
@@ -289,16 +307,17 @@ int handle_file_upload(FILE* in, R_RSA_PUBLIC_KEY& key) {
...
     strcpy(xml_signature, "");
     bool found_data = false;
     while (fgets(buf, 256, in)) {
-#if 1
-        log_messages.printf(MSG_NORMAL, "got:%s\n", buf);
-#endif
+        // intentionally set higher than debug as it may produce lots of output
+        if (config.fuh_debug_level > MSG_DEBUG) {
+            log_messages.printf(MSG_DEBUG, "got:%s\n", buf);
+        }
```
TheAspens: Doesn't this ensure that this line never gets output?  Shouldn't this be >= MSG_DEBUG?

TheAspens:
@davidpanderson - can you take a look also?

sched/file_upload_handler.cpp
ChristianBeer: No, the debug_levels in BOINC are defined as `enum { MSG_CRITICAL=1, MSG_WARNING, MSG_NORMAL, MSG_DEBUG, MSG_DETAIL };` which means the higher the level the more detailed output you get. In this case I should probably use `>= MSG_DETAIL` above (this level was added 3 moths ago and my code is older than that).

ChristianBeer: I edited my answer above since Kevin added more info to his question and my original answer looked out of context.

ChristianBeer:
> Hmmm - just saw that the permissions of the file when originally created are also set. If you make the final permissions configurable, might be worth making the initial permissions configurable as well.

I see your point but I don't want to make this a runtime option as I would need to read and parse `config.xml` for every request. I would rather make it a compile time option in the head of the file so it is easy to change. The purpose of file_upload_handler is to quickly handle the file upload. As the process itself is IO heavy I would not like to add another IO bottleneck to it.

Maybe @SETIguy can also comment as they had issues with this mechanism the last time I introduced it.

sched/file_upload_handler.cpp
TheAspens: @ChristianBeer - sorry for the editing - thought I finished before you saw it.

TheAspens:
@ChristianBeer Doesn't it already parse the config.xml file because it has to determine the location of upload and download directories?  Also - in the past we have modified files in the upload directory during validation for various reasons.  We aren't doing that at this time, but this change would impact any project that does something like that.  I would be concerned about making the file read only as a change in the behavior because this could impact any project treats the upload files as something other than read only which is why my preference is to preserve existing behavior and then use an option to change the behavior as needed.

ChristianBeer:
There are two versions of the file_upload_handler. One is the CGI version which is started for every request and indeed has to parse `config.xml` evrey time it starts. But there is also the FCGI version which parses `config.xml` only once before entering the endless FCGI loop. So if a user of the latter version is aware that changes in config.xml are not effective instantaneously but need a restart it's possible to set this mode in `config.xml`.

The question now is as always how to name this option. Since you want to be able to set the permissions like this: `<fuh_set_completed_permission>440</fuh_set_completed_permission>` I would also need to translate this into flags that are understood by `fchmod()` but would only support the three digit format without the execute bit (so possible values are: `2`, `4`, `6`). The default would be `644` I guess.

TheAspens:
I think that sounds correct and would support that implementation.  Thanks for making that change.

ChristianBeer:
Ok so we are in agreement, in case I forget here is what I need to change:

- on line 319 use `log_messages.printf(MSG_DETAIL, "got:%s\n", buf);` and remove the surrounding `if`
- on line 295 make the options configurable through `config.xml` via `<fuh_set_completed_permission>`
- on lines 150 and 170 also make the options configurable via `<fuh_set_initial_permission>`

TheAspens:
Yes - I believe so.  I think the only additional item is that the default behavior (if neither <fuh_set_completed_permission> or <fuh_set_initial_permission> are set) should behave in the current way.  Specifically, it will create the files with 664 and not change those when the file is done uploading.

bema-ligo:
I think it would be easier to just configure <fuh_set_initial_permission> and derive the "completed permission" from these by removing the "write" bits (i.e. `completed permission = initial_permission & !(S_IWUSR | S_IWGRP | S_IWOTH)`)

ChristianBeer:
We would still need two settings since Kevin said they had a use case where they wanted the completed file to still be writable. They could choose to make it only group writable which would prevent a file from being overwritten by file_upload_handler but can still be changed by other daemons running under a different useraccount. So setting the permissions explicitly (initial and completed) gives the project more choice and is more generic.

ChristianBeer:
Some notes about testing file upload handler in general (brain dump, not fully tested):
The cgi version can be tested by setting up a simple `config.xml` and request file and then sending it via `stdin` like this:
```
$ cat config.xml
<?xml version="1.0" ?>
<boinc>
  <config>
    <host>boinc-server-test</host>
    <project_dir>../</project_dir>
    <key_dir>../keys</key_dir>
    <upload_dir>../uploads</upload_dir>
    <ignore_upload_certificates/>
    <uldl_dir_fanout>0</uldl_dir_fanout>
    <fuh_debug_level>4</fuh_debug_level>
  </config>
  <daemons>
  </daemons>
  <tasks>
  </tasks>
</boinc>
$ cat cgi-bin/send-A # Note: there is no linebreak at the end of this file
<core_client_major_version>10</core_client_major_version>
<core_client_minor_version>10</core_client_minor_version>
<core_client_release>10</core_client_release>
<file_upload>
<file_info>
<name>fileA</name>
<max_nbytes>2.000000</max_nbytes>
</file_info>
<offset>0</offset>
<nbytes>1</nbytes>
<md5_cksum>7fc56270e7a70fa81a5935b72eacbe29</md5_cksum>
<data>
A
$ cd cgi-bin
$ ./fuh_cgi < send-A
Content-type: text/plain

<data_server_reply>
    <status>0</status>
</data_server_reply>
```
The fcgi version needs to be run using `spawn_fcgi` and `multiwatch` like this:
```
/usr/bin/spawn-fcgi -C 0 -n -p "8000" -a 127.0.0.1 -d "${PROJECT_DIR}/cgi-bin" -- /usr/bin/multiwatch -f "1" -- "${PROJECT_DIR}/cgi-bin/fcgi_fuh" &
```
An fcgi proxy like nginx is needed to connect to port 8000 with a simple config like this:
```
location file_upload_handler {
    include /etc/nginx/fastcgi_params;
    fastcgi_pass 127.0.0.1:8000;
}
```
Then you can use the same files as above and send them using `curl` like this:
```
curl -vv --data-binary @./send-A http://127.0.0.1/file_upload_handler
```

This is how a `get_file_size` request looks like:
```
$ cat get-size-A # Note: there is no linebreak at the end of this file
<data_server_request>
<core_client_major_version>10</core_client_major_version>
<core_client_minor_version>10</core_client_minor_version>
<core_client_release>10</core_client_release>
<get_file_size>fileA</get_file_size>
</data_server_request>
```

ChristianBeer:
Since this is going to be tested by @TheAspens experimental test framework for the server I didn't do extensive tests on my own. Kevin feel free to remove the [WIP] after you are done testing.

TheAspens:
I will test this within the next 24 hours.  Sorry for the delay.

TheAspens:
I'm testing this and it appears to work correctly for for config.fuh_set_completed_permission.  However, if I leave config.fuh_set_completed_permission unset, and test config.fuh_set_initial_permission it isn't working as desired.  Specifically if I set:

<fuh_set_initial_permission>666</fuh_set_initial_permission>

The file upload is set to 

root@boincserver:~/project# ls -lh upload/33a/test.txt 
-rw-rw-r-- 1 www-data root 45 Sep 22 14:44 upload/33a/test.txt

Hmmm.  I just realized that it is probably interpreting the number as decimal instead of the three digits that linux users are used to.  Would it be possible to assume that it is an octal number and convert it to base 10?  That would let people enter the normal permission digits that they are used to.

bema-ligo:
> Hmmm. I just realized that it is probably interpreting the number as decimal instead of the three digits that linux users are used to. Would it be possible to assume that it is an octal number and convert it to base 10? That would let people enter the normal permission digits that they are used to.

Precede the values with a '0' should help. atoi() should then read these as octals.

ChristianBeer:
The permission string is parsed using `strtol(buf, NULL, 8)` where I hardcoded the radix to 8 in order to circumvent the problem you describe. I guess that you have a modified umask because that is still applied according to `man 2 open`:
> The effective permissions are modified by the process's umask in the usual way: The permissions of the created file are (mode & ~umask).

lfield:
@ChristianBeer do you want to proceed with this PR?

ChristianBeer:
This needs independent verification. @TheAspens started but another look wouldn't hurt.

ChristianBeer:
Closing and reopening to trigger CI again.

lfield:
I put this on LHC@home dev and it seems to work. If it is already running in production for Einstein@Home then this should be good to merge. Any objections?

bema-aei:
@lfield I can confirm that this has been running on E@H for more than two years now.

PR #2065, opened 2017-08-23
Release patches 7.8.2

RichardHaselgrove:
This is intended to create a patch review list for the current release branch.

Please be kind - this is my first attempt at using git to cherry pick commits from master and turn them in to a pull request.

All these commits are already in master, but I believe should be ready for the next alpha client release test.

This list does NOT include commits targeted on Mac, Linux, or release 7.10

I failed to resolve problems with two commits I would otherwise have included:

d73d5120eb79ed8a36365876d31e5bcf26a61c63 Manager: use "client" instead of "daemon"
3a96e95d02c97c20779bcb783cd3a5177fe8d96d GUI RPC client: use std::string instead of fixed-size buffer for requests

### DIFF
--- a/client/app_control.cpp
+++ b/client/app_control.cpp
@@ -390,12 +390,12 @@ void ACTIVE_TASK::handle_temporary_exit(
     } else {
         if (is_notice) {
             msg_printf(result->project, MSG_USER_ALERT,
-                "Task postponed: %s", reason
+                "Task %s postponed for %.f seconds: %s", result->name, backoff, reason
             );
         } else {
             if (log_flags.task) {
                 msg_printf(result->project, MSG_INFO,
-                    "task postponed %f sec: %s", backoff, reason
+                    "Task %s postponed for %.f seconds: %s", result->name, backoff, reason
                 );
             }
         }
--- a/client/app_control.cpp
+++ b/client/app_control.cpp
@@ -439,6 +439,7 @@ void ACTIVE_TASK::handle_exited_app(int stat) {
         );
     }
 #endif
+    char err_msg[4096];
     bool will_restart = false;
 
     get_app_status_msg();
@@ -491,12 +492,12 @@ void ACTIVE_TASK::handle_exited_app(int stat) {
         default:
             char szError[1024];
             set_task_state(PROCESS_EXITED, "handle_exited_app");
-            gstate.report_result_error(
-                *result,
+            sprintf(err_msg,
                 "%s - exit code %d (0x%x)",
                 windows_format_error_string(exit_code, szError, sizeof(szError)),
                 exit_code, exit_code
             );
+            gstate.report_result_error(*result, err_msg);
             if (log_flags.task_debug) {
                 msg_printf(result->project, MSG_INFO,
                     "[task] Process for %s exited",
@@ -528,12 +529,12 @@ void ACTIVE_TASK::handle_exited_app(int stat) {
                 }
                 if (result->exit_status) {
                     set_task_state(PROCESS_EXITED, "handle_exited_app");
-                    gstate.report_result_error(
-                        *result,
+                    sprintf(err_msg,
                         "process exited with code %d (0x%x, %d)",
                         result->exit_status, result->exit_status,
                         (-1<<8)|result->exit_status
                     );
+                    gstate.report_result_error(*result, err_msg);
                 } else {
                     if (finish_file_present()) {
                         set_task_state(PROCESS_EXITED, "handle_exited_app");
@@ -567,9 +568,8 @@ void ACTIVE_TASK::handle_exited_app(int stat) {
                 result->exit_status = stat;
                 set_task_state(PROCESS_WAS_SIGNALED, "handle_exited_app");
                 signal = got_signal;
-                gstate.report_result_error(
-                    *result, "process got signal %d", signal
-                );
+                sprintf(err_msg, "process got signal %d", signal);
+                gstate.report_result_error(*result, err_msg);
             }
         } else {
             result->exit_status = EXIT_UNKNOWN;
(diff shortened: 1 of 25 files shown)


JuhaSointusalo:
I did shoddy job testing "client/lib: don't flush stdout and stderr in main loop" and unfortunately it is buggy. If you want it included in 7.8 I suppose I could prioritise fixing it.

RichardHaselgrove:
Hi @JuhaSointusalo - Don't worry about that, this is very much a preliminary trawl through potential candidate commits for the release: the powers-that-be will review all commits before releasing the next client. Thanks for the heads-up on the readiness status.

BOINC has been described as moving towards a community-led development model, and I was using this as a teach yourself model to see just how the git and github tools interacted (I learned a lot in the process). I've been invited to sit on a working group to discuss future development models and processes for BOINC: while we're not finished yet, and have no firm proposals to report, we seem to be moving towards [gitflow](http://www.nikola-breznjak.com/blog/miscellaneou/git-branching-done-right-gitflow-improving-code-quality-code-reviews/) principles. It might be helpful to adjust (as I'm having to) to the 'commit to a development branch, seek promotion to master via a pull request and peer review' model.

But perhaps like all neophytes, I'm being over-enthusiastic in proselytising the new model.

CharlieFenton:
@ChristianBeer, can you please comment on the following?

We need to be very careful about including any changes to localizable (translatable) strings in the client and Manager. Every such change requires updating the translation templates, and after that it can take significant time for the translators for each language to update their translations.

At the very least, any strings not in the translation files included in the public release for a language will be shown in English for that language. 

It may be more serious than that. As I recall, at least in the past, if there was even a single string that had not been translated in a particular language, none of the new translations for that language would be included. In other words, a translation file would not be updated in BOINC until it was 100% complete. I don't know if this undesirable behavior of Transifex was ever corrected.

In either case, I recommend that we should not change any translatable strings in the release branch for at least a month before a public release.

Ageless93:
I suggest we leave 79b1266, 97a3f75 and d73d512 out and move those to 7.10

RichardHaselgrove:
Because of the translation problem?

I agree about the OK / Save button - we've lived with that for years, and can change it any time. But the Daemon d73d5120eb79ed8a36365876d31e5bcf26a61c63 is new this time: wouldn't it be better to show it properly from the beginning?

Ageless93:
d73d512 also needs to be translated. It's also not 'daemon' anymore, but 'client'.

ChristianBeer:
The translation system as it is setup on transifex can only handle one set of templates. Currently it uses the templates generated from `client_release/7/7.8` and PR #2055 contains the translations that correspond to those templates which also include unfinished translations (>=98%) to circumvent the problem Charlie described because about a dozen languages are missing translations for the update version check strings and other minor strings but contain a lot of updated translations since the last translation import.

Ideally we should always update translations shortly before building the next release candidate so they are most current. I'm available to do that given that I get the information when a new version will be created.

RichardHaselgrove:
OK, I'll leave you guys to make a final decision in conjunction with the release manager. I think I've done my bit by attracting extra pairs of eyes to consider questions like this.

CharlieFenton:
This is my understanding of what @ChristianBeer meant by:
> Ideally we should always update translations shortly before building the next release candidate so they are most current. I'm available to do that given that I get the information when a new version will be created.

I think he is saying that he is waiting to be told the release schedule before updating the translations, so he will know when it is "shortly before building the next release candidate." Since @davidpanderson is the current release manager, I believe David is the logical person to make this decision and to provide this scheduling information.

ChristianBeer:
To correct Richards mentioning of gitflow above: The working group currently writing up a proposal for a community centered development workflow is favoring [GitHub Flow](https://guides.github.com/introduction/flow/) which can be seen as a simpler version of gitflow.

davidpanderson:
I suggest that the templates in transifex always correspond to master, not a release branch.
Changes to release branches shouldn't involve strings.
The translations (.po files) are copied from master when the branch is made, and don't change.

ChristianBeer:
> I suggest that the templates in transifex always correspond to master, not a release branch.
Changes to release branches shouldn't involve strings.
The translations (.po files) are copied from master when the branch is made, and don't change.

I agree that this sounds like the ideal process. But that means that translations need to be up to date before the release branch is created. This means that there needs to be a freeze phase where no string changes happen in master while the volunteers translate the templates generated from master.  Once the translations are done they can be imported to master, the release branch can be created and string changes can happen in master again.

In this scenario the release will have the translations as they were present at the time of creating the release branch. If a beta tester finds out that something is translated wrong we have no easy way of updating translations because master may already have diverged from the release branch. I guess one can see the obvious solutions. We could extend the freeze phase to include the release testing or as we are doing now use the release branch templates until the release is done and switch back to master afterwards.

RichardHaselgrove:
File lib/filesys.cpp is in conflict because it was updated in 29747d2e5c0b544d770d8288857034c0100bb3a9

CharlieFenton:
I disagree with this change. I feel you should instead change the one instance of "ignore_intel_dev" on line 414 in client/log-flags.cpp to "ignore_intel_gpu_dev>" for two reasons:
 [1] For backward compatibility. Since this is what has already been written to cc_config.xml, it will fix the issue for people who have selected this option with no further effort on their part.
 [2] To avoid misunderstanding that it may apply to the Intel CPU which also can perform OpenCL computation. This is why I originally wrote the code with "ignore_intel_gpu_dev>" and I disagreed with Rom's decision to to change it.

TheAspens:
Is there any work in this branch that is not in master?  

Since we are getting closer to starting work on the 7.10 release (once the work for auto-attach is done) which will be branched from master this pull request will not be needed.  However, I'd like to make sure that if there is changes here they are not lost if there are changes we need to keep.

RichardHaselgrove:
All patches mentioned in this PR have now been released to users as part of v7.10: this list is now redundant. Closing.

PR #2066, opened 2017-08-23
Drupal: Modified the forum subscribe link layout. 

drshawnkwang:
Removed "|" for divider, use CSS for a border divider.

Part of https://dev.gridrepublic.org/browse/DBOINCP-309

### DIFF
--- a/drupal/sites/default/boinc/themes/boinc/templates/node-forum.tpl.php
+++ b/drupal/sites/default/boinc/themes/boinc/templates/node-forum.tpl.php
@@ -76,8 +76,8 @@
 <div id="top"></div>
 
 <div id="node-<?php print $node->nid; ?>" class="<?php print $classes; ?> clearfix<?php echo ($first_page) ? '' : ' not-first-page'; ?>">
-  
-  <?php 
+
+  <?php
     if ($page) {
       // Set topic title as page title
       drupal_set_title($title);
@@ -100,7 +100,7 @@
     </div>
     <div class="subscribe">
       <?php if ($subscribe_link): ?>
-        <a href="#block-comment_form_block-comment_form">Post new comment</a> |&nbsp;
+        <a href="#block-comment_form_block-comment_form">Post new comment</a>
         <?php print $subscribe_link; ?>
       <?php endif; ?>
     </div>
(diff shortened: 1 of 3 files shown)


drupal/sites/default/boinc/themes/boinc/templates/node-forum.tpl.php
```diff
@@ -100,7 +100,7 @@
     </div>
     <div class="subscribe">
       <?php if ($subscribe_link): ?>
-        <a href="#block-comment_form_block-comment_form">Post new comment</a> |&nbsp;
+        <a href="#block-comment_form_block-comment_form">Post new comment</a>
```
tristanolive: I'm guessing there is a corresponding commit to the E@H theme (separate repo) that adds the right border to this "Post new comment" link. I don't yet have that locally, so it appears broken -- this is normal. However, the change does break the BOINC theme, which is in this repo, so that should be addressed.

This can be done either by applying the CSS changes made to the E@H theme also to the BOINC theme, or perhaps more properly using the `<ul>` "links" class and wrapping the subscribe and new comment links in `<li>` elements (each with "first" or "last" class, as appropriate) underneath that `<ul>` container. This should then automatically add the proper border in both themes without any CSS changes being required (see the comment edit / quote links, for example).

Also related to this section of code, it seems questionable that "Post new comment" is rendered only if $subscribe_link is set. In practice, it's probably ok and behaves as expected, but logically the ability to post a comment does not hinge on the subscription link being available. Making them independent does make it a bit trickier to set the right "first" and "last" classes in the `<li>` elements, though, e.g. if there's no subscribe link, the post new comment link is the first and the last link.

drupal/sites/default/boinc/themes/boinc/templates/node-forum.tpl.php
```diff
@@ -110,8 +110,8 @@
   <?php if ($unpublished): ?>
     <div class="unpublished"><?php print bts('Unpublished', array(), NULL, 'boinc:comment-action-links'); ?></div>
   <?php endif; ?>
-  
-  <?php 
+
+  <?php
```
tristanolive: All of these whitespace changes should be in a separate "clean up" commit, so as not to detract from the purpose of this commit. (Though I'm not sure it's even necessary, as they are not fixing indentation / readability issues?)

brevilo: Yes, please 👍

drshawnkwang:
It's easier for me to close this PR and issue a new one off of a new branch.

PR #2072, opened 2017-08-25
small fix in delete_spammers.php

davidpanderson:


ChristianBeer:
Could the commit message please state the issue this fix is fixing? From looking at the diff I see what it fixes but the summary description should also do that so it is easier for someone later just looking through the commit description to judge if it is worth looking at the diff or not.

PR #2073, opened 2017-08-25
Get platforms

davidpanderson:


TheAspens:
@davidpanderson  - what issue is the ati change related to?  I wanted to review that to understand the change.

The delete_spammers.php change looks fine (although if it is also related to an issue, we should note that as well).

TheAspens:
Just noticed that the delete_spammers.php change is already present in pull request #2072

JuhaSointusalo:
This change is because LHC has a plan class `native_mt`. The code matches the `ati` in it and LHC gets AMD GPU icon in [project list](http://boinc.berkeley.edu/projects.php).

Rather than improving heuristics I think a better long term solution would be to explicitly list used GPUs or other requirements in `get_project_config.php`. This way projects can freely decide how to name their plan classes and there would be no more need to patch the heuristics and add workarounds.

For projects using XML based plan classes I would copy the plan class as is into `get_project_config.php`. Any additions to plan classes would be immediately available there without any code changes. Some fancier Manager could check `get_project_config.php` to see if user's host meet the requirements, like driver version or CPU features, for any of the plan classes. For projects using code based plan classes just fall back to heuristics.

If I'm reading the code right this change as it is now will result in at least Einstein losing the AMD GPU icon in project list.

davidpanderson:
The doc says "ati":
https://boinc.berkeley.edu/trac/wiki/AppPlan#GPUplanclasses
so maybe we should revert this change and suggest that LHC change their plan class name.

JuhaSointusalo:
It never occurred to me that BOINC would have strict rules about how plan classes must, or must not, be named.

I believe that what I suggested earlier would be an improvement over the current situation. If you disagree please say so. Or if you agree but don't want to implement it right now please open an issue for it.

davidpanderson:
I think the current situation is fine; allowing arbitrary plan class names is not important.

PR #2080, opened 2017-08-29
Drupal: Modified how subscribe link layout. 

drshawnkwang:
Removed "|" for divider, use CSS for a border divider, and add permission check for new comment link.

Part of https://dev.gridrepublic.org/browse/DBOINCP-309

### DIFF
--- a/drupal/sites/default/boinc/themes/boinc/templates/node.tpl.php
+++ b/drupal/sites/default/boinc/themes/boinc/templates/node.tpl.php
@@ -74,11 +74,14 @@
  */
 ?>
 
-<?php if ($subscribe_link): ?>
-  <div class="subscribe">
+<div class="subscribe">
+  <?php if (user_access('post comments')): ?>
+    <a href="#block-comment_form_block-comment_form">Post new comment</a>
+  <?php endif; ?>
+  <?php if ($subscribe_link): ?>
     <?php print $subscribe_link; ?>
-  </div>
-<?php endif; ?>
+  <?php endif; ?>
+</div>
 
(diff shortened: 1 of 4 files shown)


drupal/sites/default/boinc/themes/boinc/templates/node.tpl.php
```diff
@@ -74,11 +74,14 @@
  */
 ?>
 
-<?php if ($subscribe_link): ?>
-  <div class="subscribe">
+<div class="subscribe">
+  <?php if (user_access('post comments')): ?>
+    <a href="#block-comment_form_block-comment_form">Post new comment</a>
+  <?php endif; ?>
+  <?php if ($subscribe_link): ?>
```
tristanolive: This does not account for the case of logged in users that do not have permission to post comments, either due to ban or insufficient credit. In that case, the subscribe link will appear with a divider on the left side even though it is the first and only link. This is why links elsewhere are contained by a `<ul class="links">` structure with `<li>` elements that are given "first" or "last" class, as appropriate. The logic has to be a bit complicated, otherwise the styling breaks down. 

Related, it looks like the same or a similar bug exists in the comment template, which is rendering a divider to the right of the last link for users who are not allowed to post.

drshawnkwang: The new commit changes the if statements around and uses the <ul> and <li> tags (and CSS classes) to generate the border between links.

The bug regarding comment links you found will be fixed separately.

drshawnkwang:
Thanks Tristan for the CSS tip. I've added it to the corresponding AEI GitLab MR.

Issue #656, opened 2015-02-04
Remove decimal points from 'Statistics' graph


romw:
**Reported by Peter Moss on 25 Feb 38440730 04:00 UTC**
It's only a small thing that I find annoying.
Your graphing functions are really very good with the multiple steps etc. depending on scale, but... Its final the labeling of the axes and the redundant decimal points that you insist in having.

i.e. My Seti graph can go from 363500.0  to 368000.0 depending if i show all or just one project. In fact ALL my graphs show the '.0' on each of the Y-Axis labels.

In our company we strip these non-essential decimalisation points as they crowd the screen! If each of the labels contains a '.' then we check to see if the legends all have the last char of '0' if all do, then strip them off. Repeat till you reach a non-zero in one or the '.', if the '.' then strip that too!

It looks neater and takes less space up!

Migrated-From: http://boinc.berkeley.edu/trac/ticket/679

Ageless93:
I think this issue can be solved now as well. Vitalii, if you have a moment, can you check if this is a lot of work? I can assume that with the multiple trillions of credits in some of the projects, that the trailing .00 are a bit excessive.

AenBleidd:
@Ageless93 Yeah, I'll check it.

AenBleidd:
Fixed in #2081

AenBleidd:
@TheAspens, could you please take a look at this ticket?
It is already fixed and ready for release.
Thanks

AenBleidd:
Fixed via #2081

TheAspens:
@AenBleidd - thanks for just proceeding to close.  Sorry for the slow response.

PR #2081, opened 2017-08-30
Manager: For Statistics graph show values on Y axis with 2 digits aft…

AenBleidd:
…er point if these two digits are not zero only

### DIFF
--- a/clientgui/ViewStatistics.cpp
+++ b/clientgui/ViewStatistics.cpp
@@ -739,8 +739,11 @@ void CPaintStatistics::DrawAxis(wxDC &dc, const double max_val_y, const double m
 	double y_start_val = ceil(min_val_y / d_oy_val) * d_oy_val;
 	d_oy_count = (int)floor((max_val_y - y_start_val) / d_oy_val);
 
+    double temp_int_part;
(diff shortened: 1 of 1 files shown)


AenBleidd:
Fixed #656

clientgui/ViewStatistics.cpp
```diff
@@ -739,8 +739,11 @@ void CPaintStatistics::DrawAxis(wxDC &dc, const double max_val_y, const double m
 	double y_start_val = ceil(min_val_y / d_oy_val) * d_oy_val;
 	d_oy_count = (int)floor((max_val_y - y_start_val) / d_oy_val);
 
+    double temp_int_part;
+    int prec = modf(d_oy_val, &temp_int_part) == 0.0 ? 0 : 2;
```
ChristianBeer: How does this cope with values like `5.001`? Shouldn't that set precision to 0? Instead it is set to 2.

AenBleidd: Ok. Will do

AenBleidd: Fixed

AenBleidd:
@davidpanderson, could you please review and merge this? Thanks

davidpanderson:
Thanks!

PR #2082, opened 2017-08-30
Locale: Update french localization files [skip ci]

ChristianBeer:
I forgot to compile the last minute change I made to fr/BOINC-Manager.po as part of a3a7b650. This brings the compiled .mo files in sync with the .po files.


davidpanderson:
Thanks.  I agree that this change doesn't merit doing a 7.8.3 release immediately.

PR #2089, opened 2017-09-06
Client: gather libc version on Linux

### LINKED ISSUE #2064: add Linux libc version info to host.os_version field
It's very helpful to know the libc version of hosts attached to a project so scientists can create compatible applications or target applications to older operating systems. Currently only the kernel version is stored in `host.os_version` e.g. `2.6.32-573.18.1.el6.x86_64` and starting with 7.8 the OS name and version too e.g.: `Debian GNU/Linux 9.1 (stretch) [4.9.0-3-amd64]`. From this you can not easily extract the libc version available on the host.

I propose to add code to the client (in [hostinfo_unix.cpp:HOST_INFO::get_os_info()](https://github.com/BOINC/boinc/blob/master/client/hostinfo_unix.cpp#L1589)) that detects the version of libc, reports it to the project server and stored it in the DB as part of `host.os_version`.

Detecting libc version on Linux based systems (mainly glibc based ones) can be done in several ways as shown on this StackExchange [question](https://unix.stackexchange.com/questions/120380/what-c-library-version-does-my-system-use). It all comes down to parsing version information either directly from `libc.so` or `ldd --version`. Unfortunately this does not work for BSD distributions (ldd doesn't support --version and libc.so is not executable). I found it easier to use `ldd --version` on Debian and RedHat based systems as the examples I gathered where simple and stable enough for this purpose. 
Example output from `PATH=/usr/bin:/bin:/usr/local/bin /usr/bin/ldd --version 2>&1` is:
Debian Stretch: `ldd (Debian GLIBC 2.24-11+deb9u1) 2.24`
Ubuntu 12.04LTS: `ldd (Ubuntu EGLIBC 2.15-0ubuntu10.18) 2.15`
Fedora 25: `ldd (GNU libc) 2.24`

I already have code that extracts the two information. The question to me now is what to do with it. Or rather how to add it to the existing `host.os_version` string so it is useful. With useful I mean that this information can be used for example in plan_classes so that certain app_versions can be targeted for specific libc versions or exclude specific libc versions. Another usage scenario is a census of available libc versions in a project to see what the minimum libc version is to use for a science app.

Here are the possibilities I see to add this new information to the existing `host.os_version` information:

1. `Debian GNU/Linux 9.1 (stretch) [4.9.0-3-amd64|libc 2.24]`
2. `Debian GNU/Linux 9.1 (stretch) [4.9.0-3-amd64|libc 2.24 (Debian GLIBC 2.24-11+deb9u1)]`
3. `Debian GNU/Linux 9.1 (stretch) [4.9.0-3-amd64|libc 2.24 (2.24-11+deb9u1)]`

The plan_class `os_version` filter uses a regular expression which just needs to start with `libc ` in order to restrict to certain versions. Using the information directly from DB and piping it through awk/sed is also easily possible to further extract the desired information although the square bracket format is not ideal.

I would like to get comments and thoughts from @BOINC/team-boinc and @BOINC/contributors  as well as a vote from @BOINC/committer on how to proceed.

ChristianBeer:
Intend.

Fixes #2064

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1584,6 +1584,39 @@ int HOST_INFO::get_memory_info() {
     return 0;
 }
 
+#if LINUX_LIKE_SYSTEM
+int get_libc_version(string& version, string& extra_info) {
+    char buf[1024];
+    string strbuf;
+    FILE* f = popen("PATH=/usr/bin:/bin:/usr/local/bin /usr/bin/ldd --version 2>&1", "r");
+    if (f) {
+        fgets(buf, sizeof(buf), f);
+        pclose(f);
+        strbuf = (string)buf;
+        strip_whitespace(strbuf);
+        std::string::size_type parens1 = strbuf.find('(');
+        std::string::size_type parens2 = strbuf.rfind(')');
+        std::string::size_type blank = strbuf.rfind(' ');
+
+        if (blank != std::string::npos) {
+            // extract version number
+            version = strbuf.substr(blank+1);
+        } else {
+            return 1;
+        }
+        if (parens1 != std::string::npos && parens2 != std::string::npos) {
+            // extract extra information without parenthesis
+            extra_info = strbuf.substr(parens1+1, parens2-parens1-1);
+        } else {
+            return 1;
+        }
+    } else {
+        return 1;
+    }
+    return 0;
+}
+#endif
+
 // get os_name, os_version
 //
 int HOST_INFO::get_os_info() {
@@ -1628,6 +1661,7 @@ int HOST_INFO::get_os_info() {
     bool found_something = false;
     char buf[256],buf2[256];
     char dist_pretty[256], dist_name[256], dist_version[256], dist_codename[256];
+    string os_version_extra = "";
     strcpy(dist_pretty, "");
     strcpy(dist_name, "");
     strcpy(dist_version, "");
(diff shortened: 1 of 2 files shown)


client/hostinfo_unix.cpp
```diff
@@ -1584,6 +1584,39 @@ int HOST_INFO::get_memory_info() {
...
+        if (parens1 != std::string::npos && parens2 != std::string::npos) {
+            // extract extra information without parenthesis
+            extra_info = strbuf.substr(parens1+1, parens2-parens1-1);
+        } else {
+            return 1;
+        }
+    } else {
+        return 1;
+    }
+    return 0;
```
AenBleidd: This code is unreachable. Exit code of this function will be always 0.

client/hostinfo_unix.cpp
```diff
@@ -1584,6 +1584,39 @@ int HOST_INFO::get_memory_info() {
...
+            return 1;
+        }
+        if (parens1 != std::string::npos && parens2 != std::string::npos) {
+            // extract extra information without parenthesis
+            extra_info = strbuf.substr(parens1+1, parens2-parens1-1);
+        } else {
+            return 1;
+        }
+    } else {
+        return 1;
```
AenBleidd: Are you sure that in case of the inability to to make open we should get exit code 1 but not 0?

ChristianBeer:
This was just a demonstration at the BOINC workshop 2017 in Paris. I'm going to redo this and provide a proper commit message.

client/hostinfo_unix.cpp
```diff
@@ -1584,6 +1584,39 @@ int HOST_INFO::get_memory_info() {
...
+
+        if (blank != std::string::npos) {
+            // extract version number
+            version = strbuf.substr(blank+1);
+        } else {
+            return 1;
+        }
+        if (parens1 != std::string::npos && parens2 != std::string::npos) {
+            // extract extra information without parenthesis
+            extra_info = strbuf.substr(parens1+1, parens2-parens1-1);
```
AenBleidd: what will happen if for some reason closing bracket will be in front of opening? e.g. ')('?
Then parens2-parens1 will be negative.
I think it is better to add such check

client/hostinfo_unix_test.cpp
```diff
@@ -135,6 +135,37 @@ void unescape_os_release(char* buf) {
     *out = 0;
 }
 
+int get_libc_version(string& version, string& extra_info) {
```
AenBleidd: We really need the clone of this function in unit tests?

client/hostinfo_unix_test.cpp
```diff
@@ -559,15 +591,28 @@ int main(void) {
...
-        strcat(buf2, os_version);
-        strcat(buf2, "]");
+        os_version_extra = (string)os_version;
         safe_strcpy(os_version, buf2);
         if (strlen(dist_name)) {
             strcat(os_name, " ");
             strcat(os_name, dist_name);
         }
     }
+
```
AenBleidd: Maybe it is better to move duplicating code to some function?

PR #2092, opened 2017-09-07
Fix build with new glibc 2.26

LocutusOfBorg:
* The nonstandard header <xlocale.h> has been removed.  Most programs should
   use <locale.h> instead.  If you have a specific need for the definition of
   locale_t with no other declarations, please contact
   libc-alpha@sourceware.org and explain.

the old glibc seems to still build correctly, build ongoing in
https://code.launchpad.net/~costamagnagianfranco/+recipe/boinc-upstream-daily

### DIFF
--- a/clientgui/AsyncRPC.cpp
+++ b/clientgui/AsyncRPC.cpp
@@ -19,7 +19,7 @@
 #pragma implementation "AsyncRPC.h"
 #endif
 
-#if !(defined(_WIN32) || (defined(__WXMAC__) && (MAC_OS_X_VERSION_MIN_REQUIRED < MAC_OS_X_VERSION_10_4)))
+#if HAVE_XLOCALE_H
 #include <xlocale.h>
 #endif
 
--- a/lib/gui_rpc_client.h
+++ b/lib/gui_rpc_client.h
@@ -809,7 +809,9 @@ struct RPC {
 
 #elif defined(__APPLE__) && (MAC_OS_X_VERSION_MIN_REQUIRED < MAC_OS_X_VERSION_10_4)
 // uselocale() is not available in OS 10.3.9 so use weak linking
+#if HAVE_XLOCALE_H
 #include <xlocale.h>
+#endif
 extern int		freelocale(locale_t) __attribute__((weak_import));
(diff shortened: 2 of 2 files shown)


ChristianBeer:
Fails on Windows with:
`..\clientgui\AsyncRPC.cpp(23): fatal error C1083: Cannot open include file: 'xlocale.h': No such file or directory`

Notice the negation at the beginning of the macro. `xlocale.h` is only included on Linux and Mac >= 10.4. I'm not sure why the header is needed but we should probably look into how to replace it with `locale.h`

CharlieFenton:
First of all, the conditional as patched in this pull request will always evaluate to true, since _WIN32 and __WXMAC__ will never both be defined. The original (unpatched) statement is correct. 

Current builds of BOINC for the Mac set a minimum OS version of OS 10.6. As explained in comments further down in AsyncRPC.cpp (and also in gui_rpc_client.h), the MAC_OS_X_VERSION_MIN_REQUIRED test was put there back when we supported older versions of OS X because neither the xlocale.h header nor the APIs uselocale(), newlocale() and freelocale() were available before OS 10.4. (Because the minimum OS X version is now 10.6, xlocale.h is no longer included by gui_rpc_client.h, but is still needed by AsyncRPC.cpp.)

Note also that config.h has:
> /* Define to 1 if you have the <xlocale.h> header file. */
because configure.ac has checks for xlocale.h.

Looking on line, it appears that uselocale(), newlocale(), freelocale() and LC_GLOBAL_LOCALE are all defined in locale.h. So lines 22 - 24 of AsyncRPC.cpp could be changed to:
#ifdef __WXMAC__
#include <xlocale.h>
#elif !defined(_WIN32)
#include <locale.h>
#endif

Or if AsyncRPC.cpp compiles on Linux without including locale.h, then simply:
#ifdef __WXMAC__
#include <xlocale.h>
#endif

or the following:
#if HAVE_XLOCALE_H
#include <xlocale.h>
#endif

LocutusOfBorg:
I pushed a fix, I'm testing right now in Ubuntu, to see how it builds, thanks you both!

LocutusOfBorg:
it worked thanks!

PR #2097, opened 2017-09-08
update french .mo file; update installshield files

davidpanderson:


AenBleidd:
@davidpanderson do we really need all this changes made in 7.8.2 to be merged back to master?

JuhaSointusalo:
As far as I can tell, the French .mo file was merged earlier and the InstallShield files were pushed to 7.8 directly without a pull request.

JuhaSointusalo:
Actually, @AenBleidd had it right. The direction is from 7.8 to master. This needs to be closed without merging.

davidpanderson:
Nope.  I made a mistake of some sort.

PR #2098, opened 2017-09-08
Removing prepended base url from links, images, css & js for easier initial server startup

grctest:
If an user is experimenting with the BOINC server docker container & they don't know about the configuration files, then the website will have broken links, images, js and css.

This pull request removes the prepended URL segment. Not having a prepended / character means that the hyperlink should append to the end of the base URL.

I have not applied this change to the external facing hyperlinks such as functions which involve emails, notices, rss, etc - since they are external and require the full hyperlink defined.

There was a note in the code that if several pages/folders deep (ie /boincserver/user/something/index.php) that the links will break, however I wasn't able to find such an area within the web server - can you point out where such issues are prevalent & I will investigate why this is the case.

Thanks.


marius311:
The problem is just that when you're running a server and it doesn't actually live at at `<master_url>` (as is usually the case in any test environment, i.e. boinc-server-docker or otherwise), the CSS and many links will be broken. This just converts several absolute paths to relative ones to avoid this happening. 

Maybe we overlooked some consequence of this, do you have an example of something this would break?

davidpanderson:
I'm not sure what you mean by "doesn't actually live at".  Example?  I'm sure the prefix stuff is there for a reason.

davidpanderson:
Some projects have PHP scripts in subdirectories of html/user, which call functions in html/inc that generate HTML with links.  These links need to be absolute.

I can probably suggest another way to fix the underlying problem.

grctest:
What projects specifically require these absolute links though? The external facing generated html/text are not changed (since they do require absolute links); these changes make the BOINC server (launched via docker) fully functional without configuration of such a master URL.

Are these PHP scripts you refer to open source for analysis within the BOINC repo? If not, why should such closed source homebrew PHP scripts be supported in a manner which imposes cosmetic/functional issues upon inexperienced developers?

davidpanderson:
e.g. SETI@home. Not closed source, and I'm not sure what you mean by "homebrew".

grctest:
Can't find the source:
http://setiathome.berkeley.edu/seti_source/ -> No permission
https://sourceforge.net/projects/setiboinc/files/?source=navbar -> Empty
http://boinc.berkeley.edu/cgi-bin/cvsweb.cgi/ -> 404
cvs -d :pserver:anonymous@alien.ssl.berkeley.edu:/home/cvs/cvsroot checkout seti_boinc -> Unknown host alien.ssl.berkeley.edu.

Homebrew as in not standard code within the core BOINC repo, no offence intended.

What about within the util.inc section introducing a check for master_url/url_base/secure_url_base -> If empty, it doesn't prepend (relative hyperlink) otherwise prepend (absolute hyperlink). That way, these subdirectory scripts still function whilst the rest of the website uses relative links?

marius311:
> Some projects have PHP scripts in subdirectories of html/user, which call functions in html/inc that generate HTML with links. These links need to be absolute.

Ahh OK, I don't think we considered this. 

> I'm not sure what you mean by "doesn't actually live at". Example? 

For example, if I run boinc-server-docker inside a VM, then to connect to my server I go the URL `$VM_IP/boincserver` rather than `127.0.0.1/boincserver`. However, since boinc-server-docker's `<master_url>` is `127.0.0.1`, the CSS and many of the links are broken. A similar thing happens when I test the C@H server locally even outside of a VM, since C@H's `<master_url>` is `cosmologyathome.org`. 

There's some workarounds involving SSH forwarding or editing `/etc/hosts`, but it would be nice to not need this, and more generally it seems weird that so much of the BOINC server depends on it knowing its own URL to function, hence this PR. 

> I can probably suggest another way to fix the underlying problem.

I'm curious to hear?

marius311:
> What about within the util.inc section introducing a check for master_url/url_base/secure_url_base -> If empty, it doesn't prepend (relative hyperlink) otherwise prepend (absolute hyperlink). That way, these subdirectory scripts still function whilst the rest of the website uses relative links?

I like this idea, unfortunately there is one key place `<master_url>` is needed and that's for sending the upload URL to the client along with jobs. So if we left `<master_url>` blank for boinc-server-docker and implemented your suggestion, the site would work, but jobs wouldn't upload right (without further changes to the client).

grctest:
What about introducing an additional ['Dir-option' to the make_file tool](https://github.com/BOINC/boinc/blob/master/tools/make_project#L41)?

Potential formats:
>--url_format=absolute|relative
>--absolute_hyperlinks=true|false

Then we could place an if/else check within the util.inc file without breaking external facing hyperlink functionality elsewhere by leaving the master_url option blank.

grctest:
So I've been looking into this, and I've gotten a bit stuck. A simpler, but more hacky solution could be:
> if (strpos (url_base, "SETI") == false) {  absolute } else { relative }

Rather than including a new field in the make_project files (c++ & python) which would require a BOINC server recompile to include the change, we could search the url_base for 'SETI' to decide upon url format within the util.inc file.

This would only work if the url_base in-fact contains the keyword, so ip-based SETI test environments may switch to relative.

davidpanderson:
Marius: I'm not understanding something.
A server has to know its URL, right?  It exports it in various places, like scheduler URLs.

make_project lets you specify the master URL; it doesn't have to be 127.0.0.1

grctest:
>  A server has to know its URL, right? It exports it in various places, like scheduler URLs.

In external facing locations such as rss, email, notices and the instructions on how to add the project to the BOINC client, the server needs to know its URL.

Throughout the website though it's not necessary (we can use relative links), apart from perhaps (can't check the source) the SETI specific user sub-directory php scripts (utilizing the util.inc file).

I've tested the strpos if/else statement within the util.inc file and it works, should I resubmit a PR using a single commit?

127.0.0.1 as the url_base works if you're both running the docker container and the web browser locally on the same linux machine, however if you spin up the docker container on an VPS or a separate machine from your browser then the use of absolute links (without changing the url_base to the external facing ip) breaks the images/css/js/links.

If you want to utilize the external facing functionality (rss/email/notices/scheduler) on an vps you'll need to specify the url_base instead of using the default 127.0.0.1 - which somewhat negates the need for this change. That said, it'd be beneficial for new boinc/docker users not to view a broken website if they skip the url_base change step (IMO).

grctest:
Should I resubmit the PR since I've now accounted for seti within the util.inc file & since this PR is no longer displaying the latest commits?

I'm still unable to view the SETI@Home source code, is there a live repo somewhere that I can view?

Cheers.

davidpanderson:
No; you need to find some other way to do what you want.

grctest:
It's not something I'm interested in maintaining as a BOINC fork over TBH, if we're able to use relative hyperlinks within the BOINC project and use absolute hyperlinks in the external facing functionality and the sub-user pages (accounting for detecting SETI within the util.inc file as implemented in the latest PR commit) then I don't think it's that disruptive of a change.

I did find [a working seti@home boinc project repo](https://setisvn.ssl.berkeley.edu/trac/browser/seti_boinc_html), can you specify the specific sub-user pages which clash with this PR please?

I'm not entirely comfortable using the above 'working' repo link though because:
* The SSL certificate [gets an 'F' grade from SSLLabs](https://www.ssllabs.com/ssltest/analyze.html?d=setisvn.ssl.berkeley.edu) (cert mismatch & insecure key cyphers)
  * I had to force a secondary web browser to disregard the SSL warnings, all downloaded files are at risk of MITM attack.
* Apache is configured [to output the verbose error message](https://setisvn.ssl.berkeley.edu/verbose_apache_error_message_identifies_apache_version.html) "Apache/2.2.15 (Scientific Linux)".
  * This [apache server version is 7 years old](https://github.com/apache/httpd/releases/tag/2.2.15) and [has 27 active vulnerabilities](http://www.cvedetails.com/vulnerability-list/vendor_id-45/product_id-66/version_id-93077/Apache-Http-Server-2.2.15.html).
    * ***That said***, looking into the [SL6 release notes lately](https://www.scientificlinux.org/category/sl-errata/slsa-20172478-1/) they've been patching the recent high score CVEs whilst maintaining the 2.2.15 version number (pretty confusing versioning). ***Probably a non-issue*** in this case.

Any chance you could throw the SETI source code up alongside the [other UCBerkeleySETI repositories](https://github.com/UCBerkeleySETI) on github?

For reference - broken source code repositories:
```
http://setiathome.berkeley.edu/seti_source/ -> 403 Forbidden error
https://sourceforge.net/projects/setiboinc/files/?source=navbar -> No files stored on sourceforge
http://boinc.berkeley.edu/cgi-bin/cvsweb.cgi/ -> 404 error
cvs -d :pserver:anonymous@alien.ssl.berkeley.edu:/home/cvs/cvsroot checkout seti_boinc -> "Unknown host alien.ssl.berkeley.edu."
```

Cheers.

Issue #2103, opened 2017-09-09
Bug: client library: inconsistent terminology for intel_gpu


RichardHaselgrove:
Previous reference: https://lists.ssl.berkeley.edu/pipermail/boinc_alpha/2017-August/022023.html

This problem was reported during alpha testing for the 7.8 branch. Charlie Fenton identified a probable cause, but it was never transferred into bug-tracking. The problem has been re-reported as an ongoing bug in v7.8.2

RichardHaselgrove:
Actually, it seems to be worse than Charlie thought. There are two separate cc_config.xml handlers:

CC_CONFIG::parse_options_client in \client\log_flags.cpp
CC_CONFIG::parse_options in \lib\cc_config.cpp

Both routines say that they should be kept in sync with each other, but they haven't been. ignore_intel_gpu_dev vs. ignore_intel_dev is merely one symptom.

I've confirmed Cliff Harding's report: when ignore_intel_dev is placed manually into cc_config.xml, it works. But when cc_config.xml is updated via the GUI 'Event log options' dialog, the intel_dev line is removed. I'm about to test updating the library file (only: three occurrences) and seeing if a new build fixes it.

RichardHaselgrove:
Yes, confirmed: changing the library file fixes Cliff's reported problem - the 'ignore' line is re-written correctly. But I'd recommend that an experienced developer re-checks the synchronisation between the two files.

RichardHaselgrove:
Commit fa0eb5d should be good to go, but I failed to make a proper pull request.

AenBleidd:
@RichardHaselgrove,
I guess you need to create a separate branch and make cherry-pick with the
necessary commit.

Thanks

Best regards,
Vitalii Koshura

2017-09-11 17:43 GMT+03:00 RichardHaselgrove <notifications@github.com>:

RichardHaselgrove:
Thanks - that seems to have worked.

RichardHaselgrove:
Issue fixed in both master and v7.8.3

PR #2109, opened 2017-09-11
client library: inconsistent terminology for intel_gpu

RichardHaselgrove:
fixes presenting problem in #2103

### DIFF
--- a/lib/cc_config.cpp
+++ b/lib/cc_config.cpp
@@ -393,7 +393,7 @@ int CC_CONFIG::parse_options(XML_PARSER& xp) {
             ignore_gpu_instance[PROC_TYPE_AMD_GPU].push_back(n);
             continue;
         }
-        if (xp.parse_int("ignore_intel_gpu_dev", n)) {
+        if (xp.parse_int("ignore_intel_dev", n)) {
             ignore_gpu_instance[PROC_TYPE_INTEL_GPU].push_back(n);
             continue;
         }
@@ -619,7 +619,7 @@ int CC_CONFIG::write(MIOFILE& out, LOG_FLAGS& log_flags) {
 
     for (i=0; i<ignore_gpu_instance[PROC_TYPE_INTEL_GPU].size(); ++i) {
         out.printf(
-            "        <ignore_intel_gpu_dev>%d</ignore_intel_gpu_dev>\n",
+            "        <ignore_intel_dev>%d</ignore_intel_dev>\n",
             ignore_gpu_instance[PROC_TYPE_INTEL_GPU][i]
         );
     }


CharlieFenton:
As I just wrote for PR #2065:

I disagree with this change. I feel you should instead change the one instance of "ignore_intel_dev" on line 414 in client/log-flags.cpp to "ignore_intel_gpu_dev>" for two reasons:
 [1] For backward compatibility. Since this is what has already been written to cc_config.xml, it will fix the issue for people who have selected this option with no further effort on their part.
 [2] To avoid misunderstanding that it may apply to the Intel CPU which also can perform OpenCL computation. This is why I originally wrote the code with

RichardHaselgrove:
Fair enough - I understand the point you're making. The trouble is that the documentation that people like Cliff Harding have been relying on for the last five years says "ignore_intel_dev", and we can't just break that without thinking of the consequences for existing users.

If we do switch to using "ignore_intel_gpu_dev" at this point, I suggest that we should revert the whole  of Rom's 516eff60b045a1d916f2be3e47e0857c1efbe364, to make internal variable name_space available for other devices: I would like to see BOINC support intel_phi, for example.

That becomes a bigger, but more honest change. The only justification for doing it the way I wrote it was that it's a simple fix which corrects the bug which Cliff discovered, and gives us a consistent platform (hopefully for v7.8.3) from which we can build a fuller ecosystem once our coders are working at full speed again.

Ageless93:
Of course the other reason for using ignore_intel_dev is that it is in the same order as the other brand-name devices that can be ignored: ignore_ati_dev, ignore_nvidia_dev. I don't think users will so easily be confused here that they need the specification that the Intel GPU is ignored, when the documentation shows all devices around that option are GPUs. 

Else the simplification here would be to just go with ignore_gpu_dev with the device number of the GPU you want to ignore.

CharlieFenton:
I do undertand that a case can be made either way, so I will leave it to what others feel is best. 

I had forgotten that Rom's changes to the (internal and external) terminology were so extensive. And I expect that it is unlikely that BOINC would ever support OpenCL processing with the CPU, which I suspect is not efficient, so we don't need to worry about how to add that in the future if <ignore_intel_dev> is already in use for the GPU.

I withdraw my objection.

TheAspens:
@CharlieFenton - since you are committer who has knowledge of the implication of this change, can you take the role of reviewer of this PR and merge it if you think it is ready?

ChristianBeer:
I support Richards and Jords point about keeping the option consistent with other options. Please note that this PR is merging into the 7.8 release branch. I'm not sure it is possible to integrate the change into master based on the release branch. There is currently no policy regarding release branches so I would advise to merge into master now and decide later on how to integrate this change into the release branch.

CharlieFenton:
I agree with Christian's comment that this should first be merged into master and from there into the 7.8 branch. I don't know how to do that from the web interface. I assume the "merge pull request" button will merge it into the 7.8 branch. I use SourceTree on my Mac to make changes and do merges, and prefer to stick to that nethod. The PR says:
> Add more commits by pushing to the intel_gpu_library_fix branch on RichardHaselgrove/boinc.

I can access Richard's repository via the GitHub web interface by clicking on the links in the above sentence, but I don't know how to merge it into master using either the web interfacer or SourceTree.

CharlieFenton:
I suppose I could just use the merge button to merge it to the 7.8 branch and then cherry pick the change back to master. Does that seem like a reasonable approach?

AenBleidd:
@CharlieFenton,

There is no way to do this from GitHub Web UI but this could be done
manually:
https://gist.github.com/alopresto/c05485f80799cd3ba02461b07f1e82de

Best regards,
Vitalii Koshura

2017-09-14 4:03 GMT+03:00 CharlieFenton <notifications@github.com>:

CharlieFenton:
@AenBleidd: If I understand correctly, the PR code is not in the official BOINC repository but rather in Richard's private repository, so one cannot just checkout the intel_gpu_library_fix branch in the pull request. If it were in the BOINC repository, I could do this easily using SourceTree.

But I see this example does use cherry picking to accomplish what we want.

AenBleidd:
@CharlieFenton,
It doesn't really matter where the code is. I just sent you alternative
method to do the same thing :)

Best regards,
Vitalii Koshura

2017-09-14 4:30 GMT+03:00 CharlieFenton <notifications@github.com>:

ChristianBeer:
On GitHub you can't change the source and target branches of a PR. What we do now depends on what we want to have as a result. Do we want to have this fix in the 7.8 release branch so it is included in the next maintenance release? Then we can merge this now and Richard can cherry-pick the commit into a new branch based on `master` and create a new PR that merges the commit into master. If we want to decide later if this should be included in the next 7.8 version we close this PR now and Richard only creates a new PR against `master` as described. We can then later still cherry pick the commit into the release branch.

RichardHaselgrove:
As we've mentioned in the working group - but not yet discussed fully - we've got ourselves into a predicament where there isn't an established procedure for handling bugs in release branches.

The v7.8 release branch was created expecting Github Flow policies - all commits via pull request and code review, direct to master: then no change is needed in the release branch. But that has a precondition: that master is up-to-date and releasable. As has become abundantly clear, our master was nowhere near fit for purpose when the release branch was created.

Now, we have to do two things:

1) dig ourselves out of the hole we've created for ourselves with a seriously compromised public release.
2) adopt and operate internally consistent release policies for the future.

CharlieFenton:
I think we are making this very simple change more complicated than necessary. I've merged it into the 7.8 branch and cherry-picked it into master.

But I agree that we need to get master and the 7.8 branch into complete synchronization before we switch to the model where master is always usable for building dependable software.

RichardHaselgrove:
I agree - especially about the triviality of the change - but I think it's served as a useful worked example of the difficulties we've created for ourselves.

CharlieFenton:
closed #2109

Ageless93:
One thing though, isn't it better that for a future version we change all these ignore_brand_dev to ignore_brand_gpu? This so we don't have the confusion if the 'dev' is a CPU or GPU? We're not ignoring any other devices via this setup anyway. There's for instance no general ignore_coproc_dev. Does BOINC even detect other coprocessors than AMD, Intel and Nvidia GPUs?

PR #2110, opened 2017-09-12
Fixed URL of "Add separate preferences" button

progger:


TheAspens:
Looks good.  I tested with a server build based on master and with your changes.

PR #2117, opened 2017-09-13
Set character set explicitly at database creation

jeremyhansen:
By default, Debian installs of MariaDB set the character set to utf8mb4. This causes the various ALTER TABLE commands from db/constraints.sql called by tools/make_project to fail with an error of "specified key was too long". The proposed change here forces the project database to the one-character-wide latin1 set, which is in line with the char arrays in db/boinc_db_types.h.

### DIFF
--- a/py/Boinc/database.py
+++ b/py/Boinc/database.py
@@ -284,7 +284,7 @@ def create_database(srcdir, config = None, drop_first = False):
     cursor = get_dbconnection().cursor()
     if drop_first:
         cursor.execute("drop database if exists %s"%config.db_name)
-    cursor.execute("create database %s"%config.db_name)
+    cursor.execute("create database %s character set latin1 collate latin1_swedish_ci"%config.db_name)
     cursor.execute("use %s"%config.db_name)
     for file in ['schema.sql', 'constraints.sql']:
         _execute_sql_script(cursor, os.path.join(srcdir, 'db', file))
--- a/py/Boinc/database.py
+++ b/py/Boinc/database.py
@@ -284,7 +284,7 @@ def create_database(srcdir, config = None, drop_first = False):
     cursor = get_dbconnection().cursor()
     if drop_first:
         cursor.execute("drop database if exists %s"%config.db_name)
-    cursor.execute("create database %s character set latin1 collate latin1_swedish_ci"%config.db_name)
+    cursor.execute("create database %s character set utf8 collate utf8_general_ci"%config.db_name)
     cursor.execute("use %s"%config.db_name)
     for file in ['schema.sql', 'constraints.sql']:
         _execute_sql_script(cursor, os.path.join(srcdir, 'db', file))


ChristianBeer:
While this fixes the immediate bug it is not future friendly. I would really like to see UTF8 being the default for new projects so they don't run into charset-hell later on. One example is the team_import from BOINC Wide teams. The team names come as UTF8 encoded and are inserted into latin1 fields and based on what charset was used for the DB connection no transformation is happening. You end up with UTF8 characters in latin1 fields.

jeremyhansen:
Given that most of the longish varchars are 254 characters long, setting the default to utf8 (rather than utf8mb4) and utf8_general_ci doesn't exceed the key limit either. Just tested it and I get no errors.

davidpanderson:
What error did Scrutinizer find?

davidpanderson:
This looks good to me.  Christian, OK to merge?

AenBleidd:
It looks like internal Scrutinizer error:

default

JuhaSointusalo:
The problem with MySQL's `utf8` is that it's not the real thing. It uses only at most 3 bytes but UTF8 requires up to 4 bytes. [MySQL Unicode Support](https://dev.mysql.com/doc/refman/5.7/en/charset-unicode.html). The canonical example of characters that need 4 bytes is U+1F4A9 PILE OF POO 💩. Even if that particular character doesn't seem important many/most other emojis seem to need the same 4 bytes. [Wikipedia](https://en.wikipedia.org/wiki/Emoji)

I'm afraid this will replace one badness with another.

Should this be accompanied with changes to code that uses the DB?

ChristianBeer:
The 3byte `utf8` is enough to support normal language. Converting later to `utf8mb4` seems trivial from the MySQL link Juha posted. The same link says that with `utf8mb4` you can store less characters in a fixed size varchar than with utf8 while also increasing database size for variable size fields. I don't see the benefit of using `utf8mb4` just to support emojis.

I just checked the schema definitions and they don't contain explicit charset definitions. The only place in the code that I'm aware of that is character set sensitive is the BWT team import/export. Other places still expect latin1 mostly. So this change might break things for new projects. Normally such a change would be noted in Release Nots of the server since we don't have that I don't see a way to warn new projects other than the usual that master is currently not stable and needs to be tested before used in production.

JuhaSointusalo:
Volunteers might use emojis in forums posts or profiles or even in user or team names. After quick testing that works currently.

From what I understand, right now BOINC depends on character set being the same on both the database and on the connection. This way MySQL doesn't try to re-code anything and simply passes the data through. It looks like the character set is not explicitly set anywhere so BOINC uses whatever default MySQL has. The actual data BOINC uses and stores is in UTF8.

If the databases's character set is set to `utf8`/`utf8mb4` but the connection is not set but remains `latin1` then MySQL will re-code the data. Re-coding UTF8 to UTF8 as if it was `latin1` and then back might work but will likely result in a headache later on.

There's some other matters to consider as well.
- a project that currently has `latin1` database and upgrades to a newer MySQL/PHP that defaults to `utf8mb4`
- a project that has already switched to `utf8`/`utf8mb4` database

[This blog post](https://mathiasbynens.be/notes/mysql-utf8mb4) points out that using `utf8` might lead to security issues.

According to DistroWatch RHEL 6.9 is supported until 2020-11, has only MySQL 5.1.73 but `utf8mb4` isn't supported until 5.5.3.

I don't really know what to do. I just think that this issue is more complex than just setting the character set for new databases.

For the "specified key was too long" error MySQL manual says one can change InnoDB settings to support longer indexes. [Converting Between 3-Byte and 4-Byte Unicode Character Sets](https://dev.mysql.com/doc/refman/5.7/en/charset-unicode-conversion.html)

lfield:
@jeremyhansen Can your original issue be solved by changing the InnoDB settings to support longer indexes as @JuhaSointusalo suggested?

JuhaSointusalo:
I think there's two tests (or perhaps more) that need to be done. I haven't checked but I assume that on Debian 8 MySQL/MariaDB defaults to `latin1` and on Debian 9 it defaults to `utf8mb4`. If I got the versions wrong just replace them with right versions.

Install BOINC server on Debian 8. Put some Unicode content in database, user names or forums posts or whatever. Then upgrade the server to Debian 9.

Do that both with stock BOINC code (temporarily increase index length) and with this patch. If I understand correctly how MySQL deals with character sets then in both cases Unicode content should be mangled.

lfield:
I have done some testing on this issue. 
Our production DB uses `latin1`. On Ubuntu 18.10 it uses `utf8mb4`.

> `mysql  Ver 15.1 Distrib 10.1.34-MariaDB`

Changing the default to `latin1` as described [here](https://stackoverflow.com/questions/1049728/how-do-i-see-what-character-set-a-mysql-database-table-column-is), inserting 💩 results in `?`. Using `utf8mb4` works but `utf8` gives the same result as `latin1`. Inserting 💩 with `utf8mb4` and altering the character set of the column to `utf8` or `latin1` results in `?`.  Converting back to `utf8mb4` results in `?`.  

As @ChristianBeer says, I don't think that this PR is future proof. My suggestion would be to close this PR and open an issue with the original problem.

PR #2122, opened 2017-09-14
update_versions: handle multiple API_VERSION strings

### LINKED ISSUE #2121: update_versions handles multiple API_VERSIONs badly
update_versions uses this code to extract API_VERSION from executables:
`strings $path | grep API_VERSION`

SixTrack released new app versions and the executables embed these API_VERSION strings:
```
API_VERSION_7.7.0
API_VERSION
LPAPI_VERSION
```

update_versions turns these into this api_version element:
```
    <api_version>7.7.0
API_VERSION
LPAPI_VERSION</api_version>
```

BOINC client in turn expects at most 16 bytes string and Windows versions react badly when the limit is exceeded.

update_versions_v6 was fixed to handle this situation in 7a35b69c642677d048329c0cff56bce8627df55b but the fix has not made it into update_versions.

JuhaSointusalo:
An executable may have more than one string that contains
"API_VERSION" such as:

API_VERSION_7.7.0
API_VERSION
LPAPI_VERSION

If there is more than one string update_versions extracts the
version number from the first one and then puts the version number
and the rest of the strings into <api_version>:

    <api_version>7.7.0
API_VERSION
LPAPI_VERSION</api_version>

Change update_versions to extract only those strings that look like
BOINC's API_VERSION and only the first one if there for some reason
is more than one such string.

Fixes #2121.

This commit adapts 7a35b69c642677d048329c0cff56bce8627df55b to the
new PHP version of update_versions.

### DIFF
--- a/tools/update_versions
+++ b/tools/update_versions
@@ -195,7 +195,7 @@ function get_api_version($a, $v, $p, $fds) {
     foreach ($fds as $fd) {
         if ($fd->main_program) {
             $path = "apps/$a/$v/$p/$fd->physical_name";
-            $handle = popen("strings $path | grep API_VERSION", "r");
+            $handle = popen("strings $path | egrep --max-count=1 '^API_VERSION_[0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*$'", "r");
             $x = fread($handle, 8192);
             pclose($handle);
             $x = strstr($x, "API_VERSION_");


JuhaSointusalo:
@davidpanderson @TheAspens 

cough

TheAspens:
@ChristianBeer - you indicated that you tested and approved these changes.  Do you want to go ahead and merge it also then?

TheAspens:
Actually, I tested this against a bunch of our binaries going back a decade and it pulled the right info out of them.  I approve also.

PR #2124, opened 2017-09-15
path fix (affects only BOINC web site)

davidpanderson:


davidpanderson:
OK, but why do all these merges appear?  Am I doing something wrong?

PR #2126, opened 2017-09-15
BOINC web: fix path for BOINC-wide teams file

davidpanderson:


ChristianBeer:
Maybe the script should be moved to boinc-site too? Maybe also remove the hardcoded paths from public repositories?

PR #2127, opened 2017-09-15
web: use "small" instead of "muted"; fixes color problem

davidpanderson:


html/user/submit_status.php
```diff
@@ -17,6 +17,7 @@
 // along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 
 // web interfaces for viewing and controlling batches
+// DEPRECATED: replaced by submit.php
```
ChristianBeer: This has nothing to do with what is described in the commit message.

PR #2133, opened 2017-09-18
job submission: don't error out if standard templates missing; change comment in transitioner

davidpanderson:
If you specify the input or output templates,
the standard ones (appname_in, appname_out) don't need to exist.

Also, in create_work():
- don't read the output template; just check that it exists.
- deprecate the result_template_filepath arg; redundant

### DIFF
--- a/html/inc/submit.inc
+++ b/html/inc/submit.inc
@@ -340,7 +340,7 @@ if (0) {
     $req->project = "http://isaac.ssl.berkeley.edu/test/";
     $req->authenticator = trim(file_get_contents("test_auth"));
     $req->app_name = "uppercase";
-    $req->batch_name = "batch_name7";
+    $req->batch_name = "batch_name9";
     $req->app_version_num = 710;
     $req->jobs = array();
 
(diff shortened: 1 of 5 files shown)


ChristianBeer:
I don't really like that the changes to `create_work()` are hidden in the commit message. IMHO they are independent from the remote submission changes and warrant their own commit so the commit subject is more descriptive.

davidpanderson:
create_work is part of remote job submission.

ChristianBeer:
`create_work` is a standalone program that can be used for local job submission as well as remote job submission. You are changing the function which affects both use cases. My point is that someone looking for a change made to the standalone program doesn't recognize that a commit message titled "remote job submission" is affecting `create_work` too.

What explanation do you have for the comment change to the transitioner? This one is certainly not part of remote job submission.

In my eyes 921dcd14982c3f8e102ae37a2a736af2b9222b4f is not atomic and contains unrelated changes.

davidpanderson:
I changed the title.  Any problem w/ the content of the PR?

ChristianBeer:
Changing the title of the PR doesn't change the commit message where the unrelated change is embedded in. This is not what is meant by atomic commits. I stand by my objection which is about atomicity not about the quality of the code (which I think is fine).

Issue #2118, opened 2017-09-13
Proof of UserID ownership using public key cryptography


grctest:
# Transparency/Background

* [Original 'project-rain' cryptocurrency field issue](https://github.com/BOINC/boinc/issues/1998): The PR was rejected in accordance with the PMC's policies against the explicit inclusion of (and reference to) virtual currency related code within the core BOINC repo (still possible on an individual project basis).
* [Single user-data field](https://github.com/BOINC/boinc/issues/2087): A scaled back version of 'project-rain' with no explicit reference to virtualcurrencies (implied use, not explicit intended use). Several legitimate risks to BOINC projects were identified during the 2017 workshop, such as painting a massive target on BOINC projects (externally/internally) to hackers monetarily motivated (since projects would have been storing the data being scraped by 3rd party systems).

# Purpose

To prove within any external system that you are the owner of an UserID for an individual BOINC project, without storing external system data within the BOINC project's servers.

### Github links

* [Pull request #2134](https://github.com/BOINC/boinc/pull/2134/)
* [BOINC Issue "Proof of UserID ownership using public key cryptography #2118"](https://github.com/BOINC/boinc/issues/2118)

# Proposed steps within BOINC

* Introduce an additional openssl public/private key pair for this feature, keeping the existing keys separate and safe. Storing the public key in a scrape-able location (perhaps alongside the stats dumps).
* Introduce a link to the user's profile, taking them to a separate php page for generating the signed message.
* Within the new page there would be:
  * Some explanation text.
  * An input textbox (external system data for full handshake).
  * A button for executing the message signing function.
  * An output textbox where the output data will temporarily be displayed.
* The message signing function will:
  * Require:
    * A minimum RAC before enabling the function - reducing the ease of claiming idle account ownership.
    * A verified email account (unless the project doesn't use email verification).
  * Use:
    * [openssl_sign](https://secure.php.net/manual/en/function.openssl-sign.php) to sign the message "$UserID $External_System_Key"
    * External_system_key to provide full handshake, preventing extraction of master UserID from one network for replay on another.
  * Limit the length and accepted characters for $External_System_Key.
  * Potentially be limited to once an hour, if computationally expensive.
     * Downside being the requirement for an additional SQL table.
  * Output the signature and original message to the in-page output textbox. It will not store this data on the BOINC server (message is temporary).

# Proposed use within external system

Pre-req: External system downloads the public-keys (used solely for this function, not the existing keys) from each of the involved projects.

[...]

PR #2134, opened 2017-09-19
Proof of UserID ownership using public key cryptography (openssl)

grctest:
# Relevant issue for more information: https://github.com/BOINC/boinc/issues/2118

Introducing an openssl signing function to prove within any external system that you are the owner of an UserID for an individual BOINC project, without storing external system data within the BOINC project's servers.

It outputs a text file download containing:
```
-----BEGIN USERID VERIFICATION-----
USER_ID: 
USER_DATA: 
MSG: User_ID USER_DATA
SHA512_MSG_DIGEST: 
BASE64_SIGNATURE: 
Verification: 
-----END USERID VERIFICATION-----
```

Would it be more appropriate to include this tool within the 'account keys' page? Or is there a better location for the hyperlink?


grctest:
Just spotted a couple problems:
* Left behind comments & commented out code in the action file
* The form php file has an inappropriate minimum RAC check in place (minimum RAC < 100), it was set to this so that I could test the functionality on a fresh server. A more appropriate RAC should be picked, but what value?

In the mean time, I'd appreciate peer review of the pull request & am open to any/all suggestions for improvement.

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,134 @@
...
+// Needs to be defined, unknown use.
+check_get_args(array("tnow", "ttok"));
+
+// Check the user is online
+$user = get_logged_in_user();
+check_tokens($user->authenticator);
+
+// For some reason, I'm only able to use numbers, no letters.
+// Need to provide security, but full functionality is likewise desirable.
+$user_data = sanitize_tags(post_str("user_data", true));
```
tomasbrod: You don't have to sanitize it if the string is going to be only hashed. These crypto algs can deal with any byte-sequence. (By design, that is. If they can't, then OpenSSL is total shit, which i hope is not.)

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,134 @@
...
+
+// Following line a potential bug.
+// $user_data = BoincDb::escape_string($user_data);
```
tomasbrod: ?

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,134 @@
...
+
+	// Check that the public key file exists where specified. If not, redirect to error page.
+	if (!file_exists($public_key_path)) {
+		error_page(tra("The public key does not exist. Proof of UserID currently not possible."));
+	}
+
+	// Fetch private key from file
+	$private_key_pem = openssl_pkey_get_private($private_key_path);
+
+	// Compute signature using the SHA512 signing algorithm
```
tomasbrod: SHA512 is not signing algorithm...
It's hash/digest algorithm and `openssl_sign` is probably using RSA signing algorithm on top.

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,134 @@
...
+//Has the user entered data with length > 0 & <= 256?
+if (($user_data) && (strlen($user_data) <= 256)) {
+	// Retrieve the logged in user's UserID.
+	$userid = $user->id;
+
+	// Create the message which will be signed.
+	$message_data = "$userid $user_data";
+
+	// Create a digest of the message_data.
+	// Without the digest, the openssl_sign function fails to sign lower case characters.
```
tomasbrod: Woah, wait. Really? How can (hybrid) signing function have problems with some letters? This needs to be investigated and the double hashing step removed.

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,134 @@
...
+	// Check that the public key file exists where specified. If not, redirect to error page.
+	if (!file_exists($public_key_path)) {
+		error_page(tra("The public key does not exist. Proof of UserID currently not possible."));
+	}
+
+	// Fetch private key from file
+	$private_key_pem = openssl_pkey_get_private($private_key_path);
+
+	// Compute signature using the SHA512 signing algorithm
+	openssl_sign($message_digest, $signature, $private_key_pem, OPENSSL_ALGO_SHA512);
```
tomasbrod: return value not checked

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,134 @@
...
+			echo "\nUSER_ID: $userid";
+			echo "\nUSER_DATA: $user_data";
+			echo "\nMSG: $message_data";
+			echo "\nSHA512_MSG_DIGEST: $message_digest";
+			echo "\nBASE64_SIGNATURE: $base64_sig";
+
+			// Check if signature was successfully validated
+			if ($sig_verification == 1) {
+				// The generated signature has been successfully verified using the public key.
+			    echo "\nVerification: GOOD";
```
tomasbrod: Just print an error if it fails (error_page), no need to print success.

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,134 @@
...
+		  // Question: Does outputting the signature as plain text corrupt the signature? Verify manually!
+		  // Verified: Firefox
+		header("Content-Disposition: attachment; filename=signature.txt");
+		header("Content-type: text/plain");
+
+		// Start outputting text to the auto-download txt file
+		echo "-----BEGIN USERID VERIFICATION-----";
+
+			echo "\nUSER_ID: $userid";
+			echo "\nUSER_DATA: $user_data";
```
tomasbrod: Why print the user data when that is contained in message?

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,134 @@
...
+// Following line a potential bug.
+// $user_data = BoincDb::escape_string($user_data);
+
+//Has the user entered data with length > 0 & <= 256?
+if (($user_data) && (strlen($user_data) <= 256)) {
+	// Retrieve the logged in user's UserID.
+	$userid = $user->id;
+
+	// Create the message which will be signed.
+	$message_data = "$userid $user_data";
```
tomasbrod: Missing the timestamp from your proposal @grctest.

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,134 @@
...
+
+// For some reason, I'm only able to use numbers, no letters.
+// Need to provide security, but full functionality is likewise desirable.
+$user_data = sanitize_tags(post_str("user_data", true));
+
+// Following line a potential bug.
+// $user_data = BoincDb::escape_string($user_data);
+
+//Has the user entered data with length > 0 & <= 256?
+if (($user_data) && (strlen($user_data) <= 256)) {
```
tomasbrod: Why exactly 256 limit?

tomasbrod:
I suggest this output format (as boinc uses XML everywhere):
```
<boinc_user_id_verification>
 <master_url>http://boinc.whateverathome.net/</master_url>
 <userid>123</userid>
 <user_data>SjZiale4A...</user_data>
 <timestamp>1505807622</timestamp>
 <signature>BASE64_SIGNATURE</signature>
</boinc_user_id_verification>

```

html/user/openssl_sign_action.php
grctest: True, it's not getting stored in SQL nor is it being stored anywhere in the website, I left this in because it's standard for pretty much every other data input & I was unsure if you could potentially maliciously escape the code without it. If it's safe to remove then we could go ahead and do so.

html/user/openssl_sign_action.php
grctest: Left behind comment - will edit out, was 3-4am when I submitted the PR. I did [spot my mistake](https://github.com/BOINC/boinc/pull/2134#issuecomment-330405745) before falling asleep though.

The boincdb escape string line isn't needed because we're not interacting with a user_data sql table field.

html/user/openssl_sign_action.php
grctest: Initially a 256 limit to prevent passing 1 mil characters in to inflate the text download, likewise hash160's are less than 256.

We could increase this to a few thousand if we want to allow say PGP user data, however I don't imagine most external systems would use more than 256 characters.

What would you propose as an appropriate max character limit?

html/user/openssl_sign_action.php
grctest: I changed my proposal after your reply on steemit, the timestamp was removed because:
* The external system key should be sufficient to prevent cross-network beacon replay attacks (stealing a registration from A to register on B).
* Since UserID doesn't change, the beacon containing this 2-way handshake data should be permanent. If a beacon is invalidated due to external network issues, then perhaps the beacon could be replayed but if the old beacon isn't available for comparison then a timestamp wouldn't stop a replay. Ultimately they could replay but they couldn't change the external system key, so an attacker would just be re-advertising your legit beacon for you (not really an attack).

Think the timestamp should be reintroduced?

html/user/openssl_sign_action.php
grctest: Without creating a message digest from the user data to feed into openssl_sign, any user data which included lower-case data would change the text file output to unrecognizable garbage like:

```
2d2d 2d2d 2d42 4547 494e 2055 5345 5249
4420 5645 5249 4649 4341 5449 4f4e 2d2d
2d2d 2d0a 5573 6572 4944 3a20 310a 5573
6572 4461 7461 3a20 2231 3233 3132 330a
4d65 7373 6167 653a 2031 2022 3132 3331
3233 0a53 6967 6e61 7475 7265 3a20 71ce
1029 7aa8 b0d6 221f 5a83 29b1 4962 3092
80b9 71a1 fce4 17ce 35e1 00d6 7b6a 3ac7
888b bbb2 361e 3ee7 982f 578d eb5d cda8
393e 0bf5 eee7 5b1b 1def 1f10 b545 678c
51ca 8d3e b117 df08 12df 2449 a6fc 646f
abbe 3b86 f6db 9220 77e6 ea4f 8947 628c
3f31 d640 5b9c c987 0771 423f 9d27 c296
a7ca 1783 c6f2 0b99 6b18 6964 aad5 1657
4929 5daa 8a1a ef27 5534 1c7b 0dda ad1f
9860 943b 4f84 0d62 5fe4 44b8 7289 6098
26b1 545c baa9 78c2 d187 0c4b c814 a53b
bc96 b5de 0866 0f9f 5e7a 217a b142 675f
b8bc bd26 10d0 f429 5498 9976 b002 7d06
6689 e30c 651d f68f b7f5 8498 c638 b85b
7f4e b7b1 8d60 49ea 38a8 ddcf 2f80 5515
191c 7d3a 0ac8 7459 029f e57c 8638 0a56
6572 6966 6963 6174 696f 6e3a 2047 4f4f
440a 2d2d 2d2d 2d45 4e44 2055 5345 5249
4420 5645 5249 4649 4341 5449 4f4e 2d2d
2d2d 2d
```

Using a message digest prevents this issue completely, I'm unsure why the above issue occurs but using a message digest seems to be an appropriate solution.

html/user/openssl_sign_action.php
grctest: https://secure.php.net/manual/en/function.openssl-sign.php

The php openssl-sign docs refer to this as a signature algorithm.

https://secure.php.net/manual/en/openssl.signature-algos.php Not verbose regarding how sha512 is implemented.

https://secure.php.net/manual/en/function.openssl-get-md-methods.php This page lists more signature algorithms (including 'sha512', 'RSA-SHA512' and 'sha512WithRSAEncryption') but it's not clear if we need to specifically use the RSA versions within the openssl_sign or if they're used by default when specifying 'OPENSSL_ALGO_SHA512' as the openssl_sign signature algorithm.

html/user/openssl_sign_action.php
grctest: The return value is '$signature' which is checked via:

```
// Check that the signature exists
if ($signature) {
    // The signature was successfully created.
} else {
    //The signature wasn't properly generated.
    error_page(tra("Unable to produce openssl signature. Please try again at a later time."));
}
```

Do you believe this is an insufficient check?

html/user/openssl_sign_action.php
grctest: Left behind debugging info for the lower-case (non-digested) user_data signing, both the USER_ID and USER_DATA could be removed since the MSG line stores both with a space delimiter. Sound good?

html/user/openssl_sign_action.php
grctest: Good point, I added this in as a post-check, but you're right that prior to printing this check should be performed and simply direct to the error_page on failure - I'll change this.

grctest:
@tomasbrod Thanks for the input, I've committed a new version with your suggestions in mind.

New XML format:
```
<boinc_user_id_verification>
 <master_url>http://ip.ip.ip.ip/boincserver/</master_url>
 <userid>1</userid>
 <user_data>testTEST1234</user_data>
 <msg>1 testTEST1234</msg>
 <msg_digest>c6107a933520e6ca013fb47c765d0e6e5b8b66f40e942a74c9baadc437ff1d76f7e787bc91b0067fd60431a8caca202a5d24038631c1040cc2e1b23225de1f50</msg_digest>
 <signature>f2zo+oqF8XEw3vTaLJLg2LF/o6YSVTyG2tHJ6jjL2HTfTBaptV/H7xI5DreeVP7J4sXm5HTlYzj7KPorlYyLc7uDxaIzQyLvwTsOZA3Uv7EH6IhYS4M01/UVm+pwdHCqbdhYPhvHql1bED5EJ7RW5nO6hqJdmhlYWhFojvO1YB5ZH+tnQuuq2IDRfAHP/ziPDO321ApYx5+ULFheONdn12naG1JG3Lwg5TtIcvX26VnSZkxhWQxug+Lkpf/koN1T6bzslk6sQv+GE+hI8Q2kiGjnFsBZE+BytquX8x0QuOUQIy3rwA2EepSAU7OWv88ctUQp3jW3ZO8MgQvpkVBHxw==</signature>
</boinc_user_id_verification>
```

html/user/openssl_sign_action.php
tomasbrod: Idk 1000? 4k?
Output of sha512 as hex is 128 chars, so current 256 limit is probably ok.

html/user/openssl_sign_action.php
```diff
@@ -92,42 +90,40 @@
...
+		if ($sig_verification == 1) {
+			// The generated signature has been successfully verified using the public key.
+
+			// Generate an signature xml file auto-download prompt
+			header("Content-Disposition: attachment; filename=signature.xml");
+			header("Content-type: text/xml");
+
+			// Start outputting xml to the auto-download xml file
+			echo "<boinc_user_id_verification>";
+			 global $master_url; // Define global master_url variable for use in xml output.
```
tomasbrod: indent error

html/user/openssl_sign_action.php
tomasbrod: `openssl_sign` return boolean, which is not saved to variable, thus can't be checked.

html/user/openssl_sign_action.php
tomasbrod: Using message digest is not a solution, it is a workaround. And that output looks really, really bad. This is a serious issue. You got me hooked, I will test myself.

grctest: Looks like I was running an old version of OpenSSL: https://github.com/marius311/boinc-server-docker/issues/26#issuecomment-330553493

Could explain the glitch..

grctest: I've managed to upgrade the openssl library to  1.0.1t, but can't upgrade any further on debian jessie despite the openssl website stating that it's unsupported now..

With this version the removal of the message digest workaround is not required.

html/user/openssl_sign_action.php
grctest: Fixed.

html/user/openssl_sign_action.php
tomasbrod: SHA512 definitely is an hash/digest algorithm, there is no debate on that.
I think it selects Asym algo by the key type you pass it (probably RSA),
and digest algo from the parameter. So using 512 is currently the best thing to do.

html/user/openssl_sign_action.php
grctest: I've changed it to check this fully now, thanks for pointing this out. 👍

grctest:
I'm open to further suggestions, anyone?

We could further add recaptcha via:

```
// form
require_once('../inc/recaptchalib.php');

if ($recaptcha_public_key) {
    form_general("", boinc_recaptcha_get_html($recaptcha_public_key));
}
```

```
// action
if ($recaptcha_private_key) {
    if (!boinc_recaptcha_isValidated($recaptcha_private_key)) {
        error_page(
            tra("Your reCAPTCHA response was not correct. Please try again.")
        );
    }
}
```

html/user/openssl_sign_action.php
tomasbrod: I am worried that it is not clear what this two functions do. What chars are safe? External systems may need to duplicate this sanitation in order to check the signature.

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,132 @@
...
+$user = get_logged_in_user();
+check_tokens($user->authenticator);
+
+$user_data = sanitize_tags(post_str("user_data", true));
+
+//Has the user entered data with length > 0 & <= 256?
+if (($user_data) && (strlen($user_data) <= 256)) {
+
+	// Retrieve the logged in user's UserID.
+	$userid = $user->id;
```
tomasbrod: I am worried that it is not clear what this two functions do. What chars are safe? External systems may need to duplicate this sanitation in order to check the signature.

grctest:
```
$user_data = sanitize_tags(post_str("user_data", true));
```

Sanitize tags:
```
function sanitize_tags($x) {
    return strip_tags($x);
}
...
```
[link to strip_tags function](https://secure.php.net/manual/en/function.strip-tags.php)

So sanitize tags could be removed because we're not worried about html tags though perhaps you could escape the input and inject php code? If an external system isn't signing html tags, just a key or hash then they don't need to worry?

We could possibly strip non-alphanumeric characters, but that could limit some external systems.

```
if (($user_data) && (strlen($user_data) <= 256)) { 
```

If 'user_data' input exists (user has input data) and the length of the input user data is less than 256 characters.

It could be changed to the following for improved consistency
```
if ((strlen($user_data) > 0) && (strlen($user_data) <= 256))
```

Regarding
```
$user = get_logged_in_user();
$userid = $user->id;
```
[link to user table schema](https://github.com/grctest/boinc/blob/07c9c3dacd51d0c1fc88170ba0aac142d9eaa2ee/db/schema.sql#L82)

userid is simply an integer, we could include any user data in the signature but we only really need userid & the user input user_data.

tomasbrod: Looks like I commented on wrong line number, ofc I ment the `sanitize_tags`, sorry.
The strlen is better the way it is, i don't think php is smart enough to optimize that double call away.
For the `sanitize_tags` I think it is not necessary. I am certain there is no place where the text could trigger code execution. Worst thing that can happen is the downloaded xml being invalid. Let's just not complicate things and pass the data directly, both strlen and openssl_sign are byte safe.

grctest: We could implement [htmlspecialcharacters()](https://secure.php.net/manual/en/function.htmlspecialchars.php) instead of sanitize_tags. 

Whilst the data isn't being stored within a database nor output to html, we should secure the user input from php injection (if that's possible, especially if the servers are running old vulnerable php versions).

grctest: I just reduced the to:
```
$user_data = post_str("user_data", true);
```

And you're able to mess with the xml output like:
```
<boinc_user_id_verification>
 <master_url>http://ip.ip.ip.ip/boincserver/</master_url>
 <userid>1</userid>
 <user_data>test</user_data><signature>testing</signature></user_data>
 <msg>1 test</user_data><signature>testing</signature></msg>
 <signature>ZzAAjI2MRR9Zz3L/0ONOgRyK2GA79o9VzRZ5huBpd7s/HJBxYyU9gu97hc7VR/lJbMiifcWDkJCYtmYF6EkyzjEgnjJiTLHpB+/gyxym4Lsi3fW78qnrwo+fL41AAkbDXzCkoW4zF0r9Esw2AcFQhcOZN2n0MwzrSPVpKgt/BRxGOXUn5u3pSN8porDTN+jDJpJuzuMr1A3cZJ/8FGvtX206wJJTdJAevdoAjoqrq9M2UbAaRajGOsQY81LVla7YgBw5z7iQufJIQdAV1kiuO5FDiDJxXaPBLCr3wScLAj95EK/OMFALw5yQlLGgZCudnWqnmlGp6fizx40IY4SHHQ==</signature>
</boinc_user_id_verification>
```

To be fair, you could just edit the xml file manually to produce the same issue in the non-signature fields, and it would probably just be an end-user problem than a problem for the whole external network. Likewise, we don't decrypt the signature but rather reproduce the signature using the msg and project's public key, so yeah it's not that much of a problem.

tomasbrod:
@grctest You should also read the comments "Scrutinizer" says.

grctest:
@tomasbrod I checked what the scrutinizer was complaining about - 'dead code' because of the commented out checks (which are optional) and indentation (which is appropriate if commented code is uncommented.

grctest:
Posted to the BOINC_DEV mailing list for input: https://lists.ssl.berkeley.edu/pipermail/boinc_dev/2017-October/023003.html

Mercosity:
Tried it out and it functions and delivers the download contents as expected. 
Tried to break it with my limited means and couldn't ..

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,148 @@
...
+
+// Input is passed in from the openssl_sign_form
+// sanitize_tags removes html elements from the user input
+// $user_data = htmlspecialchars((post_str("user_data", true)),ENT_XML1);
+
+// user data input without htmlspecialchars nor sanitize_tags
+$user_data = post_str("user_data", true);
+
+//Has the user entered data with length > 0 & <= 256?
+if ((strlen($user_data) > 0) && (strlen($user_data) <= 256)) {
```
tomasbrod: Why the `user_data` can't be empty?

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,148 @@
...
+		$pubkeyid = openssl_pkey_get_public($public_key_path);
+
+		// Base64 encode the generated signature to enable safe output to text file.
+		$base64_sig = base64_encode($signature);
+
+		// Decode base64 sig for use in sig_verification
+		$decoded_sig = base64_decode($base64_sig);
+
+		// Verify that the generated signature against the original data, using the public key.
+		$sig_verification = openssl_verify($message_data, $decoded_sig, $pubkeyid, OPENSSL_ALGO_SHA512);
```
tomasbrod: I do not think it is necessary to verify the signature right after making it. It is one more cryptographic operation on server side. This should be moved behind a condition that can be enabled for debugging/testing and disabled for production.

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,148 @@
...
+
+			// Start outputting xml to the auto-download xml file
+			echo "<boinc_user_id_verification>";
+			global $master_url; // Define global master_url variable for use in xml output.
+			echo "\n <master_url>$master_url</master_url>";
+			echo "\n <userid>$userid</userid>";
+			echo "\n <user_data>$user_data</user_data>";
+			//echo "<cpid>$cpid</cpid>"
+			//echo "<timestamp>time()</timestamp>"
+			echo "\n <msg>$message_data</msg>";
```
tomasbrod: Either remove user_data or msg elements. This is duplicate data. Or it does not matter? Idk.

html/user/openssl_sign_action.php
grctest: The intended user_data for Gridcoin would be the user's beacon public key (or some generated public key), if there was no user_data entered then it would be just proof of account ownership without a specified external purpose which could be dangerous instead of forcing external services from performing some two-way handshake.

Do you think it should accept no user_data input?

html/user/openssl_sign_action.php
grctest: The current performance is near instant, so removing the check doesn't deteriorate user experience. I do agree that it's somewhat unnecessary, however we did experience signature errors with the older openssl version. Commenting it out for debugging/testing wouldn't take much effort to be fair.

html/user/openssl_sign_action.php
grctest: The msg element is hashed, so you can't see the original user_data contents. Fair enough the external service (like the Gridcoin wallet) should know what the intended/original user_data was, but perhaps it'd be easier to import the XML and read the user_data field from there for quick reference/double-checking?

Overall, the performance improvement by removing this data is negligible, the XML download prompt is almost instant after you click the 'submit' button.

html/user/openssl_sign_action.php
tomasbrod: You are right. It does not make sense to sign empty string.

tomasbrod:
There was a suggestion of extending this mechanism to not only verify User ID but also cpid. Therefore I suggest adding CPID into the signed message and also to the output xml, so it can be verified. The message then should be formatted as `userid` *SP* `cpid` *SP* `extdata`. If CPID can be empty/undefind serverside, then a placeholder value must be substituted.
Include the CPID into the output and signature.

html/user/openssl_sign_action.php
tomasbrod: The msg element contains contents of both user_id and user_data elements. So it is kind of redundant there. You removed the digest/hashing step there, or is it?

grctest: Ah, sorry I see what you mean I didn't have the XML downloaded file to hand earlier, yeah only the 'signature' has the signed message, so the 'userid' and 'user_data' rows could be removed, sounds good - I'll removed it then.

The result will be:
```
<boinc_user_id_verification>
 <master_url>http://209.250.242.113/boincserver/</master_url>
 <msg>1 oweijfwoeijfw</msg>
 <signature>C/acKvj9QyY1hL4BQX5Pu9oUDTtowLPbjyYQR5N/G8ooZVzfLTtVuh/CluWYvsfIpftHnvBFFz4UWCp7fike7K2wrM4NmfnQjH1wwnIZNb4CqAXS5awhCLxlo7PoPrVCMeppt+SGAIn2jQ54COiJxzSLdzEVuFlpLaOOJgQnrTepqw8ETzL8FwWry+SzN/CIBPiEjkOnOWZ3n1qWuZkNHrNb+z6lNGtw3yorB6ZRKY+CgDv5ylSHpNtBdv/XOE3fjBhJaRus14mKX8VtIKmrGd5SdcBVXZxZ98q5dpcOHqf7buk0xLW6a8qoEAI9T7slU18hQ40emhyHzhJ82d5ARg==</signature>
</boinc_user_id_verification>
```

grctest:
> Tomasbrod wrote:
> There was a suggestion of extending this mechanism to not only verify User ID but also cpid. Therefore I suggest adding CPID into the signed message and also to the output xml, so it can be verified. The message then should be formatted as userid SP cpid SP extdata. If CPID can be empty/undefind serverside, then a placeholder value must be substituted.
Include the CPID into the output and signature.

We certainly could include the CPID field in the message with relative ease.

There are however a few couple with using CPID over userID:
* CPID can change either accidentally (CPID merger, email change, ...) or maliciously (someone breaking into your account & changing its email address).
* CPID aka Cross Project Identifier - whilst individual projects do share this common CPID for leaderboard coordination, is it appropriate for an individual project to sign this CPID on behalf of all other other projects? Perhaps this would be less agreeable (for project admins & BOINC devs) than verifying userID in case say an user did something malicious on another project and it appears that an external project gets blamed for 'validating' their identity?

You can easily scrape an CPID when you know the user's userID & the userID never changes, so you wouldn't need to revalidate account ownership if the CPID field changes.

If you really want the CPID field, then I would prefer to throw it in a separate signature row & message, to keep CPID/userID seperate & leave the level of verification up to the individual project admin.

grctest:
Anyone have any comments/questions/suggestions regarding this pull request? Cheers 👍

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,148 @@
...
+
+// Needs to be defined, unknown use.
+check_get_args(array("tnow", "ttok"));
+
+// Check the user is online
+$user = get_logged_in_user();
+check_tokens($user->authenticator);
+
+// Optional: Recaptcha support
+//if ($recaptcha_private_key) {
```
ChristianBeer: Since sites like gridcoin most likely directly link to this page, so it would be preferable to enable recaptcha by default if the project already uses it.

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,148 @@
...
+//    if (!boinc_recaptcha_isValidated($recaptcha_private_key)) {
+//        error_page(
+//            tra("Your reCAPTCHA response was not correct. Please try again.")
+//        );
+//    }
+//}
+
+// Input is passed in from the openssl_sign_form
+// sanitize_tags removes html elements from the user input
+// $user_data = htmlspecialchars((post_str("user_data", true)),ENT_XML1);
```
ChristianBeer: Please remove this as it implies that this is a possibility for projects to use. But it would screw the verification process later when the third party site expects tags in the message. Although if you allow tags in the `user_data` and `msg` fields you would need to encapsulate them with a `CDATA` element in the XML output.

To sumarize: either escape html tags in `user_data` (substitute `>` with `&gt;`) or use something like this: `'<![CDATA['.$message_data.']]>'`

html/user/openssl_sign_action.php
```diff
@@ -0,0 +1,148 @@
...
+	// $timestamp = time();
+	// $message_data = "$userid $user_data $timestamp $cpid";
+
+	// How to generate required keys in /project/keys/ folder:
+	// openssl genpkey -algorithm RSA -out external_system_key_private.pem -pkeyopt rsa_keygen_bits:2048
+	// openssl rsa -pubout -in external_system_key_private.pem -out external_system_key_public.pem
+
+	// FAO project admins: Change the following 2 lines to the location of the keys you have personally generated.
+	// If the following keys do not exist, then the users will face (handled) error messages.
+	$private_key_path = "file:///root/project/keys/external_system_key_private.pem";
```
ChristianBeer: This should use the existing key directory specified in `config.xml`. The same place the public code signing key is.

html/user/openssl_sign_action.php
ChristianBeer: Code currently doesn't show the result of your discussion. Also see my comment about tags in `$message_data`.

html/user/openssl_sign_action.php
ChristianBeer: Here is how to get the path out of `config.xml`:
```
$config = get_config();
$keydir = parse_config($config, "<key_dir>");
```

EnzoCaricoTri:
this change is going ahead?

html/user/openssl_sign_action.php
grctest: I've enabled this by default, but when I had it enabled it threw some pretty ugly errors in black text at the top if I got it wrong. I've not got a live demo right now - broke it about a week after the conference, need to re-implement it to check this isn't still the case with the latest BOINC codebase 👍

html/user/openssl_sign_action.php
grctest: I implemented the following which converts special characters to html equivelant:

```$user_data = htmlentities(post_str("user_data", true), ENT_QUOTES, "UTF-8");```

Is that better?

File: html/user/openssl_sign_action.php (hunk 17 above)
Comment (grctest) [reply html/user/openssl_sign_action.php]: https://github.com/BOINC/boinc/pull/2134/files#diff-91d6b04155bcf59e677280a695167c2fR51

Done.

Where is the ideal place for the public to get access to the openssl public key? Up to the project admin or is there an appropriate public download location for the pem file/text? Second copy/paste text area?

File: html/user/openssl_sign_action.php (hunk 14 above)
Comment (grctest) [reply html/user/openssl_sign_action.php]: I've removed the fields discussed above. https://github.com/BOINC/boinc/pull/2134/files#diff-91d6b04155bcf59e677280a695167c2fR111

**Comment by grctest** [comment]
Cleaned up the form a bit, fixing the recaptcha form.

I updated to the latest boinc-server-docker and encountered the following new errors:

> Warning: openssl_sign(): supplied key param cannot be coerced into a private key in /home/boincadm/project/html/user/openssl_sign_action.php on line 80

> Warning: openssl_free_key() expects parameter 1 to be resource, boolean given in /home/boincadm/project/html/user/openssl_sign_action.php on line 83

---

Since openssl version 1.1.0f is bundled with the current boinc-server-docker image, perhaps a newer version won't have this issue?

I'll look into this the next couple days, this was working at the workshop so it hopefully won't be a complex fix 👍

**Comment by ChristianBeer** [comment]
I'll take a look at the recaptcha stuff. This is working in other places and should be straight forward in this new form too.

I would propose to include the content of the public keyfile as a part of what is returned by `html/user/get_project_config.php`. There might be a problem with accessing the file since the webpage is running as user apache and the key directory is outside the web root. I'll take a look at this too to give you some pointers.

For me to be able to merge this PR I can't contribute code since that would need to be reviewed by someone else.

**Comment by ChristianBeer** [comment]
Here is how recaptcha works in BOINC:

1. You need to include: `../inc/util.inc` and `../inc/recaptchalib.php` in each file that calls a recaptcha function
2. Your call to `page_head()` should look like this: `page_head(tra("Title"), null, null, null, boinc_recaptcha_get_head_extra());` this will include the script in the head correctly.
3. The actual form is included like this:
```
global $recaptcha_public_key
if ($recaptcha_public_key) {
    form_general("", boinc_recaptcha_get_html($recaptcha_public_key));
}
```
4. Upon receiving the form you can check if the recaptcha was solved by doing something like this:
```
global $recaptcha_private_key;
if ($recaptcha_private_key) {
        if (!boinc_recaptcha_isValidated($recaptcha_private_key)) {
            show_error(
                tra("Your reCAPTCHA response was not correct. Please try again.")
            );
        }
    }
```

This is the way it was intended to be used and the way it should work. This is not done consistently in BOINC right now.

**Comment by grctest** [comment]
Some extra points:

**Comment by grctest** [comment]
Can we remove the following? 

```
// Needs to be defined, unknown use.
check_get_args(array("tnow", "ttok"));
```

If users attempt to navigate directly to the 'openssl_sign_form' page directly instead of via the hyperlink located within the user's private profile panel they get the following errors:

```
Notice: Undefined index: tnow in /home/boincadm/project/html/inc/util.inc on line 544
Notice: Undefined index: ttok in /home/boincadm/project/html/inc/util.inc on line 547
Warning: Cannot modify header information - headers already sent by (output started at /home/boincadm/project/html/inc/util.inc:544) in /home/boincadm/project/html/inc/util.inc on line 241
Warning: Cannot modify header information - headers already sent by (output started at /home/boincadm/project/html/inc/util.inc:544) in /home/boincadm/project/html/inc/util.inc on line 242
Warning: Cannot modify header information - headers already sent by (output started at /home/boincadm/project/html/inc/util.inc:544) in /home/boincadm/project/html/inc/util.inc on line 243
Warning: Cannot modify header information - headers already sent by (output started at /home/boincadm/project/html/inc/util.inc:544) in /home/boincadm/project/html/inc/util.inc on line 244
Warning: Cannot modify header information - headers already sent by (output started at /home/boincadm/project/html/inc/util.inc:544) in /home/boincadm/project/html/inc/util.inc on line 245
```

---

Related to the fopen permission denied warning, perhaps that error could be caught like:
```
    if ($fp = fopen($private_key_path, "rb"))!==false ) {
        $privkey = fopen($private_key_path, "r"); // Opening private key file
    }
```

**Comment by davidpanderson** [comment]
The tnow/ttok mechanism is there to prevent cross-site scripting attacks.
I'm not sure if that's relevant in this case.

**Comment by grctest** [comment]
I've added a commit which changes how tnow and ttok are checked in the util.inc file, the error messages no longer show when the user navigates directly to the page.

Are there any further changes that anyone would like to request?

**Comment by grctest** [comment]
@ChristianBeer Are there any remaining changes required for this PR to progress? I changed the indentation after your last review. There are changes to the util.inc file which haven't been reviewed perhaps? Thanks

**Comment by ChristianBeer** [comment]
I found some minor documentation issues but other than that it looks fine. I would like to spin up a test project to verify the procedure. I'm not sure when I'll have time (might have to wait until after Christmas).

**Comment by grctest** [comment]
@ChristianBeer The procedure thus far for implementing this is as follows:

```
1. Have the latest stable OpenSSL installed.
2. Install this PR branch's BOINC web server changes
2.1. There are 2 .inc file changes & 2 new PHP files
3. Configure recaptcha for the form
4. Generate OpenSSL keys in the /project/keys/ folder:
         openssl genpkey -algorithm RSA -out external_system_key_private.pem -pkeyopt rsa_keygen_bits:2048
         openssl rsa -pubout -in external_system_key_private.pem -out external_system_key_public.pem
5. Adjust key permissions:
         chown -R boincadm:boincadm ext*
         chmod --reference upload_private external_system_key_public.pem
         chmod --reference upload_private external_system_key_private.pem
```

Step 5 could be adjusted to perform the permission change without referencing upload_private key & there may need to be a final step where the "external_system_key_public.pem" file is shared with the public.

For testing it:
```
1. Create an account on 2134 enabled BOINC server
2. Go to user profile
3. Follow link to create signature
4. Enter a message, complete the recaptcha, submit form.
5. View result
6. Potentially verify message within local terminal
```

Are there any further required changes? Happy new year by the way 🎉

**Comment by ChristianBeer** [comment]
Since I don't have an up to date VM for BOINC and the official one needs updating I decided to create a new VM from scratch and test this PR with it. As a sideeffect I can also update the documentation on what packages to install.

File: html/user/openssl_sign_form.php [hunk 18]
```diff
@@ -0,0 +1,58 @@
...
+check_get_args(array("tnow", "ttok"));
+
+$user = get_logged_in_user();
+check_tokens($user->authenticator);
+
+page_head(tra("Generate UserID ownership signature"), null, null, null, boinc_recaptcha_get_head_extra());
+
+// Verify the user is logged in
+if ($user) {
+    // The user is logged in
```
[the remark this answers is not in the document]
grctest: We check for the public/private keys existence in the [action file](https://github.com/BOINC/boinc/pull/2134/files/29d3e1a612aae10c6a88e38f4526bef76fd5b23e#diff-91d6b04155bcf59e677280a695167c2fR68) however the error messages are generic/informative instead of instructing the user to bother the admin. So you'd prefer the check to be performed in the form rather than in the action?

grctest:
Regarding how to check the signature on the command line, first we need to store the message `1 Enter text` in `message_text.txt`, then we decode the signature from base64 to its original format:

```
base64 -d base64_signature_text.txt > decoded_signature_text.txt
```

Then we need to verify the contents:
```
openssl dgst -sha512 \
  -verify key.pem \
  -signature decoded_signature_text.txt \
  message_text.txt
```

Which fails to verify: `Verification Failure`

This is strange because the following line passes successfully: `$sig_verification = openssl_verify($message_data, $decoded_sig, $pubkeyid, OPENSSL_ALGO_SHA512);`

I'm using openssl version 1.1.0g on my test VPS, which are you running on your test server? Perhaps there are differences between the implementations of OpenSSL?

html/user/openssl_sign_form.php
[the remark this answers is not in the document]
ChristianBeer: If the user sees the form and the button to generate the signature is enabled that suggests that the system is working. Getting a generic error message after entering the data copied from an external site leads to frustration on the users side.

ChristianBeer:
I'm using `OpenSSL 1.1.0j  20 Nov 2018` on Debian 9. If this turns out to be bound to a specific OpenSSL version we can't use it since we can't pin the project owner on a specific old version of this crucial library. I suspect that an additional end of line character in the textfile is making the verification fail.

ChristianBeer:
I was able to output the public key in `user/get_project_config.php` with this code at the end:
```
$keydir = parse_config($config, "<key_dir>");
if (is_readable($keydir."/ownership_sign_public.pem")) {
    echo "<ownership_signature_public_key>";
    echo base64_encode(file_get_contents($keydir."/ownership_sign_public.pem"));
    echo "</ownership_signature_public_key>\n";
}
```

grctest:
I've successfully verified the output signature using the following python code:

```
from Crypto.PublicKey import RSA # package: pycryptodome
from Crypto.Signature import PKCS1_v1_5 
from Crypto.Hash import SHA512
from base64 import b64decode
import xmltodict # For handling XML data

with open('xml_data.xml') as fd: # Entire output XML snippet
    boinc_xml_data = xmltodict.parse(fd.read())

with open('key.pem') as f: # public key *.pem file
    public_key_data = f.read()

message = boinc_xml_data['boinc_user_id_verification']['msg']
signature = boinc_xml_data['boinc_user_id_verification']['signature']

rsakey = RSA.importKey(public_key_data) 
verifier = PKCS1_v1_5.new(rsakey)
digest = SHA512.new()
digest.update(bytes(message, encoding = "utf8"))

if verifier.verify(digest, b64decode(signature)):
    print("Successfully verified UserID ownership.")
else:
    print("Failed to verify UserID ownership.")
```

I agree with your naming convention change requests, that makes sense. I'll merge in your proposed changes in the next couple hours, thanks for helping 👍

ChristianBeer:
I took another look at my commandlines from yesterday and found the problem. I tried to verify using SHA256 instead of SHA512. I just did it again using the correct hashing algorithm and it worked. That's very comforting.

Also to be clear I would prefer the naming of `ownership signature` to distinguish it from user signatures (in forums). The term `userid signatures` are just too close to be mistaken.

grctest:
I've made the requested changes, however the following warnings are showing within my test project:

```
Warning: Cannot modify header information - headers already sent by (output started at /home/boincadm/project/html/inc/user.inc:579) in /home/boincadm/project/html/inc/util.inc on line 241
Warning: Cannot modify header information - headers already sent by (output started at /home/boincadm/project/html/inc/user.inc:579) in /home/boincadm/project/html/inc/util.inc on line 242
Warning: Cannot modify header information - headers already sent by (output started at /home/boincadm/project/html/inc/user.inc:579) in /home/boincadm/project/html/inc/util.inc on line 243
Warning: Cannot modify header information - headers already sent by (output started at /home/boincadm/project/html/inc/user.inc:579) in /home/boincadm/project/html/inc/util.inc on line 244
Warning: Cannot modify header information - headers already sent by (output started at /home/boincadm/project/html/inc/user.inc:579) in /home/boincadm/project/html/inc/util.inc on line 245
```

And within the get_project_config.php output the following error at the top:
```
 Warning: Cannot modify header information - headers already sent by (output started at /home/boincadm/project/html/inc/user.inc:579) in /home/boincadm/project/html/inc/xml.inc on line 53
```

Is this an environment issue, or related to my changes?

AenBleidd:
You have conflicts in your branch also. Looks like because of your changes...

ChristianBeer:
I'm going to look into the issue with the warnings on Sunday and would also like to advise you to rebase this branch and squash the commits so they become coherent again as a lot of things have changes since the first commit.

ChristianBeer:
Regarding the warning messages: I found the reason and it has nothing to do with this PR. I documented it in #2960.

I just found a nice guide that describes the rebase and squash workflow (https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/) when I saw that you already did the rebase. A big part of a rebase is that afterwards you have a clean history and clean commits that can stand for themselves. That is kind of not the case because of the way you did the rebase and what you used as a commit message. Can you please fix that?

In the end there should be only one commit in this PR that contains a description of the feature that is implemented in the commit. There should be no reference to any squashing or rebasing or merges that happened in an external repository. My rationale for this is that I often find myself looking at commit messages from the past that don't explain why the change was made and what the intention was. This change has a potentially big impact and I want to make sure that someone 2 or 3 years in the future can still see from the commit message why this change was made and what the intention was. For this reason I would like to Veto this PR in it's current state (the code is fine and I already tested it again).

grctest:
Sorry, I had to create a new PR to produce the requested single commit PR: https://github.com/BOINC/boinc/pull/2965

grctest:
Closing because this PR is unneccessary, #2965 is the active PR now 👍

PR #2135, opened 2017-09-20
Client: gather libc version on Linux

### LINKED ISSUE #2064: add Linux libc version info to host.os_version field

ChristianBeer:
Try to execute ldd on Linux like systems to gather the libc version installed and report it as part of host.os_version using the same pipe delimited format that is also used in host.serialnum.
A sample entry looks like this: "Debian GNU/Linux 9.1 (stretch) [4.9.0-3-amd64|libc 2.24 (2.24-11+deb9u1)]" where the OS version and codenames, kernel version and libc version information are accessible for use in plan_class restrictions or help in troubleshooting problems by project admins.

closes #2064

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1367,6 +1367,45 @@ int HOST_INFO::get_memory_info() {
     return 0;
 }
 
+#if LINUX_LIKE_SYSTEM
+// gather libc version from the system installed ldd binary
+// a usual first line looks like: ldd (Debian GLIBC 2.24-11+deb9u1) 2.24
+// where the last part is copied to version and the string between parenthesis
+// is copied to extra_info
+//
+// return 0 if at least version could be found (extra_info may remain empty)
+// return 1 if ldd couldn't be opened or no version infromation was found
+//
+int get_libc_version(string& version, string& extra_info) {
+    char buf[1024];
+    string strbuf;
+    FILE* f = popen("PATH=/usr/bin:/bin:/usr/local/bin ldd --version 2>&1", "r");
+    if (f) {
+        fgets(buf, sizeof(buf), f);
+        pclose(f);
+        strbuf = (string)buf;
+        strip_whitespace(strbuf);
+        std::string::size_type parens1 = strbuf.find('(');
+        std::string::size_type parens2 = strbuf.rfind(')');
+        std::string::size_type blank = strbuf.rfind(' ');
+
+        if (blank != std::string::npos) {
+            // extract version number
+            version = strbuf.substr(blank+1);
+        } else {
+            return 1;
+        }
+        if (parens1 != std::string::npos && parens2 != std::string::npos && parens1 < parens2) {
+            // extract extra information without parenthesis
+            extra_info = strbuf.substr(parens1+1, parens2-parens1-1);
+        }
+    } else {
+        return 1;
+    }
+    return 0;
+}
+#endif
+
 // get os_name, os_version
 //
 int HOST_INFO::get_os_info() {
@@ -1411,6 +1450,7 @@ int HOST_INFO::get_os_info() {
     bool found_something = false;
     char buf[256],buf2[256];
     char dist_pretty[256], dist_name[256], dist_version[256], dist_codename[256];
+    string os_version_extra("");
     strcpy(dist_pretty, "");
     strcpy(dist_name, "");
     strcpy(dist_version, "");
@@ -1521,14 +1561,33 @@ int HOST_INFO::get_os_info() {
             }
             strip_whitespace(buf2);
         }
-        strcat(buf2, " [");
-        safe_strcat(buf2, os_version);
-        strcat(buf2, "]");
+        os_version_extra = (string)os_version;
         safe_strcpy(os_version, buf2);
         if (strlen(dist_name)) {
             strcat(os_name, " ");
             safe_strcat(os_name, dist_name);
         }
+    }
+
+    string libc_version(""), libc_extra_info("");
+    if (!get_libc_version(libc_version, libc_extra_info)) {
+        msg_printf(NULL, MSG_INFO,
+                "[libc detection] gathered: %s, %s", libc_version.c_str(), libc_extra_info.c_str()
+            );
+        // add info to os_version_extra
+        if (!os_version_extra.empty()) {
+            os_version_extra += "|";
+        }
+        os_version_extra += "libc " + libc_version;
+        if (!libc_extra_info.empty()) {
+            os_version_extra += " (" + libc_extra_info + ")";
+        }
+    }
+
+    if (!os_version_extra.empty()) {
+        strcat(os_version, " [");
+        strcat(os_version, os_version_extra.c_str());
+        strcat(os_version, "]");
     }
 #endif //LINUX_LIKE_SYSTEM
     return 0;
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1377,15 +1377,19 @@ int HOST_INFO::get_memory_info() {
 // return ERR_NOT_FOUND if ldd couldn't be opened or no version information was found
 //
 int get_libc_version(string& version, string& extra_info) {
-    char buf[1024];
+    char buf[1024] = "";
     string strbuf;
     FILE* f = popen("PATH=/usr/bin:/bin:/usr/local/bin ldd --version 2>&1", "r");
     if (f) {
-        fgets(buf, sizeof(buf), f);
-        if (0 > pclose(f)) {
+        char* retval = fgets(buf, sizeof(buf), f);
+        strbuf = (string)buf;
+        while (fgets(buf, sizeof(buf), f)) {
+            // consume output to allow command to exit gracefully
+        }
+        int status = pclose(f);
+        if (!retval || status == -1 || !WIFEXITED(status) || WEXITSTATUS(status)) {
             return ERR_NOT_FOUND;
         }
-        strbuf = (string)buf;
         strip_whitespace(strbuf);
         string::size_type parens1 = strbuf.find('(');
         string::size_type parens2 = strbuf.rfind(')');
@@ -2221,4 +2225,3 @@ bool HOST_INFO::users_idle(bool check_all_logins, double idle_time_to_run) {
 }
 
 #endif  // ! __APPLE__
-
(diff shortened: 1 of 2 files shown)


ChristianBeer:
As a note to reviewers: `hostinfo_unix_test.cpp` is not a unittest. It is a standalone program that can be compiled without BOINC to test the detection code on new platofrms without a Client present. It is best compiled with `g++ -o hostinfo_unix_test hostinfo_unix_test.cpp`.

It would be nice to switch to a design that would allow to generate this standalone binary without needing to copy the code from `hostinfo_unix.cpp` and probably break some other OS specific stuff into separate files.

client/hostinfo_unix.cpp
```diff
@@ -1584,6 +1584,45 @@ int HOST_INFO::get_memory_info() {
...
+#if LINUX_LIKE_SYSTEM
+// gather libc version from the system installed ldd binary
+// a usual first line looks like: ldd (Debian GLIBC 2.24-11+deb9u1) 2.24
+// where the last part is copied to version and the string between parenthesis
+// is copied to extra_info
+//
+// return 0 if at least version could be found (extra_info may remain empty)
+// return 1 if ldd couldn't be opened or no version infromation was found
+//
+int get_libc_version(string& version, string& extra_info) {
```
JuhaSointusalo: I'm no big fan of BOINC's error codes but those would be nicer than magic number. BOINC_SUCCESS and ERR_NOT_FOUND or something.

client/hostinfo_unix.cpp
```diff
@@ -1584,6 +1584,45 @@ int HOST_INFO::get_memory_info() {
...
+    char buf[1024];
+    string strbuf;
+    FILE* f = popen("PATH=/usr/bin:/bin:/usr/local/bin ldd --version 2>&1", "r");
+    if (f) {
+        fgets(buf, sizeof(buf), f);
+        pclose(f);
+        strbuf = (string)buf;
+        strip_whitespace(strbuf);
+        std::string::size_type parens1 = strbuf.find('(');
```
JuhaSointusalo: Nitpick. There's `using std::string` at the top of the file. I think not having `std::` would be a bit more readable.

client/hostinfo_unix.cpp
```diff
@@ -1738,15 +1778,34 @@ int HOST_INFO::get_os_info() {
...
         if (strlen(dist_name)) {
             strcat(os_name, " ");
             safe_strcat(os_name, dist_name);
         }
     }
+
+    string libc_version(""), libc_extra_info("");
+    if (!get_libc_version(libc_version, libc_extra_info)) {
+        msg_printf(NULL, MSG_INFO,
+                "[libc detection] gathered: %s, %s", libc_version.c_str(), libc_extra_info.c_str()
```
JuhaSointusalo: This looks like debug code.

client/hostinfo_unix.cpp
```diff
@@ -1584,6 +1584,45 @@ int HOST_INFO::get_memory_info() {
...
+// return 0 if at least version could be found (extra_info may remain empty)
+// return 1 if ldd couldn't be opened or no version infromation was found
+//
+int get_libc_version(string& version, string& extra_info) {
+    char buf[1024];
+    string strbuf;
+    FILE* f = popen("PATH=/usr/bin:/bin:/usr/local/bin ldd --version 2>&1", "r");
+    if (f) {
+        fgets(buf, sizeof(buf), f);
+        pclose(f);
```
JuhaSointusalo: You need to check return value here. If `ldd` doesn't exist `pclose` returns an error code and `buf` will contain something like `sh: 1: ldd: not found`.

client/hostinfo_unix.cpp
ChristianBeer: You're right. I'm going to change it to use BOINC_SUCCESS and ERR_NOT_FOUND.

ChristianBeer:
@RichardHaselgrove : I'm not sure why I never made the changes but I would be willing to give it a quick pat so it can be included in 7.10 would that work?

RichardHaselgrove:
Yes, I think we have a couple of days left for v7.10, if the loose ends could be tidied up?

client/hostinfo_unix.cpp
ChristianBeer: Thanks, fixed in 11b5569

client/hostinfo_unix.cpp
ChristianBeer: Thanks, fixed in 11b5569

client/hostinfo_unix.cpp
ChristianBeer: It started out to be put since the hostinfo detection only runs on startup this is a nice addition to the default startup messages I think so I left it in and added a comment that it was left on purpose.

ChristianBeer:
Done. Ready for another round of review by @JuhaSointusalo.

client/hostinfo_unix.cpp
```diff
@@ -1584,6 +1584,47 @@ int HOST_INFO::get_memory_info() {
...
+// return BOINC_SUCCESS if at least version could be found (extra_info may remain empty)
+// return ERR_NOT_FOUND if ldd couldn't be opened or no version information was found
+//
+int get_libc_version(string& version, string& extra_info) {
+    char buf[1024];
+    string strbuf;
+    FILE* f = popen("PATH=/usr/bin:/bin:/usr/local/bin ldd --version 2>&1", "r");
+    if (f) {
+        fgets(buf, sizeof(buf), f);
+        if (0 > pclose(f)) {
```
JuhaSointusalo: No, still doesn't work right. Without `ldd`:
```
30-Jan-2018 23:33:09 [---] [libc detection] gathered: found,
30-Jan-2018 23:33:09 [---] OS: Linux LinuxMint: Linux Mint 17.3 Rosa [3.13.0-123-generic|libc found]
```
`pclose` returns `-1` only in case of internal error, otherwise it's the exit status as in `waitpid` and needs to be examined with `W*` macros.

Even with `ldd` in place the exit status is `WIFEXITED=1` and `WEXITSTATUS=141`. I don't know why that is. Maybe because you close the pipe before reading everything.

JuhaSointusalo: Yep. `WEXITSTATUS=141` came from closing the pipe too early. This seems to work:
```
 int get_libc_version(string& version, string& extra_info) {
-    char buf[1024];
+    char buf[1024] = "";
     string strbuf;
     FILE* f = popen("PATH=/usr/bin:/bin:/usr/local/bin ldd --version 2>&1", "r");
     if (f) {
-        fgets(buf, sizeof(buf), f);
-        if (0 > pclose(f)) {
+        char* retval = fgets(buf, sizeof(buf), f);
+        strbuf = buf;
+        while (fgets(buf, sizeof(buf), f)) {
+            // consume output to allow command to exit gracefully
+        }
+        int status = pclose(f);
+        if (!retval || status == -1 || !WIFEXITED(status) || WEXITSTATUS(status)) {
             return ERR_NOT_FOUND;
         }
-        strbuf = (string)buf;
         strip_whitespace(strbuf);
```

Or you could have the code check that the first line matches the expected format.

ChristianBeer: I suspected that closing the pipe early is going to cause problems but I hadn't had time to look into it. Thanks for testing. I'm going to adapt your code into this PR.

ChristianBeer:
My editor fixed a lot of whitespace stuff in the two files I touched with this PR which I wanted to include but got a merge conflict. So I ended up rebasing on current master and adding the whitespace changes on top. Fixing whitespace one file at a time seems better than one big whitespace PR to me.

client/hostinfo_unix.cpp
RichardHaselgrove: I'm having difficulty following github's intertwined changes and reviews, but it looks to me as if the latest change is 5b94cf8defeca4e157ae8c92c4e05c797f042fc2 two days ago (which doesn't show in this timeline). @JuhaSointusalo, does this satisfy your most recent change request?

Issue #1083, opened 2015-02-04
GUI Glitch - Actions on 2+ selected rows appear to leave every other row unselected


romw:
**Reported by JacobKlein on 8 Jun 41333780 03:33 UTC**
If 2+ rows are selected, and an action is performed (like "Show Graphics"), the UI then shows only every other row selected, but actions indicate the initial rows are still selected (like "Properties" is greyed out)

Another example would be clicking "Copy" when having several consecutive rows selected in the Event Log.

An even more interesting case would be clicking "Copy" when having several random rows selected in the Event Log.

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1107

ChristianBeer:
Does not happen in current 7.6.33

AenBleidd:
@ChristianBeer Please reopen this issue. It still can be reproduced:
1 - Select several strings
2 - Click 'Cope selected' button

![image](https://cloud.githubusercontent.com/assets/3234209/24807056/735d8072-1bbf-11e7-8ded-715c1aafbc6f.png)

Ageless93:
But if you then actually paste the contents of clipboard into a file,
you'll see that all the lines are there.
So it's just an optical illusion, it's not as if only the coloured lines
have been copied. All the ones you selected previously have been actually
copied.


-- Jord van der Elst.

ChristianBeer:
I assume you are on Windows. I just tried it again and this does not happen on Linux. Reopening either way as it should be fixed. Even if it is only an optical illusion.

AenBleidd:
@Ageless93 You're right but it is a GUI glitch still.
@ChristianBeer Yes, I've reproduced it on Win7/8.1/10 so this is a common wxWidgets Windows problem I guess

CharlieFenton:
I just tested this on 7.6.33 and it does not happen on a Mac, though I can easily reproduce it on MS Windows 7.

Ageless93:
Any progress on this?

AenBleidd:
@Ageless93, no progress now. I'll try to build BOINC with updated wxWidgets to see whether this is still an issue. If it is I'll try to fix it either on BOINC side or on wxWidgets side.

AenBleidd:
Fixed in #2137

Ageless93:
Tested with BOINC Manager built from Master. It's fixed in that one.

Ageless93:
Although now the effect is:
- select rows.
- do action (like suspend)
[...]

PR #2137, opened 2017-09-22
Manager: Fixed EventLog window GUI glitch with selected unfocused rows.

AenBleidd:
As requested in #1083: GUI Glitch - Actions on 2+ selected rows appear to leave every other row unselected

### DIFF
--- a/clientgui/DlgEventLog.cpp
+++ b/clientgui/DlgEventLog.cpp
@@ -251,14 +251,24 @@ bool CDlgEventLog::Create( wxWindow* parent, wxWindowID id, const wxString& capt
 #if EVENT_LOG_STRIPES
     m_pMessageInfoGrayAttr = new wxListItemAttr(
         wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWTEXT),
-        wxColour(240, 240, 240),
+        wxColour(244, 244, 244),
         wxNullFont
     );
-    m_pMessageErrorGrayAttr = new wxListItemAttr(*wxRED, wxColour(240, 240, 240), wxNullFont);
+    m_pMessageErrorGrayAttr = new wxListItemAttr(*wxRED, wxColour(244, 244, 244), wxNullFont);
 #else
     m_pMessageInfoGrayAttr = new wxListItemAttr(*m_pMessageInfoAttr);
     m_pMessageErrorGrayAttr = new wxListItemAttr(*m_pMessageErrorAttr);
 #endif
+    m_pMessageInfoSelectedNonFocusedAttr = new wxListItemAttr(
+        wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWTEXT),
+        wxColour(242, 242, 242),
+        wxNullFont
+    );
+    m_pMessageErrorSelectedNonFocusedAttr = new wxListItemAttr(
+        *wxRED,
+        wxColour(242, 242, 242),
+        wxNullFont
+    );
 
     SetTextColor();
     RestoreState();
@@ -989,10 +999,10 @@ wxListItemAttr* CDlgEventLog::OnListGetItemAttr(long item) const {
         item += s_bIsFiltered ? m_iTotalDeletedFilterRows : m_iPreviousFirstMsgSeqNum;
         switch(message->priority) {
         case MSG_USER_ALERT:
-            pAttribute = (item % 2 && !selected) ? m_pMessageErrorGrayAttr : m_pMessageErrorAttr;
+            pAttribute = selected ? m_pMessageErrorSelectedNonFocusedAttr : (item % 2 ? m_pMessageErrorGrayAttr : m_pMessageErrorAttr);
             break;
         default:
-           pAttribute = (item % 2 && !selected) ? m_pMessageInfoGrayAttr : m_pMessageInfoAttr;
+           pAttribute = selected ? m_pMessageInfoSelectedNonFocusedAttr : (item % 2 ? m_pMessageInfoGrayAttr : m_pMessageInfoAttr);
             break;
         }
     }
--- a/clientgui/DlgEventLog.cpp
+++ b/clientgui/DlgEventLog.cpp
@@ -126,6 +126,16 @@ CDlgEventLog::~CDlgEventLog() {
         m_pMessageErrorGrayAttr = NULL;
     }
 
+    if (m_pMessageInfoSelectedNonFocusedAttr) {
+        delete m_pMessageInfoSelectedNonFocusedAttr;
+        m_pMessageInfoSelectedNonFocusedAttr = NULL;
+    }
+
+    if (m_pMessageErrorSelectedNonFocusedAttr) {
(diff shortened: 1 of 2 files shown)


AenBleidd:
@JuhaSointusalo, could you please verify this? Thanks

JuhaSointusalo:
Sorry for taking this long to review this PR.

I would have liked a bit more text explaining what caused the problem, why it leaves stripes blue and changes non stripes to grey, why similar lists work in project and task views, and how your change fixed the problem.

I played with the code a bit and I think this change is the one that fixes the colours for non focused rows:

```
 #if EVENT_LOG_STRIPES
      m_pMessageInfoGrayAttr = new wxListItemAttr(
          wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWTEXT),
 -        wxColour(240, 240, 240),
 +        wxColour(244, 244, 244),
          wxNullFont
      );
 -    m_pMessageErrorGrayAttr = new wxListItemAttr(*wxRED, wxColour(240, 240, 240), wxNullFont);
 +    m_pMessageErrorGrayAttr = new wxListItemAttr(*wxRED, wxColour(244, 244, 244), wxNullFont);
  #else
```

At least on my system `240,240,240` happens to be the same colour as non focused selection. I'm not going to go though the list control paint code to verify but I'm guessing there is some clever code that tries to avoid redrawing unchanged rows but not clever enough to take selections into account.

I think your first commit accidentally fixed the colours by changing the colours of selected stripes and made the rows look like they had changed. wxWidgets doesn't allow programmers to choose selection colours and because of that the attributes you added should actually do nothing.

So I think the correct way to fix this is to make the changes to stripe colours and remove all other changes. But before you do that wait a bit, lets get the colours exactly right.

JuhaSointusalo:
Something like this:

```
#if EVENT_LOG_STRIPES
    m_pList->EnableAlternateRowColours();

#if wxCHECK_VERSION(3,1,0)
    wxColour stripe = m_pList->GetAlternateRowColour();
#else
    wxColour stripe = m_pList->wxListCtrl::OnGetItemAttr(1)->GetBackgroundColour();
#endif

    m_pMessageInfoGrayAttr = new wxListItemAttr(
        wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWTEXT),
        stripe,
        wxNullFont
    );
    m_pMessageErrorGrayAttr = new wxListItemAttr(
        *wxRED,
        stripe,
        wxNullFont
    );
#else
    m_pMessageInfoGrayAttr = new wxListItemAttr(*m_pMessageInfoAttr);
    m_pMessageErrorGrayAttr = new wxListItemAttr(*m_pMessageErrorAttr);
#endif
```

The thinking I have is that by using the same stripe colour as wxWidgets would have used has the best chances of working right. The wx < 3.1 code abuses implementation details a bit but it's not like there are going to be any changes now. [source](https://github.com/wxWidgets/wxWidgets/blob/WX_3_0_BRANCH/src/common/listctrlcmn.cpp#L218)

AenBleidd:
Hello @JuhaSointusalo, thank you for testing.

> I would have liked a bit more text explaining what caused the problem

This happens because we use custom colours to draw the table. We use custom colours to draw the table because we use filtering.

> why it leaves stripes blue and changes non stripes to grey

Because we draw every second line to enable zebra-style colouring.

> why similar lists work in project and task views

Because there we don't use custom colours to draw the table

>  how your change fixed the problem.

I draw now not focused line too

> I played with the code a bit and I think this change is the one that fixes the colours for non focused rows:

No. Here I change the colour to be the same as on the other grids: now if you select some rows and change the focus to another control you will see different colours on selected lines and not selected lines. Before this change these lines were the same colour.

> I think your first commit accidentally fixed the colours by changing the colours of selected stripes and made the rows look like they had changed. 

My first commit actually fixes the problem. Last 2 commits just to adjust colours with other grids.

> wxWidgets doesn't allow programmers to choose selection colours and because of that the attributes you added should actually do nothing.

Ok you can test this. For me without those lines I have the same GUI glitch.

> So I think the correct way to fix this is to make the changes to stripe colours and remove all other changes. 

Unfortunately there are no unnecessary changes.

> But before you do that wait a bit, lets get the colours exactly right.

Yes, please check colours. I check them on Win10 with colour picker tool to identify the colour.

AenBleidd:
`#if wxCHECK_VERSION(3,1,0)
    wxColour stripe = m_pList->GetAlternateRowColour();
#else
    wxColour stripe = m_pList->wxListCtrl::OnGetItemAttr(1)->GetBackgroundColour();
#endif`

If I remember correct we are not supporting wxWidgets 3.1 right now so I didn't check this on wxWidgets 3.1

JuhaSointusalo:
> Before this change these lines were the same colour.

The colours used for stripes and non focused selection are indeed very similar. FWIW, wx 3.1 uses themed list control and there it's easier to see difference between stripes and selections.

> My first commit actually fixes the problem. Last 2 commits just to adjust colours with other grids.

Could you test only the colour changes?

> Ok you can test this. For me without those lines I have the same GUI glitch.

I did test it. I set SelectedNonFocused to yellow and green background colour but I didn't get yellow and green on screen.

> Yes, please check colours. I check them on Win10 with colour picker tool to identify the colour.

I used Snipping Tool + Paint's Colour picker to check the colours. Task list has stripes `247,247,247` and non focused selection `240,240,240`.

> If I remember correct we are not supporting wxWidgets 3.1 right now so I didn't check this on wxWidgets 3.1

Google takes me more often to wxWidgets development docs than it takes to 3.0 docs. I noticed `GetAlternateRowColour` there and though why not use it.

AenBleidd:
Ok. I check this now.

AenBleidd:
@JuhaSointusalo, I am very sorry. You're completely right. I clean-up the code. Thank you.

AenBleidd:
` wxColour stripe = m_pList->wxListCtrl::OnGetItemAttr(1)->GetBackgroundColour();`

This code fails with null-reference exception. So I need to think about the different method to get alternate row colour.

AenBleidd:
@JuhaSointusalo,
I think you code is much better in this case.
Please create pull request with your changes and close this as outdated.
Thanks

JuhaSointusalo:
I can make a new PR but I don't have the powers to close this one.

AenBleidd:
@JuhaSointusalo, Ok, I'll close this by myself. Thanks

AenBleidd:
@JuhaSointusalo, do you plan to make a PR?

JuhaSointusalo:
@AenBleidd It's on the to-do pile. I thought it might be better to check
that the stripe colour won't be one that causes problems and if it is then
slightly adjust it. And I think it would be responsible thing to do to
reduce the code to a simple test case and submit it to wxWidgets developers
so that they can fix the code.

I'm going to give the VBox PR yet another look first and this I think will
be in turn next week.

AenBleidd:
@JuhaSointusalo, ok, great. I just want to be sure that it is not forgotten :-) Thank you

AenBleidd:
I close this one pull request and open more clean #2187

PR #2138, opened 2017-09-22
validator: raise the quorum for 'suspicious' results to ensure validation

bema-aei:
If a validator (in init_result()) marks a result as 'suspicious', it is replicated. However with the the quorum still being '1' (as it is in 'adaptive replication'), the replication by itself doesn't ensure that the result really gets successfully verified by another result. So in the presence of a 'suspicious' result, raise the quorum to ensure that.

### DIFF
--- a/sched/validate_util2.cpp
+++ b/sched/validate_util2.cpp
@@ -116,6 +116,16 @@ int check_set(
         good_results += suspicious_results;
     }
 
+    // if there are "suspicious" results and min_quorum < g_app->target_nresults
+    // (i.e. adaptive replication), raise min_quorum to g_app->target_nresults
+    // the abs() is there for Einstein@Home-specific use of app->target_nresults,
+    // please leave it in there.
+    //
+    if (suspicious_results && wu.min_quorum < abs(g_app->target_nresults)) {
+        log_messages.printf(MSG_NORMAL, "suspicious result - raising quorum\n");
+        wu.min_quorum = abs(g_app->target_nresults);
+    }
+
     if (good_results < wu.min_quorum) goto cleanup;
 
     // Compare results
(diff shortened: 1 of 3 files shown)


TheAspens:
@bema-ligo - do you use negative values at E@H to trigger different behavior (for feature that are not yet merged into BOINC)?  The use of the abs() is not intuitive and I'd like to see if we could figure out a way to make it clear why that is being used and that will make it easy to merge your code in over time.

bema-ligo:
@TheAspens indeed we have a feature to explicitly "trust" hosts from certain (configurable) users, and this feature hasn't been committed upstream yet. With target_nresults set to negative values (e.g. -2), replication depends solely on this "trusted users", i.e. host from trusted user => replication=1 else replication=2.

AenBleidd:
Is there is something that block this PR from being merged?

lfield:
@TheAspens is there anything else from you. It looks fine for me.

davidpanderson:
Let me look at this

ChristianBeer:
I'll give this another week for others to review and then I'm going to merge it.

davidpanderson:
What does it means for a result to be "suspicious"?  When does this arise?

ChristianBeer:
A "suspicious" result could not be marked by the validator to be good or bad without comparing it to a second result. This arises when the validator checks that the result is within the expected value ranges but by itself is not conclusive enough to be a good result. Thus the validator wants to have a second opinion and marks the result suspicious so a new task is created, sent to another user or host and validated again with the former suspicious result.

Whether a result can be marked "suspicious" or not depends on the project specific science and allowed error margins of values in the results.

This mechanism avoids sending out not needed work as a very strict validator would mark a "suspicious" result as bad then create two now jobs that validate against each other but also would validate against the first result.

bema-aei:
@davidpanderson as explained in #1635, in adaptive replication a validator can mark an unreplicated result as "suspicious" depending on own criteria. This will lead to the result being replicated and run again for validation by comparison.

PR #2142, opened 2017-09-23
client: fix writing log messages

### LINKED ISSUE #2141: client doesn't always write log messages promptly
Normally log messages appear in stdoutdae.txt right after the client has written a message there.

After 6b67fe61528356f7ebe9aa6e6060b1cd91a71d2f this doesn't work correctly.

Now when the log file has grown beyond its maximum size the client cycles the log file and after that messages are written to the file in larger chunks.

Additionally, in MinGW builds the messages are always written in larger chunks.

JuhaSointusalo:
**lib: set correct buffering mode after cycling logs**

After freopen() stream is fully buffered. In this mode client's log messages are written to stdoutdae.txt BUFSIZ (typically 512) bytes at a time.

Set stdout to line buffered mode after freopen() so that messages are written one line at a time.

Set stderr's buffering mode after making sure freopen() succeeded.

**client: explicitly flush stdout on all Windows builds**

stdout is set to line buffered mode but MSVCRT doesn't support this mode and uses fully buffered mode instead. MinGW uses MSVCRT but the code doesn't take this into account and as a result stdoutdae.txt is 
written BUFSIZ (typically 512) bytes at a time on MinGW builds.

Correct this by flushing stdout after every message regardless of which Windows compiler is used.

Fixes #2141.

### DIFF
--- a/client/client_msgs.cpp
+++ b/client/client_msgs.cpp
@@ -125,7 +125,7 @@ void show_message(
     // print message to the console
     printf("%s", evt_message);
 
-#ifdef _MSC_VER
+#ifdef _WIN32
     // MSVCRT doesn't support line buffered streams
     fflush(stdout);
 #endif
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -83,7 +83,7 @@ void log_message_startup(const char* msg) {
     );
     if (!gstate.executing_as_daemon) {
         fprintf(stdout, "%s", evt_msg);
-#ifdef _MSC_VER
+#ifdef _WIN32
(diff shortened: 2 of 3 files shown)


JuhaSointusalo:
Even though the commit message talks about MinGWn I didn't actually observe the bug in MinGW built client. This is because MinGW built client crashes on the first `printf`.

Instead I confirmed the bug with a test app that writes messages to a file and then monitoring the file's growth. The fix I confirmed by verifying that MinGW pre-defines `_WIN32` macro by running `gcc -E -dD empty.c` where `empty.c` is an empty file.

PR #2144, opened 2017-09-25
Drupal: Modified private message module to use BOINC username (and id).

drshawnkwang:
Private message module will now use the BOINC username (display name) in the "to" field, in the autocomplete, and in sending the email notification.
Added custom email body and other variables to private_messages Feature via strongarm.
Added field to User Profile which displays BOINC ID.

https://dev.gridrepublic.org/browse/DBOINCP-254


drupal/sites/default/boinc/modules/contrib/privatemsg/privatemsg.module
```diff
@@ -1422,11 +1428,13 @@ function privatemsg_user_name_autocomplete($string) {
...
   if (!empty($fragment)) {
     $query = _privatemsg_assemble_query('autocomplete', $fragment, $names);
+    db_set_active('boinc');
     $result = db_query_range($query['query'], $fragment, 0, 10);
+    db_set_active('default');
     $prefix = count($names) ? implode(", ", $names) .", " : '';
     // 3: Build proper suggestions and print.
     while ($user = db_fetch_object($result)) {
-      $matches[$prefix . $user->name .", "] = $user->name;
+      $matches[$prefix . $user->name . "_" . $user->id . ", "] = $user->name . ' - BOINC id = ' . $user->id;
```
tristanolive: A thought on formatting of the user name strings: I think readability of the suggestions list takes a hit with this change. What about just putting the BOINC ID number in parentheses following the user name and then adding a note to the form, if needed, to explain what the number is and where to validate it?

tristanolive: Along with this, the user name selected via the list also looks a bit messy with the "_[id]" suffix. This may be a limitation of the generic text box approach that the private message module uses, so that could be an improvement for another day. In general, custom templates should be created for this PM stuff rather than overriding the contrib module, but I'd say all of that can be addressed sometime in the future. Everything here is working fine as it is.

drupal/sites/default/boinc/modules/boincuser/boincuser.module
```diff
@@ -2036,3 +2036,22 @@ function boincuser_apachesolr_index_documents_alter(array &$documents, $entity,
...
+
+/**
+ * Implementation of hook_privatemsg_name_lookup();
+ */
+function boincuser_privatemsg_name_lookup($string) {
+  //Get the BOINC ID from the name string, and lookup the
+  //corresponding drupal user.
+  $boincid = substr($string, strrpos($string, '_') + 1);
+  $drupalid = get_drupal_id($boincid);
+  dpm($drupalid, 'drupalid');
```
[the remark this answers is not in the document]
drshawnkwang: Thanks, fixed with commit 35a7f27.

drupal/sites/default/boinc/modules/contrib/privatemsg/privatemsg.module
drshawnkwang: Unfortunately the boincusername_boincid construct with an underscore is necessary. This is because the recipient of the message is determined by a string, and the boincname is non-unique. Thus I need to append the boincid to the string so boincuser_privatemsg_name_lookup() can find/identify the user account.

The right way of doing this would be to pass around an array of user objects instead of just strings. In this way the name displayed could be the boincuser_name but then when lookup occurs the drupal id could be used. However, this would require recoding quite a bit of the privatemsg module, so the simpler solution of appending "_boincid" is used.

PR #2146, opened 2017-09-26
Remote job submission: fix bugs when using per-job templates

davidpanderson:
The RPC handler wasn't looking in per-job templates for rsc_fpops_est.

### DIFF
--- a/html/inc/submit.inc
+++ b/html/inc/submit.inc
@@ -42,12 +42,12 @@ function req_to_xml($req, $op) {
     <app_name>$req->app_name</app_name>
     <batch_name>$req->batch_name</batch_name>
 ";
-    if ((isset($req->output_template_filename)) && ($req->output_template_filename)) {
-        $x .= "    <output_template_filename>$req->output_template_filename</output_template_filename>
+    if ((isset($req->result_template_file)) && ($req->result_template_file)) {
+        $x .= "    <result_template_file>$req->result_template_file</result_template_file>
 ";
     }
-    if ((isset($req->input_template_filename)) && ($req->input_template_filename)) {
-        $x .= "    <input_template_filename>$req->input_template_filename</input_template_filename>
+    if ((isset($req->workunit_template_file)) && ($req->workunit_template_file)) {
+        $x .= "    <workunit_template_file>$req->workunit_template_file</workunit_template_file>
 ";
     }
     if ((isset($req->app_version_num)) && ($req->app_version_num)) {
@@ -91,11 +91,16 @@ function req_to_xml($req, $op) {
             }
             $x .= "        </input_file>\n";
         }
+<<<<<<< HEAD
         if (!empty($job->input_template)) {
             $x .= "    $job->input_template\n";
+=======
+        if (!empty($job->wu_template)) {
+            $x .= "    <wu_template>\n$job->wu_template\n    </wu_template>\n";
+>>>>>>> parent of 9bbeb05... Remote job submission: fix bugs when using per-job templates
         }
-        if (!empty($job->output_template)) {
-            $x .= "    $job->output_template\n";
+        if (!empty($job->result_template)) {
+            $x .= "    <result_template>\n$job->result_template\n    </result_template>\n";
         }
         $x .= "    </job>
 ";
@@ -151,7 +156,7 @@ function do_http_op($req, $xml, $op) {
     }
     $e = (string)$r->error_msg;
     if ($e) {
-        return array(null, "BOINC server: $e");
+        return array(null, $e);
     } else {
         return array($r, null);
     }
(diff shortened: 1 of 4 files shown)


html/inc/submit.inc
```diff
@@ -339,12 +339,12 @@ function boinc_retire_batch($req) {
 }
 
 //// example usage follows
-if (0) {
+if (1) {
```
ChristianBeer: Leftover from debugging?

html/inc/submit.inc
```diff
@@ -91,11 +91,11 @@ function req_to_xml($req, $op) {
             }
             $x .= "        </input_file>\n";
         }
-        if (!empty($job->wu_template)) {
-            $x .= "    <wu_template>\n$job->wu_template\n    </wu_template>\n";
+        if (!empty($job->input_template)) {
+            $x .= "    $job->input_template>\n";
```
JuhaSointusalo: I think there's an extra `>` after `input_template`.

davidpanderson:
I made those changes.
Only nanoHUB is using this feature; they know about the API change

TheAspens:
@davidpanderson - @JuhaSointusalo  and @ChristianBeer  have both made a inline comment in the "files changed" view of this review.  Can you take a look at their feedback and make those changes if necessary?

TheAspens:
@davidpanderson - it looks like you updated https://boinc.berkeley.edu/trac/wiki/RemoteJobs when you created this pull request.  Can you confirm that the documentation matches what the code will be after this merge?

JuhaSointusalo:
Reviewing and leaving a comment wasn't enough to clear the "requested changes", can't find how to do it in help and I don't want approve the PR because I'm not qualified to do so. :(

tristanolive:
@JuhaSointusalo It's ok for you to approve the PR if your suggested changes have been addressed. Christian's review will still be open until he likewise approves the changes or someone else dismisses his review (if he is not available), so that will still hold up the PR for further review. 

That being said, it looks like Christian's concerns may have been addressed by David, though it isn't clear to me how to tell if only nanoHUB is using a feature of the API.

davidpanderson:
Yup.  Matches the doc.  Moving to standardize terminology
("input" and "output" rather than "wu" and "result")

davidpanderson:
I'm not seeing the inline comments

ChristianBeer:
> Only nanoHUB is using this feature; they know about the API change

The general problem with that is that we don't know who is using what feature. We need to have a way to tell projects that update later that the API changed between certain versions. That is only possible with a proper server versioning. In the meantime I was hoping that such information is added to commit messages so that an API change is documented close to the code and developers can find out later. That is also necessary when compiling a change log later.

html/user/submit_rpc_handler.php
```diff
@@ -588,9 +603,9 @@ function get_batch($r) {
         $batch_name = BoincDb::escape_string($batch_name);
         $batch = BoincBatch::lookup_name($batch_name);
     } else {
-        xml_error(-1, "BOINC server: batch not specified");
+        xml_error(-1, "batch not specified");
     }
-    if (!$batch) xml_error(-1, "BOINC server: no such batch");
+    if (!$batch) xml_error(-1, "no such batch");
```
ChristianBeer: Scrutinizer found an issue about the scope of `$batch`. It does not know that `xml_error()` terminates execution so it found an execution path where `$batch` is not defined at this point. This can be solved by adding `$batch = NULL;` before the if block.

TheAspens:
@ChristianBeer  -  can you amend a commit when it is in a branch without messing other people up?   I've not developed the level of sophistication with git to feel comfortable enough to know when I can rewrite history without messing up other people.  If we can't amend the commit, is there an alternate suggestion of how to add the information that you request to the commit history?

TheAspens:
@ChristianBeer - it would be a bit wierd, but David could revert 9bbeb053bac6df96be78ecd48973958eef3f98d6 and then revert the new commit but this time make sure that the commit message had text about the api change.  Would that work?

TheAspens:
@davidpanderson - the comments had been resolved - sorry about that.

TheAspens:
I think that once the commit message and the $batch reference Christian identified are resolved then we can merge this in.

David -if you run:
git checkout dpa_submit3
git pull
git revert 9bbeb053
git log
-- git the hash of the most recent info commit
git revert ##new hash##

and then edit that commit message, then that should address the commit message discussion.

If you can then add the change @ChristianBeer  asked above about $batch then I think this will be good to merge.

tristanolive:
@TheAspens Normally, rewriting history is bad because it requires everyone else who has done any work from that point to have to manually replay changes / fix the history. In this case, it's David's personal branch that no one else is developing, so I'd think all he'd have to do is close this PR, fix the commits locally, and then open a new PR.

davidpanderson:
isaac(~/boinc) git revert 9bbeb05
Automatic revert failed.  After resolving the conflicts,
mark the corrected paths with 'git add <paths>' or 'git rm <paths>'
and commit the result.

This is silly.

TheAspens:
As per @tristanolive suggestion, I have created pull request #2172 and amended the commit to reference the api changes.  I am closing this pull request in favor of that pull request since it contains the fixed message.

PR #2148, opened 2017-09-26
Admin web: use bootstrap

davidpanderson:


TheAspens:
Tested and looks good.  Thanks

PR #2152, opened 2017-10-02
client: in get_app_config GUI RPC and no config file, don't cause error

davidpanderson:


JuhaSointusalo:
The GUI RPC protocol isn't too consistent in how it handles missing files. `handle_get_global_prefs_file` and `handle_get_global_prefs_override` return an error message. `handle_get_all_projects_list` (`read_all_projects_list_file`) and `handle_get_cc_config` return an empty reply.

This would add third way to signal a missing file. And, IMHO, signalling an error by making up a file is just wrong.

I think it would be better if `handle_get_app_config` returned an error message if the file is not found or can't be read.

CharlieFenton:
Neither my Mac nor my Windows 7 PC has an app_config.xml file in any of the project directories (for example, .../projects/setiathome.berkeley.edu/), or anywhere in the BOINC Data directory. Under what circumstances would this RPC ever find such a file? Or am I looking in the wrong place?

TheAspens:
@JuhaSointusalo and @CharlieFenton - I've tested this fix and it resolves the error message you get if you run a command like: 

[knreed@oc7811586721 BOINC]$ ./boinccmd --get_app_config http://www.worldcommunitygrid.org
Operation failed: unexpected XML tag or syntax

The app config file is not so much a missing file as it is an optional file since most users will never use it nor does it need to be there.  The <app_config></app_config> that David returns is the equivalent in this case to an empty file which would be the logical contents of the file if we were to create the file upon installation.  I am personally OK with this approach since I'd rather generate the default contents if the feature is used rather than return an error indicating that the file is not found when that is in fact the most likely outcome.

If you guys could let me know if you agree, then I will proceed with merging.

Note: I'm reviewing #2164 next.

TheAspens:
Forget everything I wrote above :-} I'm closing this one since #2164 adds the error response that @JuhaSointusalo asked for.

Issue #2153, opened 2017-10-03
Inconsistency between xml from boinc_gahp and expected xml in submit_rpc_handler 


nikolasgianna:
In commit 381e0caf the xml tag <batch> was removed from the create_batch function in submit_rpc_handler, but boinc_gahp still produces it. This breaks the remote create_batch function through boinc_gahp.

lfield:
This issue can be closed. It is not longer relevant.

PR #2154, opened 2017-10-03
Fix inconsistency between submit_rpc_handler and boinc_gahp <batch> tag

nikolasgianna:
This is to fix issue #2153. If this is breaking something else that I'm not aware of, I could change boinc_gahp instead.


lfield:
This PR can be closed. It is no longer valid.

PR #2158, opened 2017-10-06
[WIP] Handle missing gfx file

CharlieFenton:
The BOINC Manager and Screensaver Coordinator do not correctly handle the rare case when there is a project graphics executable associated with a worker executable but the is graphics executable is missing from the project directory. 

If a task with this situation is selected in the Tasks tab of the Manager, the "Show graphics" button is enabled, but pressing it does nothing.

If a task with this situation is running, the Screensaver Coordinator tries to run it, but quickly determines that it is not running and then goes on to the next task with associated graphics. But if all running tasks are associated with the missing graphics executable, then it will continuously cycle through them, constantly displaying "BOINC screensaver loading" on MS Windows or "Starting screensaver graphics.\nPlease wait ..." on Macintosh.

The changes in this PR test that the graphics executable file actually exists. If not, the Manager does not enable the "Show graphics" button, and the Screensaver Coordinator does not include  it in its list of graphics executables that can be run.

### DIFF
--- a/clientgui/ViewWork.cpp
+++ b/clientgui/ViewWork.cpp
@@ -896,10 +896,17 @@ void CViewWork::UpdateSelection() {
 
         // Disable Show Graphics button if the selected task can't display graphics
         //
-        if (!strlen(result->web_graphics_url) && !strlen(result->graphics_exec_path)) {
-            enableShowGraphics = false;
+        if (!strlen(result->web_graphics_url)) {
+            if(!strlen(result->graphics_exec_path)) {
+                enableShowGraphics = false;
+            } else {
+                if (!boinc_file_exists(result->graphics_exec_path)) {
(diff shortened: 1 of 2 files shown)


CharlieFenton:
I've changed this to WIP because I'm not sure whether it is worth the extra overhead, given how rare / unlikely it is for a project graphics executable file to be missing. I thought a missing project graphics executable might be the cause of a problem reported by Clifton Harding, but more recent information makes this seem unlikely. There are also other (similarly unlikely) scenarios which could cause these symptoms, such as a graphics executable that crashes shortly after launch, and these patches don't address those situations.

CharlieFenton:
I am closing this because there has been no evidence that this issue has ever been a problem. In particular, it does not appear that a missing graphics executable was responsible for the problem reported by Clifton harding.

PR #2159, opened 2017-10-06
Drupal: Added special characters to act as delimiters.

drshawnkwang:
BOINC usernames may contain any character, including commas. Therefore we use the special characters \xc2\xa0, which is displayed as a space, as the delimiter.

Part of
* https://dev.gridrepublic.org/browse/DBOINCP-254
Update: also see https://dev.gridrepublic.org/browse/DBOINCP-411

### DIFF
--- a/drupal/sites/default/boinc/modules/contrib/privatemsg/privatemsg.module
+++ b/drupal/sites/default/boinc/modules/contrib/privatemsg/privatemsg.module
@@ -848,7 +848,6 @@ function privatemsg_new(&$form_state, $recipients = array(), $subject = '', $thr
 
   $recipients_string = '';
   $body      = '';
-
   // convert recipients to array of user objects
   if (!empty($recipients) && is_string($recipients) || is_int($recipients)) {
     $recipients = _privatemsg_generate_user_array($recipients);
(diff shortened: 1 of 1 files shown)


brevilo:
Can you elaborate on the specific need for this? I can't find a hint on that in the references ticket.

Also, a non-breaking space (the new delimiter) might cause issues when lines are wrapped because the wrapping won't occur between names but within names, if they contain regular spaces.

drshawnkwang:
I thought I had added a message on [JIRA ticket 254](https://dev.gridrepublic.org/browse/DBOINCP-254) regarding the problem. But I guess it didn't get saved (I probably reloaded/refreshed before submitting). I'll post something there too.

The problem is that BOINC display names may contain any character: commas and spaces included. Thus the addition of a special character separator that act as the delimiter for the names. Also, the delimiter split will ignore any spaces within names because it only searches for the special character(s), which happen to display as empty spaces to the user.

brevilo:
Ok, thanks. But that doesn't resolve my concerns about the HTML rendering I discussed above (added to comment after initial statement).

drshawnkwang:
I guess I don't 100% understand what your concern is? Can you elaborate where you believe the text will (or does) wrap?

brevilo:
My concern is only relevant/applicable when the user names are ever rendered as a list-like string on a web page, that is, including the delimiter: e.g. `Oliver Bock,DELIMShawn Kwang`. In such a case any wrapping (e.g. when used in a table column) would cause the following line break when `DELIM` is a non-breaking space:

```
Oliver 
Bock,&nbsp;Shawn
Kwang
```

drshawnkwang:
Closing this PR in favor of a different approach to the problem.

PR #2161, opened 2017-10-07
Vboxwrapper: Add support for VirtualBox 5.2 via the published COM Int…

romw:
…erface

This basically adds COM support for VirtualBox 5.2


romw:
As near as I can tell, AppVeyor was using the old project file to compile vboxwrapper and that caused the compilation failure.  The commit diff clearly includes the new files.  Building locally succeeds.

JacobWKlein:
Just wanted to mention 2 things:
- I was chatting with Rom, and the discussion led to this checkin
- We used the 5.2 Beta 3 "tlb" here, since 5.2 RC1 "tlb" was unavailable in Oracle's SDK. I prompted Oracle about this here: https://forums.virtualbox.org/viewtopic.php?f=15&t=84921
- We'll want to update to the "release tlb" when it becomes available.
- When a new version of vboxwrapper is available, let me know, and I'll happily test it some. I think I remember how to do so, by going offline, using a copy of my data, and using app_info.xml to use a newer vboxwrapper.

Thanks,
Jacob Klein

samples/vboxwrapper/vboxwrapper.cpp
```diff
@@ -483,6 +486,9 @@ int main(int argc, char** argv) {
         if ((5 == vbox_major) && (1 <= vbox_minor)) {
```
JuhaSointusalo: Should this test for equality only? I think that otherwise vboxwrapper will try to create a VM first with 5.1 interface (which might not be a huge problem) if VirtualBox 5.2 is installed.

romw: Thanks for catching that.

PR #2172, opened 2017-10-17
Remote job submission: fix bugs when using per-job templates (replaces PR #2146)

TheAspens:
This pull request is a redo of David's #2146 but with the commit message changed as per request from @ChristianBeer

### DIFF
--- a/html/inc/submit.inc
+++ b/html/inc/submit.inc
@@ -42,12 +42,12 @@ function req_to_xml($req, $op) {
     <app_name>$req->app_name</app_name>
     <batch_name>$req->batch_name</batch_name>
 ";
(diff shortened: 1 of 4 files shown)


TheAspens:
With the commit message changed, the only outstanding issue from #2146 was the comment about the output from Scrutinizer.  However, the output from Scrutinizer is misleading as it does not interpret the effect of calling xml_error correctly and so its complaint about scope is less relevant (we can debate whether declaring local scope for the variable is desirable, but that is not sufficient to block merging this change which has become aged).  As a result, I am approving and merging this pull request.

TheAspens:
Note that this should not be considered a merge of my own pull request as this is really just pull request #2146 which was created by @davidpanderson

Issue #2165, opened 2017-10-10
BOINC Manager crashes when language is Spanish


Ageless93:
From https://boinc.berkeley.edu/dev/forum_thread.php?id=11871
System Windows 10 x64 v1703 v15063
When BOINC Manager language is set in Spanish, BOINC Manager will crash when the Notices tab is opened.

davidpanderson:
I couldn't repro this (Win 7, 7.8.3 64-bit)

Ageless93:
Excuse me, I read his post wrong. It's when he switches to the properties of the project that his BOINC Manager crashes. I can reproduce that wit 7.8.3
When set to Spanish, Selecting Seti@Home, then clicking Propiedades, I lose control over BOINC Manager, eventually the Windows 'busy' swirling mouse-icon comes up and after 2 minutes the manager crashes completely. The client keeps running. 

It doesn't matter which project is selected, it crashes the manager always.

Ageless93:
[stderrgui-spanish-crash.txt](https://github.com/BOINC/boinc/files/1379352/stderrgui-spanish-crash.txt)
Adding the stderrgui.txt about this crash

JuhaSointusalo:
There's quite a few bad format strings besides the Spanish one. I'll put the ones in Manager and Setup below. Client didn't have any critically bad translations. pofilter doesn't understand %1 style placeholders so it can't check web translations and I don't know how to make it check Android translations.

The strings are from master. Some of the po files have not been updated for several years and some of the bad translations may already be obsolete.

**Manager**
el
```
# (pofilter) printf: Missing printf variable: %s
#: AccountInfoPage.cpp:453
#, c-format
msgid "Are you already running %s?"
msgstr "Εκτελείτε ήδη %;"
```
es
```# (pofilter) printf: Missing printf variable: %s
#: clientgui/DlgAdvPreferencesBase.cpp:594
#, c-format
msgid ""
"Limit the percentage of disk space used by %s on the volume where it stores "
"data."
msgstr "Limita el porcentaje de disco usado por $s en la unidad donde se almacena la información."
```
```
# (pofilter) printf: Different number of printf variables
#: clientgui/DlgItemProperties.cpp:300 clientgui/DlgItemProperties.cpp:308
#, c-format
msgid "%s total, %s average"
msgstr "%s total, % medio"
```
fr
```
# (pofilter) printf: Different number of printf variables
#: clientgui/DlgOptions.cpp:710
#, c-format
msgid ""
[...]

PR #2175, opened 2017-10-18

ChristianBeer:
Fixes issues with format strings in various languages. I checked the translations with pofilter and fixed the obvious mistakes myself in transifex. I then forced an update of translations also for languages that are incomplete because the bad format strings are crashing the Manager.

See #2165


ChristianBeer:
This is a PR specifically for the 7.8 release branch. I'm going to do the same cleaning for the master branch soon.

TheAspens:
@ChristianBeer @JuhaSointusalo @RichardHaselgrove @davidpanderson 

Moving a conversation from email on to this ticket.

The discussion was about the following question.  Given that we are about to cut a branch and start testing the 7.10 client, should we merge this pull request into the 7.8 branch and release a new 7.8 build based on it?   Should we close it without merging?   Should we leave it open in case we decide we need to do a new 7.8 build?

We do owe an apology to Christian that we did not merge and do a build based on this back in October.

RichardHaselgrove:
My view was that it would be 'nice to merge' so as to leave the 7.8 branch in as tidy a condition as possible for the archivists. Some of the typos (see original issue) were so serious that they caused the Manager to crash when certain languages were in use.

JuhaSointusalo:
In my view release branch can be in two states. Either it is known to be good, that is, it was used to release something and nothing has been added since then, or release manager is actively working on next release. This is so that if there ever is need for an urgent hotfix release, say there is a security hole that's been actively exploited. If the branch is known to be good you can get the hotfix out right now with minimal testing. If there is additional code since the last release you'd need to test those too and maybe fix them delaying releasing the hotfix.

When Rom run the show he added commits to release branch without releasing a new client and sometimes even added new untested features. So basically, the supposedly stable branch was actually unstable. Because of that we have hotfix branches branched from release branch. IMHO that's wrong way to do it.

For this PR that would mean that either you merge this PR and release a new version containing just this change, or that you wait and see if a new version is needed before 7.10 is ready, or that you close without merging and hope that a new version won't be needed. I'd go with waiting.

TheAspens:
This looks like it was left hanging from the 7.8 release.  @ChristianBeer - should we do something with branch or can it be closed?

RichardHaselgrove:
> This is a PR specifically for the 7.8 release branch.

The consensus seems to be that the branch code should remain as a true record of the release, warts and all. Closing.

PR #2177, opened 2017-10-21
Mac: Updates to build all 64-bit, to prepare for OS 10.14

CharlieFenton:
Apple has announced that they will remove full support for 32-bit applications in OS 10.14. The changes in this PR modify the Xcode project to build everything as 64-bit executables, and update Mac-specific source code to fix any problems which result from building as 64-bit.

For now, the BOINC libraries will still be built as Universal (32-bit and 64-bit) binaries, to allow projects to continue building either 32-bit or 64-bit executables until they have finished converting to all 64-bit code.

Note that the dependent static libraries wxWidgets 3.0.0 (libwx_osx_cocoa_static.a) and sqlite-3.11.0 (libsqlite3.a) must be rebuilt as 64-bit libraries. I have modified the build scripts buildWxMac.sh and buildsqlite3.sh in this feature branch to build these two libraries as "universal binaries" containing both 32-bit and 64-bit versions in the same library file. This will allow both the old and the new Xcode projects to build BOINC Manager. But the CI builds will fail until the dependent libraries have been updated.


CharlieFenton:
@ChristianBeer: It appears that Travis is not rebuilding the dependent libraries libwx_osx_cocoa_static.a and libsqlite3.a using the updated build scripts (see warnings from Travis build below.) What do we need to do to fix this?

ignoring file /Users/travis/build/BOINC/boinc/3rdParty/buildCache/mac/lib/libwx_osx_cocoa_static.a, file was built for archive which is not the architecture being linked (x86_64): /Users/travis/build/BOINC/boinc/3rdParty/buildCache/mac/lib/libwx_osx_cocoa_static.a
ignoring file /Users/travis/build/BOINC/boinc/3rdParty/buildCache/mac/lib/libsqlite3.a, file was built for archive which is not the architecture being linked (x86_64): /Users/travis/build/BOINC/boinc/3rdParty/buildCache/mac/lib/libsqlite3.a

CharlieFenton:
@ChristianBeer: I updated the wxWidgets and sqlite3 build scripts to force rebuilds if they had previously been built as only 32-bit or only 64-bit binaries, in case this is the reason they were not rebuilt by Travis CI. Note that you can cut the wxWidgets build time by about one half if you call the buildWxmac.sh script with the -nodebug argument.

ChristianBeer:
@CharlieFenton I believe that Travis CI can't build 32bit libraries. I tried this for the screensaver which needed the 32bit FTGL lib. But according to https://github.com/BOINC/boinc/blob/master/mac_build/buildMacBOINC-CI.sh#L135 it didn't work so I disabled it.

Could we add a `--32` and `--64` option to the build scripts so we can suppress a 32bit build on Travis CI but enable it for manual use when the developer is sure that his system can produce 32bit libraries? My guess is that we would need to install 32bit versions of system libraries on Travis CI but which increases build time again.

ChristianBeer:
> Note that you can cut the wxWidgets build time by about one half if you call the buildWxmac.sh script with the -nodebug argument.

This is already the case.

ChristianBeer:
I deleted all Caches and triggered a rebuilt to see what errors are present when building the combined library.

CharlieFenton:
>I believe that Travis CI can't build 32bit libraries. I tried this for the screensaver which needed the 32bit FTGL lib.
That's odd. Nothing has changed in this build script. The default screensaver graphics app now requires 64-bit libraries in this feature branch, but it always used 32-bit libraries before (master and the 7.8 branch). I have always built it as a universal library containing both 32-bit and 64-bit binaries so it would be useful in graphics apps built by projects. 

How did you tell it to build on 32-bits? Th buildFTGL.sh script doesn't offer that option.

CharlieFenton:
I don't understand the error from AppVeyor, which seems to be a problem with finding the correct solution file for the Windows build. I haven't touched anything in the win_build directory.

ChristianBeer:
> How did you tell it to build on 32-bits? The buildFTGL.sh script doesn't offer that option.

It doesn't need an option it does it by default: https://github.com/BOINC/boinc/blob/master/mac_build/buildFTGL.sh#L112 and combines the 32 and64bit libraries with `lipo` at the end. That was, until recently, the only place where this combined library was produced.

The rebuilt from scratch seems to have worked but it would be better to copy the `lipo -verify_arch` checks also to `3rdParty/buildMacDependencies.sh` since that is the first place it checks if a rebuilt is needed.

> I don't understand the error from AppVeyor, which seems to be a problem with finding the correct solution file for the Windows build. I haven't touched anything in the win_build directory.

I don't understand that either. @JuhaSointusalo can you please take a look at that? Maybe a visual studio update over night? Since this PR (and especially the last two commits) have no impact on the Windows chain we can for now ignore the AppVeyor failure.

CharlieFenton:
> it would be better to copy the lipo -verify_arch checks also to 3rdParty/buildMacDependencies.sh since that is the first place it checks if a rebuilt is needed.

@ChristianBeer: I have updated 3rdParty/buildMacDependencies.sh in this feature branch to do that. Please review my changes to this file.

Cheers,
--Charlie

ChristianBeer:
Nice, thanks for the quick update. The changes to `3rdParty/buildMacDependencies.sh` are fine with me.

I looked through the other changes and they are all within `#ifdef __WXMAC__` so I don't see problems with other platforms and trust that Charlie did enough testing on Mac.

CharlieFenton:
@davidpanderson or @ChristianBeer: could one of you please merge this?

PR #2182, opened 2017-10-24
Initial commit of rpc documentation

nikolasgianna:
This is an attempt at documenting the RPC interface. Both requests and expected replies.

### DIFF
--- a/rpc_doc.md
+++ b/rpc_doc.md
@@ -1,0 +1,1782 @@
+# GUI RPC Protocol
+
+The GUI RPC protocol lets GUIs like the BOINC Manager communicate with the core client.
+
+Note that the RPC server is the core client, and the RPC client is a GUI or add-on communicating with it (such as BOINC Manager). This may seem confusing but this terminology will be used on the rest of the page.
+
+## Basic Structure
+
+The protocol is based on XML, and it's strictly request-reply. The client sends requests to the server, and waits for a reply; the server never sends anything without getting a request from the client first. Both requests and replies are terminated with the control character `0x03`.
+
+Self-closing tags must not have a space before the slash, or current client and server will not parse it correctly. For example, send `<authorized/>`, not `<authorized />`.
+
+Requests are inside `<boinc_gui_rpc_request>` elements, and replies from the RPC server are inside `<boinc_gui_rpc_reply>` elements (in both cases there is a `0x03` byte after the closing tag). The current RPC server implementation doesn't require the `<boinc_gui_rpc_request>` tag, which is handy for debugging (you can connect via ​[netcat](http://netcat.sourceforge.net) and just type `<auth1/>`); however, clients must not rely on this, and must always send the `<boinc_gui_rpc_request>` root tag.
+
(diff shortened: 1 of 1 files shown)


AenBleidd:
@nikolasgianna, I can create unit tests (with Google Test Framework) according to this documentation.
This can help us to find issues both in documentation and in current implementation of XML parser.
Also I had plans to move all XML related operations to libxml++ (or smth similar) so these unit tests can be also a very useful tool to check new parsing mechanism.
@ChristianBeer, @davidpanderson, @JuhaSointusalo, @CharlieFenton, @TheAspens:
What do you think?
Thanks

JuhaSointusalo:
@nikolasgianna

Thanks for writing the documentation. We have other documentation in [Trac wiki](http://boinc.berkeley.edu/trac/wiki). I think this should go there so that everything is in one place.

@AenBleidd

BOINC certainly needs automated testing. Rather than hijacking this PR I think it would be better to open a new issue to discuss the subject. What tools we would use and how the testing would be implemented. There is a million different testing frameworks so having some kind of comparison of some number of popular frameworks would be nice.

AenBleidd:
@JuhaSointusalo, of course, I'l create new ticket. I'm just curious whether unit testing is needed

JuhaSointusalo:
@AenBleidd Well, in my opinion, yes, BOINC needs unit tests as well as higher level tests. I think there has been some mentions now and then about wanting to plug testing into CI process.

lfield:
@JuhaSointusalo I would recommend that this documentation resides in git and not Trac. Any changes to the RPC should also include updates to the documentation.

@AenBleidd I think that tests would be very welcome. If we can document and automatically test the main interfaces of BOINC, it would help everyone. 

As this is new documentation, I see no reason not to merge it. Any mistakes can easily be updated.

lfield:
Any reason why this new documentation hasn't been merged?

TheAspens:
First of all - thank you for the detailed documentation.  It will be useful.

The technical documentation  for BOINC has up to this point always been on the trac wiki.  For an example similar to what you have documented are this page: https://boinc.berkeley.edu/trac/wiki/WebRpc

Maintaining the documentation in two places is inconsistent and makes it more difficult to maintain and find information or to know where to add things.  If you are interested in changing the standard existing standard for where the documentation is maintained, I would suggest starting a discussion on the boinc-dev mailing list and getting a consensus of the committers to agree with the change in direction.  

Also - moving all of the technical documentation into top level md files might not be sufficient to contain all of the documentation without it becoming quite disorganized.  I would suggest that a more expansive proposal of how we should maintain the technical documentation within git would be better.

In the meantime, would you object to me adding your content to to the trac wiki?

lfield:
No, I don't mind you adding the content to the trac wiki. My recommendation would be to have such specifications in git so that they can be maintained in sync with the code. It ensures that the specification agrees with the code historically and on branches.

lfield:
@TheAspens are you going to add this to the wiki?

lfield:
The trac wiki has been updated. 

https://boinc.berkeley.edu/trac/wiki/GuiRpcProtocol

PR #2187, opened 2017-10-24
Manager: Fix #1083 EventLog window GUI glitch with selected unfocused…

### LINKED ISSUE #1083: GUI Glitch - Actions on 2+ selected rows appear to leave every other row unselected

AenBleidd:
… rows.

### DIFF
--- a/clientgui/DlgEventLog.cpp
+++ b/clientgui/DlgEventLog.cpp
@@ -249,12 +249,24 @@ bool CDlgEventLog::Create( wxWindow* parent, wxWindowID id, const wxString& capt
         wxNullFont
     );
 #if EVENT_LOG_STRIPES
+    m_pList->EnableAlternateRowColours();
+
+#if wxCHECK_VERSION(3,1,0)
+    wxColour stripe = m_pList->GetAlternateRowColour();
+#else
+    const wxColour stripe = m_pList->wxListCtrl::OnGetItemAttr(1)->GetBackgroundColour();
+#endif
+
     m_pMessageInfoGrayAttr = new wxListItemAttr(
         wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOWTEXT),
-        wxColour(240, 240, 240),
+        stripe,
         wxNullFont
     );
-    m_pMessageErrorGrayAttr = new wxListItemAttr(*wxRED, wxColour(240, 240, 240), wxNullFont);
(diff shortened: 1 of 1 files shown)


AenBleidd:
This pull request is more clear then previous one: #2137

AenBleidd:
Changes are made according to @JuhaSointusalo comments.

JuhaSointusalo:
My idea wasn't well tested and that shows - the build fails on Linux and Mac.

```
DlgEventLog.cpp: In member function ‘bool CDlgEventLog::Create(wxWindow*, wxWindowID, const wxString&, const wxPoint&, const wxSize&, long int)’:
DlgEventLog.cpp:257:50: error: ‘wxListCtrl’ is not a base of ‘CDlgEventLogListCtrl’
     const wxColour stripe = m_pList->wxListCtrl::OnGetItemAttr(1)->GetBackgroundColour();
                                                  ^
```
That's because Linux and Mac use `wxGenericListCtrl` directly. That could be fixed by changing the base class used in the call but I'm having second thoughts about this. This error tells me the code is getting too clever.  I think it will be better to copy the colour setting code from `wxListCtrlBase::EnableAlternateRowColours` here rather than add more cleverness (and eventually break the build in more clever ways).

I can make a PR that I promised earlier that fixes this and adds more error checking in a bit if that's ok with you.

In related matters, have you reported or are you working on reporting the problem to wxWidgets developers? I started poking the list control sample the other day to turn it into something reportable. If you are not working on the report I could finish it up.

AenBleidd:
@JuhaSointusalo,

> I can make a PR that I promised earlier that fixes this and adds more error checking in a bit if that's ok with you.

Of course this is OK.

> In related matters, have you reported or are you working on reporting the problem to wxWidgets developers? I started poking the list control sample the other day to turn it into something reportable. If you are not working on the report I could finish it up.

I do not work on the report so please finish it.

Thank you

PR #2188, opened 2017-10-25
Build: fail if wxWidgets lib can not be found but is required

### LINKED ISSUE #1714: Building and installing “BOINC Manager 7.7.0”
I [cloned the current Git repository](https://boinc.berkeley.edu/trac/wiki/SourceCodeGit#CloningtheGitrepository "Cloning the Git repository") on my openSUSE Tumbleweed system a moment ago. It seems that the software configuration and the corresponding build succeeded then.

```
elfring@Sonne:~/Projekte/Bau/BOINC> ~/Projekte/BOINC/lokal/configure --disable-server --disable-fcgi --enable-dynamic-client-linkage --enable-optimize --prefix=/usr/local/BOINC-ab_20161119 && make -j4
…
--- Configuring BOINC 7.7.0 (Release) ---
--- Build Components: (client manager libraries) ---
…
```

The software installation worked also to some degree. Now I wonder why the subdirectory “bin” does not contain the program files which I would usually expect from the display of the content from a package like “boinc-manager 7.6.32-1.5” here. :confused:

I would appreciate your advices for a solution.

ChristianBeer:
Currently the build system will silently disable the Manager build when wxWidgets is missing or not a recent enough version.

Fix #1714

### DIFF
--- a/m4/boinc_wxwidgets.m4
+++ b/m4/boinc_wxwidgets.m4
@@ -9,7 +9,9 @@ AC_DEFUN([BOINC_OPTIONS_WXWIDGETS],[
      [])
 
    AM_OPTIONS_WXCONFIG
-     AM_PATH_WXCONFIG($1, [_ac_cv_have_wxwidgets=yes], [_ac_cv_have_wxwidgets=no])
+     AM_PATH_WXCONFIG([$1],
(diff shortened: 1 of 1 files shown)


elfring:
Thanks for your adjustment of the affected build script.

PR #2189, opened 2017-10-26
Manager: update Xcode project and scripts for wxWidgets 3.1.0 because…

CharlieFenton:
… wxWidgets 3.0.0 can not be built using Xcode 9 due to APIs no longer supported by the header files in the OS 10.13 SDK

### DIFF
--- a/3rdParty/buildMacDependencies.sh
+++ b/3rdParty/buildMacDependencies.sh
@@ -116,6 +116,14 @@ download_and_build() {
             # already built but not for correct architectures so force rebuild
             doClean="-clean"
         fi
+    else
+        # tell subsequent scripts to build everything from scratch
+        doClean="-clean"
+        # delete any FILEFLAGS for other versions of this library built previously
+        BASENAME="${DIRNAME%-*}"
+        rm -f "${PREFIX}/${BASENAME}"*
+        # delete any previous build of this library (may be redundant with -clean)
+        rm -f "${PREFIX}/lib/${PRODUCTNAME}"
     fi
     if [ ! -d ${DIRNAME} ]; then
         if [ ! -e ${FILENAME} ]; then
--- a/3rdParty/buildMacDependencies.sh
+++ b/3rdParty/buildMacDependencies.sh
@@ -148,9 +148,6 @@ if [ "${doclean}" = "yes" ]; then
     mkdir -p "${PREFIX}"
 fi
 
-# TEMPORARY CODE TO REMOVE INCORRECT FLAGFILE FROM EARLIER BUGGY SCRIPT
-rm -f "${PREFIX}/wxWidgets-3.1.0_done"
-
 # this will pull in the variables used below
 source "${ROOTDIR}/mac_build/dependencyNames.sh"
 
--- a/3rdParty/buildMacDependencies.sh
+++ b/3rdParty/buildMacDependencies.sh
@@ -148,6 +148,9 @@ if [ "${doclean}" = "yes" ]; then
     mkdir -p "${PREFIX}"
 fi
 
+# TEMPORARY CODE TO REMOVE INCORRECT FLAGFILE FROM EARLIER BUGGY SCRIPT
+rm -f "${PREFIX}/wxWidgets-3.1.0_done"
+
 # this will pull in the variables used below
 source "${ROOTDIR}/mac_build/dependencyNames.sh"
 
--- a/mac_build/boinc.xcodeproj/project.pbxproj
+++ b/mac_build/boinc.xcodeproj/project.pbxproj
@@ -2385,7 +2385,7 @@
 			attributes = {
 				LastUpgradeCheck = 0620;
 			};
-			buildConfigurationList = DD9E2381091CBDAE0048316E /* Build configuration list for PBXProject "boinc" */;
+			buildConfigurationList = DD9E2381091CBDAE0048316E /* Build configuration list for PBXProject "boinc_wx310" */;
 			compatibilityVersion = "Xcode 3.0";
 			developmentRegion = English;
 			hasScannedForEncodings = 1;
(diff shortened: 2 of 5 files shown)


ChristianBeer:
Travis used the cached 3.0.0 wxWidgets library and didn't do a rebuilt. This is because `mac_build/buildWxMac.sh` is not aware which version is in the cache and just checks for presence of a file (which is still the old one).

Best way to mitigate this right now is to use the `-clean` option when triggering a new build from `3rdParty/buildMacDependencies.sh` something like:

```
    if [ -e "${FLAGFILE}" ]; then
        lipo "${PREFIX}/lib/${PRODUCTNAME}" -verify_arch ${ARCHS}
        if [ $? -eq 0 ]; then
            echo "${DIRNAME} seems already to be present in ${PREFIX}"
            return 0
        else
            # already built but not for correct architectures so force rebuild
            doClean="-clean"
        fi
    else
        # tell subsequent scripts to build everything from scratch
        doClean="-clean"
    fi
```
This will enforce building the new version. This will still not remove the old `wxWidgets-3.0.0_done` file but I don't have a solution for that.

CharlieFenton:
Since it will not remove the old wxWidgets-3.0.0_done file, won't that cause other branches and master to be built with wxWidgets 3.1.0 instead of 3.0.0? Apparently that is not a problem in this case, but it would be in the future if an upgrade to a dependent library is not compatible with the previous version.

CharlieFenton:
An idea for deleting the FLAGFILE: The dependent library directories (and therfore the FLAGFILE) all have the generic library name followed by a dash and a version number, such as "curl-7.50.2" and "wxWidgets-3.0.0". The script could make a copy of the name up to the first dash ("curl", "wxWidgets") and then delete any existing files in the cache beginning with that string.

What do you think?

ChristianBeer:
I don't think it is enough to check for old FLAGFILEs and trigger a clean build for this dependency. The new version will overwrite any existing build products in the cache but it will not delete old products that are not used anymore from the cache.

I'm currently testing the above fix and a way to delete the old flagfile since the Linux scripts have the same problem and I didn't actually test a version upgrade in the past. But I think detecting a version upgrade based on the first part of the FLAGFILE, triggering a clean build and deleting the old FLAGFILE should work.

ChristianBeer:
Upgrading only wxWidgets is more than just building a new version. The product names (libraries and config files) contain the minor version number which for the 3.0 to 3.1 jump is relevant here.

The easiest way of course is to delete the Travis cache again and to a complete rebuild. This is not suitable for local developer environments. I'll have to think about this some more. Maybe we need to make the actual build scripts (`mac_build/buildWxMac.sh` and 3rdParty/buildWxLinux.sh`) version aware so they can delete old build products from the cache themselves.

CharlieFenton:
At least on the Mac, the libraries do not contain the version number; they are the same for all versions of a library That is why I only need to change the header and library search paths in the Xcode projects to use a different library version. 

So I believe that using the -clean argument to the build script will delete the library for any old version. I have added "rm -f ${PREFIX}/lib/${PRODUCTNAME}" to the BuildmacDependencies.sh script to ensure this, though I believe it is redundant. 

I have updated the BuildmacDependencies.sh script in this branch to try the above.

Regarding config files, I didn't think they were stored in the cache. In any case, my individual library build scripts always run configure again if the -clean argument is passed (except buildWxmac.sh, which builds wxWidgets from an Xcode project and so doesn't use configure and make.)

And if I am wrong about the config files, I believe that calling "make distclean" in my various library build scripts if the -clean argument is present should take care of the configure files; is that correct? Should I add that?

CharlieFenton:
My changes seem to have worked. The Travis log 293587755 for build 2524 clearly shows that it built wxWidgets 3.1.0. I did have to add a temporary line of code to the buildMacDependencies.sh script to delete the incorrect wxWidgets-3.1.0_done flag created by the old script. I have now removed that temporary line.

If the FLAGFILE for the desired version of a library is not present, should the script also remove all previously downloaded source files for that library? In other words, should I change line 124 from this:
        rm -f "${PREFIX}/${BASENAME}"*
to this?
        rm -fR "${PREFIX}/${BASENAME}"*

I have another question: Should not the variables in lines 128 through 138 be enclosed in quotation marks, except extra_options and doClean?

CharlieFenton:
Any branches off master that do not have my changes will incorrectly use wxWidgets-3.1.0 to build the Mac Manager and fail, and will pollute the cache with a false wxWidgets-3.0.0_done file. This will require removing the incorrect *_done files before the updated BuildMacDependencies.sh will work properly.

@ChristianBeer, do you have a way to remove the incorrect *_done files directly, or will we need to again temporarily add my change of commit 8c9375d?

If you agree with my changes to BuildMacDependencies.sh, please merge this branch so that branches created in the future will handle version changes properly.

CharlieFenton:
@ChistianBeer: is there a way for me to inspect the Travis cache to better understand what it is doing with the wxWidgets files on Macintosh?

Have you had a chance to review my changes to BuildMacDependencies.sh?

ChristianBeer:
> And if I am wrong about the config files, I believe that calling "make distclean" in my various library build scripts if the -clean argument is present should take care of the configure files; is that correct? Should I add that?

Since the wxWidgets build process is different for Mac and Linux not everything I wrote above applies to the Mac here. Please also bear in mind that `make distclean` only affects the build directory not the install directory which in the case of Travis is not the same (because we only want to cache the build product and headers not the source or intermediate artifacts.

> If the FLAGFILE for the desired version of a library is not present, should the script also remove all previously downloaded source files for that library?

There are no previously downloaded source files or build directories in the cache. Only on the developers computer and there it is the responsibility of the developer.

> I have another question: Should not the variables in lines 128 through 138 be enclosed in quotation marks, except extra_options and doClean?

Possibly, it wouldn't hurt at least.

> Any branches off master that do not have my changes will incorrectly use wxWidgets-3.1.0 to build the Mac Manager and fail, and will pollute the cache with a false wxWidgets-3.0.0_done file. This will require removing the incorrect *_done files before the updated BuildMacDependencies.sh will work properly.

Your changes are not yet part of master so the cache that is used for builds from the master branch are using the cache that contains wxWidgets 3.0.0. Only after this PR is merged the cache for the master branch is changed after 3.1.0 was built and since the script deletes the old `wxWidgets-3.0.0_done` file now, the new "master" cache will contain all the correct files afterwards. The "cache" is actually several independent caches. A pull request has it's own cache but starts of with the cache from the "master" branch. Without this the cache would be very unstable.

The changes look fine but I don't have time too look at them in detail now. I'll review this until the weekend though.

CharlieFenton:
Would @ChristianBeer or anyone please merge this? It's been ready for almost 2 weeks.

PR #2104, opened 2017-09-10
[WIP] Manager: Use wxHtmlWindow in DlgItemProperties.

AenBleidd:
Use wxHtmlWindow instead of wxWebView.
Handle mouse events to show the correct state of the 'Copy selected' button.

Signed-off-by: Vitalii Koshura <lestat.de.lionkur@gmail.com>

PR #2190, opened 2017-10-26
make webview support optional

ChristianBeer:
Some Linux distributions are dropping webkitgtk support which is needed for wxWebView to work. This will build a second wxWidgets cache that has no webview support in order to test new code.

This should in the end combine #2093, #2104 and the [Debian patch](https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=876940;filename=boinc-no-webview.patch;msg=5) and build two managers. One with webview support and one without.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,7 @@ script:
 - if [[ "${BOINC_TYPE}" == "client" ]]; then ( ./configure --disable-server --disable-manager && make ) fi
 - if [[ "${BOINC_TYPE}" == "apps" ]]; then ( ./configure --enable-apps --disable-server --disable-client --disable-manager && make ) fi
 - if [[ "${BOINC_TYPE}" == "manager" && "${TRAVIS_OS_NAME}" == "linux" ]]; then ( ./3rdParty/buildLinuxDependencies.sh && ./configure --disable-server --disable-client --with-wx-prefix=${TRAVIS_BUILD_DIR}/3rdParty/buildCache/linux && make ) fi
+- if [[ "${BOINC_TYPE}" == "manager" && "${TRAVIS_OS_NAME}" == "linux" ]]; then ( ./3rdParty/buildLinuxDependencies.sh --disable-webview --cache_dir ${TRAVIS_BUILD_DIR}/3rdParty/buildCache/linux2 && ./configure --disable-server --disable-client --with-wx-prefix=${TRAVIS_BUILD_DIR}/3rdParty/buildCache/linux2 && make ) fi
 - if [[ "${BOINC_TYPE}" == "manager" && "${TRAVIS_OS_NAME}" == "osx" ]]; then ( ./3rdParty/buildMacDependencies.sh -q && ./mac_build/buildMacBOINC-CI.sh --no_shared_headers ) fi
 - if [[ "${BOINC_TYPE}" == "libs-mingw" ]]; then ( cd lib && MINGW=x86_64-w64-mingw32 make -f Makefile.mingw ) fi
 - if [[ "${BOINC_TYPE}" == "apps-mingw" ]]; then ( cd lib && MINGW=x86_64-w64-mingw32 make -f Makefile.mingw wrapper ) fi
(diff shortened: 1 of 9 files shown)


ChristianBeer:
I restarted the manager build on Travis to see if this still builds. It seems that `make distclean` triggers a `configure` run at the end which understandably fails because there is no wxWidgets installed and it didn't even build. This needs some more thinking on (Why is `distclean` triggering a full configure run?). Unless someone else helps me out this is probably not going to be ready for 7.10 @RichardHaselgrove 

Also: this is in the wrong project. I'm moving it to Client and Manager.

ChristianBeer:
Turns out it was just a simple typo. Should be good to go now. The non-webview build needs some more testing though see referenced pull requests.

TheAspens:
@AenBleidd or @JuhaSointusalo - could one of you review this and the related pull requests see if this is ready for merge?

AenBleidd:
Reviewed and checked with VS2013 on Windows 10. Looks OK.
Will try to do the same on linux

JuhaSointusalo:
Notices are ok now but Properties dialog has the same problem as before at least on Mint 14.04. Repro:

Open a project's Properties.
Click Copy All and paste the text in editor. Works so far.
Replace clipboard with something else outside of Manager.
Select text from the dialog, lets say "location" by holding down mouse button and moving mouse.
Click Copy Selected and paste the text. Doesn't work.
Click Copy All. Doesn't work.
Click empty space to clear the selection.
Click Copy All. Doesn't work.
Double click "location" to select the word.
Click Copy Selected. Doesn't work.
Click Copy All. Doesn't work.
Click empty space to clear the selection.
Triple click "location" to select text in the entire cell.
Click Copy Selected. Now it works.
Click Copy All. Now it works. (Or maybe not. The first time it didn't work.)

So what are we going to do with this? Debian/Ubuntu and Suse have wxWebView disabled and can't release 7.10 as it is now. The options that I see are:

1. Debian/Ubuntu and Suse don't release 7.10.
2. This PR is merged and Linux version won't have working copy text function in Properties dialog. This is bit of a bummer because copying text from the dialog would be one of the What's New highlights for 7.10.
3. Debian/Ubuntu and Suse patch the Properties dialog back to what it was in 7.8. This is bit of bummer because then they don't have all the What's New stuff.
4. Releasing 7.10 is delayed until the Properties dialog is fixed or a workaround is found. This could quickly turn into a couple of weeks delay.

So bad choices all.

AenBleidd:
'Copy selected' doesn't work in linux due to some internal issue of
wxWidgets.
I propose just to disable 'Copy selected' functionality in linux

Best regards,
Vitalii Koshura

2018-02-01 22:46 GMT+02:00 Juha Sointusalo <notifications@github.com>:

JuhaSointusalo:
I think I found the problem and looks like it can be fixed but it's past my bed time already so explanation will have to wait until tomorrow.

@ChristianBeer Ok to push the fix straight to this branch?

JuhaSointusalo:
Unlike Windows or Mac, X11 has actually two (plus some more nobody uses) clipboards, `CLIPBOARD` and `PRIMARY`. `CLIPBOARD` is used in the same manner as in other systems, Ctrl-C/Ctrl-V or Copy/Paste from menu to copy or paste content.

`PRIMARY` is different. Any selected text is automatically copied to `PRIMARY`. Paste is done by mouse middle click or scrollwheel click or simultaneous left and right click depending on what kind of mouse you have.

`wxHtmlWindow` has builtin support for copying selected text to `PRIMARY` as it should have. To do that it uses `wxClipboard` which supports both `CLIPBOARD` and `PRIMARY`. But there is one catch, `wxClipboard` remembers which clipboard it used last and keeps using it until told otherwise. `wxHtmlWindow` sets the clipboard to `PRIMARY` and that is used until Manager is restarted.

So, the Copy buttons actually work, it's just that the text goes to wrong clipboard. The fix is easy, before copying text set clipboard to `CLIPBOARD`.

Selecting text with triple click "fixed" the Copy buttons because of a bug in wxWidgets, The triple click handler incorrectly copies selected text to `CLIPBOARD`.

(`wxHtmlWindow` has a few functions that access `CLIPBOARD` but they are not used in Manager.)

JuhaSointusalo:
Event Log had the same problem so I made the fix there too.

ChristianBeer:
Nice that in the end a solution was found.

TheAspens:
@ChristianBeer , @AenBleidd and @JuhaSointusalo  - is everyone satisfied with this pull request?  If so, @AenBleidd since you are indicated as the reviewer do you mind merging?  I haven't reviewed this code yet, so I can't at this time.

PR #2191, opened 2017-10-26
Build: fix distclean target

ChristianBeer:


TheAspens:
Looks fine and I've tested.

PR #2194, opened 2017-10-28
mgr: fix eventlog colors

### LINKED ISSUE #1083: GUI Glitch - Actions on 2+ selected rows appear to leave every other row unselected

JuhaSointusalo:
On Windows, `wxListCtrl` has a bug that makes it draw selected unfocused rows with colors used for selected focused rows when the row background color is `wxSYS_COLOUR_BTNFACE`.

With default color schemes, on Windows XP the list is drawn correctly, but on Windows 10 the `(240,240,240)` used in stripe color is coincidentally the same as `wxSYS_COLOUR_BTNFACE` and selected rows are drawn with wrong colors.

Fix this by using the same stripe color that wxWidgets would use by default and that is used in Projects, Tasks and Transfers lists. Future proof the code by including a workaround just in case Windows changes color scheme again. If such color scheme change occurs then the same workaround needs to be added to Projects, Tasks and Transfers lists.

Fixes #1083.

To support the above change remove `CDlgEventLogListCtrl::GetBackgroundColour()`. The method was private and not callable from CDlgEventLog and overriding the corresponding base class method wasn't useful in the first place.

### DIFF
--- a/clientgui/DlgEventLog.cpp
+++ b/clientgui/DlgEventLog.cpp
@@ -248,13 +248,47 @@ bool CDlgEventLog::Create( wxWindow* parent, wxWindowID id, const wxString& capt
         wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOW),
         wxNullFont
     );
+
 #if EVENT_LOG_STRIPES
+    m_pList->EnableAlternateRowColours();
+    wxColour stripe_color;
+
+#if wxCHECK_VERSION(3, 1, 0)
+    stripe_color = m_pList->GetAlternateRowColour();
+    if (!stripe_color.IsOk())
+#endif
+    {
+        // copied from wxListCtrlBase::EnableAlternateRowColours(bool)
+
+        // Determine the alternate rows colour automatically from the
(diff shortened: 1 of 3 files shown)


RichardHaselgrove:
I've done a test build with VS2013 and this code - looks good on Windows 7 (was bad before). Will test on Windows 10 later.

@AenBleidd - could you review this, please, for possible inclusion in v7.10

RichardHaselgrove:
@AenBleidd - could you merge this one too, please? Changes haven't yet appeared on either Github or local clone.

PR #2196, opened 2017-10-29
Manager: 'Add Project Wizard': disable 'Next' button on wrong credent…

### LINKED ISSUE #1048: When adding a project, failed username/password should not show Next, and should not place the failures on a stack.
**Reported by JacobKlein on 30 Mar 41246410 06:40 UTC**
When adding a project, if you do not use the correct email address or password, you get a dialog that says:[Failed.[[BR]([BR]]Login)]Check the email address and password, and try again.

The problem is that you are allowed to click Next.  You should not be allowed to click Next. You should be forced to click Back or Cancel or close the window.

As a side-effect, if you click Next, Try again, Fail, Next, try again, Fail... it places all of these failures on a stack, such that you can then click Back Back Back to see several failed login screens within the wizard.

Please do not allow "Next" on a failed login when adding a project.

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1072

AenBleidd:
…ials

Fixes #1048.

### DIFF
--- a/clientgui/NotFoundPage.cpp
+++ b/clientgui/NotFoundPage.cpp
@@ -192,6 +192,7 @@ void CErrNotFoundPage::OnPageChanged( wxWizardExEvent& event ) {
 
     CWizardAttach* pWA = ((CWizardAttach*)GetParent());
     pWA->DisableNextButton();
+    pWA->GetBackButton()->SetDefault();
 
     Fit();
 }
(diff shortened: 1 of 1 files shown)


AenBleidd:
This fixes #1048

AenBleidd:
@Ageless93, @ChristianBeer, could you please review this? Thanks

CharlieFenton:
Line 554 of clientgui/wizardex.cpp has `m_btnNext->SetDefault();` so the disable button will be set as the default button, which is incorrect. You should also change the default button to the 'Back' button.

AenBleidd:
@CharlieFenton, could you please review this? Thanks

CharlieFenton:
I have one minor comment: the commit message is slightly misleading. It doesn't _enable_ the back button, it makes it the _default_. But that is not a significant issue, so I will go ahead and merge this.

PR #2202, opened 2017-10-31
Add server-side support for simplified register/download/attach

davidpanderson:
Add the server parts of the scheme described here:
https://boinc.berkeley.edu/trac/wiki/SimpleAttach

This includes:
- add login_token table and function for creating login tokens
- add Web RPC login_token_lookup.php for mapping token to auth
- change concierge protocol to take project ID and login token
- update download.php so that it gets client version info
    from versions.xml (from BOINC web site)
    and displays buttons with OS name, versions and file sizes
    like on the BOINC download page
- make "register.php" work

I've tested this all the way through downloading the installer file
with the augmented name.
Changes to the installer and client are needed to complete the system.

### DIFF
--- a/html/inc/account.inc
+++ b/html/inc/account.inc
@@ -1,7 +1,7 @@
 <?php
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2015 University of California
+// Copyright (C) 2017 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -16,7 +16,21 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 
-// forms for account creation and login
+// functions related to account creation and login:
+// - forms for create / login
+// - function to make login token
+
+// make login token record; return token
+//
+function make_login_token($user) {
+    $token = substr(random_string(), 0, 8);
+    $now = time();
+    $ret = BoincLoginToken::insert("(userid, token, create_time) values ($user->id, '$token', $now)");
+    if (!$ret) {
+        return null;
+    }
+    return $token;
+}
 
 function create_account_form($teamid, $next_url) {
     global $recaptcha_public_key;
--- a/html/inc/account.inc
+++ b/html/inc/account.inc
@@ -25,7 +25,7 @@
 function make_login_token($user) {
     $token = substr(random_string(), 0, 8);
     $now = time();
-    $user->update("login_token='$login_token', login_token_time=$now where id=$user->id");
+    $user->update("login_token='$token', login_token_time=$now");
     return $token;
 }
 
--- a/html/inc/account.inc
+++ b/html/inc/account.inc
@@ -20,15 +20,12 @@
 // - forms for create / login
 // - function to make login token
 
-// make login token record; return token
+// make login token, store in user record, return token
 //
 function make_login_token($user) {
     $token = substr(random_string(), 0, 8);
     $now = time();
-    $ret = BoincLoginToken::insert("(userid, token, create_time) values ($user->id, '$token', $now)");
-    if (!$ret) {
-        return null;
-    }
+    $user->update("login_token='$login_token', login_token_time=$now where id=$user->id");
     return $token;
 }
 
--- a/html/inc/util.inc
+++ b/html/inc/util.inc
@@ -514,6 +514,8 @@ function row_heading($x, $class='bg-primary') {
     );
 }
 
+// return hard-to-guess string of 32 random hex chars
+//
 function random_string() {
     return md5(uniqid(rand(), true));
 }
--- a/html/user/download.php
+++ b/html/user/download.php
@@ -16,49 +16,140 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 
-// 2nd phase of the "join" process.
-// The user has created an account.
-// Now see if they need to download BOINC.
+// Show a page with download links and instructions.
+//
+// - get platform from user agent string
+// - find latest version for that platform (regular and vbox)
+// - Create a login token.
+// - Show download button(s)
+//   The download will be via concierge, using the login token.
+//
+// By default both regular and vbox buttons will be shown, if available.
+// You can suppress one or the other by setting
+// <disable_regular_download> or <disable_vbox_download>.
+//
+// Notes:
+// 1) You need to have the client versions file
+//    (https://boinc.berkeley.edu/download_all.php?xml=1)
+//    saved as "versions.xml" in your html/user dir
+// 2) Put your project ID in a constant PROJECT_ID
+//    (this all works only for listed projects)
 
 require_once("../inc/util.inc");
+require_once("../inc/account.inc");
 
-// return the URL of a script on the BOINC web site
-// that will send the needed cookies
-// and optionally download the appropriate installer,
+// take the client info string reported by web browser,
+// and return best guess for platform
 //
-function concierge_url($user, $download) {
-    global $master_url;
-    $project_name = parse_config(get_config(), "<long_name>");
-    $project_desc = parse_config(get_config(), "<project_desc>");
-    $project_inst = parse_config(get_config(), "<project_inst>");
-    $user = get_logged_in_user(false);
-    $x =  "http://boinc.berkeley.edu/concierge.php";
-    $x .= "?master_url=".urlencode($master_url);
-    $x .= "&project_name=".urlencode($project_name);
-    if ($project_desc) {
-        $x .= "&project_desc=".urlencode($project_desc);
-    }
-    if ($project_inst) {
-        $x .= "&project_inst=".urlencode($project_inst);
-    }
-    if ($user) {
-        $x .= "&auth=".urlencode($user->authenticator);
-        $x .= "&user_name=".urlencode($user->name);
-    }
-    if ($download) {
-        $x .= "&download=1";
-        if (parse_bool(get_config(), "need_vbox")) {
-            $x .= "&need_vbox=1";
+function client_info_to_platform($client_info) {
+    if (strstr($client_info, 'Windows')) {
+        if (strstr($client_info, 'Win64')||strstr($client_info, 'WOW64')) {
+            return 'windows_x86_64';
+        } else {
+            return 'windows_intelx86';
         }
+    } else if (strstr($client_info, 'Mac')) {
+        if (strstr($client_info, 'PPC Mac OS X')) {
+            return 'powerpc-apple-darwin';
+        } else {
+            return 'x86_64-apple-darwin';
+        }
+    } else if (strstr($client_info, 'Android')) {
+        // Check for Android before Linux,
+        // since Android contains the Linux kernel and the
+        // web browser user agent string lists Linux too.
+        //
+        return 'arm-android-linux-gnu';
+    } else if (strstr($client_info, 'Linux')) {
+        if (strstr($client_info, 'x86_64')) {
+            return 'x86_64-pc-linux-gnu';
+        } else {
+            return 'i686-pc-linux-gnu';
+        }
+    } else {
+        return null;
     }
-    return $x;
 }
 
-function show_download_page() {
-    page_head("Download required software");
+// find release version for user's platform
+//
+function get_version() {
+    $v = simplexml_load_file("versions.xml");
+    $client_info = $_SERVER['HTTP_USER_AGENT'];
+    $p = client_info_to_platform($client_info);
+    foreach ($v->version as $i=>$v) {
+        if ((string)$v->dbplatform != $p) {
+            continue;
+        }
+        if (!strstr((string)$v->description, "ecommended")) {
+            continue;
+        }
+        return $v;
+    }
+    return null;
+}
 
-    $config = get_config();
+function download_button($v, $project_id, $token) {
+    return sprintf(
+        '<form action="https://boinc.berkeley.edu/concierge.php" method="post">
+        <input type=hidden name=project_id value="%d">
+        <input type=hidden name=token value="%s">
+        <input type=hidden name=filename value="%s">
+        <button class="btn btn-info">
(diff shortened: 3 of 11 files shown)


TheAspens:
An additional comment that I have is that based on the implementation, the token is left forever in the database.  It seems that they should expire after a time.  Do we need a new daemon that runs once a day to delete tokens older than X hours?  Since an entry gets created on the table each time that the download page is viewed, this table could get large fast.

TheAspens:
During testing I found that the download page didn't work (even after I added the project_id).  This is because of 
`Warning: simplexml_load_file(): I/O warning : failed to load external entity "versions.xml" in /root/project/html/user/download.php on line 77`
which is because the server wasn't allowed to make an outbound request.  It is a common security practice to block outbound requests for servers and so that page needs to be able to gracefully handle this situation.  Perhaps by having the recommended version set on the boinc website side (passing the project_id and hash to the server to construct the url) rather than building the link on the project website side.

TheAspens:
One last comment before returning it to you for review.  I navigated manually to the download page, but I couldn't find where on the website I could otherwise reach this download page.  How do users find the download link on project websites normally?  I'm sure I'm just missing it.

davidpanderson:
Now that I think about it, it would be better to get rid of the login_token table and just put the into in the user record.  A user isn't going to have more than one outstanding token.  This will simplify things.  Any objections?

davidpanderson:
BTW, the simplexml error is misleading.  It's trying to read a local file, versions.xml, which projects must download from the BOINC server; see https://boinc.berkeley.edu/trac/wiki/SimpleAttach#Whatprojectsmustdo

davidpanderson:
- There's no link to the download page.  The system won't work until we finish the client and installer parts, and do a client release.
- we can improve random_string(), but that's outside the scope of this

brevilo:
> Any objections?

Yes, as it isn't good database/schema design. Login tokens don't describe a user entity. They are an entity of its own. Any potential (yet unanticipated) extensions to that idea can be implemented way more easily if that entity already lives in its own table.

TheAspens:
I agree with Oliver that we should leave the login_token table in place.  I also believe that the tokens should expire.  One of the scenarios that @romw and WCG talked with this mechanism is that the tokens should expire after a short period of time (can't remember if we settled on 1 hour or longer).  The reason is that people share and copy the installer between different people.  If that token becomes a permanently feature then that installed becomes permanently tied to that user.  This is going to cause significant confusion over time and result in people wondering why they are contributing as some other user (leading them to think they were hacked).  By having the token expire, it meets the need of an easy to use auto-attach mechanism but limits the possibility of this confusion when someone shares the file.

TheAspens:
> BTW, the simplexml error is misleading. It's trying to read a local file, versions.xml, which projects must download from the BOINC server; see https://boinc.berkeley.edu/trac/wiki/SimpleAttach#Whatprojectsmustdo

Given that most projects are very slow to update anything is it wise to expect them to update this file on a regular basis to point to the latest version of the client via a manual process?  Wouldn't it be better to centralize this as part of the download page on the BOINC website so that the project websites will always be directing users to the latest version of the client?

html/inc/util.inc
```diff
@@ -514,6 +514,8 @@ function row_heading($x, $class='bg-primary') {
     );
 }
 
+// return hard-to-guess string of 32 random hex chars
+//
```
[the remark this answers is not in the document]
TheAspens: Based on your comments about this change being out of scope, I've opened https://github.com/BOINC/boinc/issues/2210 to address this issue.

html/user/download.php
```diff
@@ -16,110 +16,211 @@
...
+            return 'i686-pc-linux-gnu';
+        }
+    } else {
+        return null;
     }
-    if ($download) {
-        $x .= "&download=1";
-        if (parse_bool(get_config(), "need_vbox")) {
-            $x .= "&need_vbox=1";
+}
```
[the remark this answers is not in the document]
TheAspens: After further discussion, this logic along with the processing of versions.xml should be moved to the concierge.php page on the main boinc website.

html/inc/account.inc
```diff
@@ -16,7 +16,18 @@
...
 
-// forms for account creation and login
+// functions related to account creation and login:
+// - forms for create / login
+// - function to make login token
+
+// make login token, store in user record, return token
+//
+function make_login_token($user) {
+    $token = substr(random_string(), 0, 8);
```
TheAspens: this should be login_token (or the sql below should reference the token variable)

davidpanderson:
I added a 1-day timeout.  This is unrelated to whether we use a separate table.

davidpanderson:
The reason for having versions.xml on the project is so that the buttons can say the BOINC version, the VBox version, and the download size.  This is important information.

html/inc/account.inc
```diff
@@ -16,7 +16,18 @@
...
+    $now = time();
+    $user->update("login_token='$login_token', login_token_time=$now where id=$user->id");
```
TheAspens: where id=$user->id is not required (the object is built to add that)

TheAspens:
> The reason for having versions.xml on the project is so that the buttons can say the BOINC version, the VBox version, and the download size. This is important information.

What I'm suggesting is that when you send the user to http://127.0.0.1/boincserver/download.php (this is the test server I'm using) what it does is create and store the token and then redirect the user to https://boinc.berkeley.edu/download.php?project_id=314&token=1234abcd&vm=1 (or whatever attributes are needed)

The https://boinc.berkeley.edu/download.php would use the attributes passed on the buttons as required.   Those buttons can then invoke the concierge page so that it constructs the download url properly.

This would avoid the need for projects to maintain the version information and therefore when a new release of the client became available, all projects would be recommending the new version.

This could then also be setup so that if a project didn't have a <project_id> in their config file, then http://127.0.0.1/boincserver/download.php would simply redirect to https://boinc.berkeley.edu/download.php without any parameters (that way new projects that don't have a project_id won't have their download links broken).

TheAspens:
> I added a 1-day timeout.

thanks!

davidpanderson:
One of the goals is that the user never leaves the project site (i.e. doesn't go to BOINC).  This makes the user experience more consistent.

TheAspens:
> This is unrelated to whether we use a separate table.

If we have a login_table, then each time the user goes to download the client we can generate a fresh token with a 24 hour timeout which they can use.

If we store this information on the user table, they can only have one token active at a time.  Not the end of the world, but it means that when they go to the download page, we either 1) replace the existing token and time, 2) keep the token but update the timestamp or 3) keep the token and the timestamp.

Case 1 would invalidate a previously download client and in that case if they tried to use that client, then it would just fall back to normal attach (which would be a rarer case and probably not a big deal)

Case 2 would keep the same token so the newly download file and the previously download file would have the same token.  This could lead to some of the issues occurring if the token was kept forever.  However, it is likely to be rare that would occur so this is probably ok.

Case 3 could mean that the newly download client has a token that expires quickly which would be a bad user experience.


Case 2 is probably the best option if you keep it on the user table.

If it is on a login_token table, then you can generate a fresh token for each download (and is in my opinion the best solution - and a small variant of your original solution).

TheAspens:
> One of the goals is that the user never leaves the project site (i.e. doesn't go to BOINC). This makes the user experience more consistent.

That is a good goal.  

I'm still concerned about it not being present on initial project setup and not being updated regularly.  

Would it be possible to add a script to the make_project function to fetch the latest version from the BOINC website and if it can't reach spit out an error message telling the user what to do but keep proceeding with the setup?  It would also be good to have a task that runs once every 24 hours to fetch the latest version.  Thoughts?  This would still cause an issue for projects with firewalls blocking outbound access but they will have to deal with it.

davidpanderson:
As of now, we're creating a new token each time.  This means that the most recent download will auto-attach, but not older ones.  I think this if fine.

davidpanderson:
It's not an issue for new projects, since they won't be on the vetted-projects list.  Vetted projects will have to fetch the versions list when there's a new client release.  We could automate this, I suppose.

davidpanderson:
The reason for checking "need_vbox" on the project is that if a project has only VBox apps, they should show only the VBox download button.

TheAspens:
Copying this error to this pull request since it belongs here:

From testing https://github.com/BOINC/boinc/pull/2207

I got this error:
```
[knreed@oc7811586721 test-boinc]$ cat login_token_lookup_reply.xml 
<br /> <b>Fatal error</b>: Class 'BoincLoginToken' not found in <b>/root/project/html/user/login_token_lookup.php</b> on line <b>26</b><br />
```

TheAspens:
> It's not an issue for new projects, since they won't be on the vetted-projects list. Vetted projects will have to fetch the versions list when there's a new client release. We could automate this, I suppose.

What is the logic for unvetted projects and the download button?  

Right now http://127.0.0.1/boincserver/download.php will break if there is not a versions.xml and a project_id defined.  

Will you set things so that if project_id is undefined then a download link anywhere on the site will go to https://boinc.berkeley.edu/download.php and if project_id and versions.xml is defined then they go to http://127.0.0.1/boincserver/download.php?

I'm guessing that it would be easier to just write http://127.0.0.1/boincserver/download.php so that if either project_id or versions.xml is undefined, then go to https://boinc.berkeley.edu/download.php.

If they are both defined, then render the new http://127.0.0.1/boincserver/download.php page as you have written.

davidpanderson:
My thinking was that projects would link to download.php (e.g. from their home page) only if they had a project ID and versions.xml.  Otherwise they'd link to the BOINC download page.

TheAspens:
> My thinking was that projects would link to download.php (e.g. from their home page) only if they had a project ID and versions.xml. Otherwise they'd link to the BOINC download page.

I thought a download link was standard on BOINC sites.  However, after checking E@H and S@H I didn't find download links on those sites except for one time after registering.

The link that shows up after registering goes to the BOINC website.  Since it is part of the BOINC source code is that going to change?

TheAspens:
Now I get this error with the latest code in login_token_lookup_reply.xml

```
<br />
<b>Warning</b>:  Missing argument 1 for main(), called in /root/project/html/user/login_token_lookup.php on line 44 and defined in <b>/root/project/html/user/login_token_lookup.php</b> on line <b>25</b><br />
<br />
<b>Warning</b>:  Cannot modify header information - headers already sent by (output started at /root/project/html/user/login_token_lookup.php:25) in <b>/root/project/html/inc/util.inc</b> on line <b>238</b><br />
<br />
<b>Warning</b>:  Cannot modify header information - headers already sent by (output started at /root/project/html/user/login_token_lookup.php:25) in <b>/root/project/html/inc/util.inc</b> on line <b>239</b><br />
<br />
<b>Warning</b>:  Cannot modify header information - headers already sent by (output started at /root/project/html/user/login_token_lookup.php:25) in <b>/root/project/html/inc/util.inc</b> on line <b>240</b><br />
<br />
<b>Warning</b>:  Cannot modify header information - headers already sent by (output started at /root/project/html/user/login_token_lookup.php:25) in <b>/root/project/html/inc/util.inc</b> on line <b>241</b><br />
<br />
<b>Warning</b>:  Cannot modify header information - headers already sent by (output started at /root/project/html/user/login_token_lookup.php:25) in <b>/root/project/html/inc/util.inc</b> on line <b>242</b><br />
<!DOCTYPE html>
        <html lang="en">
        <head>
        <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>Unable to handle request</title>

        <meta charset="utf-8">
        <link type="text/css" rel="stylesheet" href="http://127.0.0.1/boincserver//bootstrap.min.css" media="all">
    
        <link rel=alternate type="application/rss+xml" title="RSS 2.0" href="http://127.0.0.1/boincserver/rss_main.php">
        </head>
    <body >
<!-- SVN VERSIONS -->

```

davidpanderson:
Oops!  There was a bug.  Please update and try again.

davidpanderson:
The current (pre-autologin) model is that projects link to the BOINC download page, and registration happens in the BOINC Manager.  That will continue to be the case with unvetted projects.

TheAspens:
> The current (pre-autologin) model is that projects link to the BOINC download page, and registration happens in the BOINC Manager. That will continue to be the case with unvetted projects.

There are two paths to join (at least on S@H)

1) Click on the Join Seti@Home, download, register through client.  (this is one you refer to)
2) Clicking on the "Sign Up" and register.  After registration, you are redirected to your profile page where a download link is shown. 

What link will be used in the 2nd case?

I would like to suggest that new project download page be where all download links go.  It checks for project_id and versions.xml and if either is not present, redirect to the BOINC website for downloads and if they are present, then show your new page.  This means that any links added to download only have to go to a single page and that contains all the logic for downloads.

TheAspens:
If there is an existing login_token_lookup_reply.xml file present in the data directory, it will not replace it.

TheAspens:
Got the following, investigating why:

```
<!DOCTYPE html>
        <html lang="en">
        <head>
        <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>Unable to handle request</title>

        <meta charset="utf-8">
        <link type="text/css" rel="stylesheet" href="http://127.0.0.1/boincserver//bootstrap.min.css" media="all">
    
        <link rel=alternate type="application/rss+xml" title="RSS 2.0" href="http://127.0.0.1/boincserver/rss_main.php">
        </head>
    <body >

```

TheAspens:
client sends: 

/boincserver/login_token_lookup.php?token=1_5226c01e

server expects:

/boincserver/login_token_lookup.php?user_id=1&token=5226c01e

davidpanderson:
See last commit to dpa_autologin_client (passes user ID in RPC req)

html/user/download.php
```diff
@@ -16,110 +16,214 @@
...
+        <input type=hidden name=filename value="%s">
+        <button class="btn btn-success">
+        <font size=+1><u>Download BOINC + VirtualBox</u></font>
+        <br>for %s (%s MB)
+        <br><small>BOINC %s, VirtualBox %s</small></a>
+        </form>
+        ',
+        $project_id,
+        $token,
+        $user->id,
```
JuhaSointusalo: There is no `user_id` field in the HTML above.

TheAspens: I opened issue #2224  for this since this pull request has been merged (apparently prematurely - sorry about that).

PR #2204, opened 2017-10-31
scheduler: add <rte_no_stats> config flag to estimate job runtime wit…

davidpanderson:
…hout stats

The scheduler estimates job runtime based on statistics
of past jobs for this (host, app version).
This doesn't work well if the distribution of runtimes is very wide,
as may be the case of universal apps.

If this flag is set, runtime estimation is based solely on
CPU/GPU peak FLOPS and job FLOPs estimate.

### DIFF
--- a/sched/sched_config.cpp
+++ b/sched/sched_config.cpp
@@ -299,6 +299,7 @@ int SCHED_CONFIG::parse(FILE* f) {
         if (xp.parse_double("maintenance_delay", maintenance_delay)) continue;
         if (xp.parse_bool("credit_by_app", credit_by_app)) continue;
         if (xp.parse_bool("keyword_sched", keyword_sched)) continue;
+        if (xp.parse_bool("rte_no_stats", rte_no_stats)) continue;
 
         //////////// SCHEDULER LOG FLAGS /////////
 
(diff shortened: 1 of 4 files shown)


TheAspens:
Shouldn't this be a characteristic of a particular app rather than the whole project?  I can see cases where this is true for some apps on a project but not others.  Can the flag be moved from the project config file to a characteristic of the app table/app object?

davidpanderson:
We can do that if there's a demand.  This is for nanoHUB and TACC, which use a single VM app for all jobs.

TheAspens:
WCG is looking at exploring using VM apps for a possible set of smaller projects we are looking at running.  As a result we would want to use a setting like this for that but still use the traditional settings for our native apps.  I think that the use case is pretty strong for others to do that.  Please make this a parameter of app instead of for the whole project since it will be just as easy for nanoHUB and TACC to use it that way but it allows for much greater flexability.

davidpanderson:
I'll do this after dpa_autologin is merged (to avoid conflict on db_update.php).

davidpanderson:
But for now please merge this because nanoHUB needs it.  The per-app variant is an additional feature.

TheAspens:
Ok - sounds good.  Thanks.

PR #2207, opened 2017-11-02
client/server/web: add support for streamlined register/download/attach

davidpanderson:
- on startup, look for and parse a file containing the installer filename,
    which encodes a project ID and login token.
- look up the project ID in the all-projects list
- do an RPC to that project, mapping the login token to weak auth
- attach to that project using weak auth


TheAspens:
@ChristianBeer - can you take a look at the Travis CI build?  I was able to build it fine on my linux machine.

TheAspens:
@CharlieFenton - can you try building this on mac?  It looks like the Travis CI build for the manager failed on mac.  If there is a config change required for mac, do you mind making it and pushing it to this branch?

ChristianBeer:
I believe the changes made to the Linux and Windows build files also need to be made to the Mac build files (xcode project). The relevant files from the Mac log:
```
❌  Undefined symbols for architecture x86_64
> Symbol: PROJECT_LIST::read_file()
> Referenced from: CLIENT_STATE::process_autologin() in acct_setup.o
❌  ld: symbol(s) not found for architecture x86_64
```

As to the quality of the PR: A commit that adds two new files and just says `Add new files` in the commit message is useless. Even the files don't contain any documentation about their purpose. How is anyone in the future who wants to know why those files were added going to find that out? I'm very glad that I'm not going to be that person anymore.

TheAspens:
David, 

I've created a file called installer_filename.txt in /var/lib/boinc with the contents:

314_b9c2e254

I've built the manager from this branch and I've modified all_projects_list.xml to include

`    <project>
        <name>Localhost@Home</name>
        <id>314</id>
        <url>http://127.0.0.1/boincserver/</url>
        <general_area>Tomato Research</general_area>
        <specific_area>Ag</specific_area>
        <description><![CDATA[Pizza]]></description>
        <home>Lots of pizza</home>
    <platforms>
        <name>i686-pc-linux-gnu</name>
        <name>windows_intelx86</name>
        <name>windows_x86_64</name>
        <name>x86_64-apple-darwin</name>
        <name>x86_64-pc-linux-gnu</name>
        <name>arm-android-linux-gnu</name>
    </platforms>
      <image>https://boinc.berkeley.edu/images/radioactive.jpg</image>
      <summary>Monitor radiation levels</summary>
    </project>
`
but I cannot trigger the manager to read the file and initiate the attach.  

I'm running the server from https://github.com/BOINC/boinc/pull/2202 on this machine so it should be able connect (I can attach manually).

Any ideas what I'm doing wrong?

davidpanderson:
Did you build the client from this branch?  That's what reads the filename, not the manager.

TheAspens:
ah - I built the client, but didn't run it (left the one from epel running).  Let me retest.

TheAspens:
Ok - got to to read the file.  However I got this error from the server:

`[knreed@oc7811586721 test-boinc]$ cat login_token_lookup_reply.xml 
<br />
<b>Fatal error</b>:  Class 'BoincLoginToken' not found in <b>/root/project/html/user/login_token_lookup.php</b> on line <b>26</b><br />
`

davidpanderson:
It should work now with the dpa_autologin branch server code.

TheAspens:
If there is an existing login_token_lookup_reply.xml file present in the data directory, it will not replace it.  The code should make sure that it can replace the file with the results of a new query.

davidpanderson:
I couldn't repro the problem with existing login_token_lookup_reply.xml on Linux or Win.  What is the exact behavior?

CharlieFenton:
> If there is a config change required for mac, do you mind making it and pushing it to this branch?

Under the current rules, don't I need to create a branch and a pull request and have somebody else do the merge?

> A commit that adds two new files and just says Add new files in the commit message is useless


From this message I gather that new files need to be added to the build. Which files?

TheAspens:
> I couldn't repro the problem with existing login_token_lookup_reply.xml on Linux or Win. What is the exact behavior?

It doesn't replace the file.  It leaves the existing file in place.

TheAspens:
> Under the current rules, don't I need to create a branch and a pull request and have somebody else do the merge?

Charlie - more than one person can work on a branch/pull request.  I'm acting as the reviewer and you are another contributor on this branch.  I will be reviewing the work that you do in this branch as part of the overall review of the pull request.  Thus you can simply do the work needed so that David's changes compile properly on mac directly on this branch.

Once you have done that work, it would be good if you could indicate that it works on mac since I will only be able to validate that it compiles on mac via the Travis CI build job.

Thanks for the help!

TheAspens:
I've encountered this as I do each round of testing I have to stop the client, remove the file, start the client and retest.  You should be able to create a file with junk in it before starting the client and then see it occur.

TheAspens:
Ok - after further testing, I cannot reproduce the issue with the existing login_token_lookup_reply.xml file so that issue is resolved.

TheAspens:
I'm done with my review.  As soon as this compiles under mac I will merge this pull request.  @davidpanderson - I think @CharlieFenton had some questions for you above:

> From this message I gather that new files need to be added to the build. Which files?

davidpanderson:
You can push to this branch (just not to master).
The new files are client/project_list.cpp and project_list.h

CharlieFenton:
OK, I'll try to get to thsi on Monday. I don't have time today.

JuhaSointusalo:
I tested this with a clean data directory, as if BOINC was just installed.

I added installer_filename.txt, edited all_project_list.xml to include my test project and edited my user account on the test project to include login token. Then I started Manager and let it start the client as it normally does.

The client saw the login token and initiated fetch for account info. Then it continued with its startup code which includes listening for GUI RPCs. The Manager noticed that the client is ready and listening, asked for attached projects and because the client wasn't attached to anything yet showed Add Project wizard. After this the client finished up auto attaching to the test project.

IMHO, this feature needs more polishing.

davidpanderson:
I fixed the problem Juha reports above.

TheAspens:
@JuhaSointusalo - nice catch.

With David's additional implementation I've tested the following scenario:

1) I have a project running that includes the latest login_token_lookup.php code that is properly setup with versions.xml and a project id. 
2) I've built the client with the latest code in this branch.  I've setup the all_projects_list.xml to have an entry for my test project.
3) I've registered an account and visited the /download.php page on my new project which means that I have a new account with a login token.
4) I created a installer_filename.txt file that is setup correctly but has a invalid token (to simulate what happens if the user uses the installation file the next day or after they visited the download page again but uses an older download).

The add project dialogue does not display even though the project does not auto-attach.

One of the reasons appears to be is because if the auto-attach is not successful, then the installer_filename.txt is never removed.  I think that the code needs to make sure that the installer_filename.txt is removed if any of the failure conditions occur.

Second, it might be good that if an auto-attach mechanism was in progress then check for the pop-up of the attach project dialogue needs to be held for a short time (1.5 seconds max?  might need to check what the response time will be to the lookup vs the the usability of that popup being delayed) and then rechecked and display or not display the lookup.

TheAspens:
Scratch my comment above for now.  The client connected even without a valid token.  The server side version of this code does not check for a valid token.  It only does a lookup based on userid and then checks the timestamp.  The token itself is not checked.  That needs to be fixed and then the scenario I outlined can be retested (although i suspect it will have that issue based on reading the code).

davidpanderson:
Oops!  Fixed.

CharlieFenton:
I have added the missing files to the Mac Xcode project in this branch, and it builds OK. 

> Once you have done that work, it would be good if you could indicate that it works on mac since I will only be able to validate that it compiles on mac via the Travis CI build job.

If I understand this feature correctly, when a user new to BOINC signs up on a project's web site, the project will modify the name of the BOINC installer by appending a unique numeric or alphanumeric token. The new user downloads and runs the installer. The installer then must extract the token from the installer's file name and write it into the BOINC Data directory in some specific format for the client to read.

This will require my making changes to the Mac BOINC installer to extract the token and write the file. I am not comfortable spending a huge amount of time on this, so please point me to the detailed specifications for the exact format of the modified installer name and the file to be written to the BOINC Data directory. The standard Mac BOINC installer is named "BOINC Installer.app". I hope that there will be a known delimiter after "BOINC Installer: before the token, such as a hyphen, to make it easier to parse. Note that ".app" is the file name extension (like ".exe" on Windows) and so must must come after the token.

Note also that "BOINC Installer.app" is an OS X "bundle", not a simple file. This means it is actually a directory tree, but the Mac OS displays to the user as if it were a single file. For this reason, it must be compressed into a flat file (usually as a zip archive) before it is sent to the user. But the zip file is typically named something like boinc_7.8.3_macOSX_x86_64.zip, while the bundle it contains is named simply "BOINC Installer.app". 

The Safari browser automatically inflates and discards the zip file, so the user only sees "BOINC Installer.app", and there is no way I can think of for the installer to get the name of the original zip file. So the project web site must add the token to the installer name (which is actually the name of the top level directory of the bundle) and then compress the bundle into a flat file for downloading, using a compression scheme known to OS X. (OS X supports most standard UNIX-compatible compression schemes, including .zip, .tar, .bz2, and others.) 

I don't know if there are any Mac-specific file attributes that could be lost if this is done on a non-Mac computer. Hopefully, it will be OK as long as the server runs UNIX or Linux. 

OS X checks the file signature only after expanding the zip file, and I am fairly certain that the signature is not dependent on the bundle's name, so adding the token should not cause problems with the signature.

I'm sorry that I forgot about this complication when @davidpanderson asked me about the ability to parse the installer name 2 weeks ago.

If you want me to test this, please set up a test project to which I can upload my test installer, or which gets the installer from http://boinc.berkeley.edu/dl. When I open the project's web page access from my Safari browser, it should add the token to the installer file name and download the installer to my computer, ideally without requiring me to input much information each time, since this is for testing. 

I would appreciate it if you would first get the test project code working for the Windows client installer before we start testing for the Mac. You will need to develop and test this server code anyway. And assuming we want thorough alpha testing of this code, the test project will be needed by the alpha testers.

Another question: what happens for existing BOINC users who try to add a new project by creating an account on the project's web page rather than using the Attach Wizard within the BOINC Manager? At the very least, I would think the signup procedure on the web site should ask if they are already running BOINC and tell them to use the Attach Wizard if they answer yes.

The BOINC Manager has included code for several years to support a different approach for automatically attaching to projects using web browser cookies, originally implemented for GridRepublic; that is the reason that the rather large SQLite3 is part of the Manager, since some supported browsers (e.g., Firefox) use it for storing cookies. Please remind me why David and Rom began developing this installer file name token approach instead.

CharlieFenton:
> The Safari browser automatically inflates and discards the zip file, so the user only sees "BOINC Installer.app", and there is no way I can think of for the installer to get the name of the original zip file. So the project web site must add the token to the installer name (which is actually the name of the top level directory of the bundle) and then compress the bundle into a flat file for downloading,

I think I have thought of a way around this problem, so that the server can simply append the token to the compressed flat file name, by compressing the bundle / directory tree as a disk image file (with extension ".dmg"). I will investigate this approach further.

TheAspens:
@davidpanderson @CharlieFenton - I wanted to check on the scope of this pull request.  It is my understanding that neither the Windows, Linux or Mac installers have yet been modified to enable this feature and that work will be done in an additional branch/pull request.  Is it correct to assume that the scope of this pull request is to simply enable the reading of the installer_filename.txt file and take the contents of that file and properly connect to a project that runs the code from server side version of this work?

If that is the case, it might be worthwhile doing the work for each installer in its own branch since it is independent work.  Let me know what the plan is.

TheAspens:
@JuhaSointusalo - nice catch.

With David's additional implementation I've tested the following scenario:

- I have a project running that includes the latest login_token_lookup.php code that is properly setup with versions.xml and a project id.
- I've built the client with the latest code in this branch. I've setup the all_projects_list.xml to have an entry for my test project.
- I've registered an account and visited the /download.php page on my new project which means that I have a new account with a login token.
- I created a installer_filename.txt file that is setup correctly but has a invalid token (to simulate what happens if the user uses the installation file the next day or after they visited the download page again but uses an older download).

The add project dialogue does not display even though the project does not auto-attach.

I recommend the following changes/have the following questions:

1. After any failure scenario, the installer_filename.txt should be deleted (this includes any iusses parsing the file and any failed response from the server).
2. The code should be written so that the existing behavior (i.e popup of attach project dialogue) occurs when there is an invalid file (either invalid because it is corrupted or because the contents are not valid).
3. The manager appears to be doing a read of a local file.  In the case that the manager is used to connect a remote host this could cause odd results since the file could be present locally but not remotely.  Should the manager need to query the core client for the status of the file?

davidpanderson:
The installer changes will be done in different branches.

davidpanderson:
I added code that deletes the filename file in all error cases.
Since this is done before handling GUI RPCs, the Manager will show the attach wizard.

This feature is for the Win and Mac install process.  Doesn't apply to remote control situations.

CharlieFenton:
> I think I have thought of a way around this problem, so that the server can simply append the token to the compressed flat file name, by compressing the bundle / directory tree as a disk image file (with extension ".dmg"). I will investigate this approach further.

Unfortunately, this did not work. I had hoped that when the Macintosh disk image (*.dmg) file is renamed, then the parent directory of the mounted disk image (actually the volume name of the simulated disk) would have that new name. But the volume name of the simulated disk remains unchanged. So I have no solution better than what I wrote earlier:

> So the project web site must add the token to the installer name (which is actually the name of the top level directory of the bundle) and then compress the bundle into a flat file for downloading

CharlieFenton:
The permissions / owners of files in the Mac version of BOINC can inadvertently get corrupted, such as when a new version of the OS is installed, if a user manually copies them, etc. Although this happens rarely, it does happen. For this reason, I have written the Mac BOINC installer so it can be run again and it will fix the BOINC files' permissions and ownership. 

When the installer is run again, it will of course create the file with the original token again. So it is important that the streamlined register/download/attach code in the client check whether the targeted project is already attached and ignore the token in that case. I suspect the same situation could arise if the installers for other platforms are run a second time.

davidpanderson:
It's not a problem to do the rename/zip for Mac downloads.

CharlieFenton:
> It's not a problem to do the rename/zip for Mac downloads.

Great! Then I don't foresee any problems with the implementation for the Mac. The directory to be renamed is actually one level down in the tree. For the current 7.8.3 release, the top level directory is _boinc_7.8.3_macOSX_x86_64_ which contains one file _ReadMe.rtf_ and two directories _BOINC Installer.app_ and _extras_. It would be best to add the token to the name of the _BOINC Installer.app_ directory / bundle (before the _.app_ extension), though one could add it any of these directories, or it could even be added as an additional file under the top level directory. Adding it to the name of the _BOINC Installer.app_ directory would ensure it would work even if the installer is moved out of its parent directory. (The **contents** of the _BOINC Installer.app_ directory must not be modified as that would invalidate the bundle's signature.)

The top level directory and the tree it contains would then be zipped. For example, _boinc_7.8.3_macOSX_x86_64.zip_ is what users download for the current release.

TheAspens:
I rested the scenario above with the latest from this branch.  I can verify that the installer_filename.txt is being deleted.  

However, when I launched the manager (which then started the client), the project dialogue was not displayed and since the token was invalid, the client did not attach to any project.

Here is the log

```
Thu 09 Nov 2017 04:03:48 PM CST |  | Setting up project and slot directories
Thu 09 Nov 2017 04:03:48 PM CST |  | Checking active tasks
Thu 09 Nov 2017 04:03:48 PM CST |  | Read installer filename file
Thu 09 Nov 2017 04:03:48 PM CST |  | Doing token lookup RPC to Localhost@Home
Thu 09 Nov 2017 04:03:48 PM CST |  | Setting up GUI RPC socket
Thu 09 Nov 2017 04:03:48 PM CST |  | Checking presence of 0 project files
Thu 09 Nov 2017 04:03:48 PM CST |  | This computer is not attached to any projects
Thu 09 Nov 2017 04:03:48 PM CST |  | Visit http://boinc.berkeley.edu for instructions
Thu 09 Nov 2017 04:03:49 PM CST |  | lookup token: missing info
```

CharlieFenton:
I added the Mac installer code to this branch rather than creating a new one because it depends on the definition of `INSTALLER_FILENAME_FILENAME` in _common_defs.h_ which is not yet in Master. Otherwise, I would have needed to make my new branch a sub-branch of this one, which seems a bit messy.

CharlieFenton:
This is a summary of emails I exchanged with @davidpanderson and @TheAspens regarding this PR, to document the discussion in this PR, as requested by @TheAspens. 

@davidpanderson wrote:
For the installer stuff, you don't need to look at the installer filename.
Just copy the file _installer_filename.txt_ (which will already exist in the .zip) into the BOINC data dir. It's simpler this way.

@charliefenton replied:
The disadvantage of including the filename _installer_filename.txt_ file in the zip file (in the same parent directory as the _BOINC Installer.app_ directory / bundle) is that it won't work if the user moves _BOINC Installer.app_ out of its parent directory before running it. 

That is why I went ahead with the idea of writing the name of the renamed _BOINC Installer.app_ directory into the _installer_filename.txt_ file. That would still work even if the installer is moved out of the parent directory.

I realize this situation will happen rarely, if ever, but I wanted to mention this potential issue. I will change the code as you requested unless I hear otherwise from you. There is a nice aesthetic about not having some weird looking character string as part of the installer name (although the user may wonder what this _installer_filename.txt_ file is for in the parent directory. If you wish, we could give it a different, perhaps more descriptive name in the zip file, like _project_identifier_key.txt_ or _BOINC_setup.txt_ and the installer could rename it.)

@davidpanderson replied:
I decided to take your suggestion and change the filename
to "account_data.txt".
This is more descriptive and less likely to confuse users.

@charliefenton wrote:
I have updated the Mac installer in the dpa_autologin_client branch, and uploaded a new build with the zip file name _boinc_111217_macOSX_x86_64.zip_ to <http://boinc.berkeley.edu/dl>. Here are a few things to note:

It checks whether the project server provided the _installer_filename.txt_ file by putting it in the same parent directory as our installer bundle (the bundle's default name is _BOINC Installer.app_). If yes, copy the provided file into the BOINC Data directory. Otherwise create the _installer_filename.txt_ file and write the name of our installer bundle in it.

This allows you to implement either approach on the server. But please note that it looks for the the _installer_filename.txt_ file in the _same_ directory as the _BOINC_Installer.app_ bundle, (typically something like _boinc_7.8.3_macOSX_x86_64_), while your current server implementation puts it one directory _above_ that.

I have a number of issues to report from my first test of the test project auto-attach with the Mac:

When I tried the test project, the server sent me a file named _boinc_7.8.3_macOSX_x86_64.zip_, but its contents were inside a new directory named _boinc_7_ as follows:

```
boinc_7 (directory)
    installer_filename.txt (file)
    boinc_7.8.3_macOSX_x86_64 (directory)
         BOINC Installer.app (directory / bundle)
         Readme.rtf (file)
         extras (directory)
              Uninstall BOINC (directory / bundle)
              COPYING.txt (file)
              COPYING LESSER.txt (file)
              COPYRIGHT.txt (file)
```
In other words, the _installer_filename.txt_ file was now inside a new directory named _boinc_7_ along with the contents of the original zip file. The original top-level directory _boinc_7.8.3_macOSX_x86_64_ is now inside the new _boinc_7_ directory.

The _installer_filename.txt_ file should be inside the _boinc_7.8.3_macOSX_x86_64_ directory, alongside the _BOINC Installer.app_ directory / bundle. This needs to be fixed, and the _boinc_7_ directory needs to be eliminated. I ran a few experiments using the zip command to add a file to the archive, and found no way to accomplish what is needed directly. So you may need to inflate the zip file, add the _installer_filename.txt_ file inside the _boinc_x.x.x_macOSX_x86_64_ directory, and zip the _boinc_x.x.x_macOSX_x86_64_ directory again. Or you can inflate, add the token to the installer name, and zip again.

@davidpanderson replied:
I fixed concierge.php so it puts the file in the right place in the zip.

@charliefenton wrote:
When I moved the _installer_filename.txt_ file inside the _boinc_7.8.3_macOSX_x86_64_ directory and ran the installer, this is what the client reported:
```
12-Nov-2017 03:17:41 [---] Checking active tasks
12-Nov-2017 03:17:41 [---] Read installer filename file
12-Nov-2017 03:17:48 [---] Doing token lookup RPC to Quake Catcher Network
12-Nov-2017 03:17:48 [---] Setting up GUI RPC socket
12-Nov-2017 03:17:48 [---] Checking presence of 0 project files
12-Nov-2017 03:17:48 [---] This computer is not attached to any projects
12-Nov-2017 03:17:48 [---] Visit http://boinc.berkeley.edu for instructions
```
The content of the installer_filename.txt file sent by the server was:
`__1_1373_c8c78bdf`

The RPC returned http_op_retval of `-224 (ERR_HTTP_PERMANENT)`, and the _installer_filename.txt_ file was not deleted. I'm guessing this error occurred because the QCN server does not recognize this RPC. The Manager did not display the _Add Project_ dialog (presumably because the _installer_filename.txt_ file exists. 

@davidpanderson replied:
We need to test against a project that exports the RPC (not QCN)

@charliefenton wrote:
But even if the file were deleted after the RPC failure, the Manager would still not display the Add Project dialog because `autoattach_in_progress()` is called only when the manager first connects to the client, in `CAdvancedFrame::OnConnect()`. So the only indication was in the Event Log, which most users don't see, and new users almost certainly won't see.

Also, I believe this same logic needs to be in `CSimpleFrame::OnConnect()`, especially because the Simple View is selected by default in new installations of BOINC. 

Finally, I thought that the convention is that the Manager never accesses the client's files directly; it requests all such access through RPCs. But `autoattach_in_progress()` violates that convention.

Note that according to _all_projects_list.xml_, 1 is the ID of the Quake Catcher Network. For the auto-attach of the test project to work, I believe we need to add  the test project to the _all_projects_list.xml_ file written by the BOINC installer. 

But there is a potential issue here. When installing over an existing client, the installer does not replace the existing _all_projects_list.xml_ file, because the one already in the BOINC Data directory may be newer than the one in the installer, and the file will get updated at some point. This won't be a concern as long as the auto-attach is only used with virgin BOINC installations. But although the server web site says not to use the buttons if you already have BOINC installed, you can't count on people to follow the directions.

One possible way to solve this would be for the client to download the latest  _all_projects_list.xml_ as the first step in handling an auto-attach token.

Another concern: if users will be downloading the BOINC installer from project web sites, how can we be sure the project servers are kept up to date with the latest version of BOINC? The recent urgent hot fix release of BOINC 7.8.4 for the Mac is a case in point. Many (perhaps most) users might not have noticed that all their GPU tasks were failing because of the wrong GPU memory size reported by 7.8.3 for the Mac, especially if they use Simple View. If projects had been distributing 7.8.3, how could we be sure they would quickly upgrade their servers to distribute 7.8.4?

Could there be a chron job in the project server code to download and update the XML version list a couple of times a day? This seems like the sort of thing that should be easy to automate.

@davidpanderson replied:
I added a script to do this (branch dpa_autologin2)

@charliefenton wrote:
I notice that there is a second button _Download BOINC + VirtualBox for MacOS X ..._ and underneath the two buttons it says:
>When the download is finished, open the downloaded file to install BOINC and VirtualBox. 

But we don't offer the BOINC + VirtualBox installer option for the Mac, because we have no way to uninstall VirtualBox on the Mac if they later upgrade to a newer version. I described the problem in comments in the _mac_build/release_boinc.sh_ script in the GIT repository on May 7, 2015. 

@davidpanderson replied:
I fixed this

CharlieFenton:
This will need will need to have special temporary code for testing:
* The test project server needs to download the test builds of the client installers instead of the current release builds
* The test clients will need to use a test version of the all_projects.xml file which includes the test project, rather than the standard all_projects.xml file.

TheAspens:
Branch dpa_autologin_client now contains ALL the code related to the autologin feature.  I've updated the PR name to reflect that this pr is for the whole solution.  All remaining resolution of issues and tidying up of loose ends can now be resolved from this branch.

TheAspens:
@davidpanderson wrote 

> The way things work now:
> the client reads and parses the account info file,
> and initiates a lookup-token RPC,
> before it handles any GUI RPCs.
> If any of these things fails it deletes the file.
> 
> The Manager does a couple of GUI RPCs,
> then checks for the existence of the account info file
> and puts up the attach wizard if it's not there.
> 
> In theory this should handle all cases correctly except one:
> namely, when the RPC fails, e.g. the token has expired or the project is down.
> 
> I think the solution to this is for the client to not handle GUI RPCs
> until the lookup-token RPC has finished.
> I'll implement this when everything else is working.

This can be implemented in this branch.

davidpanderson:
Kevin:
I'm not seeing the boinccas project in the VS2010 solution.
Did you merge dpa_autologin_client_setupwin?

davidpanderson:
Ignore last comment; I added the boinccas project.

TheAspens:
I think I did miss one of them in my initial merge (didn't pull after merging on github).  I've merged from master again.  Let me know if you see anything missing.

TheAspens:
@CharlieFenton  @davidpanderson 

Travis indicates that the build failed for the mac installer due to 

▸ Compiling PostInstall.cpp
❌  /Users/travis/build/BOINC/boinc/mac_installer/PostInstall.cpp:347:22: use of undeclared identifier 'INSTALLER_FILENAME_FILENAME'
        tempDirName, INSTALLER_FILENAME_FILENAME);
                     ^
** BUILD FAILED **
	CompileC build/boinc.build/Deployment/PostInstall.build/Objects-normal/x86_64/PostInstall.o /Users/travis/build/BOINC/boinc/mac_installer/PostInstall.cpp normal x86_64 c++ com.apple.compilers.llvm.clang.1_0.compiler

TheAspens:
Actually - just a variable change based on David's code.  I made the change and pushed.

TheAspens:
While testing I noticed issue #2237.   Based on discussion earlier this won't block merge once I finish testing, but it will need to be fixed.

TheAspens:
Whilte testing I noticed issue #2238.  Not blocking merge, but something that needs to be reviewed and fix if necessary.

TheAspens:
While reviewing, I noticed some code that I think can be removed.  See issue #2239.  This would remove @CharlieFenton concern about the manager reading local files.  This won't block merge but should be considered and addressed.

TheAspens:
The travis CI job failed for issues on its side.  I'm closing and re-opening this pull request to retrigger the builds.

TheAspens:
I've opened issues for everything I've found with my testing and none of them block a merge.  The feature appears to work quite well - thanks for the implementation @davidpanderson.

@JuhaSointusalo  and @CharlieFenton also reviewed and raised some questions.  Can you guys please indicate if your questions are resolved?  

Note that I recommend we adhere to the proposed standard for code reviews listed here: https://github.com/BOINC/boinc-policy/blob/knr-code-reviews/Expectations_of_Code_Review.md

Anything that doesn't block a merge based on that criteria but should still be fixed should have an issue opened with the details.

TheAspens:
I plan to merge this code tomorrow unless I hear otherwise from anyone.

CharlieFenton:
I have opened issues #2240, #2241 and #2242. I will leave it up to @TheAspens to decide whether these need to be resolved before merging. 

I would also suggest using the term auto-attach rather than auto-login, as the code does not log the user in (at least not from the user's perspective) but rather automates the attach process. The user has manually created an account on the project's web site and in the process logged in to the project manually.

CharlieFenton:
I have built a new Mac installer from the latest source files in this branch and uploaded the zipped installer package as _boinc_111517_macOSX_x86_64.zip_ to http://boinc.berkeley.edu/dl. I have tested with this build and posted the results as a comment in issue #2238 because of their relevance to that issue.

TheAspens:
Ok based on my review and opinion of the issues open from this PR, none block the merge based on the criteria from: https://github.com/BOINC/boinc-policy/blob/master/Expectations_of_Code_Review.md

Recap of status and open issues:

- #2237  is an issue on the boinc website for download_all.php.  I've opened a pull request at https://github.com/BOINC/boinc-site/pull/2 for a proposed fix.
- #2238 adds misleading messages to users.  It should be fixed before 7.10 release, but won't block merging this.  I've added it to the 7.10 release project for tracking.
- #2239 is clean up of code that might not be required any more.  It should be reviewed and resolved before 7.10 release.  I've added it to the 7.10 release project for tracking.
- #2241 is a reminder to remove the changes made for integration testing of this feature.    I've added it to the 7.10 release project for tracking.
- #2240 addresses the issue of an auto-attach to a project not listed in the all_projects_list.xml file contained in the client build.  This should be resolved at some point.

I'll wait a couple of more hours then merge this if no-one objects.

CharlieFenton:
After discussion with @davidpanderson, I have modified the Async RPC code to increase the delay before displaying the "Communicating with client" dialog from 1.5 seconds to 60 second while the client is auto-attaching to a project. This allows more than ample time for the auto-attach before GUI RPCs are enabled. But we do display it after 60 seconds as a safety feature, so that the user can exit BOINC if the client hangs.

TheAspens:
Since this pull request has been merged, I've opened a new pull request against the branch to reflect the continued work.  Since it is a work in progress I've added the [WIP] prefix so that no-one will merge it.  I will copy Charlie's comment above into the new pull request.

PR #2220, opened 2017-11-08
Drupal: Allow for use of a read-only replica BOINC project database

drshawnkwang:
Changes to settings.php

1. Setup settings.php to use a settings.local.php where most site-specific configuration can be set, .e.g, baseurl.
1. The current 'boinc' database has been renamed 'boinc_rw'.
1. Added logic to allow for a third database, 'boinc_read' which will be used for read-only queries. This is most often configured to point to a replicate database.
1. If no third database 'boinc_read' is specified, then drupal will use 'boinc_rw' for both read-only queries and write queries.

Changes to Drupal-Boinc codebase

1. Changed instances of `db_set_active()` function to use either 'boinc_rw' or 'boinc_read' depending on queries. Any read-only query as well as Drupal Views use 'boinc_read', while others use 'boinc_rw'

https://dev.gridrepublic.org/browse/DBOINC-60


tristanolive:
This branch has a conflict since `boincteam_members_email_list()` has been removed in ef521b6d, therefore the `db_set_active()` call does not need to be updated within that function. This will be simple to resolve on merge, once testing is complete.

PR #2221, opened 2017-11-09
file_deleter: use new XML parser for wu/result XML docs

davidpanderson:
Allows arbitrary formatting of input, output templates


TheAspens:
This approach looks fine to me.  I haven't tested formally, but I assume that David has.  I'm merging and closing this pull request.

PR #2222, opened 2017-11-09
Dpa autologin client setupwin

davidpanderson:


CharlieFenton:
Would it not be better practice to `#include "filenames.h"` and write `FILE* f = fopen(INSTALLER_FILENAME_FILENAME, "w");` instead of `FILE* f = fopen("installer_filename.txt", "w");`? 

I believe this would require making this branch off of the _origin/dpa_autologin_client_ branch rather than off of _master_, so that `INSTALLER_FILENAME_FILENAME` would be defined in _filenames.h_, or just making this change in the _origin/dpa_autologin_client_ branch instead of creating a separate branch for it. I suggest making the change in _origin/dpa_autologin_client_, as I don't understand the advantage of creating this separate branch.

I am ready to make the corresponding change to the Mac installer, but I want to use `INSTALLER_FILENAME_FILENAME`, so I will wait until I se how you handle it for MSWindows.

davidpanderson:
Yes.  I'll do that later.

BTW, it occurred to me that the Mac case is easy.  Since I'm already creating a new .zip on the server, I can directly add the "installer_filename.txt" file to that zip.  No need to rename anything, or to have the installer find its own name.

CharlieFenton:
> Since I'm already creating a new .zip on the server, I can directly add the "installer_filename.txt" file to that zip. No need to rename anything, or to have the installer find its own name.

Yes, that would work, but the installer will still need to copy the file into the BOINC Data directory where the client can find it. Since the installer already gets the full path to itself, it is also a trivial change for the installer to write its own name to the file in the BOINC Data directory, so I might as well do that.

CharlieFenton:
If I understand correctly, there are two ways a user might download the installer: from the project, with the token added to the installer file name, and directly from Berkeley, in which case there would be no token. The code in _acct_setup.cpp_ assumes that there is no token only if the _installer_filename.txt_ file does not exist. Otherwise, the missing token is treated as an error.

How should the installer (for MS Windows and perhaps other platforms) decide whether to not to create the _installer_filename.txt_ file? Should it just check for the double underscore in the installer's name?

davidpanderson:
If the filename doesn't have the right syntax, the client ignores and deletes it.  I'll take out the error message.

TheAspens:
closing and reopening this pull request to retrigger the checks

TheAspens:
@davidpanderson - this pull request looks ready to merge.  The one open question from above is the "INSTALLER_FILENAME_FILENAME" issue that you and charlie were discussing.  Do you want to

A) Merge this pull request now and once PR #2207 can be merged you will then create a new pull request to iuse the constant.
B) Wait for PR #2207 to be merged to master, then we can merge master into this branch and you can update to use the constant

There are probably other ways to do this but wanted to see how you wanted this handled.

davidpanderson:
Either way.  I'd prefer to merge them both, then clean up various loose ends.

TheAspens:
Issue #2236 will done as part of PR #2207

Issue #2224, opened 2017-11-09
download.php doesn't set user_id


TheAspens:
@JuhaSointusalo reported in merged pull request: https://github.com/BOINC/boinc/pull/2202

> html/user/download.php
> ```
> 126 +        ',
> 127 +        $project_id,
> 128 +        $token,
> 129 +        $user->id,
> ```
> 
> There is no user_id field in the HTML above.

This needs to be resolved so that the download will work correctly.

PR #2225, opened 2017-11-09
Fix missing user_id on download page

TheAspens:
Fixes issue #2224 and fixes bug where the download page renders a broken vbox download button when a platform doesn't have a vbox version (the fix is to not render the button)


TheAspens:
@davidpanderson and @JuhaSointusalo  - do one or both of you mind taking a look at this?   Thanks!

davidpanderson:
I fixed this in a slightly better way in dpa_autologin2.

TheAspens:
@davidpanderson - thanks.  I thought the one box was slightly off center before - that should fix it.

PR #2234, opened 2017-11-14
Add script to get versions; don't show VBox button if no VBox version

davidpanderson:


html/ops/get_versions.php
```diff
@@ -0,0 +1,10 @@
+#! /usr/bin/env php
+<?php
+
+// get XML client version list from BOINC server
+
+file_put_contents(
+    "../user/versions.xml",
+    file_get_contents("http://boinc.berkeley.edu/download_all.php?xml=1")
```
TheAspens: Should use https not http

TheAspens:
I've also opened issue #2235.  However, as discussed over email, I'm going to merge this code into master, merge master into branch dpa_autologin_client and testing and work on this feature can be finished in that branch & pull request.

PR #2243, opened 2017-11-15
Hw virtualization detection

nikolasgianna:
Detect whether hardware virtualization is enabled and if not don't boot at all. Here the list hostinfo vboxmanage command is used.


marius311:
I'm not keeping a close eye so my apologies if this PR was discussed elsewhere and I missed it. What's the main motivation for changing the current behavior? Also, does this interfere with the current method of reporting to the BOINC client that extensions are disabled and thus no more Vbox 64bit jobs should be requested via the ERR_CPU_VM_EXTENSIONS_DISABLED string?

lfield:
No it does not interfere with the current method. It is just an addition protection for when it fails.

TheAspens:
@marius311 - do you use vbox right now?  If so, do you mind building and testing from @nikolasgianna branch?  If you are able to test and confirm that it works, then I can merge.  Thanks!

samples/vboxwrapper/vbox_common.cpp
```diff
@@ -463,6 +463,36 @@ string VBOX_BASE::read_vm_log(){
...
+            boinc_finish(1);
+	}
+	else if (vmhw == "yes")
+            vboxlog_msg("HW virtualization enabled.");
+    }
+    else {
+        vboxlog_msg("ERROR: Cannot read host info.");
+        boinc_finish(1);
+    }
+}
```
TheAspens: can you update your code so that it follows the BOINC code style guidelines: https://boinc.berkeley.edu/trac/wiki/CodingStyle

In particular your use of braces and if statements should be made consistent with the code style.  Thanks!

samples/vboxwrapper/vbox_common.h
```diff
@@ -209,6 +209,7 @@ class VBOX_BASE : public VBOX_JOB {
         virtual void dump_vmguestlog_entries();
         virtual int dump_screenshot();
         virtual std::string read_vm_log();
+	virtual void detect_hw_virt();
```
TheAspens: Can you fix the space indent on this?

marius311:
@TheAspens I'll try to take a look over the weekend, or if not early next week.

TheAspens:
That would be great - Thanks!

marius311:
Hi guys, I checked this out, I think there's a couple of problems with this approach:

1) `VBoxManage list hostinfo` will only tell you if the host supports virtualization, not if it is enabled (at least that's the behavior on my computer). Thus the error messages in this PR "HW virtualization not enabled" and "HW virtualization enabled" need to be rewritten to be less misleading. 

2) More importantly, I believe this *does* interfere with the current method of reporting to the server that a host has extensions disabled via the ERR_CPU_VM_EXTENSIONS_DISABLED string. Basically if your `detect_hw_virt` calls `boinc_finish` because if detects no extensions, then we never make it to these lines https://github.com/BOINC/boinc/blob/7db5260f87998c90e31fffd38c6f708db79887c0/samples/vboxwrapper/vboxwrapper.cpp#L882-L886 which print ERR_CPU_VM_EXTENSIONS_DISABLED on the log and thus never let server know of this fact. This means these computers will continue getting VBox jobs despite that they can't run them. 

Let me know what you think (maybe @romw can also chime in?), but otherwise I don't think this is OK to merge.

marius311:
Actually, unless I'm mistake it appears to me that the ERR_CPU_VM_EXTENSIONS_DISABLED functionality has been broken since https://github.com/BOINC/boinc/commit/0d7ecc6a168edf483f529bd1acfadf3125dde134. We should definitely fix that.

On the topic of this PR though, I still don't really totally understand what the motivation for it is unfortunately. We already have a method to detect if virtualization is disabled. @lfield mentions this is backup "for when [that] fails" but do you have examples of such failures? Without that, it seems like unnecessary complication of the already pretty complex logic of vboxwrapper.

Its also possible this is trying to fix something which is more directly fixed by undoing the bug introduced in the above commit, in which case we should do that instead. 

In any case, curious to hear your guys' opinions.

nikolasgianna:
When testing on my hosts the behaviour is the intended (detection of whether virtualisation is enabled currently and not just if it is supported). Could you please tell me  what system you are testing on and how so I can reproduce and investigate?

marius311:
What I can say is that my processor is an Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz and I'm running Linux kernel 4.4.0-112-generic and with VT-x turned off in bios `VBoxManage list hostinfo` still says `Processor supports HW virtualization: yes`. I haven't been able to track down documentation for that VBox command but certainly their phrasing "*supports*" rather than "*enabled*" is suggestive. I could definitely imagine the behavior depending on the CPU somehow though.

nikolasgianna:
I have now tested on two more hosts and I am still getting the expected behaviour. I am now searching for the documentation of the command and in case I cannot find it I will have to go into the source code to understand how it determines the virtualization state.

JuhaSointusalo:
This looks like something that could be copied to client to reset `p_vm_extensions_disabled` so I checked just how the code works.

VBoxManage reports either 'yes' or 'no' to HW virtualization. The code queries kernel driver for the information and if the driver is not loaded or otherwise not available it falls back to CPUID.

VBoxManage [`listHostInfo`](https://www.virtualbox.org/browser/vbox/trunk/src/VBox/Frontends/VBoxManage/VBoxManageList.cpp#L194)
[`Host::init`](https://www.virtualbox.org/browser/vbox/trunk/src/VBox/Main/src-server/HostImpl.cpp#L295) and [`Host::getProcessorFeature`](https://www.virtualbox.org/browser/vbox/trunk/src/VBox/Main/src-server/HostImpl.cpp#L987)
driver [`SUPR0QueryVTCaps`](https://www.virtualbox.org/browser/vbox/trunk/src/VBox/HostDrivers/Support/SUPDrv.cpp#L4382) (called through `SUPR3QueryVTCaps`)

I think the re-check code in `Host::getProcessorFeature` has a bug that makes it report 'yes' even when HW virtualization is disabled.

So 'no' is reliable and 'yes' maybe is maybe not.

There are two messages in VBox log that tells HW support is not available:
```
WARNING! 64-bit guest type selected but the host CPU does NOT support HW virtualization.
HM: HMR3Init: Falling back to raw-mode: VT-x is disabled in the BIOS for all CPU modes
```

The first one ([`Console::i_configConstructorInner`](https://www.virtualbox.org/browser/vbox/trunk/src/VBox/Main/src-client/ConsoleImpl2.cpp#L1029)) looks like it could have the same problem as VBoxManage but the second one (HW manager [`HMR3Init`](https://www.virtualbox.org/browser/vbox/trunk/src/VBox/VMM/VMMR3/HM.cpp#L654)) should be reliable.

Right now vboxwrapper checks for error codes in `VBOX_BASE::is_logged_failure_vm_extensions_disabled` and `VBOX_BASE::is_logged_failure_vm_extensions_not_supported` but the functions could be changed to also check for the full error string.

There is also [`HWVirtExPropertyType_Force`](https://www.virtualbox.org/sdkref/_virtual_box_8idl.html#ae90b543d84f42a1e477aafa7303079cb) which should remove any need for guesswork. Unfortunately that can be set only with COM API or by editing .vbox file. It is not available in VboxManage.

As for why this is needed. If you try to start 64-bit VM task without HW support VirtualBox starts the VM with 32-bit CPU and 64-bit kernel fails to boot and hangs. As far as VirtualBox and vboxwrapper knows everything is fine until the VM fails to send heartbeat. Rinse and repeat.

lfield:
I have just tried myself by turning off VT-x in the bios and running` VBoxManage list hostinfo`. It says 

> Processor supports HW virtualization: yes

So I can confirm that this PR doesn't do what we expected. As pointed out there may be another issue but we should identify it first before attempting a fix.

PR #2246, opened 2017-11-16
[WIP] Early stage of websocket proxy development

nikolasgianna:
This is a first approach and attempt at creating a proxy to allow the BOINC client to understand commands issued via websockets. The proxy should run within the same executable so that the whole procedure is transparent to the volunteers. As it stands now this is a barely working prototype and should not be expected to function in any meaningful way. For that reason this pull request is not expected or intended to be merged. It is rather a way to start a conversation around the issue and to get more eyes on such an important piece of the potential future BOINC client. 

This is a small test page: https://lhcathomedev.cern.ch/lhcathome-dev/websock_test.html that will return the version of a running client that implements these changes. The communication is done solely through websockets.

Since Firefox expects Secure Websockets (because the javascript is served from an https server) in this stage you will have to create your own certificate and key with:
`openssl req -new -newkey rsa:2048 -days 365 -nodes -x509 -keyout server.key -out server.pem`
and then add it by visiting https://localhost:8443 and following the prompts. 

Note: This is only tested on Linux and should not work on Windows or macOs

### DIFF
--- a/client/Makefile.am
+++ b/client/Makefile.am
@@ -88,7 +88,9 @@ boinc_client_SOURCES = \
     thread.cpp \
     time_stats.cpp \
     whetstone.cpp \
-    work_fetch.cpp
+    work_fetch.cpp \
+    mongoose.cpp \
+    websocket_server.cpp
 
 boinc_client_DEPENDENCIES = $(LIBBOINC)
 boinc_client_CPPFLAGS = $(AM_CPPFLAGS)
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -69,8 +69,14 @@
 #include "http_curl.h"
 #include "sandbox.h"
 
-#include "websocket_server.h"
 #include "main.h"
+
+#ifndef __APPLE__
+#ifndef WIN32
+#include "websocket_server.h"
+#endif
+#endif
+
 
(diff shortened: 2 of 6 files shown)


TheAspens:
@nikolasgianna - can you put [WIP] at the front of your pull request?  That indicates to everyone that you are interested in them taking a look but that you do not want it merged.

nikolasgianna:
There is already an older version of mongoose inside /samples/gfx_html/ . As far as I can see it is using the same license.

nikolasgianna:
In the process of reviewing websocket libraries one that caught my attention (seems to be very widely used) and that seems to meet all the requirements is this: https://github.com/uNetworking/uWebSockets

Please review and if it is ok I can go on with using it instead of mongoose.

TheAspens:
@nikolasgianna - I'm not a lawyer so my statements have no legal weight.  With that dislaimer aside, the license https://github.com/uNetworking/uWebSockets/blob/master/LICENSE appears to be the zlib license and it appears that the zlib license can be used within a LGPL licensed project https://en.wikipedia.org/wiki/Zlib_License

So I would say you could probably proceed with this library for the websocket server.

PR #2248, opened 2017-11-16
Mac installer: correctly replace screensaver when installing an older…

CharlieFenton:
… version of BOINC to replace a newer version

The Mac installer was incorrectly deleting the screensaver when an older version was being installed over a newer version, due to an incorrect setting. This fixes the setting so that the same version of the screensaver is installed as the other components.


CharlieFenton:
closing #2248

Issue #2241, opened 2017-11-15
Auto-attach: special test code needed


CharlieFenton:
To test the new auto-attach code (PR #2207), special temporary or configurable code needs to be added to the server and client:

* The _account_data.txt_ file downloaded by the test project now has a project ID of 100, but there is no project with this ID in the _all_projects_list.xml_ file. The test clients will need to use a test version of the _all_projects_list.xml_ file which includes the test project, rather than the standard _all_projects_list.xml_ file.

* The test project downloads the current release build, which does not have any of the auto-attach code. The test project server needs to download the test builds of the client installers instead of the current release builds.

It would probably be best to use a #define whose value can be modified to enable or disable the test code so that it can be re-enabled for use in future testing as well as being used for initial testing.

TheAspens:
Based on the discussion in the commit, it sounds like this issue means two things should be changed just before the release branch is cut:

1. Remove the test project from the all_projects.xml list included in the distribution
2. Change the overwrite setting on the installer so that it will not replace an existing file (check this setting for both the windows and mac installers)

I'm tagging this issue in the client 7.10 release project so that it is seen for that.  

Please confirm that the above correctly reflects the agreement on this issue.

CharlieFenton:
As I wrote in my initial comment for the issue, I _very_ strongly disagree with this approach. It will require that I modify the Mac installer to overwrite an existing _all_projects_list.xml_ file, and then change it again for final release, and we will also need to do the same for the client code to prevent it from downloading a standard _all_projects_list.xml_ file during testing. Similarly, it requires change it twice in the future every time we ned to test the auto-attach with a test project, which I expect will happen from time to time. 

Sorry, I am not willing to do make that change in the Mac installer. I have proposed a different method in this PR, and am not happy that my concerns and proposal are being ignored.

CharlieFenton:
> Please confirm that the above correctly reflects the agreement on this issue.

Sorry, no. I strongly believe that a branch with temporary test code in it should never be merged into master, nor should code with open issues that require making a change later. I feel that our goal should be that all code in master could be released for alpha testing at any time with no further changes. That is definitely not the case here. I see no reason why this branch needed to be merged prematurely. All testing could have been done within this branch.

TheAspens:
Charlie - my apologies.  I clearly did not understand your proposal.  There was no intent to ignore your concerns .  

[...]

PR #2250, opened 2017-11-16
Implement auto attach

TheAspens:
This pull request is a continuation of PR #2207 which has been merged.


TheAspens:
From pull request #2207 after the merge, @CharlieFenton commented:

> After discussion with @davidpanderson, I have modified the Async RPC code to increase the delay before displaying the "Communicating with client" dialog from 1.5 seconds to 60 second while the client is auto-attaching to a project. This allows more than ample time for the auto-attach before GUI RPCs are enabled. But we do display it after 60 seconds as a safety feature, so that the user can exit BOINC if the client hangs.

CharlieFenton:
@davidpanderson: The last lines in GET_PROJECT_LIST_OP::handle_reply() are:
```
    // were we initiated by autologin?
    //
    gstate.process_autologin(false);
```
So this RPC will alway call `process_autologin(false)`, even if the RPC was not initiated by `process_autologin(true)`. This could cause different problems every time `get_project_list_op.do_rpc()` is called (typically every 14 days), depending on whether or not the static variables _project_id_ and _user_id_ were set when the client first started. 

If _project_id_ has never been set so it is not a valid project id, the Event log will display "Unknown project ID: _project_id_". 

If _project_id_ has never been set but it is a random valid project id, the client will attach to an unintended project. (Static variables are automatically initialized to 0 on the Mac; i don't know if this is true on all platforms.)

If there was an auto-attach when the client was first started, it will display "Already attached to _project_name_" .

I believe you need to add:
`if (!autologin_in_progress) return;` between lines 299 and 300 like so:
```
    } else {
        // here the get project list RPC finished.
        //
	if (!autologin_in_progress) return;
        // check whether it failed.
        //
        autologin_in_progress = false;
```

I suspect this is what you intended by the comment `// were we initiated by autologin?`

AenBleidd:
@CharlieFenton,

> If project_id has never been set but it is a random valid project id, the client will attach to an unintended project. (Static variables are automatically initialized to 0 on the Mac; i don't know if this is true on all platforms.)

Section 6.7.8 Initialization of C99 standard (n1256) answers this question:

If an object that has automatic storage duration is not initialized explicitly, its value is indeterminate. If an object that has static storage duration is not initialized explicitly, then:

— if it has pointer type, it is initialized to a null pointer;

— if it has arithmetic type, it is initialized to (positive or unsigned) zero;

— if it is an aggregate, every member is initialized (recursively) according to these rules;

— if it is a union, the first named member is initialized (recursively) according to these rules.

CharlieFenton:
@davidpanderson: i think a race condition might be possible if there is an auto-attach with a virgin install or with a new version of BOINC on top of an older version.

Say that the  local all_projects_list.xml file does contain the project id in the token, so `gstate.process_autologin(true)` does not call `get_project_list_op.do_rpc()` but it does call `lookup_login_token_op.do_rpc()`. Since the `lookup_login_token_op.do_rpc()`is asynchronous, `CLIENT_STATE::INIT()` continues and calls `all_projects_list_check()`. Because this is a new version of the client, this calls `get_project_list_op.do_rpc()`. If the _get_project_list_op_ RPC finishes before the _lookup_login_token_op_ RPC, then `GET_PROJECT_LIST_OP::handle_reply()` will call `gstate.process_autologin(false)`. I leave it to you to determine whether this unexpected re-entry of `gstate.process_autologin()` will cause any problems.

davidpanderson:
Oops!  Fixed.

davidpanderson:
Good eye!  That is indeed a problem.  I committed a fix.

CharlieFenton:
@davidpanderson: I see that the merge conflict is due to your removal of this code from client_state.cpp:
```
    if (gstate.projects.size() == 0) {
        msg_printf(NULL, MSG_INFO,
            "This computer is not attached to any projects"
        );
        msg_printf(NULL, MSG_INFO,
            "Visit http://boinc.berkeley.edu for instructions"
        );
    }
```
But I think we still want to display these messages when there is no _login_token.txt_ file and no BOINC Manager is running (which would run the _Attach Wizard_.) There is probably no harm in displaying the messages even if the BOINC Manager displays the  _Attach Wizard_. So I suggest keeping the above code, but executing it only if `gstate.autologin_in_progress` is false. This should also eliminate the merge conflict.

davidpanderson:
The first message is OK.
The 2nd should be removed or changed because the BOINC web site front page doesn't have instructions for adding projects.

CharlieFenton:
I restored the first message to be displayed only if not doing an autologin, and resolved the merge conflict. I believe this PR is now ready to be merged into Master, assuming that all checks pass.

TheAspens:
@davidpanderson  - if this is ready for final review, can you remove WIP from the title of the pull request?

@AenBleidd - was the item that you raised fixed?

@CharlieFenton  or @davidpanderson   have either of you built the server and client from the latest version branch and done testing on this or would you like me to do so before merging?

AenBleidd:
@TheAspens, actually that was just an answer on @CharlieFenton question.
No other objections from my side.

CharlieFenton:
> @CharlieFenton or @davidpanderson have either of you built the server and client from the latest version branch and done testing on this or would you like me to do so before merging?

I haven't. I am waiting for the server code to be implemented on SETI@home to do further testing, since @davidpanderson wrote in PR #2241 16 days ago:
> Once we get the server part in a vetted project (e.g. SETI@home) we can test with the standard project list. This should happen in a week or so. We won't release it for testing before then.

TheAspens:
Will Erik deploy it to Seti@Home from this branch or is the intention for him to deploy from master after it is merged?

TheAspens:
I've also now tested this and I'm ready to merge.  Everyone who has commented on this pull request is not objecting either.  @davidpanderson - if you can confirm that you want this merged now, then I will merge.

davidpanderson:
OK to merge.  SETI@home is in the process of upgrading their beta test software to master.  They'll pull this in there, and then deploy on their production server, at which point we'll be able to do and end-to-end test and do a client test/release.

RichardHaselgrove:
@SETIguy 
Eric reported that the original Beta upgrade had caused problems with their downstream code (specifically workunit generators), but that problem is now resolved: he would deploy the code on their production server "in the next few days" - which could very well mean during maintenance today, judging by the context. I'll leave #2238 (which led me here) open for review after deployment, as now.

PR #2254, opened 2017-11-21
web: improve password reset form

davidpanderson:
- call it "Reset password"
- move authenticator-based login to a new file, login_auth.php
  This is intended for admins, not users
- Use bootstrap forms


davidpanderson:
Anyone?

Issue #2255, opened 2017-11-22
Add systemd unit 


BryanQuigley:
Both Fedora[1] and Debian/Ubuntu [2] install their own systemd unit.  It would likely be better to standardize on one.

I'm currently testing one on Ubuntu based on the modified Fedora one:
[Unit]
Description=Berkeley Open Infrastructure Network Computing Client
Documentation=man:boinc(1)
After=network-online.target

[Service]
Type=simple
Nice=19
User=boinc
WorkingDirectory=/var/lib/boinc-client
ExecStart=/usr/bin/boinc
ExecStop=/usr/bin/boinccmd --quit
ExecReload=/usr/bin/boinccmd --read_cc_config
ExecStopPost=/bin/rm -f /var/lib/boinc-client/lockfile
IOSchedulingClass=idle

[Install]
WantedBy=multi-user.target

Looks like this should live in client/scripts.  TODO handle the various differences between Fedora and Debian.  Different lockfiles and boinc client names...

[1] http://pkgs.fedoraproject.org/cgit/rpms/boinc-client.git/tree/boinc-client-systemd
[2] https://packages.debian.org/sid/amd64/boinc-client/filelist

vaneldik:
Hi Brian, it would be great if you could add a line

EnvironmentFile=/etc/sysconfig/boinc-client

in the "[Service]" part. This would allow users to define environment variables again.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1517456 (opened yesterday)

BryanQuigley:
@vaneldik had to modify it to not require the file exist (and added the two other possible Environment Files).

photonxp:
Default systemd service file for boinc on Ubuntu, pasted here for reference:
1. Make sure the directory for "--dir" exists with proper permission
2. No **ExeStop**, refer to content above.

> [Unit]
> Description=Berkeley Open Infrastructure Network Computing Client
> After=network.target
> 
> [Service]
> Nice=10
> User=boinc
> PermissionsStartOnly=yes
> ExecStartPre=/usr/bin/touch /var/log/boinc.log /var/log/boincerr.log
> ExecStartPre=/bin/chown boinc:boinc /var/log/boinc.log /var/log/boincerr.log
> #ExecStart=/usr/bin/boinc --dir /var/lib/boinc-client
[...]

PR #2260, opened 2017-11-27
Add generated systemd unit

BryanQuigley:
Both Fedora and Debian have their own systemd units for boinc;
this is based on elements of both so we stop duplicating effort.

Closes: #2255

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,6 +81,7 @@ boinc_path_config.py
 pkginfo
 prototype
 client/scripts/boinc-client
+client/scripts/boinc-client.service
 config.h
 config.log
 config.status
--- a/client/scripts/Makefile.am
+++ b/client/scripts/Makefile.am
@@ -2,16 +2,28 @@
 
 install-exec-hook:
 	chmod +x boinc-client
-	$(INSTALL) -d $(DESTDIR)$(sysconfdir)/init.d
-	$(INSTALL) -b boinc-client $(DESTDIR)$(sysconfdir)/init.d/boinc-client
+	if [ -d /etc/init.d ] ; then \
+		$(INSTALL) -d $(DESTDIR)$(sysconfdir)/init.d ; \
+		$(INSTALL) -b boinc-client $(DESTDIR)$(sysconfdir)/init.d/boinc-client ; \
+	fi
+	if [ -d /usr/lib/systemd/system ] ; then \
+		$(INSTALL) -d $(DESTDIR)/usr/lib/systemd/system/ ; \
+		$(INSTALL_DATA) boinc-client.service $(DESTDIR)/usr/lib/systemd/system/boinc-client.service ; \
+	elif [ -d /lib/systemd/system ] ; then \
+		$(INSTALL) -d $(DESTDIR)/lib/systemd/system/ ; \
+		$(INSTALL_DATA) boinc-client.service $(DESTDIR)/lib/systemd/system/boinc-client.service ; \
+	fi
 	if [ -d /etc/sysconfig ] ; then \
 	  $(INSTALL) -d $(DESTDIR)$(sysconfdir)/sysconfig ; \
-	  $(INSTALL) $(srcdir)/boinc-client.conf $(DESTDIR)$(sysconfdir)/sysconfig/boinc-client ; \
+	  $(INSTALL_DATA) $(srcdir)/boinc-client.conf $(DESTDIR)$(sysconfdir)/sysconfig/boinc-client ; \
 	elif [ -d /etc/default ] ; then \
 	  $(INSTALL) -d $(DESTDIR)$(sysconfdir)/default ; \
-	  $(INSTALL) $(srcdir)/boinc-client.conf $(DESTDIR)$(sysconfdir)/default/boinc-client ; \
+	  $(INSTALL_DATA) $(srcdir)/boinc-client.conf $(DESTDIR)$(sysconfdir)/default/boinc-client ; \
 	else \
 	  $(INSTALL) -d $(DESTDIR)$(sysconfdir) ; \
-	  $(INSTALL) $(srcdir)/boinc-client.conf $(DESTDIR)$(sysconfdir)/boinc-client.conf ; \
+	  $(INSTALL_DATA) $(srcdir)/boinc-client.conf $(DESTDIR)$(sysconfdir)/boinc-client.conf ; \
 	fi
 
+clean:
+	rm boinc-client.service
+	rm boinc-client
--- a/client/scripts/boinc-client.service.in
+++ b/client/scripts/boinc-client.service.in
@@ -1,0 +1,19 @@
+[Unit]
+Description=Berkeley Open Infrastructure Network Computing Client
+Documentation=man:boinc(1)
+After=network-online.target
+
+[Service]
+ProtectHome=true
+Type=simple
+Nice=10
+User=boinc
+WorkingDirectory=~
+ExecStart=@exec_prefix@/bin/boinc
+ExecStop=@exec_prefix@/bin/boinccmd --quit
+ExecReload=@exec_prefix@/bin/boinccmd --read_cc_config
+ExecStopPost=/bin/rm -f lockfile
+IOSchedulingClass=idle
+
+[Install]
+WantedBy=multi-user.target
--- a/configure.ac
+++ b/configure.ac
@@ -1302,6 +1302,7 @@ AC_CONFIG_FILES([
                  client/win/boinc_path_config.py:py/boinc_path_config.py.in
                  client/scripts/Makefile
                  client/scripts/boinc-client
+                 client/scripts/boinc-client.service
                  db/Makefile
                  doc/Makefile
                  doc/manpages/Makefile


smoe:
@LocutusOfBorg Please have a look.

LocutusOfBorg:
I'm wondering if we won't broke something with that file...
this is the debian one 
```
[Unit]
Description=Berkeley Open Infrastructure Network Computing Client
After=network.target

[Service]
Nice=10
User=boinc
PermissionsStartOnly=yes
ExecStartPre=/usr/bin/touch /var/log/boinc.log /var/log/boincerr.log
ExecStartPre=/bin/chown boinc:boinc /var/log/boinc.log /var/log/boincerr.log
#ExecStart=/usr/bin/boinc --dir /var/lib/boinc-client
ExecStart=/bin/sh -c '/usr/bin/boinc --dir /var/lib/boinc-client >/var/log/boinc.log 2>/var/log/boincerr.log'
ExecReload=/usr/bin/boinccmd --read_cc_config
ExecStopPost=/bin/rm -f /var/lib/boinc-client/lockfile
IOSchedulingClass=idle

[Install]
WantedBy=multi-user.target
```

We have some hacks for having log file, and I would really appreciate dropping them (if the pull request approach works), and we have a different nice level, not a big deal I would say

BryanQuigley:
@LocutusOfBorg Logging works great via the systemd journal.  I actually started working on it because the default logging wasn't helpful for Issue #2256 or #2265 (no timestamps with BOINC error log).

I was hoping it would let me figure out what app was causing an issue - by easily seeing both the standard log and error - but my issues appear to be with BOINC itself.

I'm running it on Ubuntu (artful) myself, which uses your packaging (Thanks!) and it's been working perfectly (minus my two existing issues).

LocutusOfBorg:
There are few issues with this PR:
1) the @bindir@ is not getting substituted
```
dic 06 23:39:50 systemd[1]: /lib/systemd/system/boinc-client.service:14: Executable path is not absolute: ${exec_prefix}/bin/boinc
```
2) there is no install target on Makefile.am
3) there is no clean target, so the file is not getting removed on distclean/clean
4) the file is simply not working, this one works for me
```
[Unit]
Description=Berkeley Open Infrastructure Network Computing Client
Documentation=man:boinc(1)
After=network-online.target

[Service]
Type=simple
Nice=19
User=boinc
WorkingDirectory=/var/lib/boinc-client
EnvironmentFile=-/etc/sysconfig/boinc-client
EnvironmentFile=-/etc/default/boinc-client
EnvironmentFile=-/etc/boinc-client.conf
ExecStart=/usr/bin/boinc --dir /var/lib/boinc-client
ExecReload=/usr/bin/boinccmd --read_cc_config
ExecStopPost=/bin/rm -f /var/lib/boinc-client/lockfile
IOSchedulingClass=idle

[Install]
WantedBy=multi-user.target
```

BryanQuigley:
To get to a working file (for the debian packages) I have to run configure with:
./configure --enable-pkg-client --bindir=/usr/bin --sysconfdir=/etc --localstatedir=/var/lib

This seems to follow what's true for the init script as well.  Will work on install and clean targets

BryanQuigley:
Hmm.. I'm not sure how to best specify the install location.  Seems it varies, between /usr/lib/systemd/system/* or /lib/systemd/system/*

If I do libdir I'm not sure if that would mess up other items.  I could just check to see if either directory exists, but not sure if there are more options then those two.. Thoughts?

LocutusOfBorg:
``` /usr/lib/systemd/system ``` is the correct path

**Comment by LocutusOfBorg** [comment]
or probably just /lib/systemd/system, because at least in Debian /usr/lib is for graphical tools, while /lib/systemd for system tools.

**Comment by BryanQuigley** [comment]
Reorged so all you have to do is: ./configure --enable-pkg-client --prefix=/opt

Clean now works too.

**Comment by LocutusOfBorg** [comment]
+1 from me! thanks!

**Comment by BryanQuigley** [comment]
I've been running this for quite some time now without any issues (well #2256, but that's really just hidden in the old logging).

**Comment by RichardHaselgrove** [comment]
@lfield should have a look at this as well.

**Comment by JuhaSointusalo** [comment]
Normally the client run GPU apps, wrappers and non-compute intensive apps at nice level 10. If you run the client at nice level 19 that won't be possible.

Looking this from helpdesk point of view. It would be nice if all the platforms were as similar as possible and other platforms log to std{out,err}dae.txt. On the other hand, it would be nice to have timestamps in stderrdae.txt. So that's one minus point and one plus point for this PR.

Since the client runs under different user account how easy it is to access the logs? Are the logs available without switching to root or adding one's user account to boinc group? For comparison, in Debian Wheezy (the only repository install I have) the log files are world readable.

Would it be possible to have systemd handle the logs _and_ put them in log files too? (Not sure if that's a good idea though.)

**Comment by BryanQuigley** [comment]
@JuhaSointusalo 
Thanks for the feedback!

Changed nice to 10 as it does set the compute tasks to 19.

The logs are world readable:
journalctl -u boinc-client.service will get them with options like -e to go end, or -f to tail them.

for debugging I especially like:
systemctl status boinc-client.service 
Output - https://pastebin.ubuntu.com/26462611/

There are ways to have systemd output to a file instead, but I think the journal is more useful IMHO.  See the second comment ExecStart line above.

On another note, I've been researching some of the confinement options in systemd and the following might be worth investigating in the future:
ProtectSystem=strict
ProtectHome=true
NoNewPrivileges=true
ReadWritePaths=/var/lib/boinc-client

but let's get a basic one merged first :).  Lots of project testing needed for that^.

**Comment by lfield** [comment]
One difference that I can see with the [one we are using today](https://src.fedoraproject.org/rpms/boinc-client/blob/master/f/boinc-client-systemd) is the following. 

ExecStart=/usr/bin/boinc_client --daemon --start_delay 1

**Comment by BryanQuigley** [comment]
@lfield 
I haven't found any boot slowdown by not having the start_delay.

I also found another minor issue, many config files created are executable which doesn't make sense.. Also noted a mistake in my commit to fix that - will revise.

Added Home directory protection as that doesn't require any testing IMHO - if a boinc app is breaking into a home directory something is very wrong with that app.

**Comment by RichardHaselgrove** [comment]
@JuhaSointusalo - since you've looked at this before, could you do a formal review, please? We might be able to get it into v7.10

**Comment by JuhaSointusalo** [comment]
Installing to `DESTDIR` failed because the systemd directories aren't created.

**Comment by BryanQuigley** [comment]
@JuhaSointusalo What configure line did you use?  Thanks!

**Comment by JuhaSointusalo** [comment]
@BryanQuigley @lfield 

`--daemon` makes the client log to `stdoutdae.txt` and `stderrdae.txt` so that's not compatible with the idea of having systemd manage logs. The client makes a record that it is running as daemon and that is available through GUI RPC but I don't know if anyone uses that information for anything. I didn't even know about it until I now looked at what `--daemon` does.

I don't think `--start_delay 1` is necessary and if it is `Type` could be changed to `idle` to delay startup for up to five seconds.

`ProtectHome` looks ok to me. There are some people who want to have the data directory in `/home/boinc` or something like that. They'll just have to override that setting. Sidenote: A setting to change the data directory location is requested every now and then but IMHO should not be supported. The data directory location is used in `boinc` account's home directory and in `.desktop` file. Changing `/var/lib/boinc` to a symlink pointing to the real location is then a different matter.

**Comment by JuhaSointusalo** [comment]
@BryanQuigley 

I was trying to avoid compiling the client so this is a bit messy:
```
$ ./_autosetup
$ cd ../build
$ ./config.status
## this didn't create the .service file
$ ./config.status --recheck
## this ran CONFIG_SHELL=/bin/bash /bin/bash ../src/configure --disable-server --disable-boinczip CFLAGS=-g -march=native -O3 CXXFLAGS=-g -march=native -O3 --no-create --no-recursion
## still no .service file
$ cd client/scripts/
$ make
## this ran /bin/bash ./config.status
## which finally created the .service file
$ mkdir /tmp/b
$ make install DESTDIR=/tmp/b
## which fails with
if [ -d /usr/lib/systemd/system ] ; then \
                /usr/bin/install -c -m 644 boinc-client.service /tmp/b/usr/lib/systemd/system/boinc-client.service ; \
        elif [ -d /lib/systemd/system ] ; then \
                /usr/bin/install -c -m 644 boinc-client.service /tmp/b/lib/systemd/system/boinc-client.service ; \
        fi
/usr/bin/install: cannot create regular file ‘/tmp/b/lib/systemd/system/boinc-client.service’: No such file or directory
```

**Comment by BryanQuigley** [comment]
@JuhaSointusalo Thanks, updated to handle DESTDIR per what you did.

Regarding delays, I haven't been able to get this current script to show up as blamed ever by systemd-analyze for delaying boot (even in ms).

**Comment by Germano0** [comment]
Hi, one of Fedora unit file creators here.

**SELinux CONFINEMENT**

```
Type=forking
```
and the  `--daemon` flag of line
```
ExecStart=/usr/bin/boinc_client --daemon --start_delay 1
```
are both mandatory in order to permit to SELinux to confine boinc-client. We fixed SELinux confinement problems in https://bugzilla.redhat.com/show_bug.cgi?id=1303070

Other info:

> If the program daemonizes itself (i.e. forks), please use Type=forking instead.

source: [systemd.service — Service unit configuration](https://www.freedesktop.org/software/systemd/man/systemd.service.html)


**SYSTEMD CONFINEMENT**
I like the systemd confinement flags found by @BryanQuigley 
```
ProtectSystem=strict
ProtectHome=true
NoNewPrivileges=true
ReadWritePaths=/var/lib/boinc-client
```
I think I will test them.


**LOGS**
Log files positioning should be fixed in boinc-client source code as @ChristianBeer found out in discussion
https://boinc.berkeley.edu/dev/forum_thread.php?id=11011&postid=69830#69830
I had to write a patch but I did not have the time to do it.



Concerning all the other flags, in tests I made some years ago, the lines of code in [Fedora boinc-client systemd-unit file](https://src.fedoraproject.org/rpms/boinc-client/blob/master/f/boinc-client-systemd) are the only needed to let BOINC work correctly.

BryanQuigley:
**SELinux CONFINEMENT**
Confinement works in my testing with the new unit file without changes to exec lines at all. (Or at least it says system... boinc, I'm new to reading those lines)

**Default Directories**
One thing I did note is that I have the wrong working directory on Fedora (Gives a CHDIR error to binaries, which wasn't very helpful as an error message).  It's /var/lib/boinc on Fedora, while /var/lib/boinc-client on Debian/Ubuntu. 
Is there an upstream preference for the default?  Ditto for calling /usr/bin/boinc-client vs /usr/bin/boinc

**NICE**
I don't believe 19 will cause problems for applications, but it does make sense for the managing system of those apps to have higher priority.

JuhaSointusalo:
@Germano0 

I don't think `--daemon` is necessary to get the client work with SELinux. The problem was that the package was using a Bash script wrapper to launch the client. The reporter of the bug fixed the bug by not using the wrapper script and used `--daemon` to make the client log to `std{out,err}dae.txt`. Logging to `std{out,err}dae.txt` can be accomplished with `--redirectio` just as well.

Notice that the client has its own builtin log rotator. It would not surprise me much if it didn't work the way you would want when the files are symlinks.

@BryanQuigley 

I checked a few distros and Debian seems to be the only one to have the data directory in `/var/lib/boinc-client`, others use `/var/lib/boinc`. For other files there seems to be no consensus at all. The self-extracting archive (the kind-a official package) uses `boinc` as the client name.

[Unix client package layout](http://boinc.berkeley.edu/trac/wiki/UnixClientPackage) has some thoughts about packaging the client. I think it can be summarised as use `boinc` everywhere.

BryanQuigley:
Thanks for the analysis! - changed to default /var/lib/boinc then, keeping boinc as the executable name.

Changed so the working directory only needs to be changed in one place.

Germano0:
> SELinux CONFINEMENT
> Confinement in Fedora works in my testing with the new unit file without changes to exec lines at all. (Or at least it says system... boinc, I'm new to reading those lines)

@BryanQuigley could you please show us the output of
`# ps -efZ | fgrep boinc`

@JuhaSointusalo 
> I don't think --daemon is necessary to get the client work with SELinux. The problem was that the package was using a Bash script wrapper to launch the client. The reporter of the bug fixed the bug by not using the wrapper script and used --daemon to make the client log to std{out,err}dae.txt. Logging to std{out,err}dae.txt can be accomplished with --redirectio just as well.
> 
> Notice that the client has its own builtin log rotator. It would not surprise me much if it didn't work the way you would want when the files are symlinks.

Interesting, I am going to ask DaveG for a comment, since it is more expert than me on this matter and his help would be precious

BryanQuigley:
@Germano0 
[bryan@localhost ~]$ ps -efZ | grep boinc
system_u:system_r:boinc_t:s0    boinc     1245     1  0 17:12 ?        00:00:00 /usr/bin/boinc
unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 bryan 2276 2180  0 17:19 pts/0 00:00:00 grep --color=auto boinc

Germano0:
> [bryan@localhost ~]$ ps -efZ | grep boinc
> system_u:system_r:boinc_t:s0 boinc 1245 1 0 17:12 ? 00:00:00 /usr/bin/boinc

Looks fine

dave41225:
@Germano0 
Read through the PR and discussion and I don't see any problems.

The whole thing about "--daemon --start_delay 1" was to retain logging to files without using a wrapper and dropping SELinux context.

If stdout and stderr go to Journal then there is no need for --daemon and no need for logrotate scripts or symlinks. Great! Could also drop time-stamps on the output since the Journal will add them.

As to --start_delay, I think that was probably there to work around sequencing or start dependencies. Much better to solve any issues in the unit file if and when they show up.

My view: revert to a simple ExecStart with no options and use the Journal.

Just some observations:

The WorkingDirectory assignment can take a "~" that looks up the User=boinc home directory. Distributions can use the same unit file and set the path in /etc/passwd. Unfortunately that facility is not available in ReadWritePaths so additional confinement would still need hard-coded paths. Could leave that to distribution packagers.

EnvironmentFile - does boinc use any environment variables? Not used in Fedora but not sure about other distributions.

Build system paths - distribution packagers are used to relocating upstream paths to fit their distribution choices. If upstream tries to guess during the build in can cause confusion. I would prefer upstream to deliver a "standard" tar-ball and let packagers relocate as required. i.e. /usr/lib/systemd, /var/lib/boinc etc. Keeps the build system simple (and testable) too.

Thanks, all.

BryanQuigley:
@dave41225 

>The WorkingDirectory assignment can take a "~" 
Perfect, changed!

>ReadWritePaths
We're not trying for that at the items that mean we would need that. Would want to try many more projects and those that utilize things like Virtualbox, etc.

>EnvironmentFile - does boinc use any environment variables? Not used in Fedora but not sure about other 
distributions.

On another review it appears they were mostly legacy/from init scripts, so I've removed them in latest commit.

> Build system paths -
I really like it just working with systemd on sudo make install no matter where you put it.

RichardHaselgrove:
@All - the conversation seems to have ground to a halt, and no objections have been raised in the last round of contributions. Does anyone feel they are in a position to do a formal review for a potential v7.10 release?

JuhaSointusalo:
@BryanQuigley
Thanks and sorry for taking this long.

Would you be interested in creating a reference .desktop file next? #2339

Germano0:
What do you think about the comment of systemd creator, about the position of BOINC lock file?
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/7YIZQ45SEHP3YXWMFDXWRCW6SZJGGR7D/

BryanQuigley:
@Germano0 My thought is that should be moved to /run, but it needs more investigation. Maybe open an issue on it?  It also seems there is more than just one lockfile (a starting one too?).

Germano0:
> Maybe open an issue on it?
Done
https://github.com/BOINC/boinc/issues/2365

PR #2269, opened 2017-12-08
client: include platform names in account manager request

davidpanderson:
This lets AMs like TBD make better meta-scheduling decisions

### DIFF
--- a/client/acct_mgr.cpp
+++ b/client/acct_mgr.cpp
@@ -114,6 +114,7 @@ int ACCT_MGR_OP::do_rpc(
         gstate.core_client_version.release,
         run_mode_name[gstate.cpu_run_mode.get_perm()]
     );
+    gstate.write_platforms(NULL, f);
     if (strlen(gstate.acct_mgr_info.previous_host_cpid)) {
         fprintf(f,
             "   <previous_host_cpid>%s</previous_host_cpid>\n",
--- a/client/client_state.h
+++ b/client/client_state.h
@@ -405,7 +405,7 @@ struct CLIENT_STATE {
     const char* get_primary_platform();
     void add_platform(const char*);
     void detect_platforms();
-    void write_platforms(PROJECT*, MIOFILE&);
+    void write_platforms(PROJECT*, FILE*);
     bool is_supported_platform(const char*);
 
 // --------------- cs_prefs.cpp:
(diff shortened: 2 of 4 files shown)


tristanolive:
The travis build failed for probably no good reason... restarting...

tristanolive:
Still no luck. The manager is taking over 45 minutes to build and is timing out. It only took 6 minutes when building for PR 2261. I'm not familiar enough with travis to know what its problem is, but seeing as how this is a client feature, it would seem to be irrelevant. 

Has anyone seen this sort of behavior from travis elsewhere?

ChristianBeer:
It seems the OSX cache couldn't be retrieved in the last build, so it tried to build all dependencies from scratch which takes a long time. I restarted the OSX build again in the hope that the cache is retrieved.

Please note that the OSX build also builds the Client although it says `type=Manager`. The OSX build is not as flexible as the Linux build.

tristanolive:
Thanks, Christian -- I didn't even notice the Linux vs OSX designation in the job list! [Build for 2270](https://travis-ci.org/BOINC/boinc/builds/313671336) just failed also in the same place, so we'll see if your restart of this one knocks it loose... it looks like it's fetching and building openssl, unfortunately.

ChristianBeer:
Didn't work either but the build job before could retrieve the cache and it still seems intact. I restarted it again but it failed to load the cache again.

It also seems that the dependencies are build within the time limit but the actual BOINC builds takes too long afterward. I'm looking into how to error out the build after dependencies so the cache gets saved and the build will succeed on restart with the newly build cache.

ChristianBeer:
One of the restarted builds succeeded. It seems to me that it also depends on how much other jobs are running on the Travis CI OSX build hosts. Less jobs means our build is faster and finishes within the limit. I restarted the other two builds and once the build in the master branch finishes there should be a valid cache available for future jobs.

There was maintenance on the OSX infrastructure by Travis CI two days ago which seems to have invalidated the OSX caches.

ChristianBeer:
All timed out OSX builds have been restarted and finished without problems but almost triggered the 50 minute timeout again.

tristanolive:
Hopefully those caches are back in business! This was a good learning experience, thanks for walking through it.

Issue #2113, opened 2017-09-12
Add WebSockets to the BOINC client


lfield:
To be able to create a HTML GUI, the BOINC client will need to support WebSockets.

lfield:
This is being discussed in #2275

lfield:
The feedback from PR #2275 suggests:
1. Address at code comments in #2275 
1. Explicitly consent attachment to project.
1. Authenticator in the cookie should not be used for authorization 
1. gui_rpc_auth.cfg file should be used (maybe via browser callback)
1. http://127.0.0.1 can be considered secure

lfield:
This issue is essentially addressed by PR #3214

PR #2275, opened 2017-12-19
[WIP] Updated early stage of websocket proxy development 

nikolasgianna:
(Update of #2246)

This is a first approach and attempt at creating a proxy to allow the BOINC client to understand commands issued via websockets. The proxy should run within the same executable so that the whole procedure is transparent to the volunteers. As it stands now this is a prototype and should not be expected to function in any meaningful way. For that reason this pull request is not expected or intended to be merged. It is rather a way to start a conversation around the issue and to get more eyes on such an important piece of the potential future BOINC client.

This is a small test page: https://lhcathomedev.cern.ch/lhcathome-dev/websock_test.html that will return the version of a running client that implements these changes. The communication is done solely through websockets.

Since Firefox expects Secure Websockets (because the javascript is served from an https server) in this stage you will have to create your own certificate and key with (run from client directory):
`openssl req -new -newkey rsa:2048 -days 365 -nodes -x509 -keyout server.key -out server.pem`
and then add it by visiting https://localhost:3000 and following the prompts (add exception).

Prior to implementing changes this library has to be downloaded and installed: 
https://github.com/uNetworking/uWebSockets

Note: This is only tested on Linux (Debian and CentOS) and should not work on Windows or macOs

### DIFF
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -21,6 +21,11 @@
 
 #include "cpp.h"
 
+#include <uWS/uWS.h>
+#include <thread>
+#include <iostream>
+
+
 #ifdef WIN32
 #include "boinc_win.h"
 #include "sysmon_win.h"
@@ -406,6 +411,81 @@ int boinc_main_loop() {
     return finalize();
 }
 
+//
+//
+//
+int connect_sock () {
+
+    int portno = 31416;
+    struct sockaddr_in serveraddr;
+    struct hostent *server;
+    char *hostname = "127.0.0.1";
+    int sockfd;
+
+    server = gethostbyname(hostname);
+
+    sockfd = socket(AF_INET, SOCK_STREAM, 0);
+    serveraddr.sin_family = AF_INET;
+    serveraddr.sin_port = htons(portno);
+    serveraddr.sin_addr = *((struct in_addr *)server->h_addr);
+    bzero(&(serveraddr.sin_zero), 8);
+
+    int n = connect(sockfd, (struct sockaddr *) &serveraddr, sizeof(struct sockaddr));
+    if (n < 0) {
+       printf("The connect error is : %s", strerror(errno));
+    }
+    return sockfd;
+}
+//
+//
+//
+
+//
+//
+//
+void init_websocket() {
+
+    std::vector<std::thread *> threads(/*std::thread::hardware_concurrency()*/1);
+    std::transform(threads.begin(), threads.end(), threads.begin(), [](std::thread *t) {
+        return new std::thread([]() {
+
+
+            uWS::Hub h;
+
+            h.onMessage([](uWS::WebSocket<uWS::SERVER> *ws, char *message, size_t length, uWS::OpCode opCode) {
+            int sockfd = connect_sock();
+            char buf[50000];
+            memset(buf, 0, sizeof(buf));
+            snprintf(buf, /*sizeof(buf)*/ (int) length + 1, "%s", /*(int) msg.len,*/ message);
+            int num = write(sockfd, buf, (int) length + 1);
+            if (num < 0)
+            printf("The write error is: %s\n", strerror(errno));
+            new std::thread ([&sockfd, ws, &opCode]() {
+                char recv_buf[50000];
+                memset(recv_buf, 0, sizeof(recv_buf));
+                int r = recv(sockfd, recv_buf, 50000, 0);
+                if (r < 0) printf("\n\nThe error of recv is: %s", strerror(errno));
+                    std::string received = recv_buf;
+                    const char *rc = received.c_str();
+                    ws->send(rc, received.size(), opCode);
+                });
+
+           });
+           if (!h.listen(3000, uS::TLS::createContext("server.pem",
+                          "server.key")/*, uS::ListenOptions::REUSE_PORT*/))
+           {
+                std::cout << "Failed to listen" << std::endl;
+           }
+           h.run();
+
+        });
+    });
+}
+//
+//
+//
+
+
 int main(int argc, char** argv) {
     int retval = 0;
 
@@ -534,6 +614,8 @@ int main(int argc, char** argv) {
         return ERR_USER_PERMISSION;
     }
 #endif  // SANDBOX
+
+    init_websocket();
 
     retval = boinc_main_loop();
 
--- a/client/Makefile.am
+++ b/client/Makefile.am
@@ -93,7 +93,7 @@ boinc_client_SOURCES = \
 boinc_client_DEPENDENCIES = $(LIBBOINC)
 boinc_client_CPPFLAGS = $(AM_CPPFLAGS)
 boinc_client_CXXFLAGS = $(AM_CXXFLAGS) $(SSL_CXXFLAGS)
-boinc_client_LDFLAGS = $(AM_LDFLAGS) $(SSL_LDFLAGS) -L$(top_srcdir)/lib
+boinc_client_LDFLAGS = $(AM_LDFLAGS) $(SSL_LDFLAGS) -L$(top_srcdir)/lib -l:libuWS.so
 if OS_WIN32
 boinc_client_CXXFLAGS += -I$(top_srcdir)/coprocs/NVIDIA/include
 boinc_client_SOURCES += hostinfo_win.cpp \


nikolasgianna:
Any feedback would be greatly appreciated so that implementation can proceed.

TheAspens:
I will be able to take a look at this by end of day Monday.  Sorry that I missed this earlier.

TheAspens:
I attempted to build this from your repo/branch with the following commands:
./_autosetup
./configure --enable-client --enable-manager --disable-server
make

Using those commands I can build from BOINC master and produce the client.  However, when building your branch I got the following error:

  CXX      boinc_client-gui_rpc_server_ops.o
  CXX      boinc_client-hostinfo_network.o
  CXX      boinc_client-http_curl.o
  CXX      boinc_client-log_flags.o
  CXX      boinc_client-mac_address.o
  CXX      boinc_client-main.o
main.cpp:24:21: fatal error: uWS/uWS.h: No such file or directory
 #include <uWS/uWS.h>
                     ^
compilation terminated.
make[2]: *** [boinc_client-main.o] Error 1
make[2]: Leaving directory `/home/knreed/websocket/client'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/home/knreed/websocket'

If you could let me know what is needed to build, I'll try again.  Thanks!

nikolasgianna:
The μWebSockets library (https://github.com/uNetworking/uWebSockets) has to be downloaded and installed so that the header file can be found. I have not included it in my branch.

AenBleidd:
This also should be done for Windows build.
I plan to try to create bat script to prepare 3rd party libraries folder (at least for gtest for beginning).
But this should be discussed with @davidpanderson too.

nikolasgianna:
One of the most important criterion when choosing the library was the cross platform support. uWebsockets supports Windows among others. But at this stage it is only tested on Linux and in no way plug n play. When the choice is finalised it can be fully integrated.

TheAspens:
Still having troubles getting this to run.  I got uWebSockets installed and got the client to compile.  However, I had to compile with with this option: 
`./configure --enable-client --enable-manager --disable-server CXXFLAGS='-std=c++11'`

I created the SSL certificate in the client directory and started boinc and got:

`[knreed@oc7811586721 boinc]$ ./boinc
26-Jan-2018 17:38:33 [---] cc_config.xml not found - using defaults
26-Jan-2018 17:38:33 [---] Starting BOINC client version 7.9.0 for x86_64-pc-linux-gnu
26-Jan-2018 17:38:33 [---] This a development version of BOINC and may not function properly
26-Jan-2018 17:38:33 [---] log flags: file_xfer, sched_ops, task
26-Jan-2018 17:38:33 [---] Libraries: libcurl/7.29.0 NSS/3.28.4 zlib/1.2.7 libidn/1.28 libssh2/1.4.3
26-Jan-2018 17:38:33 [---] Data directory: /home/knreed/test-boinc/boinc
26-Jan-2018 17:38:33 [---] No usable GPUs found
26-Jan-2018 17:38:33 [---] Creating new client state file
26-Jan-2018 17:38:33 [---] Host name: oc7811586721.ibm.com
26-Jan-2018 17:38:33 [---] Processor: 8 GenuineIntel Intel(R) Core(TM) i7-4810MQ CPU @ 2.80GHz [Family 6 Model 60 Stepping 3]
26-Jan-2018 17:38:33 [---] Processor features: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc aperfmperf eagerfpu pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm epb invpcid_single tpr_shadow vnmi flexpriority ept vpid fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid xsaveopt dtherm ida arat pln pts
26-Jan-2018 17:38:33 [---] OS: Linux RedHatEnterpriseWorkstation: Red Hat Enterprise Linux Workstation release 7.4 (Maipo) [3.10.0-693.11.6.el7.x86_64]
26-Jan-2018 17:38:33 [---] Memory: 15.31 GB physical, 15.31 GB virtual
26-Jan-2018 17:38:33 [---] Disk: 453.25 GB total, 143.77 GB free
26-Jan-2018 17:38:33 [---] Local time is UTC -6 hours
26-Jan-2018 17:38:33 [---] Last benchmark was 17557 days 23:38:33 ago
26-Jan-2018 17:38:33 [---] No general preferences found - using defaults
26-Jan-2018 17:38:33 [---] Preferences:
26-Jan-2018 17:38:33 [---]    max memory usage when active: 7838.44 MB
26-Jan-2018 17:38:33 [---]    max memory usage when idle: 14109.18 MB
26-Jan-2018 17:38:33 [---]    max disk usage: 143.68 GB
26-Jan-2018 17:38:33 [---]    don't use GPU while active
26-Jan-2018 17:38:33 [---]    suspend work if non-BOINC CPU load exceeds 25%
26-Jan-2018 17:38:33 [---]    (to change preferences, visit a project web site or select Preferences in the Manager)
26-Jan-2018 17:38:33 [---] Setting up project and slot directories
26-Jan-2018 17:38:33 [---] Checking active tasks
26-Jan-2018 17:38:33 [---] Setting up GUI RPC socket
26-Jan-2018 17:38:33 [---] Checking presence of 0 project files
26-Jan-2018 17:38:33 [---] This computer is not attached to any projects
26-Jan-2018 17:38:33 Initialization completed
SIGSEGV: segmentation violation
Stack trace (12 frames):
./boinc[0x4b910b]
/usr/lib64/libpthread.so.0(+0xf5e0)[0x7f09d0f355e0]
/usr/lib64/libc.so.6(+0xba7a0)[0x7f09d03fd7a0]
./boinc[0x4b97b2]
./boinc[0x4aa9d4]
./boinc[0x4abb52]
./boinc[0x4ab9ad]
./boinc[0x42734f]
./boinc[0x4810bb]
./boinc[0x481965]
/usr/lib64/libc.so.6(__libc_start_main+0xf5)[0x7f09d0364c05]
./boinc[0x407599]

Exiting...
[knreed@oc7811586721 boinc]$ cd ../../websocket/
`

Running it under gdb gave this stack trace:

`(gdb) backtrace
#0  0x00007ffff57137a0 in __opendirat () from /usr/lib64/libc.so.6
#1  0x00000000004b97b2 in dir_open(char const*) ()
#2  0x00000000004aa9d4 in get_tty_list() ()
#3  0x00000000004abb52 in all_tty_idle(long) ()
#4  0x00000000004ab9ad in HOST_INFO::users_idle(bool, double) ()
#5  0x000000000042734f in CLIENT_STATE::poll_slow_events() ()
#6  0x00000000004810bb in boinc_main_loop() ()
#7  0x0000000000481965 in main ()
`

Not sure what is going on.  The compiler reported the c++11 was experimental in my version of g++ so perhaps that is the issue.  I'm on RedHat Desktop 7 running g++ (GCC) 4.8.5 20150623 (Red Hat 4.8.5-16).  What did you build and test on?

nikolasgianna:
I have tested both on Ubuntu 16.04.3 LTS and CentOS 7.4. I should have pointed I used: CXXFLAG='-std=gnu++11' and not c++11. I am not sure which GNU extension is the one necessary but it works with that flag. Sorry for not mentioning it.

TheAspens:
I apologize for the very overdue response.  I was able to get this example working.  For those who want to try this is what I did (summarizing various responses from @nikolasgianna above).

**Setup uWebSockets**
The μWebSockets library (https://github.com/uNetworking/uWebSockets) has to be downloaded and installed.  I installed the library libuWS.so in /usr/lib64/libuWS.so

**Build Client**
Cloned: https://github.com/nikolasgianna/boinc
Checked out:  git checkout websock-test
```
./_autosetup
./configure --enable-client --enable-manager --disable-server CXXFLAGS='-std=gnu++11'
```
I copied the following binaries built in the previous step and placed them into /home/knreed/test-boinc
- boinc
- boinccmd
- boincmgr

**Create self-signed ssl***
In /home/knreed/test-boinc I ran:
```
openssl req -new -newkey rsa:2048 -days 365 -nodes -x509 -keyout server.key -out server.pem
```

**Start BOINC**
In /home/knreed/test-boinc I ran:  ./boinc

**Accept Self-Signed Certificate**
I then visited https://localhost:3000 in my browser and accepted the self-signed certificate.

**Complete Test**
I then visited https://lhcathomedev.cern.ch/lhcathome-dev/websock_test.html and saw that it corrected reported the version the client was running.

AenBleidd:
This ticket still breaks Windows build: 
4>..\client\main.cpp(24): fatal error C1083: Cannot open include file: 'uWS/uWS.h': No such file or directory
BTW, what about Windows client? Should it support this feature or it is linux-only?
Also TravisCI script should be fixed too to install μWebSockets library as a dependency

client/main.cpp
```diff
@@ -21,6 +21,11 @@
 
 #include "cpp.h"
 
+#include <uWS/uWS.h>
+#include <thread>
+#include <iostream>
+
+
```
AenBleidd: Do we really need all these empty lines?

client/main.cpp
```diff
@@ -406,6 +411,81 @@ int boinc_main_loop() {
     return finalize();
 }
 
+//
+//
+//
```
AenBleidd: Empty lines

client/main.cpp
```diff
@@ -406,6 +411,81 @@ int boinc_main_loop() {
...
+    }
+    return sockfd;
+}
+//
+//
+//
+
+//
+//
+//
```
AenBleidd: Empty lines

client/main.cpp
```diff
@@ -406,6 +411,81 @@ int boinc_main_loop() {
...
+                std::cout << "Failed to listen" << std::endl;
+           }
+           h.run();
+
+        });
+    });
+}
+//
+//
+//
```
AenBleidd: Empty lines

client/main.cpp
```diff
@@ -406,6 +411,81 @@ int boinc_main_loop() {
     return finalize();
 }
 
+//
+//
+//
+int connect_sock () {
+
+    int portno = 31416;
```
AenBleidd: Maybe this should be a parameter somewhere in cc_config.xml rather than just hardcoded number?

client/main.cpp
```diff
@@ -406,6 +411,81 @@ int boinc_main_loop() {
...
+    struct sockaddr_in serveraddr;
+    struct hostent *server;
+    char *hostname = "127.0.0.1";
```
AenBleidd: Can be const I guess.

client/main.cpp
```diff
@@ -406,6 +411,81 @@ int boinc_main_loop() {
...
+
+    sockfd = socket(AF_INET, SOCK_STREAM, 0);
+    serveraddr.sin_family = AF_INET;
+    serveraddr.sin_port = htons(portno);
+    serveraddr.sin_addr = *((struct in_addr *)server->h_addr);
+    bzero(&(serveraddr.sin_zero), 8);
+
+    int n = connect(sockfd, (struct sockaddr *) &serveraddr, sizeof(struct sockaddr));
+    if (n < 0) {
+       printf("The connect error is : %s", strerror(errno));
```
AenBleidd: I can't say anything about other code but this line will not work on Windows. You should use GetLastError() instead of errno ( https://msdn.microsoft.com/ru-ru/library/windows/desktop/ms679360%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396 )

client/main.cpp
```diff
@@ -406,6 +411,81 @@ int boinc_main_loop() {
...
+//
+//
+//
+
+//
+//
+//
+void init_websocket() {
+
+    std::vector<std::thread *> threads(/*std::thread::hardware_concurrency()*/1);
```
AenBleidd: There is no c++11 support on Windows client now and this function can fail while building on Visual Studio 2010. I know that we plan to support Visual Studio 2017 but there is no plans (as far as I know) to drop support of Visual Studio 2010. If we still want to support Visual Studio 2010 this must be compiled using it (unfortunately, current AppVeyor script use Visual Studio 2013 for building) or use boinc own implementation of threads.

client/main.cpp
```diff
@@ -406,6 +411,81 @@ int boinc_main_loop() {
...
+        return new std::thread([]() {
+
+
+            uWS::Hub h;
+
+            h.onMessage([](uWS::WebSocket<uWS::SERVER> *ws, char *message, size_t length, uWS::OpCode opCode) {
+            int sockfd = connect_sock();
+            char buf[50000];
+            memset(buf, 0, sizeof(buf));
+            snprintf(buf, /*sizeof(buf)*/ (int) length + 1, "%s", /*(int) msg.len,*/ message);
```
AenBleidd: possible buffer overflow. you buf variable is 50000 bytes long but you use input variable length that can be more than 50000

client/main.cpp
```diff
@@ -406,6 +411,81 @@ int boinc_main_loop() {
...
+            int num = write(sockfd, buf, (int) length + 1);
+            if (num < 0)
+            printf("The write error is: %s\n", strerror(errno));
```
AenBleidd: Windows has no errno constant (take a look at my comment above)

client/main.cpp
```diff
@@ -406,6 +411,81 @@ int boinc_main_loop() {
...
+            new std::thread ([&sockfd, ws, &opCode]() {
+                char recv_buf[50000];
+                memset(recv_buf, 0, sizeof(recv_buf));
+                int r = recv(sockfd, recv_buf, 50000, 0);
+                if (r < 0) printf("\n\nThe error of recv is: %s", strerror(errno));
```
AenBleidd: Windows has no errno constant (take a look at my comment above)

client/main.cpp
```diff
@@ -406,6 +411,81 @@ int boinc_main_loop() {
...
+                    std::string received = recv_buf;
+                    const char *rc = received.c_str();
+                    ws->send(rc, received.size(), opCode);
+                });
+
+           });
+           if (!h.listen(3000, uS::TLS::createContext("server.pem",
+                          "server.key")/*, uS::ListenOptions::REUSE_PORT*/))
```
AenBleidd: Is it ok to have key names hardcoded?

TheAspens:
@AenBleidd - It is my understanding that this is a very early proof of concept rather than a serious pull request.  I don't believe that the @nikolasgianna expected to have this merged in its current form but rather wanted to explore the concept and let people see it in real life and the discuss if the concept as a whole should be pursued.

AenBleidd:
In any case I've pointed places that will not definitely work on Windows. I guess it is important to keep in mind this while working on final implementation

TheAspens:
Definitely useful - thanks for doing it.  I just wanted to make sure you knew that this wasn't immediately bound for real world use.

lfield:
Here is some initial feed back to the concerns:
>     Any website in the world would be able to connect to the core client. How do we prevent malicious websites from doing harm to the users machine or pretending to be a legitimate BOINC website?
The only sensitive operation is connecting to a project as all other operations require the authenticator. By default only connecting to official projects in the white list will be allowed. Additional projects can be manually added to the client configuration to support development activities.
>     If we allow this interface to perform actions such as attach or cancel a task then how do we ensure that only actions approved by user for that particular boinc project are performed?
As far as I understand all operations require the authenticator. This is stored in a cookie after doing the Web login. 
>     How do we install a SSL certificate that would be trusted by browers so that these communications would be allowed? Most SSL certificates are only valid for 2-3 years while users often leave BOINC installed for longer than that. What happens when the certificates expire?
This can be a self-signed certificate generated during the client installation. A clean method is required to add this to the trusted certificates in the browser configuration.
>     Increased attack surface. This would allow attacks on the core client to come from a larger attack surface than currently exists (and thus more likely for an exploit to be found).
The RPC already exists so the attack surface is has not increased significantly.

RichardHaselgrove:
What's the intended target audience for the recipients of this service? I think there are some fundamental networking questions which maybe beyond some audience segments.

Users in corporate environments - may be behind a centrally-managed firewall.
Home users - are likely to be using a NAT router. I have seven computers and sundry other devices (TV, printer, NAS) sharing one external IP address.
All users - should have antivirus protection which nowadays will include some sort of firewall.

jacobbogers:
Hi folks, just discovered this PR, very very nice one, although i have familiarity with WebRTC, i have no such thing for BOINC, line wire protocol,  I will need to read into this part of BOINC if i want to contribute

TheAspens:
> > ```
> > Any website in the world would be able to connect to the core client. How do we prevent malicious websites from doing harm to the users machine or pretending to be a legitimate BOINC website?
> > ```
> 
> The only sensitive operation is connecting to a project as all other operations require the authenticator. By default only connecting to official projects in the white list will be allowed. Additional projects can be manually added to the client configuration to support development activities.

It is my opinion that a user must explicitly consent within the BOINC client installed on their machine before an attach can complete.  If we don't require this, then if (for example, and he wouldn't actually) @brevilo goes rogue and modifies E@H he could cause the website to automatically attach anyone who visits the website.  Or, someone could hack the BOINC website and add a new malicious project to the list.  They could then hack a site popular with volunteers and start attaching them to his malicious project.  Doing any form of automatic attach based on requests from a website opens up the opportunity for some very bad behavior.

> 
> > ```
> > If we allow this interface to perform actions such as attach or cancel a task then how do we ensure that only actions approved by user for that particular boinc project are performed?
> > ```
> 
> As far as I understand all operations require the authenticator. This is stored in a cookie after doing the Web login.

The use of the authenticator in the cookie as the authorization token for the website is a really insecure practice and is on my list of things to fix (I do not know when I can get to it though).  We should not be relying on this mechanism to obtain the authenticator.  Additionally, in the long run, it is my opinion that the authenticator should be reduced so that it is only used to indicate that a specific computer is owned by a specific user account and should have no other meaning or use in the system.  

Having said that, I believe that all the client RPC commands require the use of the value in the gui_rpc_auth.cfg file not the authenticator.  Since the value of the gui_rpc_auth.cfg is not available to the browser, it would probably be necessary to implement a API on the website to get a nonce that is used to authorize browser-client communications.  When the browser calls the client with the nonce, the client will have to call the server to validate that the nonce is associated with the user.  There might be another way of doing this, but that was my first thought.

> 
> > ```
> > How do we install a SSL certificate that would be trusted by browers so that these communications would be allowed? Most SSL certificates are only valid for 2-3 years while users often leave BOINC installed for longer than that. What happens when the certificates expire?
> > ```
> 
> This can be a self-signed certificate generated during the client installation. A clean method is required to add this to the trusted certificates in the browser configuration.

I think that this is dangerous because it weakens the security of the browsers.  For example, if the private key for the self-signed certificate was obtained by an attacker it would make the browsers on that computer vulnerable to man in the middle attacks.  Since we have no way to revoke the self-signed certificate, once this happened there would be nothing we could do to fix the situation.  Here is an example of where someone tried an approach similar (although much worse - I realize that we would generate the self-signed root certificate unique to each machine and then once the https certificate is generated I presume we would delete the private key of the root certificate) then this: https://www.digicert.com/blog/lenovos-superfish-adware-perils-self-signed-certificates/

Further - browsers like Firefox deploy with their own store of certificates and those get replaced on each update (I don't know what the other browsers do).  Once the browser was updated, it wouldn't have the self-signed root certificate anymore and thus the mechanism would stop working.

lfield:
> `It is my opinion that a user must explicitly consent within the BOINC client installed on their machine before an attach can complete.  If we don't require this, then if (for example, and he wouldn't actually) @brevilo goes rogue and modifies E@H so that the website automatically attaches anyone who visits the website.  Or, someone could hack the BOINC website and add a new malicious project to the approved project list.  They could then hack a site popular with volunteers and start attaching them to his malicious project.  Doing any form of automatic attach based on requests from a website opens up the opportunity for some very bad behavior.`
This should be fairly straight forward to do and is very similar to the auto-attach mechanism already in place. 

```

>  The use of the authenticator in the cookie as the authorization token for the website is a really insecure practice and is on my list of things to fix (I do not know when I can get to it though).  We should not be relying on this mechanism to obtain the authenticator.  Additionally, in the long run, it is my opinion that the authenticator should be reduced so that it is only used to indicate that a specific computer is owned by a specific user account and should have no other meaning or use in the system.  
> 
>  Having said that, I believe that all the client RPC commands require the use of the value in the gui_rpc_auth.cfg file not the authenticator.  Since the value of the gui_rpc_auth.cfg is not available to the browser, it would probably be necessary to implement a API on the website to get a nonce that is used to authorize browser-client communications.  When the browser calls the client with the nonce, the client will have to call the server to validate that the nonce is associated with the user.  There might be another way of doing this, but that was my first thought.

```
Now I am confused. I looked at the [protocol specification](https://boinc.berkeley.edu/trac/wiki/GuiRpcProtocol#Authentication) to refresh my memory but it didn't help.



```

> I think that this is dangerous because it weakens the security of the browsers.  For example, if the private key for the self-signed certificate was obtained by an attacker it would make the browsers on that computer vulnerable to man in the middle attacks.  Since we have no way to revoke the self-signed certificate, once this happened there would be nothing we could do to fix the situation.  Here is an example of where someone tried an approach similar (although much worse - I realize that we would generate the self-signed root certificate unique to each machine and then once the https certificate is generated I presume we would delete the private key of the root certificate) then this: https://www.digicert.com/blog/lenovos-superfish-adware-perils-self-signed-certificates/
> 
> Further - browsers like Firefox deploy with their own store of certificates and those get replaced on each update (I don't know what the other browsers do).  Once the browser was updated, it wouldn't have the self-signed root certificate anymore and thus the mechanism would stop working.

```
There is a [good summary ](https://letsencrypt.org/docs/certificates-for-localhost/) which states that http://127.0.0.1 can be considered secure so we may not need a certificate.

TheAspens:
> > It is my opinion that a user must explicitly consent within the BOINC client installed on their machine before an attach can complete.  If we don't require this, then if (for example, and he wouldn't actually) @brevilo goes rogue and modifies E@H so that the website automatically attaches anyone who visits the website.  Or, someone could hack the BOINC website and add a new malicious project to the approved project list.  They could then hack a site popular with volunteers and start attaching them to his malicious project.  Doing any form of automatic attach based on requests from a website opens up the opportunity for some very bad behavior.
> 
> This should be fairly straight forward to do and is very similar to the auto-attach mechanism already in place.

Great.  As long as the BOINC client requires the user acknowledge the attach request before performing it, then I think it will be good.

> 
> > The use of the authenticator in the cookie as the authorization token for the website is a really insecure practice and is on my list of things to fix (I do not know when I can get to it though).  We should not be relying on this mechanism to obtain the authenticator.  Additionally, in the long run, it is my opinion that the authenticator should be reduced so that it is only used to indicate that a specific computer is owned by a specific user account and should have no other meaning or use in the system.
> > Having said that, I believe that all the client RPC commands require the use of the value in the gui_rpc_auth.cfg file not the authenticator.  Since the value of the gui_rpc_auth.cfg is not available to the browser, it would probably be necessary to implement a API on the website to get a nonce that is used to authorize browser-client communications.  When the browser calls the client with the nonce, the client will have to call the server to validate that the nonce is associated with the user.  There might be another way of doing this, but that was my first thought.
> 
> Now I am confused. I looked at the [protocol specification](https://boinc.berkeley.edu/trac/wiki/GuiRpcProtocol#Authentication) to refresh my memory but it didn't help.

I looked at the code which lead me to this output from the boinccmd tool.  The password referred to in the link you provided is directly associated with the contents of the gui_rpc_auth.cfg file.
```
[knreed@oc7811586721 ~]$ boinccmd -h

usage: boinccmd [--host hostname] [--passwd passwd] [--unix_domain] command

default hostname: localhost
default password: contents of gui_rpc_auth.cfg
...

```

> 
> > I think that this is dangerous because it weakens the security of the browsers.  For example, if the private key for the self-signed certificate was obtained by an attacker it would make the browsers on that computer vulnerable to man in the middle attacks.  Since we have no way to revoke the self-signed certificate, once this happened there would be nothing we could do to fix the situation.  Here is an example of where someone tried an approach similar (although much worse - I realize that we would generate the self-signed root certificate unique to each machine and then once the https certificate is generated I presume we would delete the private key of the root certificate) then this: https://www.digicert.com/blog/lenovos-superfish-adware-perils-self-signed-certificates/
> > Further - browsers like Firefox deploy with their own store of certificates and those get replaced on each update (I don't know what the other browsers do).  Once the browser was updated, it wouldn't have the self-signed root certificate anymore and thus the mechanism would stop working.
> 
> There is a [good summary ](https://letsencrypt.org/docs/certificates-for-localhost/) which states that http://127.0.0.1 can be considered secure so we may not need a certificate.

That would be good.  I hope that ws://127.0.0.1 and http://127.0.0.1 are considered secure domains.  The discussion around that on links off the article you posted are extensive and somewhat confusing.  Please let us know what your testing unveils.

lfield:
The feedback has been added to issue #2113. A new PR will be provided once the issues have been resolved.

PR #2149, opened 2017-09-30
Mac os10 3 screensaver fix

CharlieFenton:
BOINC screensavers have 2 parts. The BOINC "screensaver coordinator" is an actual "native" screensaver for MS Windows (boinc.scr) or Macintosh (BOINCSaver.saver). It communicates with the BOINC client via RPCs to determine which project applications (tasks) are currently running, and which of those have associated graphics executables. Depending on the user's screensaver preference settings, it cycles through the running tasks, launching one of the associated graphics executables at a time. The graphics executables usually communicate with their associated worker app to display information about the work being done by that task.

In addition to being run by the BOINC screensaver coordinator, users can run the graphics executables from the BOINC Manager via the "Show graphics" button when they select a graphics-capable running task.

When no active tasks with associated graphics are running, BOINC's screensaver coordinator launches BOINC's default graphics executable (boincscr). Depending on the user's screensaver preference settings, it may also intersperse displaying boincscr with project application graphics.

On MS Windows, and also on Macs prior to OS 10.13, the graphics executables display their windows over the screensaver coordinator's window, giving the appearance that they are part of the native screensaver. But Mac OS 10.13 has changes which prevent any application from covering the native screensaver's window.

I designed the changes in this PR with several main goals:

 * Minimize the effort required for projects to update their graphics executables.

 * Use the same logic and methodology as before for displaying updated graphics executables on MS Windows and on Macs running versions of OS X prior to OS 10.13 (backward compatibility.)

 * Use the same logic and methodology as before for displaying graphics executables via BOINC Manager's "Show graphics" button, whether or not the graphics executable has been updated, and on any version of OS X we currently support (OS 10.6 - OS 10.13) (backward compatibility).

 * On Mac OS 10.13, give the user an explanation of why graphics executables that have not yet been updated don't appear in the screensaver, and move on as quickly and efficiently as possible to ones that can be displayed.

The fix for this problem on OS 10.13 involves establishing communication between the screensaver coordinator and the graphics executable via Mach-O inter-process communication. It also takes advantage of a feature available in OS X called IOSurfaceBuffer. IOSurfaceBuffer allows OpenGL applications to share images with another process, while continuing to take advantage of the rendering acceleration provided by the GPU. We can use this approach because BOINC graphics executables built with the standard BOINC graphics libraries use OpenGL.

To update their graphics executables, projects simply need to follow these 3 steps:
[1] Rebuild the BOINC graphics library for the Mac (libboinc_graphics2.a) from the sources in this PR
[...]

PR #2294, opened 2018-01-10
Mac screensaver: under OS 10.13+ if the BOINC screensaver cannot disp…

CharlieFenton:
…lay a graphics app with hardware acceleration using the IOSurface APIs because the app has not been linked with the current graphics libraries, then display it using the much slower CGWindowListCreateImage API.

### DIFF
--- a/clientscr/Mac_Saver_ModuleView.h
+++ b/clientscr/Mac_Saver_ModuleView.h
@@ -42,6 +42,8 @@
 - (IBAction)closeSheetSave:(id) sender;
 - (IBAction)closeSheetCancel:(id) sender;
 
+- (bool) setUpToUseCGWindowList;
+
 @end
 
 @interface SharedGraphicsController : NSObject <NSMachPortDelegate>
--- a/clientscr/Mac_Saver_ModuleView.m
+++ b/clientscr/Mac_Saver_ModuleView.m
@@ -94,7 +94,10 @@ bool isErased;
 
 static SharedGraphicsController *mySharedGraphicsController;
 static bool runningSharedGraphics;
+static bool useCGWindowList;
 static pid_t childPid;
+static int gfxAppWindowNum;
+static NSView *imageView;
 static char gfxAppPath[MAXPATHLEN];
 static int taskSlot;
 static NSRunningApplication *childApp;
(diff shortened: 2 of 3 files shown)


CharlieFenton:
I have created this PR because I believe that the project "screensavers" are an important marketing tool. When people see the graphics on a friend's computer, they can become curious and ask about what they see, helping spread word of mouth about BOINC And I believe that few BOINC projects have yet updated their graphics apps to work with the BOINC screensaver under Mac OS 10.13.

In the changes associated with PR #2149 (Mac OS 10.13 screensaver fix) I added code to both the BOINC graphics library and to the BOINC Mac screensaver code to allow the screensaver to display project graphics apps under Mac OS 10.13 High Sierra and later. On MS Windows, and also on Macs prior to OS 10.13, the graphics executables display their windows over the screensaver coordinator's window, giving the appearance that they are part of the native screensaver. But Mac OS 10.13 has changes which prevent any application from covering the native screensaver's window. To work around this restriction, it is necessary for the screensaver process to be able to access the contents of the graphics app process's window and display those contents in the screensaver's window.

The Mac has two API sets which can accomplish this: IOSurface and CGWindowListCreateImage. In the earlier PR, I chose the  IOSurface APIs because they are very efficient. They use shared memory to keep only one copy of the graphics content, and make full use of GPU hardware acceleration. But they only work with OpenGL / GLUT rendering. My implementation required only that the graphics APIs be relinked with the updated BOINC graphics libraries. When the Mac screensaver encountered graphics apps that had not been updated, it displayed a message saying that the graphics are not compatible with this version of the OS and tried a graphics app associated with a different task. Since the BOINC Manager installer includes a default graphics app which has been updated, the screensaver could always fall back to that. (The above applies only when running on Mac OS 10.13+. On older versions of Mac OS, the screensaver continues to display graphics apps as before, whether or not they have been updated.)

Several months have now passed, and I believe that few BOINC projects have yet updated their graphics apps. Also, there may be some graphics apps that do not render exclusively using OpenGL / GLUT and so may not be compatible with the IOSurface APIs. This update to the Mac screensaver code falls back to the CGWindowListCreateImage method of displaying the graphics when the graphics app has not been linked with the IOSurface support in the updated libraries. However, the animation may appear less smooth, as this method is much slower, resulting in  a low frame rate. On my new quad-core iMac (a pretty fast machine) the CGWindowListCreateImage method can render at best 8 or 9 frames per second. Nonetheless, I feel that it is better to display the screensaver graphics with a slightly jerky animation than not at all.

These changes take effect only under OS 10.13+. Under older versions of the Mac OS or on other platforms (MS Windows, etc.), the screensaver displays the graphics exactly as before.

My hope is to first merge this into GIT master, then cherry pick the changes into the 7.8 branch, and finally to release a public version 7.8.5 for the Mac only.

davidpanderson:
Thanks!

PR #2295, opened 2018-01-10
Drupal: Add additional rules to notify moderators when forum related actions occur.

drshawnkwang:
Added rules to notify moderator mailing list when a comment is:
* deleted by an admin
* editted by someone other than the original author
* hidden/unhidden by someone other than the original author
* converted to a new forum topic.

Also notify when a forum topic is:
* editted by a moderator or admin
* hidden/unhidden by moderator or admin

https://dev.gridrepublic.org/browser/DBOINCP-298


tristanolive:
Changes look good, I just have not managed any testing. Maybe moderation is set up on albert and can be used to confirm this?

PR #2298, opened 2018-01-12
[WIP] Move the gfx_html example to a separate repository

TheAspens:
Move the gfx_html example to a separate repository at https://github.com/BOINC/boinc-samples.  This is due to the gfx_html sample including the mongoose library which is under the GPL and we want clear seperation of GPL vs LGPL code.  Note that this was always a standalone sample so there was never a violation of the LGPL, but this makes the relationship clearer.

For example, this caused confusion in pull request https://github.com/BOINC/boinc/pull/2246


TheAspens:
Yea - just saw that. Thanks for commenting.  Changing this to WIP.

AenBleidd:
@TheAspens, if you do not mind I want to be a reviewer of this PR too.

davidpanderson:
See my email.  Instead of making a new confusingly-named repo, remove mongoose.* and add a link in samples/gfx_html/README

TheAspens:
@davidpanderson - last email I got from you on this subject was from Nov 16th where you wrote 

"gfx_html should be moved to its own repo (say, BOINC/gfx_html).

No one is using this (as far as Rom or I know)
and I don't think anyone legally can because of the GPL code.

-- D"

In my opinion - since we don't think anyone is using it and we don't think anyone can use.  I'd rather just delete it.  

Let me know what your opinion on this issue is.

TheAspens:
@AenBleidd  - you are also welcome to be a reviewer.

davidpanderson:
Even though no one's using it, someone might and it's a nice piece of work by Rom.  If mongoose.cpp is being used unmodified, I'd say just remove it and provide a pointer to it.

UweBeckert:
I'm using it in yafu (https://yafu.myfirewall.org/yafu) to show more information about the workunit.

TheAspens:
@UweBeckert - let me look at @davidpanderson's suggestion for removing the mongoose library (which is GPL) from the BOINC source tree (to provide clear separation from BOINC which is LGPL) but make sure the mongoose library can easily be put in place to build the gfx_html for those who want it.

I'm going to close this pull request and delete my branch and I'll open a new one with the revised approach.  I'll tag everyone in this thread when I do.

Thanks!

PR #2300, opened 2018-01-13
Mac ram size hotfix

CharlieFenton:
Fixes a bug which caused BOINC client built with SDK OS 10.11 or later to fail to get correct system RAM size when running on older versions of OS X. This should also be cherry picked to master for BOINC version 7.9.

Also replicates commit 5296088 previously made in master into 7.8 branch: correctly replace screensaver when installing an older version of BOINC to replace a newer version.


CharlieFenton:
Upon approval, I will build BOINC 7.8.6 for the Mac with these changes.

Issue #2268, opened 2017-12-06
Bugs: boinccmd --get_old_tasks


RichardHaselgrove:
1) online help at https://github.com/BOINC/boinc/blob/master/client/boinc_cmd.cpp#L80 shows "show reported tasks from last 24 hours": this was changed to 1 hour in 9024f8cc18fe06b4c9d8bededda80355e4aaceb6

2) final line of output from https://github.com/BOINC/boinc/blob/d15d0b95de6db61fc9eac3445ec280be823134bf/lib/gui_rpc_client_print.cpp#L413 is labelled "reported time": I would take this to be the time of reporting to the project server. Instead, it's the create_time of the OLD_RESULT record, which is identical to the "completed time" of the task.

From report by user naptimelabs at https://boinc.berkeley.edu/dev/forum_thread.php?id=12158

PR #2302, opened 2018-01-14
Correct boinccmd helptxt

RichardHaselgrove:
This pull request (trivially) updates the help text in boinccmd to match current code. This corrects the first part of issue #2268 - it should go into v7.10, please.

The second part of #2268 needs more work. The variable create_date is used to time-out stale entries from the OLD_RESULT vector, and can't be recycled as a reporting date. It needs a separate value report_date in the OLD_RESULT struct, initialised to 'uninitialised', and updated as part of 'got ack for task' in handle_scheduler_reply() (search 'or' vector for matching task name: update if found). Then, the final line in OLD_RESULT::print should be conditionally printed only if initialised/updated. If anybody feels able to code this in time, please be my guest.


RichardHaselgrove:
'Merge pull request' was simply me updating my working fork - not sure why it's listed. Shouldn't affect anything.

PR #2307, opened 2018-01-17

ChristianBeer:
- Deleted old translations from 2016 or older
- Imported translations from transifex that were at least 75% finished (including new languages)
- compiled all mo files

Note: BOINC-Web.po is still outdated because the template is not current. I checked for issues with printf in Manager and Client translations as described in #2165 and fixed them.


JuhaSointusalo:
Fun facts:

- 27 files were added, 74 removed and 295 changed for a total count of 396 files.
- 193 568 lines were added and 134 334 removed for a total of 327 902 lines. I'm not going to count the number of completely new lines or lines that were removed completely but with the number of added and removed files I'd say both counts are in tens of thousands. A significant number of changed lines were when the location of a string was changed in source code file but the string or the translation were not changed.
- In Git patch format, with context lines, diff directives, binary diffs and what nots, this PR is 477 967 lines long and weights 16 879 514 bytes.
- Going through the patch from top to bottom requires about 10 600 presses of Page Down or, pressing and holding down Page Down, about 6 minutes.
- In patch format view browser tab takes about 170 MB of RAM.
- In fancy formatted view the browser tab crashed several times after running out of address space. I'd guessimate that with all the diffs expanded, and with 64-bit browser, the fancy formatted view would take 10-20 GB of RAM.

RichardHaselgrove:
Wow! Thanks, Juha. I didn't realise what I was asking you to do - award yourself a gold star or something. Perhaps a beer.

PR #2309, opened 2018-01-18
db_purge: fix bugs introduced in zlib changes

davidpanderson:


SETIguy:
I'm inclined to merge despite the changes in coding style between the two versions because the bug is fairly obvious.   If there are objections we can revert.

PR #2311, opened 2018-01-20
fix client simulator build

davidpanderson:


AenBleidd:
@davidpanderson, Windows build is broken because of missed 'config.h' file.
Is it really necessary for Windows build? I guess it is linux only

ChristianBeer:
Isn't `config.h` a file that should be generated at build time on Linux to reflect the current system that is used for the build? Or are we talking about the empty `win_build/config.h` file? Do we need to populate the file on Windows like `configure`does on Linux?

AenBleidd:
@davidpanderson, there is still an issue with missed config.h file. Please, take a look at https://ci.appveyor.com/project/BOINC/boinc/build/1.0.154-vbodlnxf

davidpanderson:
Can someone please check the appveyor failure?  The solution builds for me in all configurations (32/64 bit, Debug/Release).

AenBleidd:
@davidpanderson, I can make a fix in your branch in an hour

AenBleidd:
I created a fix in #2319 
@davidpanderson, could you please review it and merge into *this* branch (dpa_sim) ?
Also I want to mention that this project (boincsim) is not built by default when building the solution.
I'm not sure what configuration should include this project to be built so could you please fix solution files (both for VS2010 and VS2013)?
Thanks

davidpanderson:
Eventually win_build/win-config.h should be renamed to win_build/config.h.
That way .cpp files can just include config.h without #ifdefs

AenBleidd:
@davidpanderson, please review and merge pull request #2329 into this branch. Thanks

TheAspens:
Are there any open issues left in this pull request?  If not, @AenBleidd - do you feel ok being the reviewer and merging this?  I haven't reviewed it yet, but if the two of you are happy with the work I think you can merge it.  Let me know if you want someone else to review.

AenBleidd:
@TheAspens, I'm OK fix this pull request. I can merge this if @davidpanderson have nothing to add.

TheAspens:
@davidpanderson - if you could let @AenBleidd know if you are happy with this change as is, then he can merge it.  Thanks!

JuhaSointusalo:
Rest of BOINC code uses the following pattern:

```
#if defined(_WIN32) && !defined(__STDWX_H__)
#include "boinc_win.h"
#elif defined(_WIN32) && defined(__STDWX_H__)
#include "stdwx.h"
#else
#include "config.h" // if necessary
// other includes
#endif
```

Unconditionally including `config.h` doesn't really fit the pattern. `boinc_win.h` defines VC++ / MinGW specific `HAVE_*`s.

`boinc_win.h` is also used as precompiled header so I think it should be the first header to be included for it to actually work as PCH.

davidpanderson:
I have nothing to add.  The goal here is to reduce #ifdefs in .cpp files.

TheAspens:
@davidpanderson and @JuhaSointusalo - Are the comments that @JuhaSointusalo made above resolved or is more discussion required?

JuhaSointusalo:
After writing my comment above I started thinking that there is a lot more room for improvements.
- `boinc_win.h` has some stuff targeting MinGW and Cygwin even though those two can run configure.
- `boinc_win.h` doubles as `config.h` and pre-compiled header for VC++.
- Autotools builds don't use pre-compiled headers at all.
- There is duplication between `boinc_win.h` and `stdwx.h`.

I think what could be done is to review the entire situation and figure out the best way to handle configuration and pre-compiled headers and all and then make the changes everywhere. But that's a big task and doesn't really fit in this PR. So I don't object if this PR is merged as it is now.

TheAspens:
Done - but we need to get 7.10 out and stop using cherry picks.  As branch 7.10 gets further and further away from master it becomes increasingly difficult to deal with these issues.

PR #2312, opened 2018-01-20
Add boinccmd commands to do acct mgr RPC, and to show acct mgr info

davidpanderson:
New commands:
--acct_mgr info: show name and URL of current AM
--acct_mgr synch: do RPC

also (for uniformity) new names for existing commands:
--acct_mgr attach URL email password
--acct_mgr detach

Tempted to remove old names, but didn't.

### DIFF
--- a/client/boinc_cmd.cpp
+++ b/client/boinc_cmd.cpp
@@ -60,7 +60,7 @@ default password: contents of gui_rpc_auth.cfg\n\
 Commands:\n\
  --acct_mgr attach URL name passwd  attach to account manager\n\
  --acct_mgr info                    show current account manager info\n\
- --acct_mgr synch                   synchronize with acct mgr\n\
+ --acct_mgr sync                    synchronize with acct mgr\n\
  --acct_mgr detach                  detach from acct mgr\n\
  --client_version                   show client version\n\
  --create_account URL email passwd name\n\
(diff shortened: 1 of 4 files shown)


ChristianBeer:
Sidenote: removing existing commands without prior warning to the community breaks backwards compatibility. There is no harm in leaving the old names for now. Ideally there should be a warning when used that they are going to be removed in the next version of BOINC (meaning 7.12).

TheAspens:
Tested legacy commands and new version of commands and everything appears to work correct.  I agree with Christian that they should be kept for now and if we want to remove in the future then we provide a notification when they are used that they are deprecated for one or two releases before we remove.

One question for confirmation: If your client was attached to a project using an account manager and then the account manager is detached.  Should the client remain attached to the projects that the account manager attached them to?  (this is the behavior I observed, but wanted to confirm that was intended).

One tiny item that I would recommend changing.  I think that it is much more common to use "sync" rather than "synch".  I think that "--acct_mgr synch" should be changed to "--acct_mgr sync".

Let me know your response to the two items above and then I can merge.  Thanks!

davidpanderson:
I changed synch to sync.

That behavior is correct: detaching from an account manager doesn't detach projects.

TheAspens:
Thanks looks good!  Merging.

Issue #1961, opened 2017-07-06
BSD build/run issues


JamesMolson:
With freebsd,

In "client/mac_address.cpp":
I seem to have SIOCGIFCONF defined but neither SIOCGIFHWADDR or SIOCGIFARP which leads to hw_addr never being initialised, which causes the call to ether_ntoa(hw_addr) to segfault.
There is no #else to catch as a fall through.

Building needs to link to -lexecinfo for backtrace calls.

In "sched/db_dump.cpp"
WEXITSTATUS needs sys/wait.h

It also fails to build with clang - for some reason a number of system headers never get included.

On netbsd,
lib/diagnostics.cpp and lib/diagnostics.h

"boinc_catch_signal(int signal, struct siginfo *siginfo, void *sigcontext)"

on netbsd siginfo seems to be defined as a union, and not a struct.

ChristianBeer:
Which version of freebsd did you test with? Did you get it to build by yourself with some changes? Can you provide a diff or a pull request?

JamesMolson:
11.1

Yes, there is a fall through that returns -1 so I just forced that, but really it should have the correct implementation.

I can make a PR (eventually), but I have a long list of other things to do first.

AenBleidd:
Is this fixed via #2313 and could be closed?

JamesMolson:
> Is this fixed via #2313 and could be closed?

That only fixed some freebsd issue from what I can see, and not the others, so no.

I've been a little too busy here to do anything on boinc directly myself :/

JamesMolson:
I have a bit of time to look into fixing things again. One issue we found was that our AVX FreeBSD executables were not being sent out.

https://github.com/BOINC/boinc/blob/master/client/hostinfo_unix.cpp#L715-L762

It seems the code here checks for just mmx, 3dnow, sse1-3, and discards everything else.

With respect to the earlier solaris comment, it does not build for me at all either; again the MAC address extraction needs looking at.

I will start to make some PRs.

AenBleidd:
@JamesMolson, is this still an issue? I don't remember you pushed cy PRs (I may be wrong).
Could you please take a look when you have time?
Thanks in advance.

AenBleidd:
If this is still an issue - please reopen it.
Closing this as 'wontfix'

PR #2313, opened 2018-01-22
client: fix BSD build (from Larry Rosenman)

davidpanderson:


lrosenman:
This also addresses some of the issues in #1961

TheAspens:
Although I cannot test the change, the code isolates the changes to BSD and it passed the build tests, so I'm going to proceed with merging this request.

TheAspens:
Thanks @lrosenman  for contributing!

lrosenman:
For the record with these patches in, my user that was segfaulting reports correct operation.  Thanks for merging.

PR #2317, opened 2018-01-24
client: remove spurious "invalid account filename" error messages

davidpanderson:
These errors happened because of the new "account_info.xml" file,
whose name resembles that of an account file.


TheAspens:
Commit 45868c1c30806ac6eacf480e8220dda199ee1fbd changed the name of the file to login_token.txt and therefore this issue was resolved previously. 

The bug report was based on a 7.9 build of the client that was created before the filename was changed by that commit.

I recommend that this pull request be closed without merging and the source branch deleted.

PR #2318, opened 2018-01-25
Web: fix factorization of project-specific code

davidpanderson:


TheAspens:
@davidpanderson - Does Seti@Home use this?  I'll bug Eric to review it if he does.

davidpanderson:
Yes.  This change is for the benefit of SETI@home, and is deployed there now.

PR #2320, opened 2018-01-25
Mac update dependencies

CharlieFenton:
In preparation for building and testing BOINC 7.9.0 for the Mac, update the following third party dependencies:
SQlite 3.11.0 -> SQlite 3.22.0
cURL 7.50.2 -> cURL 7.58.0
c-ares 1.11.0 -> c-ares 1.13.0
OpenSSL 1.1.0 -> OpenSSL 1.1.0g
FreeType 2.6.2 -> FreeType 2.9
I have previously updated wxWidgets 3.0.0 -> wxWidgets 3.1.0

Also:
* Build only 64-bit libraries for wxWidgets and SQLite, instead of 32-bit / 64-bit "universal" libraries. This will reduce the time to build them and reduce the storage requirements for Travis CI.
* Eliminate patching curlrules.h which was needed to build 32-bit BOINC Manager.

PR #2323, opened 2018-01-26
Mac: Automate updating cURL build script

CharlieFenton:
Mac: Eliminate needing to update the script which build cURL for each new version of c-ares or OpenSSL; get the names of current c-ares and OpenSSL directories from dependencyNames.sh.

This will reduce the likelihood of errors caused by forgetting to manually update this script.

### DIFF
--- a/mac_build/buildFTGL.sh
+++ b/mac_build/buildFTGL.sh
@@ -66,6 +66,9 @@ while [[ $# -gt 0 ]]; do
     shift # past argument or value
 done
 
+# needed for ftgl 2.1.3-rc5 to find our freetype 2.9 build not the system one
+export PKG_CONFIG_PATH=${libftpath}/lib/pkgconfig:${PKG_CONFIG_PATH}
+
 SRCDIR=$PWD
 
 if [ "${doclean}" != "yes" ]; then
--- a/mac_build/buildcurl.sh
+++ b/mac_build/buildcurl.sh
@@ -33,6 +33,7 @@
 # Updated 9/10/16 for curl 7.50.2 with c-ares 1.11.0
 # Updated 3/14/17 to patch curlrules.h to fix BOINC Manager compile error
 # Updated 1/25/18 for curl 7.58.0 with c-ares 1.13.0 & openssl 1.1.0g, don't patch currules.h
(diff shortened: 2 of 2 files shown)


ChristianBeer:
To make sure this works as expected I deleted the buildcache on Travis and triggered a rebuild of this PR. If this is successful I'm going to merge this.

ChristianBeer:
FTGL failed to build because it couldn't find freetype2. Is this a new dependency? Maybe we need to install via `brew install freetype`.

ChristianBeer:
Investigation update: the freetype build log was a red herring. The travis build suppresses the stdout output which is why all the `make` and `make install` output is missing. It turns out the `freetype-config` script generated during build gives precedence to a system `pkg-config` installation instead of using our defined prefix. I'll try to patch that away.

ChristianBeer:
My fix the the FTGL buildscript fixed it. This must now be merged into master so the build cache for Travis Mac can be rebuilt and the failed PRs can be restarted. I'm not sure if I'm still eligible to merge since I added a code change to this PR so I'm throwing it to some other @BOINC/committer to merge. The changes are all sound so far.

CharlieFenton:
I will merge this change into my mac_fix_travis_script branch PR #2324 in order to allow the Travis CI build to succeed for that branch, and

This change will also need to be merged into any other pending branches where the Travis CI build is failing on the Mac build.

CharlieFenton:
Cherry-picking commit ee77f9e from this branch into any other pending branches might be easier than merging. That is what i did to fix my mac_fix_travis_script branch PR #2324.

ChristianBeer:
I realize that I shouldn't have used this PR to test my changes my commit ee7f9e should have been a separate PR on it's own as it addresses a problem introduced to master in another PR. We can still salvage the situation by resetting the branch of #2324 to it's earlier state (without the cherry-pick), then merge this PR into master, then restart the failed builds on the other PRs after the merge build is done and we have a working mac cache on travis again. No need for cherry-picking at all.

JuhaSointusalo:
@ChristianBeer Since you were wondering in ee77f9e why this worked before. #2320 updated FreeType2 from 2.6.2 to 2.9. FreeType2 2.8 was the first one to try to use `pkg_config` in `freetype_config`.

PR #2327, opened 2018-01-27
boinccmd: show CPU/GPU resources in --get_task

davidpanderson:
Also only show relevant info.
E.g. don't show exit status of a job that hasn't finished.

### DIFF
--- a/lib/gui_rpc_client_print.cpp
+++ b/lib/gui_rpc_client_print.cpp
@@ -160,26 +160,37 @@ void RESULT::print() {
     foo = (time_t)report_deadline;
     printf("   report deadline: %s", ctime(&foo));
     printf("   ready to report: %s\n", ready_to_report?"yes":"no");
-    printf("   got server ack: %s\n", got_server_ack?"yes":"no");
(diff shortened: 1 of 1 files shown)


lib/gui_rpc_client_print.cpp
```diff
@@ -160,26 +160,37 @@ void RESULT::print() {
     foo = (time_t)report_deadline;
     printf("   report deadline: %s", ctime(&foo));
     printf("   ready to report: %s\n", ready_to_report?"yes":"no");
-    printf("   got server ack: %s\n", got_server_ack?"yes":"no");
```
TheAspens: Was the line printf("   got server ack: %s\n", got_server_ack?"yes":"no"); intentionally dropped?

TheAspens:
I've tested this and it looks fine.  Once I get an answer to the comment I made above I will be able to merge.

davidpanderson:
It was intentionally dropped.  Once a result is acked by the server,
it's garbage-collected within a few seconds.

TheAspens:
Ok thanks for the clarification.  I'm approving and merging.

Issue #1989, opened 2017-07-22
Android custom build app doesn't install


adamradocz:
Hi,

I build the Android Boinc without any modification and when I try to install it to my phone it doesn't install and I get the "App not installed." error message. I uninstalled the previous version of Boinc but it didn't help.

Unknown sources is allowed.
I have a Xiaomi Redmi Note 2 and Linage OS Android 6.0.1.
I also triad the same package on a LeEco Le 2 Linage OS Android 7.1 and got the same result.

![screenshot_20170722-114538](https://user-images.githubusercontent.com/16503773/28490214-a696fae6-6ed4-11e7-9bb6-f43543757b2f.png)

brevilo:
While we don't support custom builds it would be interesting to know which codebase (commit SHA1) you used for it.

adamradocz:
I downloaded the master codebase that day, when I posted the issue. I tried the vagrant environment and windows environment as well. I've deleted my development environment so I can't check the logcat records.

brevilo:
Ok, thanks. My latest builds of `master` (using the Vagrant env) are a few months old (~April) but at that time they built and worked OK on AOSP 5.1.1, AOSP 6.0.1 and OxygenOS 7.1.1.

Could you please try 981b0409d3? If that doesn't work either then it's likely to be problem at your end. If it does work you could `git-bisect` to find the commit that broke the build in `master`.

adamradocz:
The 981b040 gives the following error:
`Error:(31, 0) Invalid release tag encountered: vboxwrapper/26198. Expression: (1 == match.size())
<a href="openFile:/home/vagrant/StudioProjects/boinc/android/BOINC/app/build.gradle">Open File</a>`

The 4700280 gets the same App not installed error message.

brevilo:
Ok, the first one can be explained but the second one indicates there's something wrong on your build system or you might be picking the wrong apk (architecture). Run a logcat while installing the app. It should tell you why.

adamradocz:
The logcat doesn't show any useful unfortunately:

`--------- beginning of main
gralloc_unregister_buffer:1440: ID=373781 ref=0
gralloc_register_buffer:1325: hnd=0x7f8f489300 ID=373783 fd=37 ref=1
gralloc_unregister_buffer:1440: ID=373783 ref=0
gralloc_unregister_buffer:1440: ID=373777 ref=0
gralloc_unregister_buffer:1440: ID=373779 ref=0
[...]

PR #2330, opened 2018-01-29
Android: Make build scripts more reliable

alexander255:
Currently the Android app build scripts will not detect errors, causing spurious follow-up issues (such as configure failing with an *libcurl not found* error after libcurl failed to build or be correctly found, or `strip` dump a sh*load of spurious errors because the `cd` to the binary directory failed and it tries to parse the top-level source and text files instead, etc.).
Also it's very hard for somebody calling `./build_all.sh` from a script to figure out whether the build *actually* succeeded or not (probing the app binary directory works, but is wokey)…

### DIFF
--- a/android/build_androidtc_arm.sh
+++ b/android/build_androidtc_arm.sh
@@ -1,4 +1,5 @@
-#/bin/sh
+#!/bin/sh
+set -e
 
 #
 # See: http://boinc.berkeley.edu/trac/wiki/AndroidBuildClient#
@@ -10,6 +11,6 @@ export NDK_ROOT="${NDK_ROOT:-$HOME/NVPACK/android-ndk-r10e}"
 export ANDROID_TC="${ANDROID_TC:-$HOME/android-tc}"
 export ANDROID_TC_ARM="${ANDROID_TC_ARM:-$ANDROID_TC/arm}"
 
-if [ ! -d $ANDROID_TC_ARM/arm-linux-androideabi ]; then
-    $NDK_ROOT/build/tools/make-standalone-toolchain.sh --platform=android-9 --arch=arm --install-dir=$ANDROID_TC_ARM
+if [ ! -d "$ANDROID_TC_ARM/arm-linux-androideabi" ]; then
+    "$NDK_ROOT/build/tools/make-standalone-toolchain.sh" --platform=android-9 --arch=arm --install-dir="$ANDROID_TC_ARM" "$@"
 fi
--- a/android/build_boinc_arm.sh
+++ b/android/build_boinc_arm.sh
@@ -19,7 +19,7 @@ export TCINCLUDES="$ANDROIDTC/arm-linux-androideabi"
 export TCSYSROOT="$ANDROIDTC/sysroot"
 export STDCPPTC="$TCINCLUDES/lib/libstdc++.a"
 
-export PATH="$PATH:$TCBINARIES:$TCINCLUDES/bin"
+export PATH="$TCBINARIES:$TCINCLUDES/bin:$PATH"
 export CC=arm-linux-androideabi-gcc
 export CXX=arm-linux-androideabi-g++
 export LD=arm-linux-androideabi-ld
--- a/android/build_boinc_arm.sh
+++ b/android/build_boinc_arm.sh
@@ -1,4 +1,5 @@
-#/bin/sh
+#!/bin/sh
+set -e
 
 #
 # See: http://boinc.berkeley.edu/trac/wiki/AndroidBuildClient#
@@ -27,20 +28,20 @@ export CFLAGS="--sysroot=$TCSYSROOT -DANDROID -DDECLARE_TIMEZONE -Wall -I$TCINCL
 export CXXFLAGS="--sysroot=$TCSYSROOT -DANDROID -Wall -I$TCINCLUDES/include -funroll-loops -fexceptions -O3 -fomit-frame-pointer -fPIE -march=armv7-a"
 export LDFLAGS="-L$TCSYSROOT/usr/lib -L$TCINCLUDES/lib -llog -fPIE -pie -march=armv7-a -Wl,--fix-cortex-a8"
 export GDB_CFLAGS="--sysroot=$TCSYSROOT -Wall -g -I$TCINCLUDES/include"
-export PKG_CONFIG_SYSROOT_DIR=$TCSYSROOT
+export PKG_CONFIG_SYSROOT_DIR="$TCSYSROOT"
 
 # Prepare android toolchain and environment
 ./build_androidtc_arm.sh
 
 if [ -n "$COMPILEBOINC" ]; then
 echo "==================building BOINC from $BOINC=========================="
-cd $BOINC
+cd "$BOINC"
 if [ -n "$MAKECLEAN" ]; then
 make distclean
 fi
 if [ -n "$CONFIGURE" ]; then
 ./_autosetup
-./configure --host=arm-linux --with-boinc-platform="arm-android-linux-gnu" --with-ssl=$TCINCLUDES --disable-server --disable-manager --disable-shared --enable-static
+./configure --host=arm-linux --with-boinc-platform="arm-android-linux-gnu" --with-ssl="$TCINCLUDES" --disable-server --disable-manager --disable-shared --enable-static
 sed -e "s%^CLIENTLIBS *= *.*$%CLIENTLIBS = -lm $STDCPPTC%g" client/Makefile > client/Makefile.out
 mv client/Makefile.out client/Makefile
 fi
--- a/android/build_build_app_x86.sh
+++ b/android/build_build_app_x86.sh
@@ -1,4 +1,5 @@
-#/bin/sh
+#!/bin/sh
+set -e
 
 #
 # See: http://boinc.berkeley.edu/trac/wiki/AndroidBuildApp
@@ -24,8 +25,8 @@ export GDB_CFLAGS="--sysroot=$TCSYSROOT -Wall -g -I$TCINCLUDES/include"
 
 make clean
 
-if [ -e "./confgiure" ]; then
-./configure --host=i686-linux --prefix=$TCINCLUDES --libdir="$TCINCLUDES/lib" --disable-shared --enable-static
+if [ -e "./configure" ]; then
+./configure --host=i686-linux --prefix="$TCINCLUDES" --libdir="$TCINCLUDES/lib" --disable-shared --enable-static
 fi
 
 make
--- a/android/build_build_app_x86.sh
+++ b/android/build_build_app_x86.sh
@@ -13,7 +13,7 @@ export TCINCLUDES="$ANDROIDTC/i686-linux-android"
 export TCSYSROOT="$ANDROIDTC/sysroot"
 export STDCPPTC="$TCINCLUDES/lib/libstdc++.a"
 
-export PATH="$PATH:$TCBINARIES:$TCINCLUDES/bin"
+export PATH="$TCBINARIES:$TCINCLUDES/bin:$PATH"
 export CC=i686-linux-android-gcc
 export CXX=i686-linux-android-g++
 export LD=i686-linux-android-ld
--- a/android/build_libraries_arm.sh
+++ b/android/build_libraries_arm.sh
@@ -20,7 +20,7 @@ export TCINCLUDES="$ANDROIDTC/arm-linux-androideabi"
 export TCSYSROOT="$ANDROIDTC/sysroot"
 export STDCPPTC="$TCINCLUDES/lib/libstdc++.a"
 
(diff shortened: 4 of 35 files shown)


adamradocz:
This PR maybe solve the #1989 issue?

alexander255:
Unlikely, the BOINC Android app will compile and install just fine when the built binaries are broken/missing: Android platform simply treats the binaries as “static files”/assets and basically ignores them on install; only resources and Java bytecode (“classes.dex”) get special treatment iirc. The app *will* **hang** on startup in this case through, while filling `logcat` with “could not start client”-style messages.

No, if I had to guess what *his* issue is, I'd say he probably simply did a release build of the BOINC client app and forgot to (ie: didn't know that he has to) sign it. Only debug builds are signed by default using the standard debug keys, release builds are unsigned by default and need to be signed (using any key) before installation.

(For installation errors one usually receives much more useful error messages when attempting to install the app using `adb install <.apk>` instead of using the PackageManager GUI.)

TheAspens:
@alexander255 - I'm working on tracking down someone to review this code.  Both people I've asked are out this week on vacation but said they would try to look at it next week.  Sorry for the delay in getting it reviewed.

AenBleidd:
I think about adding Android build to CI. Will try to do this during this
week.

Best regards,
Vitalii Koshura

2018-02-05 16:33 GMT+02:00 Kevin Reed <notifications@github.com>:

TheAspens:
@AenBleidd - that would be great.  Thanks!

android/build_androidtc_arm.sh
```diff
@@ -10,6 +10,6 @@ export NDK_ROOT="${NDK_ROOT:-$HOME/NVPACK/android-ndk-r10e}"
 export ANDROID_TC="${ANDROID_TC:-$HOME/android-tc}"
 export ANDROID_TC_ARM="${ANDROID_TC_ARM:-$ANDROID_TC/arm}"
 
-if [ ! -d $ANDROID_TC_ARM/arm-linux-androideabi ]; then
-    $NDK_ROOT/build/tools/make-standalone-toolchain.sh --platform=android-9 --arch=arm --install-dir=$ANDROID_TC_ARM
+if [ ! -d "$ANDROID_TC_ARM/arm-linux-androideabi" ]; then
```
brevilo: One could argue that this is redundant but, ok, it's future-proof should line 11 change or be removed.

android/build_libraries_arm.sh
```diff
@@ -28,21 +28,21 @@ export CFLAGS="--sysroot=$TCSYSROOT -DANDROID -DDECLARE_TIMEZONE -Wall -I$TCINCL
...
 echo "==================building Libraries from $BOINC=========================="
-cd $BOINC
+cd "$BOINC"
 if [ -n "$MAKECLEAN" ]; then
 make clean
 fi
 if [ -n "$CONFIGURE" ]; then
 ./_autosetup
-./configure --host=arm-linux --with-boinc-platform="arm-android-linux-gnu" --prefix=$TCINCLUDES --libdir="$TCINCLUDES/lib" --with-ssl=$TCINCLUDES --disable-server --disable-manager --disable-client --disable-shared --enable-static --enable-boinczip
+./configure --host=arm-linux --with-boinc-platform="arm-android-linux-gnu" --prefix="$TCINCLUDES" --libdir="$TCINCLUDES/lib" --with-ssl=$TCINCLUDES --disable-server --disable-manager --disable-client --disable-shared --enable-static --enable-boinczip
```
brevilo: You missed `--with-ssl=$TCINCLUDES`

android/build_boinc_arm.sh
```diff
@@ -27,20 +27,20 @@ export CFLAGS="--sysroot=$TCSYSROOT -DANDROID -DDECLARE_TIMEZONE -Wall -I$TCINCL
 export CXXFLAGS="--sysroot=$TCSYSROOT -DANDROID -Wall -I$TCINCLUDES/include -funroll-loops -fexceptions -O3 -fomit-frame-pointer -fPIE -march=armv7-a"
```
brevilo: What's the reasoning behind not improving the contents of `CXXFLAGS`, `LDFLAGS` and `GDB_CFLAGS` in the same way (in all scripts)? I'm not sure the outer quotation marks are effective for the contained inner variables like in `-I$TCINCLUDES/include` as well. Are they?

android/build_libraries_x86.sh
```diff
@@ -14,7 +14,7 @@ MAKECLEAN="yes"
 export BOINC=".." #BOINC source code
 
 export ANDROID_TC="${ANDROID_TC:-$HOME/android-tc}"
-export ANDROIDTC="${ANDROID_TC_X86:-$ANDROID_TC/x86"
+export ANDROIDTC="${ANDROID_TC_X86:-$ANDROID_TC/x86}"
 export TCBINARIES="$ANDROIDTC/bin"
```
brevilo: Why not treat the following variables like `ANDROIDTC` above? Yes, that's already fixed but you keep depending on that. Ideally you want to protect any variable, without further dependencies.

android/build_libraries_arm.sh
alexander255: “Made with love by hand by …” – Guess I missed that one. Fixed and checked them all again.

android/build_boinc_arm.sh
alexander255: Interesting question actually… I so routinely ignore this issue I didn't have an imediate answer for this.

Turns out the answer is twofold:

  * As long as you use `autotools` there is absolutely nothing you can do here – any attempt at adding quotation or similar will make things worse! And [`cmake` is in no way better](https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/CMakeCInformation.cmake#L103) if I understand this code correctly.
  * I also checked `meson` and [it uses Python's `shlex.split`](https://github.com/mesonbuild/meson/blob/ae0e469b0692862dae10ceb02799feb3f117ce26/mesonbuild/environment.py#L924), which means it would support quoting the values correctly.

TL;DR: Yes this sucks and there is no way to fix it currently.

android/build_libraries_x86.sh
alexander255: Sorry, I don't understand what you are saying. Protect what variable, with which dependencies?

alexander255:
Hopefully all fixable issues (including `#!/bin/sh -e` vs `#!/bin/sh\nset -e`) should be fixed now.

truboxl:
Is it possible to correct the shebangs in all the scripts once and for all (missing !) ?
Probably a very small matter...

brevilo:
Yet an important one that was missed up to now! Nice catch!

alexander255:
@truboxl: Could you point out an example for a script that doesn't have a shebang?

brevilo:
As he wrote, they all lack the "!".

alexander255:
@brevilo: I looked at the scripts yesterday and I literally didn't *see* that. – `sed` fixed it just now.

brevilo:
Yeah, me too, obviously 😎

ChristianBeer:
I commented in #2409 about this but I want to raise the question here also: Why do you want to use `set -e` versus `if [ $? -ne 0 ]; then exit 1; fi` as the latter is done in all the CI scripts for Mac and Linux already. Why not be consistent and also use it for android? At least in scripts that use `if` construct the `set -e` should be removed.

AenBleidd:
@ChristianBeer, @alexander255, 
When I tested these scripts with `sh -e` on travis ci it always show no errors even when script failed.
I didn't check this on local environment but `if [ $? -ne 0 ]; then exit 1; fi` works well

alexander255:
@AenBleidd: Attempting to build BOINC using `./build_all.sh` without setting `"${OPENSSL_SRC}` properly copies all the ToolChain stuff and then “cleanly” fails with `can't cd to ~/src/openssl-1.0.2g` on the first error, setting the exit code (`$?`) to `2` on my system (`/bin/sh` is `dash`).

… and the reason is of course because I forgot to add the changed files before commiting and pushing as @truboxl correctly pointed out just now. :weary:

AenBleidd:
@alexander255, maybe we can cooperate here somehow?
I did the similar but more extended work on these scripts so maybe we can collect all fixes and make 1 clean pull request with all your changes as one squashed commit and all my changes on top of your changes as one squashed commit (2 commits in total)?
I think this can speed-up development and approval process.
Also this can save my time and your time.
What do you think?

alexander255:
Sure, I don't care. I just submitted these as I had encountered some issues when building from source and wanted to upstream it. Basically you're asking me to squash my commits so that you can reuse them as part of your PR, is that correct? If so I'd prefer to keep the first commit seperate as it addresses a specific problem that is described in detail in the commit message, while the others just fix “generic code problems”. Is that OK for you?

AenBleidd:
I'm ok with any possible ways. I just want to help you not to make the work I did already.
In any case I'll squash my commits on top of your code and resolve all conflicts.
As from my POV, it is better to review both your and my changes in scope as far as they contain changes to the same scripts.
If other guys are ok with this - I'd do it this way.
If no - I'm ok to keep them as a separate PRs.

alexander255:
@AenBleidd: Done! Please go ahead! :slightly_smiling_face:

brevilo:
@alexander255 there's still one review question pending regarding `android/build_libraries_x86.sh`.

android/build_build_app_x86.sh
```diff
@@ -25,7 +26,7 @@ export GDB_CFLAGS="--sysroot=$TCSYSROOT -Wall -g -I$TCINCLUDES/include"
 make clean
 
 if [ -e "./confgiure" ]; then
```
JuhaSointusalo: Could you fix this typo too?

JuhaSointusalo:
`set -e` has some pitfalls. See [this Server Fault answer](https://serverfault.com/a/847016) for examples as well as the linked pages.

I'm not saying you can't use `set -e` and these scripts are simple and probably don't have any issues but I think anyone working on these scripts need to be aware of the pitfalls.

alexander255:
@JuhaSointusalo: Fixed!

AenBleidd:
@brevilo, @JuhaSointusalo, are you ok with the latest changes?

JuhaSointusalo:
@AenBleidd The changes look ok to me. I don't have Android build environment set up to test this so I'll defer to @brevilo.

AenBleidd:
@JuhaSointusalo, you can take a look at my WIP PR 2437 that based on the latest changes from this PR but with my CI scripts and some additional changes. I plan to squash that PR when this one will be merged into master

alexander255:
I hate to be begging like this, but could somebody representative of the project give me a response the my second question on #2393. My [inclusion request](https://gitlab.com/fdroid/fdroiddata/merge_requests/2861) into the F-Droid repo has been open for over 2 months at this point and is currently dead-locked on that question.

brevilo:
@AenBleidd I don't even know what you mean by "latest" since none of the updates to this PR were done via incremental commits, unfortunately. As to me previous review, my [last comment](https://github.com/BOINC/boinc/pull/2330#issuecomment-376079377) still holds.

brevilo:
@alexander255 What *exactly* do you mean by "my second question on #2393". The second question or the second bullet point - presuming we're talking about the opening comment, as your second comment also contains a question. Thanks.

AenBleidd:
> @AenBleidd I don't even know what you mean by "latest" since none of the updates to this PR were done via incremental commits, unfortunately. As to me previous review, my last comment still holds.

@brevilo, I've just created new PR with new code and test on it. I also don't like this way of making commits but I used to work with different cases so it's not a problem for me to do all job one more time

alexander255:
@brevilo: Mostly second bullet, but I already got a detailed answer to both questions by CristianBeer. It's not really related to this issue either way.

AenBleidd:
@brevilo, any updates?

brevilo:
Sorry, GDPR all over I guess. I'm closing my review as the remaining questions aren't that important and the discussion just lost coherence for me...

AenBleidd:
@truboxl, are you ok with the changes now?

android/build_libraries_x86.sh
alexander255: While I always add curly braces (`{}`) around all variables in shell code *I* write, doing so is only required when doing any kind of variable substitution (ie: `${ANDROID_TC_X86:-Default}`). In practice you'll find lots and lots of shell code that only uses the braces when required and this won't cause any problems if you always consider wether the character *after* the variable name *may be part of a shell variable name* or not when writing it.

I tend to ofter *not* consider the above criterion in practice so I add the braces just to be safe, but this really only matters when you write the code; once it's properly written adding the braces everywhere is really not worth the (wasted) time, while also likely to cause hard to spot errors if the code is not reviewed very carefully (wasting everybody's time).

TheAspens:
Merging based on approval of reviewers and contributors

PR #2336, opened 2018-02-02
Credit calculation: remove old code that ignored result.app_version_id =1.  Fixes #2333

### LINKED ISSUE #2333: Information about app version with id = 1 is not updated
In validator log: missing app_version_id (1)
The reason in a strange check:
https://github.com/BOINC/boinc/blob/5c36ce25daae8532f44d81e9f4c5527bde5e0e51/sched/credit.cpp#L527-L540
This is a very old code, and perhaps this check is not relevant. As a result, information about app version with id = 1 is not updated.

davidpanderson:
... and reverted to old way of estimating job FLOPs.
Not sure why this was there.  1 is a perfectly valid ID.

### DIFF
--- a/sched/credit.cpp
+++ b/sched/credit.cpp
@@ -524,6 +524,22 @@ int get_pfc(
         }
     }
 
+    // is result from old scheduler that didn't set r.app_version_id?
+    // if so, use WU estimate (this is a transient condition
+    // while projects upgrade server software)
+    //
+    if (r.app_version_id == 0) {
+        if (config.debug_credit) {
+            log_messages.printf(MSG_NORMAL,
+                "[credit] [RESULT#%lu] missing app_version_id (%ld): returning WU default %.2f\n",
+                r.id, r.app_version_id, wu_estimated_credit(wu, app)
+            );
+        }
+        mode = PFC_MODE_WU_EST;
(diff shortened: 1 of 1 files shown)


sched/credit.cpp
```diff
@@ -524,21 +524,6 @@ int get_pfc(
...
-            log_messages.printf(MSG_NORMAL,
-                "[credit] [RESULT#%lu] missing app_version_id (%ld): returning WU default %.2f\n",
-                r.id, r.app_version_id, wu_estimated_credit(wu, app)
-            );
-        }
-        mode = PFC_MODE_WU_EST;
-        pfc = wu_estimated_pfc(wu, app);
-        return 0;
-    }
-
```
TheAspens: Do we still need to handle the case where r.app_version_id == 0?  Not all projects have updated to the new code yet so having this fall back would continue to support their transition gracefully.

JuhaSointusalo:
@davidpanderson 

Neither the commit message or pull request message references issue #2333.

davidpanderson:
Good point.  I restored the old logic for when the ID is zero (but not 1).

TheAspens:
Looks good.  Thanks.

PR #2340, opened 2018-02-06
Remove HOST_USAGE.max_ncpus.  It's not used anywhere.

davidpanderson:
avg_ncpus is always used instead.

Note: there are two other things named "max_ncpus".
One is a computing prefs field (limit on # of CPUs to use).
The other, a field of SCHED_CONFIG,
is the max # of CPUs the scheduler will believe a client has
in deciding how many jobs to send it.

### DIFF
--- a/client/client_types.cpp
+++ b/client/client_types.cpp
@@ -768,7 +768,6 @@ void APP_VERSION::init() {
     safe_strcpy(plan_class, "");
     safe_strcpy(api_version, "");
     avg_ncpus = 1;
-    max_ncpus = 1;
     gpu_usage.rsc_type = 0;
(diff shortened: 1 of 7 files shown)


RichardHaselgrove:
Please wait 10 minutes while I read it.

TheAspens:
Since this will be merged into master, then it should be on the v7.10 change list.  However, because it is removal of code that doesn't do anything, it should only be the comprehensive detailed list - not on a list of high level changes.

RichardHaselgrove:
Sure. Just a reminder to ourselves that we'll have to update (deprecate the tag) in http://boinc.berkeley.edu/wiki/Anonymous_platform when we're done

PR #2341, opened 2018-02-07

ChristianBeer:
Current translations with >98% completeness for BOINC-Manager.po checked for incorrect usage of substitution variables (%s, %1)


Ageless93:
Just asking, what variables are needed? 
I see lots of 1%s, a couple 1%d in BOINC-Manager.pot, are those correct?

RichardHaselgrove:
%s for string replacement, %d for digit replacement, seems properly placed to me. There are none that have a leading character before the %. (I found some &%s, but the accelerator key marker appeared correctly in the current manager)

The other files have %1, %2, %3 ... for replacing the first, second, third ... parameters. BOINC-Manager.pot and the associated .po files have repeated %s for multiple strings - e.g. at https://github.com/BOINC/boinc/blob/master/locale/de/BOINC-Manager.po#L3811.  Is that a change in policy, or a change in translation system requirements?

Ageless93:
What I mean is what I see in Transifex. For instance, the English source has "Identify your account at 1%s" (https://www.transifex.com/boinc/boinc/translate/#nl/manager/40473174). I take it that that's wrong then?

RichardHaselgrove:
Dunno - that one's above my pay grade ;-). I can only see the raw data files here in the repository - Transifex requires login, and I don't have an account.

https://github.com/BOINC/boinc/blob/master/locale/nl/BOINC-Manager.po#L26 has:
#: clientgui/AccountInfoPage.cpp:352
#, c-format
msgid "Identify your account at %s"
msgstr "Je account gegevens bij %s"

ChristianBeer:
Transifex adds the numbers in front of the variables and a rounded darker background to indicate the special meaning of the variable. This should help the translator to place the variables in the right places. In transifex you can click the highlighted variable and it gets copied to the translation. I warned about this at the time this was introduced as it caused confusion among translators.

BOINC uses the %1, %2, ... style on webpages (PHP code) were the translator can actually change the position of the variables according to the language. The other format %s, %d is used in programs (C code) where it is not possible to switch the positions should two replacement be in one string and it is important that the order of the variables is preserved in the translation.

PR #2343, opened 2018-02-09
Tagging for client version 7.9.1 in release branch

RichardHaselgrove:
Still showing as pre-release version


CharlieFenton:
Should we be designating this as a release candidate (still requiring testing, of course), make it version 7.10.0, create a _client_release7/7.10_ branch for it and tag it as _client_release7.10/7.10.0_? If so, we should also change version.h to have:
`#define BOINC_PRERELEASE 0`
Or do people feel it is too soon to designate it a release candidate and bump the version to 7.10.0?

CharlieFenton:
After leaving my previous comment, I see that someone has already created an _origin/client_release7/7.10_ branch, and that the proposal is to merge this PR into that. But I still wonder if we shouldn't designate this as a release candidate (still requiring testing, of course), make it version 7.10.0, tag it as _client_release7.10/7.10.0_ and change version.h to have:
`#define BOINC_PRERELEASE 0`

RichardHaselgrove:
"_someone has already created an origin/client_release7/7.10 branch_" - that was me.

I'm testing out the provisional Client Release Process at: https://github.com/BOINC/boinc-policy/blob/client_release/Client_Release_Process.md. Under 'Build and Test' (page 3), it is proposed that the initial testing cycle keeps the 'odd' testing minor version number, and that we switch to the 'even' number once we have a release candidate. The process paper is silent as to whether the test build should keep the prerelease flag: perhaps that's the first amendment required!

TheAspens:
I think that odd numbered versions should have pre-release flag enabled and even numbered versions should have it disabled.  If you want, go ahead and create a branch and open a WIP pull request against the client release process document to capture changes as we go.

I can see the arguments for either 7.9.1 or 7.10.1 so I support whichever decision Richard makes .

RichardHaselgrove:
OK, I effectively made the decision this morning by creating the branch version pull request as 'odd: prerelease'. Let's run with that for simplicity and see how it goes.

TheAspens:
@CharlieFenton - once you have reviewed Richard's comments and if you agree please go ahead and merge this  (unless someone else objects before then).

CharlieFenton:
I have merged. Please delete the update_release_version_7.9.1 branch, and tag it. Note that tagging requires an extra step in GIT beyond updating the version numbers.

RichardHaselgrove:
Both done. Please check that I've done it right - that wasn't intuitive.

CharlieFenton:
It looks correct to me. Tagging is actually pretty easy if you use a GUI like TortoiseGit or SourceTree. With TortoiseGit (which I believe is only available for MS Windows), right click on your local repository and select _Create Tag_. In the dialog which appears, you can then select the _Commit_ radio button and click on the ellipsis button (3 dots) to the right of that radio button. That displays a commit history so you can easily select the commit where you wish to place the tag.

In Sourcetree's _History_ view,  right-click on the commit where you wish to place the tag and select _Tag..._ from the contextual menu which appears. SourceTree is written in Java so it is cross-platform.

I almost always use Sourcetree on my Mac and TortoiseGit on MS Windows for GIT operations. I find them far easier than GIT's command-line interface, and both are free. You can find information on setting up TortoiseGit [here](http://boinc.berkeley.edu/trac/wiki/SourceCodeGit/Windows) and [here](http://boinc.berkeley.edu/trac/wiki/SourceCodeGit/WindowsKeygen). There is also useful information, especially for other platforms, [here](http://boinc.berkeley.edu/trac/wiki/SourceCodeGit).

RichardHaselgrove:
I've got Tortoise, but it's a matter of being guided to pick the right tool for the job. 'Make branch' is simply a matter of typing a name into a browser on GitHub: switch to 'tag' in that dialog, and the 'create' box becomes a 'search' (only) text box. GitHub help covers branching, and led me to the command-line reference for tags. That's easy from tortoise ("git bash here"), but the hardest part was switching branches from the command prompt. Of course, GitHub Help doesn't cover tools like Tortoise, so I didn't even look...

CharlieFenton:
Are we ready to begin testing 7.9.1? Should I build and upload BOINC 7.9.1 for Macintosh to <http://boinc.berkeley.edu/dl/>? Who will build and upload BOINC 7.9.1 for MS Windows and / or Linux?

RichardHaselgrove:
If you would care to refer back to the email I sent to all four build managers around midday UTC last Friday,  subject line "BOINC Client release branch 7.10 - ready for test builds as v7.9.1", you should have your answer.

Yes, please build. The code is ready. For Windows, the build has to be made by a member of UCB staff, for licencing (Installshield) and code-signing reasons. That means David (@davidpanderson). For Linux, I'm expecting builds by Laurence Field (@lfield  - Fedora/Ubuntu) and Gianfranco Costamagna (@LocutusOfBorg - Debian).

Ageless93:
For Windows builds with VirtualBox installer included, the present version 7.8.5 has 5.1.26 included. Which VBox version do you want to have included in 7.9.1? Or is that something that the Windows Build Manager has to decide upon?

RichardHaselgrove:
It will be the build manager, who has already proposed 5.2.6 and received no objection.

PR #2345, opened 2018-02-09
[WIP] [Android] Add initial CI build configuration

AenBleidd:


ChristianBeer:
Some observations (also applicable to #2347):
- Travis-CI already supports an android build image but it doesn't support multi language builds so we need to always install and build dependencies on our own if the android app is in the same repository as the C++ code. I couldn't find a way around this so I stopped looking. An interesting aspect that I didn't get around to testing is to setup an extra repository with a new `.travis.yml` file (with `language: android`) that pulls in the current master branch of the main BOINC repo thus testing the android build. One has to find out how to trigger this from a Pull request and feed information back to show the build status. Travis used to do something like this with their own repositories.
- When developing the Mac and Linux CI scripts I had a development branch with a WIP pull request to test my changes. After I was finished I made a new branch and manually "squashed" my commits into more sensible commits to avoid cluttering the commit history with "fix typo; try to fix this; try to fix that;" commits while still giving reasons for the changes. I suggest you do the same for the android CI.
- In the end you should put all custom build dependencies into `3rdParty/buildCache/android` so they are automatically cached by Travis. Please only put the build artifacts in there not the whole source and intermediate artifacts to keep the cache small.

AenBleidd:
@ChristianBeer, thank you for your advice.

- I'd like to start with all-in-one file. If we then find the solution to move the Android build to another repo - it will be easier to do with existing approach. So I want to finish this first. Also installing all necessary dependencies is not so big deal as from my POV.
- I also use this just like a temporary branch to play with scripts and make them stable. I plan to have then 2 clean pull requests: one with CI scripts, another one with fixes to existing Android scripts. Maybe there will be something else that should be included in the separate pull request.
- Yep, I plan to move custom build dependencies to the cache. At first I want to have a clean build and then move it to the cache.

In any case I want to say one more time: both this and #2347 are temporary pull requests that will be closed when the work will be done. And them I'll open create new clean pull requests with correct name and necessary changes only.

AenBleidd:
New cleaner pull request is  created #2409 
Closed as unnecessary.

PR #2354, opened 2018-02-16
update project list XML for clients

davidpanderson:


CharlieFenton:
If my understanding is correct, then the addition of Science United is meant to be temporary for testing auto-attach. If that is correct, then this should be merged into the _client_release/7/7.10_ branch, not master. If it is to be a permanent change, then it will need to be copied / cherry-picked into _client_release/7/7.10_ branch after merging into master.

davidpanderson:
it's permanent.  I updated the copy in the 7.10 branch

RichardHaselgrove:
I'll take a look when I've finished my morning reading and coffee.

RichardHaselgrove:
Argh. The copy in 7.10 uses the 'master' url for Einstein, not the 'Join' url. I'll correct that later, and bump the version.

No builds for the time being, please.

RichardHaselgrove:
The copy in the v7.9.1 Windows installer is also wrong.

This pull request intended for master leaves the current value for the Einstein 'Join' url, which is correct.

JuhaSointusalo:
I'm a bit confused now. I commented about Onboard because I thought it was still under testing. Now that I went to Science United website I see that it really is still in testing.

It is my understanding that projects are not added to the list while they are in testing phase. What kind of rules we have for adding account managers? Are the rules the same as for projects or something different?

@davidpanderson Does Science United have some bugs you know about? Or do you believe it is ready for general public but you just don't want to call it ready because it hasn't been tested by that all many people yet? Or something else?

Some Science United issues (off-topic for this PR but whatever):
Please have http point to right site or redirect to https.
Community -> Teams says teams are disabled and Profiles and UOTD say profiles are disabled. Maybe not have those in the menu at all?
Maybe you should check server config because these links work: https://scienceunited.org/wiki and https://scienceunited.org/dev . There could others too.

(I somehow managed to not notice the review request on the previous PR. Sorry about that.)

davidpanderson:
Testing SU requires making it available to testers.
In the current stage these would be alpha testers.

We're not "inviting" users of 7.8.3 to use US.
SU appears in the account manager list.
That's fine with me; SU is functional.
After testing I'll invite users to use SU.

TheAspens:
It is my opinion that Science United should be treated like any other BOINC project or BOINC account manager.  If the policy is that projects and account managers should not be added to https://boinc.berkeley.edu/project_list.php or included in the client all projects list until they have already been tested and have reached production status, then that policy should also apply to Science United as well.  I'm not sure that there has ever been a clear policy stated, but @davidpanderson - what guidelines have you used previously in determining whether or not to add a project or account manager to the all projects list?

Secondly, it is my opinion that we need to be clear about what we are testing.  Testing the 7.9/10 client does not depend on the Science United and the 7.9/10 client could be tested and released even if Science United did not exist.  As a result, I think that we should keep the process of testing each of these separate.

davidpanderson:
Auto-attach from an account manager is a feature of 7.10 and needs to be
tested.
Currently SU is the only account manager that supports auto-attach.
So testers need to create SU accounts.
They don't have to continue using SU after testing.

If people want to report other bugs in SU (via its message boards or email)
that's fine but it's not part of the 7.10 testing.

RichardHaselgrove:
Is David's comment about v7.10 testing perhaps more appropriate for PR #2355?

TheAspens:
@JuhaSointusalo @CharlieFenton @RichardHaselgrove - I've discussed with David and I'm willing to merge this.  However, you have each expressed reservations.  Can you please confirm if you object to the merge or if you are ok with it proceeding?

RichardHaselgrove:
I am NOT raising a formal objection, but I still have doubts that I think should be explored further downstream. It is not clear to me which should the authoritative source for this file - the driver for http://boinc.berkeley.edu/projects.php on isaac, or the GitHub repository. Perhaps a stronger protocol that they should be updated in sync? - isaac seems to have taken precedence recently.

CharlieFenton:
I defer to @davidpanderson and withdraw my objection.

TheAspens:
@JuhaSointusalo - can you please let me know if you are ok with proceeding with a merge of this pull request?  Thanks!

JuhaSointusalo:
I think there is two different topics here: testing 7.10 and including SU in project list.

So we need a test project that is in all_projects_list.xml to test auto-attach. Ok, no problem with that. But before auto-attach client downloads the list from Berkeley. That means the test project needs to be in official project list. Because the official list is shared between stable and test clients stable clients sees the test project too.

I think there should be two project lists, one for stable clients and one for test clients. The test project list would be the same as stable project list expect it can include test projects such as BOINC Alpha and whatever test account manager. This way it wouldn't be necessary to use stable projects to test new features like Seti Main is used now and SU once it is stable.

An easy way to do that would be adding a dev parameter to the project list PHP code the same way it's used elsewhere. The client already knows if it's pre-release version or not so adding a parameter to the project list URL is pretty trivial and the PHP code should be easy enough to change to support that.

There's a few gotchas with that. Upgrading from test client to stable version leaves all_projects_list.xml intact with the test projects. But the focus should be upgrading from one stable version to another so I don't think that's much of a problem. Another one is if when testing auto-attach Berkeley is down and downloading the list with test projects isn't possible. Not big issue either. It also wouldn't be possible to auto-attach to test project with stable client which could be a minor annoyance sometimes.


As for including SU in the official list. I see the official list as some sort of a recommendation. "These do good science, check them out!" Something similar for account managers. And I do agree with Richard's use of word invitation ("Check them out!").

Since SU is not yet ready I see inviting people to it problematic. People see that and give it a quick look. "So this is the future of BOINC? Huge disappointment! I'll go blog about it and write forum posts and trash it completely." I don't think SU needs that kind of negative PR. One day SU is going to be ready and you have ads about it and people google it and see the negative opinions and forget about it.

Writing a news post and formally inviting people to beta test SU is different. Then you can tell it's about testing the account manager functionality and you can list all the features that were planned but are not yet ready and so on. So you can make sure people have low expectations and then they can be positively surprised.

In short. I think adding SU to the official list at this stage would be bad for SU. But I'm not going to stomp my feet against it. If David doesn't mind people forming negative opinions about SU so be it.

JuhaSointusalo:
> It is not clear to me which should the authoritative source for this file - the driver for http://boinc.berkeley.edu/projects.php on isaac, or the GitHub repository.

The authoritative source is some file I can't remember which in boinc-site. A copy in this repo is needed only so that it can be included in installers.

boinc-site would have been better place to discuss adding SU but I don't watch that repo, I don't know what's going on there and I don't make comments there.

davidpanderson:
Huh??  SU is not being billed as "the future of BOINC".
It's an alternative account manager being shown to Alpha testers.
This the only way to test auto-attach from account managers.

SU works fine for me.
If you find bugs in it, report them.

RichardHaselgrove:
Hmmm. Let's pause for some rigo(u)r here.

This pull request is to bring our 'master' repo up-to-date (last change Nov 16 2017). Since then, three new projects have been added to the BOINC reprtoire (SourceFinder, Dec 7, and two from ODLK, Dec 24). This Pull Request has nothing to do with alpha testing.

Our test branch, currently at tag 7.9.3 and awaiting Windows build, has Science United included in a version of all_projects_list.xml dated Feb 17. As Release Manager, I'm happy to leave that where it is to facilitate David's testing. But when the time comes to switch that branch to Release Candidate mode, I'm minded to take advice and reconsider whether SU should remain in the installation file - that will depend, _inter alia_, on whether SU reaches full public release status before our branch does.

There are two active versions of all_projects_list under repo control. One is in 'master' in this repo. and the other is in 'client_release/7/7.10', also in this repo. But there a third copy, not subject to version control. It can't be found in the boinc-site repo as others have suggested, but it can be viewed at https://boinc.berkeley.edu/all_projects_list.xml. It it very similar to, but not identical with, the 7.9.3 repo version, with SU included.

And it's being actively distributed by 14-day (silent) client requests, not just to alpha testers, but also to users of the 'recommended stable' v7.8.3 version, and for all I know earlier recommended versions too. That's why I asked which should be the authoritative file (a) for testers, (b) for production users.

JuhaSointusalo:
The authoritative list is in https://github.com/BOINC/boinc-site/blob/master/projects.inc . Well, actually, the file in the repo is outdated so right now the authoritative file is projects.inc on the web server.

Client downloads the list through http://boinc.berkeley.edu/project_list.php .

RichardHaselgrove:
Well, up to a point. As I read it, projects.inc contains the framework for which projects to display: there's a lot more detail (platforms, hardware) which is assembled dynamically by project_list.php

I tried to follow this through with the ODLK projects, which were announced on Dec 24. According to the Wayback Machine, they were not visible on Dec 21, but visible on Dec 30: yet we have no commits between those dates. There's a finger-fumble recovery for projects.inc on Jan 9 2018, which confuses the issue.

TheAspens:
> I think there should be two project lists, one for stable clients and one for test clients. The test project list would be the same as stable project list expect it can include test projects such as BOINC Alpha and whatever test account manager. This way it wouldn't be necessary to use stable projects to test new features like Seti Main is used now and SU once it is stable.
> 
> An easy way to do that would be adding a dev parameter to the project list PHP code the same way it's used elsewhere. The client already knows if it's pre-release version or not so adding a parameter to the project list URL is pretty trivial and the PHP code should be easy enough to change to support that.

I like this idea.  I would however have a single list but that list would have an extra field for each project: 
`<status>production</status> `
or 
`<status>testing</status>`

It would then perhaps be useful to have an option during the add project or add account manager dialogue to "Show Production Projects Only." (better text needed here).   If the box was checked, then only production projects are show.  If the box is not checked, then both production and testing projects are shown.  The client would check the box by default when it is a GA build and it would not check the box when it is a development build.  There would need to be an indicator on the list about which are test vs production (perhaps by adding [test] after the project name in the list).

In all cases, the autoattach mechanism would work for either production or testing projects.

In my opinion this would preserve the seperation for end users of the difference between testing and operational projects and projects that are still prepping to be in production.  It would also allow for easy testing of features with the client and it would not hinder auto-attach from working correctly for projects in test status.

I would also suggest that the BOINC website be modified to list the projects in a Production vs Testing section as well.

TheAspens:
> SU works fine for me.
> If you find bugs in it, report them.

I have reported a couple to you, including the one that drove the need for pull request #2405.   This is why I would argue that Science United is still in a testing phase.

davidpanderson:
The problem with changing password or email requires #2405 in the client.
Any other problems?

CharlieFenton:
> I like this idea. I would however have a single list but that list would have an extra field for each project:

Since all clients download the projects list periodically, older clients would not be aware of the extra field and so probably would show the non-production projects in the _Add Project_ dialog without differentiating them from production projects. Of course, that would be no different from what happens now. 

But if we are going to implement something like this, it would be nice to do it in a way that is fully backward compatible with older clients. Ideally, it would be implemented in such a way that older clients show only production projects.

Ageless93:
One simple solution is to leave the projects list as is for production versions of BOINC, while for Beta/test versions, the list is included in the installer and the client is set not to check for a newer version of the list. 
That way you don't have to figure out how to do things with backward compatibility, with projects that show up for production versions that you do not want to show up, etc. 

For the present range that then means: 
7.9 has its own list included, with each installer, won't update unless a newer 7.9 is installed. 
7.10 uses the list from the server, updates it every 14 days from the website.

RichardHaselgrove:
While we're at it, we should probably rewrite the project list management system to be fully data (and hence database) driven. Take, for instance, the project DENIS@Home, which was re-announced as a participating project on 21 March. It's visible at https://boinc.berkeley.edu/projects.php, but it doesn't yet appear on https://boinc.berkeley.edu/all_projects_list.xml, and hence won't be in the 14-day updates - probably because boinc-site/projects.inc hasn't been updated yet. Having to make data-related changes in a code file sounds like bad practice and liable to error as personnel change.

ChristianBeer:
Having the project list as a git tracked file is better than having a DB for this. For one we want to document when a project was added or deleted and probably why. Putting this into a DB system complicates things especially in regard to traceability.

The current problem is with where the list is and how it is updated. The BOINC Governance rules state that additions to the official list need to be vetted by the PMC ([Governance](https://github.com/BOINC/boinc-policy/blob/master/Governance.md) section 2.5: "Decide on the set of “approved” projects and account managers") yet this has not happened once. Another problem is that the maintainer of the official BOINC website makes local changes that are not committed and pushed to the boinc-site github repository in time.

Additionally it seems that https://boinc.berkeley.edu/all_projects_list.xml is a locally generated file (not git tracked) that needs to be created manually each time a project is added or deleted from the list. A better way would be to automate this step so the file is always in sync with the list.

TheAspens:
@ChristianBeer is correct.  This falls under the PMC's purview.  I've started the process of getting them to review and approve/disapprove the changes contained in this pull request. 

The discussion above has diverged into questions about the best way to maintain and present the list.  I strongly encourage the creation of 1 or 2 issues based on the discussions above to be used as the bases of how to implement improvements to the process for the future.  However, making those changes is outside the scope of this pull request.

IMHO - The BOINC project page https://boinc.berkeley.edu/projects.php provides the following statement 
> The following projects are known to us at BOINC, and we believe that their descriptions are accurate.

I think that a low bar like this is the only reasonable way to maintain this list and we will need to make sure the text on the manager and on the website reflect this.  A project need not be bug free to be listed.  (again this is all IMHO and the determination needs to be made by the full PMC).

TheAspens:
This has been superseded by pull request https://github.com/BOINC/boinc/pull/2471.  As a result I am closing this pull request.

PR #2355, opened 2018-02-16
Make download.php (for auto-attach) work for both test and production

davidpanderson:
A new ?dev=1 arg tells it to show Development versions of the client

### DIFF
--- a/html/ops/get_versions.php
+++ b/html/ops/get_versions.php
@@ -5,6 +5,6 @@
 
 file_put_contents(
     "../user/versions.xml",
-    file_get_contents("https://boinc.berkeley.edu/download_all.php?xml=1")
+    file_get_contents("https://boinc.berkeley.edu/download_all.php?xml=1&dev=1")
 );
 ?>
--- a/html/user/download.php
+++ b/html/user/download.php
@@ -73,15 +73,16 @@ function client_info_to_platform($client_info) {
 
 // find release version for user's platform
 //
-function get_version() {
+function get_version($dev) {
     $v = simplexml_load_file("versions.xml");
     $client_info = $_SERVER['HTTP_USER_AGENT'];
     $p = client_info_to_platform($client_info);
+    $string = $dev?"Development":"Recommended";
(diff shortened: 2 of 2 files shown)


TheAspens:
I would recommend that you change the URL in ops/get_versions.php to use:

https://boinc.berkeley.edu/download_all.php?xml=1&dev=1

At this point, the get_versions.php script doesn't download the development versions, so when you use the script and then set ?dev=1 it causes it to redirect to the download_all.php on the BOINC site (since the development versions are not found in the versions.xml file).

davidpanderson:
done

RichardHaselgrove:
The current - 6 March - version of this commit (which is working in test mode at SETI@Home) produces a fatal error when the server is down for maintenance.

`Fatal error: Call to a member function update() on a non-object in /disks/carolyn/b/home/boincadm/projects/sah/html/inc/account.inc on line 28`

TheAspens:
Ok - this looks good.  Let me test one more time and I'll merge.  @RichardHaselgrove - I don't see how that error could be triggered by this code.  I'm going to setup my own test server and see what happens when I test it.  Do you have any other information about your error?

RichardHaselgrove:
Well, I've re-tested https://setiathome.berkeley.edu/download.php?dev=1 (and the same without the parameter) again, and the error has now gone. Maybe this was a project-specific kludge (another useful project page - https://setiathome.berkeley.edu/show_server_status.php - has become unavailable during maintenance in the last 3 hours). We'll have to ask David and Eric what project changes have been made tonight to accommodate the testing, and whether they need to be propagated into the canonical code.

RichardHaselgrove:
Checking around, the display of other SETI pages during 'in maintenance' times has changed significantly during today - but nothing has yet been committed to the project-specific SVN repository. Call this [WIP], unrelated to this pull request, and we'll take another look next week. I don't think this is related to the code change in this PR, so merge as you see fit.

TheAspens:
Tests work ok.  Merging.

PR #2374, opened 2018-02-22
Fix client crash on Mac OS 10.6 - 10.11 when built under OS 10.12 or later

CharlieFenton:
Mac: Patch ares_config.h and curl_config.h to not use clock_gettime() which is  defined in OS 10.11 SDK but was not available before OS 10.11. This caused the client built on OS 10.11 or OS 10.12 to crash on launch when run on older versions of OS X.

After this is merged into master, the changes needs to be cherry-picked into the client_release/7/7.10 branch before tagging for version 7.9.3.


CharlieFenton:
Correction: 
Mac: Patch ares_config.h and curl_config.h to not use clock_gettime() which is defined in OS 10.12 SDK but was not available before OS 10.12. This caused the client built on OS 10.12 or OS 10.13 to crash on launch when run on older versions of OS X.

PR #2377, opened 2018-02-26
add remote submit from Condor parameter tuning capability

nikolasgianna:
This pull request is to make it possible to set the following parameters: 

`rsc_fpops_est, rsc_fpops_bound, rsc_memory_bound, rsc_disk_bound, delay_bound, app_version_num `

when submitting to BOINC through Condor.  

See: https://github.com/htcondor/htcondor/pull/19

### DIFF
--- a/html/user/submit_rpc_handler.php
+++ b/html/user/submit_rpc_handler.php
@@ -201,7 +201,8 @@ function stage_files(&$jobs) {
 // submit a list of jobs with a single create_work command.
 //
 function submit_jobs(
-    $jobs, $template, $app, $batch_id, $priority, $app_version_num,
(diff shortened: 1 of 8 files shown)


TheAspens:
@davidpanderson - can you review this?  If not, it is about time I learn how to use the remote submit capability and I'll test this later this week.

davidpanderson:
We combine job parameters in a input template structure instead of passing them as separate args.
submit_batch() already lets you pass an input template for each job.  The RPC handler puts these in temp files.  We should use the same approach for batch-level templates.

davidpanderson:
I'll submit something in a few days that provides the same functionality.

TheAspens:
Closed in deference to #2416

PR #2378, opened 2018-02-26
Fix copyright year

AenBleidd:
Copyright year changed in all files except pot (and similar) files.


CharlieFenton:
Should these have been merged into the client_release/7/7.10 branch instead of (or in addition to) master?

AenBleidd:
@CharlieFenton, if this will be merged to release branch then localization
files should be updated too

Best regards,
Vitalii Koshura

2018-02-27 3:43 GMT+02:00 CharlieFenton <notifications@github.com>:

davidpanderson:
Not worth doing

RichardHaselgrove:
At the moment, I have a list of four PRs that should be in the release branch, but are currently only in master. One (#2374) is critical: the others are 'nice to have' and may as well come along for the ride. I'd put any late arrivals to the translation ball somewhere between those two levels of importance.

I've got two plans in mind:
1) Discuss and reconfirm the long-term process model at today's meeting (27 Feb)
2) Allow testing and bug-hunting to continue a little longer, then prepare a single cherrypick PR to pull all current changes into the 7.10 branch ready for the next build. That's looking like a v7.10.1 release candidate at this stage.

Ageless93:
You're going for an RC based on what test results? The Alpha site has 1
person who put his test results in. One.
If going to an RC is that easy, we can do without the alpha testers. Skip
that step. Do an immediate release to the public, see what they come back
with.


-- Jord van der Elst.

JuhaSointusalo:
Where the keyword stuff should be tested at? I'm not, for example, seeing them in any task properties.

RichardHaselgrove:
This isn't really on-topic for this particular PR, but since we're all here...

@Ageless93 - the emphasis would be on 'candidate': certainly there couldn't be a promotion to 'recommended' until the matrix is a lot fuller, and we've got a lot more positive reports from people who have tried (and failed!) to break the edge cases. The only practical effects would be an even version number and taking off the internal 'may be unstable' flag. Doesn't matter - we make that change once only, and it can be sooner or later.

@JuhaSointusalo - Good question. Where, also, is the full test for auto-attach? I tried at SETI, but could only get v7.8.3, with or without ?dev=1 on the url. Maybe we're waiting for #2355.

@davidpanderson - if I'm making a PR to the release branch anyway, it's not worth NOT doing.

RichardHaselgrove:
@JuhaSointusalo - I asked about the keywords in our WG meeting: no projects are currently using them, and they probably won't become available until the upstream TACC server for Science United has projects and tasks to distribute - which won't happen for at least a couple of months. So we'll have to cross that bridge when we come to it.

JuhaSointusalo:
Thanks. Lets hope there aren't any bad bugs then.

PR #2381, opened 2018-02-28
web: fix bugs in change-password page.

davidpanderson:
BTW, I'd like to have a "show password" checkbox here
for people with no one looking over their shoulder.
Maybe later.


TheAspens:
Appears to work correctly.

PR #2382, opened 2018-02-28
web: add "show password" checkbox to login form

davidpanderson:


TheAspens:
This is good to merge, but it would also be handy to have this on the Join page as well as the change password page.  Do you want to add those before I merge?

TheAspens:
Actually I see that you just opened another password related pull request.  Let me merge this one and then add the above comment to that one.

PR #2384, opened 2018-03-01
client: not an error if a file starts with account_ but is not accounting file

davidpanderson:
For example, if you rename account_X.xml to account_X.xml.bak,
if should not be reported as an error.
Same for statistics files.


TheAspens:
Looks good.  Thanks.

PR #2385, opened 2018-03-01
web: in download.php, use current login token if recent

davidpanderson:
Generate a new login token only if the current one is older than a day.
That way if a user does several downloads in quick succession
(for whatever reason) autoattach will work with all of them.


TheAspens:
Looks good.  Thanks.

PR #2388, opened 2018-03-02
client: show correct info about client version

davidpanderson:
On startup, the client prints a message describing itself.
This should say the architecture for which the client was compiled,
not the architecture of the host on which it's running.

### DIFF
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -257,13 +257,11 @@ void CLIENT_STATE::show_host_info() {
             "VirtualBox version: %s",
             host_info.virtualbox_version
         );
-    } else {
 #if defined (_WIN32) && !defined(_WIN64)
-        if (!strcmp(get_primary_platform(), "windows_x86_64")) {
-            msg_printf(NULL, MSG_USER_ALERT,
-                "Can't detect VirtualBox because this is a 32-bit version of BOINC; to fix, please install a 64-bit version."
-            );
-        }
+    } else {
+        msg_printf(NULL, MSG_USER_ALERT,
+            "Can't detect VirtualBox because this is a 32-bit version of BOINC; to fix, please install a 64-bit version."
+        );
 #endif
     }
 }
@@ -413,6 +411,31 @@ static void set_client_priority() {
 #endif
 }
 
+// return the compile target of the BOINC client
+// (not necessarily the same has the host architecture)
+// TODO: flesh this out for Linux (architecture, bitness)
+//
+static const char* client_target() {
+#ifdef _WIN32
+#ifdef _WIN64
+    return " for Windows 64-bit";
+#endif
+    return " for Windows 32-bit";
+#endif
+
+#ifdef __APPLE__
+    return " for Mac OS X";
+#endif
+
+#if LINUX_LIKE_SYSTEM
+    return " for Linux";
+#endif
+
+#ifdef ANDROID
+    return " for Android";
+#endif
+}
+
 int CLIENT_STATE::init() {
     int retval;
     unsigned int i;
@@ -434,13 +457,13 @@ int CLIENT_STATE::init() {
     time_stats.start();
 
     msg_printf(
-        NULL, MSG_INFO, "Starting BOINC client version %d.%d.%d for %s%s",
+        NULL, MSG_INFO, "Starting BOINC client version %d.%d.%d%s%s",
         core_client_version.major,
         core_client_version.minor,
         core_client_version.release,
-        get_primary_platform(),
+        client_target(),
 #ifdef _DEBUG
-        " (DEBUG)"
+        " (Debug)"
 #else
         ""
 #endif


JuhaSointusalo:
Microsoft has ported desktop Windows to ARM64 processors. It runs x86 programs with emulator but with limitations. Apple has been rumoured to be doing the same for years.

So "32/64-bit" isn't enough. It needs to include the processor architecture.

The way you added code for only a few operating systems means the code either won't compile or fails at runtime on BSDs and others.

`HOSTTYPE` includes both OS and processor type, is accurate and would have been readable enough. If you absolutely must have `HOSTTYPE` prettified then at least have the code fall back to `HOSTTYPE` for the "unknown" cases.

The client uses primary platform incorrectly in several places. You should have fixed those too.

And finally, you made a change to the VirtualBox message, didn't say anything about it in the commit message and actually broke the message. Now 32-bit client running on 32-bit Windows without VirtualBox tells the user to install 64-bit client.

RichardHaselgrove:
Concern was expressed yesterday that this might not work on fully 32-bit systems. I've compiled and installed this: I get

`21/03/2018 11:52:51 |  | Starting BOINC client version 7.9.3 for Windows 32-bit
21/03/2018 11:52:53 |  | VirtualBox version: 5.2.8`

which seems good enough to fix the presenting problem for v7.9.4

I agree that further development for different architectures etc. would be good for v7.12: there's a TODO in the commit

RichardHaselgrove:
On the other hand, after removing VirtualBox, I get

```
21/03/2018 12:11:06 |  | Starting BOINC client version 7.9.3 for Windows 32-bit
21/03/2018 12:11:08 |  | Can't detect VirtualBox because this is a 32-bit version of BOINC; to fix, please install a 64-bit version.
```

as @JuhaSointusalo says. That's a blocker.

davidpanderson:
:Seems correct to me.
A 32-bit client can't tell whether VBox is installed.
A 64-bit client would be able to tell.
It's always been this way.

On Wed, Mar 21, 2018 at 5:18 AM, RichardHaselgrove <notifications@github.com

RichardHaselgrove:
See my first log extract. That build successfully detected the *32-bit* VBox installed under the 32-bit version of Windows 7. The hardware (elderly Dell Precision Workstation) is fully 64-bit certified and has virtualisation-enabled Xeon CPUs. I've even run 64-bit VBox apps under the 32-bit OS simply by writing the 32-bit wrapper into an app_info.xml file.

It all works, *provided* you keep the bitness matched. 32-bit BOINC, 32-bit OS, 32-bit VBox is a working combination.

client/client_state.cpp
```diff
@@ -413,6 +411,31 @@ static void set_client_priority() {
...
+
+#if LINUX_LIKE_SYSTEM
+    return " for Linux";
+#endif
+
+#ifdef ANDROID
+    return " for Android";
+#endif
+}
+
```
TheAspens: Do we need a default return type of " for Unknown" or some message in the event that none of the flags _WIN32/_WIN64/__APPLE__/LINUX_LIKE_SYSTEM/ANDROID is set?  I don't know all the cases that things are built and if not having one of these set causes problems elsewhere in the code, but it seems like there should be a default to handle that case.  

Note:  If the code absolutely requires one of these flags to be set, then I withdraw my comment.  However, if the intention is to be able to build the client without one of these flags then a default should exist.

TheAspens: Additional comment, based on Juha's comments above, the use of HOSTTYPE in the default case would be better than "Unknown"

TheAspens:
The message "Can't detect VirtualBox because this is a 32-bit version of BOINC; to fix, please install a 64-bit version." is misleading on multiple accounts.  It is espicially bad since this now apparently shows up as a notice that there is an error.

1. The user might not be participating in projects that use VirtualBox - thus no action is required
2. The user might not want to run VirtualBox apps - thus no action is required
3. The user might want to install 32 bit Virtual Box rather than 64 bit BOINC

BOINC should not imply that an error condition exists when it is possible that there is not an error condition.  In 2 out of my 3 items above, there is not an error condition and the user does not need to take any action.  A correct message would state something like:
"VirtualBox not detected.  If you want BOINC to be able to use VirtualBox then either install a 32 bit version of VirtualBox or install a 64 bit version of BOINC"

It might be better to have a message like:
"VirtualBox not detected.  If you want BOINC to be able to use VirtualBox, the please see https://useful.wiki.page.on.boinc.site"

The user has a lot of options to getting it configured properly and pointing them to a help page will be able to provide the range of options that the message log cannot.

RichardHaselgrove:
Even worse - if the users follows the advice to install a 64-bit version of BOINC on a 32-bit version of windows (which is where the problem arises), they hit this rather clunky error message from Microsoft:

64 on 32 error

My own way of dealing with a similar problem at SETI was:

64 on 32 warning

Conversely, there is no warning or advice at all if you install a 32-bit version of BOINC on a 64-bit version of Windows. That's the user error which started this whole conversation, and which could not be deduced from the initial line in the current Event Log at startup.

RichardHaselgrove:
Would it be possible for one of the committers to do a very simple commit / PR to correct the first-line event log message only, please? That caused us considerable confusion when we started this discussion, but the extra matters aren't needed: the old usage of the VBox message was actually correct and didn't need changing.

With a quick and simple change, we can get it into v7.10 and keep the progress moving forward. Thanks.

davidpanderson:
How about if we remove the line about not being able to detect vbox64?
It's not accurate, and saying something accurate would be too complex.

RichardHaselgrove:
It's accurate for the problem it was designed for: running a 32-bit BOINC on a 64-bit operating system. The inaccuracy is having it appear when BOINC is running on a 32-bit operating system, when the advice to install a 64-bit BOINC can't be followed.

I'll re-compile and re-test with the if (!strcmp(get_primary_platform(), "windows_x86_64")) line reinstated.

RichardHaselgrove:
I've completed tests with lines 260-266 of 538696850b8f4f6f6a05c46030402bf909a62984 reinstated, in all possible combinations (32 on 32, 64 on 64, and 32 on 64), and all messages appear as I would wish them to.

If any change is desired, I'd suggest changing the mixed-mode message to 

> Can't detect VirtualBox because this is a 32-bit version of BOINC running on a 64-bit operating system; to fix, please install a 64-bit version of BOINC.

by adding "running on a 64-bit operating system" and "of BOINC" - to make it clearer which parts are in conflict, and which is easiest to correct. As @TheAspens says, the message appears as a notice, so space isn't a problem.

TheAspens:
I think I didn't state my opinion clearly enough above.  I think that only one of the two following is the correct thing to do:

- No message should be displayed
- A simple generic message is displayed. I recommend the message 

> VirtualBox not detected. If you want BOINC to be able to use VirtualBox, the please see https://useful.wiki.page.on.boinc.site

You cannot give users a very prominent message stating that something is wrong and that they must taken action when in fact that may not be true.

TheAspens:
I should also add that it definitely should not be a notice.

RichardHaselgrove:
OK, I can live with those observations. :-) The useful.wiki.page.on.boinc.site would be

http://boinc.berkeley.edu/wiki/VirtualBox

That's an existing part of the User Manual, and I've added the paragraph about avoiding mixed-mode running that's troubling us here. But being the user Wiki, we have to keep it at http until fixed, not https.

RichardHaselgrove:
Pleased to note that the useful.wiki.page.on.boinc.site would be

https://boinc.berkeley.edu/wiki/VirtualBox

The user manual is now accessible via https

RichardHaselgrove:
The core problem has been resolved in the v7.10 release. Further observations and discussions are active elsewhere. Closing this iteration as redundant.

PR #2394, opened 2018-03-07
web: improve - but not fix - password visibility checkbox

davidpanderson:
When you check the box it turns the input field type from "password" to "text".
If you navigate to another (say by clicking Submit)
and then return via the back button,
the checkbox is still checked but the input field type is back to "password"!
Further clicks toggle the type, so it's always wrong from that point.

I changed it so that clicking the checkbox sets the type
to the appropriate value rather than toggling it.
The type is still wrong after "back",
but it's right after further checkbox clicks.

Note: the example on w3schools.com
https://www.w3schools.com/howto/howto_js_toggle_password.asp
has the same bug.

### DIFF
--- a/html/inc/account.inc
+++ b/html/inc/account.inc
@@ -35,15 +35,16 @@ function passwd_visible_checkbox($name) {
     return sprintf('
 <script>
 function toggle_passwd() {
+    var c = document.getElementById("passwd_visible");
     var x = document.getElementById("%s");
-    if (x.type === "password") {
+    if (c.checked) {
         x.type = "text";
     } else {
         x.type = "password";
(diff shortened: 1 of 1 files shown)


TheAspens:
It would also be handy to have this on the Join page as well as the change password page. Do you want to add those before I test?

davidpanderson:
It's already in the SU join page.
(are you thinking of a different page?)

TheAspens:
@davidpanderson - I created pull request #2395 to show you what I meant.  If you are ok with it, please merge it (it will merge into this branch) and then I'll merge this one.  Thanks

PR #2395, opened 2018-03-07
Add 'show password' to join and change password pages

TheAspens:
web: add 'show password' to join and change password pages.  add label to text for show password to make it easier to select the show password option

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,10 @@ test/version.inc
 *.lo
 *.a
 *.la
+py/lib/Boinc
+
+## Python bytecode
+*.pyc
 
 ## Drupal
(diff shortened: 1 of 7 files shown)


davidpanderson:
To me, it doesn't make sense to have two password fields, one with "Show password" and one without.
You still have to type the (hidden) password in the 2nd field.

It would make more sense if, when "Show password" is checked, and 2nd field goes away.

The <label> thing is good.

TheAspens:
You are basically suggesting this: http://uxmovement.com/forms/why-the-confirm-password-field-must-die/ which makes sense.  Let me work this a bit more.

TheAspens:
Ok - changes made.  If you are happy with it, please merge.  Thanks.

PR #2405, opened 2018-03-10
let AMs use authenticators instead of name/password

davidpanderson:


tristanolive:
This seems reasonable to me. So if the AMS response includes an authenticator, then the client will store it in client state and use it for all subsequent AMS requests. Otherwise, backwards compatibility remains with password hashes. If I'm reading it right, t's basically an official implementation of the opaque auth method that has been used successfully for years.

davidpanderson:
More or less.  When the opaque field was used to pass an ID, the name and password fields were still populated on the client.  In fact an AM was considered to be in use if and only if these 2 fields were populated.

TheAspens:
I built this branch and tested with science united and was able to attach with my existing account.  However, when I attempted to force a sync manually I got:

```
Tue 27 Mar 2018 10:07:12 AM CDT |  | Fetching configuration file from https://scienceunited.org/get_project_config.php
Tue 27 Mar 2018 10:07:16 AM CDT |  | Contacting account manager at https://scienceunited.org/
Tue 27 Mar 2018 10:07:17 AM CDT | https://scienceunited.org/ | Message from account manager: project->update failed
```


```
[knreed@oc7811586721 boinc]$ cat acct_mgr_reply.xml 
<acct_mgr_reply>
    <error_num>-1</error_num>
    <error_msg>project->update failed</error_msg>
</acct_mgr_reply>

```

I also could not find how to change my email address or password on Science United so I was not able to test that.

TheAspens:
I also tested with BAM.  BAM worked correctly and I had no issues with changing my email address or password and being able to continue to synchronize BAM after changing those.

davidpanderson:
Please try again.
-- D

TheAspens:
I was able to connect to Science United and do follow up syncs.  However, since it is not possible to change your email address or password at science united, I was not able to test that.  Since the account manager login file does not appear to contain any credentials based on email address and password, it looks like the code should work fine from the client.

I would recommend that you allow users to change their email address and password at Science United.

TheAspens:
I should also state that I tested BAM as well and it continued to work fine.

davidpanderson:
I put back the change email/password functions.
But they'll only work once we get this feature in the client.

PR #2406, opened 2018-03-13
Fix buffer size for call of function GetWindowText

versat:
Cppcheck found this issue: Multiplying sizeof() with sizeof() indicates a logic error.
GetWindowText expects the number of characters (3. argument) that fit into the buffer (2. argument).
So a division seems to be correct here instead of a multiplication.

### DIFF
--- a/clientsetup/win/CAShutdownBOINCManager.cpp
+++ b/clientsetup/win/CAShutdownBOINCManager.cpp
@@ -83,7 +83,7 @@ UINT CAShutdownBOINCManager::OnExecution()
         hWndBOINCManagerSystray = FindWindow( _T("wxTaskBarExWindowClass"), _T("BOINCManagerSystray") );
         if ( NULL != hWndBOINCManagerSystray )
         {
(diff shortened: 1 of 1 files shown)


TheAspens:
Scrutinizer errored out.  I'm closing and re-opening the pull request in order to restart those tests.

TheAspens:
@JuhaSointusalo or @AenBleidd - can one of you review this one?  Thanks.

AenBleidd:
@TheAspens, I can but looking at this file I have 1 question? Is it still used?
As I see there is no project for VS2013 with this file and it contains in VS2010 project only (boinccas.vsxproj). If I remember correctly I was not successful building it with VS2013 Community Edition  because it contains dependencies on some old ATL libraries. Also this file (boinccas.dll) exists in git repo as pre-built object.
Is this dll still used? If yes - what for?
Thanks

AenBleidd:
I'm ok with this fix.
It can be merged despite any answer on my question

JuhaSointusalo:
@AenBleidd `boinccas` is part of the installer. I have full version of VS2010 so I can build it. I don't have InstallShield so I can't really test it though.

PR #2417, opened 2018-03-20
Web: use https links to BOINC web site in default project config

davidpanderson:


TheAspens:
Looks good - thanks.

RichardHaselgrove:
Are we going to use this on boinc-site as well? In that case, the user manual would need attention.

https://boinc.berkeley.edu/wiki/User_manual

Issue #1554, opened 2016-06-07
Display Buttons contents with three little dots


DPhilippe4:
All the Boinc Manager buttons are non dynamically resized taking into account the text length. Consequently, it is even sometimes difficult to guess the missing words.

Here is a French example but all languages are probably impacted:

![capture d ecran 2016-06-07 a 18 04 15](https://cloud.githubusercontent.com/assets/19802944/15866497/29f6c31e-2cdf-11e6-890e-b7b702b724a2.png)

![capture d ecran 2016-06-07 a 18 04 43](https://cloud.githubusercontent.com/assets/19802944/15866499/2b51b4bc-2cdf-11e6-9d90-8335b16f5e95.png)

ChristianBeer:
A workaround is to hover over the buttons and a tooltip text will appear.

Ageless93:
@CharlieFenton, are the buttons in Advanced view a set size, or can they grow when the text grows? 
Wouldn't it be better to add word wrap so the text continues on a second line?

CharlieFenton:
Looking at the code, I see that CBOINCTaskCtrl::UpdateControls() has:
        layoutChanged = 1;
and:
    if (layoutChanged) {
        Fit ();
    }
so this should already be happening. However, it also calls:
    EllipseStringIfNeeded(pItem->m_strNameEllipsed);
wich intentionally limits the string length to:
    int maxWidth = TASKBUTTONWIDTH - 10;
where:
#define TASKPANEWIDTH ADJUSTFORXDPI(200)
#define TASKBUTTONWIDTH ADJUSTFORXDPI(TASKPANEWIDTH - 55)

If I recall correctly, we actually allowed it to expand to an unlimited width several years ago, but the result was too annoying so we limited it. I believe that the problem is that the text in the buttons can changes dynamically, for example as you select or deselect rows, and the number of buttons can even change dynamically. This can cause the "task controls" area (as we call it) to keep changing size as you select / deselect rows, which people found undesirable.

It also changes the width of the area displaying the rows, which was also annoying to people. We didn't try wrapping, but that would have a similar effect on the vertical size and button spacing of the task control area.

The change to limit the size and ellipse the text was made in a series of commits on 1/25/12, but unfortunately the commit message doesn't explain the reason.

The checkin_notes file just says:
    - MGR: Make the task pane in the advanced view and its buttons fixed sizes; 
        ellipse the button contents if needed; show full button text plus 
[...]

PR #2418, opened 2018-03-20
 Mac: Fix wxWidgets 3.1.0 Mac-only bug when wxStaticBox has no label

CharlieFenton:
This patches wxWidgets 3.1.0 on the Mac to fix the bug reported by @DPhilippe4 in PR #1554. This bug was introduced in the Mac (wxCocoa) port of wxWidgets 3.1.0, and has been fixed in wxWidgets 3.1.1. But I feel it is too late in the BOINC 7.10 release cycle to upgrade wxWidgets to 3.1.1 as it won't get adequate testing. (For example, this fairly obvious problem with wxWidgets 3.1.0 was not noticed until the third alpha release 7.9.3; it was not reported in 7.9.1 or 7.9.2 though present in both.)

While analyzing this bug, I have been experimenting with wxWidgets 3.1.1 and found that several changes are needed to BOINC source code to upgrade from 3.1.0 to 3.1.1. I will put those in a separate feature branch for inclusion in the 7.11 / 7.12 release cycle.

### DIFF
--- a/mac_build/buildWxMac.sh
+++ b/mac_build/buildWxMac.sh
@@ -35,6 +35,7 @@
 # Build 64-bit library (temporarily build both 32-bit and 64-bit libraries) 10/22/17
 # Update for wxCocoa 3.1.0 10/25/17
 # Build only 64-bit library 1/25/18
+# Fix wxWidgets 3.1.0 bug when wxStaticBox has no label 3/20/18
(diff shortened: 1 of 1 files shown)


TheAspens:
Looks good.  Merging.  Thanks Charlie!

PR #2419, opened 2018-03-22
Default working directory changd to /var/lib/boinc

lfield:
This change should go into the 7.10 client release.


vaneldik:
Thanks. 

Indeed, systemd does ignores the relative path "~" on CentOS 7:
 ```
Mar 22 09:19:30 HOSTNAME.cern.ch systemd[1]: [/usr/lib/systemd/system/boinc-client.service:11] Not an absolute path, ignoring: ~ 
```

TheAspens:
I tested on Red Hat Desktop 7 and confirmed that ~ didn't work but /var/lib/boinc does.  I will merge.  Thanks!

vaneldik:
Thanks a lot!

PR #2424, opened 2018-03-27
Drupal: Changed email templates for user registration, activation

drshawnkwang:
password reset, etc.

https://dev.gridrepublic.org/browse/DBOINCP-431

### DIFF
--- a/drupal/sites/all/features/boinc_standard/boinc_standard.strongarm.inc
+++ b/drupal/sites/all/features/boinc_standard/boinc_standard.strongarm.inc
@@ -1995,7 +1995,7 @@ function boinc_standard_strongarm() {
   $strongarm->disabled = FALSE; /* Edit this to true to make a default strongarm disabled initially */
   $strongarm->api_version = 1;
   $strongarm->name = 'user_mail_password_reset_body';
-  $strongarm->value = '!mailto,
+  $strongarm->value = 'Hello,
 
 A request to reset the password for your account has been made at !site.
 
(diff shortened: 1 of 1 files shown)


drupal/sites/all/features/boinc_standard/boinc_standard.strongarm.inc
```diff
@@ -2049,21 +2049,9 @@ After logging in, you will be redirected to !edit_uri so you can change your pas
...
-You may also log in by clicking on this link or copying and pasting it in your browser:
-
-!login_url
-
-This is a one-time login, so it can be used only once.
-
-After logging in, you will be redirected to !edit_uri so you can change your password.
+  $strongarm->value = '!mailto,
 
+Thank you for registering at !site. You may now log in to !login_uri using the following email address and your passphrase.
```
tristanolive: This section of code seems to be missing the "following email address", as all that follows is the signature line of the message.

drupal/sites/all/features/boinc_standard/boinc_standard.strongarm.inc
```diff
@@ -2049,21 +2049,9 @@ After logging in, you will be redirected to !edit_uri so you can change your pas
...
-password: !password
-
-You may also log in by clicking on this link or copying and pasting it in your browser:
-
-!login_url
-
-This is a one-time login, so it can be used only once.
-
-After logging in, you will be redirected to !edit_uri so you can change your password.
+  $strongarm->value = '!mailto,
```
tristanolive: This is using the email address as the salutation of the message? Seems strange, though I understand that username is not ideal, since that is not something the user normally sees. Can display name not be used here?

drupal/sites/all/features/boinc_standard/boinc_standard.strongarm.inc
drshawnkwang: Thanks for catching this. I'll change it and update the PR.

drupal/sites/all/features/boinc_standard/boinc_standard.strongarm.inc
drshawnkwang: Unfortunately the BOINC display name is not known the core user's module. Thus the choice to use the user's email address in these emails. It's not a great solution. 

A more sophisticated piece of code would be a boincuser (sub) module that handled the emails to users for registration, changing passwords, etc.

PR #2426, opened 2018-03-27
create_account web RPC: improve parsing and error handling

davidpanderson:
Projects with old server code reply with PHP warnings
interspersed with the XML.
Parse these replies manually.


TheAspens:
built and tested using automated build and test at https://github.com/TheAspens/boinc-server-test

PR #2429, opened 2018-03-27
Cherry Picks to Client Release 7.10 branch

TheAspens:
This includes the merge commits requested by Richard (the release manager for 7.10) for the next client build.  Specifically, pull requsts #2406, #2418, #2419.  (Note that #2418 was already merged so nothing shows for it).


RichardHaselgrove:
@TheAspens - now that they've been merged, could you add #2405 and #2420 to the cherrypicks, please?

TheAspens:
Any others?  Sorry for the delay on this.

RichardHaselgrove:
Could we do #2436? I think that would be a good point to stop and build - as 7.9.4, or are we ready for a 7.10.1? At this time of night, perhaps 7.9.4 - we have the project list to sort out still.

davidpanderson:
What are the issues with the project list?

RichardHaselgrove:
Midnight here, but from memory:

DENIS isn't in the 7.10 branch file
There are keyword IDs in the latest 14-day update version, but not in the branch file
We had another scare about Einstein (their message board, a report) - SHAWN KWANG was looking into it about 10 hours ago. But I think it's OK - awaiting confirmation how it happened.

davidpanderson:
I added DENIS and the keywords (which aren't currently used, but don't
hurt) to the 7.10 file.

The URL for E@h in the file is https://einsteinathome.org
Do you know if this is correct?

-- David

RichardHaselgrove:
I believe so, and Oliver confirmed it when I changed it last time.

But the bug report thread is https://einsteinathome.org/content/project-url-out-date - judge for yourself

PR #2432, opened 2018-03-29
cppcheck: fix some errors

serval2412:
[client/cs_trickle.h:32]: (performance) Variable 'url' is assigned in constructor body. Consider performing initialization in initialization list.
[client/acct_mgr.h:170]: (performance) Variable 'global_prefs_xml' is assigned in constructor body. Consider performing initialization in initialization list.
[client/scheduler_op.cpp:937]: (performance) Variable 'message' is assigned in constructor body. Consider performing initialization in initialization list.
[client/scheduler_op.cpp:938]: (performance) Variable 'priority' is assigned in constructor body. Consider performing initialization in initialization list.
[clientsetup/win/boinccas.cpp:42]: (performance) Variable 'm_strActionName' is assigned in constructor body. Consider performing initialization in initialization list.
[clientsetup/win/boinccas.cpp:43]: (performance) Variable 'm_strProgressTitle' is assigned in constructor body. Consider performing initialization in initialization list.
[vda/vda_lib.cpp:78]: (performance) Variable 'coding' is assigned in constructor body. Consider performing initialization in initialization list.

[samples/gfx_html/mongoose.cpp:2375]: (style) Array index 'i' is used before limits check.
[zip/zip/z_fileio.c:3879]: (style) Array index 'argnum' is used before limits check.

### DIFF
--- a/client/acct_mgr.h
+++ b/client/acct_mgr.h
@@ -169,12 +169,11 @@ struct ACCT_MGR_OP: public GUI_HTTP_OP {
     int parse(FILE*);
     virtual void handle_reply(int http_op_retval);
 
-    ACCT_MGR_OP(GUI_HTTP* p) {
+    ACCT_MGR_OP(GUI_HTTP* p) : global_prefs_xml("") {
         gui_http = p;
         via_gui = false;
         error_num = BOINC_SUCCESS;
         repeat_sec = 60.0;
-        global_prefs_xml = "";
         safe_strcpy(host_venue, "");
         got_rss_feeds = false;
     }
--- a/client/cs_trickle.h
+++ b/client/cs_trickle.h
@@ -28,7 +28,7 @@ struct TRICKLE_UP_OP: public GUI_HTTP_OP {
     int error_num;
     char* req_buf;
 
-    TRICKLE_UP_OP(std::string& u) : url(u), error_num(0), req_buf(NULL) {
+    TRICKLE_UP_OP(std::string& u) : reply(), url(u), error_num(0), req_buf(NULL) {
         gui_http = new GUI_HTTP;
     }
     virtual ~TRICKLE_UP_OP(){}
--- a/client/cs_trickle.h
+++ b/client/cs_trickle.h
@@ -28,11 +28,8 @@ struct TRICKLE_UP_OP: public GUI_HTTP_OP {
     int error_num;
     char* req_buf;
 
-    TRICKLE_UP_OP(std::string& u) {
-        url = u;
-        error_num = 0;
+    TRICKLE_UP_OP(std::string& u) : url(u), error_num(0), req_buf(NULL) {
         gui_http = new GUI_HTTP;
-        req_buf = NULL;
     }
     virtual ~TRICKLE_UP_OP(){}
     int do_rpc(const char*);
--- a/client/scheduler_op.cpp
+++ b/client/scheduler_op.cpp
@@ -933,8 +933,5 @@ int SCHEDULER_REPLY::parse(FILE* in, PROJECT* project) {
 }
 #endif
 
-USER_MESSAGE::USER_MESSAGE(char* m, char* p) {
-    message = m;
-    priority = p;
-}
+USER_MESSAGE::USER_MESSAGE(char* m, char* p) : message(m), priority(p) {}
 
--- a/clientsetup/win/boinccas.cpp
+++ b/clientsetup/win/boinccas.cpp
@@ -34,13 +34,11 @@ BOINCCABase::BOINCCABase(
     MSIHANDLE hMSIHandle, 
     const tstring strActionName, 
     const tstring strProgressTitle
-    )
+    ) : m_strActionName(strActionName), m_strProgressTitle(strProgressTitle)
 {
 
     // Store the parameters for later use
     m_hMSIHandle = hMSIHandle;
-    m_strActionName = strActionName;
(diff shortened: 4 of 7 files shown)


zip/zip/z_fileio.c
```diff
@@ -3876,7 +3876,7 @@ int insert_arg(pargs, arg, at_arg, free_args)
    argnum = 0;
    newargnum = 0;
    if (args) {
-     for (; args[argnum] && argnum < at_arg; argnum++) {
+     for (; argnum < at_arg && args[argnum]; argnum++) {
```
AenBleidd: I suppose this is not important fix because it is a 3rd party component.

client/acct_mgr.h
```diff
@@ -169,12 +169,11 @@ struct ACCT_MGR_OP: public GUI_HTTP_OP {
     int parse(FILE*);
     virtual void handle_reply(int http_op_retval);
 
-    ACCT_MGR_OP(GUI_HTTP* p) {
+    ACCT_MGR_OP(GUI_HTTP* p) : global_prefs_xml("") {
```
AenBleidd: This is a question of style.
I see no particular purpose in this change

client/cs_trickle.h
```diff
@@ -28,11 +28,8 @@ struct TRICKLE_UP_OP: public GUI_HTTP_OP {
     int error_num;
     char* req_buf;
 
-    TRICKLE_UP_OP(std::string& u) {
-        url = u;
-        error_num = 0;
+    TRICKLE_UP_OP(std::string& u) : url(u), error_num(0), req_buf(NULL) {
```
AenBleidd: Same here

client/scheduler_op.cpp
```diff
@@ -933,8 +933,5 @@ int SCHEDULER_REPLY::parse(FILE* in, PROJECT* project) {
 }
 #endif
 
-USER_MESSAGE::USER_MESSAGE(char* m, char* p) {
-    message = m;
-    priority = p;
-}
+USER_MESSAGE::USER_MESSAGE(char* m, char* p) : message(m), priority(p) {}
```
AenBleidd: Same here

clientsetup/win/boinccas.cpp
```diff
@@ -34,13 +34,11 @@ BOINCCABase::BOINCCABase(
     MSIHANDLE hMSIHandle, 
     const tstring strActionName, 
     const tstring strProgressTitle
-    )
+    ) : m_strActionName(strActionName), m_strProgressTitle(strProgressTitle)
```
AenBleidd: Same here

vda/vda_lib.cpp
```diff
@@ -72,10 +72,9 @@ const char* status_str(int status) {
 META_CHUNK::META_CHUNK(
     VDA_FILE_AUX* d, META_CHUNK* par, double size,
     int coding_level, int index
-) {
+) : coding(d->policy.codings[coding_level]) {
     dfile = d;
     parent = par;
-    coding = d->policy.codings[coding_level];
```
AenBleidd: Same here

zip/zip/z_fileio.c
serval2412: Indeed but i don’t think it’d trigger some pb to include this change.

samples/gfx_html/mongoose.cpp
```diff
@@ -2372,7 +2372,7 @@ int mg_url_decode(const char *src, int src_len, char *dst,
 #define HEXTOI(x) (isdigit(x) ? x - '0' : x - 'W')
 
   for (i = j = 0; i < src_len && j < dst_len - 1; i++, j++) {
-    if (src[i] == '%' && i < src_len - 2 &&
+    if (i < src_len - 2 && src[i] == '%' &&
```
AenBleidd: This is also a question of style. src[i] will always be valid because of check if 'for' statement

client/acct_mgr.h
serval2412: It’s more optimized to use initialization list for constructors. Of course the gain is micro perf but why letting it away

samples/gfx_html/mongoose.cpp
serval2412: In for you compare i with src_len, in if you compare i with src_len  - 2, so it’s not only style

AenBleidd: I mean that there will be no out of bounds exception. So these two statements are equal in performance.

AenBleidd:
Existing code of this project doesn't have ideal style of code.
In projects like this I see no particular reason to change code style in the whole project without some real performance gains.

serval2412:
I reverted samples/gfx_html/mongoose.cpp part.
For the rest, since this project hasn't ideal style code, I don't see any reason to avoid even micro perf gains.
Of course, real perf gains would be far better but we're talking about only straightforward changes not a refactoring which could bring some regressions.

ChristianBeer:
These changes are not fixing errors (as is indicated in the title). I would prefer if all initializations happen in the body of the constructor in the same order as in the declaration. This way you can see at glance that every variable is initialized. If you pull the variables apart this is not easy to see. So in this case I would prefer readability over performance as the performance gains are negligible.

serval2412:
Argh! I misread your comment and just pushed a commit to put initializations in initialization list not in body.
Ok let's abandon this since I dislike initialization in body.

PR #2436, opened 2018-03-29
client: show build not OS platform in messages

### LINKED ISSUE #2386: How to recognize the bit-depth of the client.?
In all these years I believed that when BOINC stated in the event log "Starting BOINC client version 7.9.2 for windows_x86_64" that the windows_x86_64 part meant that the client was 64 bit, running on Windows. But through a comment by Juha it appears this is not the way. It shows what operating system it has detected, in the above case Windows 64 bit, but not necessarily that the client is 64 bit. 

When installing the 32 bit client on a 64 bit Windows, it will start with "Starting BOINC client version 7.9.2 for windows_x86_64", while it will only report "windows_intex86" when installed on a 32 bit Windows.

Now then, where can we find information about the bit-depth of the client? 
In BOINC Manager's About window it says via Version: N.N.N (x86) or N.N.N (x64)
However, when one cannot start BOINC Manager for some reason and one has to check via the command line with boinc.exe --version, the 32 bit version will report N.N.N windows_intelx86, while the 64 bit version will report N.N.N windows_x86_64. Here the windows_* indicate the bit-depth of the client, not that of the operating system. That is confusing, to me anyway.

So I am putting this discussion down, with questions. For user-friendliness, and for ease of mind of us oldies: 

1. Should we continue to have the client start by saying what OS it detected, even when that same detection is later on in the log written with the full name of the OS itself? 
I.e. _OS: Microsoft Windows 7: Ultimate x64 Edition, Service Pack 1, (06.01.7601.00)_
2. Or should the client start with information for the user what client he installed, with the same information as it is shown on the installer he used? 
I.e. _"Starting BOINC client version 7.9.2, 32-bit (x86) for Windows"_, _"Starting BOINC client version 7.9.2, 64-bit (x86_64) for Windows"_ 
3. Or is there a third option we can use? 

In any case, if 1. then IMO the command line version reporting is incorrect. That should be N.N.N intelx86/x86_64, not including the windows_ part.

JuhaSointusalo:
The client shows what kind of operating system it has detected in startup log message and elsewhere. This is confusing and not helping in troubleshooting when people expect these messages to describe the client itself.

Change the messages to use client's build platform instead.

Startup log message, output of --version and user agent string should be safe to change without breaking anything. The client still uses OS platform in various places but changing these would break stuff:

- Scheduler request XML. Scheduler expects scheduler_request.platform_name to be OS primary platform.
- client_state.xml. Client writes OS primary platform and alt platforms separately to client_state.xml. The client reads primary platform only to see if it has changed. The client doesn't read alt platforms but some other program might and depends on the platforms being as they are now. Not changing this means the client won't be able to detect when its own platform has changed. OS platform changes are still detected.
- get_state GUI RPC. The client writes primary platform to client_state.platform_name and all platforms, including primary, to client_state.platform. BOINC Manager only reads client_state.platform. As such, client_state.platform_name could be changed to client's build platform but some 3rd party manager might depend on it being OS primary platform.
- time_stats_log. The client writes platform info there. Unclear who uses that information.

Closes #2386.

### DIFF
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -438,7 +438,7 @@ int CLIENT_STATE::init() {
         core_client_version.major,
         core_client_version.minor,
         core_client_version.release,
-        get_primary_platform(),
+        HOSTTYPE,
 #ifdef _DEBUG
         " (DEBUG)"
 #else
--- a/client/cs_cmdline.cpp
+++ b/client/cs_cmdline.cpp
@@ -248,13 +248,7 @@ void CLIENT_STATE::parse_cmdline(int argc, char** argv) {
             if (i == argc-1) show_options = true;
             else safe_strcpy(update_prefs_url, argv[++i]);
         } else if (ARG(version)) {
-#ifdef __APPLE__
-            CLIENT_STATE cs;
-            cs.detect_platforms();
-            printf(BOINC_VERSION_STRING " %s\n", cs.get_primary_platform());
-#else
             printf(BOINC_VERSION_STRING " " HOSTTYPE "\n");
-#endif
             exit(0);
 #ifdef __APPLE__
         // workaround for bug in XCode 4.2: accept but ignore 
--- a/client/http_curl.cpp
+++ b/client/http_curl.cpp
@@ -73,7 +73,7 @@ static void get_user_agent_string() {
     if (g_user_agent_string[0]) return;
     snprintf(g_user_agent_string, sizeof(g_user_agent_string),
         "BOINC client (%s %d.%d.%d)",
-        gstate.get_primary_platform(),
+        HOSTTYPE,
         BOINC_MAJOR_VERSION, BOINC_MINOR_VERSION, BOINC_RELEASE
     );
     if (strlen(gstate.client_brand)) {
--- a/clientgui/mac/config.h
+++ b/clientgui/mac/config.h
@@ -337,8 +337,9 @@
 /* #undef HAVE__PROC_SELF_STAT */
 
 /* Host for this compilation */
-/* "i686-apple-darwin", "x86_64-apple-darwin" or "powerpc-apple-darwin" determined at run time */
-#define HOSTTYPE ""
+#ifdef __x86_64__
+#define HOSTTYPE "x86_64-apple-darwin"
+#endif
 
 /* "Define to 1 if largefile support causes missing symbols in C++" */
 /* #undef LARGEFILE_BREAKS_CXX */


JuhaSointusalo:
Compared to #2388:
- This PR doesn't bother pretty printing HOSTTYPE in startup log. OS and CPU info is already pretty printed a few lines later.
- Uses correct info in two more places.
- Doesn't change VirtualBox message. As for @TheAspens request, lets get rid of the no checking 64-bit VB in 32-bit client limitation in 7.12.

JuhaSointusalo:
Well that's a nice one.
```
▸ Compiling jcapimin.c
** BUILD FAILED **
	CompileC build/boinc.build/Deployment/jpeg.build/Objects-normal/i386/jcapimin.o /Users/travis/build/BOINC/boinc/samples/jpeglib/jcapimin.c normal i386 c com.apple.compilers.llvm.clang.1_0.compiler
(1 failure)
```
Is there any way to see why it failed?

CharlieFenton:
From a visual inspection of your changes, I don't see anything that should affect _jcapimin.c_. I just now synced with GIT master and built the JPEG library on my Mac. There were no warnings or errors building _jcapimin.c_ for either i386 or x86_64 architecture.

JuhaSointusalo:
I restarted the Travis build and it failed the same way.

Originally the code read like this:
```
#ifdef __x86_64__
#define HOSTTYPE "x86_64-apple-darwin"
#else
#error Unknown HOSTTYPE
#endif
```
The idea I had with the `#error` was that if someone ever wants to build BOINC on something other than x86_64 the error points to the right file to fix. On a hunch I removed the error case and now the build succeeds.

Build log made it look like the error was in the middle of `libjpeg` build but `jcapimin.c` is actually the first file built and it failed for i386 build.

CharlieFenton:
> Build log made it look like the error was in the middle of libjpeg build but jcapimin.c is actually the first file built and it failed for i386 build.

_jcapimin.c_ #includes _jpeglib.h_ which #includes _jconfig.h_ which has:
```
#ifdef _WIN32
#include "win-config.h"
#elif ( ! defined(__APPLE__))
#include "config.h"
#endif
```
Since the JPEG library _libjpeg.a_ can be used in project applications, we build it as a universal binary containing both i386 and x86_64 architectures. When Travis-CI attempted the i386 build of the JPEG library, most or all of the source files included config.h, invoking the #error directive.

Moreover, the Xcode build settings for the "boinc' base target has `-include ../clientgui/mac/config.h`, which is then carried through to the settings for all targets. This means that _config.h_ is included in **every** source file, so it is probably a bad idea for _config.h_ to have any architecture-specific statements. Since `HOSTTYPE` is used only in the client, and since _client_state.h_ is #included in every client source file and no others, one option would be to put this code in _client_state.h_ instead of in _config.h_:
```
#ifdef __x86_64__
#define HOSTTYPE "x86_64-apple-darwin"
#else
#error Unknown HOSTTYPE
#endif
```
Another possibility is to take advantage of the fact that BOINC for Mac can be built only on an x86_64 Mac and can be run only on an x86_64 Mac. So this sort of code would also do what you want here:
```
#ifdef __APPLE__
    get_primary_platform(),
#else
    HOSTTYPE,
#endif
```
Finally, note that _CLIENT_STATE::detect_platforms()_ in _client/cs_platforms.cpp_ currently has this code:
```
#elif defined(__APPLE__)

#ifdef __x86_64__
    add_platform("x86_64-apple-darwin");
    add_platform("i686-apple-darwin");
#else
#error Mac client now requires a 64-bit system
#endif

#elif defined(__linux__) && ( defined(__i386__) || defined(__x86_64__) )
```
so  you should also remove this line: `add_platform("i686-apple-darwin");`

CharlieFenton:
Correction: I overlooked the ! in 
```
#ifdef _WIN32
#include "win-config.h"
#elif ( ! defined(__APPLE__))
#include "config.h"
#endif
```
but my explanation for the Travis-CI error and other comments still hold because of the `-include ../clientgui/mac/config.h` in the  the Xcode build settings C and C++ Flags for all targets.

JuhaSointusalo:
> Xcode build settings for the "boinc' base target has `-include ../clientgui/mac/config.h`, which is then carried through to the settings for all targets.

Thanks. I had looked at Xcode project file to see if / how `config.h` was included but couldn't figure it out. The file is text all right but not really human readable.

> This means that config.h is included in every source file, so it is probably a bad idea for config.h to have any architecture-specific statements.

But that is what `config.h` is for. Yes, it is used mostly for header and function related stuff but it's not limited to that. You can for example declare types there that depend on bitness of your program and such. Normally you don't see `#if`s there but that's because it is generated by configure at build time. It's just that Xcode doesn't run configure so Mac `config.h` needs to have the #`if` there.

> So this sort of code would also do what you want here:
> ```
> #ifdef __APPLE__
>     get_primary_platform(),
> #else
>     HOSTTYPE,
> #endif
> ```

Using `get_primary_platform()` got us making these changes now. Maybe BOINC now runs only on x86_64 Mac but that doesn't have to be so forever. What if someone ports BOINC to iOS (lets ignore the roadblocks for a moment)? Newer iOS versions are probably 64-bit but that someone might want to build BOINC as 32-bit app to be able to run on older iOS versions. Then BOINC would report itself as being built for arm64 even though it's arm(32) app. So I have to disagree with you on this one too.

> CLIENT_STATE::detect_platforms() in client/cs_platforms.cpp

As far as I can tell `cs_platforms.cpp` is all about what apps the client can run. I think I may have read somewhere that Apple is dropping 32-bit support from OS X. If that is the case the code should probably check at runtime what the OS can run. I don't have Mac and I don't know enough about OS X to make that kind of change so that's something you have to do.

CharlieFenton:
> This means that config.h is included in every source file, so it is probably a bad idea for config.h to have any architecture-specific statements.

> But that is what config.h is for. Yes, it is used mostly for header and function related stuff but it's not limited to that. 

Right, but my point is that _config.h_ is #included in the library source files, which are built with both i386 and x86_64 architectures, so it can't have any statements which assume only one architecture. But I agree that something like this could be used in _config.h_:
```
#ifdef __x86_64__
#define HOSTTYPE "x86_64-apple-darwin"
#else
#define HOSTTYPE "i386-apple-darwin"
#endif
```
but it is wrong to have this under the `#else`: `#error Unknown HOSTTYPE`.

My other point is that because `HOSTTYPE` is used only by the client, the `#define HOSTTYPE "i386-apple-darwin"` won't actually get used anywhere. Since `HOSTTYPE` is used only by the client, it might be better to define it in a client-only header like _client_state.h_ instead of in _config.h_.

CharlieFenton:
I just had another idea. If you put this in _config.h_, it won't cause any errors in build targets that don't use `HOSTTYPE`, but will cause a _HOSTTYPE undefined_ error in the client  if built with an architecture other than x86_64:
```
#ifdef __x86_64__
#define HOSTTYPE "x86_64-apple-darwin"
#endif
```

CharlieFenton:
> I just had another idea.

Which of course is exactly what you had already done. Sorry to have added so much "noise" to the discussion of this PR while I processed it in my head. Let me know when you are ready to have this merged, and I will merge it.

JuhaSointusalo:
> Right, but my point is that config.h is #included in the library source files, which are built with both i386 and x86_64 architectures, so it can't have any statements which assume only one architecture.

Yeah, when I wrote the first code I was concentrated on the client and forgot `config.h` is used for building i386 libs too.

> Sorry to have added so much "noise" to the discussion of this PR

No harm done. I learned a bit more about Mac build setup.

> Let me know when you are ready to have this merged, and I will merge it.

I think it's ready but since this replaces PR #2388 I don't know if those who commented there should approve or disapprove this PR.

TheAspens:
I have no objections to this PR and deferring the wxWidgets message discussion to 7.12.  It is late in the 7.10 development cycle, so changes for 7.10 should be small and focused which this PR is.

CharlieFenton:
I have merged this into master and cherry-picked it into the 7.10 release branch.

PR #2441, opened 2018-04-01
[WIP] [Manager] Add option 'Ask user to reconnect to remote client'

### LINKED ISSUE #2440: [Manager] Auto-reconnect to remote client without user prompt
- Add new option 'Ask user to reconnect to remote client'
- Option should be checked by default to save old behavior on application upgrade
- When option is off show prompt to user

AenBleidd:
New option 'Ask user to reconnect to remote client' added.
Option is checked by default.
When option is off Manager tries to reconnect without prompting user.

This fixes #2440

### DIFF
--- a/clientgui/BOINCBaseFrame.cpp
+++ b/clientgui/BOINCBaseFrame.cpp
@@ -556,6 +556,9 @@ void CBOINCBaseFrame::ShowConnectionFailedAlert() {
                 return;
             }
         }
+    } else if (!wxGetApp().GetAskUserToReconnectToRemoteClient()) {
+        pDoc->Reconnect();
+        return;
     }
 
     // %s is the application name
--- a/clientgui/BOINCGUIApp.cpp
+++ b/clientgui/BOINCGUIApp.cpp
@@ -111,6 +111,7 @@ bool CBOINCGUIApp::OnInit() {
     m_bSafeMessageBoxDisplayed = 0;
     m_bRunDaemon = true;
     m_bNeedRunDaemon = true;
+    m_bAskUserToReconnectToRemoteClient = true;
 
     // Initialize local variables
     int      iErrorCode = 0;
@@ -171,6 +172,7 @@ bool CBOINCGUIApp::OnInit() {
     m_pConfig->Read(wxT("GUISelection"), &m_iGUISelected, BOINC_SIMPLEGUI);
     m_pConfig->Read(wxT("EventLogOpen"), &bOpenEventLog);
     m_pConfig->Read(wxT("RunDaemon"), &m_bRunDaemon, 1L);
+    m_pConfig->Read(wxT("AskUserToReconnectToRemoteClient"), &m_bAskUserToReconnectToRemoteClient, 1L);
 
     // Detect if the daemon should be launched
     m_bNeedRunDaemon = m_bNeedRunDaemon && m_bRunDaemon;
@@ -574,6 +576,7 @@ void CBOINCGUIApp::SaveState() {
     m_pConfig->Write(wxT("DisplayShutdownConnectedClientDialog"), m_iDisplayShutdownConnectedClientDialog);
     m_pConfig->Write(wxT("DisableAutoStart"), m_iBOINCMGRDisableAutoStart);
     m_pConfig->Write(wxT("RunDaemon"), m_bRunDaemon);
+    m_pConfig->Write(wxT("AskUserToReconnectToRemoteClient"), m_bAskUserToReconnectToRemoteClient);
 }
 
 
--- a/clientgui/BOINCGUIApp.h
+++ b/clientgui/BOINCGUIApp.h
@@ -108,6 +108,7 @@ protected:
     bool                m_bAboutDialogIsOpen;
     bool                m_bRunDaemon;  
     bool                m_bNeedRunDaemon;  
+    bool                m_bAskUserToReconnectToRemoteClient;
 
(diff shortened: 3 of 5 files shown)


AenBleidd:
Guys, I'm sure, this option must have better label, so, please, advice some

TheAspens:
@Ageless93 or @RichardHaselgrove - I'm tied up in the GDPR stuff both with BOINC and WCG - do you have time to build the code in this PR and testing it out and providing an answer to @AenBleidd question about the label?

RichardHaselgrove:
I can build and test on Windows - I run BOINC Manager to remotes regularly. It'll be tomorrow morning, my zone - I'll think about wording when I see what it looks like on screen.

The wording will need to be translatable, which rules it out for 7.10 - much too late to start a new translation cycle - but that gives us time to refine it for v7.12 if needed.

AenBleidd:
@RichardHaselgrove, I don't push this to 7.10 release. It is not an issue but minor feature that will be really helpful for the very small amount of users. So I can say that this PR has very low priority. It is also simple and could be reviewed and verified very quick but anyway.

RichardHaselgrove:
OK, I have a build. Only problem so far is a missing ?, but that can easily be fixed. As expected, we'll need to hook this up to the translation system:

remote connect option

I'll copy it across to my remote monitoring station, and test it there.

RichardHaselgrove:
Verification tests successful. User choice is recorded to, and read back from, local registry. In my typical use case - remote machine needed security patch installing and reboot - manager had reconnected by the time I got back to the monitoring station.

Here are the two English strings we need to assess for legibility, and eventually translate. I'm sure suggestions will be welcome while I think about it.

remote connect wording

CharlieFenton:
In my opinion, the current wording can easily be misunderstood to mean that the manager will never reconnect if the checkbox is off. I suggest something like:
> Ask user before restoring lost connection to remote client?

Another possibility might be to reverse the meaning of the checkbox, and use a label such as:
> Always restore lost connection to remote client without asking?

Perhaps the clearest option is to use radio buttons:
> if connection to remote client is lost:
> O reconnect automatically
> O ask before reconnecting

CharlieFenton:
Might this option be better in the _Select Computer_ dialog rather than the _Other options_ dialog?

Ageless93:
> Only problem so far is a missing ?
That depends, is it a question we ask, or is it an option we give? 
I personally don't feel that when unchecked it will never reconnect, it will just not ask if it needs to reconnect.

AenBleidd:
> Might this option be better in the Select Computer dialog rather than the Other options dialog?

@CharlieFenton, from my POV - definitely no. It does make no sense to put it there because after connecting to client there is no way to change this option

JuhaSointusalo:
If the remote host is up and fully functioning, in particular, any network requests are replied to immediately, but BOINC client is not running on the host, then is Manager with auto-reconnect enabled going to hog (almost full) CPU core due to trying to reconnect non-stop?

(Almost full because network latencies.)

AenBleidd:
@JuhaSointusalo, nice catch! In this case I propose to add some sleep interval that will be increased up to one 1 minute for example in case of unsuccessful reconnect. What do you think?

AenBleidd:
> Interval sounds good but it would have to be more complicated than just sleep(60 seconds). Otherwise you'll get not always responsive Manager.

Yes, I know. I have some options. Will try some of them.

> I'm having a bit trouble deciding what I think of an option to auto reconnect. Could it be a change in behaviour no option given to users? Is there some scenario where auto reconnect is bad?

> The only scenario I can think of is when user mistypes host name. In that case an error message and a prompt to try again would be appropriate. That's kind of a special case of auto reconnect. The connection was never established so it's not really about re-connecting.

Interesting idea. I need to think about it.

JuhaSointusalo:
> While connecting to the client there is already 60 seconds timeout so I see no reason to add additional one.

I don't understand this. With auto-reconnect enabled the Manager seems to keep reconnecting indefinitely. Anyways, I tested the code and the concern I had about Manager hogging CPU core is unfounded. There is some code that isn't visible in the PR that rate-limits the reconnect attempts so much that a continuously reconnecting Manager doesn't use significant CPU time.

To explain my question about whether this needs an option at all. I don't believe that having lots of options for everything possible makes a good software. Eventually you'll reach the point where it takes you half an hour to find an option, or a group of options, you want to change. And another half an hour to test and tweak the options the way you like them. I generally prefer software that automatically does the right thing without asking questions.

In my opinion the right thing in this case would be automatically reconnecting to remote hosts, without having an option to not do it. I'm not demanding that there is no option. By asking the question I was hoping that there would be some discussion about how others see availability of plethora of options but that didn't happen.

Not having an option would quite nicely take care of the problem of coming up with a name for the option :)

CharlieFenton:
>  I don't believe that having lots of options for everything possible makes a good software. 

> In my opinion the right thing in this case would be automatically reconnecting to remote hosts, without having an option to not do it.

I agree. The Manager is already much too complicated. We should be looking for ways to eliminate options rather than adding them.

RichardHaselgrove:
I suppose the question becomes 'can anyone think of a use case where you would want the manager NOT to reconnect to a remote client?'. I can't, and I'm using the opt-in from this PR. I'd support adding the behaviour but leaving out the option control.

AenBleidd:
OK. I'll update this pull request to remove the option and simplify the code.

AenBleidd:
New clean pr #2590 is opened instead of this one.

PR #2443, opened 2018-04-02
WIP:  Google Test Framework for Unit Testing BOINC

Uplinger:
Greetings,

I would like to get some feedback from the community about this approach to doing Unit Testing against the BOINC source code.  

I have documented the design and initial thoughts here:  https://github.com/Uplinger/boinc/wiki/BOINC-Google-Test-Design

The Code in this pull request reflect what is mentioned in the wiki page and gives people a chance to test it out for themselves.  

Thanks,
-Keith Uplinger

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,6 +118,11 @@ locale/*/*.flag
 .libs/
 svn_version.h
 
+## code coverage
+*.gcov
+*.gcno
+*.gcda
+
 # list of executables
 apps/1sec
 apps/concat
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,6 +87,6 @@ script:
 - if [[ "${BOINC_TYPE}" == "manager" && "${TRAVIS_OS_NAME}" == "linux" ]]; then ( ./3rdParty/buildLinuxDependencies.sh && ./configure --disable-server --disable-client --with-wx-prefix=${TRAVIS_BUILD_DIR}/3rdParty/buildCache/linux && make ) fi
 - if [[ "${BOINC_TYPE}" == "manager" && "${TRAVIS_OS_NAME}" == "linux" ]]; then ( make distclean && ./3rdParty/buildLinuxDependencies.sh --disable-webview --cache_dir ${TRAVIS_BUILD_DIR}/3rdParty/buildCache/linux2 && ./configure --disable-server --disable-client --with-wx-prefix=${TRAVIS_BUILD_DIR}/3rdParty/buildCache/linux2 && make ) fi
 - if [[ "${BOINC_TYPE}" == "manager" && "${TRAVIS_OS_NAME}" == "osx" ]]; then ( ./3rdParty/buildMacDependencies.sh -q && ./mac_build/buildMacBOINC-CI.sh --no_shared_headers ) fi
-- if [[ "${BOINC_TYPE}" == "test" ]]; then ( ./3rdParty/buildLinuxDependencies.sh && ./configure --disable-client --disable-manager --enable-coverage && make && coverage.sh ) fi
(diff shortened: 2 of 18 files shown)


Uplinger:
@AenBleidd @brevilo Can you guys take a look at this proposal as I think both of you have shown interest in adding automated testing to the BOINC framework.

brevilo:
Hopefully next week...

AenBleidd:
I'll try to take a look this weekend, Maybe also will prepare some changes to Travis to run these tests.

ChristianBeer:
I briefly looked at the implementation and it seems sound. I also looked into integrating this with travis which seems trivial although you should make the googletest install path configurable so we can put it in the travis cache under `3rdParty/buildCache/linux` treating it as a 3rdParty build dependency.

Getting coverage reports seems to be a bit more difficult as this involves running the compiled boinc binaries in the VM (I think). I haven't figured out how that works. A good service to use would be https://coveralls.io/ which has a github and Travis-CI integration.

ChristianBeer:
I already did part of the above steps in [a (hasty) commit](https://github.com/ChristianBeer/boinc/commit/a29a998638d4a2e607242475626d7b599f68d1ed) and will open a PR to see if it works.

ChristianBeer:
Final note: travis integration is working but needs some final adjustements to keep the cache small. See #2457

ChristianBeer:
I did some more testing and found a way to upload coverage reports to https://codecov.io.

Uplinger:
Christian, thanks for taking a look at this.  I haven't had a chance to view it this weekend. But I will tomorrow. Thanks again.

ChristianBeer:
As it turned out the code coverage part didn't involve additional requirements other than the codecov.io account. Integrating code coverage consists of only two changes. Adding the `--coverage` flag to the BOINC makefiles and the extra script that creates and uploads the coverage reports. I can add this as a follow up.

Converting the fixed Makefiles to autogenerated makefiles is not that difficult I think. The tricky part is to detect the presence of the googletest library and finding the needed include path. If that fails we could place a preconfigured `Makefile` in `gtest` and it's subdirectories and add it to the recursion of the root Makefile. This is currently done in the `samples` directory to build for example `condor` or `example_app`. In the past I tried to get help with the autotools scripts but ended up doing the work on my own through trial and error.

brevilo:
> Integrating code coverage consists of only two changes. Adding the --coverage flag to the BOINC makefiles and the extra script that creates and uploads the coverage reports.

Please ensure that any interaction with 3rd-party services uses an opt-in process. We'd like to be able to build BOINC without 3rd-party involvement or even an internet connection.

Thanks!

brevilo:
I read the proposal and I like it (well, standard unit testing, right?). My only comment from the one above is that you shouldn't make any assumption on where `googletest` lives, if possible. Ideally that would be configurable.

ChristianBeer:
> Please ensure that any interaction with 3rd-party services uses an opt-in process. We'd like to be able to build and test BOINC without 3rd-party involvement or even an internet connection.

The generation and upload of coverage reports only applies to the Travis CI build. Users building and executing the test cases at home will not not be bothered with this. The only requirement is downloading the googletest library to the local system before building.

ChristianBeer:
I'll try to find some time this weekend to look into changing the autotools files to support `./configure --with-gtest=` (similar to what is there for wxWidgets) and probably `make check` or `make test`

ChristianBeer:
I gave it a try and ended up with this branch: https://github.com/BOINC/boinc/commits/cb_gtest_framework which can build, install and detect googletest. It uses some outdated automake scripts that still seem to work.

The last thing to do is write some new `Makefile.am` that do the same as @Uplinger 's `Makefile.gtest` while using the `GTEST_*` variables in `Makefile.incl`. Maybe someone with a better grasp of automake can take a look at this?

What about using cmake for the unit testing stuff and have a kind of hybrid setup?

lfield:
What is that status of this PR?  There has been no progress for nearly one year.

ChristianBeer:
This is the first version of #2457 so it can be closed and the other should be reopened.

PR #2445, opened 2018-04-03
Add a mechanism allowing project admins to "delete" a user

davidpanderson:
This leaves the user record and host records in the DB
to avoid dangling references from the result table.
It:
- changes the user name, email addr, and auth to "deleted_xxx"
- removes the user from team
- removes posts by the user
- removes PMs by the user
- removes friend connection
- removes profile
- "anonymizes" the hosts (erase IP addr and domain name)

Note 1: this is accessed by admins using a script in ops/;
It's not available to users.

Note 2: this may not satisfy the requirements of EU-GDPR
since it doesn't delete the user and host records.

### DIFF
--- a/html/inc/friend.inc
+++ b/html/inc/friend.inc
@@ -115,4 +115,10 @@ function friend_accept_rss($notify, &$title, &$msg, &$url) {
     $url = secure_url_base().USER_HOME;
 }
 
+// delete friendship connections
+//
+function delete_friends($user) {
+    BoincFriend::delete_aux("user_src=$user->id or user_dest=$user->id");
+}
+
 ?>
--- a/html/inc/user.inc
+++ b/html/inc/user.inc
@@ -24,6 +24,8 @@ require_once("../inc/friend.inc");
 require_once("../inc/forum_db.inc");
 require_once("../inc/notify.inc");
 require_once("../inc/ldap.inc");
+require_once("../inc/host.inc");
+require_once("../inc/friend.inc");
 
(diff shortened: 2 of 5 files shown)


TheAspens:
David - do you mind if we hold on reviewing this to see how the discussion in #2447 goes?  Since there might be a lot of overlap between the two it would be good to have one implementation if that will meet the needs that led to this pull request as well as GDPR?

davidpanderson:
ok

TheAspens:
@davidpanderson - as I think through #2447 - what was the driver for this pull request?  I.e. I'd like to understand what are the needs that led to the creation of this pull request so that I can understand what other steps might be needed as I implement the other issue.

RichardHaselgrove:
There was a very strong request by one particular user to be erased from the BOINC-site, and potentially from project sites as well. Obviously I can't identify the user here, but they are known to me personally, and I can brief you privately if you wish.

TheAspens:
I don't need the details - just a general idea which you provided.  It sounds like whatever we implement for GDPR will have satisified that request as well.  Let me know if you agree.

RichardHaselgrove:
I would agree with that. I don't think the user concerned was consciously requesting a dummy run for GDPR, and I don't think that David was consciously operating GDPR, but in effect that's what happened. If you follow the current path of merging your legal advice with David's PR, the final result should have the best of both.

TheAspens:
I'm going to merge this to master and use it in pull request #2472

PR #2446, opened 2018-04-03
web: don't check for stop_web in page_head().

davidpanderson:
Allow projects to show (non-DB) content even if stop_web is present
(e.g. description of the project on front page).
stop_web really means "the DB is offline".


TheAspens:
approved

PR #2448, opened 2018-04-04
Drupal: Fix bug when there is no language, a default flag should appear.

drshawnkwang:
Sets locality to `en` if locality variable does not exist.

https://dev.gridrepublic.org/browse/DBOINCP-433


brevilo:
The Travis issue above is unrelated...

Issue #2451, opened 2018-04-05
Email Change Notification


TheAspens:
In issue #2447 we are discussing that there needs to be tighter controls on how to make it difficult to take over someone else's account.  This describes a change that will provide better security regarding email address changes.

Please see https://boinc.berkeley.edu/trac/wiki/EmailChangeNotification for a description of the design.

Ageless93:
>If the user attempts to visit the change email address screen while user.email_addr_change_time is within 7 days, then they will be informed that they can only change their email address once every 7 days. 

* What if the authentic user looks in here, while their account was taken over? 
* What if the authentic user's old email address was no longer valid, so he didn't get the token, his account was taken over, the nefarious one has made the change of the email address and everything is waiting for the authentic user to click the token email (that he didn't get)? 

I know, in all these years we haven't had account take-overs (that we know of), but people's laptops have been stolen with a running BOINC on board. They have been using BOINC to figure out where their laptop was, due to changing IP addresses. Which is probably where the nefarious one will try to change these things, or delete the account so this info cannot be gotten. 
I also know that the above two things may happen once in a hundred years, if ever. Just asking if we need an option for that, or can they then send a PM to the admin? (If at least the account isn't deleted, with the PM still in there before admin read it. ;-))

drshawnkwang:
Did PR #2500 resolve this issue? Is there a reason why it is still open?

PR #2453, opened 2018-04-06
Token Mechanism

TheAspens:
This pull request updates the method for creating random strings (tokens) to use cryptographically secure functions and it adds a token table and functions for interacting with the table.  This will be used by both issue #2447 and issue #2451

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -762,3 +762,13 @@ create table credit_team (
     credit_type             integer         not null,
     primary key (teamid, appid, credit_type)
 ) engine=InnoDB;
+
+create table token (
+    token                   varchar(255)    not null,
+    userid                  integer         not null,
+    type                    char            not null,
+    create_time             integer         not null default unix_timestamp(),
+    expire_time             integer,
+    primary key (token),
+    index token_userid (userid)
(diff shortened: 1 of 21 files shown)


TheAspens:
The design of this was documented under "Token Generation" and "Token Usage" here: https://boinc.berkeley.edu/trac/wiki/RightToErasure#TokenGeneration  (which is the full design document for issue #2447 )

davidpanderson:
PHP provides a function for making crypto-secure random strings:
http://php.net/manual/en/function.openssl-random-pseudo-bytes.php

TheAspens:
@davidpanderson - do you have any other objections or can you go ahead and merge this?

JuhaSointusalo:
Sorry for being late. You allow `token.expire_time` to be `null`. Was that intentional?

TheAspens:
@JuhaSointusalo - Yes - I wanted to allow for tokens that do not expire.

JuhaSointusalo:
@TheAspens Ok, thanks for explaining. I'm not sure if looking up valid non-expiring tokens works but I suppose that can wait until the more urgent stuff is done.

PR #2459, opened 2018-04-09
[Manager] Fix fields populating on 'Add Project' wizard on launch

### LINKED ISSUE #2444: [Client GUI] Add Project wizard doesn't populate all fields
This with v7.9.3 under Windows 7/64. When the 'Attach project' wizard is first opened, the fields for Research area, Organisation, and Web site are blank.

wizard initial load

After viewing other projects, or changing the category filter, the missing fields are fully populated.

wizard reselect

AenBleidd:
When opening 'Add Project' wizard it shows next fields empty:
Research area, Organisation, and Web site.
This happens because of EllipseStringIfNeeded() function that truncates the string to 0 symbols because the wizard window was not constructed yet and has undefined size.

This fixes #2444.

### DIFF
--- a/clientgui/ProjectInfoPage.cpp
+++ b/clientgui/ProjectInfoPage.cpp
@@ -561,7 +561,7 @@ void CProjectInfoPage::OnProjectSelected( wxCommandEvent& WXUNUSED(event) ) {
         m_pProjectDetailsOrganizationCtrl->SetLabel(strOrganization);
         m_pProjectDetailsOrganizationCtrl->SetToolTip(pProjectInfo->m_strOrganization);
 
-    }
+     }
 
     wxLogTrace(wxT("Function Start/End"), wxT("CProjectInfoPage::OnProjectSelected - Function End"));
 }
@@ -794,7 +794,7 @@ void CProjectInfoPage::EllipseStringIfNeeded(wxString& s, wxWindow *win) {
     wxSize sz = win->GetParent()->GetSize();
     win->GetPosition(&x, &y);
     int maxWidth = sz.GetWidth() - x - 10;
-    
+
     win->GetTextExtent(s, &w, &h);
     
     // Adapted from ellipis code in wxRendererGeneric::DrawHeaderButtonContents()
@@ -813,4 +813,10 @@ void CProjectInfoPage::EllipseStringIfNeeded(wxString& s, wxWindow *win) {
             w += ellipsisWidth;
         }
     }
+}
+
+void CProjectInfoPage::RefreshPage() {
+    // Trigger initial event to populate the list control
+    wxCommandEvent evtEvent(wxEVT_COMMAND_COMBOBOX_SELECTED, ID_CATEGORIES);
+    ProcessEvent(evtEvent);
 }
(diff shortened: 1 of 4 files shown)


AenBleidd:
Guys, could you please review this?

Personally I hate this fix because this is a dirty hack from my POV.
I found no other way to get the size of the window and make proper string truncation.

If you have another idea please feel free to close this and reimplement it in a better way.

AenBleidd:
Reopened to rerun CI

CharlieFenton:
The added space on line 564 of _clientgui/ProjectInfoPage.cpp_ appears to be wrong.
I do not know whether or not your changes to lines 123 and 139-140 in _clientgui/wizardex.cpp_ are correct.

Otherwise it looks good to me. I played with it a bit and did not find a better way to fix this. It is surprising that this has not been reported before, since it has been a bug for quite some time.

AenBleidd:
@CharlieFenton, 

> The added space on line 564 of clientgui/ProjectInfoPage.cpp appears to be wrong.

I agree. I have settings in my diff viewer to skip line ending and space changes.
I'll fix this

JuhaSointusalo:
`wxStaticText` has built-in support for ellipsizing text these days. No idea how well it works, just read it in the docs. `wxHyperlinkCtrl` of course doesn't. You could subclass `wxHyperlinkCtrl` and add ellipsizing. Maybe a lot of work.

wxWidgets' own wizard class has a page shown event. You could hook up to it and fix the labels then. Maybe a bit less of a hack but you'd need to either convert the code to use upstream wizard class or update our copy of it.

Or you could hook up to resize event of some suitable control and fix the labels then. This would add support for resizing the dialog. Not that it is resizeable but, you know, if it were.

Or you could stick to what you have done here.

Shrug.

AenBleidd:
> wxWidgets' own wizard class has a page shown event. You could hook up to it and fix the labels then. Maybe a bit less of a hack but you'd need to either convert the code to use upstream wizard class or update our copy of it.

I tried to do this but with no success. As far as I remember page shown event happened before wizard window shown event and there are still no size of the page.

> Or you could hook up to resize event of some suitable control and fix the labels then. This would add support for resizing the dialog. Not that it is resizeable but, you know, if it were.

I don't want to bind to some another control and it actually will do the same as now.

> wxStaticText has built-in support for ellipsizing text these days. No idea how well it works, just read it in the docs. wxHyperlinkCtrl of course doesn't. You could subclass wxHyperlinkCtrl and add ellipsizing. Maybe a lot of work.

This looks like a very good idea. Will play with it

JuhaSointusalo:
> I tried to do this but with no success. As far as I remember page shown event happened before wizard window shown event and there are still no size of the page.

Well bummer.

TheAspens:
@JuhaSointusalo  or @CharlieFenton  - if you guys are both good with this change can one of you merge it?  @RichardHaselgrove was mentioning he would like to include it in the next client build  for 7.9/10 if possible

AenBleidd:
@TheAspens, I think this can be merged. But in any case I want to play with the solutions suggested by @JuhaSointusalo.

PR #2463, opened 2018-04-10
client: remove /dev/input/ based idle detection

### LINKED ISSUE #2335: [Error] Could not open directory '/dev/input/mice'
Hi,

I got the following error on a headless Ubuntu Server 17.10.1. I installed the `boinc-client` package with the apt.
```
dir_open: Could not open directory '/dev/input/mice' from '/var/lib/boinc-client'.
```
I tried installing the `xorg` and the `xserver-xorg-legacy` packages, but didn't help.

JuhaSointusalo:
The code tries to open /dev/input/mice as if it was a directory. If
there are no input devices, like in a headless machine, the code
repeatedly tries to make a list of input devices. On Debian, these bugs,
together with Debian's patches to BOINC, result in useless error
messages in client's log.

On SELinux systems, services can access device nodes in /dev/input/ only
if given that right in SELinux configuration. On some Fedora versions
BOINC client doesn't have that right and the code results in numerous
SELinux warnings. These warnings in turn result in high CPU usage when
the system analyzes the warnings.

Furthermore, the code doesn't work at all. The code expects that input
drivers update timestamps on device nodes when the drivers receive input
from devices. The input drivers however have no such code. Checking
driver commit logs suggest that if such code ever existed it was so long
time ago that we don't need to care about it and testing distros of
various ages confirms that.

Since the code doesn't and can't work solve the problems on Debian and
Fedora by just removing the code altogether. The code was only used on
Linux and /dev/input/ seems to be Linux only thing anyway.

Note that the client has similar code for checking terminal devices.
Contrary to the input device code that code does work correctly. This is
because programs running on a terminal have their inputs and outputs
connected to the terminal device node. Input to or output from a program
results in a read from or write to device node and terminal drivers do
have code that updates timestamps on reads and writes.

Closes #2335.
See #1187.

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1707,65 +1707,6 @@ inline bool all_tty_idle(time_t t) {
     return true;
 }
 
-static const struct dir_input_dev {
-    const char *dir;
-    const char *dev;
-} input_patterns[] = {
-#ifdef unix
-    { "/dev/input","event" },
-    { "/dev/input","mouse" },
-    { "/dev/input/mice","" },
-#endif
-    // add other ifdefs here as necessary.
-    { NULL, NULL },
-};
-
-vector<string> get_input_list() {
-    // Create a list of all terminal devices on the system.
-    char devname[1024];
-    char fullname[1024];
-    int done,i=0;
-    vector<string> input_list;
-
-    do {
-        DIRREF dev=dir_open(input_patterns[i].dir);
-        if (dev) {
-            do {
-                // get next file
-                done=dir_scan(devname,dev,1024);
-                // does it match our tty pattern? If so, add it to the tty list.
-                if (!done && (strstr(devname,input_patterns[i].dev) == devname)) {
-                    // don't add anything starting with .
-                    if (devname[0] != '.') {
-                        sprintf(fullname,"%s/%s",input_patterns[i].dir,devname);
-                        input_list.push_back(fullname);
-                    }
-                }
-            } while (!done);
-            dir_close(dev);
-        }
-        i++;
-    } while (input_patterns[i].dir != NULL);
-    return input_list;
-}
-
-inline bool all_input_idle(time_t t) {
-    static vector<string> input_list;
-    struct stat sbuf;
-    unsigned int i;
-
-    if (input_list.size()==0) input_list=get_input_list();
-    for (i=0; i<input_list.size(); i++) {
-        // ignore errors
-        if (!stat(input_list[i].c_str(), &sbuf)) {
-            // printf("input: %s %d %d\n",input_list[i].c_str(),sbuf.st_atime,t);
-            if (sbuf.st_atime >= t) {
-                return false;
-            }
-        }
-    }
-    return true;
-}
 #ifdef __APPLE__
 
 // We can't link the client with the AppKit framework because the client
@@ -2202,15 +2143,6 @@ bool HOST_INFO::users_idle(bool check_all_logins, double idle_time_to_run) {
     }
 #endif // HAVE_XSS
 
-    // Lets at least check the dev entries which should be correct for
-    // USB keyboards and mice.  If the linux kernel doc is correct it should
-    // also work for bluetooth input devices as well.
-    //
-    // See: https://www.kernel.org/doc/Documentation/input/input.txt
-    //
-    if (!all_input_idle(idle_time)) {
-        return false;
-    }
 #else
     // We should find out which of the following are actually relevant
     // on which systems (if any)


JuhaSointusalo:
@Germano0 @lfield 

Fedora has allowed BOINC client to access input devices for some time. [boinc.te](https://github.com/fedora-selinux/selinux-policy-contrib/blob/rawhide/boinc.te)

You might want to have that access right removed from the policy at same point, after this change is in a released BOINC version and once you are sure that no-one running an older BOINC version is going to receive a policy update.

You might also want have the change made to the [upstream repo](https://github.com/TresysTechnology/refpolicy-contrib/blob/master/boinc.te) as well. I don't normally run SELinux distros and I don't know much about SELinux anyway so I'd rather not request it myself. I wouldn't see any problems that change might create.

Germano0:
> You might want to have that access right removed from the policy at same point, after this change is in a released BOINC version and once you are sure that no-one running an older BOINC version is going to receive a policy update.
> 
> You might also want have the change made to the upstream repo as well. I don't normally run SELinux distros and I don't know much about SELinux anyway so I'd rather not request it myself. I wouldn't see any problems that change might create.

Thank you @JuhaSointusalo, I will take care of updating both Fedora and upstream SELinux rules

TheAspens:
@davidpanderson or @CharlieFenton  or @lfield - can one of you review this code and confirm it is good to merge?  I'm not familiar enough with this level of linux to make the call.

JuhaSointusalo:
I had checked CentOS 4.8 and 5.11; Debian 5 and 9; Fedora 27; Mandrake 10.1; Mint 10 (=Ubuntu 10.10), 11, 14, 16 and 17 (=Ubuntu 14.04) and Scientific Linux 6.3. The oldest kernel was 2.6.8 from around 2004.

I didn't think to check Android. Android has more security features enabled than average desktop Linux so I would be surprised if the code worked on Android even if the Android/Linux kernel somehow otherwise was updating `/dev/input` timestamps.

I also checked FreeBSD 11.1; NetBSD 7.1.2; OpenBSD 6.3 and Solaris 11.3. None had `/dev/input`.

I didn't check AIX or HP-UX because they don't run on x86. Haiku is open-source clone of BeOS which I think was it's own kind of OS and probably didn't have `/dev` at all. DragonflyBSD is a fork of some other BSD and probably doesn't have `/dev/input` either. There is some OS that runs on Raspberry PI, maybe called RiscOS or something. I didn't check it and don't really know if BOINC even runs on it. I forgot about OS/2 / eComStation and I don't know if it's still alive or freely available.

Anything else?

The interesting kernel files are [`evdev.c`](https://github.com/torvalds/linux/blob/master/drivers/input/evdev.c), [`input.c`](https://github.com/torvalds/linux/blob/master/drivers/input/input.c) and [`mousedev.c`](https://github.com/torvalds/linux/blob/master/drivers/input/mousedev.c) and for comparison [`tty_io.c`](https://github.com/torvalds/linux/blob/master/drivers/tty/tty_io.c) which does update timestamps on PTY/TTY device nodes. The kernel history on GitHub goes back to 2005 and 2.6.12-rc2 which is IMHO old enough.

CharlieFenton:
I also am not familiar with this level of Linux.

Germano0:
> The interesting kernel files are evdev.c, input.c and mousedev.c and for comparison tty_io.c which does update timestamps on PTY/TTY device nodes. The kernel history on GitHub goes back to 2005 and 2.6.12-rc2 which is IMHO old enough.

IMHO the best solution for Linux idle detection is using logind features
https://github.com/BOINC/boinc/issues/1187#issuecomment-225699768

JuhaSointusalo:
One way to look at this is to run `stat /dev/input/*` (which is basically what the client does), note the times, type something on keyboard and move mouse and use other input devices, wait a few moments, run the command again and observe that the times have not changed. Repeat the test on a selection of different distros and come to the conclusion that the code doesn't work at all.

The rest is just that I wasn't satisfied with "doesn't work at all" and needed to explain to myself _why_ the code doesn't work when the similar code for TTYs does work.

TheAspens:
@JuhaSointusalo - can you let me know the status of this?  Is there anything pending that you need to do and if it is ready - who do you think the most suitable reviewers are?  Thanks!

JuhaSointusalo:
> can you let me know the status of this

It's ready.

> who do you think the most suitable reviewers are

Anyone who can check that the code doesn't work anywhere.

TheAspens:
@lfield - would you be willing to review and merge if you are ok with it?  I don't know linux at this level well enough to feel comfortable doing the review and merge.

lfield:
I am fine with this change. We can only test on a limited set of machines this will be merged for wider testing.

RichardHaselgrove:
This is an example of the sort of change which should be tested from master, long before the next release branch is forked.

Otherwise, we fall foul of "Too early to test, too late to change".

TheAspens:
@lfield - did you mean to merge the pull request or close it?

PR #2471, opened 2018-04-17
Win installer: update project list

davidpanderson:


CharlieFenton:
This PR also updates the project list for the Mac installer. The Mac installer incorporates this same file, so the project lists installed on Macintosh and Windows automatically stay in sync. (The Mac installer installs the project list only if there is not already one in the BOINC data directory, on the assumption that the one in the BOINC data directory is probably newer since the client updates it every 2 weeks.)

TheAspens:
This pull request has been approved by the PMC

PR #2472, opened 2018-04-17
Delete Account

TheAspens:
This pull request resolves issue #2447 

This will be a work in progress for a bit, so please don't merge.  However, since it is a large change I wanted to push incremental updates for review if people want to look.


Ageless93:
The one thing I notice in all the code is the blurb at the top that says:
>// This file is part of BOINC.
>+// http://boinc.berkeley.edu
>+// Copyright (C) 2018 University of California

It would be nicer if the URL was immediately made https, so that you don't have to do that again in the future.

TheAspens:
@Ageless93 - thanks for the feedback.  I've updated them to https.

html/inc/token.inc
```diff
@@ -0,0 +1,45 @@
...
+require_once("../inc/boinc_db.inc");
+require_once("../inc/util.inc");
+
+// Constants for valid token types
+define("TOKEN_TYPE_DELETE_ACCOUNT", "D");
+
+// Constants for token durations
+define("TOKEN_DURATION_ONE_DAY", "86400");
+
+function create_confirm_delete_account_token($user) {
```
Uplinger: Can this function and the is_valid_delete_account_token be generalized.  I think having them generalized will help decrease the amount of code as well as the maintenance of it going forward.  Having functions that call say create_token($userid, $type, $expire_time = null).  This way anywhere that needs to use a new token can easily call it.  At the moment we have 2 types needed, one for delete account and email change, but this can be used for many more types going forward.  

is_valid_token($userid, $token, $type) is for the other generalized function

TheAspens:
@davidpanderson - left you a message.  Give me a ring or I'll call you again tomorrow.

TheAspens:
@brevilo - the field names have are back to userid and hostid

brevilo:
Thanks Kevin.

brevilo:
@davidpanderson I presume you intended to comment on #2500 instead?

davidpanderson:
yes

TheAspens:
@brevilo @davidpanderson @tristanolive @drshawnkwang - if people are able to start reviewing, I would appreciate it.  Other than incorporating the code from the #2500 and adding logic to prevent deleting your account within 7 days of an email change, this code should be finished.  I have some additional testing to do tomorrow, but I appreciate a feedback at this time.  Thanks!

drshawnkwang:
@TheAspens , apologies for not getting to this sooner:

1) I tried testing your branch on a new test project (basically empty) and deletion worked as expected.

2) I merged your code with the Drupal code I am working on to use the functions in `inc/account_delete.inc`. Calling the delete function `wipe_account()` works. Thus with my Drupal developer hat on I believe I can use your code to delete the boinc account on our system, and then clean up the Drupal-account afterwards.

I will try to perform more tests over the next few days and report any relevant problems I find here.

TheAspens:
This is complete.  Please review and merge.

html/inc/boinc_db.inc
```diff
@@ -823,6 +835,50 @@ class BoincToken {
...
+        $now = time();
+        $cpid = md5($user->cross_project_id.$user->email_addr);
+        $clause = "(userid, public_cross_project_id, create_time) values ($user->id, '$cpid', $now)";
+        $db = BoincDb::get();
+        return $db->insert('user_deleted', $clause);
+    }
+    
+    static function delete_expired() {
+        $db = BoincDb::get();
+        $expire_time = time() - 2*30*86400; //60 days ago
```
brevilo: No sure why 60 needs to be expressed as 2*30 🤔 (same for host below)

html/inc/boinc_db.inc
```diff
@@ -303,6 +303,10 @@ class BoincTeamDelta {
         $db = BoincDb::get();
         return $db->enum('team_delta', 'BoincTeamDelta', $where_clause);
     }
+    static function delete_for_user($user_id) {
+        $db = BoincDb::get();
+        return $db->delete_aux('team_delta', "userid=$user_id");
```
brevilo: There should be an index on `team_delta.user_id` to make this query efficient.

html/inc/boinc_db.inc
```diff
@@ -823,6 +835,50 @@ class BoincToken {
         return $db->affected_rows();
     }
 
+    static function delete_for_user($user_id) {
+        $db = BoincDb::get();
+        $now = time();
+        $db->delete_aux('token', "userid=$user_id");
```
brevilo: There should be an index on `token.user_id` to make this query efficient.

html/inc/delete_account.inc
```diff
@@ -0,0 +1,187 @@
...
+// Constants for different methods of deleting accounts
+// These correspond to the value used in the config.xml 
+// field of <enable_delete_account/>
+define("DELETE_ACCOUNT_METHOD_OBFUSCATE", 1);
+define("DELETE_ACCOUNT_METHOD_WIPE", 2);
+define("DELETE_ACCOUNT_METHOD_PROJECT_DEFINED", 3);
+
+$config = get_config();
+if ( !parse_bool($config, "enable_delete_account") ) {
+    error_page(
```
brevilo: @drshawnkwang how does Drupal handle calls to `error_page()`? Is this problematic since it's part of an `inc` file (API) but sounds like a front-end feature?

html/inc/delete_account.inc
```diff
@@ -0,0 +1,187 @@
...
+        return project_delete_account($user);
+    } else {
+        error_page(
+            tra("This feature is configured incorrectly.  Please contact the project administrator.")
+        );
+    }
+}
+
+// "obfuscate" an account: leave user record (for DB consistency) but:
+//  - set email address and authenticator to "deleted_pid_time"
```
brevilo: The comment doesn't reflect the implementation (anymore?): it's `deleted_time_random_string`

html/inc/delete_account.inc
```diff
@@ -0,0 +1,187 @@
...
+    // remove user's team records
+    user_erase_team_owner($user);      // from team.inc
+    user_quit_team($user);        // from team.inc
+    user_erase_team_delta($user); // from team.inc
+    
+    // Items that do not have logic elsewhere
+    // and do not have objects in boinc_db.inc
+    if (!$db) die("no DB connection");
+    $db->do_query("delete from credited_job where userid = $user->id");
+    $db->do_query("delete from donation_paypal where userid = $user->id");
```
brevilo: There should be an index on `donation_paypal.userid` to make this query efficient.

html/inc/delete_account.inc
```diff
@@ -0,0 +1,187 @@
...
+    $db->do_query("delete from banishment_vote where userid = $user->id");
```
brevilo: There should be an index on `banishment_vote.userid` to make this query efficient.

html/inc/delete_account.inc
```diff
@@ -0,0 +1,187 @@
...
+    $db->do_query("delete from post_ratings where post in ( select id from post where user = $user->id )");
+    $db->do_query("delete from post_ratings where user = $user->id");
```
brevilo: There should be an index on `post_ratings.user` to make this query efficient.

html/inc/delete_account.inc
```diff
@@ -0,0 +1,187 @@
...
+    $db->do_query("delete from msg_from_host where hostid in (select id from host where userid = $user->id )");
```
brevilo: There should be an index on `msg_from_host.hostid` to make this query efficient.

html/inc/delete_account.inc
```diff
@@ -0,0 +1,187 @@
...
+    $db->do_query("delete from msg_to_host where hostid in (select id from host where userid = $user->id )");
```
brevilo: There should be an index on `msg_to_host.hostid` to make this query efficient.

html/inc/delete_account.inc
```diff
@@ -0,0 +1,187 @@
...
+    $db->do_query("delete from sent_email where userid = $user->id");
```
brevilo: There should be an index on `sent_email.userid` to make this query efficient.

html/user/delete_account_request.php
```diff
@@ -0,0 +1,67 @@
...
+
+require_once("../inc/util.inc");
+require_once("../inc/account.inc");
+require_once("../inc/delete_account.inc");
+require_once("../inc/user_util.inc");
+require_once("../inc/email.inc");
+
+$user = get_logged_in_user();
+
+if ($user->email_addr_change_time + 604800 > time()) {
```
brevilo: Here readability would be improved by using `7*86400` instead 💡

sched/db_dump.cpp
```diff
@@ -765,12 +796,38 @@ int ENUMERATION::make_it_happen(char* output_dir) {
             exit(retval);
         }
         break;
+    case TABLE_USER_DELETED:
+        n = 0;
+        while (1) {
+            retval = user_deleted.enumerate("order by userid", true);
```
brevilo: Why do you use `mysql_use_result()` here (via `enumerate(clause, true)`) but `mysql_store_result()` [below](a1147ebff483de86a4970ff9492cbb91R854)?

brevilo:
Static code review completed.

html/inc/delete_account.inc
drshawnkwang: Oliver is correct, this is not ideal from the Drupal standpoint. Drupal will call the `error_page()` function and display an error page, that is devoid of anything but the error message: think black text on white background.

Ideally, the config.xml check should happen elsewhere, either before the functions here are called. Or at the very beginning of each function, and then it returns false.

I don't know if that would require a major re-write.

bema-ligo:
I'm worried what happens if while this code runs, the scheduler, following a request from a host the user has forgotten about, creates a new host or modifies result properties. Thus prior to that deletion process I would invalidate the user's authenticator and wait long enough that all currently running schedulers have finished (with our CGI setting this is 2 Min).

html/inc/delete_account.inc
TheAspens: Not required.  The index msg_to_host(hostid, handled) already exists.  Since hostid is the first term in the index, it will be used efficiently in this query.

TheAspens:
@bema-ligo - since this is being done under the scope of a website request, the user experience as well as the possibility of the request timing out would be problematic.  However, to reduce the chance (of something that is already a very small chance), I will take the step to invalidate the authenticator and add a short delay of 2 seconds.  That is more time than most of our scheduler requests take and this would make the scenario you describe vanishingly unlikely.

html/inc/boinc_db.inc
```diff
@@ -823,6 +835,50 @@ class BoincToken {
...
+        $cpid = md5($user->cross_project_id.$user->email_addr);
+        $clause = "(userid, public_cross_project_id, create_time) values ($user->id, '$cpid', $now)";
+        $db = BoincDb::get();
+        return $db->insert('user_deleted', $clause);
+    }
+    
+    static function delete_expired() {
+        $db = BoincDb::get();
+        $expire_time = time() - 2*30*86400; //60 days ago
+        $db->delete_aux('user_deleted', "create_time < $expire_time");
```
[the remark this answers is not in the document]
TheAspens: added

html/inc/boinc_db.inc
```diff
@@ -823,6 +835,50 @@ class BoincToken {
...
+        $now = time();
+        $clause = "select id, host_cpid, $now from host where userid = $user->id";
+        $db = BoincDb::get();
+        return $db->insert('host_deleted', $clause);
+    }
+    
+    static function delete_expired() {
+        $db = BoincDb::get();
+        $expire_time = time() - 2*30*86400; //60 days ago
+        $db->delete_aux('host_deleted', "create_time < $expire_time");
```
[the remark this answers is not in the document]
TheAspens: added

html/inc/boinc_db.inc
TheAspens: changed

html/inc/boinc_db.inc
TheAspens: added

html/inc/delete_account.inc
TheAspens: changed

html/inc/delete_account.inc
TheAspens: added

html/inc/delete_account.inc
TheAspens: added

html/inc/delete_account.inc
TheAspens: added

html/inc/delete_account.inc
TheAspens: added

html/inc/delete_account.inc
TheAspens: added

sched/db_dump.cpp
TheAspens: and it is now changed

html/inc/delete_account.inc
TheAspens: I've re-organized the code so calls to error_page only reside in view components.

TheAspens:
I've responded to all items.  Please re-review and update your review as needed.  Thanks.

TheAspens:
@davidpanderson - if you could review also and if both you and @brevilo  are satisfied, then please merge as well.  Thanks!

davidpanderson:
Fine with me.  Oliver, merge if OK with you.

html/inc/delete_account.inc
brevilo: Agreed. I missed that one entirely.

sched/db_dump.cpp
brevilo: As long as it's an informed decision I'm fine with either.

html/inc/boinc_db.inc
brevilo: Thanks

brevilo:
> However, to reduce the chance (of something that is already a very small chance), I will take the step to invalidate the authenticator and add a short delay of 2 seconds. That is more time than most of our scheduler requests take and this would make the scenario you describe vanishingly unlikely.

@bema-ligo are you happy with this or does it (still) mean a no-merge for you?

TheAspens:
> While it almost feels like nitpicking, I'm not really happy with the abbreviated index names. I mean, they're almost never to be typed anyway so I don't see the justification to reduce their readability (contextual meaning) in favor of length. Opinions?

I've made the names long.  However, many indexes previously had shortened names so I had been following that convention.

TheAspens:
@bema-ligo and @brevilo,
I've added commit: https://github.com/BOINC/boinc/pull/2472/commits/ae7200490aa567f1781a27fe842ff90fb4258126 which make the length of the delay something that projects can configure in the project.inc file.  Since the timeouts and reasonable values depend on different conditions on different projects, this allows projects to choose a setting that works for them with a default that will work properly in the majority of cases.

brevilo:
Thanks Kevin, sounds good to me. I'll merge as soon as I here from @bema-ligo .

brevilo:
How about your review @Uplinger? Are you satisfied? If so, please acknowledge/accept the review changes.

Ideally @davidpanderson would do the same such that all reviews are visibly done (upper right).

brevilo:
Duh!

TheAspens:
I've completed my regression testing.  Once I review the scrutinizer feedback it should be good for merge.

TheAspens:
I've incorproated the scruitinizer feedback as pertains to my code.  Note that it has a false bug report that I've submitted to them for review.

TheAspens:
I'm done with the code.  Please merge when you are satisfied.

brevilo:
> Sorry for pushing, but since WCG is not up to current code, we have some work to do to integrate this into our code

Yep, same here. I reckon we can still optimize things later should Bernd have further input. So let's go...

brevilo:
Thanks Kevin for all the effort you put into this! Thanks also to all the others involved in various ways 👍

bema-ligo:
Sorry for the delay at my end. FTR: I'm fine with the solution (configurable time).

Ageless93:
Since this was merged 6 days ago, can it be tested someplace for real? I haven't seen this code be available yet at for instance Seti@Home. Not that I want to test it there, but wouldn't mind testing it at a test project. The 25th is less than two days away.

brevilo:
We already deployed a custom solution at Einstein@Home as well as our test instance Albert@Home. However, we're currently wait for #2524 to be merged which will then make use of the features added by this PR.

PR #2473, opened 2018-04-18
MySQL doesn't allow integers to have default values of functions

RadicalErin:
create_time is an integer with a default value of unix_timestamp(). MySQL doesn't allow for this, and throws syntax errors during the make_project phase. This update gives the same result in a MySQL friendly fashion.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -767,8 +767,15 @@ create table token (
     token                   varchar(255)    not null,
     userid                  integer         not null,
     type                    char            not null,
-    create_time             integer         not null default unix_timestamp(),
+    create_time             integer         default null,
     expire_time             integer,
     primary key (token),
     index token_userid (userid)
 ) engine=InnoDB;
+
+create trigger trig_token_creation_timestamp_default before insert on token for each row
+begin
+    if (new.create_time is null) then
+        set new.create_time = unix_timestamp();
+    end if;
+end$$
--- a/html/ops/db_update.php
+++ b/html/ops/db_update.php
@@ -1087,14 +1087,6 @@ function update_4_5_2018() {
             index token_userid (userid)
         ) engine=InnoDB;
     ");
-    do_query("
-        create trigger trig_token_creation_timestamp_default before insert on token for each row
-        begin
-            if (new.create_time is null) then
-                set new.create_time = unix_timestamp();
-            end if;
-        end$$
-    ");
 }
 
 function update_4_6_2018() {
@@ -1102,6 +1094,21 @@ function update_4_6_2018() {
         modify column total_credit double not null default 0.0,
         modify column expavg_credit double not null default 0.0,
         modify column seti_id integer not null default 0
+    ");
+}
+
+function update_4_18_2018() {
+    do_query("
+        alter table token
+        modify create_time integer default null
+    ");
+    do_query("
+        create trigger trig_token_creation_timestamp_default before insert on token for each row
+        begin
+            if (new.create_time is null) then
+                set new.create_time = unix_timestamp();
+            end if;
+        end;
     ");
 }
 
@@ -1160,6 +1167,7 @@ $db_updates = array (
     array(27021, "update_3_8_2018"),
     array(27022, "update_4_5_2018"),
     array(27023, "update_4_6_2018"),
+    array(27024, "update_4_18_2018"),
 );
 
 ?>
--- a/html/ops/db_update.php
+++ b/html/ops/db_update.php
@@ -1076,15 +1076,24 @@ function update_3_8_2018() {
 }
 
 function update_4_5_2018() {
-    do_query("create table token (
-        token                   varchar(255)    not null,
-        userid                  integer         not null,
-        type                    char            not null,
-        create_time             integer         not null default unix_timestamp(),
-        expire_time             integer,
-        primary key (token),
-        index token_userid (userid)
-        ) engine=InnoDB
+    do_query("
+        create table token (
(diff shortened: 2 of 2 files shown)


TheAspens:
Thanks for catching and for the fix.  I had tested on MariaDB 10.2 which does allow this so I didn't realize this would break MySQL - so my apologies for the break.

For those following the discussion here is the documentation for MySQL: https://dev.mysql.com/doc/refman/5.7/en/data-type-defaults.html which clearly documents was @RadicalErin mentions.

The script at html/ops/db_update.php needs to be fixed as well.  In particular the "update_4_5_2018" where the table is defined for those updating any existing deployment.  Do you want to add that change to your pull request?  Then I can test and merge.

Thanks for the contribution!

RadicalErin:
Updated to include db_update. You do raise a fair point, though. I have verified this in MySQL, but I have not checked against other database systems. I'm not sure what the compatibility requirements are on this project with respect to databases.

RadicalErin:
As an aside, does this require another update as well, for anyone up to current?

TheAspens:
At the moment I think we need to be compatible with MySQL 5.6/6.7 and MariaDB 10.1/10.2.

TheAspens:
> As an aside, does this require another update as well, for anyone up to current?

Good question.  I'm thinking.  If the update failed, then pull this fix will simply allow the update to succeed and they should be fine. 

I got an email from someone else indicating that they were on MariaDB 10.1 and also got his failure.  I'll take a look to see if only MariaDB 10.2 allows this so only those using 10.2 might be impacted and those would still get a default value but via a different mechanism.  The  issue would only come up if sometime in the future we removed this default value.  

However, we could avoid that if you added another update to db_update that redefined that column to not have a default and also redefined the trigger, then that would cover that case as well.  If you don't mind that would probably be a good idea.

TheAspens:
Actually - thinking slightly longer.  Change update_4_5_2018 to simple define the table without the default.

Then move the trigger definition to a new add update_4_18_2018.  update_4_18_2018 should also redefine the column on the table to not have the default.  That should get everyone to the same spot.

drshawnkwang:
I'll chime in just to say I am using MariaDB 10.0.32 w/ PHP v5.6 and received what seems to be a similar problem when upgrading the project database (db_update.php)

RadicalErin:
I've identified a second syntax error that occurs during the make_project process. Unfortunately, it's a bit hard to track down.
`Traceback (most recent call last):
  File "./make_project", line 234, in <module>
    project.install_project()
  File "/var/www/server/BOINC/py/Boinc/setup_project.py", line 594, in install_project
    drop_first = options.drop_db_first
  File "/var/www/server/BOINC/py/Boinc/database.py", line 290, in create_database
    _execute_sql_script(cursor, os.path.join(srcdir, 'db', file))
  File "/var/www/server/BOINC/py/Boinc/database.py", line 277, in _execute_sql_script
    cursor.execute(query)
  File "/usr/lib/python2.7/dist-packages/MySQLdb/cursors.py", line 174, in execute
    self.errorhandler(self, exc, value)
  File "/usr/lib/python2.7/dist-packages/MySQLdb/connections.py", line 36, in defaulterrorhandler
    raise errorclass, errorvalue
_mysql_exceptions.ProgrammingError: (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 4")
`
I don't think it's caused by this change, so much as revealed by it. I'll have to find some time to hunt that down [unless it happens to pop out to someone]. I can't necessarily say that this is ready to go, given that I can't fully build it like this.

TheAspens:
I was testing and I think it is because of the following:

```
DELIMITER $$
create trigger trig_token_creation_timestamp_default 
before insert on token 
for each row
begin
    if (new.create_time is null) then
        set new.create_time = unix_timestamp();
    end if;
end$$
DELIMITER ;
```

without changing the delimiter mysql gets confused.

TheAspens:
Here is confirmation that MariaDB 10.2.1 added the ability to use functions: https://mariadb.com/kb/en/library/create-table/#default

TheAspens:
@RadicalErin - I was talking to @davidpanderson and he had some concerns about solving this by using triggers vs moving the assignment to create_time into code.  I was also know that a number of projects utilize replication and triggers can sometimes introduce issues into replication  (I think we are in the clear here, but if statement based replication was used the value for create time would be slightly different on the replica then on the master).  

I have some other changes for tokens that I need to pull from #2472 and make available before 2472 is going to be complete.   As a result, I'm going to open a pull request shortly that will fix the issue that this pull request was working to address as well as make available the other token changes that I was working on.   Also - it is my bug so I feel some responsibility to get it fixed.

Thank you for be willing to submit a pull request to fix this bug though.  We are happy to have you involved!

RadicalErin:
https://dev.mysql.com/doc/mysql-replication-excerpt/5.7/en/replication-features-triggers.html
> With statement-based replication, triggers executed on the master also execute on the slave. With row-based replication, triggers executed on the master do not execute on the slave. Instead, the row changes on the master resulting from trigger execution are replicated and applied on the slave.

I would suggest both. I think you're safe with triggers, but it's better if they're a fallback option. Inserting that value, instead, would be preferable. It's just safer, from a DB perspective, to have that trigger as a fallback.

TheAspens:
@RadicalErin - would you mind testing #2474 to see if it fixes the bug on your systems?

RadicalErin:
Sounds like we're good, here. I'll close.

PR #2474, opened 2018-04-18
WIP: web: Fix bug creating token table with MySQL and MariaDB < 10.2

TheAspens:
This pull requests fixes a bug in creating the token table with MySQL and MariaDB earlier than version 10.2 that was introduced in pull request #2453 

It also includes code for how tokens will be used (for reference by @Uplinger and @drshawnkwang  who are already looking to use this code) from the work I'm doing in pull request #2472.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -767,7 +767,7 @@ create table token (
     token                   varchar(255)    not null,
     userid                  integer         not null,
     type                    char            not null,
-    create_time             integer         not null default unix_timestamp(),
+    create_time             integer         not null,
     expire_time             integer,
     primary key (token),
     index token_userid (userid)
--- a/html/inc/boinc_db.inc
+++ b/html/inc/boinc_db.inc
@@ -783,10 +783,11 @@ class BoincToken {
         return $db->lookup('token', 'BoincToken', $clause);
     }
     
-    static function lookup_valid_token($userid, $token) {
+    static function lookup_valid_token($userid, $token, $type) {
         $db = BoincDb::get();
         $token = BoincDb::escape_string($token);
-        return self::lookup("userid=$userid and token='$token' and expire_time > unix_timestamp()");
+        $type = BoincDb::escape_string($type);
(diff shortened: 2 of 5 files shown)


TheAspens:
Note that this is currently a WIP while I finish testing.

TheAspens:
@drshawnkwang - can you test to see if this resolves your issue?

TheAspens:
I've tested on MariaDB 10.2 and MySQL 5.7.

TheAspens:
I've removed the WIP status.  I would appreciate it if @drshawnkwang and @RadicalErin could test and validate that this resolves the problem on their systems.

RadicalErin:
That would appear to have allowed the make_project process to complete. I can't confirm or deny functionality after that point, yet, though.

drshawnkwang:
I applied the `tools/upgrade` script, which runs `db_update.php`, to upgrade my (test) projects. The database update, specifically `update_4_5_2018` worked, where as it did not before. 

I haven't tested further than this, but did get farther than I did before. I'll try going about using the token code itself tomorrow and Friday.

@TheAspens, Thanks for the fast work on this.

TheAspens:
@davidpanderson - I've made the change you requested.

I personally prefer to use time on the database server as it ensures that all clients connecting to it have the same frame of reference regardless of what the time on the client is set to.  Ideally all servers involved are properly syncing with NTP so that this difference is unimportant, but I've seen situations where it did matter.

Regardless since I assume the BOINC standard is what you indicate above, I've updated the code to comply with it.

TheAspens:
I made the changes requested by @davidpanderson and @Uplinger.  

If everyone is satisfied with the change could someone please merge since this fixes a broken issue in the master branch?

TheAspens:
@davidpanderson, @tristanolive, @drshawnkwang or anyone else who feels comfortable reviewing.  I've addressed all questions on this code and this fix resolves a blocking issue on the master branch.  Could someone please complete the review and merge the code?

drshawnkwang:
FYI, I don't have merge permissions, but I do support this PR!

PR #2480, opened 2018-04-24
Drupal: User Opt-in consent and Web registration

drshawnkwang:
Two major changes:

1. Opt-in consent - use must opt-in to a terms of use when creating a new account.
  * Additionally, existing users may continue to use the site after logging in, unless they too consent to the terms of use.
  * The following may be disabled by simply removing the terms-of-use: if the variable is empty there is no consent.
  * For now, we use the `drupal` database, `boincuser` table to record the initial consent. This will hopefully move to a BOINC upstream database table in the near future.
2. Web registration - admin has the ability to enable/disable creating new accounts using the create_account RPC and the Web registration form. These may be independently toggled.

https://dev.gridrepublic.org/browse/DBOINCP-428

### DIFF
--- a/drupal/sites/default/boinc/modules/boincuser/boincuser.module
+++ b/drupal/sites/default/boinc/modules/boincuser/boincuser.module
@@ -130,6 +130,14 @@ function boincuser_menu() {
     'access arguments' => array('access content'),
     'type' => MENU_CALLBACK,
   );
+  $items['user/optin'] = array(
+    'title' => bts('Opt In to Terms of Use', array(), NULL, 'boinc:opt-in-form'),
+    'description' => 'Opt-in to site\'s term of use.',
+    'page callback' => 'drupal_get_form',
+    'page arguments' => array('boincuser_optinform'),
+    'access callback' => 'user_is_logged_in',
+    'type' => MENU_CALLBACK,
+  );
   $items['user_control'] = array(
     'page callback' => 'boincuser_control',
     'access arguments' => array('access user profiles'),
@@ -199,6 +207,7 @@ function boincuser_menu() {
  *
  */
 function boincuser_init() {
+  global $user;
   // Skip this check for charts, which are loaded separately
   // (may get duplicate or unexpected messages otherwise)
   if (substr($_GET['q'], 0, 7) == 'charts/') {
@@ -218,6 +227,36 @@ function boincuser_init() {
   if (module_exists('boincteam')) {
     // Display any persistent team messages
     boincteam_show_messages();
+  }
+
+  // Check if user has opt-in to the terms of use. If not, send the
+  // user to the terms-of-use form. This is only makes sense if the
+  // termsofuse is enabled, by having text in the termsofuse variable.
+  $termsofuse = variable_get('boinc_weboptions_termsofuse', '');
+  if ( (!empty($termsofuse)) and ($user->uid) ) {
+    if (!boincuser_check_termsofuse($user)) {
+
+      // Admins are exempt, otherwise the admin may not be able to
+      // access the site!
+      $administrator_role = array_search('administrator', user_roles(true));
+      if (!isset($user->roles[$administrator_role])) {
+        $path = drupal_get_path_alias($_GET['q']);
+
+        // Any paths that should NOT be redirected go here.
+        $paths_to_ignore = array(
+          'user/optin',
+          'logout',
+        );
+        if (module_exists('boincuser_delete')) {
+          $paths_to_ignore[] = 'user/' . $user->uid . '/delete';
+          // @todo - once the following path has been defined: add to this array
+          //$paths_to_ignore[] = 'user/' . $user->uid . '/deleteconfirm',
+        }
+        if (!in_array($path, $paths_to_ignore)) {
+          drupal_goto('user/optin');
+        }
+      }
+    }
   }
 }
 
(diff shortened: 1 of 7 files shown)


brevilo:
I presume it should read "existing users may **not** continue to use the site", right?

drupal/sites/default/boinc/modules/boincuser/boincuser.module
```diff
@@ -1546,9 +1546,18 @@ function boincuser_create_account() {
...
         // Cross reference Drupal account with BOINC.
         $reference = db_query("INSERT INTO {boincuser} SET uid=%d, boinc_id=%d", $user->uid, $boinc_user->id);
+
         if (!$reference) {
           xml_error(-1, 'error connecting BOINC account to Drupal');
         }
+
+        // if terms of use exist, the user must have opted-in.
+        // @todo - test this block of code!
+        $termsofuse = variable_get('boinc_weboptions_termsofuse', '');
```
brevilo:
~~~This seems to be a Drupal-specific RPC customization. If I'm right, what's the point of it as long as BOINC's stock RPC isn't changed the same way. We lose API/RPC coherence for the service consumers like the BOINC client (or account managers) that would have to be updated as well to make use of that new parameter.~~~

See below, I think misinterpreted the variable and the comment above it. Please confirm: this means when this RPC gets called **and** the terms of use exist, there's no way the user can't have consented - so let's store that fact in the DB.

drupal/sites/default/boinc/modules/boincwork/boincwork.module
```diff
@@ -1041,7 +1041,133 @@ function boincwork_get_output() {
  * Get the project configuration XML; used by client software
  */
 function boincwork_get_project_config() {
-  include_boinc('user/get_project_config.php');
+  require_boinc( array('boinc_db', 'util', 'xml') );
```
brevilo: I presume there's no alternative to reimplement the whole RPC? Doing this means we always need to keep it in sync with BOINC's stock RPC.

drupal/sites/default/boinc/modules/boincwork/boincwork.module
```diff
@@ -1041,7 +1041,133 @@ function boincwork_get_output() {
...
+  $min_core_client_version = parse_config($config, "<min_core_client_version>");
+  if ($min_core_client_version) {
+    echo "<min_client_version>$min_core_client_version</min_client_version>\n";
+  }
+
+  // Display platforms
+  boincwork_show_platforms();
+
+  // Terms of use (from Drupal)
+  $termsofuse = variable_get('boinc_weboptions_termsofuse', '');
```
brevilo: Hm, so `boinc_weboptions_termsofuse` is in fact the actual text? If so, my comment above on RPC coherence is obsolete. Should this be the actual terms of use text, does this completely replace BOINC's stock [terms of use file](https://boinc.berkeley.edu/trac/wiki/TermsOfUse) (also for the BOINC client etc), presumably because you reimplemented the whole `get_project_config` RPC?

Would it be possible to use BOINC's standard terms of use file instead and thereby avoid reimplementing the RPC entirely which would largely facilitate maintenance? Encapsulation and extension would of course be fine.

drshawnkwang:
> I presume it should read "existing users may not continue to use the site", right?
Correct, I have changed the text in the PR to reflect this.

TheAspens:
@tristanolive or @brevilo  are either of you able to review and if appropriate merge this?  Thanks!

drshawnkwang:
@TheAspens - Sorry to go back & worth with `WIP`. Talking to Oliver outside of GitHub we want to merge this in next week. Basically there is another deployment that is a prerequisite. So this PR should wait a few days.

TheAspens:
Ah ok - np.  I'm just doing my periodic nudge for pull requests that look like they are going stale.  Eventually I'd like to get to non-WIP pull requests (but that is probably a dream).  Tx for adding the WIP back on.

drupal/sites/default/boinc/modules/boincuser/boincuser.module
```diff
@@ -310,26 +352,6 @@ function boincuser_user($op, &$edit, &$account, $category = NULL) {
...
-            // Save profile information for use during Insert
-            $_SESSION['profileInfo'] = array(
-              'country' => $edit['field_country'][0]['value'],
-              'zip' => $edit['field_zip'][0]['value'],                  
-              'url' => $edit['field_url'][0]['value'],
-              'background' => $edit['field_background'][0]['value'],
-              'opinions' => $edit['field_opinions'][0]['value']
-            );
-            // With BOINC validation passed, make sure name is unique
-            $edit['name'] = create_proper_drupalname($edit['boincuser_name']);
```
tristanolive: This whole block of code may be obsolete now due to earlier changes in the registration process, but we should be sure the functionality of this particular line is handled elsewhere. The Drupal username is not exposed anywhere (ideally) to the end user, so the uniqueness of it should be handled automatically in the background by this create_proper_drupalname() function. (Drupal has a unique username requirement while BOINC does not, hence the need for this feature.)

drshawnkwang:
I've updated the code to resolve the concern @tristanolive pointed out above.

Also, there were some bugs in the registration code that have been fixed.

brevilo:
To avoid confusion: this is the Drupal-only interim solution, independent of #2504, right?

drshawnkwang:
@brevilo , yes - this is independent of #2504.

PR #2486, opened 2018-04-27
Mac: update info on code signing installer in build instructions

CharlieFenton:


CharlieFenton:
This makes a correction in the Mac build instructions for the 7.10 branch.

RichardHaselgrove:
I have verified that this PR is for documentation only and is safe to merge. I make no warranty as to the accuracy of the revised build instructions.

PR #2489, opened 2018-04-28
Mac: script which builds installer to update ca-bundle.crt certificat…

CharlieFenton:
…e on very old BOINC clients running under OS 10.3.9 - OS 10.5

NOTE: this script must be run on an OS 10.5 or OS 10.6 system which has Xcode 3 and PackageMaker 3 installed.

This is not required for the 7.10.x client release, but may optionally be added to the 7.10 branch if desired.

### DIFF
--- a/mac_installer/make_certificate_updater.sh
+++ b/mac_installer/make_certificate_updater.sh
@@ -48,6 +48,8 @@ ENDOFFILE
 cp -f "/tmp/pkg_scripts/postinstall" "/tmp/pkg_scripts/postupgrade"
 
 rm -fR "../BOINC_Installer/upate_boinc_certificate.pkg"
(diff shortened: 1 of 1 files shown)


JuhaSointusalo:
Minor nitpick. Scripts are usually run without `source` unless you really want to modify the current shell environment.

CharlieFenton:
Running a script without using _source_ requires that the executable bit be set in its permissions (at least on the Mac.) In the past, I found that it was problematic to preserve that bit when checking it into and out of source control. But this was probably before we switched to GIT, so it may no longer be true.

JuhaSointusalo:
I don't think I've had problems with Git managing file permissions. Except
maybe when I do something silly like mix Windows and Linux versions of
tools or filesystems.

Issue #2370, opened 2018-02-21
Detect whether Win10 has linux distro installed


AenBleidd:
Looks like GPUGrid started to test new application that can be run on linux installed on Win10: https://www.gpugrid.net/forum_thread.php?id=4716&nowrap=true#48980
As I understand BOINC can't detect the presence of linux on Win10.
I think it is a powerful option so boinc client should have an option to make such detection and send it to project.

JuhaSointusalo:
Some thoughts about WSL support.

- Platform vs something else
Declaring support for Linux apps via alt platform would be nice but WSL can't run all Linux apps yet. GPU apps for example won't work. Linux apps also need to be run with WSL launcher. The same way VirtualBox is supported should work. Some extra info in scheduler request, plan class on server and a wrapper with the app.
- Multiple distros
Since Fall Creators Update there is now several distros available. Some projects may be picky about which Linux distros they support. BOINC should check which distros are installed. BOINC server should somehow (e.g. command line parameter) inform client side which distro was accepted.
- Distro versions
Users may upgrade a distro in place or swap the filesystem for an entirely different distro. As such the name of the distro doesn't necessarily match what's actually running. BOINC should check the distro version.
- Might not be safe to use
WSL has a very simple init system which doesn't support things like starting services or anything else really. If I have understood correctly right now people put stuff in profile to have it run automatically when starting WSL. If WSL ever gets a fancier init system then using WSL might start some services in it. So running anything in WSL, even just for querying the distro version, might not be side effect free. BOINC should have an option to not use WSL.
- Some distros may be ok to use
Users may have installed distros for non-BOINC use but may have also installed distros for BOINC to use. BOINC should have an option to list which distros it can use. The previous bullet point, option to not use WSL, could be combined with this.
- App IPC
On Windows BOINC uses Windows shared memory segments for IPC and on Unixes memory mapped files. That means client and apps can't talk to each other and things like suspending a task won't work. Unix versions have also older code that uses System V shared memory segments. Windows and System V name the memory segment differently and it seems unlikely they could be made compatible. Memory mapped files ought to be compatible but needs to be tested. If memory mapped files work then BOINC needs to switch to using them on Windows. This would also help running Windows apps on Unixes with Wine.
- Per-user distro installs
[...]

PR #2491, opened 2018-04-29
WSL detect initial implementation

AenBleidd:
Windows 10 has WSL https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux feature.
At least one project (GPUGrid) is interested in this feature support in BOINC: https://www.gpugrid.net/forum_thread.php?id=4716&nowrap=true#48980

This pull request is a first try to implement WSL feature support in BOINC,

In this pull request I tried to use already implemented linux os detection mechanism. I made it general both for Windows and linux.
Feel free to suggest better option name.
Please make a review of this.
As soon as it will be decided that everything is OK I'll create clean pull request with all code squashed into one commit.


lib/hostinfo.h
```diff
@@ -32,6 +32,16 @@
...
 
+enum LINUX_OS_INFO_PARSER {
+    lsbrelease,
+    osrelease,
+    redhatrelease
+};
+
+const char command_lsbrelease[] = "/usr/bin/lsb_release -a 2>&1";
+const char command_osrelease[] = "/etc/os-release";
+const char command_redhatrelease[] = "/etc/redhat-release";
```
ChristianBeer: Those are technically no commands but files which need to be opened. Only `lsb_release` is a command which needs to be executed. Since the difference is significant for the developer how to handle those the variable name should be correct: `file_osrealease` and `file_redhatrelease`.

lib/hostinfo.cpp
```diff
@@ -291,3 +312,158 @@ int HOST_INFO::write_cpu_benchmarks(FILE* out) {
     );
     return 0;
 }
+
+bool HOST_INFO::parse_linux_os_info(FILE* file, const LINUX_OS_INFO_PARSER parser,
```
ChristianBeer: I'm not sure this belongs in this file. The file is about parsing our own XML file into a structure and is platform neutral. The `parse_linux_os_info(...)` stuff is linux specific and has nothing to do with the XML file. The idea is valid though and this should either stay in `hostinfo_unix.cpp` or go into a new `hostinfo_linux.cpp`.

davidpanderson:
I'd like to see the bulk of this put in a separate file, say hostinfo_wsl.cpp

lib/hostinfo.cpp
AenBleidd: But here's the idea to reuse linux parsing code in windows to parse linux :)

lib/hostinfo.h
AenBleidd: Fixed

lib/hostinfo.cpp
AenBleidd: Fixed

AenBleidd:
@ChristianBeer, @davidpanderson, please review again.
@Ageless93, @RichardHaselgrove, @TheAspens, please review too.

AenBleidd:
@davidpanderson, BTW, do I need to add information about WSL somewhere else (scheduler, server code etc.)?

davidpanderson:
Probably easiest for me to do this.  I can do it in a separate PR>

AenBleidd:
@davidpanderson, I'd like to keep myself as the author of these changes. Could you please review this PR and I'll create new clear pull request that can be merged into your branch where you will add your changes? Thanks

davidpanderson:
I don't have a machine with WSL, but this works (i.e. doesn't find WSL) on my Win machine, and the code structure is fine, so I'm merging it.

JuhaSointusalo:
Does `_popen()` work with WSL processes? If it does you could have something like this

```
#if LINUX_LIKE_SYSTEM
    f = fopen("/etc/os-release", "r");
#else
    f = _popen("wsl.exe cat /etc/os-release", "r");
#endif
```

Ifdefs aren't great but `_popen` would take care of all the boilerplate for you. And you could move all of the Linux specific OS detection code to `hostinfo_linux.cpp`. I think that would make the code cleaner.

I think there would be some benefit in having the WSL version string to be the same as the Linux version string. Linux version gets kernel name and version with `uname` syscall. You get the same from WSL with `uname -s` and `uname -r`.

You run commands with `bash(.exe)`. There is some possibility of clash between bash.exe the WSL launcher and bash.exe the shell ported to Windows (from, say, Cygwin). Use `wsl.exe` instead.

IMHO, WSL is an optional software package in the same way as VirtualBox is or Python or .NET would be, not an integral part of the OS (even if the implementation heavily depends on the OS). I would like that the variables are not called `os_*` and that `get_wsl_information()` is called straight from `get_host_info()` instead of inside `get_os_info()`.

I would like that there would be an option to not use WSL, if not in this PR then at least before next client release is done. I would also like if the code enumerated all the WSL distros instead of the default one. For the reasons stated in #2370.

AenBleidd:
@JuhaSointusalo, thank you for such detailed review.

> Ifdefs aren't great but _popen would take care of all the boilerplate for you. And you could move all of the Linux specific OS detection code to hostinfo_linux.cpp. I think that would make the code cleaner.

Interesting suggestion. I'll play with it.

> I think there would be some benefit in having the WSL version string to be the same as the Linux version string. Linux version gets kernel name and version with uname syscall. You get the same from WSL with uname -s and uname -r.

If the previous will work I'll implement this too

> You run commands with bash(.exe). There is some possibility of clash between bash.exe the WSL launcher and bash.exe the shell ported to Windows (from, say, Cygwin). Use wsl.exe instead.

I had some issues passing commands to wsl.exe. I think I see another approach here so I'll try to implement it too.

>  IMHO, WSL is an optional software package in the same way as VirtualBox is or Python or .NET would be, not an integral part of the OS (even if the implementation heavily depends on the OS). I would like that the variables are not called os_* and that get_wsl_information() is called straight from get_host_info() instead of inside get_os_info().

Actually, I did this in this way because it will require to rewrite existing get_os_info() code to store the information about os version not in the string only because there is no sense to search for a WSL on os older than Win10. I can play with this too and show what I'll get as a result in a pull request.

> I would like that there would be an option to not use WSL, if not in this PR then at least before next client release is done.

Of course, this will be done. I prefer to do tasks one by one. And adding WSL support is a big task so I wanted to start with detection only. IMHO, this is a good time to create several tickets with detailed description of stuff that should be done for this WSL support.

>  I would also like if the code enumerated all the WSL distros instead of the default one.

Currently I found no way to get the list of all installed distros but I'll continue my investigation to find such way.

JuhaSointusalo:
> I had some issues passing commands to wsl.exe.

What kind of issues? From command line they seem to be very similar. `bash.exe -c command` and `wsl.exe command` and if you use distro exes `ubuntu.exe run command`.

> Actually, I did this in this way because it will require to rewrite existing get_os_info() code to store the information about os version not in the string only because there is no sense to search for a WSL on os older than Win10.

You can do something like

```
void get_wsl_info() {
    OSVERSIONINFOEX osvi = { sizeof(OSVERSIONINFOEX) };
    GetVersionInfoEx(&osvi);
    if (osvi.dwMajorVersion < 10) {
        return;
    }
    ...
}
```

That pattern is used somewhere else in BOINC too.

> Currently I found no way to get the list of all installed distros but I'll continue my investigation to find such way.

```
C:\Users\Juha>wslconfig /list
Windows Subsystem for Linux Distributions:
Ubuntu (Default)
openSUSE-42
```

[WSL-DistroLauncher](https://github.com/Microsoft/WSL-DistroLauncher) docs don't really spell it out but the names seem to match the .exes.

AenBleidd:
> What kind of issues?

I tried to run wsl.exe and then pass commands using pipe. Now I'll just run wsl.exe with command already. Looks like it will work nice. So it was not a real issue but just wrong usage.

Thank you for all this. I plan to work on this stuff on weekend to get it fixed before Monday. Then will  make WSL usage optionable.

CharlieFenton:
@AenBleidd: The file _client/hosting_linux.cpp_ should **not** be included in the Macintosh builds as it contains only code for Linux. Please undo this change as soon as possible. In the future, please check with me before modifying the BOINC Xcode project, or let me know when changes are needed and let me make them. Thank you.

AenBleidd:
@CharlieFenton, are you sure? I add this file to Xcode project because hosting_unix.cpp is there and hosting_linux.cpp just contains the code moved from hosting_unix.cpp
I can undo these changes in a couple of hours.
Sorry for that.

CharlieFenton:
I am sure. I temporarily disabled compiling that file and it built correctly. (Note that all the functions have _linux_os_ in their names.) I know you had good intentions, but it works better for me if I am the only one modifying the Xcode project.

AenBleidd:
Removed in #2545. Please merge it

ChristianBeer:
@CharlieFenton What kind of error is triggered if `hostinfo_linux.cpp` is included for you? The Travis CI job compiled the client including this file without problems. See build log: https://travis-ci.org/BOINC/boinc/jobs/384471783 If this change really broke the Mac build the Travis job should have caught this. Regardless if the file was needed or not.

CharlieFenton:
No errors, just unnecessary clutter and extra compile and link time. The project is large enough without including irrelevant files.

Issue #2496, opened 2018-05-01
Latest changes in master break make_project and (project) upgrade 


drshawnkwang:
Commit 6093ed3, part of PR #2445, causes problems elsewhere. Specifically this line:
https://github.com/BOINC/boinc/commit/6093ed3a3f0bbda7e4eed312958f858431743ca7#diff-1851d371ca1dc20d62d76a09171384d1R27

The problem is the `require_once("../inc/hosts.inc");`

When a someone runs `./tools/make_project` or `./tools/upgrade`, they will receive the following error, formatted to be more readable:

```
Fatal error: Cannot redeclare show_host() 
(previously declared in /srv/BOINC/projects/testp1/html/inc/db_ops.inc:886) 
in /srv/BOINC/projects/testp1/html/inc/host.inc on line 218
```

This is caused by having two functions named `show_hosts()`, one in `html/inc/db_ops.inc` and a second in `html/inc/host.inc`. When adding `require_once("../inc/hosts.inc");`, the result is that both functions are loaded and PHP complains. This error appears when creating  project and upgrading one.

As a temporary measure, I commented out [Line 27 in `user.inc`](https://github.com/BOINC/boinc/blob/master/html/inc/user.inc#L27) and the two scripts now work. Of course this is probably not a the real solution since it probably breaks something in @davidpanderson delete-user code. Or breaks something elsewhere.

davidpanderson:
fixed in PR #2497

TheAspens:
PR #2497 was merged to master so this should be resolved.

drshawnkwang:
Thanks for the fast work guys!

PR #2497, opened 2018-05-01
web: change names of admin functions to e.g. admin_show_user()

davidpanderson:
to avoid conflict with non-admin versions.  Fixes issue #2496.


TheAspens:
Tested with https://github.com/TheAspens/boinc-server-test  

If I had done this on the previous merge, it would have caught this issue.

TheAspens:
(also tested manually as well)

PR #2500, opened 2018-05-02
Email change notification

Uplinger:
This is to address the Email Change Notification for BOINC.  Please view this issue with more information:   https://github.com/BOINC/boinc/issues/2451


html/inc/email.inc
```diff
@@ -21,14 +21,20 @@
...
+require_once("../inc/token.inc");
 
 // send an email, using PHPMailer or not.
 //
-function send_email($user, $subject, $body, $body_html=null) {
+function send_email($user, $subject, $body, $body_html=null, $email_addr=null) {
     if (function_exists("make_php_mailer")) {
-        require_once("../inc/phpmailer/class.phpmailer.php");
+        require_once("../inc/PHPMailer/src/PHPMailer.php");
+        require_once("../inc/PHPMailer/src/SMTP.php");
```
TheAspens: Please update https://boinc.berkeley.edu/trac/wiki/ServerIntro#PHPMailer so that has the correct changes so that projects know what to do.

Uplinger: I have updated the instructions to point to 6.0.x PHPMailer and commited a code change to allow for both options, this way it won't break older version of PHPMailer.  https://github.com/BOINC/boinc/pull/2500/commits/4aa5acbc45f9e95197e3dd486bb1d8f2d8399790

TheAspens:
Although I reviewed and approved this change, since I collaborated with Keith on this and we work on the same team I think that others should do reviews and merge.

TheAspens:
@brevilo @tristanolive @davidpanderson - could one or more of you review this please?  This is now blocking me from finishing #2472

TheAspens:
Anyone else is welcome and invited to review this as well :-)

html/inc/email.inc
```diff
@@ -86,6 +99,32 @@ For further information and assistance with ".PROJECT.", visit
...
+    $subject = PROJECT." email address change.";
+    //Body for the new email address to explain how quickly they can do another email change.
+    $body_new = "Your email address was changed from ".$user->previous_email_addr.
+" to ".$user->email_addr." on ".date('F j \a\t g:i a T', $user->email_addr_change_time).
+".  You will not be able to change your email address again until ".date('F j \a\t g:i a T', $user->email_addr_change_time + $duration).
+".  If you need to undo this immediately, please look for an email from us at your ".$user->previous_email_addr." address.";
+
+    //We need to send a different version of the email to the old address.
+    $body_old = "Your email address has been changed. If you did not take this action, 
+then please click on the link below reverse this process and change your password.
```
JuhaSointusalo: "to"

html/inc/boinc_db.inc
```diff
@@ -182,9 +182,16 @@ class BoincUser {
     }
     static function lookup_email_addr($email_addr) {
         $db = BoincDb::get();
-        $email_addr = BoincDb::escape_string($email_addr);
+        $email_addr = strtolower(BoincDb::escape_string($email_addr));
         return self::lookup("email_addr='$email_addr'");
     }
+    //This is to find email addresses that have changed in the past 7 days.
+    static function lookup_prev_email_addr($email_addr) {
+        $db = BoincDb::get();
```
JuhaSointusalo: `$db` is unused.

html/user/recover_email.php
```diff
@@ -0,0 +1,64 @@
...
+
+page_head(tra("Recover email address"));
+
+$userid = get_int("id", true);
+$token = get_str("token", true);
+
+if(is_valid_token($userid, $token, TOKEN_TYPE_CHANGE_EMAIL)) {
+    $tmpuser = BoincUser::lookup_id_nocache($userid);
+    //We can only change passwd_hash if we can get the userdata. 
+    if($tmpuser) {
```
JuhaSointusalo: You might to want add an `else` just in case.

html/user/am_set_info.php
```diff
@@ -111,14 +112,27 @@ function success($x) {
 $send_email = BoincDb::escape_string($send_email);
 $show_hosts = BoincDb::escape_string($show_hosts);
 $venue = BoincDb::escape_string($venue);
+$send_changed_email = false;
 if ($email_addr) {
+    $tmpuser = BoincUser::lookup_email_addr($email_addr);
```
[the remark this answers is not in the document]
Uplinger: So, I think what the option you're asking for is.  If email_addr does not equal the same as the user of the auth_token, return an error. if it is the same email_addr for the user with auth, then set email_addr to null for local usage since that is the trigger below in the code.  Does that sound reasonable?

html/inc/boinc_db.inc
Uplinger: I'm going to make the change to remove it, also from a few other functions that call escape_string right after it, because within that function, it calls get().

html/user/recover_email.php
Uplinger: I'm going to put in this invalid token as well.  Because letting someone know a userid is invalid may give them too much information.

brevilo:
Before I final review it would be useful to see the various [changes](https://github.com/BOINC/boinc/commit/d0fcd1e729bff9ed1931b4341d2ad2381849d333) by [David](https://github.com/BOINC/boinc/pull/2472#issuecomment-387236026) incorporated first.

html/inc/email.inc
```diff
@@ -86,6 +99,36 @@ For further information and assistance with ".PROJECT.", visit
     return send_email($user, $subject, $body);
 }
 
+function send_changed_email($user) {
+    $body_new = "";
+    $body_old = "";
```
JuhaSointusalo: This is really a minor issue but Scrutinizer is flagging these as major issue. The variables are not used before they are overwritten. You could remove them from here.

Uplinger:
@brevilo @davidpanderson @TheAspens @JuhaSointusalo I have incorporated all the changes as well as fixed a few bugs from merging of code.  I have tested it locally, let me know your thoughts. 

Thanks,
-Keith

html/user/am_set_info.php
```diff
@@ -111,7 +112,24 @@ function success($x) {
...
+    if ($tmpuser) {
+        xml_error(ERR_BAD_EMAIL_ADDR, "There's already an account with that email address.");
+    }
+
+    //check if the email address is included in previous_email_addr window. 
+    //
+    $tmpuser = BoincUser::lookup_prev_email_addr($email_addr);
+    if ($tmpuser) {
+        xml_error(ERR_BAD_EMAIL_ADDR, "Email address is already in use");
```
JuhaSointusalo: This and the above should probably be `ERR_DB_NOT_UNIQUE` to match creating accounts.

html/user/edit_email_form.php
```diff
@@ -31,20 +31,24 @@
...
-
-// we need the password here not for verification,
-// but because we store it salted with email address,
-// which is about to change.
-
-form_input_text(tra("Password"), "passwd", "", "password");
-form_submit(tra("Change email address"));
-form_end();
+if ($user->email_addr_change_time + 604800 > time()) {
+    echo tra("Email address was changed within the past 7 days. Please look for an email to $user->previous_email_addr to verify this change.");
```
JuhaSointusalo: The word "verify" feels wrong to me. You can't verify the change but revert it with the email sent to the previous email address.

html/user/recover_email.php
```diff
@@ -0,0 +1,69 @@
...
+//
+// BOINC is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
+
+// ADD COMMENT
```
JuhaSointusalo: Missed this one.

Uplinger:
@JuhaSointusalo Please take another look.  I think I incorporated your suggestions completely. 

Thanks,
-Keith

JuhaSointusalo:
I haven't tested this so I don't formally stamp this Approved but it looks ok to me.

Uplinger:
@davidpanderson Can you take a look at the pull request.  I have made the change to the indexes to help when doing a query.  I have also change the order of a query to use the indexes first.  

Thanks,
-Keith

html/ops/db_update.php
```diff
@@ -1107,6 +1107,7 @@ function update_4_19_2018() {
 	add column previous_email_addr varchar(254) not null default '',
 	add column email_addr_change_time double not null default 0
     ");
+    do_query("alter table user add index user_email_time (email_addr_change_time)");
```
TheAspens: I think this needs a semi-colon at the end doesn't it?

Uplinger: Where are you thinking the semicolon goes, there is one after the do_query close bracket?  From previous examples of do_query, a semicolon is not needed before the end quotes.  Note, you should be able to see similar update query in function update_5_12_2004.

TheAspens: Ok - I was under the impression that mysql required a semi-colon at the end of statements, but if you've tested it and it works then clearly it is fine.  I retract my question.

TheAspens:
@brevilo and @davidpanderson - can either of you guys complete the review for this?  Thanks!

davidpanderson:
The new user fields also should be added to  DB_USER::db_parse() and DB_USER::db_print() (in db/boinc_db.cpp), even if they aren't used in the C++ code.  Otherwise if someone adds a field to user that IS used from C++, confusion will result.

Uplinger:
@davidpanderson My local tests succeeded.  You can merge when ready.  

Thanks for the feedback from everyone on this ticket!
-Keith

brevilo:
Sorry for being too late this time.

David fixed the `schema.sql` changes (BOINC needs add columns always at the end due to its `select * from` habit) I would have flagged and also commented on the then missing C++ API changes.

Thanks everyone!

Issue #2413, opened 2018-03-15
EU-GDPR compliance


brevilo:
On May 25th, 2018 a new [General Data Protection Regulation (GDPR)](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) is going to be effective in the EU. While most of its legal requirements only affect the projects themselves, at least one part affects BOINC directly - the sign-up process. As of that day users who want to sign-up to a project need to give their informed explicit consent to how the project processes their data, before any personally identifiable information about that new user gets processed/stored by the project. This opt-in statement of consent then needs to tracked by the project as proof.

Given the current way the BOINC Manager/Client project sign-up workflow is done, I think that all EU projects are basically forced to disable client-based login and only allow a (customized!) web sign-up process until the missing features got added. In order for BOINC being GDPR compliant I think the following needs to be changed:

* During sign-up, request the project specific declaration of consent and checkbox label (see below) from the project and display it to the user (multi-language support required, i.e. add language code to request)
* That dialog should support HTML such that links to, say, the external privacy policy could be included in the text
* Add an opt-in checkbox showing the label requested above (again, with link support)
* Allow sign-up only when checkbox is selected and transmit that as an additional boolean attribute in the account creation RPC
* Add an additional `date` type attribute `privacy_consent_dt` to the `user` table
* Store the datetime of the user's consent in the new attribute

The standard BOINC web interface also needs to support equivalently the above workflow. It might even be the case that we're required to get this consent from existing users as well, so the web interface should make the consent from equally available outside of the sign-up process (for a logged-in user in this case). We're already working on incorporating the necessary changes to the Drupal web interface.

Note to all account managers: you should be affected in the same way as the BOINC client!

AenBleidd:
Correct me if I wrong but these changes mean that there will be no backward compatibility with old clients and there will be no possibility to attach to the project using old client?
I think this will be a great issue for Android application and probably Raspberry Pi (as I can see Raspbian has 7.0.27 boinc client version only).

tristanolive:
I think old clients could only attach to existing accounts, in that case. It might be possible for an attach error to include information about how to go to the project and create an account manually; or perhaps projects could create limited or disabled accounts in these situations that require manually logging in at the website in order to accept terms & conditions and activate the account.

brevilo:
Correct, old clients mustn't be allowed to attach (read: create a new account) to any EU project - it's legally impossible.

brevilo:
[...]

Issue #1929, opened 2017-06-03
Createing an account with MacOS and Windows uses the users' computer username.


drshawnkwang:
https://einsteinathome.org/content/how-can-i-delete-my-account

Users report that when creating their account, BOINC client uses their computer's username. Account name should be a field the user fills in.

PS- I have not experienced this behavior with the Linux client.

ChristianBeer:
This is intentional and the user is redirected to the project website (at least when adding an account via Manager) to change the name. This does not happen when using the client to add an account.

The easiest way is to remove this but how should we generate usernames then? The next possibilitiy is to add a notice in the Client output that tells the user which website to visit in order to change the username. The best and more complex solution would be to ask the user for a name when signing up.

Ageless93:
I wonder when the next person comes complaining that his BOINC client shows his operating system's user name in the event log messages and how that's an infringement of their privacy.

drshawnkwang:
Since no one has worked on this in a year: an alternative approach is to modify `create_account` RPC such that it ignores the username parameter, and sets every user to have the general username 'User' or 'BOINC Volunteer'. This is not ideal: but perhaps better than nothing, i.e. the status quo?

RichardHaselgrove:
BoincView handles the problem by including the username to be used in the account creation dialog, pre-populated with the computer logon account name, but editable.

![boincview create account](https://user-images.githubusercontent.com/14886436/46549716-d9555f80-c8ca-11e8-94ec-6486f8cce5a2.png)

JuhaSointusalo:
Adding an extra text field for the user name is easy enough but what to do with the web page you go to after the account is created? It would be a bit silly to ask for a user name on the web page just after asking for a user name in the Manager.

![image](https://user-images.githubusercontent.com/15806192/46554522-44159380-c8e9-11e8-9d4c-36d1ea21fbe2.png)

PR #2504, opened 2018-05-05
User Opt-in consent

drshawnkwang:
Resolves GDPR User Opt-in Consent #2413.

This is a work in progress (WIP) merge. I wish to allow people to review what I have done and comment on it. As of this initial PR the following features are present.

1. New `consent` and `consent_type` tables which record a user's consent.
1. Web form now has terms of use (terms_of_use.txt) file which forces users to opt-in.
1. `create_account` and `am_set_info` RPCs modified to record consent.
1. HTML Ops page added for admin to manage `consent_type` table.


html/user/create_account.php
```diff
@@ -91,6 +95,25 @@
...
+        // set to 1.
+        if ($optin==0) {
+            $rc = consent_to_termsofuse($user, 1, 0, 1, $source);
+        } else  {
+            $rc = consent_to_termsofuse($user, 1, 1, 0, $source);
+        }
+        if (!$rc) {
+            xml_error(-1, "database error, please contact site administrators");
+        }
+    }
```
JuhaSointusalo: If consent is required but whoever is creating the account knows nothing of consent then what? I think the code that checks that consent was given should go before `make_user` call.

html/user/user_optin_action.php
```diff
@@ -0,0 +1,71 @@
...
+$user = $_SESSION['user'];
+$authenticator = $user->authenticator;
+$perm = $_SESSION['perm'];
+
+// Set consent in database
+$consent_result = BoincConsent::lookup("userid=$user->id AND consent_id=1");
+if ($consent_result) {
+    $cquery = "consent_flag=1, consent_not_required=0";
+    $cquery .= " WHERE userid = $user->id AND consent_id = 1";
+    $rc1 = $consent_result->update($cquery);
```
JuhaSointusalo: Update `consent_time` and `source` as well?

html/inc/consent.inc
```diff
@@ -0,0 +1,59 @@
...
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
+
+// functions dealing with the consent and consent_type tables.
+
+include_once("../inc/boinc_db.inc");
+include_once("../inc/util.inc");
+
+function consent_to_termsofuse($user, $consent_id, $consent_flag, $consent_not_required, $source, $ctime = 0) {
```
JuhaSointusalo: This function could use different name since there is now more than just the general term of use consent type.

html/inc/prefs_project.inc
```diff
@@ -137,6 +137,16 @@ $privacy_pref_descs = array (
     ),
 );
 
+$privacy_consent_descs = array(
+    new PREF_CONSENT(
+        tra("Do you consent to %1 exporting your data?", PROJECT),
+        "Data is exported to BOINC statistics aggregation Web sites.",
```
JuhaSointusalo: Needs `tra`.

html/ops/manage_consent_types.php
```diff
@@ -0,0 +1,120 @@
...
+    $rowi=1;
+    foreach ($_consenttypes as $ct) {
+        echo "<tr class=row$rowi><form action=manage_consent_types.php method=POST>\n";
+
+        echo "<input type=hidden name=consent_id value=$ct->consent_id>";
+        echo "  <td>$ct->consent_id</td>";
+
+        echo "  <td>$ct->description</td>";
+
+        if (!in_rops() and ($rowi!=1)) {
```
JuhaSointusalo: Row / consent type 2 should be protected too. There could be more stock consent types in the future so you should really protect a larger block of ids. So something like ids < 10 000 are owned by BOINC.

html/user/am_set_info.php
```diff
@@ -210,6 +222,26 @@ function success($x) {
...
 }
 
+// Build the query for the consent table. consent_id must be present.
+$cquery = "";
+$now = time();
+if ($consent_id) {
+    $cquery .= "consent_id=$consent_id";
+    if ($consent_settime==1) {
+        $cquery .= ",consent_time=$now";
+    }
```
JuhaSointusalo: I think it would be better if `consent_time` is set any time consent is given or changed and account managers have no say in the matter.

As for the rest of the consent parameters I would check that either they are all set or none of them are set. I don't think setting only some of them would work well.

JuhaSointusalo:
About revoking consent. I don't think it's possible to just record the fact and then carry on. Revoking consent requires some more actions.

If we require third parties to remove any data about a user after the user deletes their account then I would expect that we need to do the same when the user revokes permission to export data about his account. The `user_deleted` and `host_deleted` that were added for deleting accounts would work for this too I think, even though nothing was really deleted. This needs to be handled in both `am_set_info` RPC and when saving project preferences.

As for revoking general terms of use, I don't think it's possible to handle that in `am_set_info` RPC at all. Revoking consent to TOU is the same as deleting account. Quite a bit of effort went into making delete account feature secure and if it were possible to revoke the consent with `am_set_info` then it would bypass the new security additions. I think revoking consent to TOU needs to be blocked entirely and user redirected to delete account feature.

drshawnkwang:
@JuhaSointusalo , could you copy and paste your comment into the Issue ticket #2413. Let's keep discussion about the PR here, and other (good) ideas about the user consent feature in the Issue ticket.

This is because I think you have some good points

html/ops/manage_consent_types.php
brevilo: Any assumptions made here will eventually be invalid. Better come up with a flexible approach, e.g. a boolean flag `protected` that protects the row when `true`.

drshawnkwang:
> Shouldn't this update `consent_time` and maybe `source` too?

@JuhaSointusalo , you asked this a few times and the answer is: it's complicated. The assumption is you are updating a row in the `consent` table, where `consent_time` and `source` are already set. Should these be updated from their original time/source? Is the record static and the user allowed to give and then revoke consent?

Part of the answer is related to your [longer comment](https://github.com/BOINC/boinc/pull/2504#issuecomment-390789466). It depends on the design of the user opt-in functionality.

html/user/am_set_info.php
drshawnkwang: I agree with you in principle, but I received very little feedback from AM developers on what would be best for them to be able to change, hence a very flexible/open API for interaction.

html/inc/prefs_project.inc
drshawnkwang: Thanks, I'll add this soon.

html/inc/consent.inc
drshawnkwang: Good point, will change...

html/user/create_account.php
drshawnkwang: Right now there are multiple ways this Web RPC is called, including simply typing in `https://project_url/create_account.php?email_addr=xyz@example.com&passwd_hash=ahash&username=name`

This means anyone can create an account without knowledge of a project's terms of use/opt-in consent. The boinc-command line client is another place where you can create an account without ever seeing the terms of use and consenting to it.

Ultimately, projects which are in Europe and really care about GDPR should actually turn off this create_account RPC and only allow users to register via their Web site, which guarantees the user sees the terms of use.

JuhaSointusalo: I was thinking that Manager and boinccmd would be later enhanced to pass `optin` and `source` parameters. A user could be using some third party manager that doesn't show TOU and the user might not know he needs to give consent.

I was under the impression that new (non-anonymous) accounts can't be created unless the user consents to TOU. But that raises the question, if user has not given or revoked consent to TOU at all, then does that make the account automatically anonymous?

drshawnkwang:
That's okay, Oliver and I are discussing in #2413 some new changes and proposals, since we know now have a different understanding of what GDPR requires. I will actually mark this PR as WIP while we try to flush this out. Hopefully the changes will be simple for me to add. Sorry about this!

TheAspens:
No worries - our crush of GDPR changes is over so I'll be able to monitor for a status change in this and then review when it is out of WIP status.  Thanks for implementing!

drshawnkwang:
I've removed the WIP marker, and this PR is ready for more review.

html/inc/consent.inc
```diff
@@ -0,0 +1,70 @@
+<?php
+// This file is part of BOINC.
+// http://boinc.berkeley.edu
+// Copyright (C) 2017 University of California
```
TheAspens: copyright should be 2018

html/user/user_agreetermsofuse.php
```diff
@@ -0,0 +1,41 @@
+<?php
+// This file is part of BOINC.
+// http://boinc.berkeley.edu
+// Copyright (C) 2008 University of California
```
TheAspens: Copyright should be 2018

db/schema.sql
```diff
@@ -804,3 +804,27 @@ create table host_deleted (
...
+    description             varchar(255)    not null,
+    enabled                 integer         not null,
+    protected               integer         not null,
+    privacypref             integer         not null,
+    primary key (consent_id)
+) engine=InnoDB;
+
+insert into consent_type (consent_id, shortname, description, enavled, protected, privacypref)
+    values (1, 'ENROLL', 'General terms-of-use for this BOINC project.', 0, 1, 0);
+    values (2, 'STATSEXPORT', 'Do you consent to exporting your data to BOINC statistics aggregation Web sites?', 0, 1, 1);
```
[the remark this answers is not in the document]
TheAspens: @davidpanderson  I see that this content is included in db_update for existing projects.  However, for new projects, how should tables be populated with default content?

davidpanderson:
Don't use a DB table for consent types; use a PHP data structure instead.
Much simpler, and also translatable.

html/inc/consent.inc
```diff
@@ -0,0 +1,70 @@
...
+        "(id, userid, consent_name, consent_time, consent_flag, consent_not_required, source) " .
+        "values(0, $user->id, '$myn', $mytime, $consent_flag, $consent_not_required, '$mys')"
+    );
+
+}
+
+function consent_after_login($user, $perm=true, $next_url = "") {
+    session_start();
+    $_SESSION['user'] = $user;
+    $_SESSION['perm'] = $perm;
```
TheAspens: This code is the first time that BOINC uses session.  Up to this point, the website has been essentially stateless (i.e. all information needed to support the request is contained on the request or loaded from the database based on information in the request).  I have two questions about this.

1) Do any BOINC projects do load balancing using more than one web server to serve their site?  If not, then question 2 isn't important.
2) Are we expecting BOINC projects that have more than one web server to use some form of server affinity?

If there is a project using load balancing between servers, then caching the data in the session could become problematic if the users request winds up on a server that doesn't have this data cached.  

This is a pretty big architectural decision to go from stateless to stateful requests.  Perhaps it would be better to store the data you are putting in the session in a temp cookie so that we don't make this change now.

TheAspens:
I attempted to build this branch using https://github.com/TheAspens/boinc-server-test however the build fails.  I'm trying to look at what is going.  @marius311 - how do I look to see why your docker build might have failed?

drshawnkwang:
> Don't use a DB table for consent types; use a PHP data structure instead.
Much simpler, and also translatable.

The plan for this database table is to be able to create more consent_types as time goes on, as well as having individual projects create project-specific consent types. I'm not sure a PHP data structure would give us the flexibility to do all of that, as well as being able to be cross-referenced the `consent` table. Perhaps it would be possible, but a since a mysql database exists, and it's readily available for this kind of use, I used it.

html/inc/consent.inc
drshawnkwang: I admit I did not think of the high-level implications of using a stateful architecture when writing this.

I (obviously) need some kind of temporary space to store the user information  while traversing multiple pages, so the SESSION seemed like an obvious candidate. I have used it on occasion for Drupal-Web development.

But if the same functionality can be made with a temporary cookie, then I'll refactor this section of the code.

davidpanderson:
Using a PHP data structure would be best here because it's simpler and
translatable

html/inc/account.inc
```diff
@@ -140,4 +159,28 @@ function login_form($next_url) {
     form_end();
 }
 
+function user_agreetermsofuse_form($next_url) {
+    form_start(secure_url_base()."/user_agreetermsofuse_action.php", "post");
+    form_input_hidden("next_url", $next_url);
+
+    $tou_file = "../../terms_of_use.txt";
```
TheAspens: I would also reuse the define here

html/inc/account.inc
```diff
@@ -140,4 +159,28 @@ function login_form($next_url) {
...
+        $myitems = array(
+            array("agree_to_terms_of_use", "", false),
+        );
+        form_checkboxes(tra("Do you agree to the terms of use above?"), $myitems, 'tabindex="0"');
+    }
+    else {
+        // error - no terms of use for user to agree to!
+    }
+
+    form_submit("I agree");
```
TheAspens: Shouldn't this be translated?

html/inc/consent.inc
TheAspens: Yea - I would go the temp cookie route and just put the authenticator and perm flag into the temp cookie so that you have it available after the consent.

TheAspens: I agree that if session was already and established pattern on the site then this would make sense, but since it is new there might be deployment cases that this where this could be a big surprise and result in some bug reports while they tried to figure out what is going on.

db/schema.sql
```diff
@@ -804,3 +804,27 @@ create table host_deleted (
...
+    consent_id              integer         not null auto_increment,
+    shortname               varchar(255)    not null,
+    description             varchar(255)    not null,
+    enabled                 integer         not null,
+    protected               integer         not null,
+    privacypref             integer         not null,
+    primary key (consent_id)
+) engine=InnoDB;
+
+insert into consent_type (consent_id, shortname, description, enavled, protected, privacypref)
```
TheAspens: enavled should be enabled.

db/schema.sql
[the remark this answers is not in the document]
TheAspens: Also - the insert query should be this:

insert into consent_type (consent_id, shortname, description, enabled, protected, privacypref)
values (1, 'ENROLL', 'General terms-of-use for this BOINC project.', 0, 1, 0),  (2, 'STATSEXPORT', 'Do you consent to exporting your data to BOINC statistics aggregation Web sites?', 0, 1, 1);

drshawnkwang:
@The Aspens, Thanks for taking a look. I obviously have two big items and a few small ones

1. Refactor the login->terms-of-use workflow to use temporary cookies
1. Fix schema.sql, which I obviously didn't take my time with when I last changed it (sorry!)

And then add some translation and fix the copyright dates. I'll start fixing these on Monday and report back.

html/inc/consent.inc
brevilo: FWIW, I concur.

brevilo:
> However, for new projects, how should tables be populated with default content?

I don't see a problem adding that to schema.sql. Alternatively there could be a file like `constraints.sql`, say, `defaults.sql` or `content.sql`.

That said, how's `constraints.sql` even used for new project instances? `init_db` doesn't make use of it...

drshawnkwang:
@TheAspens , my latest commits should fix the schema.sql issue, as well the other items brought up. Please let me know if you find further things that need attention before merging.

TheAspens:
> That said, how's constraints.sql even used for new project instances? init_db doesn't make use of it...

It is invoked via py/Boinc/database.py function create_database.

drshawnkwang:
Hmm... not to give myself more work, but would it be worthwhile to put the `consent_type` content in a new SQL file? Call it 'initial_content.sql`. 

Then in edit `py/Boinc/database.py` to read:
```
for file in ['schema.sql', 'constraints.sql', 'initial_content.sql']:
        _execute_sql_script(cursor, os.path.join(srcdir, 'db', file))
```

TheAspens:
I think that would be the best way to go.  I would just call it content.sql.

TheAspens:
I would also encourage you to work to resolve the disagreement about the way to store content (PHP vs Table) since that dispute needs to be resolved before the pull request can be merged.  A call would be a quick way to resolve it or it can be discussed in this thread.

The teamspeak channel is available for use if you want to do a call.  Instructions are available here: https://docs.google.com/document/d/1lvCZxhwP1xmRxjHQGY-QvCyDwlO5Br-oGrNizi0NPpg/edit?usp=sharing

AenBleidd:
BTW, why don't we use https://discordapp.com ?
I found it better and easier than TeamSpeak. I have some experience using it for internal chats and voice calls in my team on my job

drshawnkwang:
> I think that would be the best way to go. I would just call it content.sql.

Okay, I'll add this tomorrow morning, it will be quick

> I would also encourage you to work to resolve the disagreement about the way to store content (PHP vs Table) 

I don't suppose we could talk about it during, or perhaps after the Contributor call Thursday? That assume all if the people who are interested in this discussion will attend the call.

Otherwise I'll setup a Doodle poll and we'll discuss the different options.

davidpanderson:
DB tables are a heavyweight (i.e. complex) option that should be used only when necessary.
Defining a data structure in a PHP file is simple and easy to change.

TheAspens:
@davidpanderson - @drshawnkwang indicated that he expects projects to add custom values to the list of consent types.  How will they be able to easily do that without having to keep a fork of the BOINC code if that data is stored in a PHP object?

Also - since the the content types are clearly foreign keys for items the content table, having their meaning and the relationship in the database is very natural and a standard practice.  I think this is a very reasonable approach.

@drshawnkwang - @davidpanderson asked how the translations would be maintained if the content types were stored in the database.  Do you have any thoughts about how that would done?

TheAspens:
@davidpanderson and @drshawnkwang - we can discuss on the contributor call at the end if we don't reach resolution by then.

AenBleidd:
> My only response is: https://xkcd.com/1810/ :-)

Got it. No problem :-)

drshawnkwang:
`consent_type` in the database contains a different types of consent. There are four types that I can identify:
1. protected=1 privacy=0 ; defined by BOINC upstream, does not deal with privacy preference,
1. protected=1 privacy=1 ; defined by BOINC upstream, is a privacy preference,
1. protected=0 privacy=0 ; project-specific, does not deal with privacy preference,
1. protected=0 privacy=1 ; project-specific, is a privacy preference.

Thus \# 2 and \# 4 will appear on the privacy preferences page. In the case of \# 2, this is sent through the `tra()` function in [prefs_project.inc, line 147 and 148](https://github.com/BOINC/boinc/pull/2504/files#diff-1d3e4e7ca23e6abd6d34499b647e657dR147).

While I am not an expert in the translation system, my understanding is that this string can now be exported to Transifex for translation, can be imported by a project, and the string will appear in another language. (Please correct my understanding of the translation workflow if I am wrong.)

For \# 4, although the string is sent through `tra()`, there is no real way we, as BOINC upstream, can translate elements that appear on a project-specific site.

For \# 1 and \# 3, there is no expectation that the `consent_type`, `description` string appear online, or in a client. The string is really for admins and not end-users.

For example: the first consent type, ‘ENROLL’, is for general consent to a site’s terms of use. And the rest of the PR is the front-end Web code that uses this: but it never shows the name or the description anywhere. Instead, all of the Web content can and should be translate-able, but the information about this consent type in the database can remain English-only.

This is part of the reason for using the database-a heavyweight construct: to have a backend that all projects can use with maximum flexibility. And then front-end code that is separately programmed so that it can be translated.

drshawnkwang:
Also, I modified the database files such that there is a separate `content.sql` file with the initial content of the `consent_type` table. I also added some indices to the new database tables, as I forgot about them, and they are important.

brevilo:
> @drshawnkwang indicated that he expects projects to add custom values to the list of consent types. How will they be able to easily do that without having to keep a fork of the BOINC code if that data is stored in a PHP object?

Exactly. This is a question of using hard-coded values vs changeability. The former requires forking and/or code changes while the latter can be easily administered. There's no point in saving time **once** during development (for BOINC) and then investing **more** time downstream for administration (by the projects). The very definitions of "simple" or "heavyweight" aren't universal, they depend on the context. From a project's point of view going through code changes just to add or change a project-specific value would considered heavyweight and complex while using a simple DB-backed admin form would be simple and easy. 

> Also - since the the content types are clearly foreign keys for items the content table, having their meaning and the relationship in the database is very natural and a standard practice. I think this is a very reasonable approach.

Thank you, yes, it's standard design for good reason.

>  Instead, all of the Web content can and should be translate-able, but the information about this consent type in the database can remain English-only. [...] have a backend that all projects can use with maximum flexibility. And then front-end code that is separately programmed so that it can be translated.

While that's true in principle I beg to differ regarding the privacy consent types. IMHO, a frontend should typically be as dynamic as its backend provides for. Thus, as we *devised* a defined list of privacy types we should list them dynamically. That would mean, for instance, to show their `description` next to a checkbox. What would be the point of having an admin maintain such a list if you'd still have to add the values to the frontend by some other means?

Thanks

davidpanderson:
Here's how to do it in PHP (this is the entire code):

define('CONSENT_TYPE_ENROLL', 1);
define('CONSENT_TYPE_STATS_EXPORT', 2);

$enroll = new StdClass;
$enroll->shortname = 'ENROLL';
$enroll->description = tra('General terms-of-use for this BOINC project');
$enroll->enabled = false;
$enroll->protected = true;
$enroll->privacy_pref = false;

$stats = new StdClass;
$stats->shortname = 'STATS_EXPORT';
$stats->description = tra('Do you consent to exporting your data to BOINC
statistics aggregation Web sites?');
$stats->enabled = false;
$stats->protected = true;
$stats->privacy_pref = false;

$consent_types = array();
$consent_types[CONSENT_TYPE_ENROLL] = $enroll;
$consent_types[CONSENT_TYPE_STATS_EXPORT] = $stats;

if (function_exists('project_consent_types')) {
    $consent_types = project_consent_types($consent_types);
}

If a project wants to add or modify types, they'd add a function like this
in their project.inc:

function project_consent_types($consent_types) {
    // add new type
    define('CONSENT_TYPE_MY', 100);
    $my_consent = new StdClass;
    $my_consent->shortname = 'MY_TYPE';
    $my_consent->description = tra('Project-specific consent');
    $my_consent->enabled = false;
    $my_consent->protected = true;
    $my_consent->privacy_pref = false;
    $consent_types[CONSENT_TYPE_MY] = $my_consent;

    // change existing type
    $consent_types[CONSENT_TYPE_ENROLL]->enabled = true;
    return $consent_types;
}

brevilo:
@drshawnkwang just noticed some flaws in the DB design:

* the `consent_type` primary key should just be called `id`
* the `consent` table shouldn't reference the `consent_type` by `id` (PK/FK relationship), not `consent_name` which causes normalization and efficiency issues
** remember to rename and change the `userid_cname` index accordingly

brevilo:
@davidpanderson Lines of code isn't the only metric here. Please consider the following:

* The `consent` table references the `consent_type` by ID (when PR got fixed). If you lack the consent type as a DB entity you can't write queries outside of PHP and use the human readable `shortname`. That has always been frustrating about the `result` (eg. `server_state`) and `workunit` (eg. `assimilate_state`) tables for instance as it makes query comprehension and maintenance harder.
* One example for the above would be `db_dump` which doesn't have access to `project.inc` AFAIK and you don't want to maintain a redundant list in some C header.
* Drupal could make use of its built-in permission system to implement consent-based restrictions that could be fully configured via its admin web interface. Requiring `project.inc` changes would just be an extra step that shouldn't be necessary. Again, this should be making the life of the admin easier and not save some lines of code for the developer.
* Even stock BOINC has a number of DB entities that can be maintained conveniently via web forms (the OPS pages). Why should we do it differently in this case?

drshawnkwang:
The parties discussed this PR at the end of the BOINC Contributor call. A brief summary:

David feels he wants to keep things less complicated from the upstream, i.e. developer, side. By putting the code in PHP, there is less overhead in the BOINC code. David prefers to not use the project database unless it is absolutely necessary.

We (Shawn and Oliver) want a more complicated database, with a design that we feel makes things less complicated from the downstream (project) side. 

Oliver also pointed out that if the consent_types are stored in the code, it must be stored in two different places with duplication: PHP (project.inc) and in the C headers, for db_dump.

At the end of the call, we wound up ‘agreeing to disagree’. People had strong opinions about this issue and at the end of the meeting the situation is that David is not convinced that the complexity introduced by this PR is justified, and prefers it not be merged.

As the result the paths forward are:

1. David removes his veto.
2. The developers asking for the PR, reprogram the code - using a PHP architecture (and C headers).
3. Appeal to Project Management Committee (PMC).
4. “Gut” the PR - Effectively remove the parts of the code that deal with consent_type. What would remain would be the consent table only, and a few things that go with it.
5. E@H goes its own way and ‘forks’. - We would wind up implementing some code for ourselves and not contribute it to BOINC.

David did provide a sixth alternative but will post some code in the PR to explain his alternative: I will admit I did not fully understand his idea at the time of the call.

davidpanderson:
consent type abstraction

PHP code

// return list of consent type objects
//
function get_consent_types()
	if (function_exists('project_get_consent_types')) {
		project_get_consent_types();
	} else {
		return PHP structure as in previous comment
	}
}

E@h would add project_get_consent_types(), which enumerates types from DB,
in its project.inc or Drupal equivalent

C++ code

in db_dump.cpp:

int export_consent_id;
#ifdef DB_CONSENT_TYPES
	export_consent_id = custom_get_export_consent_id();
		// define this function in your sched_customize.cpp;
		// enumerate consent types from DB, return ID of stats export type
#else
	export_consent_id = STATS_EXPORT_CONSENT_ID;
		// defined in common_defs.h
#endif

E@h would build its db_dump with -DDB_CONSENT_TYPES

brevilo:
Thanks Shawn for the summary. I'd prefer "complex" over "complicated" as it sounds less subjective to me. 

I'd also like to understand the "20-30 times higher complexity" figure stated during the meeting. As the veto seemingly hinges on this assertion I'd like to know which metric was used for its calculation.

TheAspens:
**Part I**
Since the meeting I've been trying to come up with a better framework for structuring contested discussions about design decisions.  Some of my thinking was the following:

Application code should be developed to minimally meet known and real use-cases.  Reason:  Unnecessary application complexity arises when you build things based on something you think you might need but don't actually have a real use case for it now.
Application code should consider possible future use cases and be implemented such that it doesn't unnecessarily inhibit reasonable future possibilities.  Reason:  Although code should be minimal and simple, it should also not be designed to block future flexibility.

The philosophy needs to be bigger than those two points, but it does cause me to ask a couple of questions based on our discussion.

Are there actually any known additional consent types that are needed?  I.e we have two consent types with known and real use cases.  Are there any others that are known and real?
If a new consent type does become known, then what is the chance that it could be known but not require changes to the BOINC code?  I.e. what is the likelihood that a project will need a new consent type that isn't something that should be extended to all boinc projects?
What makes a PHP include more difficult to use in drupal vs a database table?

If there are no other known consent use cases, and if we go with the philosophy of minimal application code to satisfy the two consent types that are defined now, then what would change with the existing implementation?

**Part II**
I've also been thinking how to characterize the right time to use database tables vs application constants.  I think that one thing to consider is that database tables should definitely be used when adding or removing a 'type' (or row) does not require changes to application logic.  For example, BOINC is designed so that app, app_version and platform are all designed so that you can add and remove rows/types from it to change the behavior of the system, but it doesn't require any corresponding changes to application logic.   This allows the system to respond without a change to application code.

For things where adding a new type does require a change to application code, then if we use a database table then we are effectively creating a lookup or reference table.  However, it is illusory that it gives us more flexibility in the behavior in the system because we would have to change the code to actually add a new type.  At this point the use of a database table is helpful for providing meaning to "magic constants" but it doesn't actually provide any additional flexibility for the application. 

What is the scenario for consent type?  If you remove a consent type, does application logic need to be removed?  If you add a consent type, does application logic need to be added?

Does thinking through and answering these questions change anyone's mind about the right way to implement consent types?

brevilo:
@TheAspens 

Part 1: 
* I agree with both your statements on feature design / complexity guidelines
* Other real consent types: the whole consent feature is a requirement that originated in the GDPR. So the question is, how are projects *interpreting* the GDPR in terms of getting *their* lawful basis (Art. 6 GDPR) for *their* type of data processing? This highly depends on the projects and their legal advisors - as we saw during our own deliberations trying to revise our privacy policy to be GDPR-compliant. That's the reason for why we built in the kind of flexibility that's provided for now, and that's also the reason why we didn't include those other types in the upstream defaults - from our *experience* they are likely to be project-specific. So there is a potential, yet very real need to ask the user for consent for each of the community features that are not required for simple account management, pursuant to GDPR Art 17.2 and 17.4. Those would be:
   * Private messages
   * Discussion forums
   * Discussion subscriptions
   * Friendships
   * Team membership
   * Public profile
* Chance of no code changes for new types: 0 with the alternative design as, by definition, it would require them to be added to at least two files (for PHP and C consumers) and require to recompile and deploy dependent C tools. With our solution you'd just add them. The likelihood that new types could be used across projects isn't something I can't answer seriously since, as outlined above, that depends on legal interpretations that are known to stray widely. I know that WCG and E@H care more about the legal aspects as other projects and even if others do care, they might come to different legal conclusions.
* PHP code is a static read-only "data store" for instance. Editing the types would mean dropping out of a simple front-end supported admin task and be compelled to change code (directly) on a server which you have to log in via the command line, change the code (or deploy it by other means) in *multiple* places that you therefor have to manually keep consistent. Also, as it's not just PHP, you have to recompile and deploy any dependent C binaries as well. That is, there are a number of needless extra tasks that increase the burden on the project admin.
* "If there are no other known consent use cases [...] what would change with the existing implementation?" Despite the "if" not being satisfied anyway (see above), I share my point of view: it wouldn't change at all because a) I'm convinced of the flexibility requirement, if not for us then for other/future GDPR-conscious projects, and b) I consider the proposed alternative bad design for reasons I already laid out earlier - and that doesn't change with the abstraction layer.

Part 2:
We're honest people so we also had this discussion during our design sessions, and I agree again with your overall philosophy and approach. Our design allows for the following:
* add/change consent types in the Drupal BOINC admin section
* dynamically change the Drupal front-end (user privacy preferences)
* map content types to Drupal roles to enable effective access control (no feature access without consent). Removing a consent type just removes any dependent role mappings.
* dynamically administer our web-frontend *in*creasingly from within itself

tl;dr: there's no need to change any code or compile and deploy any binaries.

Our philosophy here is to design a smart solution *once* with the *minimum yet useful* complexity and have its *benefits multiply* at each project using it. "Saving" complexity *once* in BOINC core just pushes it down to *multiple* projects, and I can honestly say that from years of experience. The latter is never going to be an approach that would allow to guide my design philosophy.

This is also the reason why I, for one, can't accept the abstraction layer either. Yes, it would allow us to do what *we* want, but we could simply do that in our fork anyway (as has been usual). Yet the very example above states "E@H" as if we're the only project seeing the benefits of a *consistent* DB design. In yesterday's call it became apparent that at least WCG, CERN and GridRepublic/CharityEngine do favor the current DB design (please correct me if I'm wrong) as well. And it doesn't end there: the alternative does not only fail to deliver the same benefits, I even argue that it adds to BOINC's technical debt (eg. redundant code and an incomplete DB model). Thus I just can't sign-off on the alternative, if not for us (E@H) then for the benefit of other projects - and even BOINC itself.

TheAspens:
Thanks for responding to my "thinking out loud" post.  After posting it and thinking overnight, it helped me to see that we are simply having the "ENUM vs Lookup Table" discussion that has been had thousands of times previously.   I found a large number of articles and stackexchange articles about it.  I'm sharing a few here for reference:

- https://softwareengineering.stackexchange.com/questions/305148/why-would-you-store-an-enum-in-db
- https://www.red-gate.com/simple-talk/sql/database-administration/five-simple-database-design-errors-you-should-avoid/ (see 4 - Application Encroachments into DB design)
- https://softwareengineering.stackexchange.com/questions/287433/why-should-i-use-lookup-tables-at-all
- https://softwareengineering.stackexchange.com/questions/298472/is-it-wasteful-to-create-a-new-database-table-instead-of-using-enum-data-type

Additionally, we have multiple applications that are using the single database:
- The PHP website code (PHP)
- The BOINC daemons (C/C++)
- A reporting tool (Java)
- Monitors to see if we are low on work (Nagios/Shell Scripting)
- Work management utilities that move work into and out of BOINC  (Bash/Python/C++)

Each of these applications has to have logic embedded to interpret various constants (server state, validate state, outcome, file delete state, etc).  This violates the DRY principle and adds complexity to the applications.

All of this speaks to putting the data into the database and not using application constants for "magic numbers".

brevilo:
> we should be adding foreign key relationships to enforce those

Hell, yes, please! Something I argued for all these years (alongside the DB abstraction layer to support other RDBMS than MySQL). I'll buy you a 🍺 when we get these integrated!

AenBleidd:
@brevilo, I'm afraid, I glass of beer will be not enough in this case, only smth like this?
image

brevilo:
Nice! You have a point there. I'll see what I can do!

drshawnkwang:
@TheAspens , To add to your point in your latest comment, this is why Oliver pointed out a design flaw in this PR w.r.t. me using the consent_name in the `consent` table to identify the consent types. And how it should be `consent_type_id`. Then one uses a lookup to determine what `consent_type_id` means in the `consent_type` table, which has ids, a name, and description. I will refactor the database code and commit it to this PR soon. It should affect our previous discussions about this PR.

davidpanderson:
A comparison of consent type representations is here:
https://boinc.berkeley.edu/consent.html

JuhaSointusalo:
I know this issue is in the hands of the PMC now. As if I would let that stop me from expressing my opinions on the matter...

> A comparison of consent type representations is here:
> https://boinc.berkeley.edu/consent.html

In my opinion your comparison has multiple flaws.

> consent.h (for C++ code)
> ```
> #define CONSENT_TYPE_ENROLL         1
> #define CONSENT_TYPE_STATS_EXPORT   2
> ```
> That's it! We're done.

Not true. Consent type has more properties than its ID. You'll also need to duplicate it all for all the languages used. Java, Python and shell scripts have been brought up in this thread.

> Change/add consent types or fields
> Edit consent.inc with a text editor. I like vi.

@drshawnkwang could have said
> Install phpmyadmin and use that edit consent types.

Instead he chose to add a form to allow editing the types for convenience. A form nicely takes care of typos and editing errors when duplicating the consent type definitions (not an issue for @drshawnkwang 's proposal but it is an issue for your proposal). If you want a fair comparison add such form to your proposal or ignore the form @drshawnkwang added.

> Redundancy
> The two lines in consent.h duplicate the values in consent.inc.

Except two lines is an incomplete implementation.

> No problem, since these values never change.

Not true. The purpose of `consent_type.enabled` is that it can be changed.

> content.sql
> db_update.php
> Note: projects that want to update web code without updating server code would have to do this update manually.

Some updates require updating all of server code. Why is that such a huge problem? And quite frankly, I'm not convinced that allowing projects to mix and match different server component versions is a good idea at all.

> We also need a C++ interface to the new table.

And your proposal needs C++ code as well. This C++ code needs to added only because in BOINC database definitions are duplicated in C++ and to some extent in PHP and Python. The code could have been auto-generated from the database schema. I find it questionable that you complain about @drshawnkwang following design decisions made by past BOINC developers long time ago.

> db_dump.cpp
> This will have to do a database lookup to get the ID of the export preference.

If you need to join data from multiple tables then you JOIN the tables. Relational databases have been designed to manage and process your data. Just let them do their job.

> Changing/adding fields; redundancy
> If you want to change fields (add/rename/reorder) you'll have to change 7 files.

Your proposal is no different.

In my opinion the only valid point you have is that the descriptions are not translated. Although, I can't fully agree with even that one point.

> Translation
> The consent descriptions are not translatable. 
> Making them translatable would require a significant addition to the translation mechanism,

It shouldn't be too hard to extract the strings from `content.sql` and add to server template. Any consent types projects add wouldn't be translated this way but it doesn't matter. Either projects don't have their own content translated or they don't use BOINC translation service for it. If you must have project added consent types translated nothing stops you from querying the database for the strings.

JuhaSointusalo:
I don't know if @davidpanderson's abstraction layer is still on the table in any serious manner. If it is here's some opinions of that as well.

Sometimes a new operating system version, or some other software you need, has a new feature you want to use. But at the same time you need to support older versions too so you need to have fallback code for the older versions. Or, there is something you need to be done but the different systems you support have different ideas of how that something is done so you need different code for different systems.

So there can be a technical need to have different implementations of something.

In this case however, there is no technical need for different implementations. There is only a difference of opinion of what an implementation detail should look like. Furthermore. any updates or changes would have to be duplicated to both implementations. I believe having different implementations here just results in future developers asking "what is this" and rolling their eyes.

So in my opinion, @davidpanderson's abstraction layer is the worst of the three choices.

drshawnkwang:
This PR has not been updated lately because I initiated an appeal to the BOINC PMC of a committer's  veto. The BOINC appeal hearing was held on August 22 2018, and a decision was announced on August 29 2018.

The decision was made in favor of the appeal to overrule the veto. However, it was also discussed that this PR is not yet ready to be merged. A number of technical items must be addressed.

1. Implement a mechanism so that content strings stored in the `content_type` database table are included in the translation templates.
1. Implement the necessary database interface logic in the C++ later to support the use of the new table.
1. Implement support for the STATSEXPORT consent_type in the db_dump utility.
1. Any other items that may arise during review of this PR.

drshawnkwang:
The latest commits (2018-08-29) to this PR includes some C++ code that adds access to the new `content_type` table. Additionally, the table `content_type` has a field renamed, `protected` is now `protectedct`. (The former is a reserve word in C++.)

brevilo:
> protected is now protectedct

To be honest, that's awkward and looks like a typo. How about simply "protect" instead?

drshawnkwang:
The latest commits will require a some conflict resolution. But in the meantime I wanted to ask the following MYSQL question.

I modified sched/db_dump to use a SQL statement when selecting users (and hosts). I have reproduced it below but prettier.

```
SELECT     * 
FROM       user 
INNER JOIN 
    ( 
    SELECT     b.userid 
    FROM       consent AS b 
    INNER JOIN 
        ( 
        SELECT   userid, 
            Max( consent_time) AS max_consent_time 
        FROM     consent 
        GROUP BY userid 
        ) AS a 
    ON         b.userid=a.userid 
    AND        b.consent_time=a.max_consent_time 
    WHERE      b.consent_flag=1 
    AND        b.consent_type_id=2 
    ) AS c 
ON         user.id=c.userid WHERE total_credit > 0 [ORDER BY ...]
```

I am wondering if there will be a performance hit with this more involved statement, and if we can do anything to optimize for this query. Since we have the ability to add indexes as we wish to `constraints.sql`, what could be added?

brevilo:
A few first quick comments (a deeper analysis should follow):
* Don't just `SELECT *` when you only need certain attributes. I know that this is a BOINC standard, but it's a bad one in terms of I/O efficiency, in particular with the large records in `user`, `workunit` and `result` tables.
* If you use aliases like `a`, `b` or ,`c` please don't just pick those randomly, give them context/meaning (if possible) as it helps a lot analyzing a complex statement.
* Whenever you're unsure about a statements efficiency, run it with [`EXPLAIN EXTENDED`](https://dev.mysql.com/doc/refman/5.7/en/execution-plan-information.html) to see what the optimizer does. Focus on your `WHERE` clauses and ask which indexes support those. `EXPLAIN` will show you. Add missing indexes and verify their use accordingly.

brevilo:
Hm, does that SQL statement really do what you're looking for? You're selecting the latest consent entry for a given user, irrespective of its consent type, and then check whether that single consent entry matches the type in question and equals `true`. What you want to do instead is to select the latest entry for the consent type in question and make sure it's `true`. Or am I missing something?

brevilo:
Here's a statement that could serve as a view (e.g. `latest_consent`) showing each latest consent entry per user/type:
```
SELECT id,
       userid,
       consent_type_id,
       consent_time,
       consent_flag
  FROM consent
 WHERE NOT EXISTS
       (SELECT *
          FROM consent AS filter
         WHERE consent.userid = filter.userid
           AND consent.consent_type_id = filter.consent_type_id
           AND filter.consent_time > consent.consent_time);
```
Ideally that statement would be supported by a compound index on `consent(userid,consent_type_id,consent_time)` which could simply replace the existing one on `consent(userid,consent_type_id)` (via prefix matching).

You could then join the user table against a subselect on that view, filtering for `consent_type_id=2 AND consent_flag=1`, returning the `userid` only:

```
    SELECT *
      FROM user
INNER JOIN
          (SELECT userid
             FROM latest_consent
            WHERE consent_type_id=2
              AND consent_flag=1) latest
        ON user.id=latest.userid;
```

Hope this helps

brevilo:
Now that we're trying to embrace the DB, we should really let it do its job proper: please consider adding a foreign key constraint to the `consent` table:

```
ALTER TABLE consent
ADD FOREIGN KEY(consent_type_id)
REFERENCES consent_type(id)
ON UPDATE CASCADE;
```

This prevents `consent_type` deletions of in-use types and will ensure consistent `consent_type.id` updates across tables.

drshawnkwang:
Re: `SELECT *` - I am effectively changing the original query to add the inner joins, thus I can't select the specific columns/fields neede in the user/host table.

Re: SQL query

You are correct. Thanks for catching this. The query did not correctly obtain the users whose have consented to the stats. export consent_type_id.

After reading through your comments I changed the query to a variation on one of yours:
```
SELECT    *
FROM      user
INNER JOIN
   (
   SELECT      consent.id,
               consent.userid,
               consent.consent_type_id,
               consent.consent_time,
               consent.consent_flag
   FROM        consent
   LEFT JOIN   consent AS filter 
       ON          consent.userid=filter.userid
       AND         consent.consent_type_id = filter.consent_type_id
       AND         filter.consent_time > consent.consent_time
   WHERE       filter.userid IS NULL
   AND         consent.consent_flag=1
   AND         consent.consent_type_id=2
   )
AS latest
ON user.id = latest.userid;
```

Additionally, I added an additional index to the `consent` table and the foreign key constraint

Re: SQL Views

I like the idea of using the derived SQL Views as you described. However, I think it is outside the scope of this PR, which has grown over time, to add this. I think adding Views functionality requires development of the C++ database backend. And I don't believe I am the right person to do this right now. (Additionally, my C++ is very rusty.)

Perahps we could consider the SQL Views idea as an improvment to be tackled next?

drshawnkwang:
I have merged the `master` branch into this development branch to handle some conflicts to db_upgrade. Additional changes include:

1. @ChristianBeer has provided code to handle translations from the database.
1. `db_dump` has been modified with a better SQL statement (that works!).

brevilo:
> After reading through your comments I changed the query to a variation on one of yours:

Why did you (re)add `consent.id` and `consent.consent_time` to the `SELECT` clause? They're not needed afterwards and just cause unnecessary I/O.

> I like the idea of using the derived SQL Views as you described. However, I think it is outside the scope of this PR, which has grown over time, to add this. I think adding Views functionality requires development of the C++ database backend

I don't think I agree with this assessment. One could argue that factoring out the view provides opportunities and that doesn't necessarily require a C++ API. The view can be used purely from within the DB, even just for ad-hoc admin queries, and will reduce complexity by splitting things up in a reasonable way - particularly important when statements are written in single-line form as are right now (I recommend to change that!). Having the full statement as it is right now really just tailors it for one purpose and I don't see why we should impose such a limit on it.

Regarding the indexes you chose: 
* Did you check the effectiveness of the `flag_ctid(consent_flag, consent_type_id)` index?
* Why didn't you follow the idea to use `consent(userid,consent_type_id,consent_time)` instead of just `consent(userid,consent_type_id)`?

Thanks

drshawnkwang:
Oliver and I discussed a a few things offline. That helped me understand what Oliver is talking about and I will try to summarize how that relates to the latest commits to this PR.

The SQL View doesn't have to be defined in the C++ API (as I thought it did). In fact a view (`latest_consent`) is now defined in `schema.sql` and `db_upgrade.php`. The purpose of this view is to be the latest consent status of all users and consent_types. As a result, it is defined in the DB schema because it can be used in multiple places:
  * `db_dump`: The database dump needs to know what users have consented to stats exports, thus the SQL query is now:
```
    SELECT *
      FROM user
INNER JOIN
          (SELECT userid
             FROM latest_consent
            WHERE consent_type_id = 2
              AND consent_flag = 1)
        AS lc
        ON user.id = lc.userid;
```
where `latest_consent` is the SQL View defined above (Oliver's first definition).
* BOINC Web site: needs to know the latest consent status of users when the login and when they update their project preferences. Now this SQL View is used in that code to determine if a user has consented to a specific policy.

Because there are multiple places where code needs to know what the user's latest consent status is, it makes sense to have a single point of reference where this attribute may be checked: and that is this `latest_consent` View.

Lastly, I forgot to add `consent(userid,consent_type_id,consent_time)` to the index of the consent table's indices: this has been fixed.

db/content.sql
```diff
@@ -0,0 +1,9 @@
+/*
+    This file contants inital content for any tables.
+
+    The table must be first defined in schema.sql before any content
+    is added!
+*/
+insert into consent_type (id, shortname, description, enabled, protect, privacypref)
+    values (1, 'ENROLL', 'General terms-of-use for this BOINC project.', 0, 1, 0),
+    (2, 'STATSEXPORT', 'Do you consent to exporting your data to BOINC statistics aggregation Web sites?', 0, 1, 1);
```
TheAspens: Since consent_type.id is auto-increment and you are intentionally not tying specific values of consent_type.id to behavior in the application, then you should not specify the value for the id on the insert.  

```
insert into consent_type (shortname, description, enabled, protect, privacypref)
    values ('ENROLL', 'General terms-of-use for this BOINC project.', 0, 1, 0),
    ('STATSEXPORT', 'Do you consent to exporting your data to BOINC statistics aggregation Web sites?', 0, 1, 1);
```

db/schema.sql
```diff
@@ -804,3 +805,38 @@ create table host_deleted (
...
+    source                  varchar(255)    not null,
+    primary key (id)
+) engine=InnoDB;
+
+create table consent_type (
+    id                      integer         not null auto_increment,
+    shortname               varchar(255)    not null,
+    description             varchar(255)    not null,
+    enabled                 integer         not null,
+    protect                 integer         not null,
```
TheAspens: protect describes the how the application uses this field.  In your design document you describe the field as "project specific".  Would it be better to name it project_specific rather than 'protect'?   That makes its meaning clear without having to resort to the design document.

html/inc/consent.inc
```diff
@@ -0,0 +1,71 @@
...
+    $mys = BoincDb::escape_string($source);
+    if ($ctime==0) {
+        $mytime = $user->create_time;
+    }
+    else {
+        $mytime = $ctime;
+    }
+    return BoincConsent::insert(
+        "(id, userid, consent_type_id, consent_time, consent_flag, consent_not_required, source) " .
+        "values(0, $user->id, $consent_type_id, $mytime, $consent_flag, $consent_not_required, '$mys')"
```
TheAspens: With an auto-increment value for consent.id - shouldn't it simply not have the field on the insert?  I.e.
```
 "(userid, consent_type_id, consent_time, consent_flag, consent_not_required, source) " .
 "values($user->id, $consent_type_id, $mytime, $consent_flag, $consent_not_required, '$mys')"
```

html/inc/consent.inc
```diff
@@ -0,0 +1,71 @@
...
+    );
+
+}
+
+// Checks to see if a user has consented to specfic consent_type_id.
+function check_user_consent($user, $consent_name) {
+    list($checkct, $ctid) = check_consent_type($consent_name);
+    if ($checkct) {
+        $consent_result = BoincLatestConsent::lookup("userid={$user->id} AND consent_type_id=$ctid");
+        if ( ($consent_result) and ($consent_result->consent_flag) ) {
```
TheAspens: wouldn't you add the check for consent_flag to the lookup query?

html/ops/index.php
```diff
@@ -106,6 +106,7 @@
     <ul>
         <li><a href=\"manage_apps.php\">Manage applications</a></li>
         <li><a href=\"manage_app_versions.php\">Manage application versions</a></li>
+        <li><a href=\"manage_consent_types.php\">Manage Consent types</a></li>
```
TheAspens: can you make the C in Consent lower case to match the other ones in this list?

TheAspens:
Some additional comments:

1) We need to be able to edit an existing consent type that isn't locked.  If you delete and recreate it changes to the consent_type_id changes and I can see projects needed to re-word the description after creation.
2) I can't figure out how to get terms of service to show up on the registration page nor can I figure how how to see the consent page from the user home page.  How do I trigger those to show up?

drshawnkwang:
> I can't figure out how to get terms of service to show up on the registration page nor can I figure how how to see the consent page from the user home page. How do I trigger those to show up?

1. In manage consent types (from main OPs page), you must enable the `ENROLL` consent type.
1. In config.xml, option `enable_login_mustagree_termsofuse` must be TRUE (1).

I admit this is clunky. I welcome suggestions on how to make it more intuitive. One reason for the dual set of booleans: if you enable the `ENROLL` consent type, but `enable_login_mustagree_termsofuse` is false, then new only users see the Terms of use page. This may be a desirable setting for some.

TheAspens:
thanks.  I finally figured it out.  I was testing multiple things to try to get it to work and got confused.  In the end I had put my terms_of_use.txt file in the wrong location and once I fixed that it worked.

db/content.sql
drshawnkwang: Fixed in 4cb30b1

db/schema.sql
drshawnkwang: Fixed in b44a8c8 and 992cece

html/inc/account.inc
```diff
@@ -114,6 +116,23 @@ function create_account_form($teamid, $next_url) {
             "postal_code"
         );
     }
+
+    // Add terms of use to Web form. User must agree by checking the checkbox.
+    list($checkct, $ctid) = check_consent_type('ENROLL');
```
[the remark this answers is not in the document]
drshawnkwang: Fixed in 90bdd91

html/inc/account.inc
```diff
@@ -114,6 +116,23 @@ function create_account_form($teamid, $next_url) {
...
+    if ($checkct and check_termsofuse()) {
+        $terms_of_use = trim(file_get_contents(TERMSOFUSE_FILE));
+        if ($terms_of_use) {
+            panel('Terms of Use', function() use($terms_of_use) {
```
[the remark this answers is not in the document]
drshawnkwang: Yes it does. Fixed in e4df047.

html/inc/consent.inc
drshawnkwang: Fixed in 4cb30b1

html/inc/consent.inc
drshawnkwang: Fixed in a02bc83.

html/ops/index.php
drshawnkwang: Fixed in e4df047

html/user/login_action.php
```diff
@@ -136,8 +165,16 @@ function login_with_ldap($uid, $passwd, $next_url, $perm) {
...
     }
+    // Intercept next_url if consent has not yet been given
+    list($checkct, $ctid) = check_consent_type('ENROLL');
+    $config = get_config();
+    if (parse_bool($config, "enable_login_mustagree_termsofuse") and $checkct and check_termsofuse() and (!check_user_consent($user, 'ENROLL'))) {
+        $next_url = consent_after_login($user, $perm, $next_url);
+    }
+    else {
+        send_cookie('auth', $user->authenticator, $perm);
+    }
```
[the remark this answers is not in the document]
drshawnkwang: Fixed in 1381601.

html/inc/consent.inc
drshawnkwang: This has been fixed.

html/ops/manage_consent_types.php
drshawnkwang: Fixed with redesign of `consent_type` table

drshawnkwang:
I believe this PR is coming to its end soon. I made a number of changes that deal Kevin's review. 

I also added logic to the create_account RPC: BOINC client greater and equal to v6.10.0 are presumed to present the user with the project's terms-of-use, and thus assume that consent has been granted. Because in order to continue to create an account, a user must click through the terms-of-use.

Additionally, projects should turn off all buy Web registration for new users in order to fully-comply with GDPR. I created a new [GDPR document](http://boinc.berkeley.edu/trac/wiki/GdprCompliance) in the BOINC wiki that helps describe some of this.

html/user/create_account.php
```diff
@@ -89,6 +93,47 @@
...
+        // xml_error(-1, ...).
+
+        // If BOINC client version is larger than the minimum defined,
+        // then assume consent has been given, because the user must
+        // have clicked through the terms-of-use dialog box.
+        $client_version = boinc_client_version();
+        if ($client_version >= MIN_BOINCCLIENT_VERSION_TOU) {
+            $consent_flag=1;
+            $source='BoincClient';
+        }
```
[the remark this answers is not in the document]
TheAspens: The effect is that the same information is recorded in the database whether or not the end user had the opportunity to accept a TOU on the client.

html/user/prefs_edit.php
```diff
@@ -78,6 +77,7 @@
             } else {
                 $main_prefs = $prefs;
                 prefs_privacy_parse_form($user);
+                prefs_consent_parse_update($user);
             }
 
             project_prefs_update($user, $main_prefs);
```
TheAspens: I'm not certain what the source of the issue is, but I created a new account via the client and went to enable my stats export through the project preferences.  When I submitted it I got the message:

```
Notice: Undefined index: default_venue in /root/project/html/inc/prefs.inc on line 689

Warning: Cannot modify header information - headers already sent by (output started at /root/project/html/inc/prefs.inc:689) in /root/project/html/user/prefs_edit.php on line 89
Back to preferences 
```

TheAspens: I'm blocked in further testing due to this issue.  Let me know when fixed and I will resume testing.

drshawnkwang: Quick question:
>  I created a new account via the client and went to enable my stats export through the project preferences.

I assume you mean you went to the project's Web site, went to the project preferences page, and attempted to save/submit the changes? 

I will try to debug this later today and get back to you.

TheAspens: Yes.  Complete path.  Create account either via client or via website.  Go to project preferences and edit preferences on default profile.  Attempt to save settings.  Get error.  Thanks.

drshawnkwang: Fixed in https://github.com/BOINC/boinc/pull/2504/commits/683c5629bc566c2ee79ba6b0f79a6399155b799e, I removed a necessary line accidentally in a development mistake. Reverting the line should solve the bug.

html/user/create_account.php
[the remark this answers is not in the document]
drshawnkwang: Fixed in https://github.com/BOINC/boinc/pull/2504/commits/e37e5999a93cdf46277df8dc63e97841a6c411a7, create_account RPC checks the existence of terms_of_use.txt file

drshawnkwang:
Merged master into this development branch to merge in #2727.

TheAspens:
@drshawnkwang - you appear to have merged it into your forks master branch but not into the branch for this pull request.

TheAspens:
I am going to leave this unmerged for a couple of days to give others a chance to review and comment as well.  @lfield , @tristanolive , @JuhaSointusalo  , @AenBleidd  or anyone else want to take a look?  Thanks.

TheAspens:
@brevilo - if you wanted to do a final review as well that would be good also.  I know you were involved and advising on the development, but a final check is helpful.

drshawnkwang:
I believe I did not properly `git push origin`, which I have done so. I believe this branch is now has the new development code as well as been merged into master (from a few days ago) and has no conflicts.

Sorry about the confusion.

JuhaSointusalo:
> I also added logic to the create_account RPC: BOINC client greater and equal to v6.10.0 are presumed to present the user with the project's terms-of-use, and thus assume that consent has been granted. Because in order to continue to create an account, a user must click through the terms-of-use.

One gotcha with that. BOINC Manager is not the only (G)UI there is. Three others that I know of are BoincView, BoincTasks and boinctui.

Of those, BoincView can't add new projects. It complains about some error and nothing happens on the client side. I don't know if adding projects is just buggy or is BoincView incompatible with newer clients. I didn't check older clients.

BoincTasks and boinctui can add new projects but neither shows terms of use.

I don't know if that requires changes to this PR. Just letting you know about those other (G)UIs.

RichardHaselgrove:
I also run BoincView - final release version 1.4.2, copyright 2006 (although my download is dated March 2007 - I think there was an unsuccessful Beta in between). It seems to use its own project list, which in my case is also dated 2007. I'll try to find one which is still using the same url, 

Later - Rosetta qualifies, and attachment (new account) completed successfully - download of first task confirmed on standard BOINC Manager. But Rosetta probably doesn't have any of the new opt-in code: BV certainly didn't show me any terms of use.

drshawnkwang:
@JuhaSointusalo , thanks for the information. It may change our thinking. I will have to sit on it for a bit.

@RichardHaselgrove , albertathome.org definitely has a terms-of-use it sends through `get_project_config` if you want to test things.

TheAspens:
With the comments above, I wanted to point out the following.  
- GDPR compliance can only be ensured if rpc create account is disabled. 
 <disable_account_creation_rpc>1</disable_account_creation_rpc>.  At some point in the future perhaps this can change, but in the near term this is the only way to do it.
- The code should only create a consent record for the user if they use the BOINC client that supports terms of use OR if they register on the website.  In the other cases, the code allows the registration to continue, but it does not create a consent record for the user.  If the flag <enable_login_mustagree_termsofuse>1</enable_login_mustagree_termsofuse> is turned on, then the user will be forced to approve the terms of use the next time they use the website.

These two flags provide the tools for a project to either ensure strict GDPR compliance or to provide loose GDPR compliance with backwards compatbility.   The create work rpc needs to be thought through in greater detail for the future and this can probably be improved, but I don't think that has to be part of the pull request (for example, we need to think through how to handle this well with account managers).

These settings are documented here: https://boinc.berkeley.edu/trac/wiki/GdprCompliance

RichardHaselgrove:
@drshawnkwang: I've used albertathome before, and I have an existing account. BoincView was able to attach a new host, without showing me the ToU: trying to log into the project website told me that I hadn't agreed to the ToU and said I would have to agree to proceed (I backed out).

drshawnkwang:
> The code should only create a consent record for the user if they use the BOINC client that supports terms of use.

The problem may be: if there are BOINC clients that do not support terms-of-use, such as BoincView, but we (the code) assumes that they do in fact see/support the terms-of-use, then we will mark the user as 'consent_flag=TRUE' when in fact they have never seen anything.

Options
1. Ideally we would need need a way to distinguish different BOINC clients.
1. Or we 'revert' to a slightly older code where create_account never sets the consent_flag to TRUE, and instead the Web page is the **only** way to give consent to the terms-of-use (because it's the only situation where the user is guaranteed to see it).
1. Ignore the problem, because disabling create_account RPC is only way to be fully compliant with GDPR, as mentioned above. And if you don't disable create_account RPC: the percent of users who use these 'alternative' BOINC clients is small.

TheAspens:
@RichardHaselgrove - I think that the TOU will always be presented in the manager when attaching to a project (regardless of it is a create account or attach with existing). 

Also - is albertathome running this version of the software?

RichardHaselgrove:
In that respect, my two test scenarios were distinct: BoincView did **not** present ToU, BOINC Manager 7.14.0 **did** present ToU. But the effect on the albertathome server was the same: it did not register that the ToU had been accepted. I can't speak for the AaH server code version.

davidpanderson:
GDPR does not require disabling remote account creation,
as long as no personal information is associated with the accounts.
That's the approach used by Science United, which absolutely needs to be able to create accounts.
It uses random name/email for these accounts.

We may need to make some other changes, like
- don't store IP address for hosts of such accounts
- don't set show_hosts or similar for such accounts
- some sort of white-list mechanism to ensure that creation requests
  are from known sources

RichardHaselgrove:
I was left a bit unsatisfied by my brief and hurried testing after the call last night, so I've spent some time this morning testing and thinking. These remarks apply to the 'stock' v7.14.0 client/manager combination only - no third-party tools.

I first attempted to track and capture all the RPC data between host and albertathome, but this proved frustrating and ultimately impossible.

1) All the (multiple) RPC connections between host and project during 'attach to project' are allowed to proceed even when networking is disabled. With only a 5 second gap between 'Fetching master file' and 'Project initialisation', it's hard to catch it all.
2) Updating of the Event Log is inhibited while the 'Attach' wizard is active: all messages appear in a rush only after 'Finish' is clicked, and by that time everything has already happened.

Those are for another issue.

Then I did what I should have done at the beginning, and re-read the consent processes discussed at the beginning of this PR and the original GDPR issue. And it's clear that opt-in via client/manager only applies when a new account is being created - so I was testing the wrong thing.

That being said, I was frustrated by being presented with the ToU, accepting them, but no notice being taken of my acceptance. It would be better to re-sequence the attach wizard, so that the choice of 'new or existing' is presented first, and the ToU only appear if 'new' is selected. That might annoy people who fill in their email and choose a new password, then decide that they can't accept the ToU - unless we went to a 3-step process: new/existing, ToU if needed, fill in credentials.

I'll take a breather, then try creating new accounts via both BOINC Manager and BoincView.

RichardHaselgrove:
OK, retested with making new accounts. In both cases, using the v7.14.0 client (current alpha for Windows)

a) Using BoincView as the initiator. Account creation failed, with
```
Creating project account
The account couldn't be verified.
done.
```
Nothing was recorded in the Event Log (even with <http_debug>), and I couldn't find any files on the remote machine: this seems to be an internal message from BoincView, not from the Albert server. I haven't tracked it down yet.

b) Using BOINC Manager v7.14.0. ToU were displayed, and account creation was successful. I was immediately directed to the website to complete personal details, but it was an old web address - never mind, Albert is a test site, you expect things like that.

But logging in on the correct url, I was again told I hadn't accepted ToU, and required to accept them again. The Drupal ToU page is strange:

albert tou
There are bailout links, but you can click the YES link without checking the checkbox. That leads to an error message: it's normal to de-activate the next action if there's a precondition.

drshawnkwang:
Richard, thanks for all your testing. You have a number of questions/points which I hope to address.

* Albert@Home's behavior: You noticed that you were able to read the terms-of-use (TOU), and create an account. The when you visited A@H's Web site it asked you to agree to the TOU again. This is 'correct' behavior for the code that exists on A@H.
  * A@H is not using the code in this PR. Thus creating an account using the `create_account` RPC does not record your consent of the TOU in any database. Therefore you must separately agree to the TOU using the Drupal Web interface.
  * You are correct it would be better UI to only allow the user to click 'Yes' if the checkbox is activated. I will see if there is a way to implement this in the future.

* Client vs. manager: Thanks for clarifying this for me. This is part of the problem you have correctly identified. The **manager** is responsible for displaying the TOU to the user; but the server can only check the **client** version. Thus we don't truly know if the user has seen or agreed to any TOU.

David's concern that about disabling `create_account` RPC is valid. However, I think this is a case where projects will have to decide on how they interpret GDPR and what measures they need to take to comply with it.

I think the follow TheAspen's idea is a good one: adding a `user_agreed_tou` parameter to `create_account` RPC. This allows any BOINC managers going forward to explicitly state, "I have shown the user the terms-of-use they have agreed to it." I plan to add this code change to the PR in the next day or two; and will mark the PR as WIP.

Also, I think we should **remove** code that if the boinc_client version is greater than 6.10 then we assume consent has been given. Because again, if there is no guarantee that a BOINC manager has shown the TOU to a user, we can't simply assume they have consented to anything. I will also point out: nothing I am proposing in this paragraph of the one above will prevent a user from **creating an account** with `create_account`. Only the server will **not record consent**, user agreeing to the TOU, in the database.

This will create a confusing situation that Richard just experienced at A@H: he sees the TOU in the GUI BOINC manager, agrees to it, and then when visiting the Web site, the site will ask the user to again agree to the TOU[1]. I realize this is not ideal from a user-experience PoV. But I think it is less confusing from a project administrator's PoV, and in this case may take precedence over UX . Otherwise there will be a situation where
1. A project can legitimately say 'you consented to the TOU, I see a record of it in our database',
1. And the user can legitimately say 'I never say any TOU when I signed up'.

[1] This will happen with both the bootstrap HTML Web sites well as the Drupal Web site: it is not exclusive to Einstein@Home projects.

[2] And I am admittedly biased.

JuhaSointusalo:
One more thing. When the create account process is started by BOINC Manager (and maybe other GUIs), the account is created with user name that is the same as the user's account on their computer. I would imagine this is private information. Can you store the user name without having the user agree to it? So can you store the user name and then wait for the user to log in on the website to agree to terms?

JuhaSointusalo:
> I think the follow TheAspen's idea is a good one: adding a `user_agreed_tou` parameter to `create_account` RPC.

There is that `consent_flag`. I think it would work like this:
- old Manager or third party GUI that doesn't fully support this consent stuff
  -> `consent_flag` not set
- new GUI, terms showed and user agreed
  -> `consent_flag=1`
- new or older GUI, terms showed and user disagreed
  -> no create account RPC at all

This is only when using GUIs. Account managers can still have `consent_flag=0`.

drshawnkwang:
Juha is correct, the additional parameters `user_agreed_tou` is not required. The `consent_flag` states should have all the necessary information. I will ret-con my earlier long comment. 

Re: the username, #1929, there is an issue I created. I think this is a client side issue because the username is passed to create_account RPC. No one has worked on it though.

html/ops/manage_consent_types.php
```diff
@@ -0,0 +1,192 @@
...
+
+    if (empty($shortname)) {
+        admin_error_page("The new consent type must contain a short name.</font></p>");
+    }
+    if (empty($description)) {
+        admin_error_page("The new consent type must contain a description.</font></p>");
+    }
+
+    BoincConsentType::insert(
+        "(shortname, description, project_specific) VALUES ('$shortname', '$description', 1)"
```
TheAspens: enabled and privacypref are not-null but do not have default values.  Therefore this insert fails.  Either give them default values in the db definition or add them with values in this section of the code.

TheAspens:
While testing I noticed that db_dump works correctly in these cases:

- If STATSEXPORT consent type enabled, then users and hosts that have not consented are not exported
- If STATSEXPORT consent type is disabled, then all users and hosts are exported correctly

However, the file tables.xml should reflect the same value without regards to the STATSEXPORT consent type.  Since this file is an aggregate field, GDPR restrictions are not a concern.  This file will be very important going forward since it will be the only way for sites like BOINCStats to be able to present comparisons of number of users, hosts and total credit between projects.  If data vanishes from this aggregate level file then there will be no way to see that total level at projects.

TheAspens:
Those are the only two items I found.  Once they are resolved, I'll be ready to merge.

html/ops/manage_consent_types.php
drshawnkwang: Fixed in 7d7c187

drshawnkwang:
I changed `db_dump` to have an independent count of the number of users and hosts for a project.

I also merged in master into this branch to resolve merge-conflicts.

TheAspens:
One additional question that I noticed this time testing.  If the terms of use file is present, then the manager will show the terms of use without regard to whether the consent_type 'ENROLL' is enabled or not.  However, the website only shows the terms of use if the consent_type 'ENROLL' is enabled.  I think that it is odd for it to behave differently in different places.

I would suggest that either:

A) Terms of use on a call to get_project_config.php should only be shown if both terms_of_use.txt exists and the ENROLL consent type is enabled
or
B) If terms_of_use.txt exists, show it on the manager and on the website always without regard to the ENROLL consent type.

I think A is more consent with the design of the consent_type system - but I don't have a strong opinion.

drshawnkwang:
Latest changes: db_dump correctly takes into account delete users and hosts.

From Kevin's above comment, 'Option A' implemented. - the get_project_config ignores the terms-of-use file if the ENROLL consent type is disabled

drshawnkwang:
> Latest changes: db_dump correctly takes into account delete users and hosts.

The changes were buggy: this should be fixed with the latest commit: which revamps how the SQL clause is generated.

JuhaSointusalo:
Maybe it would be good to document that after creating an account with old Manager, `enable_login_mustagree_termsofuse=1` takes effect and records are created only after logging out and back in. After adding a project and creating an account at the same the Manager opens `account_finish.php` which logs the user in with authenticator.

drshawnkwang:
Juha, You are correct, the old manager behavior is a bit confusing because it looks like you consented to the terms-of-use, when the database is not changed. I will try to document this in the [BOINC wiki](https://boinc.berkeley.edu/trac/wiki/UserOptInConsent).

Latest commits: we removed the option for admins to delete a consent_type from the `consent_type` table in the HTML OPS pages. This is because of a foreign key constraint in the `consent` table. And there may be legal reasons to keep all these records in the `consent` table. Admins may still disable a consent_type if there is no need for its use.

TheAspens:
I've tested and reviewed and I believe that open items have been resolved (with 156 comments, it is hard to be sure so if I missed something, I apologize).  Thanks for the contribution @drshawnkwang and everyone for their work reviewing and testing this piece of code!

brevilo:
Well done everyone and a big **THANK YOU** to Shawn and Kevin for taking on most of the work!

PR #2505, opened 2018-05-07
Update comments in db/schema.sql

davidpanderson:
Please read these before changing the schema.
Recent schema changes don't follow these rules.
At some point they should be revised.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -27,6 +27,9 @@
 -- Fields should generally have a default
 -- (newer MySQL versions don't have automatic defaults)
 
+-- add new fields to the end of the table
+-- (makes it easier to update C++ code)
+
 -- Engine is specified as InnoDB for most tables;
 -- supposedly this gives better performance.
 -- Some (post, thread, profile) are myISAM because it supports fulltext index
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -1,33 +1,42 @@
-/*  If you add/change anything, update
-    boinc_db.cpp,h
-    and if needed:
-    py/Boinc/
-        database.py
-    html/
-        inc/
-            host.inc (host)
-            db_ops.inc
-        ops/
-            db_update.php
-        user/
-            create_account_action.php (user)
-            team_create_action.php (team)
-    sched/
-        db_dump.cpp (host, user, team)
-        db_purge.cpp (workunit, result)
-*/
-/* Fields are documented in boinc_db.h */
-/* Do not replace this with an automatically generated schema */
-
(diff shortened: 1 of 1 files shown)


db/schema.sql
```diff
@@ -1,33 +1,45 @@
...
-   Supposedly this gives better performance.
-   The others (post, thread, profile) are myISAM
-   because it supports fulltext index
-*/
+-- Most fields are documented in boinc_db_types.h
 
-/* fields ending with id (but not _id) are treated specially
-   by the Python code (db_base.py)
-*/
+-- All fields should be not null
```
TheAspens: Why should all fields be not null?   There are definitely situations where null is a very valid value.

db/schema.sql
```diff
@@ -1,33 +1,45 @@
...
+-- (newer MySQL versions don't have automatic defaults)
+
+-- add new fields to the end of the table
+-- (makes it easier to update C++ code)
+
+-- Engine is specified as InnoDB for most tables;
+-- supposedly this gives better performance.
+-- Some (post, thread, profile) are myISAM because it supports fulltext index
+
+-- Going forward, use double for unix times (no 32-bit problem)
```
TheAspens: I would suggest that unixtimes should be bigint (8 byte integer) rather than double.  This aligns with the direction of time_t in C++.

db/schema.sql
```diff
@@ -1,33 +1,45 @@
...
+
+-- Put index definitions in constraints.sql, not here
+
+-- fields ending with id (but not _id) are treated specially
+-- by the Python code (db_base.py)
```
TheAspens: Do you mind adding some comments about how they are treated specially?   This doesn't tell me what I need to do in order to properly follow convention.

TheAspens:
Ah ok.  I read up on it and it looks like PHP might still have some problems in some situations handling a bigint value.  That's too bad.  Using double doesn't seem like the right answer to me, but due to PHP I don't have a better proposal, so I withdraw my objection.

davidpanderson:
double guarantees you 48 bits of integer.

"null" has never been used.  It isn't needed anywhere, and adds complexity

TheAspens:
> "null" has never been used.

On a vanilla new installation of BOINC and creating a user, the following fields are null:
- user.global_prefs
- user.url
- user.signature
 
Are these errors that emerged when MySQL stopped providing defaults for everything?

davidpanderson:
null doesn't hurt, but from PHP it's essentially the same as an empty
string.
In the C++ API I don't think you can distinguish null from empty string.

PR #2416, opened 2018-03-20
remote job submission: add optional "job params" for batchs

davidpanderson:
This lets you specify the rsc_* parameters and delay bound in the submit call.

Docs on new interfaces: https://boinc.berkeley.edu/trac/wiki/RemoteJobs

PR #2511, opened 2018-05-12
remote job submission: corrected job params retrieval

lfield:
Testing remote job submission using the result from PR #2416 revealed an issue with the job params retrieval.


lfield:
Yes, I had not rebuilt the boinc_gahp with the new code so it was producing different XML. Have just committed the change.

PR #2520, opened 2018-05-16
Add missing mac_spawn entries to lib/Makefile.am

JamesMolson:
When building the core boinc library on OSX it seems it is missing some files.
The library builds ok, but when linking to an external program I get linker errors on spawn().

Adding these files to the Makefile.am makes the problem go away :)

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -92,6 +92,7 @@ if OS_DARWIN
 mac_sources = \
     procinfo_mac.cpp \
     mac/mac_backtrace.cpp \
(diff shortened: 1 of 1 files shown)


CharlieFenton:
Building the client, manager and  libraries using configure & make on the Mac or Windows has not been supported or recommended for several years. That method should only be used for Linux. Please follow the instructions in the source tree at _mac_build/HowToBuildBOINC_XCode.rtf_.

davidpanderson:
Thanks (as Charlie says, build at your own risk).

PR #2530, opened 2018-05-23
Code changes needed to add the WCG brand for the mac applications

armstrdj:
The current code includes the knowledge of four brands for the Mac applications: BOINC, Grid Republic, Progress thru Processors, and Charity Engine. This pull request includes the changes needed to add a fifth brand for WCG. This will allow Mac installers built with these changes to overwrite the WCG branded manager and screen saver when they are developed in the future. The changes to allow for the building of a WCG branded installer will be included in a future pull request.

### DIFF
--- a/clientgui/AdvancedFrame.cpp
+++ b/clientgui/AdvancedFrame.cpp
@@ -1839,7 +1839,7 @@ void CAdvancedFrame::OnConnect(CFrameEvent& WXUNUSED(event)) {
                 fscanf(f, "BrandId=%ld\n", &iBrandID);
                 fclose(f);
             }
-            if ((iBrandID > 0) && (iBrandID < 4))
+            if ((iBrandID > 0) && (iBrandID < 5))
 #endif
             {
                 // If successful, hide the main window if we showed it
--- a/clientgui/BOINCGUIApp.cpp
+++ b/clientgui/BOINCGUIApp.cpp
@@ -154,6 +154,12 @@ bool CBOINCGUIApp::OnInit() {
     SetAppName(wxT("BOINC Manager"));
     SetVendorName(wxT("Space Sciences Laboratory, U.C. Berkeley"));
 
+#ifdef __WXMAC__
+    char displayName[MAXPATHLEN];
+    getDisplayNameForThisApp(displayName, sizeof(displayName));
+    SetAppDisplayName(wxString(displayName)); // {ass the display name to wxWidgets
+#endif
+
 
     // Initialize the configuration storage module
     m_pConfig = new wxConfig(GetAppName());
@@ -272,7 +278,6 @@ bool CBOINCGUIApp::OnInit() {
     // Load desired manager skin
     m_pConfig->Read(wxT("Skin"), &strDesiredSkinName, m_pSkinManager->GetDefaultSkinName());
     m_pSkinManager->ReloadSkin(strDesiredSkinName);
-
 
 #ifdef SANDBOX
     // Make sure owners, groups and permissions are correct for the current setting of g_use_sandbox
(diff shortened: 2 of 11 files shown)


davidpanderson:
This might be a good time to do a code cleanup,
i.e. define the set of brands in one place,
use for loops instead of sequences of ifs,
identify the brands with constants instead of character strings, etc.

Goal: adding a new brand consists of adding one line to an array declaration.

CharlieFenton:
Please note that the handling of branding is done very differently on Windows and on the Mac. On Windows, changing the branding requires recompiling everything. This has long been implemented for WCG as well as the other brands. This is explained in clientgui/AdvancedFrame.cpp lines 1827-1834:
`          // _GRIDREPUBLIC, _PROGRESSTHRUPROCESSORS and _CHARITYENGINE
            // are defined for those branded builds on Windows only
            //
#if defined(_GRIDREPUBLIC) || defined(_PROGRESSTHRUPROCESSORS) || defined(_CHARITYENGINE) || defined(__WXMAC__)
#ifdef __WXMAC__
            // For GridRepublic, Charity Engine or ProgressThruProcessors,
            // the Mac installer put a branding file in our data directory
            //
`

There is yet another approach taken for HTC Power To Give (only in the client code.)

On the Mac, there has been a (now obsolete) script which converts a built BOINC installer and Manager to a branded one, but it has never been implemented for WCG. @armstrdj and I have been working on updating the script and adding support for generating a WCG branded Mac installer and Manager.

On the Mac, the brands are already identified by a numeric value: 0 = BOINC (default), 1 =  GridRepublic, 2 = Progress Thru Processors, 3 = Charity Engine. While these could be replaced by symbolic names, perhaps using an enum, different source files need different strings for each brand. Unfortunately, this makes centralizing them difficult and probably impractical. For details, please see the changes in the feature branch associated with this PR.

armstrdj:
@CharlieFenton The immediate goal here is to get the code changes in that will allow the BOINC branded installer to install over a WCG branded manager. Can you review these changes and see if I missed anything to accomplish that goal?

davidpanderson:
Stylistic issues:
- there should be no numbers (like "5") in the .cpp files
- the #defines and character strings should not be in the .cpp files.
  They should be put (all in one place) in a .h file.

I can makes these changes in a separate PR later on if you want.

CharlieFenton:
> the #defines and character strings should not be in the .cpp files.
They should be put (all in one place) in a .h file.

We could consider a different approach. Currently the Branding file (if present) in the BOINC Data directory has just one line. For GridRepublic it would be:
`BrandId=1\n`
I've been thinking that it might be better to add several more lines to the Branding file, which would contain the branding-related strings currently in the .cpp files (such as lines 197 - 230 in _mac_installer/PostInstall.cpp_. I think this is a better approach than putting the strings in a header file, as it would allow for adding more (future) brands with little or no changes to the source code.

But I want to wait until Jonathan gets the WCG branding working and for a later release, as it is a larger task.

CharlieFenton:
@armstrdj: it all looks good to me, though I won't guarantee that we haven't overlooked anything. I'm looking forward to seeing the script you are writing which will convert the standard Mac BOINC installer to a WCG one. Running that will be the best test of correctness and completeness. Will that be  in a separate PR?

armstrdj:
@CharlieFenton I was planning on doing that under a separate PR but I am getting back to the work now so let me see if I can make some progress and get it in this one.

CharlieFenton:
I am merging this because I need to work on some of the same files to fix a Mac installer issue under OS 10.13 High Sierra. @armstrdj and I will work together under a new PR to get the associated branding script working and to make any additional changes to these source files needed to complete the branding work. To keep the changes for different purposes separate, I will open a third PR to address the issues stylistic issues @davidpanderson raised [here](https://github.com/BOINC/boinc/pull/2530#issuecomment-394029454), probably after the branding work is complete and the second PR has been merged.

PR #2531, opened 2018-05-23
repository: Change line endings in code of conduct to unix style 

TheAspens:
The code of conduct was showing as changed on pulling from master.  This check in converts the file to unix style line endings and removes the custom setting for the file from .gitattributes.

```
[knreed@oc7811586721 boinc]$ file CODE_OF_CONDUCT.md 
CODE_OF_CONDUCT.md: UTF-8 Unicode text, with very long lines
```


Uplinger:
Thanks for fixing this.

davidpanderson:
I'm unable to pull the repo after this merge;
git thinks I've modified the .md file

Ageless93:
This can again be fixed by doing the following from a command line:
```
git rm --cached -r .
git reset --hard
```
That's at least how I fixed the messages I had. Afterwards do a pull and it'll get all data in.

davidpanderson:
Is there something equivalent in Tortoise Git?

Ageless93:
As far as I know, even with Tortoise Git you have Git installed separately and so you should be able to do the command line commands. Tortoise Git is only a GUI that allows you to control Git. (a bit like BOINC Manager, really) 

But checking the [documentation](https://tortoisegit.org/docs/tortoisegit/tgit-dug-rename.html) you can do "git rm --cached" via TortoiseGit -> Delete (keep local)
After [that](https://tortoisegit.org/docs/tortoisegit/tgit-dug-reset.html) do "git reset" via the Reset Master to this log dialog context menu. 

I don't have Tortoise Git, or any other GUI, as none of them will show their context menus for me, so it's easier to just do everything from the command line.

davidpanderson:
Thanks, that did the trick.
We should announce this somewhere since I think all Win developers will
encounter the same problems.

CharlieFenton:
It's not just Windows developers; I had the same problem on my Mac, and the same solution worked for me.

PR #2536, opened 2018-05-28
web: add bootstrap files

davidpanderson:
Alternatively, could get them from a CDN but I think this is better


JuhaSointusalo:
Since BOINC web code requires Bootstrap it seems reasonable to have Bootstrap files installed by default. But instead of adding files could you just rename the existing files? There is already `sample_bootstrap.min.css`, `sample_jquery.min.js` and `sample_bootstrap.min.js` files in `html/user`. Please also update the names in `py/Boinc/setup_project.py`.

davidpanderson:
Oops!  I completely forgot about the sample_ versions.  Never mind!

PR #2537, opened 2018-05-28
[Manager] Fix 'Next' button on previous pages after failed login

AenBleidd:
If while using Wizard try to login to the project with wrong credentials
'Next' button is disabled forever.
This is fixed by enabling 'Next' button on AccountInfoPage by default.


davidpanderson:
I tested this and it works.

PR #2547, opened 2018-06-02
[WIP] [Manager] Add HTML TOU support to Wizard

AenBleidd:
Terms Of Use can be displayed now as HTML as well as plain text.
Fixed xml_unescape function to support escaped \r and \n tags.
Enlarged the size of Terms Of Use field to fill all wizard page.

### DIFF
--- a/clientgui/TermsOfUsePage.cpp
+++ b/clientgui/TermsOfUsePage.cpp
@@ -54,6 +54,7 @@ BEGIN_EVENT_TABLE( CTermsOfUsePage, wxWizardPageEx )
     EVT_WIZARDEX_CANCEL( -1, CTermsOfUsePage::OnCancel )
     EVT_RADIOBUTTON( ID_TERMSOFUSEAGREECTRL, CTermsOfUsePage::OnTermsOfUseStatusChange )
     EVT_RADIOBUTTON( ID_TERMSOFUSEDISAGREECTRL, CTermsOfUsePage::OnTermsOfUseStatusChange )
+    EVT_HTML_LINK_CLICKED(ID_TERMSOFUSECTRL, CTermsOfUsePage::OnLinkClicked)
 ////@end CTermsOfUsePage event table entries
  
 END_EVENT_TABLE()
@@ -138,7 +139,19 @@ void CTermsOfUsePage::CreateControls()
 
 ////@end CTermsOfUsePage content construction
 }
-  
+
+
+void CTermsOfUsePage::OnLinkClicked( wxHtmlLinkEvent& event ) {
+    wxString url = event.GetLinkInfo().GetHref();
+    if (url.StartsWith(wxT("http://")) || url.StartsWith(wxT("https://"))) {
+        // wxHtmlLinkEvent doesn't have Veto(), but only loads the page if you
+          // call Skip().
+            wxLaunchDefaultBrowser(url);
+    } else {
+        event.Skip();
+    }
+ }
+
 /*!
  * Gets the previous page.
  */
--- a/clientgui/TermsOfUsePage.cpp
+++ b/clientgui/TermsOfUsePage.cpp
@@ -224,14 +224,10 @@ void CTermsOfUsePage::OnPageChanged( wxWizardExEvent& event ) {
         _("Please read the following terms of use:")
     );
 
-    std::string tou = pc.terms_of_use;
-    xml_unescape(tou);
-    wxString terms_of_use(tou.c_str(), wxConvUTF8);
-    // HTML TOU can have no open/close html tags
-    // so I see no proper way to identify
-    // whether it is html or plain text
-    // and I have to use this dirty hack
-    if (pc.terms_of_use == tou) {
+    wxString terms_of_use(pc.terms_of_use.c_str(), wxConvUTF8);
+    // We need to replace all line endings in text TOU
+    // to make it looks properly in HTML Window
+    if (!pc.terms_of_use_is_html) {
         terms_of_use.Replace("\r\n", "<br>");
         terms_of_use.Replace("\r", "<br>");
         terms_of_use.Replace("\n", "<br>");
--- a/clientgui/TermsOfUsePage.cpp
+++ b/clientgui/TermsOfUsePage.cpp
@@ -103,6 +103,8 @@ bool CTermsOfUsePage::Create( CBOINCBaseWizard* parent )
  
 void CTermsOfUsePage::CreateControls()
 {    
+#define TERMSOFUSEWIDTH ADJUSTFORXDPI(580)
+#define TERMSOFUSEHEIGHT ADJUSTFORYDPI(250)
 ////@begin CTermsOfUsePage content construction
     CTermsOfUsePage* itemWizardPage96 = this;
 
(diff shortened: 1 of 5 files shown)


AenBleidd:
@CharlieFenton, could you please verify that the size that I set for TOU field is OK for Mac. I saw in another page of WIzard that similar field has another size for Mac.

brevilo:
Hi, will try to do that ASAP. Not sure whether I'll be able to do more (actual build / live test) than a static code review, though.

brevilo:
> could you please verify that the size that I set for TOU field is OK for Mac.

Ideally the window would be resizable and the TOU field should scale with the window on resizing. The current macOS manager's TOU field is fixed size and **way** to small. Thus it's not user-friendly and also pretty ugly design-wise.

AenBleidd:
@brevilo, Wizard window is fixed anyway. I found no easy way to resize it. In any case, on the first page we have similar field with fixed size and it looks OK for now. I made the same changes. I agree, that ideally it should be resizable. I saw that before my changes TOU window was small. Now it's much more big and it's easier to read its content

brevilo:
Thanks, **any** improvement is welcome 👍

davidpanderson:
I tested this on Windows.  It works, except that links don't work.  A busy icon appears for a few seconds, but the new page doesn't appear either here or in a browser.

AenBleidd:
@davidpanderson, I've tested this with https://albertathome.org (HTML) and https://albert.phys.uwm.edu (TXT)

AenBleidd:
https://albertathome.org/
image
https://albert.phys.uwm.edu/
image

AenBleidd:
@davidpanderson, I've just use these links in Manager and they give me the result above

clientgui/TermsOfUsePage.cpp
```diff
@@ -222,10 +224,19 @@ void CTermsOfUsePage::OnPageChanged( wxWizardExEvent& event ) {
         _("Please read the following terms of use:")
     );
 
-    m_pTermsOfUseCtrl->SetValue(
-        wxString(pc.terms_of_use.c_str(), wxConvUTF8)
-    );
-    m_pTermsOfUseCtrl->SetSelection(0,0);
+    std::string tou = pc.terms_of_use;
+    xml_unescape(tou);
```
JuhaSointusalo: I think this should go to the code that reads XML. Any objects that comes out of XML parser should be usable without any further unescaping.

JuhaSointusalo:
Hold merge. I think the spec for this needs more discussing.

AenBleidd:
> Hold merge. I think the spec for this needs more discussing.

This supposed to be a small easy fix (joke)

clientgui/TermsOfUsePage.cpp
AenBleidd: This will make it harder to identify whether we got html tou or text

JuhaSointusalo:
Goes to the same category as 'how hard could it be?'.

clientgui/TermsOfUsePage.cpp
AenBleidd: Fixed

AenBleidd:
So I'm waiting now for a final decision to make corresponded changes to this PR

CharlieFenton:
> @CharlieFenton, could you please verify that the size that I set for TOU field is OK for Mac.

I don't understand what it means to say the size is OK. Wouldn't the size required depend on the amount of text in the HTML set by the project?

If the HTML is too large to fit in the window, does it automatically add scroll bars so you can scroll through the entire message? If not, please add that.

davidpanderson:
https://albertathome.org/

On my system (VS2010, Win7) I see the links as shown above.
But they don't work (clicking on them does nothing).

AenBleidd:
> https://albertathome.org/
On my system (VS2010, Win7) I see the links as shown above.
But they don't work (clicking on them does nothing).

Response from this site gives me next TOU:

```
<ul>
<li>You are at least 16 years old</li>
<li>You have read and acknowledged the Einstein@Home <a href="/privacy">privacy policy</a> and <a href="/community/moderation">forum guidelines</a></li>
<li>You will run Einstein@Home only on devices that you own, or for which you have obtained the owner's permission</li>
<li>You understand that Einstein@Home assumes no liability for any damage, data loss, or other event or condition that may occur on your devices as a result of participating in Einstein@Home</li>
<li>You understand that scientific data supplied by Einstein@Home are not to be used for any other purpose</li>
</ul>
```

Links here are relative. I think they can be fixed from the code but it also needs to be decided.

AenBleidd:
>I don't understand what it means to say the size is OK. 

On the first page of WIzard we have a text field with the project description.
And in code I found next lines:

```
#ifdef __WXMAC__
#define DESCRIPTIONSWIDTH 350
#else
#define DESCRIPTIONSWIDTH ADJUSTFORXDPI(310)
#endif
```

As soon as I have a similar field here it looks reasonable to verify it on Mac. Unfortunately I have no Mac so there is no way for me to verify it but to ask you to do this.
Of course, maybe that piece of code is no longer needed but I just want to be sure.

CharlieFenton:
How do I check out this feature brach so I can test it? When I tell SourceTree (on Mac) or TortoiseGit (on Win7) to show all branches, I only see those branches of yours which have already been merged into master. The only branches I see which have not yet been merged are those create by @davidpanderson.

AenBleidd:
You need to add my repo aenbleidd:boinc

Best regards,
Vitalii

Sent via Android

вт, 5 июня 2018, 11:52 CharlieFenton <notifications@github.com>:

CharlieFenton:
I was able to download this branch using the GitHub browser interface and built and tested it. The size of the window is fine on the Mac, but (as @davidpanderson reported earlier), clicking on the links in the HTML does nothing. 
> Response from this site gives me next TOU:

That is not what the response to clicking  on the links should be. Clicking on the `privacy policy` link should open a web page (in your default browser) showing the privacy policy <https://albertathome.org/privacy>, and clicking on the `forum guidelines` link should open a web page describing the forum guidelines <https://albertathome.org/community/moderation>. 

Seeing the raw HTML you quoted, it appears that the links are given relative to the albert@home web site. This may be the same problem described in clientgui/NoticreListCtrl.cpp lines 186 - 204. You may also need to convert the relative links to fully qualified links, replacing `href="/privacy"` with `href="albertathome.org/privacy"` or perhaps `href=https://albertathome.org/privacy"`.

AenBleidd:
Will do

Best regards,
Vitalii

Sent via Android

вт, 5 июня 2018, 12:55 CharlieFenton <notifications@github.com>:

brevilo:
I updated the links to absolute addressing. I don't think we need the extra complexity to add "link-completion" to the manager. Yep, it's easy to do but we should keep it simple as long as there isn't a compelling reason to do otherwise.

Please re-test the link handling.

brevilo:
> How do I check out this feature brach so I can test it? 

@CharlieFenton you want to add Vitalii's repo as another (git) "remote" to your local repo.

CharlieFenton:
@AenBleidd We're almost there. You need to add a bit more code to handle the click on the link, as in clientgui/NoticreListCtrl.cpp lines 57 and 279 - 288, plus clientgui/NoticreListCtrl.h line 51. I tested this by adding the following lines to your clientgui/TermsOfUsePage.cpp:
After line 56:
`    EVT_HTML_LINK_CLICKED(ID_TERMSOFUSECTRL, CTermsOfUsePage::OnLinkClicked)`
and wherever you like:
```
void CTermsOfUsePage::OnLinkClicked( wxHtmlLinkEvent& event ) {
    wxString url = event.GetLinkInfo().GetHref();
    if (url.StartsWith(wxT("http://")) || url.StartsWith(wxT("https://"))) {
        // wxHtmlLinkEvent doesn't have Veto(), but only loads the page if you
          // call Skip().
            wxLaunchDefaultBrowser(url);
    } else {
        event.Skip();
    }
 }
```
And of course add this to clientgui/TermsOfUsePage.h:
`    void OnLinkClicked( wxHtmlLinkEvent & event );`

With these lines added, the links work correctly.

AenBleidd:
@CharlieFenton, can you send me the patch? Then I can add it to this pull request. If no - I can do this later today. Thanks

CharlieFenton:
@AenBleidd, I have sent you my modified files by separate email.

AenBleidd:
@CharlieFenton, thank you. I'll update this pr with your code asap. I believe you but in any case I want to verify it by myself today later.

AenBleidd:
Thank you, @CharlieFenton, it's now ready to be verified again.

davidpanderson:
Links now work for me (Win7).  I'll merge when Charlie signs off.

davidpanderson:
Nice work!

AenBleidd:
So at this point I'll wait for a decision to support both TXT (for old clients) and TXT/HTML (for new clients) TOU from the API side. Then I can make corresponded changes on the Manager (and client) side.

PR #2551, opened 2018-06-07
sched/db_dump: if existing stats directory contains a .htaccess file,…

bema-ligo:
… copy it to the newly created one

- this is if you want to restrict stats export via .htaccess

### DIFF
--- a/sched/db_dump.cpp
+++ b/sched/db_dump.cpp
@@ -1095,6 +1095,16 @@ int main(int argc, char** argv) {
         time_t now = time(0);
         tmp = localtime(&now);
         char base[256];
+        strncpy(base, spec.final_output_dir, sizeof(base));
+        strncat(base, "/.htaccess", sizeof(base) - strlen(base));
+        if (boinc_file_exists(base)) {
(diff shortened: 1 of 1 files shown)


bema-ligo:
CI checks failed in 'libs' and 'apps-mingw', that doesn't look related to my PR.

TheAspens:
I've tested and reviewed the code and it works as is.  However, I do not think that this is a good approach.  Apache specifically warns against the use of .htaccess files: https://httpd.apache.org/docs/2.4/howto/htaccess.html#when

Additionally, this means that the .htaccess file will be owned by the user running db_dump and it is usually not a good security practice to have security configuration files owned by application ids.

What is the reason for not creating a configuration file that specifies a <Directory> tag and putting it in the /etc/httpd/conf.d directory?

TheAspens:
If on debian the custom config directory would be /etc/apache2/conf-available and a sym link in /etc/apache2/conf-enabled

bema-ligo:
@TheAspens Thanks for pointing that out. Indeed I was under the "common misconception" mentioned in the document. Case closed.

PR #2556, opened 2018-06-08
Web: fixed bugs in delete account feature introduced in bcd853bb4f4

brevilo:
* missing include - required for delete_user()
* wrong property access syntax


TheAspens:
Sorry about that - looks like when I refactored during some of the feedback in the pull request I didn't cover all my test cases.  Thanks for fixing the bug.

brevilo:
No problem. Thanks!

PR #2558, opened 2018-06-08
Drupal: Added bts() functions to terms of use forms.

drshawnkwang:
Part of https://dev.gridrepublic.org/browse/DBOINCP-428


drshawnkwang:
Tristan correctly pointed out that the `bts()` function should send the strings to the project-specific resource. The latest commit does this.

PR #2559, opened 2018-06-08
Updated registration flow

davidpanderson:
New and improved registration flow.

Account creation is now done by default on the web, not in the BOINC Manager.
We can add consent stuff here as needed.

For projects that are vetted, and that put their project ID in config.xml,
the autoattach feature now works.
This means that new users (downloading BOINC for the first time)
will be pre-attached to the project when BOINC first runs.
No need for a separate attach/login in the Manager.

I've deployed this on SETI@home; seems to work as intended.

### DIFF
--- a/html/project.sample/project.inc
+++ b/html/project.sample/project.inc
@@ -118,7 +118,7 @@ function make_php_mailer() {
 //
 function project_banner($title, $prefix, $is_main) {
     if ($is_main) {
-        echo '<img class="img-responsive" src="img/water.jpg">';
+        echo '<img class="img-responsive" style="width:100%" src="img/water.jpg">';
     }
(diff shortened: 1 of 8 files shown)


davidpanderson:
I made diagrams of flow for new and returning users, starting from the home page:
https://boinc.berkeley.edu/signup_flow.pdf

There are different diagrams for vetted projects (which can use autoattach) and other projects.

TheAspens:
I failed to get to this before I had to leave for vacation - sorry for the delay.  If someone else can review this soon that would be very helpful.  @Uplinger @tristanolive @drshawnkwang are any of you able to help?

Otherwise I will get to it early the week of June 25.

PR #2561, opened 2018-06-16
Adding Wrapper $USER_NAME and $TEAM_NAME command line macros

Michael-DHEP:
Merge of Yoyo's and my code to add a couple of extra macros to the wrapper command_line setting.  Any non-alphanumeric characters in the names will be replaced by _ (underscore) in the expansion.

### DIFF
--- a/samples/wrapper/wrapper.cpp
+++ b/samples/wrapper/wrapper.cpp
@@ -269,11 +269,31 @@ void str_replace_all(string &str, const string& s1, const string& s2) {
     }
 }
 
+// replace non-alphanumeric characters with _
+string trim(string &str){
+    string rv = "";
+    for (unsigned int i = 0; i < str.length(); i++)
+    {
+        char current = str[i];
+        if ( (current >= 'a' && current <= 'z') ||
+             (current >= 'A' && current <= 'Z') ||
+             (current >= '0' && current <= '9') ||
+             current == '_' ) {
+            rv += current;
+        } else {
+            rv += '_';
+        }
+    }
+    return rv;
+}
+
 // macro-substitute strings from job.xml
 // $PROJECT_DIR -> project directory
 // $NTHREADS --> --nthreads arg if present, else 1
 // $GPU_DEVICE_NUM --> gpu_device_num from init_data.xml, or --device arg
 // $PWD --> current directory
+// $USER_NAME --> name of user logged into client
+// $TEAM_NAME --> user's team name
 //
 void macro_substitute(string &str) {
     const char* pd = strlen(aid.project_dir)?aid.project_dir:".";
@@ -281,6 +301,10 @@ void macro_substitute(string &str) {
 #ifdef DEBUG
     fprintf(stderr, "[DEBUG] replacing '%s' with '%s'\n", "$PROJECT_DIR", pd);
 #endif
+
+    string userName = aid.user_name, teamName = aid.team_name;
+    str_replace_all(str,"$USER_NAME",trim(userName));
+    str_replace_all(str,"$TEAM_NAME",trim(teamName));
 
     char nt[256];
     sprintf(nt, "%d", nthreads);


davidpanderson:
I suggest that trim() change non-alphanumerics to &#nn; form (like in URL encoding)
so that the app can reconstruct the name if it wants.
Also call it something other than "trim".

Michael-DHEP:
Hello David yes agreed on a more descriptive name for the function.

However I kind of agree with the sentiment of Yoyo's original no nonsense approach to completely rule out any chance of SQL injection, or HTML rendering issues further down the line.

But yes it would be nice to be able recover some of the characters in names.  However also think a full blown encoder supporting every UTF-9 character might be beyond the scope of this task at hand.  Perhaps a compromise supporting []<>-' which are commonly used in the community?

davidpanderson:
See xml_escape() in parse.cpp.  You'd need to modify this slightly.

JuhaSointusalo:
Your code basically supports only English names. That's not very nice. In my opinion you need to support full Unicode.

With full Unicode you can't reasonably make a list of allowed characters, and you shouldn't make a list of non-allowed characters because it's too easy to miss something. So you need to be able to handle any character. David's example `&#nn;` is XML escaping. BOINC has a function to do XML escaping but that won't help here. Command line and XML have different special characters. You need to have special handling for some or all of ` `, `"`. `'`, `\`, `^`. `` ` `` and possible others, depending on your command line parser. BOINC also has a function for URL encoding which does encode all of those characters. However, the function encodes more than [is necessary](https://en.wikipedia.org/wiki/Percent-encoding) and it would be unwise to take a dependency on that.

I would suggest using [Base64 encoding](https://en.wikipedia.org/wiki/Base64) instead. All the characters used in it should be safe to use on command line. BOINC has implementation available in `base64.h`.

Also note that users are not required to join a team. As such `$TEAM_NAME` might expand to an empty string. I'm not sure how much validation BOINC does on user names so it would be best that you are prepared that `$USER_NAME` expands likewise to an empty string. You need to figure out how are you going to parse an empty string. I don't think the wrapper is able to help you with that because it can't know how your command line is structured or how you parse it.

As for SQL injection and other vulnerabilities. Please understand that if someone wants to create trouble they don't have use a wrapper that filters bad characters out. They can substitute your wrapper with their own that inserts bad characters on the command line. Or they can simply start your app from terminal without the wrapper. Or they can completely bypass BOINC and directly upload bad results to your server. If your server has vulnerabilities you need to fix your server. BOINC wrapper can do absolutely nothing to help you secure your server.

samples/wrapper/wrapper.cpp
```diff
@@ -282,6 +302,10 @@ void macro_substitute(string &str) {
     fprintf(stderr, "[DEBUG] replacing '%s' with '%s'\n", "$PROJECT_DIR", pd);
 #endif
 
+    string userName = aid.user_name, teamName = aid.team_name;
```
JuhaSointusalo: Please follow conventions of the surrounding code. Wrapper uses `_` to separate words, not camelCase.

Michael-DHEP:
OK just a general comment.  All of these options require the target application be changed to process encoded command line arguments, whereas the original can be used with any existing application.  Given your comments on security JuhaSointusalo, perhaps any username that is acceptable within BOINC (we are reading aid.user_name) will also be acceptable to be expanded as a command line argument as is? ie. simply expand $USER_NAME to aid.user_name.  If it's already in the BOINC user MySQL user table, and floating the rest of BOINC code and website, it can't be that bad - ie. will be OK as a command line argument.  Except POSIX shell expansion characters (see answer of https://unix.stackexchange.com/questions/347332/what-characters-need-to-be-escaped-in-files-without-quotes) which can be prepended with a backslash \ . 

So unless I'm missing something this would allow characters such as é ñ (just as they are) and even \ / " ' $ (all prepended with \) and wouldn't require any modifications from the application code (existing applications will already be prepared for anything coming in as a command line argument so they can do their own filtering if they didn't like ' for instance).

To summarise what I'm proposing is to expand $USER_NAME to aid.user_name with any of the characters listed in https://unix.stackexchange.com/questions/347332/what-characters-need-to-be-escaped-in-files-without-quotes prepended by \ .  This would support POSIX.  I'm aware of the need to cater for any Windows specifics.  But could start crossing that bridge if you're happy with this solution in general.

Agreed with variable naming conventions, my bad.

JuhaSointusalo:
I have to admit I was thinking about shells when I listed the special characters but the command line is actually not parsed by shell. On Unixes the command line parameters are concatenated together and then split into an argument vector in [`parse_command_line`](https://github.com/BOINC/boinc/blob/master/lib/str_util.cpp#L250). You'd need to add support for escaping quotes and escapes there. Other than that, yes I think escapes and quotes would work, or at least I can't quickly come up with an example that wouldn't work.

Windows is different. The command line is passed as one string to [`CreateProcess`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425(v=vs.85).aspx) and it's up to the new process to parse the command line. Windows provides [`CommandLineToArgvW`](https://msdn.microsoft.com/en-us/library/windows/desktop/bb776391(v=vs.85).aspx) for the job and judging by code comments MS CRT seems to follow similar parsing rules. MS CRT has some special handling for two consecutive double quotes though. It might take quite a bit of experimenting and testing to have confidence that the quoting and escaping are right.

Problems come with encoding used for the command line. Internally Windows uses UTF-16, the wrapper isn't Unicode program and uses some ANSI code page. The user and team names on the other hand comes from server and use UTF-8. So when `CreateProcess` is called the bytes of UTF-8 name is converted to UTF-16 as if the bytes were ANSI characters. If the new process is not a Unicode process the command line is converted back to ANSI code page encoding. I don't think that kind of round-trip is guaranteed to produce the same string you started with. For starters, some languages are supported only in Unicode. What would the UTF-8 bytes convert to in that case? Then there's double byte code pages but the UTF-8 string might not have even number of bytes or they might not make sense when interpreted as double byte sequence.

Wrapper could convert the UTF-8 string to ANSI but that may be a lossy conversion. Unicode obviously has more characters than any ANSI code page and, again, not all languages have ANSI code page.

So I think the escapes and quotes option requires that the wrapper converts the command line to UTF-16 and uses Windows Unicode API to create new process. Converting the command line is, of course, not that straightforward. Anything from server is in UTF-8 and we can demand that wrapper's job XML is in UTF-8 but the wrapper can expand filenames on the command line. Those filenames come from ANSI APIs and need different conversion. Science apps then use `wmain(int argc, wchar_t** argv)` or `CommandLineToArgvW` or whatever to get the command line and user and team names. If the science apps only supports ANSI code pages shame on them. BOINC would still do its best to support any name.

All that ignores environment variables and shell scripts. The same code is used to expand strings for environment variables. Environment variables don't need that kind of quoting and science apps interested in passing the names though environment would have to undo the quoting. Unix shell script should get name as a single argument. If it needs to pass the name to the next process in line the script would need to re-apply escaping and quoting according to the shell's rules. Windows' `cmd.exe` requires different quoting syntax than `CreateProcess`. I think supporting names and `cmd.exe` scripts together is best forgotten. This is just something to document I think.

TheAspens:
I'm closing and reopening this to re-trigger the AppVeyor build (it failed because the 2013 license file expired - not because of the code).

lfield:
I would suggest closing this. I don't see the value in the app knowing the user or team name and it opens a can of worms.

PR #2569, opened 2018-06-24
[WIP] Mac: fixes to installer for issues under OS 10.13 High Sierra

CharlieFenton:
Fixes to installer for issues under OS 10.13 High Sierra when multiple users are to be allowed to run BOINC Manager

 - fix setting or deleting Login Items for users other than the one running the installer
 - fix setting the screensaver for users other than the one running the installer

This situation probably occurs fairly infrequently, but I recently noticed these problems when running the Mac installer, Uninstaller, and AddRemoveUser utility under OS 10.13. As far as I have been able to determine, these problems do not occur on earlier versions of OS X.

It may be desirable to back port these fixes to BOINC 7.10 once they are complete.

### DIFF
--- a/clientgui/BOINCGUIApp.cpp
+++ b/clientgui/BOINCGUIApp.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2017 University of California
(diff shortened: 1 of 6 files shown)


CharlieFenton:
This PR is ready to be merged.

PR #2571, opened 2018-06-24

ChristianBeer:
I already fixed problems with tabs and printf positioning.


davidpanderson:
Thanks

PR #2572, opened 2018-06-24
Locale: enhance translation update script

ChristianBeer:
- No testmode anymore as this was unpractical
- added instructions what to check since not everything is automatable
- translations are pulled from transifex when 97% complete. This proved practical in the last release.
- added simple po file check to prevent incorrect tabs and printf problems from translations (potential to break the manager)

### DIFF
--- a/locale/update_translations.sh
+++ b/locale/update_translations.sh
@@ -2,20 +2,9 @@
 
 set -e # abort if a command exits non-zero
 
-# update localization files from transifex and compile .po to .mo if needed
-# .po files are only updated if they are 100% translated, outdated files are not removed!
-# Then commit and push changes.
-
-testmode=0
-if test $# -gt 0; then
-  if test $1 = "-t"; then
-    testmode=1
-  else
-    echo "Usage: $0 [-t]"
-    echo "     -t  testmode (don't commit or push to git repository)"
-    exit 1
-  fi
-fi
+# update localization files from transifex and compile .po to .mo
+# .po files are updated according to minimum_perc in .tx/config, outdated files are not removed!
+# Then commit and push changes manually.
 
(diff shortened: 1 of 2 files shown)


locale/update_translations.sh
```diff
@@ -28,13 +17,12 @@ command -v tx >/dev/null 2>&1 || { echo >&2 "tx (transifex-client) is needed but
 
 # check if working directory is clean to ensure we only commit localization changes
 if test 0 -ne `git status -s -uno |wc -l`; then
-  echo "Please commit your pending changes first"
-  exit 1
+  echo "Warning: You have pending changes! Please make sure to only commit localization changes!"
```
TheAspens: Did you intentionally remove the exit if there are pending changes?

ChristianBeer: Yes, although in hindsight I should have added `read -rsp $'Press enter to continue...\n'` to make sure the user of the script sees the message. Otherwise it might vanish too soon. I might add that if no other problems arise.

TheAspens: Ok - the script doesn't automatically commit like it used to so this is probably a nice to have rather than a must.  I'm satisified with this code.

davidpanderson:
Is the use of this documented anywhere?
The official translation doc
https://boinc.berkeley.edu/trac/wiki/AdminLocalize
is out of date and doesn't mention it.

ChristianBeer:
The AdminLocalize page is not outdated. It is still accurate. Follow the link in the second sentence on this page to [TranslazionSystem](https://boinc.berkeley.edu/trac/wiki/TranslationSystem). Which needs to be slightly updated after this PR becasue of the removed test-mode.

PR #2575, opened 2018-06-26
Fix .gitattributes to use LF by default.

pluskal:
In case that boinc is cloned on Windows, git adds CRLF to files that were not explicitly specified. When those files were copied into a Linux directly, without committing them to git, the line endings were preserved, and compilation failed.

It would be much more convenient to have all files cloned with expected line endings to overcome this issue.

This issue may occur when the development is done on Windows and build on Linux using Docker for example.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,7 +3,7 @@
 #
 
 # Random stuff
-*                       text=auto
+*                       text=auto eol=lf
 
 # Source files
 #
@@ -45,3 +45,11 @@ checkin_notes_2012      text eol=lf
 *.log                   text eol=lf
 *.rtf                   text eol=lf
 *.rul                   text eol=crlf
+*.png                   binary diff=exif
+*.ttf                   binary 
+*.otf                   binary
+*.woff 	                binary 
+*.woff2                 binary 
+*.exe                   binary 
+*.dll                   binary 
+*.dat                   binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,10 +12,10 @@
 *.rc                    text eol=lf
 *.php                   text eol=lf
 *.xpm                   text eol=lf
-*.cu					text eol=lf
-*.cl					text eol=lf
-*.c					    text eol=lf
-*.hpp					text eol=lf
+*.cu                    text eol=lf
+*.cl                    text eol=lf
+*.c	                    text eol=lf
+*.hpp                   text eol=lf
 
 # Project files
 #


AenBleidd:
@pluskal, how this change will affect existing local branches? All files will be marked as changed?

pluskal:
I tried to create a branch before the change to ```.gitattributes```, do some changes/commit, change  ```.gitattributes``` and merge the testing branch, and everything was OK. So I guess that there should not be any impact on existing branches and a merge is fast-forward.

There is no change to existing files. This change only ensures that newly cloned repository will have consistent line endings.

Although, would not it make more sense to have an exhaustive list of all file extensions and define line endings per file type explicitly?  OR, to enforce LF to all text files and be done with it. Windows 10 can handle it now by default Notepad and any reasonable editors can as well.

If you want it, here is a list of all file extensions of all text files in this project [file_endings.txt](https://github.com/BOINC/boinc/files/2136051/file_endings.txt) generated by this command. This does not include files without an extension like ```Makefile```.

 ```find . -type f -exec grep -Iq . {} \; -and -print | grep -o '[^/]*$' | grep -o '\.[^.]*$' | sort | uniq >file_endings```

 But to define whether they should be CRLF or LF would be up to you.

AenBleidd:
Could someone please take a look at this pr to decide whether it should be merged or not. Thanks

tristanolive:
I'd be inclined to leave it as it is and not merge this PR. The point of "auto" is that it sets the line endings according to your environment, so if you pull changes on a Windows environment, you get CR+LF, while on Linux you get LF.

I understand this causes a problem if pulling to Windows and then manually copying files to Linux, but that seems to be a fringe case and not a typical thing to do. I think the proper approach would rather be to commit the changes on Windows to a development branch and push that branch to github; then the development branch could be pulled to the Linux system and the line endings would be converted as expected automatically.

pluskal:
Except for the case if you are developing on Windows and building a Linux docker container. Then you need to have correct line endings. 

Windows can handle LF, but Linux can not.

tristanolive:
Makes sense; but in this scenario, can't you access the repo directly from the Linux environment that is being used to compile? I can understand you probably wouldn't want a git repo in the Docker image that is being built... but I think that is the point that is confusing me, because:
- why compile on the system being used as the target Docker image; or
- why not put a git repo on the same system being used to compile, which will end up with a mess of object files, etc., created during compilation anyway?

Apologies if these are questions out of ignorance, I hope they might help others understand the need for setting explicit LF endings, too. :)

pluskal:
Thanks for these questions, I agree that they are important to clarify.

I am not building in a targeted Docker image. As you mentioned, that would leave source codes and other compilation artifacts inside, which is not desirable. Therefore, I have Dockerfile containing two images, one (temporary) for compilation, and another for the production.

You are right that it would be possible to clone repository in building container. It would work, but it has several drawbacks:
 * It would be challenging to use layers; ergo, build BOINC once and update only changed file related to a developed project, as you do when you compile over and over in the same repository. I build BOINC about 5 minutes compared to a few seconds when only changed code is compiled.
* I would need to change Dockerfile each time I work on a different branch to make this scenario possible. Dockerfile is tracked in git so that it would generate conflicts each merge. I do git-flow on this project so it would result in conflict resolutions in merge commits which is not a nice thing to have.
 * I would need to commit each change to make it compile and test in Docker. That would generate pretty messy history that would have to be cleaned up, resulting in once again slowdown in development. 

Last, you already have some files with forced LF, so why not to clean all up to be compatible with both platforms? :)

tristanolive:
It's a wide reaching change, but a compelling use case. There's a BOINC contributors call tomorrow at 1800 UTC, so this topic has been [added to the agenda](https://docs.google.com/document/d/1oORE4LQJ9uB1JuVxbYThph9uSeyDbqtTPZ54xNauCeY) (join if you're able, otherwise I'll try to advocate the convenience of this change).

pluskal:
Thanks for an invite! I am afraid that I would not be able to make it because I have already scheduled another call for that time.

I wish you luck with advocating!

tristanolive:
No problem -- and for the record, this is a bi-weekly call, every other Thursday, that has been alternating between 1800 UTC and 1200 UTC (so in two weeks it will be at the 1200 UTC time); feel free to join anytime you want to discuss things. There's usually not a hard agenda.

drshawnkwang:
I brought up this PR during the BOINC contributors call. 

A few weeks/months ago we (BOINC git repository) had an issue with the .gitattributes file that caused a number of people to have dirty working-trees. This in turn lead to a situation where people couldn't even git pull in a fix. Needless to say it was irritating and the event caused us to think: "Don't mess with this file".

Is there a way (@pluskal) can rearrange your workflow so as to not require this change? I do not claim to understand your workflow, as I do not use docker regularly. Would something like `dos2unix` help?

pluskal:
Oh, I see. No problem then.

I will just patch it locally when I pull changes from upstream. 

Thanks for discussing it!

PR #2580, opened 2018-06-27
web rpc: Added global scope to master_url variable.

drshawnkwang:
BOINC web code [refactored `$master_url` into a global variable](https://github.com/BOINC/boinc/commit/5ea948224ea15ea53bd8167c78e331951d9e1ece). The E@H Drupal installation breaks due to this change. This is due to the scope of the variable not being explicitly set  as global. The functional result is that get_project_rpc does not return the master_url for the E@H project. We believe we're the only ones affected due to our Drupal Web front-end.

If these two small changes are made, this should fix the get_project_config RPC for us. I doubt this will negatively affect the rest of the BOINC html web code, but if someone could check that would be great.

### DIFF
--- a/html/inc/util.inc
+++ b/html/inc/util.inc
@@ -37,6 +37,7 @@ require_once("../inc/bootstrap.inc");
 // parse some stuff from config (do it here for efficiency)
 //
 $config = get_config();
+global $master_url;
 $master_url = parse_config($config , "<master_url>");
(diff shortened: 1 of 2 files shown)


tristanolive:
This same sort of thing was done before to make other BOINC globals available to Drupal (e.g. 9798b829), so there should be no negative effects.

TheAspens:
Looks fine to me.

TheAspens:
Also tests fine.

PR #2581, opened 2018-06-29
Fixes for WSL subsystem

RichardHaselgrove:
set client state option on re-read
disable anonymous platform wsl plan classes when wsl not permitted


JuhaSointusalo:
While testing this I discovered that using `app_info.xml` with a WSL app and `dont_use_wsl` together results in client deleting the files the WSL app uses. The next time you set `dont_use_wsl` to `0` the client complaints that files are missing. Not your fault, the way `app_info.xml` is handled needs improving.

PR #2585, opened 2018-07-02
 [GUI] Tune advanced preferences dialog

RichardHaselgrove:
Following user feedback on v7.10.2, minor adjustments to #2043 

1) Make 'Save' the default button if (and only if) the user already has local preferences active
2) Show confirmation dialog when (and only when) switching from web to local preferences

The two commits in this PR are independent and either could be deployed separately, but I think they work better together.

Discussion and justification in [CANCEL as default on options change](http://boinc.berkeley.edu/dev/forum_thread.php?id=12515) 

@CharlieFenton - specifically requesting your review to ensure it works on other platforms too - I've only tested Windows.


RichardHaselgrove:
While working on this PR, I tested the 'Help' button - it's a mess. The code here is fine, but the target is a Wiki page...
...which is a manual redirect...
...to a page which doesn't exist...
...but wiki automation suggests a third page...
...which has had all the 'help' text deleted...
...leaving just screenshots...
...which are out of date.

If the events of 14 February 2017 weren't recorded in the final history log, I'd think we'd been hacked.

Ageless93:
It's not settings. It's preferences. It's local preferences, not local settings. You speak of web-based preferences, so why use settings for the local ones?

The present blurb at the top of the window says:
"Using local preferences. Click "Use web prefs" to use web-based preferences from {project-url}"
or
"Using web-based preferences from {project-url}. Set values and click Save to use local preferences instead."

So keep the same text all over.

CharlieFenton:
I have confirmed that the code changes under this PR work as intended on my Macintosh.
> I think it is far more important to have a confirmation dialog when switching from local to web prefs.

I had forgotten that there already was a confirmation dialog when switching from local to web prefs, and I now see that this PR does not change that. But I think that the local to web prefs confirmation could be somewhat more emphatic that your local prefs will be lost and you won't be able to "undo" the decision.

TheAspens:
It looks like scruitinizer errored out.  I"m closing and reopeninging this pull request in order to retrigger it.

RichardHaselgrove:
> I had forgotten that there already was a confirmation dialog...

Yes, I saw that while working on it - that's partly what triggered my comment about BOINC's terse style in the discussion. I didn't feel the need to change it, because the button you click is much more specific - less likelihood of being unaware of the consequences. But we could use the opportunity to make the two styles more similar:

 "Changing to use web preferences will delete the local preferences from all four pages of this dialog. You won't be able to 'undo' this operation later. Do you want to proceed?" 

It would make life easier for the translation team if we make both changes together, if at all.

clientgui/DlgAdvPreferencesBase.cpp
```diff
@@ -37,6 +37,8 @@
 #define STATICBOXVERTICALSPACER 10
 #define DAYOFWEEKBORDERSIZE 10
 
+bool usingLocalPrefs;
```
JuhaSointusalo: Make this a member variable instead (move to header inside `class` and rename to `m_bUsingLocalPrefs`).

clientgui/DlgAdvPreferencesBase.cpp
```diff
@@ -1080,6 +1085,11 @@ bool CDlgAdvPreferencesBase::doesLocalPrefsFileExist() {
     return local_prefs_found;
 }
 
+// to make result available externally
+bool CDlgAdvPreferencesBase::isUsingPrivatePrefs() {
```
JuhaSointusalo: I don't think I've heard them called private prefs before. Maybe `isUsingLocalPrefs` instead?

clientgui/DlgAdvPreferences.cpp
```diff
@@ -1148,6 +1153,18 @@ void CDlgAdvPreferences::OnOK(wxCommandEvent& ev) {
     ev.Skip();
 }
 
+bool CDlgAdvPreferences::ConfirmSetLocal() {
+    wxString strMessage     = wxEmptyString;
+	strMessage.Printf(
+            _("Changing to use the local BOINC preferences defined on this page. BOINC will ignore your web-based preferences, even if you subsequently make changes there. Do you want to proceed?")
```
JuhaSointusalo: @CharlieFenton Does this need branding?
@RichardHaselgrove Or maybe you can find a good wording that doesn't mention BOINC?

davidpanderson: "Changing to use" is awkward.
Can omit "BOINC".
Can omit "defined on this page".
Could say "This will override web-based preferences".

RichardHaselgrove:
It's always better if you don't try to write the words in the same session as wrestling with the code. I'll look at both of these in the morning.

CharlieFenton:
> Does this need branding?

Yes; anywhere "BOINC" or "BOINC Manager" appears in the UI, it needs to allow for branding.

JuhaSointusalo:
@RichardHaselgrove 

I forgot earlier. Could you make the same changes to simple GUI?

clientgui/DlgAdvPreferencesBase.cpp
RichardHaselgrove: I'm afraid I've completely failed to work out how to add a boolean value to the class definition - I only need to export one lousy bit! It should be wxBool or something, but searching the online manual only finds wxBook or books about wool.

clientgui/DlgAdvPreferencesBase.cpp
RichardHaselgrove: You're right. Changed.

clientgui/DlgAdvPreferences.cpp
RichardHaselgrove: I've removed the need for branding, and used override instead of ignore. I'd like to keep 'defined on this page', because this whole PR is addressed at users encountering the difference between web and local preferences for the first and only time.

I haven't found a better alternative for 'Changing to use', but it's up for better suggestions.

RichardHaselgrove:
I've sorted out 'spaces for tabs', and found out where Notepad++ keeps the default - so many standards to choose from! The global tidyup has caught several more instances.

I'll do the simple GUI version as a separate PR later, when we're all happy with this one.

JuhaSointusalo:
> I'll do the simple GUI version as a separate PR later, when we're all happy with this one.

Ok.

JuhaSointusalo:
About naming variables. BOINC code base isn't too consistent in that but I think this `m_bUsingLocalPrefs` style is used more often in Manager code than other styles.

JuhaSointusalo:
Code looks fine. I'm not merging yet because I don't know if we have reached an agreement on the wording.

clientgui/DlgAdvPreferences.h
```diff
@@ -28,35 +28,36 @@
...
-	bool EnsureTabPageVisible(wxTextCtrl* txtCtrl);
-	bool ConfirmClear();
-	wxString DoubleToTimeString(double dt);
-	double TimeStringToDouble(wxString timeStr);
+    bool IsValidTimeChar(const wxChar& ch);
+    bool IsValidTimeValue(const wxString& value);
+    void ShowErrorMessage(wxString& msg,wxTextCtrl* errorCtrl);
+    bool EnsureTabPageVisible(wxTextCtrl* txtCtrl);
+    bool ConfirmClear();
+    bool ConfirmSetLocal();
```
AenBleidd: I think only this line is added to this document. other changes are because of different settings for tabulations and blanks used on your environment. Please don't blame me that I'm picky but could you please revert unnecessary changes from this file?

clientgui/DlgAdvPreferencesBase.cpp
```diff
@@ -161,13 +161,16 @@ CDlgAdvPreferencesBase::CDlgAdvPreferencesBase( wxWindow* parent, int id, wxStri
...
+    }
     buttonSizer->Add( m_btnOK, 0, wxALL, 5 );
 
     m_btnCancel = new wxButton( m_panelButtons, wxID_CANCEL, _("Cancel"), wxDefaultPosition, wxDefaultSize, 0 );
     m_btnCancel->SetToolTip( _("Close the dialog without saving.") );
-    m_btnCancel->SetDefault();
-
+    if (!m_bUsingLocalPrefs) {
+        m_btnCancel->SetDefault();
+    }
```
AenBleidd: Can this block be combined with the block on line 164 to be smth like this:

```
if (m_bUsingLocalPrefs) {
    m_btnOK->SetDefault();
} else {
    m_btnCancel->SetDefault();
}
```

I think this will make the code more clear for understanding

clientgui/DlgAdvPreferences.h
RichardHaselgrove: @JuhaSointusalo asked me, in effect, to ensure that my own changes complied with BOINC's  [Coding Style](https://boinc.berkeley.edu/trac/wiki/CodingStyle#indent). I've seen other committers clean up other files when they happen to revisit them, so it seemed 'the thing to do' while I was here, but I can remove those changes if people feel I've mis-interpreted the policy.

clientgui/DlgAdvPreferencesBase.cpp
RichardHaselgrove: I tried that first of all, but it didn't work. My conclusion was that the SetDefault action had to take place between the 'new' and 'add' actions. In the code we both inherited, all the actions for each of the three buttons were grouped together: OK/Save, Cancel, Help each went through their new/tooltip/default/add sequence. To use if...else  (it appeared to me) that we would have to re-sequence all the actions (do all the news, then all the sets, then all the adds) which was a bigger change than I felt confident making.

clientgui/DlgAdvPreferences.h
JuhaSointusalo: When code and whitespace changes are mixed together it becomes harder to see the code changes. So it's usually better to do whitespace only changes separately. If the lines you are editing have wrong indentation then of course fix the indentation too.

I could have said something about it but in trying to balance between relaxed, everything goes reviewer and, the more typical of me, hard-ass nitpicker I decided to give you some slack. This is a somewhat small PR and the whitespace changes are fairly isolated from code changes. Since the whitespace changes are already in I can live with them continuing to be here.

If you do remove the whitespace changes could you squash the commits or make a new, clean PR? Otherwise this gets a bit silly.

clientgui/DlgAdvPreferencesBase.cpp
AenBleidd: Sorry, I didn't know about this. So I'm OK with this.

JuhaSointusalo:
The current wording is
> Changing to use the local preferences defined on this page. This will override your web-based preferences, even if you subsequently make changes there. Do you want to proceed?

Nobody has said they don't like it so maybe we can assume everyone is happy with it and proceed with getting this PR merged.

@RichardHaselgrove

@AenBleidd made some requests. Could you address them?

RichardHaselgrove:
I think @AenBleidd and I went through all the queries and agreed at least a reason for leaving them as they stand. The wording was discussed in the message board thread and, again, although there were minor differences of opinion, everyone accepted that the current wording was close enough. I'd like to think we could merge and move on to the simple GUI case?

RichardHaselgrove:
Thanks. I'll work on the simple GUI tomorrow (Monday 10 Sept) and hope to have a PR ready the same day - it would be good to have both versions in sync before 7.14 is branched.

Issue #2515, opened 2018-05-14
Android:  Invalid resource directory name


Rudloff:
Hello,

I'm trying to build 8d90c57ca963067ac0a78e6f3198f62b4cec4120 for Android.
I get the following error when I run `./gradlew assemble` in `android/BOINC`:
```
Execution failed for task ':app:mergeDebugResources'.
> /home/pierre/www/boinc/android/BOINC/app/src/main/res/values-sr@latin: Error: Invalid resource directory name
```

PR #2586, opened 2018-07-03
Android, Fix issue 2515

blackleg:
Fix [Issue 2515](https://github.com/BOINC/boinc/issues/2515)

The `values-sr@latin` localization folder isn't a correct android resource, for Serbian with Latin script is `values-b+sr+Latn`. 

https://stackoverflow.com/questions/15518597/how-to-include-2-variants-of-serbian-language-with-latin-letters-and-with-cyril


blackleg:
@brevilo I split the change to values-az in a new pull request:  https://github.com/BOINC/boinc/pull/2605

blackleg:
@brevilo 
* How about earlier versions?
Earlier versions doesn't support Serbian with Latin script without a workaround.

* AFAIK, we do officially support versions older than that?
According the android project the oldest version supported is android 4.1

* How does you implementation behave on such systems?
I checked it in a Android 4.4 device that not support Serbian and it is working.

brevilo:
> I checked it in a Android 4.4 device that not support Serbian and it is working.

Thanks, that what I wanted to know. So to ensure Android 4.1 compatibility of your implementation it would be nice to have someone confirm that.

blackleg:
> So the remaining question is, does BOINC support anything older than Android 4.4?
The oldest version of android supported by Boinc is the 4.1 version.

brevilo:
Right, and as far as I understand your implementation is compatible with Android >= 7 only. Despite your check on Android 4.4 we don't yet know whether it's also compatible with Android 4.1, 4.2 and 4.3.

blackleg:
@brevilo I will try to test this into an android 4.1 version. Could you review this pull request? https://github.com/BOINC/boinc/pull/2605

blackleg:
I have tested the changes into a virtual android device with 4.1 version and doesn't have any error about this changes.

brevilo:
Ok, thanks.

PR #2588, opened 2018-07-07
add virtualization info to host info page

marius311:
This adds a row to the host info page that tells the user what version of Virtualbox they have, if they have hardware virtualization support available and whether it is enabled. Having this should make it much easier for users to debug problems stemming from the fact that their CPU doesn't have hardware support or they have it disabled.

### DIFF
--- a/html/inc/host.inc
+++ b/html/inc/host.inc
@@ -251,15 +251,26 @@ function host_nresults($host) {
     return BoincResult::count("hostid=$host->id");
 }
 
+
+// Parse the Virtualbox version information from inside the serialnum field.
+// Prior to BOINC commit 6121ce1, the DB entry looked like e.g. "[vbox|5.0.0]"
+// where Virtualbox 5.0.0 gave the version number. After 6121ce1, the entry was
+// "[vbox|5.0.0|1|1]", where now two additional flags give information about
+// hardware virtualization support. Older clients may have the old-style
+// serialnum in the DB despite the server being upgraded.
 function vbox_desc($x){
-    if (preg_match("/\[vbox\|(.*)\|([01])\|([01])\]/",$x,$matches)){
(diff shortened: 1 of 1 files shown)


TheAspens:
@marius311 

I tested this but it I wasn't able to get it to show any information.  I pulled various strings from WCG production (see the list below) and these always rendered as "Virtualization: None".

- [BOINC|6.12.34][CUDA|GeForce 8600M GT|1|512MB|18681][vbox|4.1.2]
- [BOINC|7.10.2][CAL|AMD Radeon HD - FirePro D700 Compute Engine|2|6144MB|][vbox|5.1.22r115126]
- [BOINC|7.6.22][CUDA|Quadro K600|1|1024MB|34066][vbox|5.1.6]
- [BOINC|6.12.34][vbox|4.0.6]

I tested by updating the serialnum field on host entry that was created via a device attach using my docker server test setup.

In all cases the GPU and the BOINC version updated on the page correctly.

marius311:
Thanks for testing. I'm guessing you guys don't have https://github.com/BOINC/boinc/commit/6121ce1b580634037249055268aec0cf212de39e ? That turned the serialnum style from e.g. `[vbox|5.1.22r115126]` to `[vbox|5.1.22r115126|x|x]` where the final two flags tell you about hardware extensions. Or maybe you do and you just got unlucky and didn't grab any newer style serialnum's from your DB?

In any case, I should add support older style serialnums, its easy enough and lots of projects may have users with older entries in the DB. Will ping you once thats done.

TheAspens:
We are on 3 year old server code so that would be why.  It would be good to support the older formats as there will be a lot of old records in the db.  Thanks for being willing to add them.

marius311:
Ready for further review.

TheAspens:
Looks good - thanks!

Issue #2583, opened 2018-07-01
Daily schedules not allowing to select the entire day


maciejdlugosz:
In Options->Computing preferences->Daily schedules it is impossible to set a day, when computation or transfering are not limited. E.g. we would like to compute on every day only at night, but on weekend the entire day. Before #2044 it was possible to pass a range 0:00 to 24:00.

Passing 0:00 to 23:59 is not a huge cost until one has "Leave non-GPU tasks in memory while suspended" checked.

I suggest not to add a new checkbox, but rather allow passing "0:00 to 24:00" or more intuitively "0:00 to 0:00".

AenBleidd:
@CharlieFenton, 
why did this change was done 37ae3de4 ?
It looks like that next code accepts equal start and end time without any problems:

```
bool TIME_SPAN::suspended(double hour) const {
    if (start_hour == end_hour) return false;
    if (start_hour == 0 && end_hour == 24) return false;
    if (start_hour == 24 && end_hour == 0) return true;
    if (start_hour < end_hour) {
        return (hour < start_hour || hour > end_hour);
    } else {
        return (hour >= end_hour && hour < start_hour);
    }
}

TIME_SPAN::TimeMode TIME_SPAN::mode() const {
    if (end_hour == start_hour || (start_hour == 0 && end_hour == 24)) {
        return Always;
    } else if (start_hour == 24 && end_hour == 0) {
        return Never;
    }
    return Between;
}
```

CharlieFenton:
@AenBleidd 

> why did this change was done 37ae3de ?

Change 37ae3de has nothing to do with this PR. Even though _GIT blame_ lists me as the author of that code, commit 98cfb8d on 26 September 2008 lists @davidpanderson as the author of this code. But please note that the code you reference above is is not called by the Manager's dialog. I also don't see anywhere in the client or Manager where this code is called. 

As stated in [the original complaint](https://github.com/BOINC/boinc/issues/2583#issue-337308547), the issue was caused by your PR  #2044 (and related PR #2083) and your commits f39005c, 7b8cfd6 and 7b8cfd6. I believe you should change this:
```
    //verify hours
    if (h < 0 || h > 23) {
        return false;
    }

```
to this:
```
    //verify hours
    if (h < 0 || h > 24) {
        return false;
    }

```
The 24 case should be handled by this code (to ensure that it is _24:00_:
```
    //all chars are valid, now what is with the value as a whole ?
    wxDateTime dt;
[...]

PR #2589, opened 2018-07-07
[Manager] Fix time verification to allow '24:00' value in Preferences…

AenBleidd:
… dialog

Time verification simplification.


AenBleidd:
This fixes #2583 and contains patch made by @CharlieFenton

AenBleidd:
Yes, I verified and the fix works correctly

PR #2591, opened 2018-07-07
[Client] Change default end time in daily schedules from 23:59 to 24:00

AenBleidd:

### DIFF
--- a/lib/prefs.cpp
+++ b/lib/prefs.cpp
@@ -284,9 +284,9 @@ void GLOBAL_PREFS::enabled_defaults() {
     max_bytes_sec_down = 100*KILO;
     max_bytes_sec_up = 100*KILO;
     cpu_times.start_hour = 0;
-    cpu_times.end_hour = 23.983;    // 23:59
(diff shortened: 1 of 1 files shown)


AenBleidd:
Small fix related to #2583 and #2589

CharlieFenton:
@davidpanderson: I don't know whether the web-based preferences also have a default value which needs changing. But I just now looked at the computing preferences on the SETI@home website and see that the "Compute only between" and "Transfer files only between" fields allow only times on the hour (0:00, 1:00, 2:00 ... 23:00.) There is no way to enter times like 3:30.

AenBleidd:
@TheAspens, @RichardHaselgrove, @davidpanderson, could you please take a look at this PR and decide whether it can be merged or not. Thanks

TheAspens:
@AenBleidd - looks good - thanks for the fix.  I reviewed it and also noted that @CharlieFenton set those defaults originally so since he approved the change as well I think this is good.

The comments about the web interface should be handled in a different issue/pull request if a change is required.

PR #2576, opened 2018-06-26
Generic Mac build script for branded version of Mac manager/screen saver

armstrdj:
@CharlieFenton here is the current script. I still need to look at your suggestion for PostInstall.cpp to fix finding the branding file in the installer. Also I need to add support for setting up an account manager in the script. We have not finalized all the images/content for the WCG branded installer/manager so I have not checked in all of those images yet. 

Also I am still looking into some issues that I think are unrelated to the installer but may be issues in general with the manager or installer code.
1. The snooze/disconnect image is not showing up over the WCG icon in the taskbar.
2. The logo in the upper right notification bar uses the appropriate WCG logo defined in the skin. However if I switch the skin to default it properly switches the logo to boinc but does not switch the logo back when switching back to WCG skin.
3. The BOINC Installer icon displays in the screen saver popup during install instead of the bundle branded installer icon.

PR #2596, opened 2018-07-11
Manager: Fix branded link in About Dialog

CharlieFenton:
Manager: in the About dialog "For more information, visit" hyperlink, set the value of the link to the same string as the text displayed for the link when using a branded skin.

NOTE: Despite the "mac" in the name of this branch, this fix is needed for all platforms.


CharlieFenton:
@armstrdj Note that this fixes an omission from your PR #2576.

PR #2597, opened 2018-07-11
Mac: Consolidate branding strings

CharlieFenton:
Mac: Consolidate branding strings in one source file and one header file shared among all files which use them; add checks for missing branding strings.

PR #2603, opened 2018-07-14
Add support for new nvc_config.xml file for branded installers

CharlieFenton:
Move client_version_check_url, client_download_url and client_new_version_text tags from cc_config.xml into new nvc_config.xml file.

If nvc_config.xml file is absent, use default values.
Branded installers can create or replace this file to customize these values.
Standard (unbranded) BOINC installers should either delete the file or create or replace it with one containing default values.


CharlieFenton:
I still need to update the mac_installer/Postinstall.cpp, and lib/mac/mac_branding.cpp,.h files to add support for this to the Mac installer, but I am waiting for someone to merge my PR #2597 to ensure there are no merge conflicts. I will then test this code.

CharlieFenton:
There is a problem with the current implementation of the _client_new_version_text_ tag, which is used in the text of the notices the client generates when a new version is available. When the tag is empty or absent, the string: `_("A new version of BOINC is available")` is used, which is translatable via Transifex. But if the _client_new_version_text_ tag has a custom string, the tag's value replaces the entire string I just quoted. The problem is that the tag's value cannot be translated since it is not compiled with the client.

I propose changing the way this tag is used, so it just contains the text to replace `BOINC` in the notice. Il the tag is not empty, then the notice would display the string `_("A new version of %s is available")` where the `%s` is filled in with the contents of the tag. So for a WCG branded installer, the nvc_config file would contain `<client_new_version_text>World Community Grid</client_new_version_text>` so the notice would say _A new version of World Community Grid is available_, which would be fully translatable.

@davidpanderson, @armstrdj, @AenBleidd and anyone else: Does anyone feel this change in the use of the _client_new_version_text_ tag is not a good idea?

CharlieFenton:
A couple of notes about this PR:

[1] Regarding the `new_version_text` tag, I wrote:
> I propose changing the way this tag is used, so it just contains the text to replace BOINC in the notice. Il the tag is not empty, then the notice would display the string _("A new version of %s is available") where the %s is filled in with the contents of the tag.

I have made this change. Consistent with this change, I renamed the  _NVC_CONFIG_ struct member `new_version_text` to `new_version_name`.

[2] As in the past, when the client detects that a newer version is available, it writes the _<newer_version>_ tag into the client_state file with the newly available version number. Each subsequent time the client is launched, it calls `newer_version_startup_check()` at startup which checks for this tag and posts the new version message in the _Notices_ tab. But the newest available version may differ between brands, so if a different brand of BOINC is installed, that message may no longer apply.

To deal with this, the installer must rename any previous _nvc_config.xml_ file to _old_nvc_config.xml_ before installing the new one. I have added code in `newer_version_startup_check()` to compare the old and new `client_version_check_url` fields. If they are different, it deletes the _<newer_version>_ tag from the client_state file. This logic also correctly handles the case where there is either not an old or not a new _nvc_config.xml_ file. The absence of this file implies "use standard BOINC defaults."

I have already added code to the Mac installer to rename any previous _nvc_config.xml_ file to _old_nvc_config.xml_; this **must** also be done for the installers on MS Windows and any other platforms which might have branded builds. I do not have the expertise to do it for platforms other than the Mac.

CharlieFenton:
> I still need to update the mac_installer/Postinstall.cpp, and lib/mac/mac_branding.cpp,.h files to add support for this to the Mac installer, but I am waiting for someone to merge my PR #2597 to ensure there are no merge conflicts. I will then test this code.

I thought of a better way to accomplish what I wanted, so there this PR has no files common with that one. My testing is complete. If the requested reviewers approve this PR, then it is ready to be merged into master.

Once this has been merged, I will update the [cc_config documentation](http://boinc.berkeley.edu/wiki/Client_configuration).

CharlieFenton:
@armstrdj  Please also review this.

CharlieFenton:
@TheAspens, @davidpanderson, @armstrdj: The current WCG branded installer for MSWindows also sets (or tries to set) `<network_test_url>http://www.ibm.com</network_test_url>`. Should we move the `network_test_url` tag to the new nvc_config.xml file?

Note my earlier comments on the boinc_dev email list, which is why I write "(or tries to set)":
> I ran some tests on what the current WCG branded MS Windows installer does. I found several problems with it:
> If there is no cc_config.xml file, the WCG installer correctly creates one with only <client_version_check_url>, <client_download_url> and <network_test_url> tags. 
> If there is a pre-existing cc_config.xml file with the WCG values for the <client_version_check_url> and <client_download_url> tags, it correctly leaves the file untouched.
> But if the pre-existing cc_config.xml file has different values for those two tags, it deletes the entire file and replaces it with one containing only <client_version_check_url>, <client_download_url> and <network_test_url> tags. This destroys any log flags or options that were previously set.
> And if the pre-existing cc_config.xml file does not have either of the two tags, it incorrectly leaves the file untouched, causing it to use the wrong (default unbranded BOINC) URLs when checking for a new version.
>

armstrdj:
@CharlieFenton I tested and didn't find any issues. We will continue to set the network_test_url to ibm.com so from my perspective it makes sense to move it also to nvc_config.

ChristianBeer:
Point for Continuous Integration testing! Last commit broke Windows build with:
```
>..\client\sysmon_win.cpp(303): error C2039: 'network_test_url' : is not a member of 'CC_CONFIG'
>          C:\projects\boinc\lib\cc_config.h(144) : see declaration of 'CC_CONFIG'
>..\client\sysmon_win.cpp(303): error C2228: left of '.c_str' must have class/struct/union
```

CharlieFenton:
@davidpanderson and @TheAspens: This PR is complete and awaiting your review and merging.

TheAspens:
@CharlieFenton - thanks for doing this work.  I checked in with @armstrdj  and he built and tested it.

CharlieFenton:
PR #2623 eliminates the need for the installer to rename a previous nvc_config.xml file to old_nvc_config.xml, superseding part [2] of [this comment above](https://github.com/BOINC/boinc/pull/2603#issuecomment-405083071).

PR #2606, opened 2018-07-15
web: code cleanup in delete user area

davidpanderson:
- functions generally return zero on success
- improve function names
- comment formatting

See https://boinc.berkeley.edu/trac/wiki/CodingStyle

todo: fix translatable strings


davidpanderson:
translatable strings should not contain HTML tags.
Do something like

tra("blah blah %1link%2", "<a href=foo>", "</a>")

PR #2608, opened 2018-07-17
update installshield files; update version #s; fix Win custom action build

davidpanderson:


win_build/installerv2/BOINC.ism
```diff
@@ -2461,7 +2461,7 @@
 		<row><td>WEBx86</td><td>ProductCode</td><td/></row>
 		<row><td>WEBx86</td><td>ProductName</td><td/></row>
 		<row><td>WEBx86</td><td>ProductVersion</td><td/></row>
-		<row><td>WEBx86</td><td>SetupFileName</td><td>boinc_7.7.2_windows_intelx86</td></row>
+		<row><td>WEBx86</td><td>SetupFileName</td><td>boinc_7.12.1_windows_intelx86</td></row>
```
TheAspens: Should these be 7.13.1? (and the others in the other .ism files)

TheAspens: Or rather 7.13.0 to match version.h

davidpanderson: I'll update those when I actually make installers

Issue #2449, opened 2018-04-05
EU-GDPR - Privacy by design and default


brevilo:
Let's use this issue for privacy-by-design-related topics that need to be addressed. I start with this one:

* On account creation [the user's computers are publicly visible](https://github.com/BOINC/boinc/blob/master/html/inc/user_util.inc#L76) by default. Needs to be changed/inverted (`show_hosts`).
* FYI, we don't consider `send_email` to be problematic is this doesn't cause personal data to be shared.

TheAspens:
I also think that when it comes to the data export, there should be a flag the user can opt into that indicates that the user agrees to export their data.   If not set, then the data should not be exported.

In order to mitigate the fact that this would cause significantly fewer devices and users to be exported over time, we could add a summary.xml to the export that includes:
- Total credit on project (sum of user.total_credit)
- Total recent credit on project (sum of user.expavg_credit)
- Total number of users on project (count of user.total_credit > 0)
- Total number of active users on project (count of user.expavg_credit > 1)
- Total number of hosts on project (count of host.total_credit > 0)
- Total number of active hosts on project (count of host.expavg_credit > 1)

Stats sites could then use this data to provide aggregate data graphs for the project.

RichardHaselgrove:
Remember that BOINC itself participates in this merry-go-round of exported and re-imported stats to drive pages like https://boinc.berkeley.edu/chart_list.php

I think the data is aggregated by http://boinc.netsoft-online.com/e107_plugins/boinc/bp_home.php (another page which can't be accessed by https, it seems), which also provides the data for the 'Projects in which Richard Haselgrove is participating' panel on public pages like https://setiathome.berkeley.edu/show_user.php?userid=5509 - I'm not sure whether WCG participates in this process, so @TheAspens may not be so familiar with it.

Provided netsoft-online updates its code and processes along the lines we've been discussing, proper GDPR compliance should follow automatically, but it might be wise to reach out to James Drews sooner rather than later, because his site is rather less independent from BOINC than some of the others.

brevilo:
The good things is that as soon as we lock down our stats exports, everyone we missed will surely be in contact very soon. But yes, Willy and James are on our radar. Thanks.

AenBleidd:
Is there something left here or this ticket can be closed?

drshawnkwang:
The main item for this ticket, setting `show_host` to false by default has been merged. I will ask at the next contributors call whether there are other privacy options people care about.

[...]

PR #2610, opened 2018-07-18
user: set show_host default to false.

drshawnkwang:
This PR changes a single flag the BoincUser::insert statement, setting the flag `show_host` to false. Thus by default a user's hosts are not in publicly view.

Part of #2449

### DIFF
--- a/html/inc/user_util.inc
+++ b/html/inc/user_util.inc
@@ -92,6 +92,13 @@ function is_valid_user_name($name, &$reason) {
     return true;
 }
 
+function default_show_hosts() {
+    global $config;
+    // If enable privacy by default is TRUE, then show_hosts' default
+    // is FALSE.
+    return parse_bool($config, "enable_privacy_by_default") ? 0 : 1;
+}
+
 // the following DB-escapes its args
 //
 function make_user(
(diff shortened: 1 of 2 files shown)


davidpanderson:
should probably be a config flag

drshawnkwang:
@davidpanderson , I don't disagree. Yet if we go that route we should decide 1) does this one database option get its own config flag 2) or should we have a 'privacy' config flag that controls this database option as well as other GDPR/ privacy related items. Which then leads to 2a) what are the other GDPR privacy related options? Just some food for thought.

Obviously the PR was meant to be the 'simplest' change possible. And I haven't heard much from other people regarding #2449 . Maybe it's something to bring up in a contributors call.

drshawnkwang:
I added a configuration option, `enable_privacy_by_default`, in order to control the default behavior of `show_hosts`. If `enable_privacy_by_default` is 0 (default), then `show_hosts` is 1. And vice-versa.

Other privacy related options could be added to this config.xml option in the future as we see fit.

drshawnkwang:
@lfield - This could probably go into the upcoming server release. I don't think I have the ability to tag or add it to a project board.

drshawnkwang:
@davidpanderson , can I ask you to (re-)review this PR? It would be nice to have this code merged in as part of the first server release.

davidpanderson:
Better implementation (avoid duplication):

function default_show_hosts() {
    global $config;
    return parse_bool($config, "enable_privacy_by_default")?0:1;
}

...
BoincUser::insert( .... default_show_hosts()...)

drshawnkwang:
Good idea David. I've marked this as WIP and will try to add the code changes soon.

drshawnkwang:
Refactored the code to check the project's config in a function.

TheAspens:
I updated https://boinc.berkeley.edu/trac/wiki/ProjectOptions and http://boinc.berkeley.edu/trac/wiki/GdprCompliance with this field.

PR #2616, opened 2018-07-20
script_validator segfault fix

AenBleidd:
Current script_validator implementation contains next statements:

`sprintf(cmd, "../bin/%s", init_script[0].c_str());`

and

`sprintf(cmd, "../bin/%s", compare_script[0].c_str());`

It is dangerous to use this arrays
that are filled by parsing command line
without validating whether they contain any elements
and leads to next behavior:

Program received signal SIGSEGV, Segmentation fault.
```
0x000000000040f838 in compare_results (r1=..., r2=..., match=match@entry=@0x7ffffffb6c17: false) at script_validator.cpp:154
154	script_validator.cpp: No such file or directory.
(gdb) bt
#0  0x000000000040f838 in compare_results (r1=..., r2=..., match=match@entry=@0x7ffffffb6c17: false) at script_validator.cpp:154
#1  0x000000000040e85d in check_set (results=std::vector of length 2, capacity 2 = {...}, wu=..., canonicalid=@0x7ffffffb6dd0: 0,
    retry=@0x7ffffffb6dcf: false) at validate_util2.cpp:134
#2  0x000000000040944e in handle_wu (validator=..., items=std::vector of length 3, capacity 4 = {...}) at validator.cpp:411
#3  0x000000000040a9f2 in do_validate_scan () at validator.cpp:730
#4  0x000000000040aae4 in main_loop () at validator.cpp:756
#5  0x00000000004039f6 in main (argc=5, argv=0x7fffffffdd48) at validator.cpp:929
```

This fix checks whether these arrays contain some elements
and output a readable error of missed inout parameter.

### DIFF
--- a/sched/script_validator.cpp
+++ b/sched/script_validator.cpp
@@ -76,9 +76,9 @@ int validate_handler_init(int argc, char** argv) {
         }
     }
 
-    if (!init_script.size() && !compare_script.size()) {
+    if (!init_script.size() || !compare_script.size()) {
         log_messages.printf(MSG_CRITICAL,
-            "script names missing from command line\n"
+            "init_script and/or compare_script names are missing from command line\n"
         );
         return 1;
     }
--- a/sched/script_validator.cpp
+++ b/sched/script_validator.cpp
@@ -112,7 +112,7 @@ int init_result(RESULT& result, void*&) {
     }
 
     if (init_script.size() == 0) {
-        fprintf(stderr, "init_result() failed: init_script parameter was not specified\n")
+        fprintf(stderr, "init_result() failed: init_script parameter was not specified\n");
         return 1;
     }
 
@@ -158,7 +158,7 @@ int compare_results(RESULT& r1, void*, RESULT const& r2, void*, bool& match) {
     }
 
     if (compare_script.size() == 0) {
-        fprintf(stderr, "compare_results() failed: compare_script parameter was not specified\n")
+        fprintf(stderr, "compare_results() failed: compare_script parameter was not specified\n");
         return 1;
     }
 
--- a/sched/script_validator.cpp
+++ b/sched/script_validator.cpp
@@ -110,6 +110,12 @@ int init_result(RESULT& result, void*&) {
         fprintf(stderr, "get_output_file_paths() returned %d\n", retval);
         return retval;
     }
+
+    if (init_script.size() == 0) {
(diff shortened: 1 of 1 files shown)


marius311:
Thanks for the quick reply, I think that's exactly it. 

I think the solution though is instead fixing:
https://github.com/BOINC/boinc/blob/fa6c27805dafe208353c714e5f74d8fc25c49aba/sched/script_validator.cpp#L79

It looks like the logical is just screwed up, should be an `or` instead of an `and`, no? I think validating the arguments up-front is better too, otherwise the program could run for a long time before alerting the user the arguments were wrong, and that would be easy to miss (as I apparently have had).

AenBleidd:
@marius311,
The statement you mention probably should be changed too but I'm not the right guy to decide this because I'm not awared about the logic of this script. I can fix that too but please take a look at the description:

https://github.com/BOINC/boinc/blob/fa6c27805dafe208353c714e5f74d8fc25c49aba/sched/script_validator.cpp#L39-L42

It is said that this parameter is additional so it can be the case when they are not used. It is also a reason why I didn't add verification in `validate_handler_init()` function but closer to the function when these arguments are really used.

marius311:
To me it seems clear that line was meant to ensure both `init_script` and `compare_script` arguments were provided (since the script is not written to handle anything otherwise). This is in line with the docs which say "script_validator takes two additional command-line arguments" (https://boinc.berkeley.edu/trac/wiki/ValidationSimple). However, due to minor logic flaw it seems to only check that one or the either is provided. So I think the only thing this PR needs to do is to turn that `&&` into a `||` .

AenBleidd:
@marius311, ok, done, but I left verification in-place also just in case someone will decide to call these functions without 'validate_handler_init()' function call.

marius311:
I don't mean to picky, just to keep the codebase as clear as possible, but why do we need the two redundant checks and not just the one in validate_handler_init? The validate_handler_init function will always be called first unless the user somehow compiles without using validator.cpp but does use script_validator.cpp and then skips calling validate_handler_init, which seems to me such an obscure / difficult thing to do that we can basically disregard it, no?

AenBleidd:
I prefer to keep such redundancy because this order can be changed. We do not have here data encapsulation and access control: every function that uses a data from outside without any verification can be called from everywhere. And this can lead to the issue you faced. I'm pretty sure that this additional 'in-place' check will not have so much effort on performance to remove it and keep dangerous code.

PR #2624, opened 2018-07-26
[WIP]: add detection of physical cores and ability to limit jobs to the number of physical (rather than logical) cores

### LINKED ISSUE #2542: vboxwrapper should use ncpus = physical cores
From https://www.virtualbox.org/manual/ch03.html:

> You should not, however, configure virtual machines to use more CPU cores than you have available physically (real cores, no hyperthreads).

However currently vboxwrapper uses the number logical cores. There is some discussion of this in https://github.com/BOINC/boinc/issues/1501 and https://github.com/marius311/boinc-server-docker/issues/33.

I had thought this was some minor sub-optimality, but thanks to @zonca getting a totally reproducible scenario working, I recently made the connection that this could be causing a significant number of the failures we see at C@H which I had previously not understood. 

I don't have a solution, see https://github.com/BOINC/boinc/issues/1501 for discussion of the issues involved in getting the number of physical cores (anyone know a good API?). I'm opening this though so we can brainstorm / track progress on this specific issue.

marius311:
WIP, do not merge. 

- [ ] Client detection
  - [x] Linux
  - [ ] Windows
  - [ ] Mac
- [x] Receive info on server, update DB, and display in host info pages
- [ ] Add planclass (?) to allow limiting to physical cores
- [ ] Use this info in scheduling

Fixes #2542

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -39,6 +39,8 @@
 #include <vector>
 #include <string>
 #include <cstring>
+#include <tuple>
+#include <set>
 #endif
 
 #if HAVE_XSS
@@ -122,6 +124,9 @@
 #include "hostinfo.h"
 
 using std::string;
+using std::tuple;
+using std::make_tuple;
+using std::set;
 
 #ifdef __APPLE__
 #include <IOKit/IOKitLib.h>
@@ -463,6 +468,8 @@ static void parse_cpuinfo_linux(HOST_INFO& host) {
     bool model_info_found=false;
 #endif
     char buf2[256];
+    int physical_id=-1, core_id=-1;
+    set<tuple<int,int>> unique_physid_coreid;
 
     FILE* f = fopen("/proc/cpuinfo", "r");
     if (!f) {
@@ -663,7 +670,29 @@ static void parse_cpuinfo_linux(HOST_INFO& host) {
                 features_found = true;
             }
         }
+        
+        // /proc/cpuinfo prints out a "physical id" (this is a misleading name,
+        // its actualy the socket id) and "core id", for each logical CPU. two
+        // logical CPUs that are just hyperthreads on the same physical CPU will
+        // have identical "physical id" / "core id", so we can get the number of
+        // physical CPUs by counting the number of *unique* "physical id" /
+        // "core id" pairs
+        if (strstr(buf, "physical id")) {
+            physical_id = atoi(strchr(buf, ':') + 2);
+        }
+        if (strstr(buf, "core id")) {
+            core_id = atoi(strchr(buf, ':') + 2);
+            unique_physid_coreid.insert(make_tuple(physical_id, core_id));
+        }
     }
+    
+    // here we total the number of unique "physical id" / "core id" entries
+    // (unique_physid_coreid is a set thus contains no duplicates). if no
+    // entries were found, here we will set p_ncpus_phys = 0, which is an
+    // indication detection failed (eg ARM/PowerPC/HPPA which dont have
+    // hyperthreading)
+    host.p_ncpus_phys = unique_physid_coreid.size();
+
     safe_strcpy(model_buf, host.p_model);
 #if !defined(__aarch64__) && !defined(__arm__)
     if (family>=0 || model>=0 || stepping>0) {
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -203,10 +203,17 @@ void CLIENT_STATE::show_host_info() {
         host_info.domain_name
     );
     nbytes_to_string(host_info.m_cache, 0, buf, sizeof(buf));
-    msg_printf(NULL, MSG_INFO,
-        "Processor: %d %s %s",
-        host_info.p_ncpus, host_info.p_vendor, host_info.p_model
-    );
+    if (host_info.p_ncpus_phys == 0) {
+        msg_printf(NULL, MSG_INFO,
+            "Processor: %d logical CPUs, %s %s",
+            host_info.p_ncpus, host_info.p_vendor, host_info.p_model
+        );
+    } else {
+        msg_printf(NULL, MSG_INFO,
+            "Processor: %d logical / %d physical CPUs, %s %s",
+            host_info.p_ncpus, host_info.p_ncpus_phys, host_info.p_vendor, host_info.p_model
+        );
+    }
     if (ncpus != host_info.p_ncpus) {
         msg_printf(NULL, MSG_INFO, "Using %d CPUs", ncpus);
     }
--- a/client/hostinfo_win.cpp
+++ b/client/hostinfo_win.cpp
@@ -1269,21 +1269,92 @@ typedef DWORD (WINAPI *GAPC)(WORD);
 #ifndef ALL_PROCESSOR_GROUPS
 #define ALL_PROCESSOR_GROUPS 0xffff
 #endif
-int get_processor_count(int& processor_count) {
+
+typedef BOOL (WINAPI *LPFN_GLPI)(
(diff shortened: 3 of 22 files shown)


client/hostinfo_unix.cpp
```diff
@@ -39,6 +39,8 @@
 #include <vector>
 #include <string>
 #include <cstring>
+#include <tuple>
```
AenBleidd: I'm not sure this is available in VS2010

marius311: Thanks, didn't know. That's annoying. So the client needs to be buildable with VS2010 on Windows? What's the motivation for supporting such an old version?

marius311: Nevermind, I've got the VS2010 story explained to me at the workshop :) For now I'll get a workaround.

ChristianBeer: `std::tuple` is also a C++11 extension and thus not available for the Linux part too.

davidpanderson:
For a plan class with <physical_cpus>, a job might get --nthreads 4 but use 8 logical CPUs (client scheduling is in terms of logical CPUs).  The volunteer would see that the job uses 8 CPUs, but the CPU meter would show 50% usage.  This may be confusing to overly attentive volunteers.

davidpanderson:
The semantics of <max_threads>N 1</max_threads> is that the job uses the
number of physical CPUs minus one, i.e. leave one free to avoid slowing down desktop response.
If the user has specified "use X% of the CPUs" in their computing prefs with X < 100,
this is probably not the desired behavior; there will already be a free CPU.
I propose changing the semantics so that the minus one only applies to hosts
where all CPUs are usable.

marius311:
Sorry I've not had time to finish this up. I will try this week. 

> I propose changing the semantics so that the minus one only applies to hosts
where all CPUs are usable.

Seems reasonable to me. 

> For a plan class with <physical_cpus>, a job might get --nthreads 4 but use 8 logical CPUs (client scheduling is in terms of logical CPUs). The volunteer would see that the job uses 8 CPUs, but the CPU meter would show 50% usage. This may be confusing to overly attentive volunteers.

Can you explain in more detail why this would happen? I don't quite follow, and this seems a pretty serious deficiency.

davidpanderson:
The Windows Task Manager (the default CPU monitor) shows logical CPU usage.
So if an app is using 4 physical cores, the task manager will show 50% CPU usage.

This is not a big deal, and there's nothing we can do about it except to explain
to users what's going on.

marius311:
> The Windows Task Manager (the default CPU monitor) shows logical CPU usage.
So if an app is using 4 physical cores, the task manager will show 50% CPU usage.

OK, I'm on the same page with that. 

However, the way I was thinking about this is that on a 4 phys core / 8 logical core machine, if the user's CPU preference was to 100%, then they could run e.g.

* two separate nthreads=4 vboxwrapper jobs, or 
* one nthreads=4 vboxwrapper job, and the remaining 4 logical cores could run other BOINC jobs if that's what happened to be available

In either case, the vboxwrapper job should show "running with 4 CPUs" in the boinc manager. 

Basically, I don't think that `<physical_cpus>` should imply that the job needs exclusive use of the entire physical CPU, because, at least as the Virtualbox docs are written, this is not a requirement of Virtualbox. However, I realize now we've never actually tested this. I've asked a colleague to test this here https://github.com/marius311/boinc-server-docker/issues/33#issuecomment-410889201, his reply could inform what we do.

davidpanderson:
The way I implemented things, the client will run only one 4-thread VM;
i.e. the client will treat it like it's using 8 logical CPUs (and will show
it as such).

Changing job scheduling on the client to distinguish between logical and
physical CPUs
would be a large job, which I suggest we defer.

davidpanderson:
I'm not sure the <max_threads>N M<max_threads> semantics are useful after all.
You can tell it to not use a CPU (M=1).
But the client will then run a single-thread job on that CPU.
The only way to keep CPUs free is with the "use at most X% of the CPUs" preference.

marius311:
> But the client will then run a single-thread job on that CPU.

No that's OK, as long as that ends up not being a VBox job. The problem is that the VBox job can't be set to low priority, but a normal native job filling that spot can be, so the user would suffer from less desktop degradation. Lots of C@H users currently have this set up succesfully. 

> The way I implemented things, the client will run only one 4-thread VM;

Yea, I see that now. I admit I don't totally understand scheduling, but why can't the new system simply set, for these vbox jobs, nthreads_max=4, and then let the client pack two of them in like I think would happen with any other two native nthreads=4 jobs ? I worry many users are going to think we're wasting 50% of their resources and hate VBox even more than they already do.

davidpanderson:
We could.

Running 2 VMs at once would have twice the working set size,
which could cause increased paging.
I'm curious to know whether (on a typical Win PC)
- 2 VMs gives more or less throughput
- 2 VMs slows down interactive response.
Is there any way to set up such a test?

davidpanderson:
This question applies to FP-intensive apps in general, on hyperthreaded
machines:
does using all the virtual CPUs give more or less throughput?

ChristianBeer:
From time to time there is some intensive testing in the German Rechenkraft.net forum about the effect of hyperthreading on different CPU and different projects. For VirtualBox projects @JacobWKlein did a good comparison and found out that disabling Hyperthreading helps the VBox jobs to complete faster. This is also the case for other science applications (but not all).

RichardHaselgrove:
The question is frequently discussed at SETI, but without any methodical testing. The general consensus is that with HT active, ordinary CPU tasks run slower, but total throughput is greater because more tasks run concurrently. My personal view is that we ignore the differences between integer-heavy and FP-heavy applications at our peril - especially on hardware with FPU shared between logical cores.

marius311:
I would suggest we decouple,

1) The ability to limit certain jobs to run with a per-job max nthreads = (# of phys cores). This is needed by VBox jobs so they don't outright crash. 

2) The ability for the user or for project admins to (potentially) improve performance for VBox or native jobs by running without hyperthreading. 

I think (1) would should be the goal of this PR, then we can potentially think about (2) (of course we can keep it in mind while coming up with the way to do (1)) @ChristianBeer and @RichardHaselgrove your guys' info is helpful suggesting we could indeed want ((2)). 

So if you agree, then I suggest @davidpanderson change the scheduling to just limit <phys_core> jobs to (# of phys cores), but otherwise allow the client to pack multiple of them together.

lfield:
What is that state of this? There hasn't been any update for a while.

marius311:
If I remember correctly, in writing this I used some C++11 extensions which I didn't know I shouldn't. But it was explained to me that someone was working on updating the compilation process, I think using a newer version of Visual Studio or something like that, so that this would be fine. I don't know what's the status of this work. Does anyone know or can link and Issue/PR?

AenBleidd:
@marius311, #2348
But I'm not sure that there are some changes or finished implementations, @JuhaSointusalo car tell more details.
But meanwhile I'd suggest change the code to be compatible with C++ 2003 to move this functionality further

RichardHaselgrove:
Remember that the current release builds were done with VS2010.

CharlieFenton:
Remember that hostinfo_unix.cpp is also used by Mac builds.

marius311:
> But meanwhile I'd suggest change the code to be compatible with C++ 2003 to move this functionality further

I admit I'm pretty unmotivated to do that. I may try and see if there's anything I can help with in #2348 however, that seems like the better solution anyway.

lfield:
I am closing this PR as more work is needed until it is ready. Discussion can continue on the [issue](https://github.com/BOINC/boinc/issues/2542) while branch [phys_cores](https://github.com/BOINC/boinc/tree/phys_cores) is being worked on.

Issue #1485, opened 2016-01-31
parts of global_prefs get reported as variety in trickle up messages


ChristianBeer:
I just found 26 entries in msg_from_host that accumulated over the last weeks that didn't have the correct variety of "cpu_time" instead they had the beginning of the global prefs XML or a something else stored (I deleted the newlines as they were messing with the layout).

```
| id     | create_time | hostid | variety                                                                                                                                                                                                                                                       | handled | xml                                                                                                                                                                                          |
| 433101 |  1435373824 |  33244 | <source_project>http://bam.boincstats.com/</source_project><mod_time>1406799579</mod_time>                                                                                                                                                                   |       0 | <result_name>cmsvm_GA-p[e30-50MB_Lin64f]_1_Oryza-sativa-Japonica-Group_CM000138.lin.EMBL_RF00177_SSU_rRNA_5_1330438623_83865_63</result_name><time>1435370520</time>         |
| 443986 |  1436319757 |  27843 | <source_project>http://bam.boincstats.com/</source_project><mod_time>1432043467</mod_time><venue name="New: 2010-12-07 07:24:28"><ram_max_used_idle_pct>0</ram_max_used_idle_pct><work_buf_min_days>0.01</work_buf_min_days><cpu_usage_limit>100</cpu_us |       0 |       <result_name>cmsvm2_GA-p[e30-50MB_Lin64f]_1_Anopheles-gambiae-str.PEST_CM000358.lin.EMBL_RF00028_Intron_gpI_1330438623_1400_14</result_name>      <time>1436319500</time>         |
| 447532 |  1436637105 |  13748 | <source_project>http://gerasim.boinc.ru/</source_project>    <source_scheduler>http://gerasim.boinc.ru/sched</source_scheduler>    <mod_time>1404481426</mod_time>   <cpu_scheduling_period_minutes>120</cpu_scheduling_period_minutes>    <idle_time_to |       0 |       <result_name>cmsvm2_GA-p[e20-30MB_Lin64f]_1_Oryza-sativa-Japonica-Group_AP008214.lin.EMBL_RF00028_Intron_gpI_1349111823_52164_33</result_name>      <time>1436636371</time>        |
[...]

PR #2630, opened 2018-08-02
server: add the ability to have > 1 stages of trickle handling

davidpanderson:
Some projects (namely CPDN) need to process trickle-up messages
with two different daemons.
We do this as follows:
- daemon 1 enumerates messages with handled==0, and sets handled=1
- daemon 2 enumerates messages with handled==1, and sets handled=2
You can then purge records with handled==2

To implement this, trickle_handler.cpp has two global vars,
int handled_enum and int handled_set.
By default these are 0 and 1.
For daemon 2, set them to 1 and 2 in trickle_handler_init()

### DIFF
--- a/html/ops/purge_trickles.php
+++ b/html/ops/purge_trickles.php
@@ -40,7 +40,7 @@ if ($argc == 1) {
     $n = (int)$argv[2];
     $db->do_query("delete from msg_to_host where handled = $n");
 } else {
-    echo "usage\n";
+    echo "echo "usage: purge_trickles.php [msg_from_host | msg_to_host]\n";
 }
 
 ?>
--- a/html/ops/purge_trickles.php
+++ b/html/ops/purge_trickles.php
@@ -19,11 +19,28 @@
 // along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 
 // purge already-handled trickle messages from the DB
+//
+// no args: delete both up and down messages with handled != 0
+// msg_from_host N
+//    delete trickle ups with handled==N
+// msg_to_host N
+//    same, trickle down
 
 require_once("../inc/boinc_db.inc");
 $db = BoincDb::get();
 if (!$db) die("no DB connection");
-$db->do_query("delete from msg_from_host where handled <> 0");
-$db->do_query("delete from msg_to_host where handled <> 0");
+
+if ($argc == 1) {
+    $db->do_query("delete from msg_from_host where handled <> 0");
+    $db->do_query("delete from msg_to_host where handled <> 0");
+} else if ($argv[1] == "msg_from_host") {
+    $n = (int)$argv[2];
+    $db->do_query("delete from msg_from_host where handled = $n");
+} else if ($argv[1] == "msg_to_host") {
+    $n = (int)$argv[2];
+    $db->do_query("delete from msg_to_host where handled = $n");
+} else {
+    echo "usage\n";
+}
 
 ?>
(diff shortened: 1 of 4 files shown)


ChristianBeer:
While at it, why not fix the problem with rogue xml in the trickle message? See: #1485

ChristianBeer:
Code looks good. I'll do a final test soon and then merge.

ChristianBeer:
We should add a comment in code and in documentation that the [purge_trickles.php](https://github.com/BOINC/boinc/blob/113f9c534c97bfa2ae896aa6cae06b40534db02f/html/ops/purge_trickles.php) script needs to be changed when using multiple stages.
I personally also find it more useful to use `-1` as an error indicator by default. That makes it easier to look for problematic content.

davidpanderson:
Done

html/ops/purge_trickles.php
```diff
@@ -19,11 +19,28 @@
...
+    $db->do_query("delete from msg_from_host where handled <> 0");
+    $db->do_query("delete from msg_to_host where handled <> 0");
+} else if ($argv[1] == "msg_from_host") {
+    $n = (int)$argv[2];
+    $db->do_query("delete from msg_from_host where handled = $n");
+} else if ($argv[1] == "msg_to_host") {
+    $n = (int)$argv[2];
+    $db->do_query("delete from msg_to_host where handled = $n");
+} else {
+    echo "usage\n";
```
ChristianBeer: Is the actual usage message missing? The single word `usage` seems not very descriptive. We have to keep in mind that not all projects read the source code of the ops scripts.

ChristianBeer:
A short update on review progress. I want to add some automated tests for this part of the code to the server integration testing which is almost ready. Once the minor documentation issue in purge_trickles.php is fixed I'm going to merge.

TheAspens:
@ChristianBeer - is the only delay on merging this your note about?
> We should add a comment in code and in documentation that the purge_trickles.php script needs to be changed when using multiple stages.

davidpanderson:
Actually you don't need to change the script; see
https://boinc.berkeley.edu/trac/wiki/TrickleApi#Purgingtricklemessages

ChristianBeer:
The documentation is fine. What I meant is the error message in `purge_trickles.php` that is shown when the wrong number of parameters is used. It currently just says "usage" and the user needs to take a look at the code to get the proper information. Since the usage is rather simple the error message should contain what is basically the comment at the top of the file. 
```
// no args: delete both up and down messages with handled != 0
// msg_from_host N
//    delete trickle ups with handled==N
// msg_to_host N
//    same, trickle down
```
This is what you expect from other programs too.

TheAspens:
@davidpanderson - I think the only thing holding up @ChristianBeer from merging this pull request is his comment above.  If you could review his comment and respond I think Christian will be able to merge. 

Thanks - I just wanted to nudge this one along so that we can merge it.

davidpanderson:
This is a one-off for CPDN; fine as is.

TheAspens:
@davidpanderson - This looks like a useful ops function that we might use as well (we use trickle messages on one of our applications).  I suspect that it will be used on more projects than just CPDN.  Could you change the usage section from what you have now:

```
} else {
    echo "usage\n";
}
```
to
```
} else {
    echo "usage: purge_trickles.php [msg_from_host | msg_to_host]\n";
}
```
and then I think this will be ready for merge.  Thanks!

AenBleidd:
@davidpanderson, could you please make the fixes @TheAspens asked for? Thanks

PR #2631, opened 2018-08-03
web: if a thread or post creation fails...

davidpanderson:
(e.g. because it has too many links) show the user an explanation.

Aside: the forum code (which was written by students a long time ago)
uses a hodge-podge of return conventions.
At some point we should standardize on 0 = success, nonzero = error code


TheAspens:
Looks good.  Thanks for the contribution!

PR #2632, opened 2018-08-03
add script to show host breakdown by OS and client version

davidpanderson:


html/user/host_stats.php
```diff
@@ -0,0 +1,217 @@
...
+
+require_once("../inc/util.inc");
+require_once("../inc/cache.inc");
+
+$min_credit = .1;   // only count hosts with this much RAC
+$total_eac = 0;
+
+function show_type($s, $x) {
+    global $total_eac;
+    $p = number_format(100*$x->c/$total_eac, 4);
```
TheAspens: On a new project (admittedly a edge case), this causes a division by zero error and the following is output in the PHP.  

> Warning: Division by zero in /root/project/html/user/host_stats.php on line 30
> 
> Warning: Division by zero in /root/project/html/user/host_stats.php on line 30
> 
> Warning: Division by zero in /root/project/html/user/host_stats.php on line 30

Something like the following will prevent this:
```
function show_type($s, $x) {
    global $total_eac;
    if ( $total_eac == 0 ) {
        $p = 0;
    } else {
        $p = number_format(100*$x->c/$total_eac, 4);
    }
    row_array(array($s, $x->n, "$p %"));
}
```

TheAspens:
In addition to the comment above, there are a lot of one character variable names which did not provide any meaning to me when I was reading the code.  This violates the "Names should be descriptive without being verbose" rule from the [BOINC style guide](https://boinc.berkeley.edu/trac/wiki/CodingStyle#naming)

TheAspens:
Other than those comments, the code looks good.   Thanks for the addition.  We enjoyed looking at it on Seti@Home during the projects call today.

davidpanderson:
Good points!  I incorporated your suggestions

TheAspens:
Thanks for the changes.  Looks good!

PR #2637, opened 2018-08-06
Client, Win: Fix WSL-related crash

davidpanderson:
The client was crashing in cs_benchmark.cpp on the line
    bdp->host_info = host_info;
... in something involving copying WSL info.

I don't know why this was happening, but I fixed it by moving WSLS out of HOST_INFO.
It's probably better this way in any case.

### DIFF
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -252,26 +252,6 @@ void CLIENT_STATE::show_host_info() {
         tz<0?"":"+", tz
     );
 
-#ifdef _WIN64
-    if (host_info.wsl_available) {
-        msg_printf(NULL, MSG_INFO, "WSL detected:");
-        for (size_t i = 0; i < host_info.wsls.wsls.size(); ++i) {
-            const WSL& wsl = host_info.wsls.wsls[i];
-            if (wsl.is_default) {
-                msg_printf(NULL, MSG_INFO,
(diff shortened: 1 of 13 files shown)


AenBleidd:
Linux client is failed to built with next error:
```
  CXX      boinc_client-acct_mgr.o
In file included from acct_mgr.cpp:41:0:
client_state.h:112:5: error: ‘WSLS’ does not name a type
     WSLS wsls;
     ^
```

AenBleidd:
Also on Windows x86 client can't be built too because WSLS type is not defined, in `wslinfo.h` this struct is surrounded with `#ifdef _WIN64`

AenBleidd:
Now it builds ok but due to unnecessary merge this PR contains a lot of unrelated commits.

JuhaSointusalo:
@davidpanderson

Could you share the complete crash log? Oh, and please also upload debugging symbols to symbol store.

Since you don't know why the client crashed and you fixed it by shuffling some code around, you probably didn't really fix the crash, rather you moved it somewhere else.

> moving WSLS out of HOST_INFO.
> It's probably better this way in any case.

Why is it better to have WSL info in `CLIENT_STATE` instead of `HOST_INFO`? `HOST_INFO` stores information about host's capabilities so to me the logical place for info about WSL is indeed  in `HOST_INFO`.

PR #2640, opened 2018-08-08
Update apache configuration in make_project

lfield:
Making Apache v2.4 the default as v2.2 is not longer supported.


davidpanderson:
OK, but leave in the comment (reversed) for people running 2.2

lfield:
I have put back the comment.

Issue #2579, opened 2018-06-27
[GUI] 'Add project' wizard: distinguish 'web' and 'attach' urls


RichardHaselgrove:
The latest online update to all_projects_list.xml contains distinct urls:
```
        <url>http://www.cpdn.org/cpdnboinc/</url>
        <web_url>http://climateprediction.net/</web_url>
```
(and likewise for Einstein@Home). This is a good move - it resolves problems I encountered with 7.10 - but it needs placing in the repository for @CharlieFenton and others to use. And the attach wizard needs updating to display it: this is from a fresh checkout/build from master today, with the new file in place.

![attach-wizard](https://user-images.githubusercontent.com/14886436/41988145-25138492-7a33-11e8-99d6-cfe09e60691e.png)

Attaching a reconciliation of the various versions of all_projects_list as at 27 June 2018 (Excel speadsheet)

[Comparison of All_Projects_List versions.zip](https://github.com/BOINC/boinc/files/2142201/Comparison.of.All_Projects_List.versions.zip)

drshawnkwang:
Just my unenlightened opinion: I don't think #2580 is related, but I could be wrong.

davidpanderson:
CPDN's PHP site could redirect to the Wordpress site for non-BOINC user
agents.
Then they'd just need one URL

RichardHaselgrove:
But the non-attach BOINC functions (stats, help desk) still need to be on the designated BOINC server.

RichardHaselgrove:
PIcking up a comment from #2438, and extending the (overdue?) review of the 'Attach project' wizard.

1) intel_gpu not tested for
2) other GPU tests search for "[cuda", "[ati" (i.e. keyword at start of plan class string). SETI is not detected as ATI/AMD capable, because its plan classes start "[opencl". Test needs generalising.
3) it would be user-friendly to add tooltips to the icons: it requires one line of code per icon, such as

`m_pProjectDetailsSupportedPlatformWindowsCtrl->SetToolTip(_("Windows"));`

adamradocz:
@RichardHaselgrove I'd suggest to increase the icons size at least to 36 or 48 pixel wide. Is it possible. Currently this size is barely see-able and not so informative.

RichardHaselgrove:
We have another conference call tomorrow: perhaps we can ask again for some documentation. A search for "web_url" in this repository finds only my opening remark for this issue, so the current all_projects_list hasn't yet been committed.

[...]

PR #2643, opened 2018-08-10
[GUI] Refresh ProjectInfoPage in attach wizard

RichardHaselgrove:
1) Use web_url when present
2) Parse all plan class names from
   http://boinc.berkeley.edu/trac/wiki/AppPlan#Planclassnames
   (except WSL)
3) Update AMD, NV GPU icons to modern style
4) Add icons for Linux/ARM and Intel_GPU
   (icons downscaled from hi-res resource used for projects.php)
5) Re-sequence icons to group OS first, then resources
6) Add tooltips to all icons

Items 2-6 now match projects.php display. Implements #2579


AenBleidd:
It is not related directly to this ticket but anyway.
Does android icon looks 'noisy' for me only (white dots on the 'chest')?
image

RichardHaselgrove:
I think they're eyes (if above the shoulder), or a highlight. See https://boinc.berkeley.edu/images/android.png

AenBleidd:
It is a highlight that looks not so fine on small icon with reduced number of colors.

Ageless93:
More and free: http://pngimg.com/imgs/logos/android_logo/

RichardHaselgrove:
I'm finding that gimp (required tool for xpm export) is completely over-the-top for the sort of image work we're doing here. The best I've come up with is

new attach icons2

I'd prefer to leave it 'as is', otherwise this PR will languish for weeks like my previous attempts. And they want 7.14 out of the door...

clientgui/ProjectInfoPage.cpp
```diff
@@ -42,10 +42,14 @@
 #include "res/windowsicon.xpm"
 #include "res/macosicon.xpm"
 #include "res/linuxicon.xpm"
+#include "res/androidicon.xpm"
 #include "res/freebsdicon.xpm"
+#include "res/linuxarmicon2.xpm"
```
JuhaSointusalo: Why `2` in this and others?

clientgui/ProjectInfoPage.cpp
```diff
@@ -42,10 +42,14 @@
...
 #include "res/atiicon.xpm"
+#include "res/amdicon2.xpm"
```
JuhaSointusalo: Merge ATi and AMD here and in the rest of the code.

clientgui/ProjectInfoPage.cpp
```diff
@@ -42,10 +42,14 @@
...
 #include "res/nvidiaicon.xpm"
```
JuhaSointusalo: If you are not going to use the old icon and want to have a new file with new name then remove the old one.

clientgui/ProjectInfoPage.cpp
```diff
@@ -512,8 +559,9 @@ void CProjectInfoPage::OnProjectSelected( wxCommandEvent& WXUNUSED(event) ) {
 
         CProjectInfo* pProjectInfo = (CProjectInfo*)m_pProjectsCtrl->GetClientData(m_pProjectsCtrl->GetSelection());
 
-        wxString strURL = pProjectInfo->m_strURL;
-        EllipseStringIfNeeded(strURL, m_pProjectDetailsURLCtrl);
+//        wxString strURL = pProjectInfo->m_strURL;
```
JuhaSointusalo: Remove.

clientgui/ProjectInfoPage.cpp
```diff
@@ -643,25 +702,42 @@ void CProjectInfoPage::OnPageChanged( wxWizardExEvent& event ) {
...
 
-                    if (strProjectPlatform.Find(_T("[cuda")) != wxNOT_FOUND) {
+                    if (strProjectPlanClass.Find(_T("cuda")) != wxNOT_FOUND) {
                         pProjectInfo->m_bProjectSupportsCUDA = true;
 						if (!pDoc->state.host_info.coprocs.have_nvidia()) continue;
                     }
 
-                    if (strProjectPlatform.Find(_T("[ati")) != wxNOT_FOUND) {
+                    if (strProjectPlanClass.Find(_T("nvidia")) != wxNOT_FOUND) {
+                        pProjectInfo->m_bProjectSupportsIntelGPU = true;
```
JuhaSointusalo: Bug, though checking for `nvidia` makes sense to catch NVIDIA+OpenCL.

clientgui/res/amdicon2.xpm
```diff
@@ -0,0 +1,308 @@
+/* XPM */
+static char *amdicon2_xpm[] = {
```
JuhaSointusalo: These must be `static const char*` to avoid compiler warnings, even if it's against XPM spec.

clientgui/res/amdicon2.xpm
```diff
@@ -0,0 +1,308 @@
...
+"o.p.u q.r.s.t.u.v.w.x.y.z.A.B.C.p D.E.",
+"F.G.u 0.H.I.f g J.K.j L.M.m N.C p O.P.",
+"Q.R.S.T.d U.V.W.X.Y.Z.`. +.+N.++p @+#+",
+"$+%+&+*+=+-+;+>+,+'+,+)+!+~+{+]+^+/+(+",
+"_+:+<+[+}+|+1+2+3+4+5+6+9 7+8+}+9+0+a+",
+"b+c+d+d+d+e+f+g+h+i+j+$+k+l+d+d+m+n+7.",
+"o+p+q+r+s+t+u+v+w+x+y+z+A+B+C+D+E+F+G+",
+"H+I+J+K+L+M+N+O+P+Q+R+S+T+a+U+V+W+X+H+"};
+// This file is part of BOINC.
+// http://boinc.berkeley.edu
```
JuhaSointusalo: `https`

clientgui/res/amdicon2.xpm
```diff
@@ -0,0 +1,308 @@
...
+// as published by the Free Software Foundation,
+// either version 3 of the License, or (at your option) any later version.
+//
+// BOINC is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
```
JuhaSointusalo: `https`

clientgui/ProjectInfoPage.cpp
RichardHaselgrove: To clearly differentiate my second generation icons from the previous set.

clientgui/ProjectInfoPage.cpp
RichardHaselgrove: Code has to handle plan class names using either convention.

clientgui/ProjectInfoPage.cpp
RichardHaselgrove: I wanted my PR to be non-destructive, at least until previous developers (Charlie and Rom) have had a chance to review.

clientgui/ProjectInfoPage.cpp
RichardHaselgrove: Yes, it's probably safe to do that.

clientgui/ProjectInfoPage.cpp
RichardHaselgrove: Good catch. I'll check for that and other copy'n'paste errors in the morning.

clientgui/res/amdicon2.xpm
RichardHaselgrove: The compiler objected to what GIMP had written. I tweaked it to match what the first generation icons already had.

clientgui/res/amdicon2.xpm
RichardHaselgrove: Then we'll have to change the old ones I copied from too.

clientgui/res/amdicon2.xpm
RichardHaselgrove: ditto.

CharlieFenton:
While this is not entirely part of this PR, why is the code trying to translate URLs from English? I realize Richard was just expamnding on the first line below, which was added by @romw on June 15, 2011:
`            pProjectInfo->m_strURL = wxGetTranslation(wxString(m_apl->projects[i]->url.c_str(), wxConvUTF8));
            pProjectInfo->m_strWebURL = wxGetTranslation(wxString(m_apl->projects[i]->web_url.c_str(), wxConvUTF8));
`

I haven't found any additional problems within the scope of this PR, but I see a few concerns with the logic in ProjectInfoPage.cpp.

[1]  I think there is a shortcoming in the way the icons are displayed and in the way it warns that a select project may not have work for this type of computer. A project may support certain GPUs only on certain operating systems. Say a project supports only AMD on Windows and only NVIDIA on Macs. Then it should not display the AMD icon on a Mac, and should display the warning if selected on a Windows PC without an AMD GPU. The code now does not display the warning if the project has any support for the OS on which BOINC is currently running, even if there is no support for the GPU.

[2] I wonder how much sense it makes to show Linux and Mac icons on a Windows PC, etc. While this might be useful to the tiny  percentage of BOINC users who have multiple computers with different operating systems, it might be confusing to most users.

[3] It is possible for a plan class to just have `opencl` without specifying the GPU brand; ideally OpenCL code should work on AMD, NVIDIA and Intel GPUs. In that case, should all 3 icons be displayed? Or should there be a separate icon for OpenCL support?

[4] While this PR adds checks for `arm-unknown-linux-gnu` and `arm-unknown-linux-gnueabihf`, the all_projects_list also has such entries as `aarch64-unknown-linux-gnu` and `x86-android-linux-gnu`. And all of these would trigger the display of the Linux icon and suppress the warning, even though applications for X86 won't run on ARM and vice-versa.  But I realize there is a practical limit on how thorough this logic can be, especially as new combinations of hardware and OS continue to appear.

RichardHaselgrove:
@CharlieFenton: I'm just starting work on the items Juha requested, but a quick initial response to your additional comments.

[1] I agree, but this shortcoming applies to the projects.php page on the BOINC website as well. "only AMD on Windows and only NVIDIA on Macs" should be visible in both places, but I think that's beyond the scope of this PR - one for the TODO list.

[2] Just a thought - how would this work if BOINC Manager is being used to control a remote PC, say a Windows tablet being used to control a Raspberry Pi? Again, food for thought, but outside the scope of this refresh.

[3] OpenCL runs on CPUs too, though I doubt any project would use it in preference to native apps. Developers have told me in the past that in most cases the applications do have to be separately compiled to suit the target GPU type, Again, this change would have to be applied to projects.php too.

[4] Again, I think that's out of scope here. Another for TODO, and an issue for BOINC v8.0?

RichardHaselgrove:
Using https in the licence links opens a much bigger can of worms. While working on this PR, I've been referring to a copy of all_projects_list.xml (14-day autoupdate version) downloaded on 05 August 2018. It contains 34 project <web_url> entries - only 12 of which are https, and 22 still plain http. Although it's well documented that there are problems changing project master urls, that doesn't apply to either get_project_config or web display requests. I think we're guilty of a collective groupthink here, and we need to train ourselves out of it.

Of the 34 projects, just 8 can't be contacted by https from here. They are:

www.acousticsathome.ru/boinc/
sat.isa.ru/pdsat/
www.enigmaathome.net/
gerasim.boinc.ru/
srbase.my-firewall.org/sr5/
casathome.ihep.ac.cn/ (I can't reach that one by http either)
quakecatcher.net/sensor
radioactiveathome.org/boinc/

Downloading a fresh copy of all_projects_list.xml today, I find that all <web_url> elements have been removed again, so maybe David is working on that - though it's hard to know, because nothing has been committed yet.

Ageless93:
@CharlieFenton, 
[2] How about running BOINC in a VM on another host OS? Which of the two OSes does it recognize? 
But in that same frame of mind for the user who only has one OS, wouldn't it be overkill then to show the icon for the OS you're on? Because even though there _could_ be other OSes supported by this project, you wouldn't care as you're not on any of those.

[3} Not all projects have OpenCL applications for all GPUs. Aside from that, one has to specifically install the drivers with OpenCL support for the device, else it won't be recognized as OpenCL capable, despite it being an industry standard. E.g. AMD GPU drivers do contain OpenCL, but this OpenCL doesn't allow BOINC to detect that the Intel CPU and GPU are OpenCL capable.

clientgui/res/amdicon2.xpm
JuhaSointusalo: Could you show the error message from compiler? The XPM data is made of string literals which are constant. Not having the data declared `const` generates [tons of warnings](https://travis-ci.org/BOINC/boinc/jobs/414841363#L2786). I think previously there was only one file not declaring the data `const`.

clientgui/ProjectInfoPage.cpp
JuhaSointusalo: Ok, thanks for explaining.

clientgui/ProjectInfoPage.cpp
JuhaSointusalo: I had misunderstood your intentions. I thought you were going to have ATi and AMD icons both which would not have made much sense.

clientgui/res/amdicon2.xpm
JuhaSointusalo: I was aiming for having new files use `https` and old files could be mass changed in some other PR. I could have communicated that better...

clientgui/res/amdicon2.xpm
RichardHaselgrove: GIMP exports the line as
`static char * D:\Sources_build\boinc_7_12_1\clientgui\res\androidicon2_xpm[] = {`

That's clearly not usable, and VS2013 responds with

Error	3	error C2143: syntax error : missing ';' before ':'	d:\sources_build\boinc_7-12-1\clientgui\res\androidicon2.xpm	2	1	boincmgr
Error	4	error C2059: syntax error : ':'	d:\sources_build\boinc_7-12-1\clientgui\res\androidicon2.xpm	2	1	boincmgr
Error	5	error C2017: illegal escape sequence	d:\sources_build\boinc_7-12-1\clientgui\res\androidicon2.xpm	2	1	boincmgr
Error	6	error C2146: syntax error : missing ';' before identifier 'boinc_7_12_1'	d:\sources_build\boinc_7-12-1\clientgui\res\androidicon2.xpm	2	1	boincmgr
Error	7	error C4430: missing type specifier - int assumed. Note: C++ does not support default-int	d:\sources_build\boinc_7-12-1\clientgui\res\androidicon2.xpm	2	1	boincmgr
Error	8	error C2146: syntax error : missing ';' before identifier 'clientgui'	d:\sources_build\boinc_7-12-1\clientgui\res\androidicon2.xpm	2	1	boincmgr
Error	9	error C2146: syntax error : missing ';' before identifier 'res'	d:\sources_build\boinc_7-12-1\clientgui\res\androidicon2.xpm	2	1	boincmgr
Error	10	error C2146: syntax error : missing ';' before identifier 'androidicon2_xpm'	d:\sources_build\boinc_7-12-1\clientgui\res\androidicon2.xpm	2	1	boincmgr

I'll try again with const - I missed that specific point last time.

Yes, `static const char *androidicon2_xpm[] = {` is clean. I'll change it - I must have looked at the one file without `const` when fixing the first problem.

RichardHaselgrove: Yes, it was the original Android icon (which was added later, and doesn't have the licence lines either). I'll fix that while I'm here. Actually, no - I'll pull in that flat colour Android2 icon, to deal with Vitali's objection at the same time.

RichardHaselgrove:
> The code doesn't check for intel_gpu plan classes.

I must have distracted myself when writing that bug you found first. I've put it back in, but on my development machine - which has both NVidia and Intel GPUs active - it doesn't show.

This is where my ignorance of C++ shows. Can somebody please talk me through the logic and purpose of the 'continue' statements in the detection area - around line 700? I'm worried that they may skip later tests if triggered (by no ATI in my case). The intel_gpu detection works if I comment them out.

I'll commit the code as it stands, but please treat this as a WIP until I can clarify what's going on.

RichardHaselgrove:
After much careful scrutiny and abortive debugging (can't write to the message log from the Manager), I can't find what spooked me last night - unless it was LHC's [n**ati**ve_mt] plan class being picked up as ATI. Given that, I'll take off the WIP and invite reviewers to consider this as if ready to merge.

CharlieFenton:
> can't write to the message log from the Manager

You can add temporary printf() statements in the manager, and they will be written to the Manager's _stdoutgui.txt_ file. I think that on Windows you will usually find that file at this location:
`C:/Users/{username}/AppData/Roaming/BOINC/stdoutgui.txt.`
That is where it is on my Windows 7 system.

CharlieFenton:
@ChristianBeer: Can you please check why Travis CI builds are failing for all current PRs?

AenBleidd:
@CharlieFenton, I'm not @ChristianBeer, but it looks like that there are some issues with Ubuntu image on TravisCI. I've restarted failed jobs

RichardHaselgrove:
Well. 'abortive' in the sense of 'no bugs found', but I managed to get a GUI debugger working. Your way would have been quicker, 

mgr debug

PR #2646, opened 2018-08-10
Client, Win: Fix WSL-related crash (2nd try)

davidpanderson:
The client was crashing in cs_benchmark.cpp on the line
        bdp->host_info = host_info;
    ... in something involving copying WSL info.

    I don't know why this was happening, but I fixed it by moving WSLS out of HO
ST_INFO.
    It's probably better this way in any case.

### DIFF
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -252,26 +252,6 @@ void CLIENT_STATE::show_host_info() {
         tz<0?"":"+", tz
     );
 
-#ifdef _WIN64
-    if (host_info.wsl_available) {
-        msg_printf(NULL, MSG_INFO, "WSL detected:");
-        for (size_t i = 0; i < host_info.wsls.wsls.size(); ++i) {
-            const WSL& wsl = host_info.wsls.wsls[i];
-            if (wsl.is_default) {
-                msg_printf(NULL, MSG_INFO,
-                    "   [%s] (default): %s (%s)", wsl.distro_name.c_str(), wsl.name.c_str(), wsl.version.c_str()
-                );
-            } else {
-                msg_printf(NULL, MSG_INFO,
-                    "   [%s]: %s (%s)", wsl.distro_name.c_str(), wsl.name.c_str(), wsl.version.c_str()
-                );
-            }
-        }
-    } else {
-        msg_printf(NULL, MSG_INFO, "No WSL found.");
-    }
-#endif
-
     if (strlen(host_info.virtualbox_version)) {
         msg_printf(NULL, MSG_INFO,
             "VirtualBox version: %s",
@@ -287,6 +267,28 @@ void CLIENT_STATE::show_host_info() {
 #endif
     }
 }
+
+#ifdef _WIN64
+void WSLS::show() {
+    if (available) {
+        msg_printf(NULL, MSG_INFO, "WSL detected:");
+        for (size_t i = 0; i < wsls.size(); ++i) {
+            const WSL& wsl = wsls[i];
+            if (wsl.is_default) {
+                msg_printf(NULL, MSG_INFO,
+                    "   [%s] (default): %s (%s)", wsl.distro_name.c_str(), wsl.name.c_str(), wsl.version.c_str()
+                );
+            } else {
+                msg_printf(NULL, MSG_INFO,
+                    "   [%s]: %s (%s)", wsl.distro_name.c_str(), wsl.name.c_str(), wsl.version.c_str()
+                );
+            }
+        }
+    } else {
+        msg_printf(NULL, MSG_INFO, "No WSL found.");
+    }
+}
+#endif
 
 int rsc_index(const char* name) {
     const char* nm = strcmp(name, "CUDA")?name:GPU_TYPE_NVIDIA;
@@ -625,6 +627,12 @@ int CLIENT_STATE::init() {
     host_info.get_host_info(true);
     set_ncpus();
     show_host_info();
+#ifdef _WIN64
+    if (!cc_config.dont_use_wsl) {
+        wsls.get_information();
+        wsls.show();
+    }
+#endif
 
     // this follows parse_state_file() because that's where we read project names
     //
--- a/client/client_state.h
+++ b/client/client_state.h
@@ -109,7 +109,9 @@ struct CLIENT_STATE {
     NET_STATS net_stats;
     ACTIVE_TASK_SET active_tasks;
     HOST_INFO host_info;
+#ifdef _WIN64
     WSLS wsls;
+#endif
 
     // the following used only on Android
     DEVICE_STATUS device_status;
--- a/client/client_state.h
+++ b/client/client_state.h
@@ -55,6 +55,7 @@ using std::vector;
 #include "project_list.h"
 #include "scheduler_op.h"
 #include "time_stats.h"
+#include "wslinfo.h"
 
 #ifdef SIM
 #include "../sched/edf_sim.h"
@@ -108,6 +109,7 @@ struct CLIENT_STATE {
     NET_STATS net_stats;
     ACTIVE_TASK_SET active_tasks;
     HOST_INFO host_info;
+    WSLS wsls;
 
     // the following used only on Android
     DEVICE_STATUS device_status;
--- a/client/cs_statefile.cpp
+++ b/client/cs_statefile.cpp
@@ -421,6 +421,11 @@ int CLIENT_STATE::parse_state_file_aux(const char* fname) {
             }
             continue;
         }
+#ifdef _WIN64
+        if (xp.match_tag("wsls")) {
+            wsls.parse(xp);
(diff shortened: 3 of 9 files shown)


davidpanderson:
It's better because it disentangles WSLS from HOST_INFO

davidpanderson:
But I don't have strong feelings about this.
If you can fix the crash some other way, feel free.

AenBleidd:
I was not successful trying to reproduce the original issue. @davidpanderson, what are the preconditions/steps to reproduce?

davidpanderson:
When the client runs CPU benchmarks, it crash in the line
    bdp->host_info = host_info;
in cs_benchmark.cpp.
Stack trace:

>	msvcr100d.dll!_CrtDbgBreak()  Line 85	C
 	msvcr100d.dll!_VCrtDbgReportW(int nRptType, const wchar_t * szFile, int nLine, const wchar_t * szModule, const wchar_t * szFormat, char * arglist)  Line 502	C
 	msvcr100d.dll!_CrtDbgReportWV(int nRptType, const wchar_t * szFile, int nLine, const wchar_t * szModule, const wchar_t * szFormat, char * arglist)  Line 242	C++
 	msvcr100d.dll!_CrtDbgReportW(int nRptType, const wchar_t * szFile, int nLine, const wchar_t * szModule, const wchar_t * szFormat, ...)  Line 258 + 0x2c bytes	C++
 	msvcp100d.dll!std::_Debug_message(const wchar_t * message, const wchar_t * file, unsigned int line)  Line 13 + 0x22 bytes	C++
 	boinc.exe!std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > >::_Compat(const std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > & _Right)  Line 239	C++
 	boinc.exe!std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > >::operator==(const std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > & _Right)  Line 203	C++
 	boinc.exe!std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > >::operator!=(const std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > & _Right)  Line 208 + 0xf bytes	C++
 	boinc.exe!std::vector<WSL,std::allocator<WSL> >::erase(std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > * _First_arg, std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > * _Last_arg)  Line 1194 + 0xf bytes	C++
 	boinc.exe!std::vector<WSL,std::allocator<WSL> >::clear()  Line 1218 + 0xe1 bytes	C++
 	boinc.exe!std::vector<WSL,std::allocator<WSL> >::operator=(const std::vector<WSL,std::allocator<WSL> > & _Right)  Line 715 + 0xa bytes	C++
 	boinc.exe!WSLS::operator=(const WSLS & __that)  + 0x32 bytes	C++
 	boinc.exe!HOST_INFO::operator=(const HOST_INFO & __that)  + 0x458 bytes	C++
 	boinc.exe!cpu_benchmarks(BENCHMARK_DESC * bdp)  Line 219	C++
 	boinc.exe!win_cpu_benchmarks(void * p)  Line 234	C++
 	kernel32.dll!00000000779d59cd() 	
 	[Frames below may be incorrect and/or missing, no symbols loaded for kernel32.dll]	
 	ntdll.dll!0000000077b0a561()

AenBleidd:
It's interesting. It works OK with VS2013 and, as I see from the stacktrace, it crashes with VS2010.

AenBleidd:
@davidpanderson, could you please try the fix provided in #2650 to verify whether it solves your initial issue?

JuhaSointusalo:
So the code crashes here:
```cpp
bdp->host_info = host_info;
```
`bdp` is a pointer to `BENCHMARK_DESC` object. `BENCHMARK_DESC` looks like this, with sub-objects inlined and other not interesting members removed:
```
struct BENCHMARK_DESC
    class HOST_INFO
        struct WSLS
            class std::vector<WSL, std::allocator<WSL> >
                ...
        struct COPROCS
            struct COPROC
                struct PCI_INFO
                struct OPENCL_DEVICE_PROP
            struct COPROC_NVIDIA : public COPROC
                struct CUDA_DEVICE_PROP
            struct COPROC_ATI : public COPROC
                struct CALdeviceattribsRec
                struct CALdeviceinfoRec
            struct COPROC_INTEL : public COPROC
        struct OPENCL_CPU_PROP
            struct OPENCL_DEVICE_PROP
```

`bdp` is a part of an array which is allocated with this:
```cpp
benchmark_descs = (BENCHMARK_DESC*)calloc(bm_ncpus, sizeof(BENCHMARK_DESC));
```
`calloc` is basically `malloc`+`memset`. `malloc` doesn't run constructors or do any other initialisation, `memset` trashes `vptr` and assumes all member variables can have the same value.

Until WSL detection was added all those classes were close enough to POD types that allocating the array with `calloc` didn't break anything too much. But, because of the `std::vector`, `WSLS` is not at all POD type and can't be allocated that way. `std::vector` inside it is basically initialised to garbage.

I haven't run the code under debugger yet to confirm this but allocating C++ objects that way looks very suspicious.

AenBleidd:
@JuhaSointusalo, thank you for such detailed investigation. I guess I have a solution that will fix the original issue. It will be ready today later.

AenBleidd:
#2650 updated

JuhaSointusalo:
After having so much fun with debugger I can tell that `calloc` indeed misconstructs `std::vector`.

`std::vector` is actually a number of classes. One of the classes holds a pointer to the `std::vector` in debug builds. I'm not sure but I think the idea is that the pointer can be used to check for any corruption or that when using iterators they came from the same object and things like that.

With `calloc`ed vector the pointer is `NULL` which looks like corrupt vector or some other bug and CRT throws an assert.

With VS2013 the assert is still there but the implementation of `std::vector::clear` has been changed in such way that the asserting code path is no longer taken.

AenBleidd:
#2651 created. Please review

TheAspens:
@davidpanderson - @AenBleidd created two possible fixes for this #2651 and #2650 .  Would it be possible for you to review those two solutions and pick the one that is better to merge?  Thanks to everyone working on this issue!

@ChristianBeer - would you also have the time to take a look at all three as well?  @AenBleidd thought you might be good to review as well.

I can review if no-one else is able, but my C++ is weaker than other people's and I would rather defer to them if possible.

ChristianBeer:
I took a quick look over the three implementations and currently prefer the one in #2651 although I need to check that it solves the original issue. I'll look into this over the weekend.

JuhaSointusalo:
I tested both #2651 and #2650 with VS2010 and 2013, Release and Debug, x64 and Win32, and 1-4 CPUs. Also tested on Linux release build with 1 and 2 CPUs. No crashes with either PR.

I like #2651 because it's more C++ style. #2650 with manual memory management and raw pointers is more plain C style. I made an error in proposing to use `std::vector::resize()` instead of `reserve()` but it's not critically bad and can be fixed sometime later with other improvements to the benchmark code.

(The other improvements being: actually use the vector to keep count of running benchmarks instead of `bm_ncpus`, synchronise access to benchmark data, use threads on all platforms (on multithreaded programs only async-signal-safe functions are safe to use after `fork()`), remove `HOST_INFO` from `BENCHMARK_DESC` and `cpu_benchmarks()` (doesn't make sense to instantiate several copies of `HOST_INFO` just to use them to save a couple of variables), release memory after benchmark run, and anything else that might come up).

ChristianBeer:
I reviewed #2651 which I prefer for the same reasons as @JuhaSointusalo. But why would you switch `reserve()` for `resize()`? The latter automatically creates the object that we would need to create on our own otherwise and the size of the vector doesn't change during runtime since it is fixed to the number of CPUs.

JuhaSointusalo:
`resize()` creates objects but their states are not necessarely valid. The rest of the benchmark code assumes that all `BENCHMARK_DESC` objects have valid state. The class could have a constructor that sets some reasonable initial state. But what is that initial state, in particular thread and process ids?

On Windows we have `INVALID_HANDLE_VALUE` which as the name says is not valid handle and any operation using it fails. On Unixes there is no good invalid process id value. `abort_benchmark()` passes the PID to `kill()` and all PID values have some meaning to `kill()`, none of them are invalid. So you'd need to pick some magic value, set it in constructor and check for that magic value in `abort_benchmark()`. `reserve()` avoids that in that the vector would have objects in valid state only.

In practise `resize()` vs `reserve()` doesn't matter much because the code doesn't check for errors anyway.

ChristianBeer:
Closed in favor of #2651

PR #2650, opened 2018-08-14
[WIP] Possible fix for cs_benchmard.cpp issue:

AenBleidd:
Stack trace:

>	msvcr100d.dll!_CrtDbgBreak()  Line 85	C
 	msvcr100d.dll!_VCrtDbgReportW(int nRptType, const wchar_t * szFile, int nLine, const wchar_t * szModule, const wchar_t * szFormat, char * arglist)  Line 502	C
 	msvcr100d.dll!_CrtDbgReportWV(int nRptType, const wchar_t * szFile, int nLine, const wchar_t * szModule, const wchar_t * szFormat, char * arglist)  Line 242	C++
 	msvcr100d.dll!_CrtDbgReportW(int nRptType, const wchar_t * szFile, int nLine, const wchar_t * szModule, const wchar_t * szFormat, ...)  Line 258 + 0x2c bytes	C++
 	msvcp100d.dll!std::_Debug_message(const wchar_t * message, const wchar_t * file, unsigned int line)  Line 13 + 0x22 bytes	C++
 	boinc.exe!std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > >::_Compat(const std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > & _Right)  Line 239	C++
 	boinc.exe!std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > >::operator==(const std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > & _Right)  Line 203	C++
 	boinc.exe!std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > >::operator!=(const std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > & _Right)  Line 208 + 0xf bytes	C++
 	boinc.exe!std::vector<WSL,std::allocator<WSL> >::erase(std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > * _First_arg, std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > * _Last_arg)  Line 1194 + 0xf bytes	C++
 	boinc.exe!std::vector<WSL,std::allocator<WSL> >::clear()  Line 1218 + 0xe1 bytes	C++
[...]

PR #2651, opened 2018-08-15
[WIP] Fix cs_benchmark.cpp heap destruction

AenBleidd:
Use std::vector instead of allocated piece of memory to store multiple objects.
This hopefully fixes crash originally described in #2646 with heap destruction while copying data and next stacktrace:

>	msvcr100d.dll!_CrtDbgBreak()  Line 85	C
 	msvcr100d.dll!_VCrtDbgReportW(int nRptType, const wchar_t * szFile, int nLine, const wchar_t * szModule, const wchar_t * szFormat, char * arglist)  Line 502	C
 	msvcr100d.dll!_CrtDbgReportWV(int nRptType, const wchar_t * szFile, int nLine, const wchar_t * szModule, const wchar_t * szFormat, char * arglist)  Line 242	C++
 	msvcr100d.dll!_CrtDbgReportW(int nRptType, const wchar_t * szFile, int nLine, const wchar_t * szModule, const wchar_t * szFormat, ...)  Line 258 + 0x2c bytes	C++
 	msvcp100d.dll!std::_Debug_message(const wchar_t * message, const wchar_t * file, unsigned int line)  Line 13 + 0x22 bytes	C++
 	boinc.exe!std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > >::_Compat(const std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > & _Right)  Line 239	C++
 	boinc.exe!std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > >::operator==(const std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > & _Right)  Line 203	C++
 	boinc.exe!std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > >::operator!=(const std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > & _Right)  Line 208 + 0xf bytes	C++
 	boinc.exe!std::vector<WSL,std::allocator<WSL> >::erase(std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > * _First_arg, std::_Vector_const_iterator<std::_Vector_val<WSL,std::allocator<WSL> > > * _Last_arg)  Line 1194 + 0xf bytes	C++
 	boinc.exe!std::vector<WSL,std::allocator<WSL> >::clear()  Line 1218 + 0xe1 bytes	C++
 	boinc.exe!std::vector<WSL,std::allocator<WSL> >::operator=(const std::vector<WSL,std::allocator<WSL> > & _Right)  Line 715 + 0xa bytes	C++
 	boinc.exe!WSLS::operator=(const WSLS & __that)  + 0x32 bytes	C++
 	boinc.exe!HOST_INFO::operator=(const HOST_INFO & __that)  + 0x458 bytes	C++
 	boinc.exe!cpu_benchmarks(BENCHMARK_DESC * bdp)  Line 219	C++
 	boinc.exe!win_cpu_benchmarks(void * p)  Line 234	C++
 	kernel32.dll!00000000779d59cd()
 	[Frames below may be incorrect and/or missing, no symbols loaded for kernel32.dll]
 	ntdll.dll!0000000077b0a561()

### DIFF
--- a/client/cs_benchmark.cpp
+++ b/client/cs_benchmark.cpp
@@ -276,7 +276,7 @@ void CLIENT_STATE::start_cpu_benchmarks() {
         benchmark_descs[i].error = false;
 #ifdef _WIN32
         benchmark_descs[i].handle = CreateThread(
-            NULL, 0, win_cpu_benchmarks, &benchmark_descs, 0,
+            NULL, 0, win_cpu_benchmarks, &benchmark_descs[i], 0,
             &benchmark_descs[i].pid
         );
         int n = host_info.p_ncpus;
--- a/client/cs_benchmark.cpp
+++ b/client/cs_benchmark.cpp
@@ -262,15 +262,13 @@ void CLIENT_STATE::start_cpu_benchmarks() {
     remove_benchmark_file(BM_TYPE_INT);
     cpu_benchmarks_start = dtime();
 
-    if (!benchmark_descs.empty()) {
-        benchmark_descs.clear();
-    }
+    benchmark_descs.clear();
+    benchmark_descs.resize(ncpus);
+
     bm_ncpus = ncpus;
     benchmarks_running = true;
 
     for (i=0; i<bm_ncpus; i++) {
-        benchmark_descs.push_back(BENCHMARK_DESC());
-
         benchmark_descs[i].ordinal = i;
         benchmark_descs[i].done = false;
         benchmark_descs[i].error = false;
--- a/client/cs_benchmark.cpp
+++ b/client/cs_benchmark.cpp
@@ -292,7 +292,7 @@ void CLIENT_STATE::start_cpu_benchmarks() {
                 perror("setpriority");
             }
 #endif
-            int retval = cpu_benchmarks(&benchmark_descs);
+            int retval = cpu_benchmarks(&benchmark_descs[i]);
             fflush(NULL);
             _exit(retval);
         } else {
--- a/client/cs_benchmark.cpp
+++ b/client/cs_benchmark.cpp
@@ -68,6 +68,8 @@
 #include "log_flags.h"
 #include "client_state.h"
 
+#include <vector>
+
 // defaults in case benchmarks fail or time out.
 // better to err on the low side so hosts don't get too much work
 
@@ -119,7 +121,7 @@ struct BENCHMARK_DESC {
 #endif
 };
 
-static BENCHMARK_DESC* benchmark_descs=0;
+static std::vector<BENCHMARK_DESC> benchmark_descs;
 static double cpu_benchmarks_start;
 static int bm_ncpus;
     // user might change ncpus during benchmarks.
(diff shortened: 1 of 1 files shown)


client/cs_benchmark.cpp
```diff
@@ -260,20 +262,21 @@ void CLIENT_STATE::start_cpu_benchmarks() {
...
     for (i=0; i<bm_ncpus; i++) {
+        benchmark_descs.push_back(BENCHMARK_DESC());
+
         benchmark_descs[i].ordinal = i;
         benchmark_descs[i].done = false;
         benchmark_descs[i].error = false;
 #ifdef _WIN32
         benchmark_descs[i].handle = CreateThread(
-            NULL, 0, win_cpu_benchmarks, benchmark_descs+i, 0,
+            NULL, 0, win_cpu_benchmarks, &benchmark_descs, 0,
```
JuhaSointusalo: Shouldn't that be `&benchmark_descs[i]`?

AenBleidd: Good catch! Thank you again.

JuhaSointusalo:
I'm not sure if the original problem classifies as heap corruption. But constructing C++ objects with `calloc` is probably undefined behaviour, in which case anything and everything is possible.

Your code works with one CPU but, rather ironically, it corrupts the heap when there is more than one CPU. 

In MS CRT `std::vector`'s capacity starts with 0 and it is grown by 50% any time it's needed. So the capacity changes 0 -> 1 -> 2 -> 3 -> 4 and so on. Every time the capacity changes the memory block backing the vector is moved (or at least that happened every time I looked at it in debugger). Since the memory block is moved the pointer that was given to `CreateThread()` and to the new thread is no longer valid.

Allocating the `BENCHARK_DESC`s upfront like so seems to fix it:
```diff
diff --git a/client/cs_benchmark.cpp b/client/cs_benchmark.cpp
index 9518b3e2b9..45ff48ff9c 100644
--- a/client/cs_benchmark.cpp
+++ b/client/cs_benchmark.cpp
@@ -262,15 +262,13 @@ void CLIENT_STATE::start_cpu_benchmarks() {
     remove_benchmark_file(BM_TYPE_INT);
     cpu_benchmarks_start = dtime();

-    if (!benchmark_descs.empty()) {
-        benchmark_descs.clear();
-    }
+    benchmark_descs.clear();
+    benchmark_descs.resize(ncpus);
+
     bm_ncpus = ncpus;
     benchmarks_running = true;

     for (i=0; i<bm_ncpus; i++) {
-        benchmark_descs.push_back(BENCHMARK_DESC());
-
         benchmark_descs[i].ordinal = i;
         benchmark_descs[i].done = false;
         benchmark_descs[i].error = false;
```
(But do test it carefully.)

AenBleidd:
Ok. At this point I think that the original idea with vector of pointers
was significantly better idea. What do you think? Maybe we can return to
it? More changes but it's better than objects that can be moved.

Best regards,
Vitalii

Sent via Android

пт, 17 авг. 2018 г., 23:08 Juha Sointusalo <notifications@github.com>:

AenBleidd:
I've reopened #2650 and add necessary fixes. Could you please review it and test?
I definitely think that the idea with manual pointers is better even when this adds additional code changes

davidpanderson:
BTW, my suggestion of removing WSLS from HOST_INFO was motivated in part by work I've been doing where I needed to reuse parts of large existing software, and the parts I needed were entangled with parts I didn't want, and it took brain surgery to separate them.

If someone in the future wants to use the WSLS code (perhaps for a rewrite of the BOINC client) it would help them if WSLS were unentangled.  I think we should do it this way, not necessarily in this PR.

RichardHaselgrove:
I found I had a machine with VS2010 Express loaded, so I tried giving it a go. Nightmare getting x64 support into VC2010 (final key component seemed to be https://www.microsoft.com/en-us/download/details.aspx?id=4422) but I'm now getting 

`Required file "" is missing`

on all BOINC components in 64-bit build configuration (host is Win7 x64: sources are fresh clone of master and boinc_depends_win_vs2010)

Most Win32 builds succeeded: the x64 errors list target line numbers which I can investigate tomorrow. But that's enough for tonight.

JuhaSointusalo:
@AenBleidd prefers #2650 over #2651 but left the decision to the team. There's two likes for #2651 and one like for #2650 (not saying votes because a formal voting hasn't been requested). So this PR wins and I'll merge it.

JuhaSointusalo:
Now that I went to inbox to archive the email thread I see that there was a WIP in front of the subject. I believe the WIP was only because @AenBleidd wasn't sure if this fixes the crash in benchmarks. I did verify that the fix works so no harm done other than having an entry called WIP in repo history.

AenBleidd:
Yes, WIP label was there because there was two 2 PRs to fix one problem and the team had to decide which is better.
@JuhaSointusalo, thank you for merging this.

PR #2655, opened 2018-08-23
Add indexes to host table

TheAspens:
When a user has a very large number of hosts attached to a project (i.e. at least 50k), the scheduler requests can slow down as the existing index on userid still returns a significant number of records that have to be checked.  These two indexes address the issues associated with this situation.

See discussion in boinc-projects: https://groups.google.com/a/ssl.berkeley.edu/forum/#!topic/boinc_projects/Z59_4zIrSXE

### DIFF
--- a/html/ops/db_update.php
+++ b/html/ops/db_update.php
@@ -1162,6 +1162,12 @@ function update_5_9_2018() {
     ");
 }
 
+function update_8_23_2018() {
+    $retval = do_query("alter table host add index host_cpid (host_cpid)");
+    return $retval && do_query("alter table host add index host_domain_name (domain_name)");
+}
+
+
 // Updates are done automatically if you use "upgrade".
 //
 // If you need to do updates manually,
@@ -1219,7 +1225,8 @@ $db_updates = array (
     array(27023, "update_4_6_2018"),
     array(27024, "update_4_18_2018"),
     array(27025, "update_4_19_2018"),
-    array(27026, "update_5_9_2018")
+    array(27026, "update_5_9_2018"),
+    array(27027, "update_8_23_2018")
 );
 
 ?>
--- a/html/ops/db_update.php
+++ b/html/ops/db_update.php
@@ -1163,7 +1163,8 @@ function update_5_9_2018() {
 }
 
 function update_8_23_2018() {
-    $retval = do_query("alter table host add index host_cpid (host_cpid)");
+    $retval = do_query("alter table host add index host_userid_cpid (userid, host_cpid)");
+    $retval = $retval && do_query("alter table host drop index host_user");
     return $retval && do_query("alter table host add index host_domain_name (domain_name)");
 }
 
--- a/db/constraints.sql
+++ b/db/constraints.sql
@@ -83,6 +83,10 @@ alter table msg_to_host
 alter table host
     add index host_user (userid),
         -- html_user/host_user.php
+    add index host_cpid (host_cpid),
+        -- scheduler request for user with many hosts
+    add index host_domain_name (domain_name),
+        -- scheduler request for user with many hosts
     add index host_avg (expavg_credit desc),
         -- db_dump.C
     add index host_tot (total_credit desc);
--- a/db/constraints.sql
+++ b/db/constraints.sql
@@ -81,12 +81,11 @@ alter table msg_to_host
         -- for scheduler
 
 alter table host
-    add index host_user (userid),
+    add index host_userid_cpid (userid, host_cpid),
         -- html_user/host_user.php
-    add index host_cpid (host_cpid),
-        -- scheduler request for user with many hosts
+        -- sched/handle_request.cpp for user with many hosts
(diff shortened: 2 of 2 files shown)


html/ops/db_update.php
```diff
@@ -1162,6 +1162,12 @@ function update_5_9_2018() {
     ");
 }
 
+function update_8_23_2018() {
+    $retval = do_query("alter table host add index host_cpid (host_cpid)");
+    return $retval && do_query("alter table host add index host_domain_name (domain_name)");
```
AenBleidd: I understand that it is a common practice in this script to add indexes this way but why don't we check index existence before adding it? I'm not very familiar with MySql but it's can be done easily in MSSQL and I'm pretty sure that there is a way to do the same here.

TheAspens: So I looked into this and the easiest way to do this is simply to check the errno returned if the create failed and if it is 1061 then it means that the key name is a duplicate.  I was going to implement this but then I started thinking through different scenarious about what could be going on and I came up with the following:

1) The project created their own index that is named the same.  In this case stopping processing and having the project take a look at what is going on is the right thing.
2) The project applied the index early by seeing a needed index and adding it.  In that case it would be ok to skip adding it.

I didn't come up with other scenarios.   Given that these two scenarios suggest different behavior with regards to the appropriate action to take I'm not sure how to improve on the existing behavior.

The script does inform the user of the following in the event of a failure:
```
performing update update_8_23_2018
Doing query:
alter table host add index host_cpid (host_cpid)
Failed:
Duplicate key name 'host_cpid'
WARNING: database upgrade failed.
MySQL error message: Duplicate key name 'host_cpid'
Please find the update queries in html/ops/db_update.php
and perform them manually.
When done, edit PROJECT_DIR/db_revision so that it contains the line 27027
```
which provides a pretty straight forward instructions on how to respond.

thoughts?

brevilo: Hm, what happens if there's already an index for the column in question but under a different name? Never tried that, I think...

brevilo:
Kevin, I couldn't find the actual queries that you're trying to support here. Could you please provide those? That'll help reviewing this PR.

TheAspens:
For the index host_domain_name the query is found in sched/handle_request.cpp function find_host_by_other(): 
```
select * from host where userid=? and id>? and domain_name=? and last_ip_addr=? and os_name=? and p_model=? and m_nbytes=? order by id desc
```

For the index host_cpid the query is found in sched/handle_request.cpp function  find_host_by_cpid():
```
select * from host where userid=? and host_cpid=? order by id desc
```

Both of these indexes become important when a single user starts to have many host entries (for example a university lab that re-images devices nightly or weekly or that backfills a grid cluster).   We needed these on WCG for this reason and when Eric Driver reported the same issue on boinc_dev we realized it was an issue for more than just us so we created this pull request.

brevilo:
That's what I thought.

In the case of the second query, however, I recommend to create a compound/composite index which replaces the existing regular index on `userid`: `alter table host add index host_userid_cpid (userid, host_cpid)`. The replacement will support queries filtering on **both**, `userid` **and** `userid, host_cpid` at the same time, while avoiding the overhead of maintaining an additional index.

The first query is a more complex one for the optimizer to deal with and adding the additional index on domain_name might help. Have you ever tested which index gets used for the query above after you added the extra one? Granted, such a test result depends on the index cardinalities (read: your particular DB content) and won't be universally true anyway.

TheAspens:
```
MariaDB [BOINC]> explain select * from host where userid=X and id>Y and domain_name='localhost' and last_ip_addr='1.1.1.1' and os_name='Linux' and p_model='Fast CPU' and m_nbytes=1024 order by id desc
    -> ;
+------+-------------+-------+-------+------------------------------------+------------------+---------+------+------+-------------+
| id   | select_type | table | type  | possible_keys                      | key              | key_len | ref  | rows | Extra       |
+------+-------------+-------+-------+------------------------------------+------------------+---------+------+------+-------------+
|    1 | SIMPLE      | host  | range | PRIMARY,host_user,host_domain_name | host_domain_name | 261     | NULL |   11 | Using where |
+------+-------------+-------+-------+------------------------------------+------------------+---------+------+------+-------------+
1 row in set (0.01 sec)
```

TheAspens:
Modified as suggested

html/ops/db_update.php
brevilo: @TheAspens in case you missed it, there's still this open question of mine 😏

TheAspens:
@brevilo - If you get a chance can you go ahead and check my changes and if you are satisified merge?  Thanks!

brevilo:
Sure, but please have a look at https://github.com/BOINC/boinc/pull/2655#discussion_r215924659 first.

html/ops/db_update.php
TheAspens: MariaDB happily adds a index with a different name even if it duplicates an existing index

TheAspens:
Responded.

PR #2659, opened 2018-08-25
Travis: restructure travis configuration

ChristianBeer:
Instead of excluding unwanted jobs this now specifically includes the jobs needed. This allows to run multiple environments or languages within a single repository. This is not fully documented but was confirmed to be supported by Travis via email.

Configuration can be tested on this page: http://config.travis-ci.org/
Original idea came from this blog entry: http://axelnilsson.tech/2017/08/04/travisci-with-multi-language-repository/

A problem with `libcurl` and the updated Travis CI OSX VM occurred because the buildcache needed to be rebuilt. Therefore I added some code to get more information from failed builds and fixed the problem with `libcurl`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,12 @@
 language: cpp
-os:
-  - linux
-  - osx
 compiler:
   - gcc
-#  - clang
 
 branches:
   only:
     - master
     - coverity_scan
 
-#notifications:
-#  email:
-#    recipients:
-#      - boinc_cvs@ssl.berkeley.edu
-#    on_success: always
-#    on_failure: always
-
-# Use Trusty image and enable sudo to install apt dependencies
 dist: trusty
 sudo: required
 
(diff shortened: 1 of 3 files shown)


ChristianBeer:
@CharlieFenton could you please sign off the change in `mac_build/buildcurl.sh`? This should not affect your local dev environment.

@TheAspens or @Uplinger : since the changes are needed to proceed with adding integration testing and unit testing I would like to ask you to review and merge this.

CharlieFenton:
@ChristianBeer I tested the change in `mac_build/buildcurl.sh` and I approve it.

Uplinger:
@ChristianBeer I have reviewed the changes and they look good.  I'm assuming that with https://transfer.sh we don't need an account to upload the log files and can go over the 10GB since it clears them within 14 days.  I have not utilized that website and wanted to make sure it wouldn't cause issues when building say 100 times in 24 hours.  

Thanks,
-Keith

ChristianBeer:
They don't offer accounts and I don't know if the 10GB limit is per file or per IP. I only use it to upload the logfile which is <10MB in case of a build failure. Even if a lot of builds fail and we hit some kind of limit (because we share the IP with other Travis CI users) they would just block the upload and we don't get the log. In that case we can just restart the build later.

Uplinger:
I think this is great.  I have looked at their website and the "Become a Patron" is what I thought required a login possibly.  However if we like the service we could fire up our own (highly unlikely to happen) by pulling the source code from github as it is available.  Anyways, that's for the response and I'll merge it.  Thanks.

-Keith

PR #2667, opened 2018-08-28
Travis: Add integration testing to Travis CI

ChristianBeer:
This spins up a project using docker and runs some PHPUnit based tests against it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ env:
     - BOINC_TYPE=manager
     - BOINC_TYPE=libs-mingw
     - BOINC_TYPE=apps-mingw
-    - BOINC_TYPE=integration-test
+    - BOINC_TYPE=unit-test
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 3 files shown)


ChristianBeer:
@TheAspens this is now basically working. This is still WIP as I have a few questions and comments.
1. Where do we move the boinc-server-test repository to? As a separate repository in the BOINC organization or do we copy the files into the new `integration_test` directory?
2. I would like to have colored more verbose output from PHPUnit even if all test are successful.Right now it's just a boring grey line in the Travis log. A list of green checkmarks (or whatever PHPUnit can output) is a bit more impressive.
3. I looked into caching the docker layers and we would gain some time but this doesn't look trivial with the ansible setup we use. Right now we are still under 10 minutes runtime where the actual tests are only several seconds. So adding more tests does not add too much a burden and we are still far away from the 30 minute limit.

TheAspens:
I've add the repository here: https://github.com/BOINC/boinc-server-test

And issues: 
https://github.com/BOINC/boinc-server-test/issues/1
https://github.com/BOINC/boinc-server-test/issues/2

ChristianBeer:
Thanks for moving the directory. I changed the install script. Tests are still fine.

ChristianBeer:
I just removed the WIP tag as this is ready for review now. I just want to note that the actual unit tests are in a separate repository. This is fine with me for now but we should think about moving the tests into `integration_test` in the future.

TheAspens:
Looks good.  Thanks for doing this!

PR #2669, opened 2018-08-29
[WIP] Android ci

AenBleidd:


ChristianBeer:
Some observations regarding the travis jobs. I would like to keep the number of jobs small considering that we now need to specify each job separately. Therefore the `manager-android` job was meant to be a catch all job for all architectures (if possible). This would minimize the overhead added by the VM since the runtime of the individual build seems small enough.

Also there is not really a distinction between manager and client for android. It's one application that contains both. Therefore we should maybe use a better name and use `android-app` instead of `manager-android`. The build artifact from that job should be an apk file for each architecture.

AenBleidd:
I don't agree.
1 - As I can see it now, we can build and should build client separately from manager (we're not making a final product for deployment). I didn't try to build Manager (actually, it's my next step after finalizing and cleaning this script). If it is possible to build Manager separately from client - it will be great because it is better to see each 'component' build atomically to identify possible issue. Unless we have no limit to number of jobs - I think, it's better.
2 - I tried to build all 6 versions in one job and got too long log for Travis CI (actually, it is the main reason to split 1 job to 6).

ChristianBeer:
The problem with multiple jobs is this snippet:
```
    android:
        components:
          - android-23
          - build-tools-25.0.0
          - extra-google-m2repository
          - extra-android-m2repository
```
in order to have it applied to all android jobs you need to copy this to every android job in the matrix. Or install the dependencies all yourself. Also keep in mind that `buildCache` is shared between all android jobs.

The too long log error should be circumvented by limiting the log output to errors only or save it locally and upload in case of an error. See 69a96eb892479c62cacbacf1e29d4808831286d0 for my solution for OSX where I also ran into the too long log problem. Splitting the jobs is also possible but we still have a long log for each job. I usually suppress log output from building dependencies like openSSL or curl once I verified they are working. In my experience this makes the log more comprehensible.

I'm fine with splitting client and manager for Android if that makes sense for development and troubleshooting but I would still like to have one job that builds the apk that can be run on a weekly basis to have a current snapshot of the Android app that can be downloaded by testers.

AenBleidd:
This snippet is not necessary for Android client build.
Ok, I'll rework the script to run all configuration in one job by
suppressing log output for openssl and curl.
Thank you

Best regards,
Vitalii

Sent via Android

сб, 1 сент. 2018 г., 20:37 Christian Beer <notifications@github.com>:

AenBleidd:
@ChristianBeer, just another thing to think about:
There are 6 jobs now, each one takes ~9 minutes. While running in parallel it will take not longer than 20 minutes. But if they will be build as one jobe it will take ~50 minutes.
If we really want to have weakly build ready to be download then it will be easier to create some dedicated job that will be switched off be default and which can really run all these stuff within one build.
I think it's more reasonable. What do you think?

brevilo:
Hi guys. I know I'm late to the game but I'd like to chime in now 😬 

I'd like to support AenBleidd's motion to have separate jobs for the client(s) and manager as these are completely separate. Yes, the manager (read: BOINC Android app) ships (with) the clients for all supported platforms but they are built independently. Also, in terms of CI their build should be triggered whenever the `BOINC/client` (and presumably other parts of the) sources change, whereas the manager build should be triggered on `BOINC/android` changes. Yep, I'm no Travis expert so I'll leave this to you guys to figure out.

brevilo:
@AenBleidd, please let me know when your PR got adapted to #2680 and is ready for review.

AenBleidd:
Not needed anymore

PR #2672, opened 2018-08-29
clears the Launch services database during a mac client install

armstrdj:
This is a code change from Charlie Fenton which will clear the launch services database during a Mac install. The issue is that on a Mac machine which has built both generic boinc installers and branded boinc installers they may display the wrong app name and app icon in notifications.

### DIFF
--- a/mac_installer/PostInstall.cpp
+++ b/mac_installer/PostInstall.cpp
@@ -84,6 +84,9 @@
 #include <time.h>       // for time()
 #include <vector>
 #include <string>
+#define DLOPEN_NO_WARN
+#include <mach-o/dyld.h>
+#include <dlfcn.h>
 #include "url.h"
 
 #include "mac_branding.h"
@@ -116,6 +119,7 @@ OSErr GetCurrentScreenSaverSelection(passwd *pw, char *moduleName, size_t maxLen
 OSErr SetScreenSaverSelection(char *moduleName, char *modulePath, int type);
 void SetSkinInUserPrefs(char *userName, char *nameOfSkin);
 Boolean CheckDeleteFile(char *name);
+static void FixLaunchServicesDataBase(uid_t userID, long brandID);
 void SetEUIDBackToUser (void);
 static char * PersistentFGets(char *buf, size_t buflen, FILE *f);
 static void LoadPreferredLanguages();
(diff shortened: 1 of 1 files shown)


CharlieFenton:
Could a committer please merge this PR?

mac_installer/PostInstall.cpp
```diff
@@ -1220,6 +1224,95 @@ Boolean CheckDeleteFile(char *name)
...
+    }
+
+    saved_uid = geteuid();
+    CFStringRef bundleID = CFSTR("edu.berkeley.boinc");
+
+    if (LSCopyApplicationURLsForBundleIdentifier) { // Weak linked; not available before OS 10.10
+        seteuid(userID);    // Temporarily set effective uid to this user
+        appRefs = LSCopyApplicationURLsForBundleIdentifier(bundleID, NULL);
+        seteuid(saved_uid);     // Set effective uid back to privileged user
+            if (appRefs == NULL) {
```
AenBleidd: Additional indent level should be removed because it confuses a user who reads the code

mac_installer/PostInstall.cpp
```diff
@@ -1220,6 +1224,95 @@ Boolean CheckDeleteFile(char *name)
...
+                printf("Call to LSCopyApplicationURLsForBundleIdentifier returned NULL\n");
+                goto registerOurApp;
+            }
+            n = CFArrayGetCount(appRefs);   // Returns all results at once, in database order
+    } else {
```
AenBleidd: Could be moved outside of the else block before
`if (LSCopyApplicationURLsForBundleIdentifier) { // Weak linked; not available before OS 10.10`

mac_installer/PostInstall.cpp
```diff
@@ -1220,6 +1224,95 @@ Boolean CheckDeleteFile(char *name)
...
+        n = 500;    // Prevent infinite loop
+    }
+
+    for (i=0; i<n; ++i) {     // Prevent infinite loop
+        if (appRefs) {
+            CFURLRef appURL = (CFURLRef)CFArrayGetValueAtIndex(appRefs, i);
```
AenBleidd: What will be returned here in case of invalid i value?

mac_installer/PostInstall.cpp
```diff
@@ -1220,6 +1224,95 @@ Boolean CheckDeleteFile(char *name)
...
+                CFRetain(appURL);
+                CFStringRef CFPath = CFURLCopyFileSystemPath(appURL, kCFURLPOSIXPathStyle);
+                CFStringGetCString(CFPath, boincPath, sizeof(boincPath), kCFStringEncodingUTF8);
+                if (CFPath) CFRelease(CFPath);
+                CFRelease(appURL);
+                appURL = NULL;
+            }
+        } else {
+            seteuid(userID);    // Temporarily set effective uid to this user
+            // GetPathToAppFromID() returns only first result from database
```
AenBleidd: Please removed unused commented code

mac_installer/PostInstall.cpp
CharlieFenton: This is a void function so nothing is returned. But if it fails after 500 tries (extremely unlikely), it falls through to the code for not having found a previous entry. There is really nothing else that we can do in that situation. 

As the comment says, the for loop is there merely as a safety measure to guarantee there can never be an infinite loop here.

mac_installer/PostInstall.cpp
CharlieFenton: That is not unused code. It is a comment explaining that the the loop is needed because the  GetPathToAppFromID() API (called in the next line) returns only the first result from the database, explaining why the loop is needed.

mac_installer/PostInstall.cpp
AenBleidd: I understand, but as I understand from the name of the function, it returns a value by index from array. What will function do if there will be only 400 elements in array and you'll pass 401 as i parameter to this function? Will it crush or just do nothing?

mac_installer/PostInstall.cpp
CharlieFenton: It could, but I think it is better as is, because it is closer to the for() statement, improving readability. It's really a matter of coding style and individual preference.

mac_installer/PostInstall.cpp
AenBleidd: Sorry, misunderstood, my bad

mac_installer/PostInstall.cpp
AenBleidd: Ok, I prefer to get variable initialized with the default value but I'm not forcing you or anyone else to change this piece of code according to my preferences :)

mac_installer/PostInstall.cpp
AenBleidd: According to documentation https://developer.apple.com/documentation/corefoundation/1388767-cfarraygetvalueatindex?language=objc
passing invalid index leads to undefined behavior

CharlieFenton: Since only BOINC creates executables with that bundle ID, there should never be more than 2 or 3 on an end-user's system. The high number (500) is an aid to developers who may have many different built versions of BOINC on their hard drives, and chosen to be far greater than even a developer (like myself, who has been building BOINC Managers for many years) will ever have. Though only one version can be actually active at a time,  the Launch Services Database can keep references to all the ones that exist on the hard drives, so this is designed to deal with that.

CharlieFenton: if (apparefs) mans that `n = CFArrayGetCount(appRefs);` was called, not `n = 500;` so i is guaranteed valid.

AenBleidd: `CFArrayGetCount(appRefs)` is called only when `LSCopyApplicationURLsForBundleIdentifier` is true.
Sorry but I see here a potential issue.
I'm afraid, before calling `CFArrayGetValueAtIndex(appRefs, i)` length of the `appRefs` should be verified.
Maybe I'm wrong, so lets wait for review from someone more experienced than me

CharlieFenton: And _appRefs_ is non-null only when `LSCopyApplicationURLsForBundleIdentifier` is true, so `CFArrayGetValueAtIndex(appRefs, i)` is called only when `n = CFArrayGetCount(appRefs)` has been called. `if (apparefs)` means `LSCopyApplicationURLsForBundleIdentifier` is true which means `CFArrayGetValueAtIndex(appRefs, i)` has been called successfully. Note the line `CFArrayRef appRefs = NULL;` and the block:
```
if (appRefs == NULL) {
                printf("Call to LSCopyApplicationURLsForBundleIdentifier returned NULL\n");
                goto registerOurApp;
            }
```

AenBleidd: Ok, then this adds complexity:
If `appRefs' is non-null when `LSCopyApplicationURLsForBundleIdentifier` is true then there is no sense to iterate 500 times next code
```
err = GetPathToAppFromID('BNC!', bundleID,  boincPath, sizeof(boincPath));
            seteuid(saved_uid);     // Set effective uid back to privileged user
            if (err) {
                printf("Call %ld to GetPathToAppFromID returned error %d\n", i, err);
                break;
            }
```
BTW. it never uses i variable here. Maybe just another issue on the code

CharlieFenton: >If appRefs is non-null when LSCopyApplicationURLsForBundleIdentifier is true, then there is no sense to iterate 500 times next code.

Correct. That is why If `LSCopyApplicationURLsForBundleIdentifier` is true and succeeds, then `n = CFArrayGetCount(appRefs)`, **not** `n = 500`. Otherwise (`else`) `n = 500` and the following are called until either all entries have been removed or _i_ reaches 500:
`GetPathToAppFromID()` is called and the first entry is removed from the database.

I prefer the use of `else` here to make it clear these are two mutually exclusive situations. If the system is OS 10.10 or later, it uses `appRefs = LSCopyApplicationURLsForBundleIdentifier(bundleID, NULL);` and `n = CFArrayGetCount(appRefs)`. If the OS is OS 10.9.5 or earlier, n = 500 and it calls `GetPathToAppFromID()` as many times as necessary.

>BTW. it never uses i variable here to pass it to GetPathToAppFromID. Maybe just another issue on the code

In that case it uses `for (i=0; i<n; ++i)` with _n_ = 500as a safety measure to guarantee we never have an infinite loop.

I'm sorry, but I've explained all this several times now. @armstrdj and I have both thoroughly tested this code.

PR #2673, opened 2018-08-29
Updates for the wcgrid brand to support the mac installer

armstrdj:
Added graphics, license file, and readme file for the Mac installer.


TheAspens:
It is usually discouraged to have someone from the same organization review and merge the code from someone else in the same organization (@armstrdj and I both work on WCG).  However, in this case the changes are specific to the branded version of the installer for WCG and thus impact only WCG.  As a result, I will merge these changes.

PR #2675, opened 2018-08-30
client: parse and store user and team name from account manager.

davidpanderson:
If the AM is "dynamic", show those for projects attached via AM,
rather than e.g. "Science United user 123456"

### DIFF
--- a/html/inc/web_rpc_api.inc
+++ b/html/inc/web_rpc_api.inc
@@ -27,6 +27,7 @@ function fetch_url($url) {
     } else {
         $ch = curl_init();
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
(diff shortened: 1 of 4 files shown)


JuhaSointusalo:
@davidpanderson

I think you opened the PR from wrong branch. The description of this PR and the changes in it don't seem to be related at all.

TheAspens:
I've tested code as is and it works and they seem like reasonable changes.  However as per @JuhaSointusalo's  comment, the commit and pull request name and description don't match the code changes so I want to understand the mismatch.

ChristianBeer:
I fixed Scrutinizer but I had to reset the CodeSniffer and selected some arbitrary tests I think are resembling our coding style.

davidpanderson:
D'oh!  My bad.  Will do new PR

PR #2678, opened 2018-08-30
Test: Remove old test code

ChristianBeer:


ChristianBeer:
I want to start proper unit and integration testing with a clean slate. For this the old code is just a distraction.

I agree that the more abstract ideas of the tests, that describe what to test, are helpful. But this is better summarized on a wiki page not in several code files in different languages in a directory that can easily be mistaken to contain working test cases.

davidpanderson:
Yeah.  So name the new directory test_new

ChristianBeer:
Would it help if I rename "old" to "unmaintained"? There was at least one occasion in the past where a new contributor wanted to look at the test infrastructure and only found this part of the code. If we keep the code it's just distracting potential new contributors as they are looking at unmaintained code and not the actual tests that are executed. I never was a fan of having those `test_old`, `test_new`, test_old_backup` or `even_older_test` directories or files around. I mean that is what the VCS is for. We could place a tag on the last commit with the old test system and everyone interested can still access it through git history.

davidpanderson:
Maybe put unit test code in unit_tests/.
Do we have any end-to-end tests?
Those in test/ might still work.

lfield:
@ChristianBeer Do you have any updates for this?

ChristianBeer:
No real update just that I want to get rid of this stuff and make room for a new test hierarchy that could look like this:
```
tests/unit_tests # containing the actual unit test cases
tests/gtest # containing the test framework of choice (googletest in this case)
tests/integration_tests # containing the actual integration tests to be run
tests/docker_test # containing scripts to bootstrap the integration testing (essentially what is now in the boinc-server-test repo)
```
This would unclutter the base directory and new contributors could easily find the place where all tests are.

As a compromise I would suggest to rename the current `test` directory to `tests/old` and review deletion again in the future when a proper test system is in place.

lfield:
@ChristianBeer I agree with @davidpanderson that it is better to keep these tests visible until we have replacement tests. Hence I would suggest moving to tests/old/. If you agree I will close this PR and you can open a new one with the move. I can then merge.

ChristianBeer:
Closed in favor of #3097

PR #2686, opened 2018-09-09
Add languages to language selection menu for which we have translatio…

davidpanderson:
…n files

but which, for some reason, were missing from the list (?).
Namely: Azerbaijani, Basque, Italian, Norwegian, Persian, and Ukranian.


AenBleidd:
Fixed spelling of Ukrainian language string. As soon as it's my native language I can say that it is correct to have such spelling

TheAspens:
Thanks @davidpanderson and @AenBleidd !

PR #2689, opened 2018-09-10
[GUI] Tune simple preferences dialog

RichardHaselgrove:
Changes match those applied to the advanced dialog in #2585:

Make 'Save' the default button if (and only if) the user already has local preferences active
Show confirmation dialog when (and only when) switching from web to local preferences


TheAspens:
@AenBleidd - since you approved the changes and there have been no further comments in the past 7 days, can you go ahead and merge?  Thanks!

AenBleidd:
Merged. Thanks

PR #2699, opened 2018-09-20
[Android][CI] Android CI implementation

AenBleidd:
Initial Android client CI is implemented.
Some fixes in build scripts made to enable CI build cache for OpenSSL and CURL.

### DIFF
--- a/android/buildAndroidBOINC-CI.sh
+++ b/android/buildAndroidBOINC-CI.sh
@@ -1,0 +1,145 @@
+#!/bin/sh
+set -e
+
+# This file is part of BOINC.
+# http://boinc.berkeley.edu
+# Copyright (C) 2018 University of California
+#
+# BOINC is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License
+# as published by the Free Software Foundation,
+# either version 3 of the License, or (at your option) any later version.
+#
+# BOINC is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+export ANDROID_HOME=$HOME/Android/Sdk
+export NDK_ROOT=$HOME/Android/Ndk
+export ANDROID_TC=$HOME/Android/Toolchains
+
+# checks if a given path is canonical (absolute and does not contain relative links)
+# from http://unix.stackexchange.com/a/256437
+isPathCanonical() {
+  case "x$1" in
+    (x*/..|x*/../*|x../*|x*/.|x*/./*|x./*)
+        rc=1
+        ;;
+    (x/*)
+        rc=0
+        ;;
+    (*)
+        rc=1
+        ;;
+  esac
+  return $rc
+}
+
+doclean=""
+cache_dir=""
+arch=""
+while [ $# -gt 0 ]; do
+    key="$1"
+    case $key in
+        --cache_dir)
+        cache_dir="$2"
+        shift
+        ;;
+        --clean)
+        doclean="yes"
+        ;;
+        --arch)
+        arch="$2"
+        shift
+        ;;
+        *)
+        echo "unrecognized option $key"
+        ;;
+    esac
+    shift # past argument or value
+done
+
+if [ "x$cache_dir" != "x" ]; then
+    if isPathCanonical "$cache_dir" && [ "$cache_dir" != "/" ]; then
+         PREFIX="$cache_dir"
+     else
+         echo "cache_dir must be an absolute path without ./ or ../ in it"
+         exit 1
+     fi
+else
+    cd ../
+    PREFIX="$(pwd)/3rdParty/buildCache/android"
+    cd android/
+fi
+
+mkdir -p "${PREFIX}"
+
+if [ "${doclean}" = "yes" ]; then
+    echo "cleaning cache"
+    rm -rf "${PREFIX}"
+    mkdir -p "${PREFIX}"
+fi
+
+export COMPILEOPENSSL="no"
+export OPENSSL_VERSION=1.0.2p
+if [ ! -d "$PREFIX/openssl-${OPENSSL_VERSION}" ]; then
+    wget -O /tmp/openssl.tgz https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz
+    tar xzf /tmp/openssl.tgz --directory=$PREFIX4
+    export COMPILEOPENSSL="yes"
+fi
+export OPENSSL_SRC=$PREFIX/openssl-${OPENSSL_VERSION}
+
+export COMPILECURL="no"
+export CURL_VERSION=7.61.0
+if [ ! -d "$PREFIX/curl-${CURL_VERSION}" ]; then
+    wget -O /tmp/curl.tgz https://curl.haxx.se/download/curl-${CURL_VERSION}.tar.gz
+    tar xzf /tmp/curl.tgz --directory=$PREFIX
+    export COMPILECURL="yes"
+fi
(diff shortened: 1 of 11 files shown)


android/buildAndroidBOINC-CI.sh
```diff
@@ -0,0 +1,145 @@
...
+         echo "cache_dir must be an absolute path without ./ or ../ in it"
+         exit 1
+     fi
+else
+    cd ../
+    PREFIX="$(pwd)/3rdParty/buildCache/android"
+    cd android/
+fi
+
+mkdir -p "${PREFIX}"
```
ChristianBeer: You actually need two directories at this point. One where you are going to download the source and build the dependencies (`buildDir`) and one where you are going to store the build products of those dependencies (`cacheDir`). Right now you are using the same directory for both which bloats the cache. 

At the end of this script you need to set `ANDROID_TC` to be `cacheDir` so the other scripts following can install the libraries into the right directory. For the `buildDir` you need to have another directory specified. For Linux I hardcoded this to `3rdParty/linux`. Since you are downloading the sources here and export the build directory via a variable you can do the same or you could add another argument to the script so the `buildDir` can be specified.

I would set the variables like this for the default case:
```
BUILD_DIR="$(pwd)/3rdParty/android"
PREFIX="$(pwd)/3rdParty/buildCache/android-tc
```
Note that `PREFIX` now has the same meaning as in the Linux and Mac build scripts. Following comments will refer to those two variables.

android/buildAndroidBOINC-CI.sh
```diff
@@ -0,0 +1,145 @@
...
+    echo "cleaning cache"
+    rm -rf "${PREFIX}"
+    mkdir -p "${PREFIX}"
+fi
+
+export COMPILEOPENSSL="no"
+export OPENSSL_VERSION=1.0.2p
+if [ ! -d "$PREFIX/openssl-${OPENSSL_VERSION}" ]; then
+    wget -O /tmp/openssl.tgz https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz
+    tar xzf /tmp/openssl.tgz --directory=$PREFIX
```
ChristianBeer: use `$BUILD_DIR` instead of `$PREFIX`

android/buildAndroidBOINC-CI.sh
```diff
@@ -0,0 +1,145 @@
...
+    export COMPILEOPENSSL="yes"
+fi
+export OPENSSL_SRC=$PREFIX/openssl-${OPENSSL_VERSION}
```
ChristianBeer: use `$BUILD_DIR` instead of `$PREFIX`

android/buildAndroidBOINC-CI.sh
```diff
@@ -0,0 +1,145 @@
...
+
+export COMPILECURL="no"
+export CURL_VERSION=7.61.0
+if [ ! -d "$PREFIX/curl-${CURL_VERSION}" ]; then
+    wget -O /tmp/curl.tgz https://curl.haxx.se/download/curl-${CURL_VERSION}.tar.gz
+    tar xzf /tmp/curl.tgz --directory=$PREFIX
```
ChristianBeer: use `$BUILD_DIR` instead of `$PREFIX`

android/buildAndroidBOINC-CI.sh
```diff
@@ -0,0 +1,145 @@
...
+    export COMPILECURL="yes"
+fi
+export CURL_SRC=$PREFIX/curl-${CURL_VERSION}
```
ChristianBeer: use `$BUILD_DIR` instead of `$PREFIX`

android/buildAndroidBOINC-CI.sh
```diff
@@ -0,0 +1,145 @@
...
+
+export NDK_VERSION=17c
+wget -O /tmp/ndk.zip https://dl.google.com/android/repository/android-ndk-r${NDK_VERSION}-linux-x86_64.zip
+unzip -qq /tmp/ndk.zip -d $HOME
+export NDK_ROOT=$HOME/android-ndk-r${NDK_VERSION}
+
+export ANDROID_TC=$PREFIX/android-tc
```
ChristianBeer: use `$PREFIX/` instead of `$PREFIX/android-tc`

android/buildAndroidBOINC-CI.sh
AenBleidd: Done. Please make one more review. Thanks

ChristianBeer:
Changes look good but I just triggered a rebuild with an empty cache and it is again at 1024 MB. I tested the arm architecture on my desktop and can see that the NDK takes up a lot of space in the cache. It might not be preferable to put the NDK into the cache. See Travis Caching guidelines: https://docs.travis-ci.com/user/caching/#things-not-to-cache

> Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source

So when not caching the NDK we would save Travis some bandwidth (Caches are stored in AWS) and put it on Googles tap while only adding a small amount of build time to our jobs. My proposal would be to keep the cache small and not cache the NDK.

I'm doing a little experiment now. The current fresh build of `client-android-arm` took 8 minutes 8 seconds. I'm going to rebuild this job only and we can compare the time it takes to retrieve the 1GB cache without building the NDK.

ChristianBeer:
It seems we are getting a huge benefit from using the cache even when it is 1GB. This second run only took 5 minutes and that included downloading the NDK without installing it. There needs to be a flagfile guarding the NDK download as is with openssl and curl. This would be the last change and then it is ready to be merged (even with the big cache).

AenBleidd:
Done. @ChristianBeer, please take a look

brevilo:
Sorry for the delay. I had to deal with some urgent matters. I'll start to have a look now...

android/buildAndroidBOINC-CI.sh
```diff
@@ -0,0 +1,175 @@
...
+    echo "cleaning cache"
+    rm -rf "${PREFIX}"
+    mkdir -p "${PREFIX}"
+    echo "cleaning build dir"
+    rm -rf "${BUILD_DIR}"
+    mkdir -p "${BUILD_DIR}"
+fi
+
+export COMPILEOPENSSL="no"
+export OPENSSL_VERSION=1.0.2p
```
brevilo: Let's move any version details (OpenSSL, cURL, NDK) to the top of the file or, even better, refactor them into a separate config-like file in the `<root>/android` directory. Ideally that file would use a syntax that's usable by all of the build scripts (incl. `Vagrantfile`) to avoid redundant settings.

AenBleidd: @brevilo, It is a good idea to have config file. What exact type of config file will work for you to use it in Vagrant?

AenBleidd:
> Sorry for the delay. I had to deal with some urgent matters. I'll start to have a look now...

It's ok. I'm not in hurry with these tickets.
BTW, I'll be on vacation before Monday, 1st of October, so any changes and fixed could be implemented from my side starting from Monday only.

android/buildAndroidBOINC-CI.sh
brevilo: It would basically have to be ruby, so that's out of the question I suppose since I don't see how to combine bash and ruby declarations in the same file (they share the comment operator). So just factor out version details for all the shell/bash files first of all. Maybe we'll find a way to make use of those for Vagrant as well at some later stage.

Thanks

AenBleidd: @brevilo,
Hm, so it sounds like a refactoring. I think, this should be done out of the scope of current task.
Will it work for you for now just to move all version declarations to the top of the file and then add a new task to think how can we move all dependencies versions to dedicated file?

brevilo: Yeah, that'd be fine.

ChristianBeer: Have a look at https://github.com/BOINC/boinc/blob/master/mac_build/dependencyNames.sh where Charlie did something similar for the Mac build. The script gets `source`d from other build scripts and we only track the versions in one place. Maybe this can be used to create a ruby syntax on the fly with some `sed` rules?

But I also agree that this should be done afterwards.

AenBleidd:
Done, Ready for further review

PR #2707, opened 2018-09-21
client/scheduler: update NVidia detection

RichardHaselgrove:
for new architectures released since 2014, including RTX range released September 2018


davidpanderson:
I can't test this but it looks OK

RichardHaselgrove:
I got help from NVidia users at SETI and from Ray Hinchliffe, author of System Information Viewer. We couldn't test either until RTX deliveries commence, but I can supply sources for the information used. The code built cleanly, and we'll test for generated values once test builds are available.

PR #2710, opened 2018-09-21
[Android] Add Android Manager CI configuration

AenBleidd:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,12 @@ matrix:
     - language: android
       env: BOINC_TYPE=client-android-x86_64
     - language: android
+      android:
+        components:
+            - build-tools-28.0.2
+            - android-26
+            - extra-google-m2repository
+            - extra-android-m2repository    
       env: BOINC_TYPE=manager-android
 
 before_install:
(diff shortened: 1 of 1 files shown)


AenBleidd:
@ChristianBeer, here's only the Manager build without client build so I don't think that this apk will be useful for testing or something else.
Also I don't agree to add both Manager and client (4 configurations) into one build because it will take ~30 minutes to be built in one thread and it's pretty much, I'm afraid.
I'd rather consider to create separate configuration to build both Manager and client as one job for release branch only, because it makes more sense. First of all because 99% of commits in master are not related to Android application and will just slow the CI verification.
If you're not agree with you we need some kind of polling to identify which way is better:
- 5 separate jobs to verify whether master still could be built
- 1 job to get working Android apk.
Personally I (and @brevilo too) believe that first way is better but I'm open for discussion because I want to make the job that will be useful for everyone not just me.

ChristianBeer:
I've looked at #2699 and runtimes look fine. I'm not against the 5 separate jobs approach and see the benefits. If you could identify build artefacts from those 5 separate jobs that would be useful to make available to testers that would be great. If not, that's not a problem.
I would still like to have a 6th job that builds an (unsigned) apk though. This can be added so that it only builds for non pull requests (e.g. upcoming weekly build of master). Vice versa we can change the 5 separate jobs (or a subset) to only build when testing a pull request.

Do you have a preferred order when merging this in respect to #2699? I'm going to give this my thumbs up but please see the review of the other PR as there is still some work to do.

AenBleidd:
> If you could identify build artefacts from those 5 separate jobs that would be useful to make available to testers that would be great.

Ok, I'll do this.

> I would still like to have a 6th job that builds an (unsigned) apk though. This can be added so that it only builds for non pull requests (e.g. upcoming weekly build of master).

I'd prefer this way. I can do this when weekly build implementation will be finished.

> Do you have a preferred order when merging this in respect to #2699?

No, they are not linked to each other so can be merged separately in any order.

> but please see the review of the other PR as there is still some work to do.

I guess, you're talking about #2699 ? Ok, I'll wait for your review and make appropriate fixes there.

brevilo:
FWIW, I concur with the CI job separation as outlined above.

.travis.yml
```diff
@@ -51,6 +51,12 @@ matrix:
       os: osx
       env: BOINC_TYPE=manager-osx
     - language: android
+      android:
+        components:
+            - build-tools-28.0.2
+            - android-26
```
brevilo: Hm, can you envision a way to extract those details from https://github.com/BOINC/boinc/blob/master/android/BOINC/app/build.gradle#L49 and use/integrate them here? Can you include other files (with variable definitions) in the travis config? The idea here again would be to avoid redundant config settings that have to be kept in sync manually otherwise - somethings that's asking for inconsistencies and thus problems.

AenBleidd: I'm afraid no. @ChristianBeer could know more about this

ChristianBeer: I don't know a way to include other files into the travis config. I guess we have to put a comment in the gradle file to remind whoever changes it to also change the CI system dependency. Of course this point is mood if the gradle file is changed by a UI like Android Studio.

ChristianBeer:
Could you please rebase your commit on current master as the merge of #2699 created a conflict in .travis.yml? I'm going to merge after that if @brevilo doesn't change his review.

Issue #2029, opened 2017-08-13
(Windows) GPU detection fails when started from command line


RichardHaselgrove:
Previous reference https://lists.ssl.berkeley.edu/pipermail/boinc_alpha/2017-July/021908.html

If the client is started from the Windows command line with

prompt>boinc --detach_console

GPU detection fails, and any previous coproc_info.xml file is deleted.

Starting the client with

prompt>boinc.exe --detach_console

allows GPUs to be detected and used as normal.

Changing the commandLine used in

https://github.com/BOINC/boinc/blob/master/client/main.cpp#L488

to use an explicit .exe extension also solves the problem - as was found by the author of https://github.com/BOINC/boinc/blob/master/client/gpu_detect.cpp#L640 (though I prefer the idea of passing the command line here too, rather than hard-coding the executable name)

Tests reconfirmed with code as at the v7.8.1 tag.

Ageless93:
Is PR #2036 a fix for this problem? (https://github.com/BOINC/boinc/pull/2036)

Ageless93:
Seeing how PR#2036 was added to Master, I updated my local repository and built a new client and manager from Master. Then ran those from the command line, with `boinc --detach_console`, and got this:

```
20-Aug-2017 00:21:39 [---] Starting BOINC client version 7.8.3 for windows_x86_64 (DEBUG)
20-Aug-2017 00:21:39 [---] This a development version of BOINC and may not function properly
20-Aug-2017 00:21:39 [---] log flags: file_xfer, sched_ops, task, benchmark_debug, checkpoint_debug
20-Aug-2017 00:21:39 [---] log flags: cpu_sched, file_xfer_debug, sched_op_debug, task_debug
20-Aug-2017 00:21:39 [---] Libraries: libcurl/7.42.1 OpenSSL/1.0.2a zlib/1.2.8
20-Aug-2017 00:21:39 [---] Data directory: M:\ProgramData\BOINC
20-Aug-2017 00:21:39 [---] Running under account Ageless
20-Aug-2017 00:21:39 [---] Project for statistics_abcathome.com.xml not found - ignoring
20-Aug-2017 00:21:39 [---] Project for statistics_albert.phys.uwm.edu.xml not found - ignoring
20-Aug-2017 00:21:39 [---] Project for statistics_aqua.dwavesys.com.xml not found - ignoring
20-Aug-2017 00:21:39 [---] Project for statistics_convector.fsv.cvut.cz.xml not found - ignoring
[...]

PR #2713, opened 2018-09-22
client: fixes for launching GPU detection

### LINKED ISSUE #2657: (Windows) GPU detection fails when started from command line (take 2)
Previous references: issue #2029 , PR #2036.

A year ago we tackled an issue where GPU detection failed if the client was started from the command line with "boinc", rather than "boinc.exe"

The problem has been re-reported for BOINC v7.10.2 and v7.12.1, but this time when the invocation is from a batch file or from BoincTasks, if the correct working directory is not used. Compare these screenshots (from Process Explorer)

boinc client 2 - v7 8 3
boinc client 2 - v7 12 1

After all the multiple restarts to detach from the command window and do the GPU detection, the final running client instance was still started with the full path on the command line (D:\BOINC\... on my machine). By contrast, v7.12.1 starts with the now hard-wired "boinc.exe", but no path.

I think commit 7930686128709433b29741a034dd890c27a616f1 just moved the problem around a bit, without fully fixing it. We need both the path and the .exe

JuhaSointusalo:
Launching GPU detection has a few issues.
- doesn't work reliably with `--detach_console`
- doesn't work if client is run from `$PATH`
- if it fails the user is presented with hard to understand error code
- the code throws compiler warnings

This PR fixes those issues. To support the fixes `get_real_executable_path()` is implemented for several more operating systems and `run_program()` is changed to use `$PATH`. See individual commit messages for details.

Fixes #2657.

### DIFF
--- a/client/gpu_detect.cpp
+++ b/client/gpu_detect.cpp
@@ -661,8 +661,6 @@ int COPROCS::launch_child_process_to_detect_gpus() {
          NULL
     }; 
 
-    chdir(client_dir);
-    
     retval = run_program(
         client_dir,
         client_path,
@@ -672,8 +670,6 @@ int COPROCS::launch_child_process_to_detect_gpus() {
         prog
     );
 
-    chdir(data_dir);
-    
     if (retval) {
         if (log_flags.coproc_debug) {
             msg_printf(0, MSG_INFO,
--- a/client/gpu_detect.cpp
+++ b/client/gpu_detect.cpp
@@ -595,6 +595,7 @@ int COPROCS::launch_child_process_to_detect_gpus() {
 #else
     int prog;
 #endif
+    char quoted_client_path[MAXPATHLEN];
     char quoted_data_dir[MAXPATHLEN+2];
     char data_dir[MAXPATHLEN];
     int retval = 0;
@@ -612,20 +613,31 @@ int COPROCS::launch_child_process_to_detect_gpus() {
         }
     }
     
+    // use full path to exe if possible, otherwise keep using argv[0]
+    char execpath[MAXPATHLEN];
+    if (!get_real_executable_path(execpath, sizeof(execpath))) {
+        client_path = execpath;
+    }
+
     boinc_getcwd(data_dir);
 
 #ifdef _WIN32
+    strlcpy(quoted_client_path, "\"", sizeof(quoted_client_path));
+    strlcat(quoted_client_path, client_path, sizeof(quoted_client_path));
+    strlcat(quoted_client_path, "\"", sizeof(quoted_client_path));
+
     strlcpy(quoted_data_dir, "\"", sizeof(quoted_data_dir));
     strlcat(quoted_data_dir, data_dir, sizeof(quoted_data_dir));
     strlcat(quoted_data_dir, "\"", sizeof(quoted_data_dir));
 #else
+    strlcpy(quoted_client_path, client_path, sizeof(quoted_client_path));
     strlcpy(quoted_data_dir, data_dir, sizeof(quoted_data_dir));
 #endif
 
     if (log_flags.coproc_debug) {
         msg_printf(0, MSG_INFO,
             "[coproc] launching child process at %s",
-            client_path
+            quoted_client_path
         );
         msg_printf(0, MSG_INFO,
             "[coproc] relative to directory %s",
@@ -639,7 +651,7 @@ int COPROCS::launch_child_process_to_detect_gpus() {
             
     int argc = 4;
     char* const argv[5] = {
-         const_cast<char *>(CLIENT_EXEC_FILENAME), 
+         const_cast<char *>(quoted_client_path),
          const_cast<char *>("--detect_gpus"), 
          const_cast<char *>("--dir"), 
          const_cast<char *>(quoted_data_dir),
--- a/client/gpu_detect.cpp
+++ b/client/gpu_detect.cpp
@@ -58,6 +58,7 @@
 
 #ifdef _WIN32
 #include "boinc_win.h"
+#include "win_util.h"
 #ifdef _MSC_VER
 #define snprintf _snprintf
 #define chdir _chdir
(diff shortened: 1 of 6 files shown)


JuhaSointusalo:
@CharlieFenton Please test.

`get_real_executable_path()` is implemented on OSX using `_NSGetExecutablePath()`. Since GPU detection falls back to `argv[0]` if `get_real_executable_path()` fails you'll need to add some test code. I used something like this at the top of client's `main()`:
```c
char path[MAXPATHLEN] = "";
int ret = get_real_executable_path(path, sizeof(path));
fprintf(stderr, "ret %d path %s\n", ret, path);
exit(0);
```
If `_NSGetExecutablePath()` doesn't work or you think it's not appropriate for the job an alternative I found is sysctl `{CTL_KERN, KERN_KERN_PROCNAME}` (`kern.procname`), though that one looks like not exactly well documented.

RichardHaselgrove:
Built and tested under Windows 7 / VS2013.

Checked both problems reported in #2657 and #2029: successful GPU detection in both cases.

CharlieFenton:
@JuhaSointusalo: `_NSGetExecutablePath()` works correctly and is appropriate. However, I disagree in part with your commit 1296c86. I agree that the command `chdir(client_dir)` is redundant, since that is also done inside `run_program()`. However, I believe that without `chdir(data_dir)` after `run_program()` to restore the working directory, the original instance of the client will no longer properly access the BOINC Data directory. Am I missing something here?

JuhaSointusalo:
> However, I believe that without `chdir(data_dir)` after `run_program()` to restore the working directory, the original instance of the client will no longer properly access the BOINC Data directory.

On Windows, the directory that is passed to `run_program()` is only used in a call to `CreateProcess()`. `CreateProcess()` shouldn't modify the parent process in any way.

On others, the directory is used in a call to `chdir()` but that happens after `fork()`, that is, it's changing the current working directory of the child process.

https://github.com/BOINC/boinc/blob/460a6dba1477e222621c2f0b2fc1288d992a8466/lib/util.cpp#L471-L476

Maybe it's the `return` after failed `chdir()` that's confusing here. That's actually a bug in `run_program()` and the code should call `_exit()` instead. If the code returns we'll have two processes running, the parent and a crippled copy of it.

I intend to review all child process launching code in BOINC to see if these bugs are made elsewhere but that's another day another PR.

davidpanderson:
Has anyone tested this on Mac?
I was hoping to make the 7.14 branch today.

JuhaSointusalo:
@CharlieFenton 

I logged current working directory before and after `run_program()` and in both log entries it was the data directory.

RichardHaselgrove:
Although Windows isn't the primary concern of recent comments, the private build I made yesterday has been running since

23-Sep-2018 14:07:24 [---] Starting BOINC client version 7.13.0 for windows_x86_64

The report I made earlier on tests for the original reported problems was with a secondary instance: the primary client has been running constantly and updating files in the data directory as normal. The list of files updated since I started typing this report is

client_state.xml
daily_xfer_history.xml
stdoutdae.txt
client_state_prev.xml
job_log_setiathome.berkeley.edu.txt
statistics_setiathome.berkeley.edu.xml
sched_reply_setiathome.berkeley.edu.xml
sched_request_setiathome.berkeley.edu.xml
job_log_einstein.phys.uwm.edu.txt

CharlieFenton:
@JuhaSointusalo: You are correct. I misread the run_program code. I now see that `chdir(dir)` is in the forked process, not the parent. Since the arguments passed to the child include `-dir quoted_data_dir`, the _client_dir_ argument to `run_program` is not needed and NULL could be passed instead. If you do that, **all** references to _client_dir_ should be removed, including the entire `COPROCS::set_path_to_client()` method and the call to it  in `main()`.

This _log_flags_ message should also be removed, as it is no longer correct since you have modified the code to use an absolute path:
```
       msg_printf(0, MSG_INFO,
            "[coproc] relative to directory %s",
            client_dir
        );

```

CharlieFenton:
> Has anyone tested this on Mac?
I was hoping to make the 7.14 branch today.

@davidpanderson: I tested it and further changes are needed, as I wrote in my last comment. At the very least, the incorrect _log_flags_ message must be removed. All references to the now unnecessary static variable _client_dir_ should also be removed, including  the entire `COPROCS::set_path_to_client()` method and the call to it in `main()`.

davidpanderson:
Juha, can you please address Charlie's suggestions?
Thanks -- D

JuhaSointusalo:
> since you have modified the code to use an absolute path

Well, the code _tries_ to use absolute path but that doesn't work everywhere. I know it doesn't work on OpenBSD at all because OpenBSD provides no way to find out the filename and path of the current program. It also doesn't work on NetBSD 7 if `/proc` is not mounted (the sysctl is NetBSD 8 thing). I don't know for sure it won't ever fail on other systems.

One could argue that OpenBSD doesn't support GPU crunching and never will and we don't need to care about it but I still would like to keep `argv[0]` as a backup. `argv[0]` could be something like `./boinc` and then we need to know what directory the client was started in. Calling the variable `client_dir` is a bit misleading but I'm not sure what to call it then. `dir_that_was_cwd_at_startup`?

Sidenote: I suppose implementing `get_real_executable_path()` on BSDs and Solaris was kinda unnecessary. I was searching for something and came across info on how to find out executable path on those. So I figured why not use the info, if we ever decide to have good support for BSDs then there won't be need to find that information again.

CharlieFenton:
>  I still would like to keep argv[0] as a backup

Fair enough. Perhaps this is getting too fussy, but it would be nice if this log_flags message would appear **only** when `get_real_executable_path()` failed:
```
        msg_printf(0, MSG_INFO,
            "[coproc] relative to directory %s",
            client_dir
        );

```

JuhaSointusalo:
@CharlieFenton Ready for re-review.

CharlieFenton:
@JuhaSointusalo I tested this and it works as intended. Thank you. I approve this PR for merging into master.

Issue #2696, opened 2018-09-19
Devise mechanism for updating all_projects_list.xml in github


RichardHaselgrove:
all_projects_list.xml is dynamic data which is distributed to new users as a part of the installation process. Historically, it's held in this repository at master/win_build/installerv2/redist (and hence automatically incorporated in new release branches): although the location flags it as a Windows build file, the same resource is also used in Mac installer builds.

The file is also fetched periodically (every 14 days) by the BOINC client, unless disabled by corporate policies like <dont_contact_ref_site>.

At the moment, the file in master is 5 months old, and thus breaks the policy that master should be held in a state of readiness for the Release Manager when he or she decides to make a new client release branch.

This new issue is intended to act as a reminder that we have 'unfinished business' from the last few client releases, so that previous references like #2247 and #2579 can be closed or remain closed. This issue should be read in conjunction with @ChristianBeer comments in #2642, where the confusion between 'project config' and 'master' urls is discussed: this affects whether http or https is used in the primary url field in all_projects_list - see note on Einstein@Home in attached spreadsheet showing the current differences between versions of the file.

[Comparison of All_Projects_List versions 2018-09-19.zip](https://github.com/BOINC/boinc/files/2396794/Comparison.of.All_Projects_List.versions.2018-09-19.zip)

KeithMyers:
Einstein has changed their URL to https now.  All the BOINC files will need to be changed also.

RichardHaselgrove:
@KeithMyers - see the discussion on #4926. I think the Einstein change was in response to @AenBleidd's email to the projects list: one down, six to go.

But all should see the discussion in https://einsteinathome.org/content/master-url-changed-use-https: the change hasn't been welcomed in all quarters, especially by those users not yet running v7.16.20 or later.

KeithMyers:
But of course you know the Catch-22 people are faced with when responding to the advisory messages about being connected twice to Einstein.  If you follow the instructions to remove and rejoin the project, you are going to use the Join Project mechanism which relies on the all_projects_list.xml file which still has Einstein listed with a http URL.

So you achieve nothing by taking action on the advisory.  Until that projects list file gets updated with the correct https URL for Einstein you are back to the original problem. Same thing applies to all the other projects still on http URL's.

I've been responding to posts in the forums complaining that the project managers aren't doing their job and I have to explain the project has nothing to do with the issue and the BOINC devs need to resolve it.

[...]

Issue #2642, opened 2018-08-09
Make attach process the same for Client and Manager


ChristianBeer:
Currently the Manager uses a 4 step process to attach the Client to a project:

1. The User selects or enters a `URL` in the Attach project Wizard. The Manager makes a GET request to `URL/get_project_config.php` to fetch the `master_url`
1. The Manager tells the Client to attach to `master_url`
1. The Client fetches the content of `master_url` to scrape the `scheduler_url` from the HTML
1. The Client uses the `scheduler_url` to actually attach to the project

This is a problem when a User uses `boinccmd` to attach to a project on the commandline since he/she has to use the ` master_url` or there will be an annoying message in the Client logfile each time the Client connects to the project. The process using `boinccmd` instead of the Manager looks like this:

1. The User calls `boinccmd --project_attach URL account_key`
1. The `boinccmd` tool tells the Client to attach to `URL`
1. The Client treats `URL` as the `master_url` and fetches the content of `URL` to scrape the `scheduler_url` from the HTML
1. The Client uses the `scheduler_url` to actually attach to the project

The problem is that there are projects that have a different `master_url` and `website_url`. For example Einstein@home redirects all web requests to `https://einsteinathome.org` but for historical reasons the `master_url` in `https://einsteinathome.org/get_project_config.php` is still `http://einstein.phys.uwm.edu/` (because it is not possible to just change this without all users re-attaching to the project).

For consistency either the `boinccmd` tool or the Client itself should do a request to `URL/get_project_config.php` like the Manager does.

Issue #1345, opened 2015-02-04
BOINC should push projects to use HTTPS for better security


romw:
**Reported by bryanquigley on 12 Nov 44964364 11:49 UTC**
Right now all downloads happen over HTTP and AFAICT the only protection is code signing and the sandbox.  HTTPS projects would have another layer protecting against MITM attacks.

Code signing won't protect against tampering with the data and/or results back to the project.  Tampering with the data as it goes to the client could lead to an exploit.

In all I'm just asking if BOINC can change to explicitly recommend using HTTPS for the project website and project URL. 

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1374

romw:
**Commented by bryanquigley on 6 Jan 44964369 12:32 UTC**
I would be happy to help with work to make the BOINC website more secure to lead by example.

romw:
**Commented by Nicolas on 21 Mar 45018814 12:54 UTC**
Using HTTPS for the scheduler requests would be enough, as it would protect the MD5 hashes sent by the server. There's no need to use HTTPS for the downloads themselves too.

romw:
**Commented by bryanquigley on 22 Sep 45020797 06:32 UTC**
If it was just an md5sum it wouldn't be enough given exploits against it.. I believe if you check the size though it might be (for now).

Opened a new ticket for supporting sha256 (https://boinc.berkeley.edu/trac/ticket/1375)

grctest:
Pushing BOINC projects to use HTTPS for everything would be great, I've used Letsencrypt via EFF's certbot to quickly provide SSL for my BOINC project.

The Gridcoin community has been attempting to chase up projects regarding implementing SSL (only 4 have not replied): https://cryptocointalk.com/topic/49717-continued-boinc-project-ssl-discussion/

This recent boinc_project mailinglist thread is interesting & talks about implementing SSL (whilst redirecting all HTTP traffic to HTTPS): http://lists.ssl.berkeley.edu/mailman/private/boinc_projects/2016-September/011836.html

I raised the question:
I've been discussing potential security issues with the BOINC web server in the BOINC_Dev mailinglist and found that the majority of BOINC projects in the 'add project' window do not list the URL as HTTPS despite its availability, is this something else we'll need to change or is it purely a cosmetic issue? 
https://github.com/BOINC/boinc/blob/master/doc/projects.inc

Preferably, nothing vital like password hashes/account keys should be sent plain-text over the public Internet.

[...]

PR #2716, opened 2018-09-25
Add project

davidpanderson:


RichardHaselgrove:
I've checked the list against the comparison chart I attached to #2696, and all seems good - including the new project, DBN Upper Bound, which has not yet been publicly welcomed.

I'm still a bit concerned that our usage of HTTPS is inconsistent. Now that the <web_url> tag is active, we could and perhaps should set the web_urls for these projects to https:

https://www.rnaworld.de/rnaworld/
https://www.gpugrid.net/
https://boinc.bakerlab.org/rosetta/
https://climateprediction.net/
https://asteroidsathome.net/boinc/
https://www.cosmologyathome.org/
https://milkyway.cs.rpi.edu/milkyway/
https://setiathome.berkeley.edu/
https://www.rechenkraft.net/yoyo/
https://www.worldcommunitygrid.org/
https://moowrap.net/
https://www.primegrid.com/
https://www.primaboinca.com/
https://boinc.multi-pool.info/latinsquares/

And, for completeness, these projects cannot yet use https to access their websites. Perhaps we could nudge them gently?

http://quakecatcher.net/sensor
http://radioactiveathome.org/boinc/
http://www.acousticsathome.ru/boinc/
http://casathome.ihep.ac.cn/
http://sat.isa.ru/pdsat/
http://www.enigmaathome.net/
http://gerasim.boinc.ru/
http://srbase.my-firewall.org/sr5/
http://anthgrid.com/dbnupperbound/

Finally, I'm still not quite sure why the plain < url > tags for

http://denis.usj.es/denisathome/
http://einstein.phys.uwm.edu/

have been reverted to http this time. Both projects support using https in

Fetching configuration file from https://denis.usj.es/denisathome/get_project_config.php
Fetching configuration file from https://einstein.phys.uwm.edu/get_project_config.php

which is the primary purpose of this commit. I'm aware of the command line problem raised by @ChristianBeer in #2642, but how many users use this file as the source for command line urls? The number of users still using Version 4 or earlier BOINCs is tiny, and the only other users I can think of being affected are

a) self-ports to unsupported operating systems
b) bulk deployments by corporate users wishing to maintain central control

I think both classes could be expected to do their own research and apply the master url directly where needed?

AenBleidd:
I think http://sat.isa.ru/pdsat/ has no work for 2 years already as far as I heard they have no plans to continue but I can be wrong.

I can gently push guys of these projects on Russian boinc forum :
http://www.acousticsathome.ru/boinc/
http://gerasim.boinc.ru/

ChristianBeer:
For me the commit message is very non-descriptive and doesn't reflect what is in the code changes. I can already hear someones head banging on his desk in the future because he/she wants to know why the changes were made and "Add project" just doesn't do that.

As for the changes related to Einstein@Home they are fine if the intention of <url> is to represent the `master_url` and <web_url> represents the url to be used in the browser by the user.

As for the general separation of <url> and <web_url> I think that was needed because the file is used in multiple places where the distinction was needed. I can't remember the exact details. But the `master_url` is also needed by users who install BOINC on systems with no UI, for example RaspberryPi or similar boards where they attach to existing accounts via `boinccmd`.

TheAspens:
My understanding is that the source of the all projects list is this: 

https://github.com/BOINC/boinc-site/blob/master/projects.inc
which is used by 
https://boinc.berkeley.edu/project_list.php
and the output of that creates the file submitted in this pull request.

If we want to get things updated in the original source, then we need to submit a pull request against https://github.com/BOINC/boinc-site/blob/master/projects.inc with the needed changes.  @davidpanderson  - is this correct?

I think that things like trying to persuade projects to use things like Let's Encrypt or otherwise move to https should be handled and discussed on tickets outside of this as those requests confuse the action needed on this pull request.  Issue #1345 that @RichardHaselgrove refers to is a great place to discuss that.

@RichardHaselgrove  - could you create a pull request against https://github.com/BOINC/boinc-site/blob/master/projects.inc that proposes what needs to be corrected based on how things are today (not how we wish they might be in the future).

RichardHaselgrove:
@TheAspens: My recollection is that https://github.com/BOINC/boinc-site/blob/master/projects.inc was cited as the definitive source during our working party deliberations, too. And that file was updated 11 days ago, so we should be close to equivalence. I'll try and work up a pull request tomorrow, in time for the committers' call on Thursday. The change of format (data to code) will slow things down, as will validating the other data sources (keywords via KeywordID, plan classes). I think I feel a database coming on...

"What needs to be corrected [today]" would be based on a bias towards using https wherever supported, especially for <web_url> which is used for browser access only: my feeling is (as it was during my stint as release manager) that this presents a professional image for both BOINC and the projects in this age of the promotion of browser security.

The join/master url debate still needs to be resolved through #2642, but I think @ChristianBeer's direction of travel there is to resolve the command line (boinccmd) attachment problem by implementing get_project_config there too. Which leaves Account Managers...

CharlieFenton:
Both the Mac and Windows installers get the all projects list from the release branch at the following path:`win_build/installerv2/redist/all_projects_list.xml`. So, from the point of view of building a public release, the copy in the release branch at that location is the one that matters.

I programmed the Mac installer so that it writes the _all_projects_list.xml_ file to the _BOINC Data_ directory ** only** if there is not already one there (in the case of installing over a previously installed BOINC.) This is because an exiting _all_projects_list.xml_ file will usually be more recent, since the client checks the server every 14 days and downloads the file if the one on the server is newer than the local copy. However, I don't think the Windows installer has this logic.

CharlieFenton:
I have confirmed that the _all_projects_list.xml_ file added to the 7.14 branch by commit 50e4ed872f is identical to all the following:
* the file associated with this PR
* the file in master
* the current list at https://github.com/BOINC/boinc-site/blob/master/projects.inc (which BOINC downloaded on 26 September and 1 October on my two Macs and also on 20 September on my Windows 7 box.)

So I will go ahead and tag 7.14.1 and build a Mac installer.

RichardHaselgrove:
> I have confirmed ...

Thanks Charlie - that saves me a tough job (though I might still build that database at leisure to facilitate future tests).

I'll still make the PR for updating web urls to https where viable, but let's not regard it as a blocker for v7.14

RichardHaselgrove:
I have opened https://github.com/BOINC/boinc-site/pull/3 for these changes. I find I don't currently have write access to the boinc-site repo, so these changes are derived from a personal fork.

davidpanderson:
Why hasn't this been merged?

PR #2723, opened 2018-09-27
Drupal: Add BOINC username blacklist

drshawnkwang:
User may not change or select a BOINC displayname/username on a black list set by the admin.

Blacklist of names has defaults set in boinc_standard Feature.

https://dev.gridrepublic.org/browse/DBOINCP-462


drshawnkwang:
Marked as WIP as I realized afterwards there is a problem that needs fixing.

drshawnkwang:
Removed WIP: Added sensible defaults to blacklist.

PR #2727, opened 2018-09-28
Web: Eliminate PHP warning

TheAspens:
While testing and allowing PHP to report warnings, attempts to create a venue resulted in the following being shown on the page:

Warning: Creating default object from empty value in /root/project/html/inc/prefs_project.inc on line 395

This is because the variable $new_prefs was not defined before being passed into prefs_project_parse_form and the application was relying on implicit object creation.

This change initializes $new_prefs before passing it in so that the warning is not triggered.

### DIFF
--- a/html/user/add_venue.php
+++ b/html/user/add_venue.php
@@ -35,6 +35,8 @@ if ($action) {
     if ($subset == "global") {
         $prefs = prefs_parse_global($user->global_prefs);
         $prefs->$venue = $prefs;
+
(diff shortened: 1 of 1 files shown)


RichardHaselgrove:
This error message has been reported many times over the years - e.g.

https://www.gpugrid.net/forum_thread.php?id=4817&nowrap=true#50228

Tracking it down once and for all would be greatly appreciated.

PR #2733, opened 2018-10-03
web: use HTTPS for all pages if it's available

davidpanderson:


html/inc/util.inc
```diff
@@ -1010,16 +1010,20 @@ function do_download($path,$name="") {
...
+// if have SSL but not using it, redirect
+//
+function redirect_to_secure_url() {
     if (defined('SECURE_URL_BASE')
         && strstr(SECURE_URL_BASE, "https://")
-        && (!isset($_SERVER['HTTPS']) || !$_SERVER['HTTPS'])
+        && empty($_SERVER['HTTPS'])
     ) {
-        Header("Location: ".SECURE_URL_BASE."/$url");
+        Header("Location: https://".$_SERVER['HTTP_HOST'].$_SERVER['REQUEST_URI']);
```
TheAspens: URL_BASE might have a value of http://www.computeathome.org and SECURE_URL_BASE might have a value of https://secure.computeathome.org.  If a project had done this and their certificate was only good for secure.computeathome.org then this code change would break the site.

Can we use 
Header("Location: ".SECURE_URL_BASE.$_SERVER['REQUEST_URI']);

instead?

davidpanderson:
That doesn't work if the URL contains a directory, e.g. boinc.berkeley.edu/test/

I doubt that any project would use a different domain name for https

TheAspens:
Item 1: A redirect of a POST is bad since it turns into a GET so we might want to only redirect on GETs.

Item 2: I was looking at the functions secure_url_base and url_base and their usage throughout the code.  

Right now the site will continue to construct and send out a lot of URL's based on URL_BASE which will cause additional redirects as requests come if the project has defined URL_BASE with http and they have a SECURE_URL_BASE.  Also, 

I was also thinking about the goal of this change to use https as much as possible.  My thought is that we should be deprecating the use of SECURE_URL_BASE and encouraging everyone to just put an https base into URL_BASE. 

secure_url_base and url_base should become the same function and it should do the following:

- if URL_BASE starts with https, then return URL_BASE
- If URL_BASE doesn't start with https but SECURE_URL_BASE starts with https, then return SECURE_URL_BASE

It should no longer care if the current request is HTTP or HTTPS - it should construct URLs with HTTPS if HTTPS is available.

The use of secure_url_base() or url_base() throughout the code base could then be reduced down to just the one function.

thoughts?

davidpanderson:
Agreed about using the secure URL for url_base().

The original suggestion had to do with outside links, e.g. to
http://boinc.berkeley.edu
The goals was to redirect these to https.
POST/GET is not an issue in this case.

PR #2734, opened 2018-10-04
Update GAHP version string

timtheisen:
The existing code depended on building within a git repository.  It
included the GIT_REVISION in the startup message.  Unfortunately,
if you built from a downloaded tarball, the build fails. I thought
that the version of BOINC that the GAHP was compiled against would
be useful information as it is logged.  The SVN_VERSION is still
available by using the --version command line option.  Because of
the small change in behavior, I also bumped the version number of
the GAHP to 1.0.2.

### DIFF
--- a/samples/condor/Makefile
+++ b/samples/condor/Makefile
@@ -7,7 +7,7 @@ distclean: clean
 
 distclean-recursive: clean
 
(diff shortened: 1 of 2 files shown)


timtheisen:
I think that this is my last update for the foreseeable future.

davidpanderson:
Thanks.

PR #2735, opened 2018-10-04
Mac screensaver: If no application graphics running, don't display bad message

CharlieFenton:
Under OS 10.13 and later, there was a bug in the Mac screensaver coordinator when the default (overview) screensaver _boincscr_ took longer than 5 seconds to open its window. This incorrectly triggered the message "Screensaver of application _XXX_ is not compatible with this version of OS X" where _XXX_ is normally the name of the worker application associated with the faulty graphics app. 

But instead of substituting one worker app name for _XXX_ in the displayed string, it appended the names of all inactive apps (all those with slot set to -1) up to the buffer size limit of 1024 characters.

Probably the worst problem created by this bug is that it caused the logic to decide that _boincscr_ was defective and stop trying to use it. 

Fortunately, none of these problems persisted once the screensaver coordinator was dismissed. The next time the screensaver coordinator was invoked _boincscr_ was again able to run (though the problem would recur if _boincscr_ again took too long to display its window.)

Also fortunately, the fix involved a very small change to just one file. However, I needed to add `#if  DEBUG_UNDER_XCODE` in a few more places to allow the debugging logic to work properly so that I could diagnose the problem, and I have included them in this PR.

### DIFF
--- a/clientscr/Mac_Saver_ModuleView.m
+++ b/clientscr/Mac_Saver_ModuleView.m
@@ -169,7 +169,7 @@ static bool mojave;
 #define MAXDELTA 16
 
 // On OS 10.13+, assume graphics app is not compatible if no MachO connection after 5 seconds
-#define MAXWAITFORCONNECTION 5.0
+#define MAXWAITFORCONNECTION 8.0
 #define MAX_CGWINDOWLIST_TRIES 3
 
 int signof(float x) {
--- a/clientscr/Mac_Saver_ModuleView.m
+++ b/clientscr/Mac_Saver_ModuleView.m
@@ -347,10 +347,12 @@ void launchedGfxApp(char * appPath, pid_t thePID, int slot) {
     [ super stopAnimation ];
 
     if ([ self isPreview ]) return;
+#if ! DEBUG_UNDER_XCODE
     NSRect windowFrame = [ [ self window ] frame ];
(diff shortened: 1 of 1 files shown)


CharlieFenton:
As an added precaution, I have extended the time to wait for graphics apps to launch.

CharlieFenton:
It would be nice to include these small changes before public release of 7.14, but it is probably best to wait before bumping the version number to 7.14.2 in case alpha testers find other issues which need addressing.

armstrdj:
This code change looks reasonable to me and I don't see any issues.

TheAspens:
Based on @armstrdj  review, I"m proceeding with the merge.  Thanks @CharlieFenton

TheAspens:
@brevilo - In the boinc-dev discussion list, did we decide that the person who does the merge should delete the branch or the person who opened the pull request should delete the branch?

AenBleidd:
I think this should be done by the person who created a PR because (e.g. in my case) the branch could be in different repository (fork)

TheAspens:
@CharlieFenton - I created #2736 to allow this to be merged into the 7.14 release branch when it is time for the 7.14.2 version.

TheAspens:
@AenBleidd - good point.  It should be left to the person who created the pull request then since they will always have access to the branch.  Thanks.

CharlieFenton:
> it is probably best to wait before bumping the version number to 7.14.2

Sorry, that will be 7.14.3

Issue #2688, opened 2018-09-10
Server Version


lfield:
A mechanism is required to publish the server version.

lfield:
Are there any suggestions on how to do this?

marius311:
If I remember right, at the workshop it was suggested to put it in the get_project_config.xml ? I'm in favor of that idea. 

One thing I'd request is that it be easy for boinc-server-docker to append something to the published version to include that the server is running in boinc-server-docker, in a way that the admin doesn't have to do anything.

lfield:
This is fine. We can also show it in the footer. The question is what is the source for this information? Where does the php pick it up from and how does it get there. Also does the drupal also need updating?

lfield:
Does this work with the zip download?

brevilo:
Erm, nope, but that should be discouraged anyway, right? I mean how are admins supposed to run controlled updates of such instances?

brevilo:
Also, may I ask why publishing the server version is wanted in the first place? Projects might not want to do that due to security (by obscurity) concerns.

lfield:
At least the Fedora build uses the zip file. Publishing the server version is useful. The security risks are limited. Most security is provided by the OS, the details of which are not published.

brevilo:
Not sure I can agree to that. Compromised BOINC server components (by their nature) endanger the sensitive BOINC database content, and that's way easier if you know what server version you are about to attack. You might even go as far as arguing that this in violation of Art. 32 GDPR. This is of course debatable, but I think we should therefore at least have that discussion. One way or another, this feature must be made optional (via **opt-in**) and thus nothing must ever depend on the version being available.

ChristianBeer:
With SVN the revision was put into a file at checkout which was accessible to apache and shown on the server status page. Something like this can be done with the server version (not the git sha1) and it is up to the project to enable this or not. This file could be part of the zip archive. For projects that use a local fork of the BOINC repo the SHA1 hash is more meaningful though. They could publish this SHA1 but since the local fork is private nobody but the project admins know what version is really running (assuming they put their local changes in the same repo).

Publicly showing the exact SHA1 that is running is not advisable as that makes it easy for attackers to see if this specific version has a vulnerability without interacting with the project.

I also want to stress that no other tool should rely on this version number and expect certain features or behaviors. There are better ways for service discovery these days.

lfield:
[...]

PR #2737, opened 2018-10-05
Added server version detection and publishing

lfield:
This PR modifies the make_project command so that it will detect the git commit, branch and tag if it is run from a git directory. The result is stored in release.inc. The get_project_config.php and server_status.php views have been updated to show the git commit and version (from the tag) if it is available. This address #2688

### DIFF
--- a/html/user/get_project_config.php
+++ b/html/user/get_project_config.php
@@ -18,8 +18,8 @@
 
 require_once("../inc/util.inc");
 require_once("../inc/xml.inc");
-if(file_exists('../inc/release.inc'))
-    include '../inc/release.inc';
+if(file_exists('../../release.inc'))
+    include '../../release.inc';
 
 BoincDb::get(true);
 xml_header();
--- a/html/user/get_project_config.php
+++ b/html/user/get_project_config.php
@@ -18,6 +18,8 @@
 
 require_once("../inc/util.inc");
 require_once("../inc/xml.inc");
+if(file_exists('../inc/release.inc'))
+    include '../inc/release.inc';
 
 BoincDb::get(true);
 xml_header();
@@ -62,6 +64,14 @@ echo "<project_config>
     <master_url>$master_url</master_url>
     <web_rpc_url_base>".secure_url_base()."</web_rpc_url_base>
 ";
+
+if ( isset($git_commit) ) {
+    echo "<git_commit>$git_commit</git_commit>\n";
+}
+
+if ( isset($server_version) ) {
+    echo "<server_version>$server_version</server_version>\n";
+}
 
 if (parse_config($config, "<account_manager>")) {
     echo "    <account_manager/>\n";
--- a/html/user/server_status.php
+++ b/html/user/server_status.php
@@ -30,8 +30,8 @@ require_once("../inc/cache.inc");
 require_once("../inc/util.inc");
 require_once("../inc/xml.inc");
 require_once("../inc/boinc_db.inc");
-if(file_exists('../inc/release.inc'))
-    include '../inc/release.inc';
+if(file_exists('../../release.inc'))
+    include '../../release.inc';
 
 if (!defined('STATUS_PAGE_TTL')) {
     define('STATUS_PAGE_TTL', 3600);
@@ -186,8 +186,11 @@ function show_status_html($x) {
     }
     end_table();
 
+    global $server_version;
+    global $git_commit;
+
     if ( isset($server_version) && isset($git_commit) ) {
-        echo "Server version: $server_version ($git_commit)\n";
+        echo "Server version: $server_version (<a href=https://github.com/BOINC/boinc/commit/$git_commit>$git_commit)</a>)<br>";
     }
 
     if ($j->db_revision) {
--- a/html/user/server_status.php
+++ b/html/user/server_status.php
@@ -186,9 +186,6 @@ function show_status_html($x) {
     }
     end_table();
 
-    global $server_version;
-    global $git_commit;
-
     if ( isset($server_version) && isset($git_commit) ) {
         echo "Server version: $server_version (<a href=https://github.com/BOINC/boinc/commit/$git_commit>$git_commit)</a>)<br>";
     }
--- a/html/user/server_status.php
+++ b/html/user/server_status.php
@@ -185,7 +185,9 @@ function show_status_html($x) {
         ";
     }
     end_table();
-
+    
+    global $server_version;
+    global $git_commit;
     if ( isset($server_version) && isset($git_commit) ) {
         echo "Server version: $server_version (<a href=https://github.com/BOINC/boinc/commit/$git_commit>$git_commit)</a>)<br>";
     }
--- a/html/user/server_status.php
+++ b/html/user/server_status.php
@@ -30,6 +30,8 @@ require_once("../inc/cache.inc");
 require_once("../inc/util.inc");
 require_once("../inc/xml.inc");
 require_once("../inc/boinc_db.inc");
+if(file_exists('../inc/release.inc'))
+    include '../inc/release.inc';
 
 if (!defined('STATUS_PAGE_TTL')) {
     define('STATUS_PAGE_TTL', 3600);
@@ -183,6 +185,11 @@ function show_status_html($x) {
         ";
     }
     end_table();
+
+    if ( isset($server_version) && isset($git_commit) ) {
+        echo "Server version: $server_version ($git_commit)\n";
+    }
+
     if ($j->db_revision) {
         echo tra("Database schema version: "), $j->db_revision;
     }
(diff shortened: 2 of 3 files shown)


AenBleidd:
@lfield, I see that PR is in [WIP] but just to let you know that one include file is missed.
The piece of log from CI:

```
Traceback (most recent call last):
  File "./make_project", line 234, in <module>
    project.install_project()
  File "/root/boinc/py/Boinc/setup_project.py", line 585, in install_project
    install_boinc_files(self.dir(), True, not self.web_only)
  File "/root/boinc/py/Boinc/setup_project.py", line 343, in install_boinc_files
    os.remove(srcdir('html/inc/release.inc'))
OSError: [Errno 2] No such file or directory: '../html/inc/release.inc'
```

lfield:
Not a missing file. Missing try/catch.

py/Boinc/setup_project.py
```diff
@@ -339,6 +340,31 @@ def dir(*dirs):
...
         os.system("rm -f "+dir('html/languages/translations/*'))
         install_glob(srcdir('html/languages/translations/*.po'), dir('html/languages/translations/'))
+        try:
+            os.remove(srcdir('html/inc/release.inc'))
+        except OSError:
+            pass
+        try:
+            s = subprocess.Popen(["git", "rev-parse", "HEAD"], stdout=subprocess.PIPE, stderr=open(os.devnull, 'w'))
+            commit = s.stdout.read()[:7]
+            s = subprocess.Popen(["git", "branch"], stdout=subprocess.PIPE, stderr=open(os.devnull, 'w'))
```
marius311: If you're in a detatched HEAD state (which will be the case for boinc-server-docker builds) this will give you something that looks like `(HEAD detached at 91ceaae7c8)` which is probably undesired. `git rev-parse --abbrev-ref HEAD` might be better, which will just return `HEAD` in that case instead. 

Actually though, what's the motivation for even storing the branch, given that we have the commit? I'd say we probably don't need the branch at all.

py/Boinc/setup_project.py
```diff
@@ -339,6 +340,31 @@ def dir(*dirs):
...
+            branch = s.stdout.read().split('*')[1].split('\n')[0].strip()
+            s = subprocess.Popen(["git", "tag", "-l", "--points-at", "HEAD"], stdout=subprocess.PIPE, stderr=open(os.devnull, 'w'))
```
marius311: Might be best to ensure you're looking only at "server_release" tags here, which I imagine is what you want. You could do `git describe --match "server_release/*" --tags --exact HEAD`. This will error if the current commit doesn't have such a tag, in which case you could leave it blank, or you might use the output of `git describe --match "server_release/*" --tags --abbrev=0  HEAD` and append a `-DEV` or something like that so you get a version which looks like `server_release/0.9-DEV`, which would indicate the most recent server_release tag is 0.9 but you're some number of commits ahead.

html/user/server_status.php
```diff
@@ -183,6 +185,11 @@ function show_status_html($x) {
         ";
     }
     end_table();
+
+    if ( isset($server_version) && isset($git_commit) ) {
+        echo "Server version: $server_version ($git_commit)\n";
```
marius311: Maybe have this a hyperlink to `https://github.com/BOINC/boinc/commit/$git_commit` ?

html/user/get_project_config.php
```diff
@@ -63,6 +65,14 @@ function show_platforms() {
     <web_rpc_url_base>".secure_url_base()."</web_rpc_url_base>
 ";
 
+if ( isset($git_commit) ) {
+    echo "<git_commit>$git_commit</git_commit>\n";
```
marius311: Maybe have this a hyperlink to https://github.com/BOINC/boinc/commit/$git_commit ?

lfield:
@davidpanderson 
> release.inc should probably go in the project root dir rather than html/inc
> (in case C programs want to access it)
ok
> Also, the "upgrade" script will need to change it too
The [upgrade script](https://github.com/BOINC/boinc/blob/master/tools/upgrade#L81) uses the same [install_boinc_files](https://github.com/BOINC/boinc/blob/master/py/Boinc/setup_project.py#L303) function.

py/Boinc/setup_project.py
lfield: I will remove the branch detection

lfield:
Please could someone check my use of global variables.

html/user/server_status.php
```diff
@@ -183,6 +185,14 @@ function show_status_html($x) {
         ";
     }
     end_table();
+
+    global $server_version;
+    global $git_commit;
```
ChristianBeer: Those two `global` statements are not needed since you already declared both variables global in `release.inc`.

py/Boinc/setup_project.py
```diff
@@ -339,6 +340,30 @@ def dir(*dirs):
...
+global $git_commit;
+global $server_version;
+$git_commit = {commit}
+$server_version = {version}
+
+\?>
+'''.format(commit=commit, version=version)
+            f = open(os.path.join(dest_dir, 'release.inc'), 'w')
+            f.write(content)
+            f.close()
```
ChristianBeer: Be aware that the user that executes the upgrade python script is not the same as the user the webserver runs as. If the permissions on this file are too restrictive the webserver can't read it.

html/user/server_status.php
```diff
@@ -183,6 +185,11 @@ function show_status_html($x) {
         ";
     }
     end_table();
+
+    if ( isset($server_version) && isset($git_commit) ) {
+        echo "Server version: $server_version (<a href=https://github.com/BOINC/boinc/commit/$git_commit>$git_commit)</a>)<br>";
+    }
+
```
TheAspens: The variables $server_version and $git_commit are not visible within the function.  I had to user

```
    global $server_version;
    global $git_commit;
    if ( isset($server_version) && isset($git_commit) ) {
        echo "Server version: $server_version (<a href=https://github.com/BOINC/boinc/commit/$git_commit>$git_commit)</a>)<br>";
    }
```
For it to render correctly.

py/Boinc/setup_project.py
```diff
@@ -339,6 +340,30 @@ def dir(*dirs):
...
+            s = subprocess.Popen(["git", "rev-parse", "HEAD"], stdout=subprocess.PIPE, stderr=open(os.devnull, 'w'))
+            commit = s.stdout.read()[:7]
+            s = subprocess.Popen(["git", "describe", "--match", "server_release/*", "--tags", "--exact", "HEAD"],
+                                 stdout=subprocess.PIPE, stderr=open(os.devnull, 'w'))
+            version = s.stdout.read().split("/")[-1].strip()
+            content = '''<?php
+global $git_commit;
+global $server_version;
+$git_commit = "{commit}";
+$server_version = "{version}";
```
TheAspens: I believe that this should be able to retrieve the 0.9 server release tag.  However, it is not found and the file generated looks like this:

root@boincserver:~/project# cat release.inc 
```
<?php
global $git_commit;
global $server_version;
$git_commit = "3d7281d";
$server_version = "0.9";
```

Also - shouldn't the file end with `?>` to be proper?

html/user/get_project_config.php
```diff
@@ -18,6 +18,8 @@
 
 require_once("../inc/util.inc");
 require_once("../inc/xml.inc");
+if(file_exists('../../release.inc'))
+    include '../../release.inc';
```
TheAspens: Do you test for file existence simply to reduce the warning message that may occur using the include?

Also - I think the practice used in BOINC is to go ahead and use include_once even though it isn't strictly necessary in this case.

lfield:
@TheAspens Please check again. Here it is working on LHC@home Dev

https://lhcathomedev.cern.ch/lhcathome-dev/server_status.php
https://lhcathomedev.cern.ch/lhcathome-dev/get_project_config.php

TheAspens:
Laurence,

I've tested via boinc-server-test as well as just cloning your repo, checking out your branch and building form there.  In both cases the server version is not found.  The produced release.inc is:

```
[knreed@oc7811586721 boinc]$ cat release.inc 
<?php
global $git_commit;
global $server_version;
$git_commit = "842f76b";
$server_version = "";

?>
```

If I go into the clone of your repo and run this command (which I believe is used to obtain the server version number):

```
[knreed@oc7811586721 boinc]$ git describe --match server_release/* --tags --exact HEAD
fatal: no tag exactly matches '842f76b26dddfffe4321efa130b337b936fb7a6a'
```

It doesn't appear to obtain the reference correctly.


Are you sure you don't want to simply use: `git describe --all --match server_release/* --tags`

That returns the most recent server_release tag in the current checked out base.  Or do you intentionally not want a release number found in the event that the current commit is not a tagged commit?

lfield:
It is intentional that the version is not given if the current commit is not a tagged commit.

TheAspens:
Ah - then I have been over testing this :-)

It is good as is.

PR #2738, opened 2018-10-05
Fix issues to build android release

TheAspens:
While trying to create an unsigned release apk for Android, the following issues were found.

Attempt to use system only permission:
```
 > Lint found errors in the project; aborting build.
  
  Fix the issues identified by lint, or add the following to your build script to proceed with errors:
  ...
  android {
      lintOptions {
          abortOnError false
      }
  }
  ...
  
  Errors found:
  
  /home/knreed/BOINC/android/BOINC/app/src/main/AndroidManifest.xml:41: Error: Permission is only granted to system apps [ProtectedPermissions]
      <uses-permission android:name="android.permission.READ_LOGS"/>
```

Version regex in gradle build file required a seperator character for the suffix in all cases.  Added a ? to indicate that it was not required.

### DIFF
--- a/android/BOINC/app/src/main/AndroidManifest.xml
+++ b/android/BOINC/app/src/main/AndroidManifest.xml
@@ -38,7 +38,6 @@
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-    <uses-permission android:name="android.permission.READ_LOGS" />
     <uses-permission android:name="android.permission.RESTART_PACKAGES" />
 
     <!-- Features required for Android TV, consoles, and set-top boxes like Nexus Player, OUYA,
--- a/android/BOINC/app/build.gradle
+++ b/android/BOINC/app/build.gradle
@@ -20,7 +20,7 @@ def buildVersionCode() {
 
 // Derive version name from release tag and add commit SHA1
 def buildVersionName() {
-    def pattern = /client_release\/\d+\.\d+\/(?<major>\d+)\.(?<minor>\d+)\.(?<revision>\d+)(?<suffix>[-_\.].*)/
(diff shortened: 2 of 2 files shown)


android/BOINC/app/src/main/AndroidManifest.xml
```diff
@@ -38,7 +38,6 @@
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-    <uses-permission android:name="android.permission.READ_LOGS" />
```
AenBleidd: I'm not sure but does this mean that there will be no possibility to show logs for the user?

TheAspens: This permission only pertains to the app reading the system logs.  The logs generated by the app should be fine.  On the debug build I'm running on my phone I have no issues seeing the boinc log messages.

TheAspens: Here is the reference https://developer.android.com/reference/android/Manifest.permission.html#READ_LOGS

Nice and vague....

AenBleidd: Thank you! :thumbsup:

TheAspens: Better answer: https://www.youtube.com/watch?feature=player_embedded&v=WDDgoxvQsrQ#t=1369s

When Jelly Bean was released, all apps get access to their logs without requiring additional permissions.  Additionally, apps are no longer allowed to ask for or use the READ_LOGS permission.

Since we only support API 19 and higher, this permission should be able to be safely removed.

brevilo:
Can you please share two different examples of patterns that the regexp should match?

Thanks

TheAspens:
> Can you please share two different examples of patterns that the regexp should match?

For valid release:
client_release/7.14/7.14.1
client_release/7.8/7.8.26

For situation where the gradle file will block:
client_release/7.14/7.14.1.dirty

PR #2740, opened 2018-10-07
Mac installer: Changes to avoid Mojave security dialog

CharlieFenton:
OS 10.14 Mojave has added some new security safeguards. As a result, Macs who have multiple user accounts will see a scary warning when each user other than the one running the installer logs in, and asks for permission to proceed. As a temporary measure, I added the following to the installer's ReadMe file for generic (unbranded) BOINC:

> New for Mojave: If your Mac has more than one user account, each user may see the following dialog upon their first login after you have installed BOINC: "sh" wants to control "System Events...." The BOINC installer needs this access to finish adding or removing its login item for each user; please click "OK".

With the changes in this PR I am trying to either avoid that warning or at least make it easier to understand.

The reason for this problem is that the installer needs to adjust the Login Items for each user. (The Login Item causes BOINC Manager to be launched at login.)

For users authorized to manage BOINC, the installer deletes any old Login Items for branded / unbranded Managers other than the branded / unbranded one being installed, and adds the Login Item for the one being installed. 

For users NOT authorized to manage BOINC, the installer deletes any old Login Items for all branded / unbranded Managers.

The method the installer had been using to do this prior to OS 10.13 High Sierra no longer worked as of OS 10.13. My fix was for the installer to add a LaunchAgent in each user's account which runs the next time the user logs in. It runs a shell script to update the Login Items (which requires the use of the SystemEvents app) and then the LaunchAgent deletes itself. Since the app executing the shell script is _sh_, that is the application named in the warning dialog:
> "sh" wants to control "System Events....

Since the warning dialog says _sh_ rather than _BOINC_, it is not obviously associated with the BOINC installer, this PR adds a small CLI app named something like "BOINC_Finish_Install" to do be run by the LaunchAgent, in the hope that the warning dialog will then say something like 

> BOINC_Finish_Install" wants to control "System Events...."

substituting the branded name for _BOINC_ in the case of a branded Manager.

My early testing found that this approach seems to completely eliminate the warning dialog. If that holds true, then this is an even better solution than I had expected.

In addition to the BOINC installer, I will need make the corresponding changes to the _uninstaller_ and the _AddRemoveUser_ utility (all for Macintosh only.)

### DIFF
--- a/mac_build/boinc.xcodeproj/project.pbxproj
+++ b/mac_build/boinc.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 44;
+	objectVersion = 45;
 	objects = {
 
 /* Begin PBXAggregateTarget section */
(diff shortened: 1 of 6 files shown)


CharlieFenton:
I just realized that this also fixes a potential security issue. The security alert now says:
> “sh“ wants access to control “System Events“. Allowing control will provide access to documents and data in “System Events“, and to perform actions within that app.

When the user answers _OK_ to that alert (as they must to allow the BOINC installer to set up their Login Items), I believe that permanently gives the shell interpreter application _sh_ permission to send commands to the _System Events_ utility. While it is unlikely that could ever cause any real problems, it seems imprudent since that would give all shell scripts access. As far as I know, there is no easy way for a user to revoke this permission once granted.

After implementation of this PR, any such permission will be given only to the _BOINC_Finish_Install_ app, whose capabilities are very limited. It can only add or remove Login Items for the _BOINC Manager_ and launch or kill the _BOINC Manager_.

CharlieFenton:
I have finished my testing of this PR and it is ready to merge.

davidpanderson:
Thanks!

PR #2748, opened 2018-10-09
run_program() fixes

JuhaSointusalo:
lib: change `run_program()` to use `file` parameter as program name on Windows
mgr,scr: fix `argv`s in calls to `run_program()`
client: remove quoting client path in launching GPU detection

See commit messages for details.

Fixes launching graphics apps with Manager and screensaver as reported on boinc_alpha.

### DIFF
--- a/client/gpu_detect.cpp
+++ b/client/gpu_detect.cpp
@@ -598,7 +598,6 @@ int COPROCS::launch_child_process_to_detect_gpus() {
 #else
     int prog;
 #endif
-    char quoted_client_path[MAXPATHLEN];
     char quoted_data_dir[MAXPATHLEN+2];
     char data_dir[MAXPATHLEN];
     int retval = 0;
@@ -625,22 +624,17 @@ int COPROCS::launch_child_process_to_detect_gpus() {
     boinc_getcwd(data_dir);
 
 #ifdef _WIN32
-    strlcpy(quoted_client_path, "\"", sizeof(quoted_client_path));
-    strlcat(quoted_client_path, client_path, sizeof(quoted_client_path));
-    strlcat(quoted_client_path, "\"", sizeof(quoted_client_path));
-
     strlcpy(quoted_data_dir, "\"", sizeof(quoted_data_dir));
     strlcat(quoted_data_dir, data_dir, sizeof(quoted_data_dir));
     strlcat(quoted_data_dir, "\"", sizeof(quoted_data_dir));
 #else
-    strlcpy(quoted_client_path, client_path, sizeof(quoted_client_path));
     strlcpy(quoted_data_dir, data_dir, sizeof(quoted_data_dir));
 #endif
 
     if (log_flags.coproc_debug) {
         msg_printf(0, MSG_INFO,
             "[coproc] launching child process at %s",
-            quoted_client_path
+            client_path
         );
         if (!is_path_absolute(client_path)) {
             msg_printf(0, MSG_INFO,
@@ -656,7 +650,7 @@ int COPROCS::launch_child_process_to_detect_gpus() {
             
     int argc = 4;
     char* const argv[5] = {
-         const_cast<char *>(quoted_client_path),
+         const_cast<char *>(client_path),
          const_cast<char *>("--detect_gpus"), 
          const_cast<char *>("--dir"), 
          const_cast<char *>(quoted_data_dir),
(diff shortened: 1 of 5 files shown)


AenBleidd:
This PR fixes #2657
@JuhaSointusalo, correct me if I wrong

RichardHaselgrove:
@AenBleidd 

> This PR fixes #2657

I think that's a little premature. @JuhaSointusalo specifically referenced a quite separate conversation on boinc_alpha about screensavers and graphics apps. Let's leave #2657 open again (good idea to re-open it) until we're all sure we're all on the same page.

JuhaSointusalo:
> This PR fixes #2657

More like #2713.

AenBleidd:
@JuhaSointusalo, Is this ready for review?
If yes - @davidpanderson, @TheAspens, could you please assist?

JuhaSointusalo:
> Is this ready for review?
>
Yes.

AenBleidd:
@davidpanderson, @TheAspens, could you please review this?

AenBleidd:
Just a kind reminder, @davidpanderson, @TheAspens

CharlieFenton:
If I am remembering correctly, this PR should **not** be merged but should just be closed.  I believe this issue was fixed of by PR #2713 which was merged on Sep 26, 2018. Likewise, I believe PR #2657 should be closed without merging.

AenBleidd:
@CharlieFenton, could you please double check? I believe this particular PR fixes the issue presented in #2713 and have better implementation than #2657. Thanks

TheAspens:
I have removed myself as a reviewer as this is an area that I do have expertise in.  However, should we have a meeting to discuss the differences between this pull request and #2742 ?

TheAspens:
If we should have a meeting on this, and all the related pull requests, who all needs to be there?  @davidpanderson , @JuhaSointusalo as the key developers - who else?

RichardHaselgrove:
I have been involved with this issue since it first appeared here as #2029. I did record in #2713 that I had tested a Windows build from that code, and that it successfully addressed all the issues I was aware of.

I'm happy to perform similar tests on the alternative formulation, or to join a meeting, if required, but for the time being I think we're OK

CharlieFenton:
@AenBleidd Sorry, I now did a quick review of the commits in #2713 and see that they are independent of the changes in this PR. So I withdraw my objection. However, I have neither the time nor the inclination to wend my way through this long convoluted string of PRs. I believe that we agree that  PR #2657 should be closed without merging; as far as I can see there were no commits in that PR.

I defer to @JuhaSointusaloand others to decide what should be done.

davidpanderson:
What problem does this PR fix, and how?

JuhaSointusalo:
Quick recap. There was a problem with GPU detection when client was started without using path to it but searched from `%PATH%` (plus some other conditions that don't matter to this discussion). For GPU detection another copy of client is started with `run_program()`. Here's what the function signature was originally:

`int run_program(const char* dir, const char* file, int argc, char *const argv[], double nsecs, HANDLE& id)`

`argv[]` is Unix `exec*()`style list of arguments where `argv[0]` is program name. As an implementation detail, the function didn't actually require `argv[0]` to be anything useful and instead used `file` for launching the program.

In #2713 I changed the function to use `argv[0]` as program name and ignore `file` on Windows. This was done so that the function could search `%PATH%` for the program. I carefully tested the code and found that it works correctly for GPU detection. What I neglected to was test the entire BOINC codebase.

Turns out the code that launches graphics apps and screensavers exploited the previously mentioned implementation detail and passed in rubbish `argv[0]`. That with my changes broke launching graphics apps and screensavers. My bad for not checking the other users of `run_program()`.

That's the history part. On to this PR.

Passing in rubbish parameters just because you can get away with it today is plain and simply bad programming so this PR fixes the other users of `run_program()`. This PR also changes `run_program()` to use `file` instead of `argv[0]` as program name on Windows. POSIX says `argv[0]` should be the program name only without path so any code that actually does things according to POSIX wouldn't have worked with the changes I originally did. The last change is to not quote path in the code that launches GPU detection. How `run_program()` needs to quote paths is again an implementation detail that the function's callers should not need to know or care about.

Since there was a request to compare this PR and #2742. #2742 was opened because David didn't understand the changes I did, didn't read the commit messages, didn't bother ask me to explain the changes and instead though it would be a good idea to revert the changes I had intentionally done. This PR lets `run_program()` search `%PATH%`, #2742 doesn't.

PR #2745, opened 2018-10-08
NeedRunDaemon option should be applied when connecting to localhost o…

AenBleidd:
…nly.

This commit fixes a situation when trying to connect to remote client with option 'Run client' on leads to unability to connect to selected client.

Signed-off-by: Vitalii Koshura <lestat.de.lionkur@gmail.com>

PR #2749, opened 2018-10-09
[Manager] Fix an issue when 127.0.0.1 was not recognized as local add…

### LINKED ISSUE #2746: 127.0.0.1 doesn't recognized as local address
In the function below 127.0.0.1 doesn't recognized as local address.

https://github.com/BOINC/boinc/blob/650127da844e38ace599001d6225cdbf8efb4da3/clientgui/MainDocument.cpp#L255-L278

Should this function be fixed in the corresponding way?

AenBleidd:
…ress

This issue fixes #2746.
It adds an option to recognize '127.0.0.1' as local address.Also code was reorganized to get rid off multiple strMachine.Lower() function calls.Signed-off-by: Vitalii Koshura <lestat.de.lionkur@gmail.com>

### DIFF
--- a/clientgui/MainDocument.cpp
+++ b/clientgui/MainDocument.cpp
@@ -265,15 +265,26 @@ bool CNetworkConnection::IsComputerNameLocal(const wxString& strMachine) {
 
     if (strMachine.empty()) {
         return true;
-    } else if (wxT("localhost") == strMachine.Lower()) {
-        return true;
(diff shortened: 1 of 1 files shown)


RichardHaselgrove:
Added this commit to the source I used for the tests on #2745 on Monday. All worked OK - including the problematic `boincmgr /n 127.0.0.1 /nd`. I won't do a formal review / approve, because I can't test the other OSs (a Linux user should check it out too), but seems good for Windows.

PR #2758, opened 2018-10-12
Manager, simple view: if project name not known yet, show URL

davidpanderson:


TheAspens:
Looks fine.  Thanks for the fix.

PR #2760, opened 2018-10-12
Web: let projects customize their certificate

davidpanderson:
define CERT_FILENAME in your project.inc to whatever you want;
default is "cert1.php"


TheAspens:
works fine - thanks!

PR #2712, opened 2018-09-22
Travis: copy successful builds to bintray

ChristianBeer:
Copy a weekly build from master to https://bintray.com/boinc/boinc-ci so it is available for testers. Only specified build artefacts are copied (see deploy/prepare_deploy.sh). The copy operation is triggered via a Travis CI cronjob or a pull request from within the BOINC/boinc gitub repository.
Pull requests from forks are ignored because they can not have access to the bintray API key, this is a security restriction enforced by Travis CI.

PR #2763, opened 2018-10-14
fix bintray deploy

ChristianBeer:
- provide sensible defaults for AppVeyor deploy to bintray; Until there is a better mechanism to control deployments on AppVeyor a deployment will happen for every merge into master. This will prevent it from failing.
- stop Travis deploy script when unsupported event happens
- fix intendation

Follow up hotfix for #2712

### DIFF
--- a/deploy/prepare_deployment.bat
+++ b/deploy/prepare_deployment.bat
@@ -4,6 +4,11 @@ for /f "tokens=2-4 delims=/ "  %%a in ("%date%") do (set MM=%%a& set DD=%%b& set
 set build_date=%YYYY%-%MM%-%DD%
 set git_rev=%APPVEYOR_REPO_COMMIT:~0,8%
 
+rem Default values because %bintray_deploy% is currently unused
+set pkg_name=master
+set git_rev=%APPVEYOR_REPO_COMMIT:~0,8%
+set pkg_version=master_%build_date%_%git_rev%
+
 if defined APPVEYOR_PULL_REQUEST_NUMBER (
     set pkg_name=pull-requests
     set git_rev=%APPVEYOR_PULL_REQUEST_HEAD_COMMIT:~0,8%
--- a/deploy/deploy_to_bintray.sh
+++ b/deploy/deploy_to_bintray.sh
@@ -86,6 +86,7 @@ if [[ $CI_RUN == "true" ]]; then
         ;;
         *)
         echo "event $TRAVIS_EVENT_TYPE not supported for deployment"
+        exit 0;
         ;;
     esac
 fi
--- a/deploy/deploy_to_bintray.sh
+++ b/deploy/deploy_to_bintray.sh
@@ -34,9 +34,9 @@ fi
 
 ROOTDIR=$(pwd)
 if [[ $# != 1 ]]; then
-  echo "Usage: $0 SOURCE_DIR"
-  echo "SOURCE_DIR : relative path where binaries are, last component is used as BOINC_TYPE"
-  exit 1
+    echo "Usage: $0 SOURCE_DIR"
+    echo "SOURCE_DIR : relative path where binaries are, last component is used as BOINC_TYPE"
+    exit 1
 fi
 
 SOURCE_DIR="$1"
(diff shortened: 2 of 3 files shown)


deploy/prepare_deployment.bat
```diff
@@ -4,6 +4,11 @@ for /f "tokens=2-4 delims=/ "  %%a in ("%date%") do (set MM=%%a& set DD=%%b& set
 set build_date=%YYYY%-%MM%-%DD%
 set git_rev=%APPVEYOR_REPO_COMMIT:~0,8%
 
+rem Default values because %bintray_deploy% is currently unused
+set pkg_name=master
+set git_rev=%APPVEYOR_REPO_COMMIT:~0,8%
```
ChristianBeer: This is declared above already but since we have to change this section again when introducing `deploy_to_bintray.bat` this shouldn't be a show stopper.

AenBleidd: Isn't this a duplicate string thes same as 5?

ChristianBeer: Yes, and I realized it at the same time as you. See my comment about that.

AenBleidd:
Thanks for a quick fix

RichardHaselgrove:
Well done to everyone. I have downloaded and just started running the c9bacd5a.7z 'artefacts'.

One gotcha to remember when the time comes to write up the process for testers. AppVeyor is making VS2013 debug builds: a standard user / first time tester will be running VS2010 release builds. The DLLs are incompatible.

BOINC Manager requires msvcp120d.dll and msvcr120d.dll
The BOINC client requires those two, plus additionally libcurld.dll and zlib1d.dll

The debug DLLs are hard for end users to come by: I had them handy because of my own previous VS2013 building and testing. It would be helpful to place the debug DLLs in a handy download location linked from the testing instructions.

I also wonder how useful the debug builds will be without the associated .pdb symbol files - though they might add considerably to the download/storage sizes.

RichardHaselgrove:
Both client and Manager are running normally, though the first observation is that the artefact Manager is much slower to update the display. I think this is related to the 'debug' nature of the build: replacing this Manager with a recent VS2013 local release build restores normal update speed, but reveals 
 a possible unrelated problem. I'll investigate and report separately.

Mentioning the slower display update in the testing documentation might be wise.

AenBleidd:
I think, we can build Release binaries for Windows on weekly basis with scheduled builds. It is already implemented as a part of binded PR merged earlier but now it builds Degud version only. I plan to fix this soon.
Test only issue I found no way to build automatically VS2010 version. As far as I see now BOINC couldn't be built using msbuild because of Post-build script. And it couldn't be built either using VS command line because it looks like that it opens the Studio GUI iyselg and just wait for user interaction.

ChristianBeer:
This kind of didn't work as expected. The merge commit in master was uploaded as a weekly scheduled build. For now this is not a big problem. I'll try to keep an eye on this.

RichardHaselgrove:
The msvcp*.dll files are supplied by Microsoft when you install VS2013

When you run a build, the appropriate version (debug or release) is copied to the build output folder, alongside the compiled binaries. The pdb files are in the same folder.

JuhaSointusalo:
IIRC, MS CRT's license forbids redistributing the debug version.

RichardHaselgrove:
My system doesn't appear to have a license file (apart from some extension components), and Microsoft have removed most VS2013 documentation - the choice now is between VS2015 and VS2017.

Reading https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/ms235299(v=vs.120) and the Visual C++ Runtime Files section of https://visualstudio.microsoft.com/license-terms/distributable-code-microsoft-visual-studio-2013-microsoft-visual-studio-2013-sdk-includes-u-2/ I'd agree that the debug files are excluded by omission.

The best set of answers (including workrounds) I could find is https://stackoverflow.com/questions/21707992/msvcp120d-dll-missing

AenBleidd:
> I think it would be easiest to change AppVeyor to build Release version

Will be done soon

AenBleidd:
#2775 will switch AppVeyor from Debug to Release build

Issue #694, opened 2015-02-04
db_dump is inefficient


romw:
**Reported by Nicolas on 10 Dec 38606274 05:20 UTC**
Let's compare.
- Write a potentially big uncompressed XML file (over 2GB for SETI@Home), then use  system() to run `gzip` on it (which reads it '''again''', writes something like 300MB, then deletes the original that took so long to write). Total I/O: 2GB read, 2.3GB write, plus the database reads (which are quite a lot too).
- Use zlib to write the file in compressed form in the first place. I/O: 300MB write, plus database reads.

Which would you choose?

I'll see if I can modify db_dump myself to use zlib, and post the patch. Otherwise, anyone is free to take the task.

Migrated-From: http://boinc.berkeley.edu/trac/ticket/717

romw:
**Commented by Nicolas on 3 Jul 38606279 21:20 UTC**
In case it isn't clear: the first is what db_dump currently does :)

ChristianBeer:
For reference: the db_purge tool already does 'in place' compression using pipes: [sched/db_purge.cpp](https://github.com/BOINC/boinc/blob/master/sched/db_purge.cpp)

denravonska:
I added a proposal for this in #2767. Ideally, it should handle zip the same way but I was unsure if BOINC  had an existing libzip dependency.

AenBleidd:
Fixed in #2767

PR #2767, opened 2018-10-17
db_dump: Add support for dumping straight to gzip.

denravonska:
A refactor of db_dump which splits the compression logic out of ZFILE and into separate output stream classes. This opens up for on-the-fly gzip compression for #694.

I have tested this in the VM image but it could be good if someone with a larger database tries an export to see if there are any concrete time gains.

### DIFF
--- a/sched/db_dump.cpp
+++ b/sched/db_dump.cpp
@@ -26,6 +26,7 @@
 //    I'll get to this someday.
 
 #include "config.h"
+#include <zlib.h>
 #include <cstdio>
 #include <cstring>
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


JuhaSointusalo:
I thought you were redoing earlier changes but no, it was actually `db_purge` the last time. Both deal with compressed files so there could some code sharing opportunities for some future enhancement. Your approach with classes look nicer than `db_purge`s `void` pointers.

sched/db_dump.cpp
```diff
@@ -209,32 +210,172 @@ int DUMP_SPEC::parse(FILE* in) {
...
+        if(!is_open())
+            return false;
+
+        gzclose(gz);
+        gz = 0;
+        return true;
+    }
+
+    void write(const char* fmt, va_list args) {
+        gzvprintf(gz, fmt, args);
```
JuhaSointusalo: `gzvprintf()` is relatively new addition. RHEL 6 users are not going to like you using it and I don't think RHEL 7 users either. Another problem is that it uses 8 kB buffer and at least team descriptions can be larger than that.

denravonska: Good point. I considered the small buffer but discarded it as an issue as I only saw small writes. I'll have another look at it to see what the alternatives are.

denravonska: How about using `vasprintf` and then dumping the entire buffer to libz? It's going to be a bit slower but you are not bound to a buffer size.

Out of curiosity I double checked printf and its min buffer size is apparently 4095 character.

sched/db_dump.cpp
```diff
@@ -209,32 +210,158 @@ int DUMP_SPEC::parse(FILE* in) {
     return ERR_XML_PARSE;
 }
 
+//
+// File streams
+//
+class OutputStream {
```
denravonska: Should the class names be uppercase?

TheAspens: The coding standards are here: https://boinc.berkeley.edu/trac/wiki/CodingStyle  which I think stats that class names should be upper case.  @davidpanderson can confirm if needed.

denravonska: @TheAspens  Thank you! I'll check it out and tweak the classes.

JuhaSointusalo:
I started thinking. This code with classes and all is good. But...

Just before GDPR came into effect I downloaded host stats from all the projects BOINCstats knows of. Every single project had gzipped files. `db_dump´ could just as well always unconditionally output gzipped files and nobody would notice.

So... I wonder if those other output options should be removed to clean up the code a bit and hard code `db_dump` to always gzip files? The [note at the top of `db_dump.cpp`](https://github.com/BOINC/boinc/blob/master/sched/db_dump.cpp#L21) kind of hints at that direction.

denravonska:
@JuhaSointusalo I had the same thought when I did this. I assumed the flexibility was needed but I have yet to see an export which was not gzip besides the `tables.xml` in the VirtualBox sample.

TheAspens:
For WCG we have always used the gzip option - so we wouldn't object to that change either.

TheAspens:
@denravonska - if you decide you are going to remove those options (that are apparently unused) then make add WIP to the pull request.  I will try to review this either over the weekend or monday - but I'll leave it alone if i see the WIP.

TheAspens:
But one thought is - since this change is complete (except for the review of coding standards) maybe leave this pull request as is and make the additional changes to remove the unused options in another pull request.

TheAspens:
(sorry for multiple messages) - either way - if you could just post in this thread when you feel you are ready for review, that would let me know for sure.  Thanks!

denravonska:
Changed the class names to conform to the standard. Any help testing would be appreciated as I'm still trying to find the password to the database in the virtual machine :)

JuhaSointusalo:
I'll leave it for @TheAspens to test this on a real project database.

TheAspens:
I tried this three times against a database with a million host rows and a bunch of user rows as well and it ran in roughly the same time.  However, since the files were small enough that they would be cached in memory it may not have caught the issue Rom outlined in the associated issue.

It produced identical files as the version in master.  

Sorry for the delay in getting this tested.  However, it looks good and I'm going to merge it.  Thanks for the contribution!

denravonska:
Great! Out of interest, what was the rough size of the gzip?

TheAspens:
The host records I created was 105M compressed and 732M uncompressed.

denravonska:
That is a pretty large data set and I would expect some speedup. Unless the host is very beefy.

PR #2778, opened 2018-10-23
[AppVeyor] Switch default build to Release x64

AenBleidd:
Add cache dependency to invalidate it properly

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -53,8 +53,8 @@ skip_commits:
     - xcompile/
 
 configuration:
-  - Debug
-#  - Release
+#  - Debug
+  - Release
 
 environment:
   matrix:
@@ -64,7 +64,7 @@ environment:
       depends_path: C:\projects\boinc_depends_win_vs2013
 
 cache:
-  - C:\projects\boinc_depends_win_vs2013
+  - C:\projects\boinc_depends_win_vs2013 -> appveyor.yml
 
 before_build:
   - call win_build\load_dependencies.bat %depends_git_path% %depends_path% %platform% %configuration%
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -64,7 +64,7 @@ environment:
       depends_path: C:\projects\boinc_depends_win_vs2013
 
 cache:
-  - C:\projects\boinc_depends_win_vs2013 -> appveyor.yml
+  - C:\projects\boinc_depends_win_vs2013 -> appveyor.yml, win_build\load_dependencies.bat
 
 before_build:
   - call win_build\load_dependencies.bat %depends_git_path% %depends_path% %platform% %configuration%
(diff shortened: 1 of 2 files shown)


AenBleidd:
Android Manager build failed because the file support-vector-drawable-26.0.2.jar is not present in https://jcenter.bintray.com/com/android/support/support-vector-drawable/26.0.2/ anymore.
I hope it is a temporary issue but in any case this PR is not related to it

appveyor.yml
```diff
@@ -64,7 +64,7 @@ environment:
       depends_path: C:\projects\boinc_depends_win_vs2013
 
 cache:
-  - C:\projects\boinc_depends_win_vs2013
+  - C:\projects\boinc_depends_win_vs2013 -> appveyor.yml
```
ChristianBeer: Could you add `win_build\load_dependencies.bat` as a dependency too?

ChristianBeer:
@RichardHaselgrove Artifacts are here: https://ci.appveyor.com/project/BOINC/boinc/builds/19737699/artifacts Could you please test if they are faster than the debug builds? You might need the release msvcp120.dll's from your local build for now since they are not included in the artifacts right now.

RichardHaselgrove:
@ChristianBeer - yes, the artifact version you linked runs at normal speeds (slow display update in the manager has gone).

I cleaned out my BOINC program folder and re-installed v7.14.2, and then dropped the artifact executables over the top. Ran first time, nothing else needed. BUT: I'd followed my own advice from https://boinc.berkeley.edu/trac/wiki/CompileClient#Windows years ago, so the msvc stuff was installed locally as a shared resource (in system32) and available to all. That's the proper way to do it.

AenBleidd:
@RichardHaselgrove, if you're ok with the results, could you please merge this?

RichardHaselgrove:
Being still a very junior contributor, I'm cautious about getting above my station. Three observations:

1) the code changes seem well-sandboxed into the intended area, and the effect is good for testing. That's a plus.
2) I haven't seen any discussion of whether debug builds were useful, and hence whether we'd be losing anything.
3) The Travis failure on Android is completely unrelated to this, but it still makes me nervous.

So I give notice that I will merge 24 hours from now, unless someone more experienced jumps me, or a significant objection is raised on my point (2).

AenBleidd:
About point [2]: this behaviour could be changed and there is a ticket for this. In the near future I plan to upgrade build scripts not to use git located on Berkeley website because even to fetch it is a hard operation. 
About point [3]: necessary package is missed here: https://bintray.com/bintray/jcenter/com.android.support%3Asupport-vector-drawable
I've already send a message to the owner of this repo to update the build. Currently got no answer

ChristianBeer:
I was just waiting for Richards confirmation about the timing and then live happened. We can easily switch back to debug builds if needed in the future.

PR #2780, opened 2018-10-24
Sandboxing tools installation fixes

EvaSDK:
While comparing packaging in Debian and Gentoo, I came across the `switcher` tool. Haven't really found it if it is ever used but the following modifications would make it install in a FHS compatible location. Also switcher makes calls to `setprojectgrp` which wasn't installed at all so I added some changes for that as well as adding missing config.h include to it.

### DIFF
--- a/client/Makefile.am
+++ b/client/Makefile.am
@@ -24,7 +24,7 @@ endif
 
 bin_PROGRAMS = boinc_client boinccmd
 if !OS_WIN32
-bin_PROGRAMS += switcher
+pkglibexec_PROGRAMS = switcher
 endif
 
 boinccmd_SOURCES = boinc_cmd.cpp
@@ -106,7 +106,17 @@ else
 if OS_DARWIN
 boinc_client_LDFLAGS += -Wl,-flat_namespace,-undefined,dynamic_lookup
 else
-boinc_client_SOURCES += hostinfo_unix.cpp
+boinc_client_SOURCES += \
+	hostinfo_unix.cpp \
+	file_names_unix.h
+
+edit = $(SED) -e 's|@pkglibexecdir[@]|$(pkglibexecdir)|g'
+
+file_names_unix.h: file_names_unix.h.in Makefile
+	$(AM_V_GEN)$(edit) $@.in > $@
+
+BUILT_SOURCES = file_names_unix.h
+CLEANFILES = file_names_unix.h
 endif
 endif
 boinc_client_LDADD = $(LIBBOINC) $(LIBBOINC_CRYPT) $(BOINC_EXTRA_LIBS) $(PTHREAD_LIBS)
--- a/client/Makefile.am
+++ b/client/Makefile.am
@@ -24,7 +24,7 @@ endif
 
 bin_PROGRAMS = boinc_client boinccmd
 if !OS_WIN32
-pkglibexec_PROGRAMS = switcher
+pkglibexec_PROGRAMS = switcher setprojectgrp
 endif
 
 boinccmd_SOURCES = boinc_cmd.cpp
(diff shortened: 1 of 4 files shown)


AenBleidd:
@EvaSDK
Renamed file `file_names.h` breaks both Windows and MacOS build.
Please fix this or use another approach without file rename

EvaSDK:
Unfortunately, I don't have the tools to edit Windows & Xcode projects.

EvaSDK:
I could make this build-time switchable for unix platforms so MacOS and Windows builds stick to current codepath.

AenBleidd:
> I could make this build-time switchable for unix platforms so MacOS and Windows builds stick to current codepath.

Please do this. BOINC is cross-platform project so it's codebase should be valid to be built on every supported platform.

Thank you

EvaSDK:
Should be fine now.

CharlieFenton:
The sandboxing that uses switcher and setprojectgrp are used **only** for Macintosh. Since building for Mac using config / make has not been supported for several years, switcher and setprojectgrp should not be included in any of the configure or make tools. It is **essential** that they be installed the BOINC Data directory inside the switcher subdirectory, with certain permissions.

Please see the comments at the top of clientgui/makefile.am, which says:
```
## To build for Macintosh (Darwin), please read the instructions at 
## mac_build/HowToBuildBOINC_XCode.rtf
```
In addition, the Macintosh build uses the custom _config.h_ located at _clientgui/mac/config.h_.

Do **NOT** make any changes to anything related to these files; you are likely to break the security on Macintosh. However, feel free to remove all mentions of _darwin_ in the __autosetup_, _config_ and _make_ files. That should have been done years ago. Even better would be to have configure fail with an appropriate message if the _darwin OS_ is detected.

CharlieFenton:
Sandboxing is explained [here](http://boinc.berkeley.edu/trac/wiki/SandboxUser) and [here](http://boinc.berkeley.edu/sandbox.php). Unfortunately, the latter link is broken. I have asked @davidpanderson to fix it, but he has not yet done so.

CharlieFenton:
> Sandboxing is explained [here](http://boinc.berkeley.edu/trac/wiki/SandboxUser) and [here](http://boinc.berkeley.edu/sandbox.php). Unfortunately, the latter link is broken.

There is a new URL for the formerly broken link. The sandbox design diagram is now found [here](http://boinc.berkeley.edu/sandbox_design.php).

ChristianBeer:
@EvaSDK yes you understand correctly. The only supported way to build on Mac is the Xcode project. Everything else should be removed but since we lack people with autotools knowledge this was never a high priority. Although I want to switch the build system to cmake in the future I fully support removing unsupported parts of the current autotools files in the hope that makes the switch to cmake easier.

JuhaSointusalo:
I don't think removing `switcher` and `setprojectgrp` from Autotools build is a problem but before the rest of Mac specific stuff is removed maybe it's better to check with projects how many of them are using Autotools to build their science apps. I think there has been a few patches submitted just because some project wanted to do that.

I'm not saying that @CharlieFenton should officially support Autotools for building science apps. More like if a project wants to use Autotools then they also volunteer to keep it working.

EvaSDK:
If there is a will to change to cmake eventually, have you considered meson ?

CharlieFenton:
> before the rest of Mac specific stuff is removed maybe it's better to check with projects how many of them are using Autotools to build their science apps.

I agree with @JuhaSointusalo that some projects may want to build the BOINC libraries for the Mac using autotools. I should have suggested removing only those parts of configure and make which are used to build the _client_, _Manager_, _screensaver coordinator_ (_BOINCSaver.saver_) and _default graphics_ (_boincscr_) for the Mac / Darwin. It makes sense to still allow the 6 Mac BOINC libraries to be built using autotools: __libboinc.a_, libboincapi.a_, _libboinc_graphics2.a_, _libboinc_opencl.a_, _libjpeg.a_ and _libboinc_zip.a_.

I am not familiar enough with autotools to do any of this.

For those who want to build for the Mac from the command line, I have provided a series of scripts in the mac_build directory to build all or any portion of BOINC (including the 6 libraries) using the CLI API xcodebuild, as well as scripts to build all the Mac dependencies. For the 6 libraries this is an alternative to using autotools. This is all documented for each version of BOINC in the file _mac_build/HowToBuildBOINC_XCode.rtf_ in the git repository for that version.

TheAspens:
@EvaSDK and @CharlieFenton - it appears that discussion on this item has stalled and that there is not support for moving forward on this pull request.

Can you please comment on if my understanding is correct?  If so, can we close this ticket?

Note:  in the absence of any further discussion in the next two weeks, we will close this pull request.

TheAspens:
Given the lack of further discussion, we decided on the contributor call to close this ticket.  If this is in error, please re-open and let us know what the next steps are.

PR #2782, opened 2018-10-24
Drop bashisms in configure.ac and macros

EvaSDK:
Most platforms ship bash nowadays but it's still bad practice to rely on it as far as autotools go.
This can be tested using `CONFIG_SHELL=/bin/dash ./configure` for example which produces errors like `./configure: 36970: test: yes: unexpected operator`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -439,20 +439,19 @@ dnl m4/boinc_check_fcgi.m4
 BOINC_CHECK_FCGI
 fi
 dnl ---------- SSL (m4/check_ssl.m4)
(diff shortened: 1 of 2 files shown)


AenBleidd:
@davidpanderson, @ChristianBeer, could you please review this? Thanks

davidpanderson:
Not my area of expertise; I defer to Eric Korpela

ChristianBeer:
The code looks fine but I would like to do a local test if the logic is still correct. I'll probably do this over the weekend.

PR #2786, opened 2018-10-29
Image File Compression

Peppernrino:
Ran a series of lossless compression algorithms to make the images smaller. Total reduction of ~13MB.


CharlieFenton:
Is there any reason to even keep _clientgui/res/templates/icons_ and _clientgui/res/templates/skins_ in the repository? I don't think they are used anywhere.

Peppernrino:
If anything can be removed entirely, I'm all for it.

As for the other files displaying properly, I don't see why they wouldn't. They're the same pictures, just smaller. Unless something is calling the images by their SHA hashes, we should be good. :)

CharlieFenton:
My question was directed primarily to @romw and @davidpanderson, because they seem to be artwork which was used to create the images tat are actually used, but they also may be obsolete or at least no longe needed:
> Is there any reason to even keep clientgui/res/templates/icons and clientgui/res/templates/skins in the repository? I don't think they are used anywhere.

Peppernrino:
Ah! Pardon me, friend. Just anxious to contribute to something big and useful. Watching the pot boil. lol :D

TheAspens:
@Peppernrino - which tool did you use for the compression and what settings?

Peppernrino:
Example here: https://pastebin.com/MTtcVck2 :D

Peppernrino:
Another gentleman has squeezed a bit more juice from the image files, and I am currently sorting out a few things with him. It affects this pull. Please hold until further notice. Should be later today. :)

Peppernrino:
Ready to go. The other gentleman scrapped his work. :)

Rytiss:
I had the most concern about JPGs, but they don't seem to show any extra conversion artifacts. I also checked a somewhat random sample of files on Windows 10 and all seem to be displayed correctly using standard Windows image viewer.

As for removing unused assets, that should not block the merge of this commit. An issue should be raised and somebody can review the files later.

lfield:
I looked at a random selection of images and they seem fine to me. Any objection to merging?

CharlieFenton:
I have no objection

PR #2789, opened 2018-10-31
client: prevent a situation where a job is not throttled

davidpanderson:
I noticed an Amicable Numbers 4-CPU job wasn't being throttled,
while other jobs were.
Two types of jobs are exempt from throttling: GPU and non-CPU-intensive.
This app version was neither.
And yet it was marked as <dont_throttle/> in client_state.xml.
This is read at startup, and prevents that app version from being throttled
even if it not GPU and is not non-CPU-intensive.

I'm not sure how this situation arose.
Possibly the project had a temporary misconfiguration
where that plan class (mt) used a GPU.
This set the dont_throttle flag, and it stuck.

Anyway, the fix is to not parse <dont_throttle/> from app versions
in the client state file.
There's no need to - after parsing the app_version record,
dont_throttle is set according to whether it uses a GPU.

### DIFF
--- a/client/client_types.cpp
+++ b/client/client_types.cpp
@@ -891,7 +891,6 @@ int APP_VERSION::parse(XML_PARSER& xp) {
             }
             continue;
         }
-        if (xp.parse_bool("dont_throttle", dont_throttle)) continue;
         if (xp.parse_bool("is_wrapper", is_wrapper)) continue;
         if (xp.parse_bool("needs_network", needs_network)) continue;
         if (log_flags.unparsed_xml) {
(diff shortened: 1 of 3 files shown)


TheAspens:
@davidpanderson - I was reviewing the documentation for this setting here: https://boinc.berkeley.edu/trac/wiki/AppVersionNew

> dont_throttle: set if the app does its own CPU throttling (e.g., the Virtualbox wrapper). 

Are you sure that we want to make this change?  

@lfield - you use the virtualbox wrapper - what would be the the impact if this field was changed?

davidpanderson:
You're right - projects are supposed to be able to specify dont_throttle if the app does its own throttling.
The problem here is actually that the Amicable Numbers app specifies it, but doesn't throttle.
I'll notify them.

Issue #2792, opened 2018-11-01
Scalability of query batches RPC


lfield:
Calling query batches with O(150K) tasks in the system runs out of memory. 

`    response = query_batches(req)`

```
<br />
<b>Fatal error</b>:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 32 bytes) in <b>/boincdata/boinc/project/lhcathome-dev/html/inc/db_conn.inc</b> on line <b>144</b><br />
```

PR #2801, opened 2018-11-06
remote job submission: increase handler memory limit to 4G

davidpanderson:
address issue #2792


lfield:
Tested and can confirm it fixes the issue. Thanks.

PR #2806, opened 2018-11-07
scheduler: parse <wsl> and <wsl_available>

davidpanderson:
Currently these generate a server log error message.
TODO: add plan class support for WSL


TheAspens:
Looks fine.  Thanks.

PR #2810, opened 2018-11-10
Check-in notes removed

adamradocz:
I removed the check-in notes to make the repository cleaner. They had no any meaningful purpose, but if you disagree you can close this PR.


RichardHaselgrove:
Might I respectfully suggest that we DON'T do this? The checkin notes don't really get in the way, but some of the early history and reasoning behind changes isn't available anywhere else. There are a lot of *** empty log message *** in the early SVN commits: one which caught my eye was aad7a928b2f061a746668c1d690b3e33839c82e8, which would be meaningless without the notes. Just one example.

JuhaSointusalo:
The checkin notes were updated as part of the commit, [like so](https://github.com/BOINC/boinc/commit/aad7a928b2f061a746668c1d690b3e33839c82e8#diff-53182a67f97d205ff3b5ec2ac59e2bc7). Removing the files now would remove the changes to the checkin notes in old commits.

So I guess the question is, how often you need to not view old commits but search through the old commits messages?

RichardHaselgrove:
I have kept my own search list of SVN commits, from the time of the first (failed) transfer to Github. I refer to that - not often, but sometimes. So, personally, I'm probably OK: but it's not about me, it's about the whole community. Different people have different ways of working.

davidpanderson:
I refer to them frequently.  Keep them.

PR #2811, opened 2018-11-10
lib: fix xml_escape(), again

### LINKED ISSUE #2809: Stats export with ] in them break after #2784
After PR #2784 went in, the export of team names with ] in them is broken. The PR changed the logic to silently eat the ] when it wasn't part of the ending tag.

JuhaSointusalo:
Commit ff0b8d08985707f089f934816b0d28841b709c80 (fix CDATA escaping) introduced two bugs to xml_escape():

- it eats ']' characters that are not part of ']]>'
- it eats the next character after ']]>'

Fixes #2809

### DIFF
--- a/lib/parse.cpp
+++ b/lib/parse.cpp
@@ -376,7 +376,9 @@ void xml_escape(const char* in, char* out, int len) {
             if (!strncmp(in, "]]>", 3)) {
                 strcpy(p, "]]&gt;");
                 p += 6;
-                in += 3;
+                in += 2;  // +1 from for loop
+            } else {
+                *p++ = x;
             }
         } else {
             *p++ = x;


JuhaSointusalo:
The code still has (at least :/) one bug.  For some reason `xml_escape()` eats characters less than 32. If it's given a string like `]\xf]\xf>` it munches the `\xf`s and leaves `]]>` behind, putting us back to where we started from.

I think there's two ways to fix it. Either escape not `]]>` but all `>` characters or don't eat characters. The first one is allowed by XML standard. The second one would require figuring out why such a low level function gets to decide which characters are good and which are bad. Both would require full testing of entire BOINC and I'm not too keen on doing that right now.

Quite frankly, I'm beginning to think that the next time someone finds a bug in BOINC's XML code it's time to start porting the code to some XML library.

lfield:
Looks ok to me. I would suggest that if there is any other issue we back out these changes, keep the original issue for this release and then move to an XML library afterwards.

PR #2816, opened 2018-11-12
[WIP] Move Windows dependencies from git to AWS S3

AenBleidd:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -60,14 +60,13 @@ environment:
   matrix:
     - TOOLCHAIN_VERSION: 12.0
       solution_name: win_build\boinc_vs2013.sln
-      depends_git_path: http://boinc.berkeley.edu/git/boinc_depends_win_vs2013.git
       depends_path: C:\projects\boinc_depends_win_vs2013
 
 cache:
   - C:\projects\boinc_depends_win_vs2013 -> appveyor.yml, win_build\load_dependencies.bat
 
 before_build:
-  - call win_build\load_dependencies.bat %depends_git_path% %depends_path% %platform% %configuration%
+  - call win_build\load_dependencies.bat %depends_path% %platform% %configuration%
   - if %TOOLCHAIN_VERSION%==12.0 call "%VS120COMNTOOLS%\vsvars32.bat"
 
 build_script:
--- a/win_build/load_dependencies.bat
+++ b/win_build/load_dependencies.bat
@@ -1,69 +1,28 @@
-@echo off
-set dependencies_git_path=%1
-set dependencies_path=%2
+rem @echo off
+set dependencies_path=%1
 set revision_lst_path=%dependencies_path%\revision.lst
-set platform=%3
-set configuration=%4
+set platform=%2
+set configuration=%3
+
+powershell -Command "(New-Object Net.WebClient).DownloadFile('https://s3-us-west-2.amazonaws.com/boinc-win-dependencies/latest/revision.lst', '%TEMP%\revision.lst')"
+set /p last_revision=<%TEMP%\revision.lst
 
 if exist %revision_lst_path% (
   set /p temp_saved_revision=<%dependencies_path%\revision.lst
   set saved_revision=%temp_saved_revision%
 
-  for /f %%i in ('git ls-remote %dependencies_git_path% HEAD') do (
-    set last_revision=%%i
-    if "%last_revision%" == "%saved_revision%" (
-       echo Dependencies are up-to-date
-       goto :EOF
-    )
-    goto :PROCESS
+  if "%last_revision%" == "%saved_revision%" (
+    echo Dependencies are up-to-date
+    goto :EOF
   )
 )
-
-:PROCESS
 
 if exist %dependencies_path% (
   rd /s /q %dependencies_path%
 )
 
-git clone -q --branch=master %dependencies_git_path% %dependencies_path%
-  
-rd /s /q %dependencies_path%\.git
-
-if "%platform%" == "Win32" (
-  rd /s /q %dependencies_path%\curl\mswin\x64
-  rd /s /q %dependencies_path%\freetype\mswin\x64
-  rd /s /q %dependencies_path%\ftgl\mswin\x64
-  rd /s /q %dependencies_path%\openssl\mswin\x64
-  rd /s /q %dependencies_path%\sqlite3\mswin\x64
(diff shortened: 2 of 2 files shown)


adamradocz:
I think it's fine now for temporarily, cause I strongly advocate to to a package manager like vcpkg as proposed at #2348.
There won't be server cost either.

AenBleidd:
#2348 still in progress and while we're awaiting it we can move forward.
> temporarily

There is nothing more constant than temporary solution ;)

adamradocz:
IMO, developing and maintaining a script is more work than using a package manager. What're the benefits of self compiling the 3rd party libraries?

ab1vl:
Hi Adam,

Of topic: I want to get a few GPUs for my aging fleet of 2 and 4 core
mobos.  I got a Nvidia GT 710 with 1GB memory and it runs well on linux and
windows.cc stated as 3.6.  Good enough for qualitative testing I think.
I guess I can't go wrong with nVidia, and the GT 710 above goes for $35
from Amazon. The GT 730 is about $65 but I wonder if I would see any
difference  as the rated cc  is still 3.5 and it draws much more current.

Question: does BOINC limit GPU tasks to 1 per GPU, or more.  Running
Einstein, SETI and some WCG WUs use 85-100% of the memory and run about 65C
on the GT-710 according to Psensor or GPU-Z. I typically limit CPU to
50-70% to keep temps 55-65C.

Do you have any thoughts about other GPUs than nVidia; I see you are keen
on AMD.

tnx, chuck

adamradocz:
@ab1vl Here is the compatison of your cards: [https://gpu.userbenchmark.com/Compare/Nvidia-GeForce-710M-vs-Nvidia-GeForce-GT-730/m10751vsm12582](https://gpu.userbenchmark.com/Compare/Nvidia-GeForce-710M-vs-Nvidia-GeForce-GT-730/m10751vsm12582)

If you're on budget and if I were you, I'd buy a GTX 1030 or a used GTX 1050. You can check them too in the website I linked. IMO, the 710 or 730 is useless.

Please open an issue for this topic, if you have more question.

tristanolive:
I think this makes sense to be pulled as a temporary workaround if:

- there is a current need to update dependencies
- the package manager approach is some ways off
- multiple developers would benefit and none should be hindered

AenBleidd:
@tristanolive,
> there is a current need to update dependencies

There is no need right now but it would require.
E.g. if we decide to move from own XML parser implementation we will require this library in pre-build dependencies. While moving on some XML parser it would be good to make some unit tests to verify that we would not broke anything. This will require adding GTest to pre-build dependencies.

AenBleidd:
@JuhaSointusalo,

> git provides change log and checksums. A bit unfortunate to lose those but if it's temporary I guess it's ok.

It doesn't mean a lot with prebuilt binaries, I think.

> It seems to me that this just moves the management elsewhere but doesn't necessarily make it easier. You still need someone to actually do the management.

I plan to do this next several years. If someone else will decide to do this I can give him access to this particular bucket. Or we can use any other account or provider to store there files.

I do agree that creating the script will be better (I don't know about vcpkg).

In any case, I just want to let anyone speak on this and decide whether we require this or not.

By the way, this will decrease load on Berkeley servers because on every pull request build it asks whether there were changes on git and this requires to make quite big pull that is very heavy because of nature of files stored there.

In a day there will be a contributors call and I'd like to hear opinion of everyone.

CharlieFenton:
> IMO, developing and maintaining a script is more work than using a package manager. 

I had never heard of package managers until your mentioned them. If they allow building the dependencies locally more easily than a script, that is fine with me.

> What're the benefits of self compiling the 3rd party libraries?

Isn't that what _vcpkg_ does? One benefit is that we don't need to keep such large binaries in a repository, which to me seems cumbersome, especially since you can't keep them in GitHub. If I understand _vcpkg_ correctly, it is a command line utility, so one would use some sort of scripts or configuration files to run it automatically on the local computer to build the dependencies. This script (or whatever the input to a package manager uses) can be kept on GitHub along with the other BOINC source files. This would permit versioning, change log and checksums.

 But I don't really have a strong opinion one way or the other.

davidpanderson:
I'd be happy to update the library repo as needed.

adamradocz:
@CharlieFenton [Vcpkg is a C++ package manager](https://docs.microsoft.com/en-us/cpp/vcpkg?view=vs-2017), and yes, you're right it downloads sources instead of binaries. I thought it downloads binaries, my bad :).

When you mentioned writing a script, I thought you want to write a script from scrath rather than writing a script for Vcpkg. Also my bad :).

AenBleidd:
Title changed by adding [WIP] prefix to prevent from accidental merging before final decision will be made by community

brevilo:
FWIW, did you guys ever consider a good compromise between git and S3? How about [Git LFS](https://git-lfs.github.com)? It's even directly [supported](https://help.github.com/articles/configuring-git-large-file-storage/) by GitHub. Couldn't this be providing the best of both worlds...?

JM2C

TheAspens:
It looks like conversation on this issue has stalled.  My understanding is that there are several options pending:
1) Use the s3 that @AenBleidd  setup (which is what this pull request is)
2) Keep using the existing boinc repo (which above @davidpanderson offers to update)
3) Use GIT LFS as @brevilo has suggested
4) Start building with VS2017 so that we don't need this

This was discussed on today's contributor call and we wondered how to move forward for resolution to this discussion.  If this cannot be decided via discussion here, would it be possible to discuss on the next contributor call?  If not then, can we set up a time for a discussion so that we can make a decision?  Thanks!

AenBleidd:
Sorry I missed contributors call today.
I believe that point 4 is better because it solves more issues:
- move dependencies to vcpkg
- upgrade solution to VS2017 that is more modern (maybe, to VS2019 now?)
- remove both solutions for VS2010 and VS2013 that makes development and testing easier
- add possibility to have true CI: now release version that shipped to users if being built with VS2010 but on AppVeyor VS2013 is used for commit-build (I found to easy way to let VS2010 solution being built in non-interactive mode)
But as far as I understand, porting to VS2017 is stalled also. I could probably take a look at @JuhaSointusalo branch and possibly help him to finish this porting.

adamradocz:
If I may, I vote for option 4, but moving to VS 2019.

AenBleidd:
Closing it in favor of #2348

PR #2818, opened 2018-11-13
added default project value disable_account_creation_rpc

Germano0:
While testing [GDPR compliance](https://boinc.berkeley.edu/trac/wiki/GdprCompliance) I have noticed that in `project/config.xml` the attribute `disable_account_creation_rpc` was missing

### DIFF
--- a/py/Boinc/setup_project.py
+++ b/py/Boinc/setup_project.py
@@ -514,7 +514,7 @@ class Project:
         config.max_wus_to_send = 50
         config.daily_result_quota = 500
         config.disable_account_creation = 0
-        config.disable_account_creation_rpc = 1
+        config.disable_account_creation_rpc = 0
         config.disable_web_account_creation = 0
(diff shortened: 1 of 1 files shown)


davidpanderson:
? this should not be the default value.

Germano0:
> ? this should not be the default value.

What value should be the default?

lfield:
Disable account create should by default probably be set to true.

lfield:
The account creation RPC needs to be enabled for the integration tests.

RichardHaselgrove:
It depends how early in the configuration process you want members of the public to start joining your test team - which in turn depends on whether you are testing integration on a public or private IP or url.

I think most projects do test in public, but with account creation requiring an invitation code: the two defaults should be considered in tandem.

lfield:
The php unit tests use this RPC to create a test account so the Travis integration tests fail if this is not enabled.

Issue #2819, opened 2018-11-14
Update the community profile on GitHub


adamradocz:
I propose to update the [community profile](https://github.com/BOINC/boinc/community).

1. The Contributor Covenant Code of Conduct template suggested by GitHub is similar that the BOINC uses but more detailed but still simple. I propose to use that template.
2. Contributing guidelines would make easier for new people how can they contribute.
3. IMO standardizing the issues with issue templates would make our work a little bit easier, by showing people step-by-step instructions on how to give a clear issue description. I propose 3 templates, Bug report, Feature request, and Other issues. [GitHub's temples](https://github.com/BOINC/boinc/issues/templates/edit) could be the base for each template.
4. I think we could also benefit a Pull request template too.

Any thoughts on this?

AenBleidd:
@adamradocz, fell free to create PR with changes you think are good

PR #2821, opened 2018-11-14
Update code of conduct

adamradocz:
Response to the first item at #2819

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,31 +1,76 @@
-•  Be welcoming
+# Contributor Covenant Code of Conduct
 
-Aim to welcome and support people of all backgrounds and identities, of any race, ethnicity, culture, national origin, color, immigration status, social and economic class, educational level, sex, sexual orientation, gender identity and expression, age, size, family status, political belief, religion, and mental and physical ability.
+## Our Pledge
 
-•  Be respectful and friendly
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
+level of experience, education, socio-economic status, nationality, personal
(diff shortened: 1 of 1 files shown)


lfield:
This change requires general project consensus. Is this a PMC or Committers vote?

adamradocz:
> This change requires general project consensus. Is this a PMC or Committers vote?

No, this change hasn't voted. I just proposed an idea, to move a more standardized, but very similar code of conduct that we have.

TheAspens:
This would require a PMC vote for approval, but I would like to have a majority of committers indicate their support of this as well as a good discussion of it before taking it before the PMC.

tristanolive:
I don't think we gain anything by using the github template. If I recall, that template (among others) was used as a starting point for the BOINC code of conduct and ideas from different resources were melded together into what we have.

If there are specific shortcomings, it makes sense to address them, but I'd view this move as unnecessary and maybe even a step backwards in terms of readability (I may just be missing the specifics in the diff).

brevilo:
I had a brief look at the proposed changes. Before I dive any deeper I'd like to understand the motivation for those changes:
* why are these changes necessary?
* what problem do they address?
* how are they an improvement over the current doc?

adamradocz:
> I had a brief look at the proposed changes. Before I dive any deeper I'd like to understand the motivation for those changes:
> 
> * why are these changes necessary?
> * what problem do they address?
> * how are they an improvement over the current doc?

I thought a standardized version of the COC is easier to understand, because of its standardized nature. That's it, nothing fancy.

brevilo:
Thanks, I see your point Adam. I don't have a strong opinion on this: I'm neither sure the COC represents such a standard, nor do I think we stray far from any COC as they're all structured similarly, while still representing each project's specifics. Again, I'm neutral on this as I mostly care about the content and its clear communication and not so much about a perceived standardization. Up to the PMC...

adamradocz:
I'm going to close this issue, because the trouble bigger with it, then the expected gain.

PR #2833, opened 2018-11-20
[AppVeyor] Deploy to Bintray Implementation

AenBleidd:


AenBleidd:
@ChristianBeer, could you please verify why I got an error while uploading to bintray? Thanks
`{"message":"This resource requires authentication"}`

ChristianBeer:
It seems that the api_key specified in the AppVeyor deploy configuration is not available as the environment variable 'BINTRAY_API_KEY'. You should add a check for that and abort deployment in deploy_to_bintray.bat if this variable doesn't exist or is empty. I'll provide the proper yaml entry to have this variable in the build stage.

ChristianBeer:
You need to change the `environment`section in the `appveyor.yml` file and add this:
```
  BINTRAY_API_KEY:
    secure: kZI9k0Kh2bFSCbXfkz+J16fGNAee1ToRMl10D8QPQsKpC2PqhF/uVMpd6gRC+OSI
```

AenBleidd:
Looks like ready for review

ChristianBeer:
Looks good, except:

1. The SHA1 in `APPVEYOR_REPO_COMMIT` is not what we want. I had to experiment with this on Travis too. Try to use `APPVEYOR_PULL_REQUEST_HEAD_COMMIT` next. This should work and create the correct version string ending with the commit SHA1 from the pull request that is tested.

1. The link to the job log is also not working. See [deploy/deploy_to_bintray.sh#L127](https://github.com/BOINC/boinc/blob/master/deploy/deploy_to_bintray.sh#L127) where I had to add `'` around the URL for bintray to treat this as text and not fetch the content of the URL. This seems to be an undocumented feature.

AenBleidd:
@ChristianBeer, regarding point 1, it is already used here:
https://github.com/BOINC/boinc/blob/bffe87cef3a79f41181b9faba5a30aeb59f4554d/deploy/prepare_deployment.bat#L14

ChristianBeer:
Nope that both didn't work either. We might have to find another way to get the same SHA1 as `TRAVIS_PULL_REQUEST_SHA` or file a feature request with AppVeyor.

You should probably comment out the release notes stuff as this conflicts with Travis. I just realized this. We might need to find a different solution.

ChristianBeer:
I figured out what's wrong with the batch scripts. We need to adjust the prepare and the deploy script to set the correct variable values. It turns out Windows is expanding variables at parse time and not at execution time. So changing a variable and reusing it in the same block has no effect unless you turn on delayed expansion and use `!` instead of `%` around the variable.

So here is what is needed: `setlocal EnableDelayedExpansion` needs to be added as first line to each batch file in `deploy`. Every variable that gets reset within the script needs to be enclosed in `!` not `%` especially the ones in the if blocks. See my test code in https://github.com/BOINC/boinc/pull/2841/files.

If you could fix both scripts I'm able to merge this PR.

appveyor.yml
```diff
@@ -62,6 +62,8 @@ environment:
       solution_name: win_build\boinc_vs2013.sln
       depends_git_path: http://boinc.berkeley.edu/git/boinc_depends_win_vs2013.git
       depends_path: C:\projects\boinc_depends_win_vs2013
+  BINTRAY_API_KEY:
```
ChristianBeer: Please also fix the indentation of this line to match the level above. The next line needs to get two more spaces so it is slightly more indented than the first line.

deploy/deploy_to_bintray.bat
```diff
@@ -0,0 +1,93 @@
...
+echo Creating package %pkg_name%...
+set data={\"name\": \"%pkg_name%\", \"desc\": \"%pkg_desc%\", \"desc_url\": \"auto\", \"website_url\": [\"%WEBSITE_URL%\"], \"vcs_url\": [\"%VCS_URL%\"], \"issue_tracker_url\": [\"%ISSUE_TRACKER_URL%\"], \"licenses\": [\"LGPL-3.0\"]}
+%CURL% -H Content-Type:application/json -X POST -d "%data%" "%API%/packages/%BINTRAY_REPO_OWNER%/%BINTRAY_REPO%"
+
+for /f "tokens=2-4 delims=/ "  %%a in ("%date%") do (set MM=%%a& set DD=%%b& set YYYY=%%c)
+if not defined build_date (
+    set build_date=%YYYY%-%MM%-%DD%
+)
+if not defined git_rev (
+    set git_rev=%APPVEYOR_REPO_COMMIT:~0,8%
```
ChristianBeer: Please keep this and the next two blocks as they are now. If this is a PR or a scheduled build the variables should already be set correctly.

AenBleidd:
> If you could fix both scripts I'm able to merge this PR.

Will do

appveyor.yml
AenBleidd: @ChristianBeer, are you sure? I thought that BINTRAY_API_KEY should be under environment section, not matrix or matrix entry level....

ChristianBeer: You are right. It's on the correct level.

ChristianBeer:
Is it possible that variables set in prepare_ are not available in deploy_ ?

AenBleidd:
> Is it possible that variables set in prepare_ are not available in deploy_ ?

Do not know yet, investigating...

AenBleidd:
Looks like it won't work:
https://stackoverflow.com/questions/33898076/passing-variable-out-of-setlocal-code

AenBleidd:
@ChristianBeer, what do you think about merging these two scripts into one?
I think it is better than trying to resolve this issue.

AenBleidd:
Ok, I've found the way out.

ChristianBeer:
I'm not fond of combining the scripts as I wanted to have it separated in case we want to deploy to some other service. But right now I don't have a solution to the problem at hand. I would still like to keep the two sections separate and the structure the same as the linux scripts. This way I hope it is easier to change things in the future. If I have to change something in the *.sh scripts I have a good idea where to find the same thing in the *.bat scripts if the overall structure is the same. please call the combined script  `deploy_to_bintray.bat`

AenBleidd:
@ChristianBeer, done. Please review it one more time

PR #2838, opened 2018-11-23
Server version publication improvements

lfield:
These improvements were the result of discussions with Einstein@home.

### DIFF
--- a/html/user/get_project_config.php
+++ b/html/user/get_project_config.php
@@ -66,10 +66,6 @@ echo "<project_config>
     <web_rpc_url_base>".secure_url_base()."</web_rpc_url_base>
 ";
 
-if ( isset($git_commit) ) {
-    echo "<git_commit>$git_commit</git_commit>\n";
-}
-
 if ( isset($server_version) ) {
     echo "<server_version>$server_version</server_version>\n";
 }
--- a/html/user/server_status.php
+++ b/html/user/server_status.php
@@ -189,7 +189,7 @@ function show_status_html($x) {
     global $server_version;
     global $git_url;
     if ( isset($server_version) && isset($git_url) ) {
-        echo "Server version: $server_version (<a href=\"$git_url\"</a>)<br>";
+        echo "Server version: <a href=\"$git_url\">$server_version</a><br>";
     }
 
     if ($j->db_revision) {
--- a/html/user/server_status.php
+++ b/html/user/server_status.php
@@ -187,9 +187,9 @@ function show_status_html($x) {
     end_table();
     
     global $server_version;
-    global $git_commit;
-    if ( isset($server_version) && isset($git_commit) ) {
-        echo "Server version: $server_version (<a href=https://github.com/BOINC/boinc/commit/$git_commit>$git_commit</a>)<br>";
+    global $git_url;
+    if ( isset($server_version) && isset($git_url) ) {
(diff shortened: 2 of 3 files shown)


html/user/get_project_config.php
```diff
@@ -66,10 +66,6 @@ function show_platforms() {
     <web_rpc_url_base>".secure_url_base()."</web_rpc_url_base>
 ";
 
-if ( isset($git_commit) ) {
```
brevilo: Does this mean that the project config should not communicate any codebase details anymore? Why not replace this with the newly generated common tag base?

html/user/server_status.php
```diff
@@ -187,9 +187,9 @@ function show_status_html($x) {
     end_table();
     
     global $server_version;
-    global $git_commit;
-    if ( isset($server_version) && isset($git_commit) ) {
-        echo "Server version: $server_version (<a href=https://github.com/BOINC/boinc/commit/$git_commit>$git_commit</a>)<br>";
+    global $git_url;
+    if ( isset($server_version) && isset($git_url) ) {
```
brevilo: I recommend to make the link optional but always show the version (if available). Right now `$git_url` is set in a way that makes the following assumptions:
- the source repo was cloned via `https`, not `ssh`
- the source repo's relevant remote is called "origin"

I think that both assumptions shouldn't be made. At least the code should prepare for the case that they're no met, i.e. make the link/URL optional and don't just assume you got a valid `$git_url`.

py/Boinc/setup_project.py
```diff
@@ -345,25 +345,29 @@ def dir(*dirs):
...
-
+$git_url = "{git_url}";
 ?>
-'''.format(commit=commit, version=version)
-            f = open(os.path.join(dest_dir, 'release.inc'), 'w')
-            f.write(content)
-            f.close()
-            os.chmod(os.path.join(dest_dir, 'release.inc'), 0o644)
+'''.format(git_url=git_url, version=version)
+                f = open(os.path.join(dest_dir, 'release.inc'), 'w')
```
brevilo: Is this file only written once during the initial project instantiation? Shouldn't the version be a dynamic detail that always represents the current state?

html/user/get_project_config.php
lfield: No it now only uses the tag.

py/Boinc/setup_project.py
lfield: It is also created when doing upgrade.

html/user/server_status.php
lfield: The release.inc file can always be edited/deleted.

html/user/get_project_config.php
brevilo: Ah, ok, I [found it](https://github.com/BOINC/boinc/blob/57e2e5b66dfd588575812638d13f01b49c29de25/html/user/get_project_config.php#L69) myself.

html/user/server_status.php
brevilo: What's the point of this PR if projects are meant to edit `release.inc` manually afterwards? What I'm trying to convey here is that the current implementation makes assumptions that won't always be met. Even worse, it doesn't handle gracefully the case when the assumptions aren't met, producing an invalid HTML link.

lfield:
It has now been done a different way.

PR #2840, opened 2018-11-23
[Android] Changed the RESTART_PACKAGES to KILL_BACKGROUND_PROCESSES

adamradocz:
**Description of the Change**
RESTART_PACKAGES constant was deprecated in API level 8. The ActivityManager.restartPackage(String) API is no longer supported.

Replacement constant is KILL_BACKGROUND_PROCESSES.

**Release Notes**
N/A


adamradocz:
The KILL_BACKGROUND_PROCESSES constant was introduced in API level 8, and we require minimum 19, so it is backward compatible.

This PR doesn't interfere with the #2839 PR.

brevilo:
Thanks. One minor recommendation. Please add a little background/reasoning to your commit messages as well such that the git log becomes a bit more meaningful. By just saying **what** you state the obvious. The **why** is way more interesting and we can't ensure this PR will always remain available.

Issue #2826, opened 2018-11-15
BOINC refuses to add exclusive application it can't execute, like Windows UWP apps


GreenReaper:
On Windows 10, UWP apps such as the Xbox app (which may be impacted, particularly by GPU execution) cannot be added as exclusive applications because BOINC does not recognize them as executable:

> 'XboxApp.exe' is not an executable application.

This is even though the apps _can_ be executed, and indeed may be running, as that user or another.

Here is [a screenshot of the advanced security settings for this executable](https://www.dropbox.com/s/tufitql9gjwm2fd/UWPSecurity.png?dl=0).
(I wasn't able to upload it to the issue - "Something went really wrong".)

As you can see it provides a condition to executablilty:
> WIN://SYSAPPID Contains "MICOSOFT.XBOXAPP_..."

This is an AppContainer [package moniker](https://books.google.co.uk/books?id=y83LDgAAQBAJ&pg=PT981&lpg=PT981&dq=%22SYSAPPID%22+appcontainer&source=bl&ots=kHarMBMpkw&sig=3pUMUMJWS7nHgVcOosJY1rX5MLc&hl=en&sa=X&ved=2ahUKEwiD54PvptbeAhVTT8AKHawMDoUQ6AEwAnoECAUQAQ#v=onepage&q=%22SYSAPPID%22%20appcontainer&f=false) attached to the impersonation security token upon app execution.

The issue is that BOINC is checking for the ability to open for executability by the BOINC process:
https://github.com/BOINC/boinc/blob/f14d96d2be2e5a290177f7b5391b0802fbd3756c/clientgui/DlgExclusiveApps.cpp#L331

In practice this file _can_ be opened as executable, even by the user, but only when actually executing the app. Indeed it seems that this problem is already known (to an extent) based on a line just a little further down:
https://github.com/BOINC/boinc/blob/f14d96d2be2e5a290177f7b5391b0802fbd3756c/clientgui/DlgExclusiveApps.cpp#L360

---

I therefore suggest that executability not be tested - after all, you might want to disable BOINC when applications are running which the user in question cannot execute, even as an app, if it interferes with them.

At the end of the day, BOINC needs a list of filenames to processes it can see running. If if it gets a validly-_formatted_ filename (not necessarily an existing file), as `fn.IsOk()` [implies](http://proton-ce.sourceforge.net/rc/wxwidgets/docs/html/wx/wx_wxfilename.html#wxfilenameisok), should that not be enough?

---

[...]

PR #2843, opened 2018-11-23
Manager: in select exclusive app, don't check for existence, executability, or filename extension

davidpanderson:
... because the file may be inaccessible to or not executable by user.
Also don't require .exe or .app extension.

The user can use the file picker to locate the file, but they don't have to;
they can just type in the file name.

Addresses issue #2826

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/clientgui/DlgExclusiveApps.cpp
+++ b/clientgui/DlgExclusiveApps.cpp
@@ -282,71 +282,62 @@ void CDlgExclusiveApps::AddToListBox(wxListBox * theListBox) {
     bool hostIsWin = false;
     bool isDuplicate;
     wxArrayString appNames;
-    wxChar *extension = wxT("");
     wxString errmsg;
     CMainDocument* pDoc = wxGetApp().GetDocument();
 
     wxASSERT(pDoc);
     wxASSERT(wxDynamicCast(pDoc, CMainDocument));
 
+    // ? why use these instead of ifdefs?
     if (strstr(pDoc->state.host_info.os_name, "Darwin")) {
         hostIsMac = true;
-        extension = wxT(".app");
     } else if (strstr(pDoc->state.host_info.os_name, "Microsoft")) {
         hostIsWin = true;
-        extension = wxT(".exe");
     }
 
     pDoc->GetConnectedComputerName(strMachineName);
     if (pDoc->IsComputerNameLocal(strMachineName)) {
+        // if the client is local, use a file chooser dialog.
+        // This eliminates spelling errors.
+        // Note: the excluded app may be in an inaccessible directory,
+        // or the user might not know where it is.
+        // In this case they can just type the filename into the dialog
+        //
 #ifdef __WXMAC__
-        wxFileDialog picker(this, _("Applications to add"),
-            wxT("/Applications"), wxT(""), wxT("*.app"),
-            wxFD_OPEN|wxFD_FILE_MUST_EXIST|wxFD_MULTIPLE|wxFD_CHANGE_DIR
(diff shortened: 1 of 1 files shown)


clientgui/DlgExclusiveApps.cpp
```diff
@@ -282,71 +282,62 @@ void CDlgExclusiveApps::AddToListBox(wxListBox * theListBox) {
...
     bool isDuplicate;
     wxArrayString appNames;
-    wxChar *extension = wxT("");
     wxString errmsg;
     CMainDocument* pDoc = wxGetApp().GetDocument();
 
     wxASSERT(pDoc);
     wxASSERT(wxDynamicCast(pDoc, CMainDocument));
 
+    // ? why use these instead of ifdefs?
```
AenBleidd: I think the answer is 'because Manager could be run on one OS (e.g. Windows) and manage a client that run different OS (e.g. MacOS))

CharlieFenton:
The use of a picker rather than free text input helps prevent against typographical / spelling errors when adding to the list, and so I feel the existing approach is better. Similarly, requiring the correct extension also helps guard against faulty additions to the list. I believe that eliminating these checks will result in users complaining that the _Exclusive Applications_ feature does not work when the problem is actually due to bad data entry on their part, causing unnecessary work for our tech support volunteers.

Why can't the problem raised by issue #2826 be solved by simply eliminating the `if (!fn.IsOk() || !fn.IsFileExecutable()` checks and leaving everything else as is?

CharlieFenton:
On closer examination, I see that these changes do not eliminate the use of the file chooser dialog. However, please be aware that only the Windows file chooser dialog allows a user to enter free text (type into the dialog.) The Mac file picker does not allow this.

The concerns I expressed in my previous comment about not passing the extension `.app` or `.exe` from the chooser dialog, and about not checking for that extension still apply .

davidpanderson:
Keep in mind that the original goal was to let people specify apps they couldn't see due to permissions.  These won't be visible in a file picker.  (of course, they can always edit the config file but that's ugly).

CharlieFenton:
> But someone might want a regular Unix program (like ssh) to be exclusive, right?

You are correct that requiring the extension makes it impossible to add command-line executables. But I don't know how likely it is that anyone would actually need to add one of those to the exclusive applications list. 

So the question is whether the ability to select non-GUI apps is more important / useful than the simplicity and error reduction provided by the file chooser filtered by extension. I don't know the answer to that.

I wonder if there might be a practical way to offer the best of both worlds: use the extension-filtered file chooser as the default, but also allow a more general unfiltered free text entry as an advanced option. (This  may be a bit more complicated to implement on the Mac since the Windows chooser dialog always allows free text entry while the Mac chooser dialog never allows it. So it would require opening a separate text entry window on Macs.)

TheAspens:
@AenBleidd - you added yourself as a reviewer to this pull request.  Can you update us with what the next steps are to move this PR forward?  Let me know if you want someone else to move it forward.  Thanks!

clientgui/DlgExclusiveApps.cpp
AenBleidd: This comment should be removed from source code because it could just confuse someone

AenBleidd:
@TheAspens, thank you for reminder

AenBleidd:
@davidpanderson, could you please remove the comment I mentioned above? Then I believe this PR will be OK to me

AenBleidd:
Looks fine. Thanks!

PR #2845, opened 2018-11-24
Add Hyper-V support to Vagrant

adamradocz:
This PR is based on #2842, please merge that first.

**Description of the Change**
On Windows, the [VirtualBox can't be used](https://forums.virtualbox.org/viewtopic.php?f=6&t=85438) if someone uses Hyper-V by default.

The user doesn't have to do anything differently. If VirtualBox is installed, the Vagrant detects it and use it as default.

**Release Notes**
N/A

### DIFF
--- a/android/Vagrant.README.md
+++ b/android/Vagrant.README.md
@@ -30,7 +30,7 @@ Provide a turn-key VM for Android development
    1. Wait until the final reboot finished
    1. **From this point on you don't need Vagrant anymore**
       1. Don't run `vagrant up` again!
-      1. Just use VirtualBox to stop/start your new shiny VM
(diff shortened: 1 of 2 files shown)


adamradocz:
The ideal situation would be that Vagrant install `qemu-kvm` on Hyper-V, but not install the `virtualbox-guest-x11` package, but I couldn't figure out, how to handle it with Vagrant, so I put an extra note in the Readme.

adamradocz:
Is that possible travis gives false unsuccessful test?

AenBleidd:
I've restart the job. It could help

brevilo:
Thanks!

PR #2851, opened 2018-11-26
Add config parameter to hide server status page version and db_schema…

drshawnkwang:
… if desired.

Part of https://github.com/BOINC/boinc/issues/2688

**Description of the Change**
Adds a config.xml parameter to hide the server version and db_schema on the server status page.

**Release Notes**
N/A

### DIFF
--- a/html/user/server_status.php
+++ b/html/user/server_status.php
@@ -185,15 +185,18 @@ function show_status_html($x) {
         ";
     }
     end_table();
-    
-    global $server_version;
-    global $git_commit;
-    if ( isset($server_version) && isset($git_commit) ) {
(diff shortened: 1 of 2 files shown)


TheAspens:
@drshawnkwang - can you resolve the conflict and update to the latest server code?  Tx.

TheAspens:
Negative names are confusing.  Can we call the variable "show_server_status_version" and make the default value 1?

That way the logic will be "if show then" rather than "if not hide then".

drshawnkwang:
Good news: I believe PR #2861 (merged) does what this PR is trying to do, and in a better way as well. Thus I am closing the PR as it is no longer needed. If people (@TheAspens) disagree, then we can revisit and reopen this later.

PR #2852, opened 2018-11-26
Travis Bintray: add attributes to versions for better cleanup

ChristianBeer:
**Description of the Change**
Attributes can be used to easily delete old versions. A script can check if a PR is closed and delete all versions for this PR on bintray. Or it could delete all versions older than a threshold.
Attribtes only need to be added once. Currently this is done in the Travis build because this is running for every PR.

**Alternate Designs**
None

**Release Notes**
N/A

### DIFF
--- a/deploy/deploy_to_bintray.sh
+++ b/deploy/deploy_to_bintray.sh
@@ -111,6 +111,15 @@ data="{
 set +x
 ${CURL} -H Content-Type:application/json -X POST -d "${data}" "${API}/packages/${BINTRAY_REPO_OWNER}/${BINTRAY_REPO}/${PKG_NAME}/versions"
 
+echo "Adding attributes to version ${VERSION}"
+# this can be used to cleanup old versions
+pr="${TRAVIS_PULL_REQUEST:-0}"
+created=$(date -Is)
+data="[{\"name\": \"PR\", \"values\": [\"${pr}\"], \"type\": \"string\"},
+            {\"name\": \"create_time\", \"values\": [\"${created}\"], \"type\": \"date\"}]"
+echo $data
+${CURL} -H Content-Type:application/json -X POST -d "${data}" "${API}/packages/${BINTRAY_REPO_OWNER}/${BINTRAY_REPO}/${PKG_NAME}/versions/${VERSION}/attributes"
+
 echo "Uploading and publishing ${SOURCE_DIR}/${BOINC_TYPE}.7z..."
 if [ -f "${SOURCE_DIR}/${BOINC_TYPE}.7z" ]; then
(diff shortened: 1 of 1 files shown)


ChristianBeer:
Just FYI: I already have a script to cleanup bintray using those two attributes. I need a bit more time to polish it.

deploy/deploy_to_bintray.sh
```diff
@@ -111,6 +111,15 @@ data="{
 set +x
 ${CURL} -H Content-Type:application/json -X POST -d "${data}" "${API}/packages/${BINTRAY_REPO_OWNER}/${BINTRAY_REPO}/${PKG_NAME}/versions"
 
+echo "Adding attributes to version ${VERSION}"
+# this can be used to cleanup old versions
+pr="${TRAVIS_PULL_REQUEST:-0}"
+created=$(date -Is)
+data="[{\"name\": \"PR\", \"values\": [\"${pr}\"], \"type\": \"string\"},
+            {\"name\": \"create_time\", \"values\": [\"${created}\"], \"type\": \"date\"}]"
+echo $data
```
AenBleidd: I think this was added for debug purposes and should be removed

AenBleidd: Ok, but this is the only information that is shown this way in this script and it makes duplicate output in log:

```
Adding attributes to version PR2852_2018-11-26_07111b32
[{"name": "PR", "values": ["2852"], "type": "string"}, {"name": "create_time", "values": ["2018-11-26T20:15:52+0000"], "type": "date"}]
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   271    0   124  100   147    472    560 --:--:-- --:--:-- --:--:--   561
[{"name":"create_time","type":"date","values":["2018-11-26T20:15:52.000Z"]},{"name":"PR","type":"string","values":["2852"]}]
```

So personally I think that it is not required here

ChristianBeer: I started to output the request data in the API call before the call for the cleanup script thus I also used it for this little snippet. This way it's easier to see what information was send and you also see the response. In this particular case the API responds with the values it recognized (they do type recognition if you omit the type parameter). In the other cases you see other response messages.

In my testing I had problems to get the json payload correct. So in case this breaks I don't want to add the output as part of the investigation but already see what the request was and develop a fix.

I probably add this kind of output to the other API calls in this script to be consistent once I polished the cleanup script.

AenBleidd:
Ok, I suppose this should be done separately from this pull request.
Merging it now.
Thanks

PR #2854, opened 2018-11-27
Fix vagrant

brevilo:
Fixes #

**Description of the Change**
* Updated vagrant to get it up and running again
* Updated OpenSSL and cURL dependencies
* Fixed and improved installation procedure

**Release Notes**
N/A

### DIFF
--- a/android/Vagrantfile
+++ b/android/Vagrantfile
@@ -113,7 +113,7 @@ Vagrant.configure("2") do |config|
   config.vm.provision "shell", name: "Preparing desktop environment...", inline: <<-SHELL
       add-apt-repository ppa:lyzardking/ubuntu-make
       apt-get update
-      apt-get upgrade
+      apt-get dist-upgrade
       apt-get --assume-yes install ubuntu-mate-core virtualbox-guest-x11
       apt-get --assume-yes install ubuntu-make git automake libtool pkg-config
       update-locale LC_ALL=en_US.UTF-8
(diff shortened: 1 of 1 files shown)


android/Vagrantfile
```diff
@@ -137,11 +137,11 @@ Vagrant.configure("2") do |config|
   end
 
   config.vm.provision "shell" do |s|
-      s.name = "Clone BOINC repository and configuring toolchain..."
+      s.name = "Cloning BOINC repository and configuring toolchain..."
       s.privileged  = false
       s.inline  = <<-SHELL
-          export OPENSSL_VERSION=1.0.2p
-          export CURL_VERSION=7.61.0
+          export OPENSSL_VERSION=1.0.2q
```
AenBleidd: You should update android/buildAndroidBOINC-CI.sh with both OPENSSL_VERSION and CURL_VERSION you use here

brevilo: That's your file 😈

No, honestly, I wasn't aware of it and strictly speaking it's independent of vagrant. Ideally that would be just one file but I'm not sure how to do that... But ok, I'll quickly update those.

brevilo: Done

AenBleidd: Yes, it's mine but they are doind almost the same so it is nice to keep them both updated.
I have some thoughts how to keep these versions in one place but I need a little bit more time to convert these thoughts into code :smile:

PR #2860, opened 2018-11-30
Avoid project.inc including itself

lfield:
Include only the function we need to avoid pulling in other stuff.

**Description of the Change**
util.inc includes project.inc but only the function parse_config that is required and this can be found int util_basic.inc

### DIFF
--- a/html/project.sample/project.inc
+++ b/html/project.sample/project.inc
@@ -8,7 +8,7 @@
 // see inc/translation.inc and
 // https://boinc.berkeley.edu/trac/wiki/TranslateProject for details
 
-require_once("../inc/util.inc");
+require_once("../inc/util_basic.inc");
 
 //-------------- Project name and owner
 


TheAspens:
so project.inc makes calls to:

- In project_banner() it calls sample_navbar() which is from inc/bootstrap.inc
- In project_footer() it calls time_str() which is from inc/util.inc
- In auth_ops_example() it calls auth_ops_userid() and auth_ops_privilege() which is from inc/util_ops.inc
- In multiple functions it calls tra() which is defined in inc/translation.inc

I didn't finish checking the file, but the change as it is will cause issues with the execution of project.inc.  There needs to be a more to the change to prevent util.inc calling project.inc and vice versa.

TheAspens:
@lfield - can we close this pull request?  The proposed change will break a lot of other code as described above.

lfield:
I will close this as it is not so important and we can fix our own project.inc.

PR #2861, opened 2018-11-30
Alternative approach for server versioning

### LINKED ISSUE #2858: Duplication of versioning
The recent server versioning PR #2838 duplicates the functionality of [generate_svn_version.sh](https://github.com/BOINC/boinc/blob/master/generate_svn_version.sh). For example running `db_dump --version` on the server outputs:

> 2
> 2018-11-30 13:26:19.6759  db_dump starting
> 19a64836a416268ee8c9ca67fe3340efcbacbc0f [https://github.com/BOINC/boinc.git] (boincai05.cern.ch:/usr/local/boinc/software/boinc_src/boinc [(detached from server_release/1.0/1.0.1)])
The version should be stored on one place place and found by one mechanism. Thoughts?

lfield:
Fixes #2858 

**Description of the Change**

1. The version detection is now done in generate_svn_version.sh rather than setup_project.py. 
2. release.inc is created in html/inc and added to .gitignore
3. The link to the repository has been removed to avoid further discussion

### DIFF
--- a/generate_svn_version.sh
+++ b/generate_svn_version.sh
@@ -19,6 +19,7 @@ elif [ -d .git ]; then
     fi
     URL=`git config --get remote.$remote.url`
     DATE=`git log -n1 --pretty="format:%ct"`
+    SERVER_VERSION=`git describe --abbrev=0 --match server_release/*` 
 elif [ -d .svn ]; then
     CMD="svn info"
 else
@@ -60,4 +61,17 @@ if cmp "$HEADER" "$TMPFILE" >/dev/null 2>&1; then
 	rm -f "$TMPFILE"
 else
 	mv "$TMPFILE" "$HEADER"
+fi
+
+if [ ! -z ${SERVER_VERSION} ]; then
+    SERVER_VERSION=`echo ${SERVER_VERSION} | sed 's#.*/##'`
+    cat << EOF > html/inc/release.inc
+<?php
+
+global \$server_version ;
+\$server_version = "${SERVER_VERSION}";
+
+?>
+EOF
+
 fi
--- a/html/user/server_status.php
+++ b/html/user/server_status.php
@@ -30,8 +30,8 @@ require_once("../inc/cache.inc");
 require_once("../inc/util.inc");
 require_once("../inc/xml.inc");
 require_once("../inc/boinc_db.inc");
-if(file_exists('../../release.inc'))
-    include '../../release.inc';
+if(file_exists('../inc/release.inc'))
+    include '../inc/release.inc';
 
 if (!defined('STATUS_PAGE_TTL')) {
     define('STATUS_PAGE_TTL', 3600);
@@ -187,9 +187,8 @@ function show_status_html($x) {
     end_table();
     
     global $server_version;
-    global $git_commit;
-    if ( isset($server_version) && isset($git_commit) ) {
-        echo "Server version: $server_version (<a href=https://github.com/BOINC/boinc/commit/$git_commit>$git_commit</a>)<br>";
+    if ( isset($server_version) ) {
+        echo "Server version: $server_version <br>";
     }
 
     if ($j->db_revision) {
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,7 @@ py/Boinc/version.py
 py/setup.py
 stamp-h1
 test/version.inc
+html/inc/release.inc
 
 ## files created by make:
 *.o
--- a/html/user/get_project_config.php
+++ b/html/user/get_project_config.php
@@ -19,8 +19,8 @@
 require_once("../inc/consent.inc");
 require_once("../inc/util.inc");
 require_once("../inc/xml.inc");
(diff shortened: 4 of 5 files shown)


generate_svn_version.sh
```diff
@@ -19,6 +19,7 @@ elif [ -d .git ]; then
     fi
     URL=`git config --get remote.$remote.url`
     DATE=`git log -n1 --pretty="format:%ct"`
+    SERVER_VERSION=`git describe --abbrev=0 --match server_release/*`
```
brevilo: I think this should also be added to https://github.com/BOINC/boinc/blob/61d50f69ce140ce05eb4908e58b1db2d7eb12074/generate_svn_version.sh#L30, right at the end of `SVN_VERSION`, e.g. like this: `($HOST:$PWD [$BRANCH]) [Server-Release: $SERVER_VERSION]`

html/user/server_status.php
```diff
@@ -187,9 +187,8 @@ function show_status_html($x) {
     end_table();
     
     global $server_version;
-    global $git_commit;
-    if ( isset($server_version) && isset($git_commit) ) {
-        echo "Server version: $server_version (<a href=https://github.com/BOINC/boinc/commit/$git_commit>$git_commit</a>)<br>";
+    if ( isset($server_version) ) {
+        echo "Server version: $server_version <br>";
```
brevilo: Since you now added `--abbrev=0` this can't be considered the "Server version" anymore. It's rather "Based on server version:".

brevilo:
Way better. Now projects can decide when and how often to update those details.

html/user/server_status.php
lfield: I don't understand. Please can you explain.

generate_svn_version.sh
lfield: I have made this change

html/user/server_status.php
brevilo: See our long mail discussion: the `git describe` you now use just gives the *latest common* server release tag used by the server, it doesn't represent the server's version itself. Thus I recommend a more precise label for that version which really describes what it is.

lfield: ok so ...
```
lfield:~/boinc (server_release/1.0) $ git describe
server_release/1.0/1.0.0-3-g19a64836a4
```
and 

```
lfield:~/boinc (server_release/1.0) $ git describe --abbrev=0 
server_release/1.0/1.0.0
```
git log shows the following (some output suppressed)
```

commit 19a64836a416268ee8c9ca67fe3340efcbacbc0f (HEAD -> server_release/1.0, tag: server_release/1.0/1.0.1, origin/server_release/1/1.0)
...
commit cefcf545b825504571e080ff35578508907e679e
...
commit 771620daa7abe057a7eba030650086866a46c5c1
...
commit e72c05ace220190c82550e9f91745590fb2d8560 (tag: server_release/1.0/1.0.0)
```
So my guess is that using `git describe` without `--abbrev=0 `appends -3-g19a64836a4 where this is the number of commits since the tag and the last commit. So your argument is that as the branch may be few changes ahead if we made and local changes, we should use the words _based on_ ?

bema-aei: Actually, if you are restricting the server version to *official* (annotated) tags, a link to the official repository might be useful (`https://github.com/BOINC/boinc/releases/tag/server_release%2F1.0%2F1.0.1` for the tag or `https://github.com/BOINC/boinc/tree/server_release/1.0/1.0.1` directly to the code).

lfield: I think I would still use '_server version_'. After installation, files can get hot fixed, modified by configuration management systems or tweaked manually. If we do want to make this explicit, there is probably a better word to use like series.

lfield: ok. decided between the two of you if you want a link or not and let me know.

brevilo: @bema-aei wants to use a fixed URL to github/boinc, not a dynamically generated one from the actual repo's remote. That's fine with me, as long as you clearly refer to the `latest common server release tag` only, that is, use `--abbrev=0`.

brevilo: > I think I would still use 'server version'.

The reason I don't like this is that 'server version' is ambiguous. It doesn't tell the reader whether that refers to upstream BOINC or the project's deployment.

lfield: I have just done this. Please check. Can we agree on the text? Does it need to be changed from server version?

lfield: Can you think of anything better than based on? Upstream version? Powered by ? Built from?

brevilo: "Upstream version" and "Powered by" are both fine with me 👍

lfield: What about _server release_? This suggests something a little more general.

brevilo: Upstream server release?

lfield:
Upgraded LHC@home Dev. Seems to work fine. 
https://lhcathomedev.cern.ch/lhcathome-dev/server_status.php

PR #2862, opened 2018-12-01
[Android] Add work unit buffer to preference

### LINKED ISSUE #2755: [Android] Add "Store at least X days of work" preference
On Android the "Store at least X days of work" and the "Store up to an additional X days of work" preferences would be nice.

adamradocz:
Fixes #2755

**Description of the Change**
Add WU buffer option to the Preference menu.

**Release Notes**
[Android] Add "Work unit buffer" option

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/PrefsFragment.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/PrefsFragment.java
@@ -413,7 +413,7 @@ public class PrefsFragment extends Fragment {
          			   item.ID == R.string.prefs_disk_min_free_gb_header ||
          			   item.ID == R.string.prefs_disk_access_interval_header ||
          			   item.ID == R.string.prefs_other_store_at_least_x_days_of_work_header ||
-					   item.ID == R.string.prefs_other_store_up_to_an_additional_x_days_of_work_header) {
+         			   item.ID == R.string.prefs_other_store_up_to_an_additional_x_days_of_work_header) {
          		   EditText edit = dialog.findViewById(R.id.Input);
          		   String input = edit.getText().toString();
          		   Double valueTmp = parseInputValueToDouble(input);
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/PrefsFragment.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/PrefsFragment.java
@@ -245,7 +245,7 @@ public class PrefsFragment extends Fragment {
 				valueWrapper.ID == R.string.prefs_disk_max_pct_header || 
 				valueWrapper.ID == R.string.prefs_cpu_time_max_header ||
 				valueWrapper.ID == R.string.prefs_cpu_other_load_suspension_header || 
-				valueWrapper.ID == R.string.prefs_memory_max_idle_header ) {
+				valueWrapper.ID == R.string.prefs_memory_max_idle_header) {
 			Double seekBarDefault = valueWrapper.status / 10;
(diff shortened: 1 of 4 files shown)


adamradocz:
@AenBleidd could you help me out? I can't find the function that links the Android preferences with the global_prefs_override.xml. How can I serialize it?

adamradocz:
I think I got it.

adamradocz:
Only the "Store at least N days of work" added, the "Store up to an additional N days of work" not.
Is that necessary?

brevilo:
> Is that necessary?

Well, you requested it in the first place 😁 Since these are advanced options I personally don't care. Whatever gets implemented can be useful but there's no strict requirement. Bottom line: if you wan't to take it on it would certainly be appreciated and if you don't no one will blame you 👍

android/BOINC/app/src/main/java/edu/berkeley/boinc/PrefsFragment.java
```diff
@@ -407,7 +410,8 @@ public void onClick(View v) {
          	   } else if(item.ID == R.string.prefs_network_daily_xfer_limit_mb_header ||
          			   item.ID == R.string.battery_temperature_max_header ||
          			   item.ID == R.string.prefs_disk_min_free_gb_header ||
-         			   item.ID == R.string.prefs_disk_access_interval_header) {
+         			   item.ID == R.string.prefs_disk_access_interval_header ||
+					   item.ID == R.string.prefs_other_store_at_least_x_days_of_work_header) {
```
brevilo: Please remove the additional indentation (one level).

adamradocz: Done.

adamradocz: Done

adamradocz: Now should be fine.

brevilo:
I presume you tested your latest implementation?

brevilo:
Hm, any idea where or why that happens? While technically not a huge problem this is of course confusing to the user. So, if feasible, this should be fixed first.

adamradocz:
Unfortunately, I've no idea.

AenBleidd:
@you need to specify additional `PrefsListItemWrapperValue.Unit` type because default is integer, see:
https://github.com/BOINC/boinc/blob/9e18843a8dc69b6edc8da287ed57f1f9e5acfdf9/android/BOINC/app/src/main/java/edu/berkeley/boinc/adapter/PrefsListAdapter.java#L91-L93

adamradocz:
Thanks @AenBleidd.
@brevilo It's done.

android/BOINC/app/src/main/java/edu/berkeley/boinc/adapter/PrefsListAdapter.java
```diff
@@ -106,6 +108,9 @@ public View getView(int position, View convertView, ViewGroup parent) {
 	    				case GIGABYTES:
 	    					value = Formatter.formatShortFileSize(this.activity, (long)(item.status.doubleValue() * 0x40000000));
 	    					break;
+						case DAYS:
```
AenBleidd: Please fix indents here.
BTW, maybe it is better to add `UNIT` type `DECIMAL `as more generic than `DAYS`?

adamradocz: You're right, changed and fixed.

brevilo:
👍

PR #2865, opened 2018-12-01
[Android] Remove redundant castings

adamradocz:
**Description of the Change**
Micro-optimization. The redundant castings removed.

**Release Notes**
N/A


brevilo:
Hi, can you share some insight we these explicit downcasts have been used so far? I'd think that they were added on purpose by the original author. Why are they superfluous now?

Also, did you `grep` (search) the whole codebase for all instances of `findViewById`? Thus are there any calls left to checked for those casts or are you sure you changed all of them?

Cheers!

adamradocz:
I removed the leftover castings, I did search for `findViewById`.
It is an Android Studio recommendation, plus there are no visual benefits for using them.
Eg, The declaration shows clearly it's a button.
`Button returnB = (Button) dialog.findViewById(R.id.returnB);`

brevilo:
Thanks guys!

PR #2871, opened 2018-12-05
client: unsuspend jobs before telling them to quit

davidpanderson:
If a job is suspended (e.g. because of CPU throttling)
its timer thread is suspended and won't process any messages,
including quit messages.
This could cause a delay in client exit,
and possibly job processes that don't go away.

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/client/app_control.cpp
+++ b/client/app_control.cpp
@@ -194,6 +194,14 @@ static void print_descendants(int pid, vector<int>desc, const char* where) {
 // Send a quit message, start timer, get descendants
 //
 int ACTIVE_TASK::request_exit() {
+    // unsuspend the process.
+    // If it's suspended, the timer thread is suspended and
+    // won't process the quit message
+    //
+    if (task_state() == PROCESS_SUSPENDED) {
+        unsuspend();
+    }
+
     if (app_client_shm.shm) {
         process_control_queue.msg_queue_send(
             "<quit/>",


davidpanderson:
Can someone please merge this?

TheAspens:
Tested and works correctly.  Thanks for the contribution!

JacobWKlein:
The solution for this PR, is not correct.

On Windows 10 x64, with "Leave non-GPU tasks in memory while suspended" checked
... When exiting BOINC while it is suspended, previously in-progress jobs get resumed *for a full minute*, before finally stopping. They shouldn't run for a full minute when we're trying to do the exact opposite - EXIT.

I relayed my findings via the BOINC Alpha email list. David helped me confirm that reversion of this PR makes the client behave normally again. Logs confirmed that the PR solution is in fact leaving tasks running for a minute, before being killed by a timeout. Logs are attached as a text file, to this post.

I suggest that this solution needs to be excluded from public/test releases, to be reworked before being put into a future test release.

Thanks,
Jacob Klein

[20191001 BOINC Resuming Suspended Tasks on Exit.txt](https://github.com/BOINC/boinc/files/3676953/20191001.BOINC.Resuming.Suspended.Tasks.on.Exit.txt)

PR #2872, opened 2018-12-05
web: add ability to put extra stuff in <head> element of all pages

davidpanderson:
project.inc can include a constant HEAD_EXTRA.
Its contents are inserted at the beginning of <head> tags.

This makes it possible to use Google Analytics,
which requires that a script be added at that point.

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/html/inc/util.inc
+++ b/html/inc/util.inc
@@ -257,8 +257,8 @@ function page_head(
         <html lang="en">
         <head>
     ';
-    if (defined('HEAD_EXTRA')) {
-        echo HEAD_EXTRA;
+    if (defined('GLOBAL_HEAD_EXTRA')) {
+        echo GLOBAL_HEAD_EXTRA;
     }
     echo '
         <meta name="viewport" content="width=device-width, initial-scale=1">
--- a/html/inc/util.inc
+++ b/html/inc/util.inc
@@ -243,15 +243,24 @@ function page_head(
 
     if (!$caching) {
         header("Content-type: text/html; charset=utf-8");
-        header ("Expires: Mon, 26 Jul 1997 05:00:00 UTC");    // Date in the past
-        header ("Last-Modified: " . gmdate("D, d M Y H:i:s") . " UTC"); // always modified
-        header ("Cache-Control: $cache_control_extra no-cache, must-revalidate, post-check=0, pre-check=0");  // HTTP/1.1
-        header ("Pragma: no-cache");                          // HTTP/1.0
+        header("Expires: Mon, 26 Jul 1997 05:00:00 UTC");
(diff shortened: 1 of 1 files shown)


TheAspens:
The function page_head includes this parameter
```
    $head_extra=null
        // extra stuff to put in <head>. E.g.:
        // reCAPTCHA code (create_profile.php)
        // bbcode javascript (forums)
```
and in the code down lower there is your new code followed by the existing code
```
    if (defined('HEAD_EXTRA')) {
        echo HEAD_EXTRA;
    }
    echo '
        <meta name="viewport" content="width=device-width, initial-scale=1">
    ';
    if ($head_extra) {
        echo "\n$head_extra\n";
    }
```

based on the description of the function parameter, it exists to serve the same purpose as your new code.  It seems like there is overlap between your code and this existing mechanism.  Does the existing mechanism not meet your need?  Do we know if anyone uses the existing mechanism?  Is there some way that the two mechanisms can be combined so that there isn't any overlap?

davidpanderson:
Google analytics says their script must go at the start of the <head>

Rytiss:
The Google documentation (https://developers.google.com/analytics/devguides/collection/analyticsjs/) specifies "near the top of the <head> tag".

I've experienced zero difference in the performance of the tracking code based on tag locations - it just works. I usually put it as the last tag in the `<head>`.

I believe there is no need for an additional custom head section, especially considering that the new and old sections are only separated by a single meta tag, which does not load any assets and thus does not slow down the page load.

davidpanderson:
The main thing is, this puts it in every page,
without modifying every single page_head() call

TheAspens:
Ah - so the parameter on the function is for use on a "per page" basis while the variable you have defined is more of a "global" setting.

Is it worth calling it GLOBAL_HEAD_EXTRA to clarify the difference?  Let me know what you think and then I'll merge.

TheAspens:
(process note - since Rytiss voiced an concern - he needs to indicate he is ok with the solution as well before I merge)

Rytiss:
I'm ok with this.

davidpanderson:
Either name is fine.  Probably not EXTRA_HEAD though.

TheAspens:
I think GLOBAL_HEAD_EXTRA make a good distinction between the other mechanism and this and will make it easier for someone in the future reading the code and deciding how to use the feature.  If you could make that change then I will merge it.  Thanks!

TheAspens:
@davidpanderson - can you review my last request for a change?  I think it can be merged after that.  Thanks.

TheAspens:
Thanks for the update and contribution!

PR #2874, opened 2018-12-05
Add SupplementaryGroups=video to systemd

candrews:
**Description of the Change**

By including the video group, the boinc client can access video (GPU) hardware making boinc-client "just work" without the user having to perform additional steps.

All major distros (including Fedora, Gentoo, and Debian) have a video group.
**Alternate Designs**

I don't think there are any alternative designs.
**Release Notes**

N/A

### DIFF
--- a/client/scripts/boinc-client.service.in
+++ b/client/scripts/boinc-client.service.in
@@ -8,6 +8,7 @@ ProtectHome=true
 Type=simple
 Nice=10
 User=boinc
+SupplementaryGroups=video
 WorkingDirectory=/var/lib/boinc
 ExecStart=@exec_prefix@/bin/boinc
 ExecStop=@exec_prefix@/bin/boinccmd --quit


JuhaSointusalo:
Debian adds the boinc user to video group.

Could you list pros and cons between these two approaches?

candrews:
This change would mean that Debian (as well any other distro) wouldn't need to do that any more. Less distro specific work, less complexity, more upstream uniformity - those are all good things :-)

Germano0:
Concerning video group, please read the whole thread
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/HEXV3WJASEVAZV7P55KBA7RI7ELD7FNS/#FXS2MPBVWFQEQOR2WZKSXW7QJN2ERFAY
On Fedora / EPEL the long term plan is to implement this by using udev/ACLs on dri/render.

Yamakuzure:
> Video group is a dangerous thing for security, please read the whole thread
> https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/HEXV3WJASEVAZV7P55KBA7RI7ELD7FNS/#FXS2MPBVWFQEQOR2WZKSXW7QJN2ERFAY

I can not see the slightest hint on why video group is a dangerous thing for security in that thread, only some speculations what you _might_ be able to do from within an application that belongs to this group.
The noteworthy part is:

> Note sure what "boinc-client" does, but if this isn't turstworthy then
> it probably shouldn't be able to get access to "video".

If the boinc-client is not trustworthy, then we have a major issue here. Distributions add the client to the video group anyway, so I guess the argument is somewhat moot.

> On Fedora / EPEL the long term plan is to implement this by using udev/ACLs on dri/render.
So Fedora plans on patching boinc-client? How does this fit in with OpenCL/CUDA?

Yamakuzure:
> > I can not see the slightest hint on why video group is a dangerous thing for security
> 
> From [Arch Linux Users and groups wiki](https://wiki.archlinux.org/index.php/users_and_groups#Pre-systemd_groups):
> _Video group: Access to video capture devices, 2D/3D hardware acceleration, framebuffer._
> 
> So:
> 
> * If a BOINC working unit can access to framebuffer it can see what a user is doing on his screen.
> * If a BOINC working unit can access to video capture devices, it can see you through your webcam.
>   Moreover: _This way has been deprecated in favour of udev marking the devices with a uaccess tag_
>
> > If the boinc-client is not trustworthy, then we have a major issue here.
> 
> BOINC client != BOINC working unit, which is binary code that cannot be inspected so it must be treated as not trusted and it must be confined.

I see... so in the end it is not about the client at all, but the working units. I didn't expect them to be able to "reach through" in such a manner.

> 
> > Distributions add the client to the video group anyway
> 
> Does Client mean BOINC client? That is not true for Fedora / CentOS / Red Hat Enterprise Linux / Scientific Linux

Okay, "Some" (or even "most", I don't know) distributions.
So this way is considered "deprecated" in favour of uaccess. I see problems for people coming along, who do not which to run [e]udev on their (static) machines. And people who do not wish to use either systemd-logind or elogind.

So I duly hope this "new way of things" does not suddenly destroy the userspace for people, whose systems have been running fine for ages. That just wouldn't be nice.

> > So Fedora plans on patching boinc-client? How does this fit in with OpenCL/CUDA?
> 
> Once the work will be completed, BOINC client will no longer have problems in detecting using OpenCL / CUDA videocards.

That sounds promising! :+1:

Germano0:
> Do you have any kind of timetable for this? Earlier in the thread you said "long term plan" which could easily mean a few years.

Next week I will start studing how this should be implemented. I cannot provide any timetable

> Also, I'm curious, what this set up would roughly look like?

?

JuhaSointusalo:
CC: @LocutusOfBorg

Regarding the security issues. I tried to find a way to take a screenshot but wan't successful. There's plenty of tips on how to do that with standard Gnome/KDE/etc tools but those are of no use here. I did find a tool that grabs the contents of framebuffer device but on a physical machine the tool's output didn't really have anything at all and on a VM the tool crashed. I suspect the tool failed because X wasn't using framebuffer but DRI/DRM. I couldn't find a tool that would allow accessing display contents with DRI/DRM with a reasonable amount of searching.

Accessing webcamera is a different. I was able to create a science app that takes a snapshot of the user with a webcam and uploads it to the server. Making the science app was pretty easy.

Having GPU computing work out of the box would be super convenient. But I don't think we can decide on the behalf of our users that they want GPU computing so much that they are willing to risk their privacy. So I have to say thanks but no thanks to this pull request.

Until such time that we have something better available maybe package install scripts could ask the user if they would like to enable GPU computing explaining the benefits but also the risks.

Note that we have similar situation on Windows with the small difference that there is no simple hack that enables services to access GPUs. If GPU computing is wanted then BOINC has to be run with user's credentials under normal login session. I would hope that no-one expects any kind of security in that case.

Germano0:
@JuhaSointusalo thank you for having runned those tests.

> Until such time that we have something better available maybe package install scripts could ask the user if they would like to enable GPU computing explaining the benefits but also the risks.

As Fedora maintainer let me stress that (at least) on such distro (RHEL/CentOS should be the same) the user can use GPU without having to do anything, as I have seen from recent tests I have runned.

At the moment the only very specific use case that needs BOINC privileges to tweaked is when a person with a Radeon wants to use AMDGPU-Pro propertary stack without having to install it on the system. This can be useful on distros that are not supported by AMDGPU-Pro, like Fedora. You simply unpack the drivers somewhere and use the flag `Environment=LD_LIBRARY_PATH=foo_dir` BOINC in systemd unit file as I stated in Fedora devel discussion linked in https://github.com/BOINC/boinc/pull/2874#issuecomment-445046063
But let me say it again, it is a very specific use case

PR #2875, opened 2018-12-05
[Android] The setBackgroundDrawable was deprecated in API level 16.

adamradocz:
**Description of the Change**
The [setBackgroundDrawable ](https://developer.android.com/reference/android/view/View.html#setBackgroundDrawable(android.graphics.drawable.Drawable)) was deprecated in API level 16. Android.com says to use the setBackground instead which was added in API level 16.

We target minimum API 19.

**Release Notes**
N/A


AenBleidd:

PR #2876, opened 2018-12-05
[Android] Remove unnecessary SDK_INT checks

adamradocz:
**Description of the Change**
We target a minimum API 19 which is always bigger. The conditions are unnecessary.

**Release Notes**
N/A


AenBleidd:

PR #2878, opened 2018-12-06
[Android] Fix 'equals()' between objects of inconvertible types

adamradocz:
**Description of the Change**
The `match.size()` return a `long`, and it was compared with an `int`.
The number 1 now used as long.

Fixes the following inspected error:  '==' between objects of inconvertible types 'Integer' and 'long'

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/app/build.gradle
+++ b/android/BOINC/app/build.gradle
@@ -28,7 +28,7 @@ def buildVersionName() {
     def match = (tag =~ pattern)
 
     // Sanity checks for tag format
-    if(match.hasGroup() && 1L == match.size() && 5 == match[0].size() && head.commitDistance == 0) {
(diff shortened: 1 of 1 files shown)


android/BOINC/app/build.gradle
```diff
@@ -28,7 +28,7 @@ def buildVersionName() {
     def match = (tag =~ pattern)
 
     // Sanity checks for tag format
-    if(match.hasGroup() && 1 == match.size() && 5 == match[0].size() && head.commitDistance == 0) {
+    if(match.hasGroup() && 1L == match.size() && 5 == match[0].size() && head.commitDistance == 0) {
```
AenBleidd: Does `match[0].size()` returns int or long? Maybe it should be fixed too?

adamradocz: The inspection didn't show error for that, but seems reasonable.

AenBleidd:

PR #2884, opened 2018-12-07
[Android] Fix string performance warnings

adamradocz:
**Description of the Change**
Reports String concatenation used as the argument to StringBuffer.append(), StringBuilder.append() or Appendable.append(). Such calls may profitably be turned into chained append calls on the existing StringBuffer/Builder/Appendable, saving the cost of an extra StringBuffer/Builder allocation.

Reports String concatenation in loops. As every String concatenation copies the whole String, usually it is preferable to replace it with explicit calls to StringBuilder.append() or StringBuffer.append().

**Release Notes**
N/A


AenBleidd:

PR #2887, opened 2018-12-07
[Android] Update Charger Type attribute to use the API's constant

adamradocz:
**Description of the Change**
The previous solution used the fix number 4, because it was necessary prior API 17. Now we target minimum API 19, so updated to use the standard method.

**Release Notes**
N/A


AenBleidd:

PR #2889, opened 2018-12-07
[Android] Fix Java language level migration aids

adamradocz:
**Description of the Change**
Fixed the following code inspections in the `Java language level migration aids` category:

_'for' loop replaceable with 'foreach'_
Reports for loops which iterate over collections or arrays, and can be replaced with the foreach iteration syntax, available in Java 5 and newer.

_'StringBuffer' may be 'StringBuilder'_
Reports any variables declared as java.lang.StringBuffer which may be more efficiently declared as java.lang.StringBuilder.

_Unnecessary boxing_
Reports explicit boxing, i.e. wrapping of primitive values in objects. Explicit manual boxing is unnecessary under Java 5 and newer, and can be safely removed.

_Unnecessary unboxing_
Reports "unboxing", e.g. explicit unwrapping of wrapped primitive values. Unboxing is unnecessary under Java 5 and newer, and can be safely removed.

_Explicit type can be replaced with <>_
This inspection reports all new expressions with type arguments which can be replaced with diamond type <>

_'try' can use automatic resource management_
Reports try finally statements which can use Java 7 Automatic Resource Management. A quickfix is available to convert the try finally statement into a try with resources statement.

_Identical 'catch' branches in 'try' statement_
Reports identical catch sections in try blocks under JDK 7. A quickfix is available to collapse the sections into a multi-catch section.

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/adapter/PrefsListAdapter.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/adapter/PrefsListAdapter.java
@@ -103,10 +103,10 @@ public class PrefsListAdapter extends ArrayAdapter<PrefsListItemWrapper>{
 	    					value = NumberFormat.getInstance().format(item.status) + this.activity.getString(R.string.prefs_unit_celsius);
 	    					break;
 	    				case MEGABYTES:
-	    					value = Formatter.formatShortFileSize(this.activity, (long)(item.status.doubleValue() * 0x100000));
+	    					value = Formatter.formatShortFileSize(this.activity, (long)(item.status * 0x100000));
 	    					break;
 	    				case GIGABYTES:
-	    					value = Formatter.formatShortFileSize(this.activity, (long)(item.status.doubleValue() * 0x40000000));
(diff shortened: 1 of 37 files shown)


android/BOINC/app/src/main/java/edu/berkeley/boinc/adapter/PrefsListAdapter.java
```diff
@@ -103,10 +103,10 @@ public View getView(int position, View convertView, ViewGroup parent) {
 	    					value = NumberFormat.getInstance().format(item.status) + this.activity.getString(R.string.prefs_unit_celsius);
 	    					break;
 	    				case MEGABYTES:
-	    					value = Formatter.formatShortFileSize(this.activity, (long)(item.status.doubleValue() * 0x100000));
+	    					value = Formatter.formatShortFileSize(this.activity, (long)(item.status * 0x100000));
```
AenBleidd: @adamradocz, did you verify that this logic still produces valid numbers?

adamradocz: `item.status` returns `double` value by default.

AenBleidd:

PR #2891, opened 2018-12-07
[Android] Fix 'Obsolete Gradle Dependency' inspection.

adamradocz:
**Description of the Change**
Fix 'Obsolete Gradle Dependency' inspection. The new Gradle has newer dependency.

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/app/build.gradle
+++ b/android/BOINC/app/build.gradle
@@ -71,5 +71,5 @@ android {
 }
 
 dependencies {
-    implementation 'com.android.support:appcompat-v7:26.0.2'
+    implementation 'com.android.support:appcompat-v7:26.1.0'
 }


android/BOINC/app/build.gradle
```diff
@@ -71,5 +71,5 @@ android {
 }
 
 dependencies {
-    implementation 'com.android.support:appcompat-v7:26.0.2'
+    implementation 'com.android.support:appcompat-v7:26.1.0'
```
AenBleidd: @adamradocz, in #2896 you have the same change bu to higher version. I suppose this one can be just closed without merging

adamradocz: Yes, that issue closes this one. I created two different issues because this related to Android 8, while the other one bumps the version of Android to 9 too, so a little bigger change.

PR #2894, opened 2018-12-07
Drupal: Modified work_and_host_stats Feature.

drshawnkwang:
**Description of the Change**

As part of PHP7 upgrade, this Feature cannot be reverted. This attempts
to fix that by changing the Feature.

https://dev.gridrepublic.org/browse/DBOINCP-468


brevilo:
Are you sure the original code isn't required?

drshawnkwang:
After working on this some more, I must admit this is hack.

`$pages` for this Feature contain exactly one `$page`; meaning it doesn't need a key (`page_stats`). However, if one tries to add a second page to this Feature, then it would overwrite the first page- since there are no keys in the `$pages` array.

This means the **teams** Feature cannot be 'hacked' in this way since it contains three pages set in the `$pages` array.

So in retrospect I am reluctant to make this change; I'll close the PR for now and we can discuss things offline, or at our next telecon.

PR #2895, opened 2018-12-07
Drupal: Modified team_forums Feature.

drshawnkwang:
**Description of the Change**
**Description of the Change**
As part of PHP7 upgrade, this Feature cannot be reverted. This attempts to fix that by changing the Feature.

https://dev.gridrepublic.org/browse/DBOINCP-468


brevilo:
Are you sure the original code isn't required?

drshawnkwang:
See https://github.com/BOINC/boinc/pull/2894#issuecomment-445853359

PR #2896, opened 2018-12-07
[Android] Update target SDK to 28 (Android 9)

adamradocz:
Closes #2891

**Description of the Change**
Fix 'Target SDK attribute is not targeting latest version' inspection.

When your application runs on a version of Android that is more recent than your targetSdkVersion specifies that it has been tested with, various compatibility modes kick in. This ensures that your application continues to work, but it may look out of place. For example, if the targetSdkVersion is less than 14, your app may get an option button in the UI.  To fix this issue, set the targetSdkVersion to the highest available value. Then test your app to make sure everything works correctly. You may want to consult the compatibility notes to see what changes apply to each version you are adding support for: http://developer.android.com/reference/android/os/Build.VERSION_CODES.html as well as follow this guide: https://developer.android.com/distribute/best-practices/develop/target-sdk.html

**Release Notes**
Add Android 9 (Pie) compatibility.

### DIFF
--- a/android/BOINC/app/build.gradle
+++ b/android/BOINC/app/build.gradle
@@ -47,13 +47,13 @@ def buildVersionName() {
 }
 
 android {
-    compileSdkVersion 26
+    compileSdkVersion 28
     buildToolsVersion '28.0.3'
 
     defaultConfig {
         applicationId "edu.berkeley.boinc"
         minSdkVersion 19
-        targetSdkVersion 26
+        targetSdkVersion 28
         versionCode buildVersionCode()
         versionName buildVersionName()
     }
@@ -71,5 +71,5 @@ android {
 }
 
 dependencies {
-    implementation 'com.android.support:appcompat-v7:26.1.0'
+    implementation 'com.android.support:appcompat-v7:28.0.0'
 }
--- a/android/BOINC/app/build.gradle
+++ b/android/BOINC/app/build.gradle
@@ -71,5 +71,5 @@ android {
 }
 
 dependencies {
-    implementation 'com.android.support:appcompat-v7:26.0.2'
+    implementation 'com.android.support:appcompat-v7:26.1.0'
 }


brevilo:
The change itself is obviously trivial so I could sign-off on it in this sense. But I don't think I have access to an Android 9 device so I can't test this myself. How about Travis, does it build Android for different Android flavors? Maybe @AenBleidd or @ChristianBeer can answer this?

Raising the `compileSdkVersion` and `targetSdkVersion` doesn't affect the minimum supported Android version, correct?

adamradocz:
@brevilo Correct, `compileSdkVersion` and `targetSdkVersion` doesn't affect the minimum supported Android version.
I use this release about for 3 days now, on my device which runs Android 9.

brevilo:
Too bad that the target API guide doesn't yet have a section for "Earlier than 28" to describe where to put our attention when coming from API level 26 (not from below it). 👎

AenBleidd:
@brevilo, sorry, I missed your question. No, it doesn't. 
But if we want it could be possible to get several builds for several platforms. This will make our CI slower but maybe it is really necessary to do to help other people test Android builds.
Question: what if we create a separate branch (e.g. Android) to push all reviewed Android tickets there and make test builds of merged stuff there? Then we can just merge that branch to master when all changes will be verified.
I think, it could be a good question on a call today.
Agree?

brevilo:
If that slow down would affect *all* other CI builds, then yes. OTOH, the same could be said about the client builds for various platforms, right? So that's a general question I think. Yet I'm not sure that having various "staging" branches and thus increased complexity outweighs the build time increase.

AenBleidd:
Separate branch could allow to make weekly builds with several changes included rather than every other change separately. Of course, this way also have both negative and positive sides. @brevilo, will you attend today call?

brevilo:
No, sorry, too many local issues :-/

PR #2898, opened 2018-12-07
[Android] Fix misused 'continue' statements.

adamradocz:
**Description of the Change**
The `continue` statement has been used wrongly a couple of times. I changed to `break` or deleted them, but it changes the logic of the loops, so please check thoroughly.

**Alternate Designs**
Some `for + break` loop could have changed to `while`.

**Release Notes**
[Android] Minor improvement.

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/PrefsFragment.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/PrefsFragment.java
@@ -219,7 +219,6 @@ public class PrefsFragment extends Fragment {
 		for (PrefsListItemWrapper item: data) {
 			if(item.ID == ID){
 				((PrefsListItemWrapperBool) item).setStatus(newValue);
-				continue;
 			}
 		}
 	}
@@ -231,7 +230,6 @@ public class PrefsFragment extends Fragment {
 		for (PrefsListItemWrapper item: data) {
 			if(item.ID == ID){
 				((PrefsListItemWrapperValue) item).status = newValue;
-				continue;
 			}
 		}
 	}
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
@@ -158,7 +158,7 @@ public class ProjectsFragment extends Fragment {
 		for(int x = 0; x < data.size(); x++) {
 			if(data.get(x).isMgr) {
 				mgrIndex = x;
-				continue;
+				break;
 			}
 		}
 		if(mgrIndex < 0) { // no manager present until now
@@ -185,7 +185,7 @@ public class ProjectsFragment extends Fragment {
 			for(int x = 0; x < data.size(); x++) {
 				if(rpcResult.master_url.equals(data.get(x).id)) {
 					index = x;
-					continue;
+					break;
 				}
 			}
 			if(index == null) { // result is new, add
@@ -207,7 +207,7 @@ public class ProjectsFragment extends Fragment {
 			for(Project rpcResult: latestRpcProjectsList) {
 				if(listItem.id.equals(rpcResult.master_url)) {
 					found = true;
-					continue;
+					break;
 				}
 			}
 			if(!found) iData.remove();
@@ -225,7 +225,6 @@ public class ProjectsFragment extends Fragment {
 						project.addServerNotice(serverNotice);
 						noticeFound = true;
 						mappedServerNotices++;
-						continue;
 					}
 				}
 				if(!noticeFound) project.addServerNotice(null);
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
@@ -158,46 +158,46 @@ public class ProjectsFragment extends Fragment {
 		for(int x = 0; x < data.size(); x++) {
 			if(data.get(x).isMgr) {
 				mgrIndex = x;
-				break;
+				//break; // This function has to be revised. Are we searching the firs or the last account manager? Are there more or only one possible?
 			}
 		}
 		if(mgrIndex < 0) { // no manager present until now
-			if(Logging.VERBOSE) Log.d(Logging.TAG,"no manager found in layout list. new entry available: " + acctMgrInfo.present);
+			if(Logging.VERBOSE) Log.d(Logging.TAG,"No manager found in layout list. New entry available: " + acctMgrInfo.present);
 			if(acctMgrInfo.present) {
 				// add new manager entry, at top of the list
 				data.add(new ProjectsListData(null,acctMgrInfo,null));
-				if(Logging.DEBUG) Log.d(Logging.TAG,"new acct mgr found: " + acctMgrInfo.acct_mgr_name);
+				if(Logging.DEBUG) Log.d(Logging.TAG,"New acct mgr found: " + acctMgrInfo.acct_mgr_name);
 			}
 		} else { // manager found in existing list
-			if(Logging.VERBOSE) Log.d(Logging.TAG,"manager found in layout list at index: " + mgrIndex);
+			if(Logging.VERBOSE) Log.d(Logging.TAG,"Manager found in layout list at index: " + mgrIndex);
 			if(!acctMgrInfo.present) {
 				// manager got detached, remove from list
 				data.remove(mgrIndex);
-				if(Logging.DEBUG) Log.d(Logging.TAG,"acct mgr removed from list.");
+				if(Logging.DEBUG) Log.d(Logging.TAG,"Acct mgr removed from list.");
 			}
 		}
 		
 	// ATTACHED PROJECTS	
-		//loop through all received Result items to add new results
+		//loop through all received Result items to add new projects
 		for(Project rpcResult: latestRpcProjectsList) {
-			//check whether this Result is new
+			//check whether this project is new
 			Integer index = null;
 			for(int x = 0; x < data.size(); x++) {
 				if(rpcResult.master_url.equals(data.get(x).id)) {
 					index = x;
-					break;
+					//break; // Need more further investigation.
 				}
 			}
-			if(index == null) { // result is new, add
-				if(Logging.DEBUG) Log.d(Logging.TAG,"new result found, id: " + rpcResult.master_url + ", managed: " + rpcResult.attached_via_acct_mgr);
+			if(index == null) { // Project is new, add
+				if(Logging.DEBUG) Log.d(Logging.TAG,"New project found, id: " + rpcResult.master_url + ", managed: " + rpcResult.attached_via_acct_mgr);
 				if(rpcResult.attached_via_acct_mgr) data.add(new ProjectsListData(rpcResult,null, mapTransfersToProject(rpcResult.master_url, ongoingTransfers))); // append to end of list (after manager)
 				else data.add(0, new ProjectsListData(rpcResult,null, mapTransfersToProject(rpcResult.master_url, ongoingTransfers))); // put at top of list (before manager)
-			} else { // result was present before, update its data
+			} else { // Project was present before, update its data
 				data.get(index).updateProjectData(rpcResult,null,mapTransfersToProject(rpcResult.master_url, ongoingTransfers));
 			}
 		}
 		
-		//loop through the list adapter to find removed (ready/aborted) Results
+		//loop through the list adapter to find removed (ready/aborted) projects
 		// use iterator to safely remove while iterating
 		Iterator<ProjectsListData> iData = data.iterator();
 		while(iData.hasNext()) {
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/TasksFragment.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/TasksFragment.java
@@ -122,7 +122,7 @@ public class TasksFragment extends Fragment {
 			for(int x = 0; x < data.size(); x++) {
 				if(rpcResult.name.equals(data.get(x).id)) {
 					index = x;
-					continue;
+					break;
 				}
 			}
 			if(index == null) { // result is new, add
@@ -142,7 +142,7 @@ public class TasksFragment extends Fragment {
 			for(Result rpcResult: newData) {
 				if(listItem.id.equals(rpcResult.name)) {
 					found = true;
-					continue;
+					break;
 				}
 			}
 			if(!found) iData.remove();
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/ClientInterfaceImplementation.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/ClientInterfaceImplementation.java
@@ -184,18 +184,16 @@ public class ClientInterfaceImplementation extends RpcClient{
 	 */
 	
 	public Boolean checkProjectAttached(String url) {
-		Boolean match = false;
 		try{
 			ArrayList<Project> attachedProjects = getProjectStatus();
 			for (Project project: attachedProjects) {
 				if(Logging.DEBUG) Log.d(Logging.TAG, project.master_url + " vs " + url);
 				if(project.master_url.equals(url)) {
-					match = true;
-					continue;
+					return true;
 				}
 			}
 		} catch(Exception e){}
-		return match;
+		return false;
 	}
 	
 	/**
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/ClientStatus.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/ClientStatus.java
@@ -589,7 +589,7 @@ public class ClientStatus {
 					for(Result task: results) {
 						if(task.active_task) { // this result has corresponding "active task" in RPC XML
 							activeTask = true;
-							continue; // amount of active tasks does not matter.
+							break; // amount of active tasks does not matter.
 						}
 					}
 				}
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
@@ -823,14 +823,14 @@ public class Monitor extends Service {
     	for (int x = 0; x < headers.length; x++) {
     		if(headers[x].equals("PID")) {
     			PidIndex = x;
-    			continue;
+    			break;
     		}
     	}
 		if(Logging.DEBUG) Log.d(Logging.TAG,"getPidForProcessName(): PID at index: " + PidIndex + " for output: " + processLinesAr[0]);
     	
 		Integer pid = null;
+		Boolean found = false;
     	for(int y = 1; y < processLinesAr.length; y++) {
-    		Boolean found = false;
     	    String [] comps = processLinesAr[y].split("[\\s]+");
     	    for(String arg: comps) {
     	    	if(arg.equals(processName)) {
@@ -843,7 +843,7 @@ public class Monitor extends Service {
 	    	    	pid = Integer.parseInt(comps[PidIndex]);
 	        	    if(Logging.ERROR) Log.d(Logging.TAG,"getPidForProcessName(): pid: " + pid); 
 	    	    }catch (NumberFormatException e) {if(Logging.ERROR) Log.e(Logging.TAG,"getPidForProcessName(): NumberFormatException for " + comps[PidIndex] + " at index: " + PidIndex);}
-	    	    continue;
+	    	    break;
     	    }
     	}
     	// if not happen in ps output, not running?!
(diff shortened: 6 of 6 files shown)


adamradocz:
`/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java`

```
// figure out what index PID has
     	String [] headers = processLinesAr[0].split("[\\s]+");
     	Integer PidIndex = 1;
    	for (int x = 0; x < headers.length; x++) {
    		if(headers[x].equals("PID")) {
    			PidIndex = x;
    			break;
    		}
    	}
```
I don't understand why Integer PidIndex gets an initial 1 value.
Isn't that should be less than 0 or null, because we pass x to PidIndex and x can be 0 and 1 too. Or the for should start from 2, because the first two loop is unnecessary.

Hope I was clear :)

adamradocz:
`/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java`

Line 832:
```
for(int y = 1; y < processLinesAr.length; y++) {
    		Boolean found = false;
    	    Boolean found = false;
    	    String [] comps = processLinesAr[y].split("[\\s]+");
    	    for(String arg: comps) {
    	    	if(arg.equals(processName)) {
     	    		if(Logging.DEBUG) Log.d(Logging.TAG,"getPidForProcessName(): " + processName + " found in line: " + y);
     	    		found = true;
     	    	}
     	    }
     	    if(found) {
 	    	    try{
	    	    	pid = Integer.parseInt(comps[PidIndex]);
	        	    if(Logging.ERROR) Log.d(Logging.TAG,"getPidForProcessName(): pid: " + pid); 
	    	    }catch (NumberFormatException e) {if(Logging.ERROR) Log.e(Logging.TAG,"getPidForProcessName(): NumberFormatException for " + comps[PidIndex] + " at index: " + PidIndex);}
	    	    break;
    	    }
    	}
```

I replaced the `continue` statement to `break` after the `catch` tag, but I think it was intended in the `foreach` loop under the `found = true;`. Am I right?

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
```diff
@@ -158,7 +158,7 @@ private void updateData(ArrayList<Project> latestRpcProjectsList, AcctMgrInfo ac
 		for(int x = 0; x < data.size(); x++) {
 			if(data.get(x).isMgr) {
 				mgrIndex = x;
-				continue;
+				break;
```
brevilo: The question here is: can there be more than one AM entry? If so, which one should be recorded, the last (original) or the first (your version)? Given the context I tend to accept your change.

android/BOINC/app/src/main/java/edu/berkeley/boinc/PrefsFragment.java
```diff
@@ -219,7 +219,6 @@ private void updateBoolPref(int ID, Boolean newValue) {
 		for (PrefsListItemWrapper item: data) {
 			if(item.ID == ID){
 				((PrefsListItemWrapperBool) item).setStatus(newValue);
-				continue;
```
brevilo: Ok.

android/BOINC/app/src/main/java/edu/berkeley/boinc/PrefsFragment.java
```diff
@@ -231,7 +230,6 @@ private void updateValuePref(int ID, Double newValue) {
 		for (PrefsListItemWrapper item: data) {
 			if(item.ID == ID){
 				((PrefsListItemWrapperValue) item).status = newValue;
-				continue;
```
brevilo: Ok.

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
```diff
@@ -185,7 +185,7 @@ private void updateData(ArrayList<Project> latestRpcProjectsList, AcctMgrInfo ac
 			for(int x = 0; x < data.size(); x++) {
 				if(rpcResult.master_url.equals(data.get(x).id)) {
 					index = x;
-					continue;
+					break;
 				}
 			}
 			if(index == null) { // result is new, add
```
brevilo: I think all the comment/logging references to "result" or wrong here and likely a copy 'n' paste issue. Shouldn't all of those occurrences refer to "project" instead??? This doesn't really help understanding the original code's purpose.

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
```diff
@@ -185,7 +185,7 @@ private void updateData(ArrayList<Project> latestRpcProjectsList, AcctMgrInfo ac
 			for(int x = 0; x < data.size(); x++) {
 				if(rpcResult.master_url.equals(data.get(x).id)) {
 					index = x;
-					continue;
+					break;
```
brevilo: Should be ok as well as finding a single match seems to be the intention here.

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
```diff
@@ -207,7 +207,7 @@ private void updateData(ArrayList<Project> latestRpcProjectsList, AcctMgrInfo ac
 			for(Project rpcResult: latestRpcProjectsList) {
 				if(listItem.id.equals(rpcResult.master_url)) {
 					found = true;
-					continue;
+					break;
```
brevilo: ok

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
```diff
@@ -225,7 +225,6 @@ private void updateData(ArrayList<Project> latestRpcProjectsList, AcctMgrInfo ac
 						project.addServerNotice(serverNotice);
 						noticeFound = true;
 						mappedServerNotices++;
-						continue;
```
brevilo: Ok.

android/BOINC/app/src/main/java/edu/berkeley/boinc/TasksFragment.java
```diff
@@ -122,7 +122,7 @@ private void updateData(ArrayList<Result> newData) {
 			for(int x = 0; x < data.size(); x++) {
 				if(rpcResult.name.equals(data.get(x).id)) {
 					index = x;
-					continue;
+					break;
```
brevilo: Ok.

android/BOINC/app/src/main/java/edu/berkeley/boinc/TasksFragment.java
```diff
@@ -142,7 +142,7 @@ private void updateData(ArrayList<Result> newData) {
 			for(Result rpcResult: newData) {
 				if(listItem.id.equals(rpcResult.name)) {
 					found = true;
-					continue;
+					break;
```
brevilo: Ok.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/ClientInterfaceImplementation.java
```diff
@@ -184,18 +184,16 @@ public Boolean attachProject(String url, String projectName, String authenticato
...
 	public Boolean checkProjectAttached(String url) {
-		Boolean match = false;
 		try{
 			ArrayList<Project> attachedProjects = getProjectStatus();
 			for (Project project: attachedProjects) {
 				if(Logging.DEBUG) Log.d(Logging.TAG, project.master_url + " vs " + url);
 				if(project.master_url.equals(url)) {
-					match = true;
-					continue;
+					return true;
```
brevilo: Ok (and all related changes)

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/ClientStatus.java
```diff
@@ -589,7 +589,7 @@ private void parseComputingStatus() {
 					for(Result task: results) {
 						if(task.active_task) { // this result has corresponding "active task" in RPC XML
 							activeTask = true;
-							continue; // amount of active tasks does not matter.
+							break; // amount of active tasks does not matter.
```
brevilo: Certainly OK (as the comment says!)

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
```diff
@@ -823,27 +823,27 @@ private Integer getPidForProcessName(String processName) {
...
 		Integer pid = null;
     	for(int y = 1; y < processLinesAr.length; y++) {
-    		Boolean found = false;
+    	    Boolean found = false;
     	    String [] comps = processLinesAr[y].split("[\\s]+");
     	    for(String arg: comps) {
     	    	if(arg.equals(processName)) {
     	    		if(Logging.DEBUG) Log.d(Logging.TAG,"getPidForProcessName(): " + processName + " found in line: " + y);
     	    		found = true;
+                    break;
```
brevilo: Please use the same indentation level or adjust the previous two lines as well. Otherwise ok.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
```diff
@@ -823,27 +823,27 @@ private Integer getPidForProcessName(String processName) {
...
     	    	}
     	    }
     	    if(found) {
 	    	    try{
 	    	    	pid = Integer.parseInt(comps[PidIndex]);
 	        	    if(Logging.ERROR) Log.d(Logging.TAG,"getPidForProcessName(): pid: " + pid); 
 	    	    }catch (NumberFormatException e) {if(Logging.ERROR) Log.e(Logging.TAG,"getPidForProcessName(): NumberFormatException for " + comps[PidIndex] + " at index: " + PidIndex);}
-	    	    continue;
```
brevilo: Ok.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
```diff
@@ -823,27 +823,27 @@ private Integer getPidForProcessName(String processName) {
     	for (int x = 0; x < headers.length; x++) {
     		if(headers[x].equals("PID")) {
     			PidIndex = x;
-    			continue;
+    			break;
```
brevilo: Ok.

brevilo:
> I don't understand why `Integer PidIndex` gets an initial `1` value.

That doesn't make sense indeed. So, yes, I think it should be initialized with `-1`.
And it's worse than just that: the following debug log statement as well as the next code block don't check if the `PID` column was found at all. Lines 829 to 848 should be wrapped in a conditional that executes `if(PidIndex != -1)`.

brevilo:
> I replaced the `continue` statement to `break` after the `catch` tag, but I think it was intended in the `foreach` loop under the `found = true;`. Am I right?
> 
> **Update:** I moved the `break` from the `if` to the `foreach` statement. If OK, I'll fix the indentation.

Sorry, I'm not what exactly is being referred to here: the snippet your posted or the change you committed? Please clarify.

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
adamradocz: It looks fine to me.

adamradocz: Seems fine to me.

adamradocz:
> > I don't understand why `Integer PidIndex` gets an initial `1` value.
> 
> That doesn't make sense indeed. So, yes, I think it should be initialized with `-1`.
> And it's worse than just that: the following debug log statement as well as the next code block don't check if the `PID` column was found at all. Lines 829 to 848 should be wrapped in a conditional that executes `if(PidIndex != -1)`.

Lines between 829 and 850 should be wrapped in a conditional statement.

adamradocz:
> > I replaced the `continue` statement to `break` after the `catch` tag, but I think it was intended in the `foreach` loop under the `found = true;`. Am I right?
> > **Update:** I moved the `break` from the `if` to the `foreach` statement. If OK, I'll fix the indentation.
> 
> Sorry, I'm not sure what exactly is being referred to here: the snippet you posted or the change you committed? Please clarify.
> 
> I think the `break` you [added](https://github.com/BOINC/boinc/pull/2898/files/facb70ee9887e88ebfa003563823a498d04a6d1c#diff-1b828896f8d964b096d9187399c8e0e8R839) to the for-each is as justified as all the others you added elsewhere (here: stop when matching process found). But I also think that it shouldn't just replace the `continue` you [removed](https://github.com/BOINC/boinc/pull/2898/files/facb70ee9887e88ebfa003563823a498d04a6d1c#diff-1b828896f8d964b096d9187399c8e0e8L846). That instead should also be replaced by a `break` to stop the search when the process' PID was found.
I added back the 2nd `break`, but honestly I don't like the two `break`s here. It makes the `for` loop unreadable, but I will leave it for now.

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
brevilo: How? This code section doesn't deal with results at all but with projects. Just to be clear, this not about your change, it's an additional change that would be good to get in, since you're at it anyway.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
```diff
@@ -819,35 +819,48 @@ private Integer getPidForProcessName(String processName) {
...
-		if(pid == null) if(Logging.ERROR) Log.d(Logging.TAG,"getPidForProcessName(): " + processName + " not found in ps output!");
+		for (int y = 1; y < processLinesAr.length; y++) {
+			Boolean found = false;
+			String[] comps = processLinesAr[y].split("[\\s]+");
+			for (String arg : comps) {
+				if (arg.equals(processName)) {
+					if (Logging.DEBUG)
+						Log.d(Logging.TAG, "getPidForProcessName(): " + processName + " found in line: " + y);
+					found = true;
+					break;
```
brevilo: Since you find it confusing, just add a comment like "break out of inner for-each (comps) loop"

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
```diff
@@ -819,35 +819,48 @@ private Integer getPidForProcessName(String processName) {
...
+			if (found) {
+				try {
+					pid = Integer.parseInt(comps[PidIndex]);
+					if (Logging.ERROR)
+						Log.d(Logging.TAG, "getPidForProcessName(): pid: " + pid);
+				} catch (NumberFormatException e) {
+					if (Logging.ERROR)
+						Log.e(Logging.TAG, "getPidForProcessName(): NumberFormatException for " + comps[PidIndex] + " at index: " + PidIndex);
+
+					break;
```
brevilo: Since you find it confusing, just add a comment like "break out of outer for (processLinesAr.length) loop"

brevilo: Also, I think this `break` should go after the `try/catch` block as the outer for loop should be exited after the process was found, regardless of whether the PID could be extracted or not.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
```diff
@@ -819,35 +819,48 @@ private Integer getPidForProcessName(String processName) {
...
     	for (int x = 0; x < headers.length; x++) {
     		if(headers[x].equals("PID")) {
     			PidIndex = x;
     			break;
     		}
     	}
-		if(Logging.DEBUG) Log.d(Logging.TAG,"getPidForProcessName(): PID at index: " + PidIndex + " for output: " + processLinesAr[0]);
-    	
+
+		if (PidIndex == -1)
```
brevilo: Please review the indentation of all lines between 829 and 864. At least GitHub shows things are really off right now. Thanks.

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
AenBleidd: I think changing `continue` to `break` will break the logic here. I'd rather just remove `continue` here.
BTW, I don't fully understand the purpose of this function at all.
In case of mane AMs present in the list current implementation will overwrite the last AM entry.
New proposed implementation will overwrite first only.
I think more detailed investigation is necessary here to understand the purpose and desired behavior of this function

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
AenBleidd: Same as above. Need more info

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
AenBleidd: I agree with @brevilo. I believe comments and log messages should be changed from `result` to `project`

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
```diff
@@ -819,35 +819,48 @@ private Integer getPidForProcessName(String processName) {
...
+			return null;
```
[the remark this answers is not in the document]
adamradocz: Fixed.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
adamradocz: Fixed.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
adamradocz: True. Fixed.

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
adamradocz: I disagree. The comment above says, "Check whether this Result is new". If we pass the `index` a value, the `if (index == null)` is already false, so we don't need to find more `X`s.

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
adamradocz: Fixed.

adamradocz:
Done.

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
AenBleidd: @adamradocz, I agree with you but the open question for me now it why this function is implemented this way. Maybe the current behavior is incorrect at whole but changing the logic without full understanding of the situation in worse than just leaving it as is for a while.
Please understand me correctly, I also want to make this product better, as all of us I believe.
And thank you for your hard job!

brevilo: I think the whole codebase shows that the original author had no clear understanding of what `continue` means. So we have to take that into account while trying to determine the **originally intended** functionality.

adamradocz: I made the changes that @AenBleidd proposed.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
```diff
@@ -827,38 +827,35 @@ private Integer getPidForProcessName(String processName) {
...
+    			}
+    		}
+    		if (found) {
+    			try {
+    				pid = Integer.parseInt(comps[PidIndex]);
+    				if (Logging.ERROR) Log.d(Logging.TAG, "getPidForProcessName(): pid: " + pid);
+    			} catch (NumberFormatException e) {
+    				if (Logging.ERROR) Log.e(Logging.TAG, "getPidForProcessName(): NumberFormatException for " + comps[PidIndex] + " at index: " + PidIndex);
+    			}
+    			break;// Break out of inner for (processLinesAr) loop
```
brevilo: This breaks out of the *outer* loop

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
brevilo: Ok, so the decision is to not optimize this right now, but to flag this as an open question (providing how we think it should be written). FTR: removing the `continue` doesn't change the semantics.

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectsFragment.java
brevilo: Same as above: removing `continue` doesn't change the semantics and the proposed solution is visible, yet disabled until further clarification.

brevilo:
Waiting for CI to finish. Then I'll merge...

PR #2906, opened 2018-12-11
web: add options for not collecting or showing URL and country

davidpanderson:
Add constants NO_USER_URL and NO_USER_COUNTRY.
If you set these to true (in project.inc)
there will be no query or display of user web site URL and user country, respectively.

### DIFF
--- a/html/inc/account.inc
+++ b/html/inc/account.inc
@@ -102,7 +102,7 @@ function create_account_form($teamid, $next_url) {
         ),
         "passwd", "", "password",'id="passwd"',passwd_visible_checkbox("passwd")
     );
-    if (!NO_USER_COUNTRY) {
+    if (USER_COUNTRY) {
         form_select(
             sprintf('<span title="%s">%s</span>',
                 tra("Select the country you want to represent, if any."),
--- a/html/inc/account.inc
+++ b/html/inc/account.inc
@@ -102,14 +102,16 @@ function create_account_form($teamid, $next_url) {
         ),
         "passwd", "", "password",'id="passwd"',passwd_visible_checkbox("passwd")
     );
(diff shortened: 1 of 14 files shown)


TheAspens:
@davidpanderson - I'm starting to test this.  One request though.  The two new flags you create are 'negative' flags.  NO_USER_URL and NO_USER_COUNTRY which makes the logic in the if statements reads as `if ( !NO_USER_COUNTRY ) then ...... `

Could you use USER_COUNTRY and USER_URL instead so that the logic will read `if ( USER_COUNTRY) then....` which is easier to read.

TheAspens:
Another question, shouldn't db_dump be changed also since it exports user country and user url as well?  If it does, should the config option be moved into the project config.xml file?

davidpanderson:
I made it negative so that projects won't see a change in behavior.

TheAspens:
You can preserve the existing behavior by changing this (from boinc/util.inc)
```
if (!defined('NO_USER_COUNTRY')) {
    define('NO_USER_COUNTRY', 0);
}
if (!defined('NO_USER_URL')) {
    define('NO_USER_URL', 0);
}
```
to
```
if (!defined('USER_COUNTRY')) {
    define('USER_COUNTRY', 1);
}
if (!defined('USER_URL')) {
    define('USER_URL', 1);
}
```

Also - what about db_purge?

TheAspens:
sorry - not db_purge - I meant db_dump.l

JuhaSointusalo:
Now that we have server releases it is ok to make changes that require projects to take action. All you need to do is make sure those changes are properly documented in release notes.

TheAspens:
Looks good thanks for the changes!

PR #2909, opened 2018-12-13
[MINGW] Added include windows.h before winhttp.h when building with mingw

vkbo:
**Description of the Change**
If building with MINGW, include `windows.h` a bit earlier in `boinc_win.h`. Otherwise the build fails on datatypes in winhttp.h.

This has been an issue for a while when building SixTrack for LHC@home on Windows. Previously we've just fixed it with a bash script.

**Release Notes**
Include `windows.h` before `winhttp.h` when building on Windows/MINGW

### DIFF
--- a/lib/boinc_win.h
+++ b/lib/boinc_win.h
@@ -146,11 +146,6 @@
 #define SECURITY_WIN32
 #endif
 
-/* MINGW needs windows.h before winhttp.h */
-#ifdef __MINGW32__
-#include <windows.h>
-#endif
-
 #if !defined(__CYGWIN32__) || defined(USE_WINSOCK)
 
 /* If we're not running under CYGWIN use windows networking */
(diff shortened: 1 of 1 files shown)


vkbo:
I did not add a `#ifndef __MINGW32__` at the regular include point of `windows.h`, maybe that is cleaner?

vkbo:
Moved `winhttp.h` instead, but not sure which `#ifdef` you're referring to here. I'm now moving it outside the `#ifdef` it was sitting in though.

JuhaSointusalo:
> which `#ifdef` you're referring to here

The one guarding `winhttp.h`, `#ifdef HAVE_WINHTTP_H`. But that's ok, the file needs clean up and better organising anyway. That can be done later. I'm just waiting for CI builds to finish before merging.

JuhaSointusalo:
Thanks.

PR #2911, opened 2018-12-15
Enforcing idle CPU scheduling policy

Germano0:
**Description of the Change**
Pull request https://github.com/BOINC/boinc/pull/1878
changed CPU scheduling policy from SCHED_BATCH to SCHED_IDLE. This pull request enforces such configuration by adding a flag in BOINC systemd unit file

**Release Notes**
Enforcing idle CPU scheduling policy in BOINC systemd unit file

Issue #2859, opened 2018-11-30
[Linux] Use Scheduler Policy rather than Niceness


truboxl:
**Describe the problem**
Using niceness 19 puts BOINC tasks on idle but can cause UI tasks (GNOME, KDE, Firefox, even the BOINC manager) lagging/choppy on Linux machines. It is more noticeable when using 100% CPUs, 100% of CPU time. This does not happen on Windows.

**Describe the solution you'd like**
Use `SCHED_IDLEPRIO` (preferably) or `SCHED_BATCH` will improve the situation much better and on par with Windows ~~performance and~~ responsiveness.
~~Bonus: May solve the problem why the benchmark score on Linux is always lower compared to on Windows.~~

**Additional context**
`SCHED_BATCH` is available since mainline Linux 2.6.16 (20 March 2006)
`SCHED_IDLEPRIO` requires patches but recent distros (eg: Fedora 29) support it

From `man schedtool`:
>...
>POLICY OVERVIEW
>...
>SCHED_BATCH  [  since 2.6.16 in mainline ] SCHED_BATCH was designed for non-interactive, CPU-bound applications.  It uses longer timeslices (to better exploit the cache), but can be interrupted anytime by other processes in other classes to guaratee interaction  of  the  system.  Processes in this class are selected last but may result in a considerable speed-up (up to 300%). No interactive boosting is done.
>...
>SCHED_IDLEPRIO  [  patch  needed  ]  SCHED_IDLEPRIO is similar to SCHED_BATCH, but was explicitely designed to consume only the time the CPU is idle. No interactive boosting is done.  If you used SCHED_BATCH in the -ck kernels this is what you want since 2.6.16
>...

You can use the `schedtool` to run BOINC under these policies (not sure about other ways), eg:
for SCHED_IDLEPRIO: `schedtool -D -e boinc`
for SCHED_BATCH: `schedtool -B -e boinc`

EDIT: Cross-out irrelevant info

AenBleidd:
There was a pr not so long ago: #1878 to switch from SCHED_BACH to SCHED_IDLE
Looks like it need to be reviewed one more time

Germano0:
> There was a pr not so long ago: #1878 to switch from SCHED_BACH to SCHED_IDLE
> Looks like it need to be reviewed one more time

`lfield merged commit ac7ac37 into BOINC:master on 19 Oct`
so it will likely be available on next BOINC client release. The 7.14.2 stable release still contain the old code.
[...]

PR #2913, opened 2018-12-17
[Linux] change nice level to 19

Germano0:
**Description of the Change**
In past, In BOINC systemd unit file, nice level has been setted to 10, because in [this comment](https://github.com/BOINC/boinc/pull/2260#issuecomment-360619802) 

> Normally the client run GPU apps, wrappers and non-compute intensive apps at nice level 10. If you run the client at nice level 19 that won't be possible.

I runned some tests on Fedora 29 with systemd unit file flag nice=19 and:
- all processes related to BOINC (working units, wrappers, VirtualBox virtual machines,etc.) run at same nice=19 level;
- GPU calculus works fine;
- working units that use VirtualBox (like Theory Simulation 263.90 (vbox64_mt_mcore)) works fine.
I will also continue the testing in next days.

**Release Notes**
[Linux] changed nice level to 19

### DIFF
--- a/client/scripts/boinc-client.service.in
+++ b/client/scripts/boinc-client.service.in
@@ -6,7 +6,7 @@ After=network-online.target
 [Service]
 ProtectHome=true
 Type=simple
-Nice=10
+Nice=19
 User=boinc
 WorkingDirectory=/var/lib/boinc
 ExecStart=@exec_prefix@/bin/boinc


JuhaSointusalo:
> > Normally the client run GPU apps, wrappers and non-compute intensive apps at nice level 10. If you run the client at nice level 19 that won't be possible.
> 
> I runned some tests on Fedora 29 with systemd unit file flag nice=19 and:
> 
> * all processes related to BOINC (working units, wrappers, VirtualBox virtual machines,etc.) run at same nice=19 level;

Let's extend my earlier comment with a few more words:

> Normally the client run GPU apps, wrappers and non-compute intensive apps at nice level 10 _if it's allowed to do so_. If you run the client at nice level 19 that won't be possible _because the client is not allowed be lower niceness (only root can do that)_.

I would prefer for the controlling programs to have higher priority than the science apps so that they are ensured to have control over the science apps. vboxwrapper is particularly picky about it getting to poke VMs when it wants to and not a moment later.

Besides, the client doesn't use any more CPU time than other background apps. It doesn't need to run at nice=19.

TheAspens:
I agree with @JuhaSointusalo  - the controlling programs are not cpu intensive and it is important that they get their time slice to do their job when they need it.  They should be left at nice level of 10.

Has someone observed an issue with these running at nice level 10?  What is the motivation for wanting to lower it?

Ferroin:
> Has someone observed an issue with these running at nice level 10? What is the motivation for wanting to lower it?

Not sure that this is the motivation, but somewhat paradoxically, increasing the nice level of a process (for example, going from 10 to 19) under the default scheduler on Linux can actually _improve_ the latency under certain circumstances (namely, if the process mostly just needs short infrequent bursts of CPU time).  It's a side effect of how the scheduler decides what to run next, how it sorts the runqueue, and how Linux actually implements nice levels (it doesn't do priority scheduling with them, but instead scales the time-slice that the process is given, nicer processes get shorter time slices).

truboxl:
On Windows, I don't see there's the need for this mess of manual scheduling.

If the user really need to nice 19 on boinc client, can't they use `lower_client_priority` in cc_config.xml or I have mistaken its purpose?

On the other hand why not put the client at nice 0 same as Windows?

The way of manually nice-ing anything on Linux just seem to jeopardize more of the system stability and UI responsiveness, and it will be better if we look at alternatives like the Scheduling Policies instead #2859 ...

Germano0:
You are right, this pull request is useless. I left a more detailed comment in my other related pull request: 
https://github.com/BOINC/boinc/pull/2911#issuecomment-452273979

PR #2916, opened 2018-12-19
web: add consent preferences to column view of project preferences page.

### LINKED ISSUE #2857: GDPR not showing in alternate view
**Describe the bug**
The GDPR options are not showing in alternative view.

**Steps To Reproduce**
https://lhcathomedev.cern.ch/lhcathome-dev/prefs.php?subset=project&cols=1

**Expected behavior**
https://lhcathomedev.cern.ch/lhcathome-dev/prefs.php?subset=project&cols=0

drshawnkwang:
Fixes #2857

**Description of the Change**
Project Preferences column view now has the consent preferences displayed.

**Release Notes**
Include consent preferences in the Project Preferences column view.

### DIFF
--- a/html/inc/prefs_project.inc
+++ b/html/inc/prefs_project.inc
@@ -303,6 +303,7 @@ function print_prefs_display_project($user, $columns=false) {
         start_table();
         row_heading(tra("Combined preferences").$switch_link);
         prefs_show_privacy($user, true);
+        prefs_show_consent($user, true);
         venue_show($user);
         row_top(tra("Project specific settings"));
         prefs_show_project($project_prefs, true);
--- a/html/inc/prefs_util.inc
+++ b/html/inc/prefs_util.inc
@@ -198,6 +198,9 @@ class PREF_CONSENT extends PREF {
         } else {
             echo "<td>---</td>";
         }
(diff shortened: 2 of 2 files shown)


Rytiss:
Having looked through the code but not having ran it, I believe that while this shows the pref in the column view, it mismatches the number of table cells: `consent_show_cols()` (/html/inc/prefs_util.inc#L190) has the global `$venues` but does not use it, so it will only generate a single column, whereas other prefs such as privacy prefs will generate as many columns as there are venues.

Now, consent is a global, venue-spanning option, so it makes sense for it to not be repeated for each venue. However, so as not to break the table layout, `colspan` param could be applied to the value cell, matching the venue count, in both `show_value()` and the hardcoded table cell with dashes in `consent_show_cols()`.

`venue_show()` method also does not generate extra columns, so it is visible how it behaves differently. It should probably be fixed too.

drshawnkwang:
@Rytiss, the latest comment should fix what you are asking about. I attached a screenshot of the table generated by the newest code. Please evaluate if this is what you are looking for.

boincpr2916_table01

Rytiss:
Looks good, merged.

PR #2918, opened 2018-12-21
Client: fix job scheduling bug

davidpanderson:
- There was a scenario (# 164 in fact) where CPUs were starved
because CPU weren't being added to the run list.
The basic problem was the the max_concurrent stuff was being
called in make_run_list().
It doesn't belong there - only in enforce_run_list().

- add the ability to handle app_config.xml files in the client emulator.

- fix a performance bug that caused extremely long run lists;
in make_run_list(), check for exclusion at the project level, not global.

- do max_concurrent logic only if a max_concurrent rule was given.

- fix bug where the emulator would assign the wrong
version number to results, then fail to find their app version.

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**


RichardHaselgrove:
I've recreated the environment which led to the opening of #1677, on a much simpler machine than Keith's - 4 cores, 2 NVidia GPUs, Intel GPU. The Intel GPU is constrained to run Einstein only, and I don't think it needs to concern us here.

SETI can run on both NVidia and CPU. For testing, it has a max_concurrent (for the application, not the project) of 6

With the stock BOINC v7.14.2 client, <cpu_sched_debug> shows the same symptoms we've been discussing:

[cpu_sched_debug_v7-14-2.txt](https://github.com/BOINC/boinc/files/2703616/cpu_sched_debug_v7-14-2.txt)

Although 6 SETI tasks are selected as runnable, all are for GPU, and only two can run. Replacing the boinc client with the AppVeyor build for 4a7bb390af525f8d814213854a255c361a135a34, and making no other changes, SETI CPU tasks are chosen and scheduled.

[cpu_sched_debug_4a7bb390.txt](https://github.com/BOINC/boinc/files/2703625/cpu_sched_debug_4a7bb390.txt)

So, functionally, this PR cures the bug it was designed to address. Thanks, David, and well done. I'll leave it to more experienced coders to perform the formal review: I'll keep this version running, and try tweaking the configuration to ensure nothing unexpected turns up, but at this stage I would have no objections to merging.

RichardHaselgrove:
Unfortunately, I spoke too soon. When the running SETI/CPU task finished, no replacement was started - not even one of the part-processed SETI tasks 'waiting to run'. Here's the cpu_sched_debug: you'll see at the end that a second max_concurrent has come into play.

[cpu_sched_debug_4a7bb390_1.txt](https://github.com/BOINC/boinc/files/2703829/cpu_sched_debug_4a7bb390_1.txt)

I'll pop the files into the emulator, but I don't think I can upload two app_config files at once? The second, for NumberFields@home, is very simple:

```
<app_config>
   <project_max_concurrent>1</project_max_concurrent>
</app_config>
```
The target, at this point in testing, is to run:

SETI on two NV GPUs, with full-core support for the OpenCL code
SETI on one CPU core
NumberFields on one CPU core
Einstein on Intel GPU only

(I have cached work from other, unrelated, testing to finish off)

RichardHaselgrove:
Created [Scenario 166](https://boinc.berkeley.edu/sim_web.php?action=show_scenario&name=166), simulation 0

This is the same configuration as scenario 165, but by utilising the second app_config, the timeline now shows the same behaviour as I'm seeing locally: one CPU core starved by the NumberFields max_concurrent=1, although SETI CPU tasks are available.

RichardHaselgrove:
New AppVeyor build at aed33a47 loaded and running OK. On initial startup, seems to have solved the preceding problems, but I'll throw some extra configurations at it in the course of the next couple of days. The new feature to report max_concurrent settings at startup is helpful - immediately caught a typo!

I thought this might be a difficult one, but I didn't think it would be quite this hard - thanks for all the work.

RichardHaselgrove:
After nearly 24 hours running, I haven't come across any problems yet (scrutiny continues), but I have two comments/queries.

1) Compared with previous builds, I see a vastly increased incidence of NO_WORK_MSG displays: both as notices, and as high priority (red letter) lines in the Event Log. With work fetch kept reasonably modest, I'm getting five alerts, from three projects, every hour (after every project RPC to the affected projects).

That seems overkill. In my case, a blocked resource isn't a "problem" to be "fixed": it's a "preference" which could be "changed". I'd prefer a less strident notification level, perhaps only once after each client restart.

2) The commit note

> - work fetch: if a project is at a max concurrent limit, don't fetch work from it.

is unclear. I do have one project running at `project_max_concurrent` level, and it's fetching work normally: but the point isn't whether it's running **now** at `project_max_concurrent` level. The point is, does it have enough work cached to continue running at that level through the next anticipated running period (as set by the user), but without risking any deadline misses. That makes it more an RR_SIM test, not a work fetch test. If that's what was intended by the commit note, I'm happy with it, but it's not how it reads.

RichardHaselgrove:
Yes, that's better. The notices are there at startup, which is when you do need to check that you made your changes properly: once you're into steady running, you only want to be alerted to significant changes.

I think we're nearly ready for review and merge - I'll try to get Keith to give it a run. We may need some help with a Linux build at 9f8f52b7824164091828bc586189771971b399d5

RichardHaselgrove:
I'm still concerned about starvation by work fetch blocking with max_concurrent in operation. I've created [Scenario 167](https://boinc.berkeley.edu/sim_web.php?action=show_scenario&name=167), but the bare bones are:

I'm examining NumberFields@home. All projects except this are on NNT so I can play with cache sizes.

I had (when I started testing) 63,420 seconds 'remaining to run' across 8 tasks - one running, two pre-empted by setting max_concurrent, and five ready to start. I set work fetch to 1 day (86,400 seconds), so I was in shortfall even for the single running CPU. The lines I'm worried about are:

31/12/2018 19:33:55 |  | [work_fetch] ------- start work fetch state -------
31/12/2018 19:33:55 |  | [work_fetch] target work buffer: 86400.00 + 4320.00 sec
31/12/2018 19:33:55 | NumberFields@home | [work_fetch] REC 2638.046 prio -1.000 can't request work: at max_concurrent limit
31/12/2018 19:33:55 |  | [work_fetch] --- state for CPU ---
31/12/2018 19:33:55 |  | [work_fetch] shortfall 321471.08 nidle 1.40 saturated 0.00 busy 0.00
31/12/2018 19:33:55 | NumberFields@home | [work_fetch] share 0.000

I think that work fetch should have scaled back the buffer to 86,400, resulting in a shortfall of 22,980 seconds - and been allowed to fetch that much. Observations?

(The timeline in the scenario shows the current task running until 04:11:00 - that's 3 hours 11 minutes from start at 01:00:00. That's the local estimated for a full unstarted task, so doesn't take account of the 75% progress, 45 minutes remaining, at the time I uploaded the files. Not critical, but confusing)

KeithMyers:
I tried the [https://bintray.com/boinc/boinc-ci/pull-requests/PR2918_2018-12-31_9f8f52b7#files](https://bintray.com/boinc/boinc-ci/pull-requests/PR2918_2018-12-31_9f8f52b7#files) client and manager.  It did not work.  My app_config <project_max_concurrent>16</project_max_concurrent> statement was not being enacted after the file was read.  All 24 cpu threads were engaged.

davidpanderson:
The work fetch policy includes: don't fetch work from a project that's at a max_concurrent limit.  That's because the new jobs wouldn't be able to run, the shortfall would remain the same, and we'd fetch unbounded work.

KeithMyers:
I wasn't talking about fetching any work.  I was just stating that the app_config project max concurrent statement was not enforced.  It always has worked until the gpu_exclude statements were added to the cc_config.

RichardHaselgrove:
> The work fetch policy includes: don't fetch work from a project that's at a max_concurrent limit. That's because the new jobs wouldn't be able to run, the shortfall would remain the same, and we'd fetch unbounded work.

I'm sorry, but I still don't understand either your comment or the reasoning behind it.

I'll read your code as best I can to identify exactly what test is used to determine "at a max_concurrent limit". But even at a limit of 1, which is what I was testing yesterday, work is still being processed, completed and returned. Eventually it will run out.

Even with a max_concurrent in place, steady-state processing should require that the supply of work on hand matches the user's "Store at least xxx days of work", expressed in wall-clock units. That's not unbounded. It will be less than the amount required if multiple resources are occupied, sure. But I was below that target level yesterday, and I feel that work fetch should have been restricted proportionately, rather than blocked entirely.

RichardHaselgrove:
OK, understood. Presumably the client will drop out of "at a max_concurrent limit" at some point - Keith and I can monitor that and advise accordingly later on.

RichardHaselgrove:
I've now observed exactly how the max_concurrent buffer limit operates in practice.

If max_concurrent=1, buffer limitation applies until the very last cached task is completed:

```
02/01/2019 06:02:41 | NumberFields@home | Computation for task wu_sf5_DS-13x12_Grp27912of2000000_0 finished
02/01/2019 06:02:41 | NumberFields@home | Starting task wu_sf5_DS-13x12_Grp30813of2000000_0
02/01/2019 07:03:20 | NumberFields@home | Sending scheduler request: To report completed tasks.
02/01/2019 07:03:20 | NumberFields@home | Reporting 1 completed tasks
02/01/2019 07:03:20 | NumberFields@home | Not requesting tasks: at max_concurrent limit
02/01/2019 10:11:13 | NumberFields@home | Computation for task wu_sf5_DS-13x12_Grp30813of2000000_0 finished
02/01/2019 10:11:13 | NumberFields@home | Sending scheduler request: To fetch work.
02/01/2019 10:11:13 | NumberFields@home | [sched_op] CPU work request: 56138.84 seconds; 4.00 devices
02/01/2019 10:11:16 | NumberFields@home | Scheduler request completed: got 6 new tasks
02/01/2019 10:11:19 | NumberFields@home | Finished download of sf5_DS-13x12_Grp78068of2000000.dat
02/01/2019 10:11:19 | NumberFields@home | Starting task wu_sf5_DS-13x12_Grp78068of2000000_0
```
Providing that doesn't happen during maintenance, the interruption to work flow is minimal.

If max_concurrent>=2, buffer limitation is not applied, and work fetch continues as normal:

```
02/01/2019 14:21:10 | NumberFields@home | Computation for task wu_sf5_DS-13x12_Grp78497of2000000_0 finished
02/01/2019 14:36:46 | NumberFields@home | Computation for task wu_sf5_DS-13x12_Grp78046of2000000_0 finished
02/01/2019 15:01:39 | NumberFields@home | Sending scheduler request: To fetch work.
02/01/2019 15:01:39 | NumberFields@home | [sched_op] CPU work request: 17621.29 seconds; 1.00 devices
```
I think that we can live with that pretty much indefinitely, now that we know what's happening and we can document it for users.

The request for 56138.84 seconds of work with the cache completely empty surprised me: my work demand is for 0.25 + 0.05 days - 25,920 seconds, or 103,680 seconds for all four devices together. I declare (through app_config) that 1.6 avg_ncpus are reserved for GPU support: I think that the work request is scaled to accommodate this. Which is what led me to muse on more accurate CPU support estimates in #2949.

KeithMyers:
So would changing my GPUGrid project app_config max_concurrent statement to something other than 1 for both long and short tasks allow the cpu tasks to be run?  That is the only project I have max_concurrent set to 1. No cpu tasks run when I have a <project_max_concurrent>16</project_max_concurrent> statement in app_config for Seti.

RichardHaselgrove:
As we discussed in private messages last night (not currently accessible due to SETI maintenance), we would need to get the current 9f8f52b7824164091828bc586189771971b399d5 test code into a working condition on your particular flavour of Linux (you should only need to compile the client - your existing Manager should be adequate without change). Then we can examine logs or whatever else is needed to determine exactly what is happening, and whether there are any undesired side effects. The latter stages of the discussion here are about whether new work is **fetched** when max_concurrent is set to 1, not about whether it is **run**. But it runs for me.

KeithMyers:
I have compiled a client from the aforementioned branch and it appears to be functional.  It is classified weirdly compared to what I am used to though.  It is a shared library (application/x-sharedlib) instead of an (application/x-executable) like my previous clients and managers.

I attached my 7.8.3 boincmgr to it and it connects.  I am replicating all the projects that I am normally attached to with the same app_configs and cc_config files that cause the cpu tasks to not run on the RTX 2080 host.  But I am running it on a different host that has the experimental Ubuntu 18.10 partition.  Even though this host only has 16 threads instead of 24 and only 3 gpus instead of 4, it appears to run cpu tasks with the gpu:0 excluded from Einstein and GPUGrid projects. 

And it is obeying my <project_max_concurrent>11</project_max_concurrent> Seti app_config statement.

So I think it will work over on the real host that has that problem that the original bug was affecting.  That will be the true test I believe.  But I think the current test system is probably close enough to proclaim the #2918 commit good.

KeithMyers:
In the later stages of my test with the client code, I had finished off all my Seti work.  But I still had work for MilkyWay, Einstein and GPUGrid.  With that work available I was only running on 2 out of 3 gpus for some reason. The gpu:1 card was never used.  Each project I mentioned can run a maximum of two tasks so should have been computing on all 3 cards.  All 3 cards were available in the Event Log startup.  Never saw any reason mentioned the Event Log why one card was not being used.  Even with coprocessor_debug, the output never explained why.  I only had a statement that 2 cpus out of 16 were being used.

KeithMyers:
I just tried the client built off the https://github.com/BOINC/boinc/commit/9f8f52b7824164091828bc586189771971b399d5[](url) commit that Richard wanted me to use.

It has major issues.  Seems the Seti <project_max_concurrent>16</project_max_concurrent> statement is interpreted as the Seti cache is only allowed 16 tasks.

It never requests work because it believes the maximum tasks have been exceeded.  It never updates after a 305 second scheduler countdown.  The only way to get the client to report finished work is to do a manual update.

I have provided further information in the Questions & Answers forum thread.
[https://setiathome.berkeley.edu/forum_thread.php?id=83645&postid=1974416](url)

davidpanderson:
Not fetching work if max_concurrent is reached is the intended behavior.

RichardHaselgrove:
> I have provided further information in the Questions & Answers forum thread.

Reading through [Not requesting tasks_ at max_concurrent limit.txt](https://github.com/BOINC/boinc/files/2746156/Not.requesting.tasks_.at.max_concurrent.limit.txt) from that post, I see

[cpu_sched_debug] skipping [task name]; max concurrent limit 0 reached

for tasks 17, 18, 19 in the preliminary and final job lists. The code suggests (and I agree) that this should read

max concurrent limit 16 reached

reflecting the user's settings.

RichardHaselgrove:
@davidpanderson wrote:
> I'm going to stop here with this PR. Maintaining a full buffer in the presence of max_concurrent would require nontrivial changes to the RR simulation which I might do later.

I think we've reached that point. Apart from that minor suggestion about the debug message, both @KeithMyers and I find that the client is scheduling tasks as required.

But I do hope that you will be able to reconsider the new design decision on work fetch, and replace "don't fetch" with "fetch proportionately". I'll go into the reasons in more details at tonight's contributors' call - I hope you'll be able to join us.

client/client_state.h
```diff
@@ -62,130 +62,12 @@ using std::vector;
...
-    // if a project is uploading,
-    // and the last upload started within this interval,
-    // don't fetch work from it.
-    // This allows the work fetch to be merged with the reporting of the
-    // jobs that are currently uploading.
-
-#define RESULT_REPORT_IF_AT_LEAST_N 64
-    // If a project has at least this many ready-to-report tasks, report them.
-
 #define WF_EST_FETCH_TIME 180
```
JuhaSointusalo: I guess `WF_EST_FETCH_TIME` should go with the rest of the constants around line 570.

client/work_fetch.cpp
```diff
@@ -325,30 +322,13 @@ void RSC_WORK_FETCH::clear_request() {
...
-
 void PROJECT_WORK_FETCH::reset(PROJECT* p) {
     for (int i=0; i<coprocs.n_rsc; i++) {
         p->rsc_pwf[i].reset();
     }
 }
 
-void PROJECT_WORK_FETCH::rr_init(PROJECT* p) {
-    project_reason = compute_project_reason(p);
+void PROJECT_WORK_FETCH::rr_init(PROJECT*) {
```
JuhaSointusalo: I guess you could have just removed the parameter altogether.

client/work_fetch.cpp
```diff
@@ -144,20 +138,21 @@ int RSC_PROJECT_WORK_FETCH::compute_rsc_project_reason(
         if (n_runnable_jobs >= n_not_excluded
             && queue_est > (gstate.work_buf_min() * n_not_excluded)/rwf.ninstances
         ) {
-            return DONT_FETCH_BUFFER_FULL;
+            printf("%s: setting reason to BUFFER_FULL\n", p->project_name);
```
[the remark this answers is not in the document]
RichardHaselgrove: Please refer to https://setiathome.berkeley.edu/forum_thread.php?id=83645&postid=1973170#1973170 - looks like we have cause and effect here.

JuhaSointusalo:
> `[cpu_sched_debug] skipping [task name]; max concurrent limit 0 reached`

should be fixed with this, though I couldn't really test it because I didn't get enough tasks in the `run_list`:

```diff
diff --git a/client/cpu_sched.cpp b/client/cpu_sched.cpp
index ac9eea70e8..aa7421a15f 100644
--- a/client/cpu_sched.cpp
+++ b/client/cpu_sched.cpp
@@ -1237,9 +1237,21 @@ bool CLIENT_STATE::enforce_run_list(vector<RESULT*>& run_list) {
 
         if (have_max_concurrent && max_concurrent_exceeded(rp)) {
             if (log_flags.cpu_sched_debug) {
+                const char* type = "unknown";
+                int limit = -1;
+                switch (max_concurrent_exceeded(rp)) {
+                case CONCURRENT_LIMIT_APP:
+                    type = "app";
+                    limit = rp->app->max_concurrent;
+                    break;
+                case CONCURRENT_LIMIT_PROJECT:
+                    type = "project";
+                    limit = rp->project->app_configs.project_max_concurrent;
+                    break;
+                }
                 msg_printf(rp->project, MSG_INFO,
-                    "[cpu_sched_debug] skipping %s; max concurrent limit %d reached",
-                    rp->name, rp->app->max_concurrent
+                    "[cpu_sched_debug] skipping %s; %s max concurrent limit %d reached",
+                    rp->name, type, limit
                 );
             }
             continue;

```

client/work_fetch.cpp
```diff
@@ -144,20 +138,23 @@ int RSC_PROJECT_WORK_FETCH::compute_rsc_project_reason(
         if (n_runnable_jobs >= n_not_excluded
             && queue_est > (gstate.work_buf_min() * n_not_excluded)/rwf.ninstances
         ) {
-            return DONT_FETCH_BUFFER_FULL;
+            if (log_flags.work_fetch_debug) {
+                printf("%s: setting reason to BUFFER_FULL\n", p->project_name);
```
JuhaSointusalo: Erm...

```suggestion
                msg_printf(p, MSG_INFO, "[work_fetch] setting reason to BUFFER_FULL");
```

JuhaSointusalo:
@KeithMyers

> In the later stages of my test with the client code, I had finished off all my Seti work. But I still had work for MilkyWay, Einstein and GPUGrid. With that work available I was only running on 2 out of 3 gpus for some reason. The gpu:1 card was never used. Each project I mentioned can run a maximum of two tasks so should have been computing on all 3 cards

I don't think you ever followed up with that. If you think there is a bug in the code (instead of client configuration) you need to provide something to work with, like a simulation run David can debug.

TheAspens:
@davidpanderson @JuhaSointusalo - what are the next steps with this pull request?

davidpanderson:
ready to go as far as I'm concerned

RichardHaselgrove:
Artifact win-client_PR2918_2019-01-28_ae90be52 is still showing

28/01/2019 14:39:29 | NumberFields@home | [cpu_sched_debug] skipping wu_sf5_DS-13x12_Grp1398092of2000000_0; max concurrent limit 0 reached

(should be 1, in my test case)

davidpanderson:
fixed

RichardHaselgrove:
Um. I think I'd call simply removing the number 'suppressed', rather 'fixed', but at least it isn't false. @JuhaSointusalo - could you take another look, please? Even incomplete, this might be enough to let us make progress.

client/client_state.h
JuhaSointusalo: @davidpanderson

Please move the `#define` and the comment following it. It got relocated in the merge commit.

client/cs_statefile.cpp
```diff
@@ -366,20 +366,25 @@ int CLIENT_STATE::parse_state_file_aux(const char* fname) {
...
             }
+#endif
             rp->avp = lookup_app_version(
                 rp->wup->app, rp->platform, rp->version_num, rp->plan_class
             );
             if (!rp->avp) {
                 msg_printf(project, MSG_INTERNAL_ERROR,
-                    "No application found for task: %s %d %s; discarding",
-                    rp->platform, rp->version_num, rp->plan_class
+                    "No application found for task %s: platform %s version %d plan class%s; discarding",
```
JuhaSointusalo: Needs space after "plan class".

JuhaSointusalo:
@davidpanderson

Please take care of this

https://github.com/BOINC/boinc/blob/d9bb933d7988593b4d8f6fbb60fdd18cb4d22cfe/client/client_state.h#L65-L69

and this

https://github.com/BOINC/boinc/blob/d9bb933d7988593b4d8f6fbb60fdd18cb4d22cfe/client/cs_statefile.cpp#L386

davidpanderson:
the #define is there because it's used in an inline function.

someone merge this; I don't want to have to do it myself.

ChristianBeer:
It looks strange that all defines except one were moved to the bottom of the file and the only one left at the top of the file is there because it is used in the middle of the file. Why not keep all of the defines at the top of the file?

JuhaSointusalo:
> the #define is there because it's used in an inline function

Well, you could have said that sooner. I had not noticed the inline fuction.

JuhaSointusalo:
Even though I merged this I don't consider the job done yet. As previously discussed in my opinion the work fetch code now needs improving.

- If work fetch is blocked as long as there are max_concurrent or more tasks in buffer then before work fetch resumes there has to be an idle device.
- Those who want to use max_concurrent will now have to choose between using max_concurrent and having work buffered. My gut feeling is that people will choose work buffer and that means they won't be able to upgrade BOINC beyond 7.14 as long as the work fetch works the way it does now.

KeithMyers:
I concur with this assessment.  I won't be using any of the new branches because of the merge commits.
As Juha states, you won't be able to maintain the standard work cache buffer if you use a max_concurrent statement and you won't be using all of your devices.

I will continue to use the configurations that work for me now and which may change in the future once Einstein and GPUGrid properly support Turing architecture and I can get rid of my exclude_gpu statements.

PR #2921, opened 2018-12-23
Travis Bintray: enable cleanup of bintray repositories

ChristianBeer:
**Description of the Change**
Switch from date attribute to number attribute in order to get versions older than a specified date. This is a limitation of the API.
The cleanup_bintray.sh script needs to be run by the BOINC bintray maintainer. See comments in file on how to use this.

**Alternate Designs**
None

**Release Notes**
N/A

### DIFF
--- a/deploy/cleanup_bintray.sh
+++ b/deploy/cleanup_bintray.sh
@@ -1,0 +1,145 @@
+#!/bin/bash
+
+# This file is part of BOINC.
+# http://boinc.berkeley.edu
+# Copyright (C) 2018 University of California
+#
+# BOINC is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License
+# as published by the Free Software Foundation,
+# either version 3 of the License, or (at your option) any later version.
+#
+# BOINC is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


AenBleidd:
Can't we use client_id/client_secret instead to use server2server scenario?

ChristianBeer:
I don't understand what you mean.

AenBleidd:
Sorry for not clear message.
I mean you said that unauthorized access has limitations. But is it possible to use client_id/client secret pair instead of OATH key like it said here: https://developer.github.com/v3/#oauth2-keysecret ? And in the script we can use encrypted client_secret like Travis/AppVeyor secrets?

ChristianBeer:
The github rate limit is only a problem if the bintray maintainer doesn't run the script in a regular interval. And even then it is not a problem to run the script again after waiting an hour. Since I'm currently the one maintaining the Travis/Bintray/AppVeyor accounts I'm aware of the limits. Whenever this becomes a problem it's easy to implement whatever authentication method is required at the time. So I didn't bother to implement the github authentication. 

This is really starting to get relevant when we have more than 60 pull requests to check on a regular basis.

PR #2925, opened 2018-12-24
RSS feed: add <description> element to make it validate

davidpanderson:


Rytiss:
Looks good - merging.

PR #2931, opened 2018-12-25
Mac documentation: add a PDF version of Mac build instructions

CharlieFenton:
Adds a PDF version of Mac build instructions to allow automatically linking to the latest version from the TRAC WIKI page <https://boinc.berkeley.edu/trac/wiki/MacBuild>

**Description of the Change**
It is tedious to update the MacBuild TRAC Wiki page each time there are changes to the build instructions. I always maintain correct Macintosh build instructions for each version of BOINC in the source tree at `mac_build/HowToBuildBOINC_XCode.rtf`, but copying them to the wiki requires rewriting them using TRAC Wiki formatting. I use Apple's native _TextEdit_ application to edit the RTF file, and it allows me to also export it as a PDF.

Adding the PDF version will allow me to simply put a link to `https://github.com/BOINC/boinc/raw/master/mac_build/HowToBuildBOINC_XCode.pdf` on that wiki page, which will automatically always point to the latest version of the file.

**Alternate Designs**
I also tried converting it to HTML and enclosing the HTML in the TRAC html processor `{{{#!html` and `}}}` but found that I also needed to remove the `<html>`, `<head>` and `<body>` tags, which would not have updated the document automatically.


adamradocz:
Wouldn't be easier maintening the wiki here in GitHub?

CharlieFenton:
I strongly feel that the appropriate (and best) place for the build instructions is in the source tree for each version. Having a copy on the web (whether in a wiki or anywhere else) is an extra nicety, as far as I am concerned. For one thing, the web is not as conducive to showing the different build instructions for different versions of the software.

The main reason I have been putting a copy in the wiki has been to show people where to find the build instructions for any given software version they wish to build. The secondary reason is that this document shows them how to get the source tree; if the only way to get this document were by first downloading the source tree, they might not know how to do that (though that information _is_ available elsewhere in the TRAC wiki.)

As far as the question of TRAC vs GitHub for web-based documentation, I have no experience with using GitHub for that purpose so I don't have an opinion about that.

PR #2932, opened 2018-12-25
[Android] Fix number comparison

adamradocz:
**Description of the Change**
Fix number comparison using '==', instead of 'equals()'.
Reports any use of == or != to test for Number equality, rather than the equals() method. With auto-boxing it is easy to make the mistake of comparing two Integer (or other subclasses of java.lang.Number) objects instead of two ints.

**Release Notes**
N/A


AenBleidd:
Looks fine. Thanks! Merging...

PR #2935, opened 2018-12-26
[Android] Fix Keyboard inaccessible widget issue

adamradocz:
**Description of the Change**
A widget that is declared to be clickable but not declared to be focusable is not accessible via the keyboard. Please add the focusable attribute as well.

**Release Notes**
[Android] Minor layout fixes.


brevilo:
Thanks for taking care of this Adam!

PR #2951, opened 2019-01-04
web: fix undefined var on host page when <show_results> is false

davidpanderson:


TheAspens:
Looks good - thanks for the fix!

PR #2956, opened 2019-01-10
[Android] Reformat source code, to resolve the Whitespace-Tab issue.

adamradocz:
**Description of the Change**
I used the Android Studio 'Reformat Code' function, without the 'Imports' optimization.
This will resolve the Tab-Whitespace issue, plus a lot more styling issues as well.
I couldn't specify the function to reformat only the whitespace and nothing else.

**Release Notes**
N/A


brevilo:
Gee, I can't possibly review 236 files with almost 30,000 changes in them! If this were only whitespace changes I would probably just merge them after I understood which rules those changes follow. I just can't do that for *all* changes, though, as I'm not even aware of what types of changes went into this.

Furthermore: before we do a general code reformatting of one whole sub-system we should try to agree on one standard that (most of) the subsystem's core developers are fine with. Also this agreed coding style needs to be actionable by those affected. What I mean by that is, that they should ideally be in a position to adjust their editor or code formatter to the agreed style in order to be able to use it safely. Otherwise every formatter will spew out different code which will hinder/preclude our change reviews.

adamradocz:
If you skim through the files, you can see that basically the indentation, whitespace issue, extra space have been reformatted, nothing else.

It has to be done, because of the whitespace issue, the code cleanup took almost 30% time of my PRs.

brevilo:
Sure, I'm not opposing the idea Adam. I know we need to do something about it rather sooner than later. I'll look into this next week when I'm back in office. Stay tuned.

ChristianBeer:
Since you didn't mention it and I also struggled in the past with automated reformatting tools I wanted to ask about the settings you used. There is a Style Guideline for how to format sourcecode in BOINC: https://boinc.berkeley.edu/trac/wiki/CodingStyle which got written over time out of necessity and doesn't strictly follow an established standard.

Some time ago when I investigated CI tools I also tried automated code formatters that detect style violations and create patches to resolve them. Since the BOINC coding style is an amalgamation of different rules I always ended up writing custom rulesets or configuring my code editor in a way that resembled the BOINC coding style the most. But with all the esxception within the codebase this got a nightmare quickly.

To summarize the BOINC coding style a bit one could say that the rules in the linked page above apply to everything except (not sure I covered everything):
- the Drupal part of the PHP code (that follows the Drupal coding style which is different from ours
- the Manager code (everything in `clientgui/`) should follow most of the rules but is exempt from the naming conventions (it uses CamelCase a lot since the original developer liked that)
- the Android code also uses CamelCase because that is the preferred way to write Java code
- Comments are also structured different based on the language and Editor the developer at the time preferred. Most of the Android code is using Javadoc style but the rest of the code doesn't which makes it hard to parse with tools like doxygen. There was an effort to introduce doxygen style comment blocks some years ago (before BOINC switched to the Community driven development model) which was declined at the time

To have exceptions on a component or language level is fine with me and would be a major hassle to reformat. What should be made similar for all the code is the indentation and braces and also maybe the comment blocks.

adamradocz:
I used the standard Android studio conding style. The indentation, braces, comments are the same as boinc coding standard.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/ClientInterfaceImplementation.java
```diff
@@ -27,485 +29,542 @@
...
+                    loop = true; //no result yet, keep looping
+                } else {
+                    //final result ready
+                    if (reply2.error_num == 0) {
+                        if (Logging.DEBUG)
+                            Log.d(Logging.TAG, "ClientInterfaceImplementation.synchronizeAcctMgr: acct mngr reply retrieved.");
+                    } else {
+                        if (Logging.DEBUG)
+                            Log.d(Logging.TAG, "ClientInterfaceImplementation.synchronizeAcctMgr: final result with error_num: " + reply2.error_num);
+                    }
```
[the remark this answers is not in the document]
adamradocz: I can't see any additional brace.
diff

adamradocz:
I left out the assets folder in the new PR #2961.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/ClientInterfaceImplementation.java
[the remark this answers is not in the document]
ChristianBeer: I checked again and yes there is no additional brace. It was just missing because it didn't show me pure whitespace changes.

ChristianBeer:
Closing this PR and opening a new one is not so ideal with this much changes. What I meant with backing out the XML changes was to restore the old file and commit it to the existing branch as a second commit. With the new PR I have to do a new review and look at all the changes again to make sure they don't deviate from what I reviewed earlier.

ChristianBeer:
Since github still knows about the branch we might be able to restore it and then do the necessary commit to exclude the asset files. Which would mean a quicker merge than a complete new review.

adamradocz:
I can't push to this PR, it says I don't have permission.

brevilo:
> What I meant with backing out the XML changes was to restore the old file and commit it to the existing branch as a second commit.

Will start my review (ignoring whitespaces) as soon as that's done. Cheers!

ChristianBeer:
Adam: You need to checkout this branch (specifically 7aacf79) again and then add the extra commit. Then push it again and that should go through without a --force. Forcing the push has basically the same effect as opening a new PR.

adamradocz:
Thanks. Reverted the `cc_config.xml`.

ChristianBeer:
Thanks for fixing that. I'm fine with the changes.

android/BOINC/app/src/main/java/edu/berkeley/boinc/BOINCActivity.java
```diff
@@ -59,298 +60,308 @@
...
+                    startActivity(new Intent(this, SelectionListActivity.class));
+                    break;
+                case R.string.tab_preferences:
+                    ft.replace(R.id.frame_container, new PrefsFragment());
+                    fragmentChanges = true;
+                    break;
+
+                default:
+                    if (Logging.ERROR)
+                        Log.d(Logging.TAG, "dispatchNavBarOnClick() could not find corresponding fragment for " + item.getTitle());
```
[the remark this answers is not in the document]
adamradocz: Done.

android/BOINC/app/src/main/java/edu/berkeley/boinc/BOINCActivity.java
```diff
@@ -59,298 +60,308 @@
...
+                    final Dialog dialog = new Dialog(this);
+                    dialog.requestWindowFeature(Window.FEATURE_NO_TITLE);
+                    dialog.setContentView(R.layout.dialog_about);
+                    Button returnB = dialog.findViewById(R.id.returnB);
+                    TextView tvVersion = dialog.findViewById(R.id.version);
+                    try {
+                        tvVersion.setText(getString(R.string.about_version) + " "
+                                + getPackageManager().getPackageInfo(getPackageName(), 0).versionName);
+                    } catch (NameNotFoundException e) {
+                        if (Logging.WARNING) Log.w(Logging.TAG, "version name not found.");
```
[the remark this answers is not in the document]
adamradocz: Done.

android/BOINC/app/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
```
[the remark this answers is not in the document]
adamradocz: I couldn't find anywhere this option.

brevilo:
Is it me or GitHub (again)? I can only see useless XML line break changes but no `if` changes: https://github.com/BOINC/boinc/pull/2956/commits/f4df4c2d5148088363deb9a1f2e69d32fd6577b7?utf8=%E2%9C%93&diff=split&w=1

android/BOINC/app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
brevilo: Too bad. Thanks for checking.

adamradocz:
I don't understand. Push it again...

adamradocz:
Should be good now.

adamradocz:
I can resolve to the oneliner debug `if` statements don't get braces, but I have to make a new PR for that.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/NoticeNotification.java
```diff
@@ -135,34 +136,25 @@ private Notification buildNotification() {
...
+                    setLargeIcon(NoticeNotification.getLargeProjectIcon(this.context, projectName));
+        }
+        else {
             // multi notice view
-            nb.setNumber(notices)
-                    .setLargeIcon(BitmapFactory.decodeResource(
-                            this.context.getResources(),
-                            R.drawable.ic_stat_notify_boinc_normal))
-                    .setSubText(this.context.getString(R.string.app_name));
+            nb.setNumber(notices).setLargeIcon(BitmapFactory.decodeResource(this.context.getResources(), R.drawable.ic_stat_notify_boinc_normal)).setSubText(this.context.getString(R.string.app_name));
```
brevilo: This is a type of change that makes things worse. The human editor had clear and justified intentions (with the line breaks) which should be preserved.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/NoticeNotification.java
```diff
@@ -172,27 +164,16 @@ private Notification buildNotification() {
...
-            if (Log.isLoggable(Logging.TAG, Log.DEBUG)) Log.d(
-                    Logging.TAG,
-                    e.getLocalizedMessage(),
-                    e
-            );
-            return BitmapFactory.decodeResource(
-                    context.getResources(),
-                    R.drawable.ic_stat_notify_boinc_normal
-            );
+            return (projectIconBitmap = Monitor.getClientStatus().getProjectIconByName(projectName)) !=
```
brevilo: This whole block is a type of change that makes things worse. The human editor had clear and justified intentions (with the line breaks) which should be preserved.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/AccountOutParser.java
```diff
@@ -42,26 +42,30 @@ public static AccountOut parse(String rpcResult) {
...
     }
 
     @Override
     public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException {
         super.startElement(uri, localName, qName, attributes);
-        if (localName.equalsIgnoreCase("error_num") ||
-                localName.equalsIgnoreCase("error_msg") ||
-                localName.equalsIgnoreCase("authenticator")) {
-            if (mAccountOut == null)
+        if (localName.equalsIgnoreCase("error_num") || localName.equalsIgnoreCase("error_msg") ||
```
brevilo: This is a type of change that makes things worse. The human editor had clear and justified intentions (with the line breaks) which should be preserved.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/CcStateParser.java
```diff
@@ -132,11 +133,9 @@ public void startElement(String uri, String localName, String qName, Attributes
             mResultsParser.startElement(uri, localName, qName, attributes);
         }
         if (localName.equalsIgnoreCase("core_client_major_version") ||
-                localName.equalsIgnoreCase("core_client_minor_version") ||
-                localName.equalsIgnoreCase("core_client_release") ||
-                localName.equalsIgnoreCase("have_ati") ||
-                localName.equalsIgnoreCase("have_cuda")
-                ) {
+            localName.equalsIgnoreCase("core_client_minor_version") ||
```
brevilo: This is a type of change that makes things worse. The human editor had clear and justified intentions (with the line breaks) which should be preserved.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/GlobalPreferences.java
```diff
@@ -92,15 +92,7 @@ public void writeToParcel(Parcel dest, int flags) {
...
-        dest.writeBooleanArray(new boolean[]{
-                run_on_batteries,
-                run_if_user_active,
-                run_gpu_if_user_active,
-                leave_apps_in_memory,
-                dont_verify_images,
-                override_file_present,
-                network_wifi_only
-        });
+        dest.writeBooleanArray(new boolean[]{run_on_batteries, run_if_user_active, run_gpu_if_user_active, leave_apps_in_memory, dont_verify_images, override_file_present, network_wifi_only});
```
brevilo: This is a type of change that makes things **clearly** worse. The human editor had clear and justified intentions (with the line breaks) which should be preserved.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/Project.java
```diff
@@ -163,20 +163,7 @@ public void writeToParcel(Parcel dest, int flags) {
...
-                scheduler_rpc_in_progress,
-                attached_via_acct_mgr,
-                detach_when_done,
-                ended,
-                trickle_up_pending,
-                no_cpu_pref,
-                no_cuda_pref,
-                no_ati_pref
-        });
+        dest.writeBooleanArray(new boolean[]{master_url_fetch_pending, non_cpu_intensive, suspended_via_gui, dont_request_more_work, scheduler_rpc_in_progress, attached_via_acct_mgr, detach_when_done, ended, trickle_up_pending, no_cpu_pref, no_cuda_pref, no_ati_pref});
```
brevilo: This is a type of change that makes things **clearly** worse. The human editor had clear and justified intentions (with the line breaks) which should be preserved.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/ProjectConfig.java
```diff
@@ -77,13 +81,7 @@ public void writeToParcel(Parcel dest, int flags) {
...
         dest.writeString(termsOfUse);
 
-        dest.writeBooleanArray(new boolean[]{
-                usesName,
-                webStopped,
-                schedulerStopped,
-                accountCreationDisabled,
-                clientAccountCreationDisabled,
-                accountManager});
+        dest.writeBooleanArray(new boolean[]{usesName, webStopped, schedulerStopped, accountCreationDisabled, clientAccountCreationDisabled, accountManager});
```
brevilo: This is a type of change that makes things worse. The human editor had clear and justified intentions (with the line breaks) which should be preserved.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/Result.java
```diff
@@ -125,19 +125,7 @@ public void writeToParcel(Parcel dest, int flags) {
...
-                project_suspended_via_gui,
-                coproc_missing,
-                gpu_mem_wait,
-                active_task,
-                supports_graphics,
-                too_large,
-                needs_shmem,
-                edf_scheduled
-        });
+        dest.writeBooleanArray(new boolean[]{ready_to_report, got_server_ack, suspended_via_gui, project_suspended_via_gui, coproc_missing, gpu_mem_wait, active_task, supports_graphics, too_large, needs_shmem, edf_scheduled});
```
brevilo: This is a type of change that makes things **clearly** worse. The human editor had clear and justified intentions (with the line breaks) which should be preserved.

adamradocz:
I'll see what can I do.

adamradocz:
Should be OK now.

brevilo:
Sorry to say this, but no. If you have a look at 37311ab you'll see that it doesn't deal with line breaks at all. It mostly changes whitespace in front of parentheses and restructures a few imports.

brevilo:
That's strange. What did you change in the config between the last two commits? Why did it change the parentheses spacing and not the line breaks?

brevilo:
While at it, could you please [export](https://www.jetbrains.com/help/idea/2018.2/exporting-and-importing-settings.html) the relevant AndroidStudio settings and add them to the repo? This way other contributors can simply import and use them as well, establishing a de-facto coding standard.

I suggest limiting the exported settings to "Code Style" only and place the JAR file in the repo at `android/AndroidStudio/settings.jar`. Future style changes should then include an updated `settings.jar` in the same commit.

Thanks 👍

adamradocz:
> That's strange. What did you change in the config between the last two commits? Why did it change the parentheses spacing and not the line breaks?

I set all the settings to keep the line breaks and one-liners. Plus I removed the spaces before parentheses. I think it's more convenient to code.

> While at it, could you please export the relevant AndroidStudio settings and add them to the repo? 

I exported the setting from the File menu, and I also exported the Code Style scheme from the Cody Style setting tab. It looks like, the jar file not containing all of the data?

brevilo:
> I set all the settings to keep the line breaks and one-liners. Plus I removed the spaces before parentheses. I think it's more convenient to code.

Don't get me wrong, I didn't mean to criticize that. I was just wondering why only the spacing changed when you set out to revert the line break changes. Do you see any chance to fix the line breaks? Now that I have your config I hope to give this a try myself, if time allows.

> It looks like, the jar file not containing all of the data that the exported XML does?

The JAR file currently contains the "Default" scheme, not the "Project" scheme the XML contains, so that explains the difference. Any chance to switch the scheme for the JAR export? If not, we should just provide the XML and add a note how to import/apply that.

adamradocz:
> The JAR file currently contains the "Default" scheme, not the "Project" scheme the XML contains, so that explains the difference. Any chance to switch the scheme for the JAR export? If not, we should just provide the XML and add a note how to import/apply that.

I thought there is something like this. Unfortunately, there is no option which scheme to export.

brevilo:
> Unfortunately, there is no option which scheme to export.

In that case please just remove JAR file for the time being as we don't wan't to enforce our settings on all projects a contributor might work on.

> Doesn't the Android project contains the project scheme and only the .gitignore file should be modified, to not ignore it?

No idea. Where did you get the single XML file from? If that's part of an ignored path we should probably change that to keep things easy. Related, how does one apply it anyway?

adamradocz:
> Where did you get the single XML file from?

I exported it from the File -> Setting -> Editor -> Code Style

> Related, how does one apply it anyway?

File -> Setting -> Editor -> Code Style, then Gear wheel icon -> Import scheme

brevilo:
Go it. I agree with you that we should retain the coding style settings as part of the project itself, not as a separate file. I had a look and the (most) relevant file can be added to the repo by slightly modifying `.gitignore` like this:

```
diff --git a/.gitignore b/.gitignore
index b1469eb219..9fc82a2e11 100644
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,8 @@
 /android/**/*.iml
 /android/**/.gradle
 /android/**/local.properties
-/android/**/.idea
+/android/**/.idea/*
+!/android/**/.idea/codeStyles
 /android/**/build
 /android/**/captures
 /android/**/.externalNativeBuild
```
However, I don't know which other files Android Studio might need in order to have a working "project" so the code style file alone might now be enough. Do you know those details?

brevilo:
There you go:
```
x b1469eb219..1152eb6869 100644
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,13 @@
 /android/**/*.iml
 /android/**/.gradle
 /android/**/local.properties
-/android/**/.idea
+/android/**/.idea/caches
+/android/**/.idea/libraries
+/android/**/.idea/gradle.xml
+/android/**/.idea/modules.xml
+/android/**/.idea/tasks.xml
+/android/**/.idea/usage.statistics.xml
+/android/**/.idea/workspace.xml
 /android/**/build
 /android/**/captures
 /android/**/.externalNativeBuild
```
Just add that in a dedicated commit and remove `android/AndroidStudio/Code Style (Scheme).xml` again.

brevilo:
>>  Do you see any chance to fix the line breaks?
> There is no option for keeping the formatting in case of chain method calls.

What I was trying to get at was that this particular change was part of 906cc9527e. So you remember which options you changed for that commit you should be able to revert that part of it, no?

Another closely related one could be the hard wrap at X characters which should have caused a few of the changes I complained about. I think we should turn that off entirely and leave wrapping (line breaks!) to the devs. I you can't disable it, use a very high value (e.g. 500). Unfortunately that's probably not going to fix already reformatted code, so you'd have to go back to before 906cc9527e to see whether that works. But that shouldn't be a problem as long as you retain the code style XML.

adamradocz:
I did a hard reset and ran the Reformat Code function again with the modified setting to preserve all the line breaks, and it gave this result. I don't know if there any option to prevent this.

brevilo:
By "this" I presume no difference to the latest commit, I presume. Thanks for trying Adam.

Sh*t, we should have started tracking the code style definition earlier. That would make it clear what caused that change in https://github.com/BOINC/boinc/commit/906cc9527e4efb642b1630a45a68ade49fa38dba.

I need to find the time to look into this myself as the changes I flagged above aren't acceptable.

brevilo:
Update: I began looking into this, based on your branch. Hang in there... 👍

brevilo:
Hi @adamradocz, I think I found a fix for the line break issue 😃 

I first rolled back the last two commits of your's, right before https://github.com/BOINC/boinc/commit/906cc9527e4efb642b1630a45a68ade49fa38dba which was the one introducing the issues. I then modified the code style settings until I reached what I was aiming for.

Please have a look at [this diff](https://github.com/BOINC/boinc/compare/android-reformat-sourcecode...brevilo:android-reformat-sourcecode) which compares your branch against my improved one. As you can see the original line breaks are back again and I'm 95% happy with the result!

* What do you think?
* Please check out my branch and open it (fresh) in your Android Studio. Then run another reformat on `android/BOINC/app` to verify that **no** extra changes appear. This should prove that the code style settings are now portable.

If we agree on the current settings I'll push my changes into your branch (this PR) such that we can finally merge the PR and carry on with more important changes 👍

adamradocz:
Tested, looks good.

brevilo:
Thanks for checking. I pushed my changes and approved the PR. I'll merge as soon as Scrutinizer is done...

brevilo:
Finally! Thanks for your patience 😄

PR #2967, opened 2019-01-17
web: keep server version # in a file (html/inc/server_version.inc)

davidpanderson:
... rather than generating it during configure.
This way things work for projects that don't do "configure",
e.g. because they use only the web code.
And it's simpler.

Maintain this as for client version #s.
The minor version # is even in release branches,
odd in development branches like master.

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,7 +96,6 @@ py/Boinc/version.py
 py/setup.py
 stamp-h1
 test/version.inc
-html/inc/release.inc
 
 ## files created by make:
 *.o
--- a/generate_svn_version.sh
+++ b/generate_svn_version.sh
@@ -62,16 +62,3 @@ if cmp "$HEADER" "$TMPFILE" >/dev/null 2>&1; then
 else
 	mv "$TMPFILE" "$HEADER"
 fi
-
-if [ ! -z ${SERVER_VERSION} ]; then
-    SERVER_VERSION=`echo ${SERVER_VERSION} | sed 's#.*/##'`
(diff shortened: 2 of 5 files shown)


davidpanderson:
What's the status of this?  Without this I'm getting?
PHP Parse error:  syntax error, unexpected '[' in /mydisks/a/users/boincadm/projects/test/html/user/server_status.php on line 192, referer: https://boinc.berkeley.edu/test/

Rytiss:
That is an unrelated issue. With PHP 5.3 you cannot use `explode()[0]` (you have to put into a variable first); with PHP 5.4+ you can.

I will not comment on this particular PR as I don't have a strong opinion about the functionality change it does, but I created a separate one that fixes the issue David has encountered - @davidpanderson please review and merge #3034.

davidpanderson:
This is the right way to store server version #, and the PHP fix is a by-product.

TheAspens:
This looks fine to me.

lfield:
I know that this PR has already been merged but in my view this is a backwards step as the server version is now hard-coded rather than generated from the tag. As far as I can tell if anyone installs from master, they will publish version 1.1.0. This approach would require the release manager to update the server_version.inc file before each tag. At least the default version in master should identify that this is not a proper release such as -1,-1,-1.

davidpanderson:
That's right; the version manager has to update that file,
similar to how the client release works

PR #2968, opened 2019-01-17
remote job submission: add API for getting job counts

davidpanderson:
If a remote job submission system wants to maintain a constant
level of unsent jobs, it needs a way to get this value.
This commit adds a Python API for this.
It also returns other counts such as in-progress results,
and WUs needing validation/assimilation/file delete.

I could add PHP and C++ interfaces too if needed.

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**


TheAspens:
@lfield - can you review this change as well?  We don't use the remote job submission API so I am less familiar with it.

TheAspens:
I merged from master and cleaned up the merge conflicts.

lfield:
This works as intended but does not show the break down per app. This may be required by some projects for this use case but can be addressed in a separate PR.

Issue #1348, opened 2015-02-04
Closing manager window minimizes it instead (Linux)


romw:
**Reported by Nicolas on 4 Apr 45021490 12:25 UTC**
On Linux, clicking the Close button on the Manager titlebar
minimizes the window.
Not to the tray, but to the normal task bar.
In other words, the Close button does
exactly the same as the Minimize button,
making it pretty useless.

FileClose Window also minimizes the manager.
I couldn't find any way to minimize to tray.

My desktop environment is KDE Plasma 4.11.13.

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1377

shalnoff:
agree. Moreover, the window doesn't minimise to tray at all (Mate, Linux Mint 32bit. BOINIC 7.0.27)
would be great to fix it. 

thank you for great applicaiton!

Songworks:
Any plans to fix this?
I like the tray icons Snooze feature. But the icon is obviously only present if the managers window is also open.
System: Linux Mint 17.3 running boinc 7.6.2 64bit.

eMPee584:
Yes, please make BOINC Manager minimize to tray. I like to keep it running for the "snooze" option, but it wastes space in the task list.

eMPee584:
Actually, found a way to snooze BOINC from the CLI (f.e. for 15 minutes):
`boinccmd --set_run_mode never $((60*15))`
This will probably obliterate the GUI for me.

elfring:
Can this special software behaviour happen because the support for the functionality “Minimize window into system tray” is unclear for [the usage of the member function “wxTopLevelWindow::Iconize”](http://docs.wxwidgets.org/trunk/classwx_top_level_window.html#ae461559192e33ef111c3f147999a26bf "Clarification for usage of “wxTopLevelWindow::Iconize”?") which is called by [the member function “CBOINCBaseFrame::OnClose”](https://github.com/BOINC/boinc/blob/e6607a63402a2b4ce2d88588efa2c011ffae9b86/clientgui/BOINCBaseFrame.cpp#L325 "Closing an application window") so far?

ChristianBeer:
Problem still present with 7.6.33 on Debian 9 (KDE)

Ageless93:
@lfield, one for your list of fixes?

lfield:
[...]

PR #2969, opened 2019-01-17
Fixes window close action on Linux

### LINKED ISSUE #2604: Should we allow ALT+F4 to actually close BOINC?
Testing ALT+F4 on an open BOINC (Manager) on Windows 7, it 'closes' the BOINC Manager window, but leaves everything actually running. It just moves it to the system tray. 
Same thing for right-clicking the BOINC icon on the task bar and clicking 'Close window'. 

In both cases it's as if one's clicking the red X in the top right corner. 

I'm sure this behaviour was added at one time to force the user to actually go through BOINC Manager or BOINCCMD to close everything down correctly - and in a timely fashion so it doesn't abruptly close the science application. But I'm wondering if we shouldn't allow ALT+F4 to close the manager and client down, as I can think of times when you quickly want rid of everything and going the cumbersome way of via the GUI or typing the whole command can cost precious time you don't want to lose in closing programs down 'the way the developer intended' (think breaking hardware). 

What does everyone think? 
And is what I surmise the actual reason that ALT+F4 doesn't close the program down?

Germano0:
Fixes #2604 #1348 

**Description of the Change**
On Linux when you close the BOINC Manager window, it will be minimized instead of closed.
The patch closes the BOINC Manager window correctly.

**Alternate Designs**
Read
https://github.com/BOINC/boinc/issues/2604#issuecomment-445826926
and
https://github.com/BOINC/boinc/issues/1348#issuecomment-444534306
(they are the same comment)

**Release Notes**
On Linux, when a user closes BOINC Manager window, it will be closed rather than minimized.

### DIFF
--- a/clientgui/BOINCBaseFrame.cpp
+++ b/clientgui/BOINCBaseFrame.cpp
@@ -330,9 +330,8 @@ void CBOINCBaseFrame::OnClose(wxCloseEvent& event) {
         Destroy();
     } else {
 #ifdef __WXGTK__
-        // Apparently aborting a close event just causes the main window to be displayed
-        // again.  Just minimize the window instead.
-        Iconize();
+        wxGetApp().FrameClosed();
+        Destroy();
 #else
         Hide();
 #endif


AenBleidd:
Personally I'm not sure that it is a fine that the same application has different behavior of the same action on different OSs.

Ferroin:
@AenBleidd It arguably already has different behavior on different OSes for closing the window.  ATM, the manager minimizes itself to the taskbar on Linux instead of closing the window and persisting in the tray like it does on Windows.  The net result is that the manager still shows up in the window list on Linux, but not on Windows, and therefore on Linux it does not behave consistently with _every other application on the system_, which I would argue is a bigger consistency issue than inconsistent behavior across OSes.

Germano0:
I also noticed that before this patch, GNOME users were not able to close BOINC Manager application by using the close button on the upper right. The Manager was still in "activities" GNOME menu.
Now they are able to close it regulary.

CharlieFenton:
Will this change also cause the Manager to tell the client to exit?

Germano0:
> Will this change also cause the Manager to tell the client to exit?

No, the client is not affected

RichardHaselgrove:
Can you talk us through all the possible permutations, please, so Linux users can test that all of them work.

When the Close button is clicked,

1) The graphical window is hidden, but the Manager application continues to run and can be re-activated by the user in the same session
2) The Manager application exits, and may or may not - depending on selected option - display the exit dialog
3) The Manager application exits, and may or may not - depending on the persistent option in the Exit dialog - also stop the Client.

Germano0:
> Can you talk us through all the possible permutations, please, so Linux users can test that all of them work.
> 
> When the Close button is clicked,
> 
>     1. The graphical window is hidden, but the Manager application continues to run and can be re-activated by the user in the same session

[X] button (right upper side of the window) always behaved like "Close Window" File menu entry.
On Linux there was this weird thing to minimize the Manager the **first** time the user tried to close Manager window. If user right clicked on system taskbar, and clicked on "close", finally the Manager will close behaving like "Close Window" File menu entry.

>     2. The Manager application exits, and may or may not - depending on selected option - display the exit dialog

Clicking on the [X] button (right upper side of the window) never triggered display exit dialog, even before this patch.

>     3. The Manager application exits, and may or may not - depending on the persistent option in the Exit dialog - also stop the Client.

`Exit from Manager` has always had a different behaviour from the [X] button (right upper side of the window). The latter always behaved like "Close Window" File menu entry.

Concerning Exit dialog, please refer to pull request [Removed "Exit from BOINC Manager" menu entry on Linux](https://github.com/BOINC/boinc/pull/3028)

RichardHaselgrove:
I think you know that I am primarily a Windows user. My reason for asking the questions was that many users are transitioning from Windows to the new consumer-facing Linux distros, and will expect to be able to bring their prior BOINC experience with them.

Hence the distinction between 'Close' and 'Exit'. Windows users expect to be able to control the client on demand, in both service and user modes. For example, there's s difference between 'stop client by user request' (which allows time to show dialogs and process user responses), and 'stop client, plus child processes, unconditionally because of system closedown'. The latter suppresses all dialogs, and has been known to kill processes "with extreme prejudice", resulting in loss of work. For this reason, many users choose to stop the client manually (from the Manager interface) prior to performing system maintenance. It is also possible to re-start the Client from the Manager, even in service mode: the utility program boincsvcctrl.exe is installed with administrative privileges for this purpose.

I was just a little concerned that some of the subtle and sophisticated nuances of the original BOINC client design might be being swept aside, but I'll leave it at that.

Germano0:
> My reason for asking the questions was that many users are transitioning from Windows to the new consumer-facing Linux distros, and will expect to be able to bring their prior BOINC experience with them.
As Fedora / RHEL /CentOS / SL BOINC package maintainer I say that if a Windows user wants to migrate to Linux he has to adapt to Linux behaviour, not viceversa.

> Hence the distinction between 'Close' and 'Exit'. Windows users expect to be able to control the client on demand, in both service and user modes. For example, there's s difference between 'stop client by user request' (which allows time to show dialogs and process user responses), and 'stop client, plus child processes, unconditionally because of system closedown'. The latter suppresses all dialogs, and has been known to kill processes "with extreme prejudice", resulting in loss of work. For this reason, many users choose to stop the client manually (from the Manager interface) prior to performing system maintenance. It is also possible to re-start the Client from the Manager, even in service mode: the utility program boincsvcctrl.exe is installed with administrative privileges for this purpose.

Linux users simply suspend the BOINC client when they want, from BOINC Manager. It frees RAM from working units, and simply leaves in memory the `boinc` process that uses only 4384 kB of RAM. No loss of work, even with VirtualBOX virtual machines.
When the user halts the system, systemd will call the [BOINC ExecStop script](https://github.com/BOINC/boinc/blob/master/client/scripts/boinc-client.service.in#L13)

Ferroin:
> I think you know that I am primarily a Windows user. My reason for asking the questions was that many users are transitioning from Windows to the new consumer-facing Linux distros, and will expect to be able to bring their prior BOINC experience with them.

I have three issues with this argument as someone who uses BOINC on both Linux and Windows:

1. As mentioned in https://github.com/BOINC/boinc/pull/2969#issuecomment-455183030, things are already inconsistent across platforms, and this isn't making that any worse while making consistency with the rest of the system better.
2. Providing an easy transition isn't the job of the applications, it's the job of the distribution.  Yes, BOINC is an odd case compared to most applications, but that doesn't change the fact that the current behavior is making things harder for _actual_ Linux users to supposedly cater to _potential_ Linux users.  Put slightly differently, if BOINC is the _only_ thing someone who is switching from Linux to Windows has issues with, then they are very likely going to be able to work around those issues because they managed the rest of the transition without issue, while if it's one of many things, they probably aren't going to continue to use Linux anyway, so us making things 'easy' for them doesn't matter.
3. There's no way to be truly, portably, consistent with how things work on Windows on Linux, because there are just too many ways to handle running BOINC.  At minimum, any of the following may be the case for how it gets run:
    * It may be managed directly by a traditional `init` process, in which case there is exactly zero way for _anybody_ to keep it from running.
    * It may be managed by a traditional init system that does not provide service management (for example, the traditional Debian init setup), in which case it can be trivially stopped by _some_ users, but can only be restarted by the root user unless a significant amount of complicated secondary configuration is done or the security of the system is severely compromised.
    * It may be managed by a dedicated service manager independent of the init system, in which case it may or may not be possible to reliably stop, start, or restart it as any given user without similar constraints to the above case.
    * It may be handled by systemd, in which case it's actually _probably_ not hard to handle, but still requires non-trivial configuration that will likely have to be customized to the particular system.

    On top of that, you can get complicated combinations or interactions between the above.  I know people who run a dedicated VM or container just for BOINC, which adds an extra layer of indirection to any of the above.  Alternatively, you have people like me who use a traditional init system to start and stop the client, but actually manage it through a service manager that is responsible for invoking the init scripts to start and stop the client.

lfield:
Tested and works as intended.

Germano0:
new version of this pull request that has been closed by mistake https://github.com/BOINC/boinc/pull/3095

PR #2973, opened 2019-01-22
Scrutinizer: adjust config to remove issues in dependencies

ChristianBeer:
**Description of the Change**
There are lots of issues with the drupal modules that are copied here as dependencies. Treat them that way to only show issues with our own code.

**Release Notes**
N/A

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -9,11 +9,12 @@ filter:
         - '*.tpl.php'
         - '*.min.js'
         - 'html/inc/htmLawed.php'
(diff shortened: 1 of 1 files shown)


ChristianBeer:
The difference between exclusion and dependency is that the first does not get scanned at all and the second does get scanned but not analyzed. The effect is that if the remaining code uses a class that is defined in a dependency it does not get flagged as an issue as it would be in case of the exclusion. So it is in general better to mark the existing code as a dependency than to exclude it. Which in turn means I might have to move `html/inc/htmLawed.php` into the dependencies too.

PR #2979, opened 2019-01-23
Do not allow page_head() to be called twice

### LINKED ISSUE #2960: Web: wrong use of error_page() function throughout the code
**Describe the bug**
The current implementation of `error_page()` in [html/inc/util.inc#L367](https://github.com/BOINC/boinc/blob/3979d30c41ff629e2fbbd23b8efc41792eb669e7/html/inc/util.inc#L367) contains a call to `page_head()` which among other things outputs header information (via `header()`. This means that after a page called `page_head()` itself it can't use the current `error_page()` function or the header is output twice. This leads to duplication of HTML code (two navbars) and PHP warnings `Cannot modify header information - headers already sent by`.

**Steps To Reproduce**
1. Create a page that calls `error_page()` after `page_head()`
2. View this page with a browser

**Expected behavior**
There should be no warnings and only one vaigation bar on the top. Therefore the second call to `page_head()` when calling `error_page()` **after** `page_head()` should be avoided. When `error_page()` is called **before** `page_head()` then of course it needs to call `page_head()` itself.

**Scope**
This is a rather big change as one has to check each PHP file and modify the `error_page()` call that happens after a `page_head()` call. So I would propose to do this in chunks and document progress in this issue.

Rytiss:
**Description of the Change**

error_page() calls page_head(), but there are instances of code that call error_page() having already called page_head() before, thus attempting to send headers twice and creating double page header.

Fixes #2960 

**Release Notes**
Prevent page_head() from being called multiple times. Notice to projects: If you have a page_head() function override in your project.inc you need to apply this change there too.

### DIFF
--- a/html/inc/util.inc
+++ b/html/inc/util.inc
@@ -243,6 +243,10 @@ function page_head(
     global $caching, $cache_control_extra, $did_page_head;
     global $is_login_page, $fixed_navbar;
 
+    if ($did_page_head) {
+        return;
+    }
+    
     $did_page_head = true;
     $url_base = url_base();
 


ChristianBeer:
This definitely fixes the problem but there should be an entry in the server release notes because existing projects that have this function overridden in their project config need to also change those or still suffer from the problem.

Rytiss:
> This definitely fixes the problem but there should be an entry in the server release notes because existing projects that have this function overridden in their project config need to also change those or still suffer from the problem.

Updated the release notes field in the original description above. Do I need to do anything else?

ChristianBeer:
I added: **"Notice to projects: If you have a page_head() function override in your project.inc you need to apply this change there too."** to the release note and confirmed that it fixes the issue.

Thanks for your contribution.

PR #2980, opened 2019-01-24
Strip HTML tags for the page title

sergeyklay:
**Description of the Change**
Strip HTML tags for the page title.

**Screenshots**
image

**Release Notes**
- Fixed title for the pages that were subject to localization

### DIFF
--- a/html/user/team_members.php
+++ b/html/user/team_members.php
@@ -48,7 +48,7 @@ if (!$team) {
     set_cached_data(TEAM_PAGE_TTL, serialize($team), $cache_args);
 }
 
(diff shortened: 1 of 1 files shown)


davidpanderson:
Let's fix this by changing the call to page_head() in team_members.php.
That's the only call that passes HTML.

sergeyklay:
@davidpanderson Should I leave it as is or change the call to page_head() in team_members.php instead?

davidpanderson:
Please make that change.

sergeyklay:
Done

PR #2982, opened 2019-01-24
Drupal: PHP7 fixes for boinc modules

drshawnkwang:
**Description of the Change**
Modified boinc modules to fix PHP 7 compatibility issues

* mysql_xyz() functions replaced with mysqli_mxyz() functions.
* Removed references to ini settings `safe_mode`.
* Updated a class constructor to use modern PHP standards.

**Release Notes**
N/A

Part of https://dev.gridrepublic.org/browse/DBOINCP-468

### DIFF
--- a/drupal/sites/default/boinc/modules/boincimport/boincimport.module
+++ b/drupal/sites/default/boinc/modules/boincimport/boincimport.module
@@ -240,11 +240,6 @@ function boincimport_process_form() {
   // Causes problems with form api redirect
   //ini_set('display_errors', TRUE);
 
-  // Adjust how long you want the script to run...
-  if (!ini_get('safe_mode')) {
-    set_time_limit(variable_get('boincimport_time_limit', 0));
-  }
-  
   // Check for a lock on the import process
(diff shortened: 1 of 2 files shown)


drshawnkwang:
Did some more testing: Scrutinizer reports bugs on the new `mysqli_num_rows()` functions:

```
$variable of type `false|resource` is incompatible with the type `mysqli_result` expected by parameter `$result` of `mysqli_num_rows()`. ( Ignorable by Annotation )
```

I tested the code and dumped out the class type of `variable` with `get_class()`, and it is an object of class `mysqli_result`. I don't know why the Scrutinizer flags the variable as being something else. I believe that the code should work. `db_query` will return a `mysqli_result` which will be passed to `mysqli_num_rows()` and be processed correctly.

As a result I am re-opening this PR.

ChristianBeer:
I also noticed a false positive involving incompatible types. I opened a bug report with scrutinizer about that. You could probably for now ignore those kind of bugs (while still doing a quick check though).

PR #2997, opened 2019-01-31
Drupal: Boincimport - fix errant character problem.

drshawnkwang:
**Description of the Change**
Fixes problem with errant chracters

Part of https://dev.gridrepublic.org/browse/DBOINCP-468

**Release Notes**
N/A


tristanolive:
Thanks for the catch! Maybe using the online editor was not such a great idea...

PR #2999, opened 2019-02-01
Drupal: Removed PHP ini setting 'arg_separator.output'.

drshawnkwang:
**Description of the Change**
With this setting present, reCAPTCHA does not work. Removing this setting, un-breaks reCAPTCHA.

Part of PHP 7.0 compatibility.

https://dev.gridrepublic.org/browse/DBOINCP-484

**Release Notes**
N/A

### DIFF
--- a/drupal/sites/default/settings.php
+++ b/drupal/sites/default/settings.php
@@ -162,7 +162,6 @@ else {
  * settings are used there. Settings defined here should not be
  * duplicated there so as to avoid conflict issues.
  */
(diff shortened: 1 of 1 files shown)


tristanolive:
That ini_set() call was apparently added to Drupal in 2003 with the following commit message:

> "Added 'php_value arg_separator.output "&"' as suggested by Curtis"

...and to this day no one has any idea what the purpose of it was. This should be a safe merge. :cheese:

Issue #2988, opened 2019-01-25
Incorrect peak FLOPS for AMD GPU


davidpanderson:
The OpenCL GPU detection for a particular AMD GPU reports 43 ExaFLOPS - about 10,000X too big.
(this is 7.10.2, but same in 7.14.2)

22-Dec-2018 12:28:23 [---] cc_config.xml not found - using defaults
22-Dec-2018 12:28:23 [---] Starting BOINC client version 7.10.2 for windows_x86_64
22-Dec-2018 12:28:24 [---] OpenCL: AMD/ATI GPU 0: AMD Radeon(TM) Vega 8 Graphics (driver version 2580.6 (PAL,HSAIL), device version OpenCL 2.0 AMD-APP (2580.6), 6567MB, 6567MB available, 43980464 GFLOPS peak)

adamradocz:
Maybe driver issue? The Intel OpenCL driver shows me different result too, then the Beignet driver,

RichardHaselgrove:
Whatever the cause, we probably need to update [lib/coproc.cpp](https://github.com/BOINC/boinc/blob/4f6e52f1cbac1e776b9ea0b6853b3e3e156a342b/lib/coproc.cpp#L840) for AMD/ATI, as we did last year for NVidia in #2707

davidpanderson:
Addendum from user:

Update: On a hunch, I set the underclocked GPU back to auto (400-1100MHz).  The spurious peak GFLOPS issue went away. 02/02/2019 15:28:02 \|  \| OpenCL: AMD/ATI GPU 0: AMD Radeon(TM) Vega 8 Graphics (driver version 2766.5 (PAL,HSAIL), device version OpenCL 2.0 AMD-APP (2766.5), 6567MB, 6567MB available, 1127 GFLOPS peak)On underclocking it again, this time to a slightly different nominal speed, we get the old issue back:02/02/2019 15:57:59 \|  \| OpenCL: AMD/ATI GPU 0: AMD Radeon(TM) Vega 8 Graphics (driver version 2766.5 (PAL,HSAIL), device version OpenCL 2.0 AMD-APP (2766.5), 6567MB, 6567MB available, 43980464 GFLOPS peak)CPUZ seems to report the correct GPU speed in each case.

JuhaSointusalo:
Workarounded in #3001.

JuhaSointusalo:
Eeh, worked around...

PR #3001, opened 2019-02-03
client: fix possible overflow in peak FLOPS calculation for OpenCL GPUs

davidpanderson:
... and generate a warning message if impossible values
Attempt to fix issue #2988

### DIFF
--- a/client/gpu_amd.cpp
+++ b/client/gpu_amd.cpp
@@ -377,6 +377,9 @@ void COPROC_ATI::get(
         cc.atirt_detected = atirt_detected;
         cc.device_num = i;
         cc.set_peak_flops();
+        if (cc.bad_gpu_peak_flops("CAL", s)) {
+            warnings.push_back(s);
+        }
         get_available_ati_ram(cc, warnings);
         ati_gpus.push_back(cc);
     }
--- a/client/gpu_nvidia.cpp
+++ b/client/gpu_nvidia.cpp
@@ -240,6 +240,7 @@ void COPROC_NVIDIA::get(
     char buf[256];
     int j, itemp;
     size_t global_mem = 0;
+    string s;
     COPROC_NVIDIA cc;
 
 #ifdef _WIN32
@@ -444,6 +445,9 @@ void* cudalib = NULL;
         cc.cuda_version = cuda_version;
         cc.device_num = j;
         cc.set_peak_flops();
+        if (cc.bad_gpu_peak_flops("CUDA", s)) {
+            warnings.push_back(s);
+        }
         get_available_nvidia_ram(cc, warnings);
         nvidia_gpus.push_back(cc);
     }
--- a/client/gpu_opencl.cpp
+++ b/client/gpu_opencl.cpp
@@ -202,6 +202,7 @@ void COPROCS::get_opencl(
     vector<int>devnums_pci_slot_sort;
     vector<OPENCL_DEVICE_PROP>::iterator it;
     int max_other_coprocs = MAX_RSC-1;  // coprocs[0] is reserved for CPU
+    string s;
 
     if (cc_config.no_opencl) {
         return;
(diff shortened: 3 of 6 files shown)


RichardHaselgrove:
This problem has been traced to a bad value returned by the AMD OpenCL driver stack (confirmed by clinfo), and is thus outside our immediate control at source.

This patch attempts to catch insane values and replace them with failsafe values. Unfortunately, the artifact from e77a6c3b6cb8e127c984d6a59449b20a4bb8f18b has been tested by the reporting user, and didn't meet this objective.

My reading of the logic flow in the patched file is

```
435 if (is_NVIDIA(prop.vendor)) {
509 else if (is_AMD(prop.vendor)) {
567 else if (is_intel(prop.vendor)) {
590 else { //other vendor
616 David's patch
627 } endif
```
So far, I have only seen reports of the problem affecting AMD devices: the safety net could be applied in the line 509 code block, or could be applied globally outside this structure.

The effect on end users is that tasks are reported to have zero estimated runtime, and fail with error 197 EXIT_TIME_LIMIT_EXCEEDED. It only affects the minority of users who are running their devices at non-default clocks, but for that minority, it's critical.

Uplinger:
Closing and reopening to trigger travis-ci

RichardHaselgrove:
At last, I've got a report back from one of the affected users:

```
2/11/2019 4:54:29 PM |  | OpenCL: AMD/ATI GPU 0: AMD Radeon(TM) Vega 8 Graphics (driver version 2766.5 (PAL,HSAIL), device version OpenCL 2.0 AMD-APP (2766.5), 7206MB, 7206MB available, 100 GFLOPS peak)
2/11/2019 4:54:29 PM |  | [coproc] AMD OpenCL reported bad GPU peak FLOPS 43980464128000000.000000; using 100000000000.000000
2/11/2019 4:55:42 PM | Milkyway@Home | Aborting task de_modfit_84_bundle5_3s_NoContraintsWithDisk200_1_1548330302_5851041_0: exceeded elapsed time limit 19.99 (45115500.00G/2323910.58G)
2/11/2019 4:56:03 PM | Milkyway@Home | Aborting task de_modfit_80_bundle5_3s_NoContraintsWithDisk200_1_1548330302_5850766_0: exceeded elapsed time limit 18.10 (42060400.00G/2323910.58G)
```
Given the log times, I assume that the two MilkyWay tasks were allocated and downloaded when the old, faulty, client was in operation. He's probably down to a low daily quota, but I'll watch for the next round of tasks.

The format of the 'bad flops' message is clunky - I never know what we really do with a micro-flop - but it's a debug message, and we can live with it. Not a blocker.

JuhaSointusalo:
I had tested Nvidia and Intel code earlier with debugger (couldn't figure out a better way). Good to have a real test report.

For the server side, there is also the C++ version of plan classes. But that has a "todo: get rid of this" comment and it's not on par with the XML version feature-wise anyway. So I guess updating the XML version is enough. I'm not sure if plan class code is the right place for the sanity check. Maybe after parsing scheduler request would have been better. But the location of the sanity check is not a huge issue so I'll merge this.

JuhaSointusalo:
@davidpanderson

I was looking at the tasks from one of the affected users and noticed that he is still getting too low runtime estimates for Milkyway tasks even after he changed the client to the one from this PR. And then the importance of the server changes in this PR dawned on me. Without patching server it will recompute the peak flops using the bad input values from the client and ends up with bad peak flops and bad runtime estimates. So the server must be patched as well. Correct?

If that is correct then I guess @lfield has to do a server hotfix release.

RichardHaselgrove:
The server patch component of this PR has now been deployed and tested on the SETI@Home Beta server: the test was successful. Tasks run on https://setiweb.ssl.berkeley.edu/beta/show_host_detail.php?hostid=87175 show evidence of the faulty driver, but have been allowed to run to completion.

PR #3003, opened 2019-02-04
[Android] Add option to manually enter Account Manager URL.

### LINKED ISSUE #1582: Android - Allow adding other account managers
The Windows client allows to enter an account manager URL, so it's very flexible.
On Android (V 7.4.53) currently only 3 account managers are shown and can be used:

BAM    @   http://bam.boincstats.com/
http://gridrepublic.org/
http://www.extremadurathome.org/extremadurathome/

Would it be possible to allow adding an account manager by URL ?
(similar to the existing 'Add project by URL')

Or to include the Gridcoin.co Pool account manager @  http://pool.gridcoin.co   in the list of account managers ?

If the 2nd option is implemented please note

> How do I fix "[error] Inconsistent signing key from account manager"?
> Please make sure to use "**http://pool.gridcoin.co**" as the URL and NOT "http://pool.gridcoin.co/", this is a problem with how BOINC treats URLs.

from   http://pool.gridcoin.co/faq/#how_do_i_fix_inconsistent_signing_key_from_account_manager

adamradocz:
Fixes #1582

**Description of the Change**
This is the first part of the #2867, to lets that issue moving. I don't think that the 2nd part is feasible with the current design.
 
Added the option to use a custom account manager.

**Release Notes**
[Android] Allow entering custom Account Manager URL.

### DIFF
--- a/android/BOINC/.idea/misc.xml
+++ b/android/BOINC/.idea/misc.xml
@@ -35,5 +35,5 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
 </project>
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/AcctMgrFragment.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/AcctMgrFragment.java
@@ -41,6 +41,7 @@ import android.view.View;
 import android.view.View.OnClickListener;
 import android.view.ViewGroup;
 import android.view.Window;
+import android.widget.AdapterView;
 import android.widget.ArrayAdapter;
 import android.widget.Button;
 import android.widget.EditText;
@@ -55,6 +56,7 @@ public class AcctMgrFragment extends DialogFragment {
     private boolean asIsBound = false;
 
     private Spinner urlSpinner;
+    private EditText urlInput;
     private EditText nameInput;
     private EditText pwdInput;
     private TextView warning;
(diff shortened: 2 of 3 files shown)


brevilo:
Hi Adam, I'm not sure what the "first part" actually comprises. Can you share one (or more) screenshots to show the effect? Thanks.

adamradocz:
screenshot_20181211-135430_boinc

brevilo:
Ok, thanks.

But now I'm confused. I had a close look at #2867 again and it seems as if you have resolved all issues now:
   * ✅ There's a list of service names to choose from, [instead of](https://github.com/BOINC/boinc/pull/2867#pullrequestreview-180695300) URLs
   * ✅ I don't see any hard-coded values in this PR so you probably [use the XML file instead](https://github.com/BOINC/boinc/pull/2867#issuecomment-446148933), avoiding redundant service list maintenance

I'm curious, [previously](https://github.com/BOINC/boinc/pull/2867#issuecomment-446646613) using the canonical service list (XML file) couldn't be used unless the client was already attached to a first project, or it would crash. Did you resolve that? If so, can't we also close #2867 after merging this PR? If not, what am I missing?

adamradocz:
I resolved all the issues except one. It's working only if at least 1 project has been added. Because if no project has been added, then at the initial starting phase, the `Monitor` class isn't running yet.

brevilo:
Ok, so this is indeed about the [crash](https://github.com/BOINC/boinc/pull/2867#issuecomment-446646613).

However, I don't see any provisions that prevent a crash. If that's indeed the case, please add a check that verifies that at least one project is already added, and please show a popup letting the user know if that's not yet the case.

AenBleidd:
Personally I don't happy with such solution. This issue need to be solved, Unfortunately I'm overloaded on my job and can't help Adam with this.

adamradocz:
I totally agree with @AenBleidd. Plus it is confusing in terms of user experience.

brevilo:
You're both right, but why this new PR then? It's effectively in the same state as #2867, right? If so, I suggest to close this one and follow up on #2867 instead.

adamradocz:
I opened a new PR just to add the ability to add custom account managers. This will resolve a lot of headaches for people who want to use any other AM which not in the list right now.

brevilo:
Guys, you're really losing me here. We seemed to be in agreement that a crash in case of a lacking project attachment isn't acceptable. Yet, if there is no crash then #2867 would be resolved. I'm really confused as I think I'm reading contradictory statements.

So, please, explain to me what works and what doesn't in/after this PR.

adamradocz:
This PR is not crashing because it doesn't use the official project list.

brevilo:
Ok, **one** mystery solved then. Yet, I don't see where the service/URL mapping list is held and I'm still lacking a comprehensive description of what this PR does and doesn't and how it does what it does. 

Unfortunately we've run out of time now as I'm out of office for a week. Hopefully someone else can review this, if it's up for review already (no one assigned).

adamradocz:
This PR adds the most basic functionality to be able to enter an account manager URL.
Currently, if you want to use an AM, you have to choose one from the list. If you want to use anything else than that basically you screwed, because there is no input option.

brevilo:
Ok, thanks.

PR #3010, opened 2019-02-08
[wip] build: fix building Manager with wxWidgets 3.1

### LINKED ISSUE #2271: Manager fails to link with wxWidgets 3.1
When building Manager with wxWidgets 3.1 linking fails with the following error:

```
  CXXLD    boincmgr
boincmgr-NoticeListCtrl.o: In function `CNoticeListCtrl::Create(wxWindow*)':
/home/juha/dev/boinc/build-wx3.1.0/clientgui/../../src/clientgui/NoticeListCtrl.cpp:87: undefined reference to `wxWebViewNameStr'
/home/juha/dev/boinc/build-wx3.1.0/clientgui/../../src/clientgui/NoticeListCtrl.cpp:87: undefined reference to `wxWebViewBackendDefault'
/home/juha/dev/boinc/build-wx3.1.0/clientgui/../../src/clientgui/NoticeListCtrl.cpp:87: undefined reference to `wxWebViewDefaultURLStr'
/home/juha/dev/boinc/build-wx3.1.0/clientgui/../../src/clientgui/NoticeListCtrl.cpp:87: undefined reference to `wxWebView::New(wxWindow*, int, wxString const&, wxPoint const&, wxSize const&, wxString const&, long, wxString const&)'
boincmgr-NoticeListCtrl.o: In function `wxEventTableEntry':
/home/juha/dev/wxwidgets/inst-3.1.0/usr/local/include/wx-3.1/wx/event.h:3182: undefined reference to `wxEVT_WEBVIEW_NAVIGATING'
/home/juha/dev/wxwidgets/inst-3.1.0/usr/local/include/wx-3.1/wx/event.h:3182: undefined reference to `wxEVT_WEBVIEW_ERROR'
boincmgr-DlgItemProperties.o: In function `CDlgItemProperties::CDlgItemProperties(wxWindow*)':
/home/juha/dev/boinc/build-wx3.1.0/clientgui/../../src/clientgui/DlgItemProperties.cpp:65: undefined reference to `wxWebViewNameStr'
/home/juha/dev/boinc/build-wx3.1.0/clientgui/../../src/clientgui/DlgItemProperties.cpp:65: undefined reference to `wxWebViewBackendDefault'
/home/juha/dev/boinc/build-wx3.1.0/clientgui/../../src/clientgui/DlgItemProperties.cpp:65: undefined reference to `wxWebViewDefaultURLStr'
/home/juha/dev/boinc/build-wx3.1.0/clientgui/../../src/clientgui/DlgItemProperties.cpp:65: undefined reference to `wxWebView::New(wxWindow*, int, wxString const&, wxPoint const&, wxSize const&, wxString const&, long, wxString const&)'
collect2: error: ld returned 1 exit status

```

This is because BOINC configure doesn't request any particular set of wxWidgets components and wxWebView is no longer part of the standard set in 3.1.

3.0:
```
$ wx-config --libs
-L/usr/local/lib -pthread   -lwx_gtk2u_xrc-3.0 -lwx_gtk2u_webview-3.0 -lwx_gtk2u_html-3.0 -lwx_gtk2u_qa-3.0 -lwx_gtk2u_adv-3.0 -lwx_gtk2u_core-3.0 -lwx_baseu_xml-3.0 -lwx_baseu_net-3.0 -lwx_baseu-3.0
```
3.1
```
$ wx-config --libs
-L/usr/local/lib -pthread   -lwx_gtk2u_xrc-3.1 -lwx_gtk2u_html-3.1 -lwx_gtk2u_qa-3.1 -lwx_gtk2u_adv-3.1 -lwx_gtk2u_core-3.1 -lwx_baseu_xml-3.1 -lwx_baseu_net-3.1 -lwx_baseu-3.1
```

JuhaSointusalo:
**WORK IN PROGRESS - DO NOT MERGE!!!**

**Description of the Change**
Link Manager with wxWebView if wxWidgets was built with it.
Update wxWidgets M4 macros.

Fixes #2271 

**Alternate Designs**
None

**Release Notes**
Fix building Manager with wxWidgets 3.1.

### DIFF
--- a/m4/boinc_wxwidgets.m4
+++ b/m4/boinc_wxwidgets.m4
@@ -9,9 +9,11 @@ AC_DEFUN([BOINC_OPTIONS_WXWIDGETS],[
      [])
 
    AM_OPTIONS_WXCONFIG
-     AM_PATH_WXCONFIG([$1],
+   WX_CONFIG_CHECK([$1],
        [_ac_cv_have_wxwidgets=yes],
-       [AC_MSG_ERROR([No suitable wxWidgets library found.])])
+       [AC_MSG_ERROR([No suitable wxWidgets library found.])],
+       , ,
+       [webview])
    AC_CACHE_CHECK([if wxWidgets works],[ac_cv_have_wxwidgets],
     [ac_cv_have_wxwidgets="${_ac_cv_have_wxwidgets}"])
    AC_CACHE_SAVE
--- a/m4/wxWidgets.m4
+++ b/m4/wxWidgets.m4
@@ -1,13 +1,48 @@
 dnl ---------------------------------------------------------------------------
+dnl Author:          wxWidgets development team,
+dnl                  Francesco Montorsi,
+dnl                  Bob McCown (Mac-testing)
+dnl Creation date:   24/11/2001
+dnl ---------------------------------------------------------------------------
+
+dnl ===========================================================================
+dnl Table of Contents of this macro file:
+dnl -------------------------------------
+dnl
+dnl SECTION A: wxWidgets main macros
+dnl  - WX_CONFIG_OPTIONS
+dnl  - WX_CONFIG_CHECK
+dnl  - WXRC_CHECK
+dnl  - WX_STANDARD_OPTIONS
+dnl  - WX_CONVERT_STANDARD_OPTIONS_TO_WXCONFIG_FLAGS
+dnl  - WX_DETECT_STANDARD_OPTION_VALUES
+dnl
+dnl SECTION B: wxWidgets-related utilities
+dnl  - WX_LIKE_LIBNAME
+dnl  - WX_ARG_ENABLE_YESNOAUTO
+dnl  - WX_ARG_WITH_YESNOAUTO
+dnl
+dnl SECTION C: messages to the user
+dnl  - WX_STANDARD_OPTIONS_SUMMARY_MSG
+dnl  - WX_STANDARD_OPTIONS_SUMMARY_MSG_BEGIN
+dnl  - WX_STANDARD_OPTIONS_SUMMARY_MSG_END
+dnl  - WX_BOOLOPT_SUMMARY
+dnl
+dnl The special "WX_DEBUG_CONFIGURE" variable can be set to 1 to enable extra
+dnl debug output on stdout from these macros.
+dnl ===========================================================================
+
+
+dnl ---------------------------------------------------------------------------
 dnl Macros for wxWidgets detection. Typically used in configure.in as:
 dnl
 dnl     AC_ARG_ENABLE(...)
 dnl     AC_ARG_WITH(...)
 dnl        ...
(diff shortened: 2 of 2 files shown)


JuhaSointusalo:
The updated macros need polishing and upstreaming, the unicode stuff needs to be left for `WX_CONFIG_CHECK` (because right now the code checks for wxWidgets twice and the later check is missing optional libs) and it all needs to be tested.

ChristianBeer:
I'm investigating this issue too and will use some of the changes to the upstream wxwidgets macros made in this PR. This way it is clearer what needs to be pushed upstream to the wxwidgets developers.I'm also looking into the unicode check. I believe this is unnecessary now as this is done by the upstream macros. The m4/wxwidgets.m4 macros in BOINC (I suppose originally copied from upstream) were not updated since wxWidgets 2.6 (in 2005).

ChristianBeer:
I finished: https://github.com/BOINC/boinc/compare/cb_wxwidgets_31_compatibility please take a look. I basically did the same code changes but made more atomic commits. I was looking into getting rid of the second check for the unicode stuff but left it as is for the moment.

I don't know if we ever want a non-unicode wxWidgets build and just make a hard dependency on wxWidgets to be unicode enabled. That would get rid of a lot of code. Thus it seemed a bigger change worthy of it's own discussion and PR. As of now we don't loose anything by just keeping the second check as it is.

JuhaSointusalo:
I would like that the changes done to support optional libs in `WX_CONFIG_CHECK` is upstreamed first, copied back once upstream is happy and commits rewritten so that it looks like we never customized it. If there ever is need to update the file again it should be less work if there is no custom code in the file or in the history.

For upstreaming the optional libs stuff needs adding comments about it and checking that it's added to all the right places.

As for non-Unicode builds, I'd love to throw that option away. wxWidgets build instructions discourages using non-Uncode build and says it's going away in the future. A good excuse to remove the option from BOINC would be if it didn't work. Unfortunately Manager can be built with non-Unicode wxWidgets and it still works (at least on otherwise Unicode enabled Linux).

ChristianBeer:
I agree that we should push this upstream but we don't need to keep our history clean. We already use wxWidgets 3.1 on Mac and should be ready to update it on Linux and Windows too. Making the build system ready for that now has a higher priority in my opinion than pushing the change upstream, wait for integration and pulling it back into BOINC.

One thing we need to change for upstream is that the `--optional-libs` argument was introduced with wxWidgets 2.9 so it doesn't work with 2.8 which needs to be taken care of. The question to me is who is going to make the PR against upstream? Are we integrating my branch into BOINC before upstream accepts the PR?

JuhaSointusalo:
Let's sort out upstream later.

JuhaSointusalo:
I polished the changes to `wxwin.m4` and opened a pull request.

I suppose you could open a pull request for your changes. At some point I'm going to need to build on the Manager updates you have done.

PR #3011, opened 2019-02-09
Remote job submission: add logging

davidpanderson:
Remote job submission requests and errors are now logged to log_host/submit_log.txt.

### DIFF
--- a/html/inc/submit_util.inc
+++ b/html/inc/submit_util.inc
@@ -21,6 +21,26 @@
 
 require_once("../inc/submit_db.inc");
 
+$log_file = null;
+$verbose = true;
+
+function log_write($x) {
+    global $verbose, $log_file;
+
+    if (!$verbose) return;
+    if (!$log_file) {
+        $hostname = gethostname();
+        $log_dir = "../../log_$hostname";
+        if (!file_exists($log_dir)) {
+            mkdir($log_dir);
+        }
+        $log_file = fopen("$log_dir/submit_log.txt", "a");
+        if (!$log_file) return;
+    }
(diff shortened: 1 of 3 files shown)


davidpanderson:
can someone please merge this?

html/inc/submit_util.inc
```diff
@@ -21,6 +21,26 @@
...
+    global $verbose, $log_file;
+
+    if (!$verbose) return;
+    if (!$log_file) {
+        $hostname = gethostname();
+        $log_dir = "../../log_$hostname";
+        if (!file_exists($log_dir)) {
+            mkdir($log_dir);
+        }
+        $log_file = fopen("$log_dir/submit_log.txt", "a");
```
lfield: This fails with `fopen(../../log_boincai05.cern.ch/submit_log.txt): failed to open stream: No such file or directory in <b>/boincdata/boinc/project/lhcathome-dev/html/inc/submit_util.inc`

Note that the existing directory is log_boincai05 so $hostname should not contain cern.ch

lfield: The issue is that the log dir should come from the configuration. Elsewhere it uses
parse_config(get_config(), "<log_dir>");

lfield: To be consistent the file should be called submit.log rather than submit_log.txt.

lfield:
Please could you briefly explain the difference to the remote_submission.log?

davidpanderson:
I fixed all the issues.

lfield:
Tested and it looks good.

PR #3012, opened 2019-02-09
Curl options

serval2412:
Add ssl_verifyhost and ssl_verifypeer options

### DIFF
--- a/client/acct_mgr.h
+++ b/client/acct_mgr.h
@@ -169,12 +169,11 @@ struct ACCT_MGR_OP: public GUI_HTTP_OP {
     int parse(FILE*);
     virtual void handle_reply(int http_op_retval);
 
-    ACCT_MGR_OP(GUI_HTTP* p) {
+    ACCT_MGR_OP(GUI_HTTP* p) : global_prefs_xml("") {
         gui_http = p;
         via_gui = false;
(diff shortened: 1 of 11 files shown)


serval2412:
My fault, I hadn't notice my local branch wasn't up to date. I'm not good enough to 

I submitted this patch because I got an error about certificate on a machine. Of course, the pb must be on this machine since I don't reproduce it on another machine but wanted to provide a mean to bypass these options.

Anyway, I don't need to work more on this since it seems the patch will be refused anyway.

So let's abandon this one and sorry for the noise.

JuhaSointusalo:
@serval2412

From the description you now gave it might be that the certificate bundle on that problem machine is outdated. If the problem is that simple then yes, I'd say the patch would get rejected. But sometimes people have problems and we end up patching BOINC.

For example, AMD released a driver that reports, via OpenCL, that the GPU runs at 50 THz (or something equally silly). While that is definitely not our fault we have a pull request open that adds a workaround for this problem (or rather the silly peak GFLOPS that is the result of silly clock frequency) and we might need to do a hotfix release because of that.

So, always open a new issue here or a thread on forums to discuss the problem first. If it looks like it's going to need code changes only then start coding.

PR #3019, opened 2019-02-12
client: increase finish-file timeout

### LINKED ISSUE #3017: Finish file present too long
When running BOINC under minium priority and the system is under heavy load and a science app finishes, BOINC aborts the task and does not submit the result. This is unfortunate, because the task is already successfully done, yet the result is discarded.

The app does not need to do any complicated processing  at exit. I have `exit(boinc_finish(0));` at the end my app and it still triggers. This can also happen during swap thrashing.

Relevant line: https://github.com/BOINC/boinc/blob/f14d96d2be2e5a290177f7b5391b0802fbd3756c/client/app_control.cpp#L140

I propose to check the process CPU time instead of wall clock time. Alternatively, just kill the app, but do not abort the task. After all, boinc_finish wass called with success and all output files are present.

davidpanderson:
When an app finishes, it writes a "finish file",
which ensures the client that the app really finished.

If the app process is still there N seconds after the finish file appears,
the client assumes that something went wrong, and it aborts the job.

Previously N was 10.
This was too small during periods of heavy paging.
I increased it to 300.

It has been pointed out that if the app creates the finish file,
and its output files are present,
it should be treated as successful regardless of whether it exits.
This is probably true, but right now we don't have a mechanism
for killing a job and marking it as success.
The longer timeout makes this moot.

Fixes #3017

### DIFF
--- a/client/app_control.cpp
+++ b/client/app_control.cpp
@@ -135,9 +135,16 @@ bool ACTIVE_TASK_SET::poll() {
             ACTIVE_TASK* atp = active_tasks[i];
             if (atp->task_state() == PROCESS_UNINITIALIZED) continue;
             if (atp->finish_file_time) {
-                // process is still there 10 sec after it wrote finish file.
-                // abort the job
-                atp->abort_task(EXIT_ABORTED_BY_CLIENT, "finish file present too long");
+                if (gstate.now - atp->finish_file_time > FINISH_FILE_TIMEOUT) {
+                    // process is still there 5 min after it wrote finish file.
+                    // abort the job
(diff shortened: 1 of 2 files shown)


JuhaSointusalo:
> It has been pointed out that if the app creates the finish file,
> and its output files are present,
> it should be treated as successful regardless of whether it exits.

FWIW, it has also been argued that if an app hangs at exit there's no trusting anything that happened before the hang. The context in that discussion was GPU apps. If GPU drivers cause an app to hang at exit then the drivers might not have processed computing commands correctly either.

PR #3020, opened 2019-02-13
client: sanity-check job runtime limits

davidpanderson:
A job is assigned a max runtime as:
max_elapsed_time = rp->wup->rsc_fpops_bound/rp->avp->flops
The purpose is to eventually abort jobs that are in an infinite loop.

Various problems (e.g. bad GPU peak FLOPS calculations)
can cause this limit to be too small, e.g. one second,
in which case the job is aborted almost immediately.

In this change, if the calculated limit is < 2 minutes,
it's assumed to be in error, a limit of 30 minutes is used instead,
and an event log message is written.

Of course the underlying problem still must be addressed.
But this change will, in some cases, prevent a situation where
thousands of jobs are dispatched and immediately aborted.

**Release Notes**

If calculated job time limit is < 2 minutes, use a limit of 30 minutes instead.


AenBleidd:
> So I'd set the safe default value to 12 hours. 

If I remember correctly, GPUGrid tasks on not top-performance cards can run even longer, for a couple days. I can be wrong here because I have no GPUs that can run GPUGrid tasks for more than 3 years now.

Maybe this is not this case but anyway

JuhaSointusalo:
Yes, my 940MX takes three or four days to complete GPUGRID tasks.

The 12 hours, or whatever number David comes up with, is a compromise for when the server can't be trusted. We can't have the safe default be long enough for all projects because RNAWorld has tasks that take _a year_ to run.

davidpanderson:
I changed it to 12 hours.
The goal here, BTW, is just to slow down the whole system when something
is resulting in 1-second runtime bounds, so that clients don't chew through millions of jobs.

JuhaSointusalo:
> The goal here, BTW, is just to slow down the whole system when something
> is resulting in 1-second runtime bounds, so that clients don't chew through millions of jobs.

Oh. I thought the goal was to make sure a host with buggy GPU driver can participate in at least some projects.

Anyways. I requested two changes. You still need to do something about the log message. See the review comment for details.

davidpanderson:
The message is printed in pathological situations, e.g. the host has a ridiculous FLOPS estimate.  We want the user to see it.

lfield:
@JuhaSointusalo This PR provides protection for a scenario that could overload the server. It does not fix the cause as that is not the purpose and there could be other causes that we have not found. While I think that the setting of the limits is sub-optimal as the server stats page already shows min, mean and max runtimes, it is good enough for now. The messages are helpful for gathering stats on how often this condition occurs. Hence I see not reason why this PR should not be merged.

TheAspens:
As all remaining objections to this pull request have been removed, I'm proceeding with the merge.

RichardHaselgrove:
Just seen my first occurrence of this.
```
09-Mar-2021 18:35:01 [Einstein@Home] Starting task h1_0600.00_O2C02Cl4In0__O2MDFS3a_Spotlight_600.35Hz_199_2
09-Mar-2021 18:52:43 [Einstein@Home] Computation for task h1_0600.00_O2C02Cl4In0__O2MDFS3a_Spotlight_600.35Hz_199_2 finished
09-Mar-2021 18:52:43 [Einstein@Home] Elapsed time limit 103.834983 < 120.000000; setting to 43200.000000
```
The task was reported to the Einstein server as https://einsteinathome.org/task/1084172095

The interesting thing for me is that the Einstein website reports a perfectly normal (this host, this GPU, this app, this task type) run time of 1,062.47 seconds - which matches the duration from start to finish.

So where did the client get `Elapsed time limit 103.834983` from? There's nothing unusual in the log.

PR #3028, opened 2019-02-22
Removed "Exit from BOINC Manager", "Shut down connected client" menu entries on Linux

Germano0:
**Description of the Change**
On Linux distributions, BOINC runs as a service. Users must not be able to try stopping the service from Manager exit menu entry.
This leads to unexpected behaviour, like:
- service being killed;
- service still running.

Moreover, the Manager will no longer be able to connect to the client, unless the user connects to 127.0.0.1. Then if the Manager is connected to the client by using 127.0.0.1 address, the "Exit from BOINC Manager" entry will not show any frame asking the user if he wants to stop the service.
"Close Window" menu entry is enough to let the user close the Manager window.

**Alternate Designs**

**Release Notes**
Removed "Exit from BOINC Manager", "Shut down connected client" menu entries on Linux


AenBleidd:
@Germano0, please remove *.orig files from this pull request

Germano0:
> @Germano0, please remove *.orig files from this pull request

Right, I am sorry

Germano0:
> Also what will happen if before installing this version of BOINC Manager user already has option on Exit Confirmation Dialog Window 'Stop running tasks when exiting the BOINC Manager 'checked?

Probably the client will keep running because the only action triggered is [wxWindow::Close](https://docs.wxwidgets.org/trunk/classwx_window.html#a3e44f4a494fc9ef4346c4fba70c8de0c)

**Source code of BOINC Manager action for wxWindow::Close**
[void CBOINCBaseFrame::OnCloseWindow(wxCommandEvent& WXUNUSED(event)) ](https://github.com/BOINC/boinc/blob/master/clientgui/BOINCBaseFrame.cpp#L345)

**Source code of BOINC Manager action for wxWindow::wxID_EXIT**
[void CBOINCBaseFrame::OnExit(wxCommandEvent& WXUNUSED(event))](https://github.com/BOINC/boinc/blob/master/clientgui/BOINCBaseFrame.cpp#L354)

Germano0:
I deleted the two .orig files, but Github is still stuck on "changes requested"

JuhaSointusalo:
First of all, not all our Linux users run BOINC as service. You can argue to us that those users will just have to learn the Linux way but you are arguing to the wrong people. You'd need to tell our users how they can use BOINC. Good luck with that. Expect tar and feathers.

Second, if there really is a problem connecting to localhost after the client has been stopped and restarted then it's a bug and it needs to be fixed by, well, fixing it. Removing a menu option is not fixing the bug.

As for the changes. If only this PR is merged then users can't exit Manager. That is unacceptable. If this PR is merged along with your other PR that changes the Close window action to exit then users can exit Manager by selecting a menu entry that has a misleading name. That's bad UI/UX.

Another problem is that there is more than one way to exit Manager and have it stop the client. Trying to handle it by taking out menu options one by one seems hackish to me and a good way to miss some code paths.

Before you rush into making further changes. Right now I'm leaning strongly towards a no. While we have users stopping the service without realizing there is a service I don't believe this is the right way to deal with that. I think a better way would be have a better integration with service managers. That is, to continue to have the menu entries and dialogs in Manager but forward start and stop actions to service manager if the client is installed as a service. And to have that across platforms.

RichardHaselgrove:
I agree with @JuhaSointusalo 

Please cross-reference to the discussion in https://github.com/BOINC/boinc/pull/2969#issuecomment-466423510 and following.

truboxl:
Is removing "Shutdown connected client" necessary?
If I want to remotely shutdown a client on another PC from the GUI Manager then what should I do?

I may be the niche-st user out there using the Linux BOINC Manager as a standalone application to control other platforms of BOINC Client,
and I seriously do not want starting the Manager also starts the Client (meaning treating Manager and Client as separate programs).

It always dawns on me that I can't just
install BOINC Manager --> install BOINC Manager + BOINC Client
unlike
install BOINC Client --> install BOINC Client
in major Linux distros.

AenBleidd:
> If I want to remotely shutdown a client on another PC from the GUI Manager then what should I do?

If I remember correctly, there is no option to shutdown remote client in BOINC Manager. This option is available for local client only

RichardHaselgrove:
There is in Windows - I've just done it. What you can't do is restart the remote client remotely - fortunately this one was on the desk in front of me (which is why I'm sure it shut down, leaving a blank Manager on its own monitor).

AenBleidd:
Damn, I forgot about menu option. I just told about Exit Manager window. My bad.
@RichardHaselgrove, thank you for your answer

lfield:
The manager should not shutdown the client on Linux. By default it is run as a service. For those who start it a different way, they can use the same method to stop it. There is also an asymmetry with providing the stop functionally but no start functionally. I agree with @JuhaSointusalo that the connection issues is a bug but should be addressed separately. 

My only comment therefore would be about the code comments. The bit about the manager connection should be removed. And the rest should be shortened to one line. After making those changes I am happy to merge.

Germano0:
Aborting this pull request since I erased my fork repository so I am no longer able to edit the commit.
I opened a new pull request https://github.com/BOINC/boinc/pull/3094

tckosvic:
I find now that closing boinc manager shuts down the client.  I find this need to keep the manager running very annoying.  I currently use openSUSE 15.2.
Before I upgraded to 15.2, while running 15.,1 the boinc manager could be started and stopped and the client would keep running.  I dont know what version of boinc manager was running under 15.1 but i would like to return to it.  I accrued  1.5 million statistics points running 15.1  and starting boinc manager countless times.

The boinc manager with the annoying behavior is 7.16.6 version.  In the file tab it says "shut down connected client"  with no button for yes or no.

I would like to revert back to allow boinc client to keep running when manager it stopped.  I think services should only be turned off by OS system root purposefully and a manager turning off a service client is wrong to me.

I saw some discussion about this here and thought it a good place to post.

tom kosvic

LUGO022 (commented):
Thank you

PR #3034, opened 2019-02-26
Make server_status.php compatible with PHP 5.3

Rytiss:
**Description of the Change**
Array dereferencing is only supported in PHP 5.4+. Use a variable instead.

See http://php.net/manual/en/language.types.array.php#example-63

**Release Notes**
N/A

Relates to pull request #2967.

### DIFF
--- a/html/user/server_status.php
+++ b/html/user/server_status.php
@@ -188,9 +188,10 @@ function show_status_html($x) {
     
     global $server_version;
     if ( isset($server_version) ) {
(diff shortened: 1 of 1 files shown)


davidpanderson:
#2967 subsumes this fix; let's use that instead.

TheAspens:
@Rytiss - since I merged #2967 should we close this pull request?

Rytiss:
Superseded by #2967, closing.

PR #3043, opened 2019-03-02

ChristianBeer:
**Description of the Change**
Current translations from Transifex


android/BOINC/app/src/main/res/values-de/strings.xml
```diff
@@ -1,396 +1,364 @@
...
+  <string name="nonexcl_dialog_text">Eine weitere Volunteer Computing App läuft auf diesem Gerät. Es kann immer nur eine App verwendet werden.</string>
+  <string name="nonexcl_dialog_exit">Beenden</string>
+  <!--social integration-->
+  <string name="social_invite_button">Freunde einladen</string>
+  <string name="social_invite_intent_title">Wie möchten Sie andere informieren?</string>
+  <string name="social_invite_content_title">Ich betreibe Forschung mit meinem Smartphone!</string>
+  <string name="social_invite_content_body">Ich benutze mein %1$s um für die Forschung zu rechnen. Du kannst das auch! Hier gibt es die App: %2$s</string>
+  <!--first parameter: device manufacturer, second: URL-->
+  <string name="social_invite_content_url_google">https://play.google.com/store/apps/details?id=edu.berkeley.boinc</string>
+  <string name="social_invite_content_url_amazon">https://www.amazon.de/Space-Sciences-Laboratory-U-C-Berkeley/dp/B00DCCGLAG/ref=sr_1_1?ie=UTF8&amp;qid=1467183304&amp;sr=8-1&amp;keywords=boinc</string>
```
JuhaSointusalo: I'm not comfortable with URLs like this. I can't tell if it's got an affiliate ID baked into it or is it some search query.

When I went to the previous URL I got
https://www.amazon.de/gp/product/B00DCCGLAG/ref=mas_dl/257-3709454-5925656
on the first try and
https://www.amazon.de/gp/product/B00DCCGLAG/ref=mas_dl
on the second try. Likewise with the original URL I got
https://www.amazon.com/gp/product/B00DCCGLAG/ref=mas_dl/135-0457377-5148255
on the first try and
https://www.amazon.com/gp/product/B00DCCGLAG/ref=mas_dl
on the second try.

Please clean up the URL and do the same for es and it-rIT which have similar odd URLs.

ChristianBeer: This is related to the search string used when looking on the regional amazon portal. I forgot to clean the URL when I changed the URL. I cleaned it on transifex and used this one: https://www.amazon.it/gp/product/B00DCCGLAG (exchanging the TLD).

ChristianBeer: I updated the German one but couldn't update the Italian one because the translation on transifex is not complete enough. The Spanish one is just a copy of the original string.

JuhaSointusalo: Thanks. The Spanish one is odd in repo but if it's the same as original on Transifex I guess that's ok. We'll just have repo updated later.

PR #3045, opened 2019-03-06
Changed SIGSTOP to SIGTSTP

### LINKED ISSUE #3026: The wrapper should use SIGTSTP rather than SIGSTOP
As far as I understand, it is [not possible](https://en.wikipedia.org/wiki/Signal_(IPC)#POSIX_signals) to [catch SIGSTOP](https://github.com/BOINC/boinc/blob/master/lib/proc_control.cpp#L250). SIGTSTP should be used instead.

lfield:
Fixes #3026

Changed SIGSTOP to SIGTSTP as SIGSTOP can not be caught. 

The use of  SIGTSTP allows the code run by the wrapper to act upon this signal such as pausing a running container.


TheAspens:
Seems to make sense to me.

PR #2835, opened 2018-11-20
client, mgr: add support for registering user's consent to project's terms of use

JuhaSointusalo:
This PR adds the missing pieces of code to the consent work done in #2504 making it possible for users to consent to project's terms of use using Manager and having that consent recorded in project database.

This is implemented by
- Adding `<name>` element to `<exchange_versions>` GUI RPC. The name is something like "BOINC Manager 7.15.0". `<major>`, `<minor>` and `<release>` in `<exchange_versions>` is now interpreted as API version.
- Adding `<consented_to_terms/>` element to `<create_account>` GUI RPC to mark that the project had terms of use and the user consented to the terms.
- Including the GUI name and consent in `create_account` Web RPC.

I didn't add consent functionality to `boinccmd` because I don't think we can do that. `boinccmd` doesn't know if the user has seen the terms. `boinccmd` could be made to print terms first and ask for consent but I see it as a very simple tool over the GUI RPCs. It just converts command line options to XML and pretty prints results.

Release Notes:
client, mgr: add support for registering user's consent to project's terms of use

PR #3049, opened 2019-03-11
[web][android][lib] Add conditional terms of use consent requirement to create_account RPC

brevilo:
**Description of the Change**
* Added a new config.xml setting "account_creation_rpc_require_consent" (default: disabled)
* Added condition to create_account RPC to return an error if consent details aren't given but required
* Defined a dedicated error number and default message (for C and Android use)

**See also**
* #2413
* #2835

**Release Notes**
Allow projects to require explicit consent for RPC-based account creation to comply with GDPR.

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/AcctMgrFragment.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/AcctMgrFragment.java
@@ -265,6 +265,9 @@ public class AcctMgrFragment extends DialogFragment {
             case BOINCErrors.ERR_INVALID_URL:
                 stringResource = R.string.attachproject_error_invalid_url;
                 break;
+            case BOINCErrors.ERR_ACCT_REQUIRE_CONSENT:
+                stringResource = R.string.attachproject_error_consent_required;
+                break;
             default:
                 stringResource = R.string.attachproject_error_unknown;
                 break;
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/utils/BOINCErrors.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/utils/BOINCErrors.java
@@ -47,4 +47,5 @@ public class BOINCErrors {
     public final static int ERR_BAD_PASSWD = -206;
     public final static int ERR_NONUNIQUE_EMAIL = -207;
     public final static int ERR_ACCT_CREATION_DISABLED = -208; // i.e. account creation currently disabled
+    public final static int ERR_ACCT_REQUIRE_CONSENT = -242; // project requires consent
 }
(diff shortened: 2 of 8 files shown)


drshawnkwang:
I don't have any comments to add. From the perspective of `create_account`, I think it's a good idea we start requiring consent to be set by the clients.

brevilo:
Just to explain why I tagged this as "major": right now we **don't** allow RPC-based sign-ups, that is, including the BOINC Manager as well BAM!, etc. That's obviously a major hurdle for recruiting new volunteers. Integrating PR would allow us to reenable RPC-based sign-up as we can impose filtering based on explicit consent.

drshawnkwang:
@JuhaSointusalo , that's a really good point, I'm sorry I missed it. If account creation requires consent, then the flow of this RPC needs to be altered.

brevilo:
@JuhaSointusalo nice catch. I'll update the PR.

brevilo:
Oh, wow, we have an actual integration test for `create_account.php`? Didn't expect that. I'll look into updating it to cover the changes in this PR...

ChristianBeer:
The current integration test only tests against a default project configuration. This obviously still works. It would be nice to add a test that checks the file with the GDPR project configuration. The actual test is here: https://github.com/BOINC/boinc-server-test/blob/master/tests/user-tests/create_account.php but I don't have an idea how to change the project configuration from within the unit tests.

brevilo:
> The actual test is here: https://github.com/BOINC/boinc-server-test/blob/master/tests/user-tests/create_account.php but I don't have an idea how to change the project configuration from within the unit tests.

Yep, that's what I figured in the meantime.

drshawnkwang:
Later today I will find the time to test out this new create_account RPC code with my development BOINC project that I used in the past to develop the 'opt-in constent' features.

brevilo:
Thanks Shawn!

drshawnkwang:
I've tested the create_account RPC logic with a devel project. Summary: it works as indended.

If I set the config setting `account_creation_rpc_require_consent` is TRUE, one cannot create an account without having the parameters `consent_flag` and `source` in the GET request. If the setting is FALSE, then one can create an account w/o these two parameters, and the user will be logged in w/o consenting. This is obviously not optimal; but It is the case where admins are expected to disable the create_account RPC with `disable_account_creation_rpc` set to TRUE, in order so that users may only create an account using the Web site.

Therefore admins who want to use `account_creation_rpc_require_consent` (set to TRUE) should also set `disable_account_creation_rpc` to FALSE.

brevilo:
Thanks guys! So who's going to merge it? 😄  @TheAspens, do you wanna have the final word...?

TheAspens:
I scanned it real quick and it looks good to me.  I'd recommend one of the formal approvers do the merge - @JuhaSointusalo  or @AenBleidd

brevilo:
Thanks everyone!

JuhaSointusalo:
@brevilo

I don't think you have documented this yet?

brevilo:
True, thanks for the reminder

brevilo:
Docs updated:
* https://boinc.berkeley.edu/trac/wiki/UserOptInConsent#ConsenttoTerms-of-Use
* https://boinc.berkeley.edu/trac/wiki/ProjectOptions#misc

PR #3052, opened 2019-03-12
Disabled Manager system tray icon for Linux builds

Germano0:
**Description of the Change**
Disables BOINC Manager system tray icon on Linux platforms.
On Linux BOINC client runs regardless the Manager is running or not, so having a Manager system tray icon may suggest the user that calculus are running only when the Manager is running, leading to misunderstandings.
Moreover removing Manager system tray icon, prevents also the user from accessing to tray icon right-clic menu that can let the user interfere with client service, as stated in pull request ["Removed "Exit from BOINC Manager", "Shut down connected client" menu entries on Linux"](https://github.com/BOINC/boinc/pull/3028)

**Release Notes**
Disablec BOINC Manager system tray icon on Linux platforms.

### DIFF
--- a/clientgui/AsyncRPC.cpp
+++ b/clientgui/AsyncRPC.cpp
@@ -944,11 +944,13 @@ void CMainDocument::HandleCompletedRPC() {
 
     if (m_bNeedTaskBarRefresh && !m_bWaitingForRPC) {
         m_bNeedTaskBarRefresh = false;
+#ifndef __WXGTK__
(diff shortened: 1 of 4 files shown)


Ferroin:
Adding to the reasoning here, the system tray icon API's in wxGTK just don't consistently work in most desktop environments on Linux anymore, at least with the versions of wxGTK BOINC is using (not sure about newer versions).  While I can't prove it, I would be willing to bet that most current BOINC users on Linux aren't able to use this functionality anyway.

PR #3058, opened 2019-03-13
web: make it possible to send HTML email

davidpanderson:


html/inc/email.inc
```diff
@@ -63,11 +64,18 @@ function send_email($user, $subject, $body, $body_html=null, $email_addr=null) {
...
-        if ($email_addr) {
-            return mail($email_addr, $subject, $body, $headers);
-        } else {
-            return mail($user->email_addr, $subject, $body, $headers);
+        if (!$email_addr) {
+            $email_addr = $user->email_addr;
+        }
+        if ($body_html) {
+            $body = "<html>\n"; 
+            $body .= '<body style="font-family:Verdana, Verdana, Geneva, sans-serif; font-size:12px;">\n';
```
TheAspens: Wouldn't it be better to leave off the style tag on the body element?  If a project is using HTML emails they are probably trying to replicate the styles they are using on their website and things like font and size may be something they want to control.

html/inc/email.inc
```diff
@@ -63,11 +64,18 @@ function send_email($user, $subject, $body, $body_html=null, $email_addr=null) {
...
+        if (!$email_addr) {
+            $email_addr = $user->email_addr;
+        }
+        if ($body_html) {
+            $body = "<html>\n"; 
+            $body .= '<body style="font-family:Verdana, Verdana, Geneva, sans-serif; font-size:12px;">\n'; 
+            $body = $body_html; 
+            $body .= "</body>\n"; 
+            $body .= "</html>\n"; 
+            $headers .= 'Content-type: text/html; charset=iso-8859-1' . "\r\n";
```
TheAspens: Given the charset iso-8859-1, if a user has registered a username with Chinese characters, how will this be rendered on the email client?

davidpanderson:
I changed it to UTF-8

html/inc/email.inc
```diff
@@ -63,11 +64,16 @@ function send_email($user, $subject, $body, $body_html=null, $email_addr=null) {
...
-        if ($email_addr) {
-            return mail($email_addr, $subject, $body, $headers);
-        } else {
-            return mail($user->email_addr, $subject, $body, $headers);
+        if (!$email_addr) {
+            $email_addr = $user->email_addr;
+        }
+        if ($body_html) {
+            $body = "<html><body>\n"; 
+            $body = $body_html;
```
TheAspens: You need to append to the string on this line.  At the moment this overwrites <html><body> from the previous line.

TheAspens:
Sorry for the delay in getting back to this.  I found a small bug that needs fixing otherwise it should be good to merge.

davidpanderson:
Fixed; thanks!

TheAspens:
Looks fine - merging.  Thanks for the change.

PR #3062, opened 2019-03-17
Manager: add account manager functions to simple view

davidpanderson:
Add AM functions (join, leave, synch) to simple view Tools menu, same as Advanced view.

Also did some code cleanup/reorg.
The simple view (CBOINCSimplFrame) and advanced view (CBOINCAdvancedFrame)
each had their own functions for invoking the attach wizard.
Instead, have only one function, and make it a member of CBOINCBaseFrame,
from which each of these classes derives.

Other stuff is duplicated between simple and advanced also,
and we should consolidate it at some point.

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/clientgui/AdvancedFrame.cpp
+++ b/clientgui/AdvancedFrame.cpp
@@ -1959,15 +1959,6 @@ void CAdvancedFrame::OnNotebookSelectionChanged(wxNotebookEvent& event) {
     wxLogTrace(wxT("Function Start/End"), wxT("CAdvancedFrame::OnNotebookSelectionChanged - Function End"));
 }
 
-void CBOINCBaseFrame::ResetReminderTimers() {
-#ifdef __WXMSW__
-    wxASSERT(m_pDialupManager);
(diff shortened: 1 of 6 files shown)


clientgui/AdvancedFrame.cpp
```diff
@@ -2118,8 +1959,7 @@ void CAdvancedFrame::OnNotebookSelectionChanged(wxNotebookEvent& event) {
     wxLogTrace(wxT("Function Start/End"), wxT("CAdvancedFrame::OnNotebookSelectionChanged - Function End"));
 }
 
-
-void CAdvancedFrame::ResetReminderTimers() {
+void CBOINCBaseFrame::ResetReminderTimers() {
```
AenBleidd: I believe this should be moved to clientgui/BOINCBaseFrame.cpp

davidpanderson: Done.

lfield:
Test on Ubuntu 18.10 and seems to work as intended.

PR #3063, opened 2019-03-18
Drupal: Fixed registration form terms-of-use checkbox

drshawnkwang:
**Description of the Change**
Fixes broken user registration form. 

**Release Notes**
N/A

https://dev.gridrepublic.org/browse/DBOINCP-497


brevilo:
Just to make sure, `boincuser_termsofuse_form_validate()` shouldn't be used here?

drshawnkwang:
No, that function is for a second form that is separate. I believe that form is for users who had accounts before terms-of-use was implemented. They are forced to see the form when the login. [It has the same construct](https://github.com/BOINC/boinc/blob/master/drupal/sites/default/boinc/modules/boincuser/includes/boincuser.forms.inc#L768). That was the bug: I changed the check if-statement in the function `boincuser_termsofuse_form_validate` and not the other `boincuser_register_validate`, which is being fixed in this PR.

brevilo:
Yeah, I understand the solution but just wanted to avoid the apparent code redundancy. But agreed, that could be followed up separately.

PR #3064, opened 2019-03-20
Manager: task list scrolling

### LINKED ISSUE #2147: Linux Manager: bug/workround no longer needed?
Commit 6479868b7c4c2efa87ff69d9e0910d43d2c33df7 introduced a workround for a wxWidget bug at wxWidget v3.0 / BOINC v7.4

Although that commit introduces the workround into the v7.4 release branch, the same code is still present in master. It has been asserted in testing that the workround is no longer needed, and itself now introduces unwanted side-effects.

RichardHaselgrove:
Reduce the number of times the task list scrolls to the bottom under routine operations.

Fixes #2147

**Description of the Change**
6479868b7c4c2efa87ff69d9e0910d43d2c33df7 (Linux only) scrolls the task list viewport to display the bottom line every time the number of items in the list changes during reporting/requesting new work. This is unnecessary, and goes beyond the design statement for the commit.

**Release Notes**
Linux only: eliminate unnecessary 'jump to bottom' events in Manager task list, caused by over-enthusiastic wxWidgets bug fix in 6479868b7c4c2efa87ff69d9e0910d43d2c33df7

### DIFF
--- a/clientgui/BOINCBaseView.cpp
+++ b/clientgui/BOINCBaseView.cpp
@@ -278,7 +278,9 @@ void CBOINCBaseView::OnListRender(wxTimerEvent& event) {
                     // to draw the bottom items.  This could happen
                     // if the list was scrolled near the bottom and
                     // the user selected "Show active tasks."
-                    m_pListPane->EnsureVisible(iDocCount - 1);
+                    if (m_pListPane->GetItemCount() <= m_pListPane->GetCountPerPage()) {
+                        m_pListPane->EnsureVisible(iDocCount - 1);
+                    }
 #endif
                     m_bNeedSort = true;
                 }


CharlieFenton:
@RichardHaselgrove Have you confirmed that the original drawing bug does not occur when there are far more items than fit on one page? Since I made this change over 4 years ago, I don't remember the details clearly but I don't recall this happening only when there was a small number of items. Since I don't have a Linux box, and setting up the test on a VM is a significant effort for me, I can't easily test this here.

RichardHaselgrove:
No, I haven't, for the same reason as you. What I did do was to test a Windows build with the `#ifdef __WXGTK__` taken out, and observe the 'jump to bottom' problem for myself: I can see why the Linux users are complaining - I certainly couldn't live with that myself. These are people with powerful, multi-GPU computers with possibly thousands of tasks ready to start, and contacting project servers every few minutes.

@ChristianBeer commented in https://github.com/BOINC/boinc/issues/2147#issuecomment-473265469 that "The workaround is still required to handle the initial problem but it seems the side-effect described by Richard are gone now.": unfortunately, other testers found that the side-effects were still present. If Christian or other testers could explore more deeply, that would be great. The simplest way is probably to over-fill the cache with tasks from a short-running project, and then set 'no new tasks'. Tasks will be reported as they complete, and the list will jump each time.

CharlieFenton:
Just to make sure we understand each other, I think there is no question that the workaround in 6479868 (I would not really call it a _fix_) does cause the annoying side effect that the display jumps to the end each time the number of items in the list changes. (Though this is not noticeable if the list was already scrolled to the end.) 

But the original problem as described in the associated comment in the code was a very serious one, so some sort of workaround or fix is definitely needed. What is needed is a better fix or workaround, or at least coming up with a way to call `EnsureVisible` only when actually needed.

Furthermore, it is not clear to me exactly how the list should behave when it is scrolled to items somewhere in the middle of the list and items are added to or removed from the list. For example, say the window is sized to show only 5 lines, but there are dozens of items in the list and we are currently viewing lines 10 through 14 (whose contents I will call items A through E.) If lines 9 through 11 are deleted, then those original items C, D and E (which were in lines 12 - 14) now become lines 8 - 10. Should the display now show items 6 - 10 so these 3 remaining items C, D and E stay at the bottom of the display, or should it show the newly numbered lines 10 - 14, which would scroll items C and D out of view? Another complication is that the order of the items in the list can change when values in the current sort column change. 

Back when I introduced this clumsy workaround, I remember thinking that the logic to do the correct thing when scrolled to the middle of the list might require some fairly complex logic.

lfield:
I have tested this on a local build and the scroll bar does not jump to the bottom when I get new tasks.

PR #3066, opened 2019-03-21
Fix bug: in abort/retire/expire batch, update state instead of projec…

### LINKED ISSUE #2793: Batches Failing To Be Retired
The python function `retire_batch(req)` seems not to work as expected. It looks like the code tries to update the column [state](https://github.com/BOINC/boinc/blob/master/html/inc/submit_util.inc#L146) but the actual column name is [project_state](https://github.com/BOINC/boinc/blob/master/db/schema.sql#L344). This also would affect `abort_batch` and `expire_batch`. If my understanding is correct I can provide a PR with the fix.

davidpanderson:
…t_state

Fixes #2793

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/html/inc/submit_util.inc
+++ b/html/inc/submit_util.inc
@@ -163,7 +163,7 @@ function abort_batch($batch) {
     foreach ($wus as $wu) {
         abort_workunit($wu);
     }
-    $batch->update("project_state=".BATCH_STATE_ABORTED);
+    $batch->update("state=".BATCH_STATE_ABORTED);
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


TheAspens:
This fix was already implemented by @lfield in pull request #3007 

I'm very confused why this shows this as a change.  If you look at the file in master: https://github.com/BOINC/boinc/blob/master/html/inc/submit_util.inc#L161 it already has these changes.

Is someone able to explain what is going on?

JuhaSointusalo:
> I'm very confused why this shows this as a change. If you look at the file in master: https://github.com/BOINC/boinc/blob/master/html/inc/submit_util.inc#L161 it already has these changes.

If you look at Insights -> Network it looks like this branch, dpa_submit7, predates the merge of #3007 and it also looks like this branch was used earlier for another PR.

davidpanderson:
Oops!  Ignore this change.

PR #3067, opened 2019-03-22
build: remove Fortran checks

JuhaSointusalo:
**Description of the Change**
While science apps may use Fortran BOINC itself doesn't use it making
the checks useless.

Removing the checks reduces verbosity of configure a little bit making
its output more accessible.

**Alternate Designs**
None

**Release Notes**
N/A


lfield:
The CI tests pass so I guess the build is fine.

PR #3070, opened 2019-03-25
Fix prototype warnings in the BOINC API headers

### LINKED ISSUE #861: Prototype warnings in the BOINC API headers
**Reported by gombasg on 24 Nov 39325552 20:26 UTC**
Hi,

When the BOINC API headers are included from C source code that gets compiled with "-Wstrict-prototypes", they generate a lot of "function declaration isnt a prototype" warnings. The attached patch fixes it by turning "foo()" to "foo(void)".

Gabor

Migrated-From: http://boinc.berkeley.edu/trac/ticket/884

AenBleidd:
Apply patch reported almost 10 years ago.
It doesn't make so much affect now but make API headers more consistent.
Also removed header of the function that was removed many years ago.
This is cosmetical change and make doesn't touch existing functionality.
Fixes #861

### DIFF
--- a/api/boinc_api.h
+++ b/api/boinc_api.h
@@ -76,7 +76,7 @@ typedef struct BOINC_STATUS {
 
 extern volatile BOINC_STATUS boinc_status;
 
-typedef void (*FUNC_PTR)();
+typedef void (*FUNC_PTR)(void);
 
 struct APP_INIT_DATA;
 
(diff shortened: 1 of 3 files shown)


davidpanderson:
I prefer foo() to foo(void).  What's the point of void?

AenBleidd:
@davidpanderson, 
In C++ there is no difference. In pure C foo() in definition means that function could have some arguments but at this point they are not specified, and foo(void) defines that function takes no arguments.
Because this is a header file of API functions and it could be used in some applications written on pure C in could be useful to avoid warnings.

AenBleidd:
@JuhaSointusalo, fixed

davidpanderson:
OK, I didn't realize that.  No objection.

AenBleidd:
@JuhaSointusalo, please merge if you're ok with fixed commit message. Thanks

PR #3075, opened 2019-03-29
remote job submission: fix race condition

davidpanderson:
Scenario:
- a submit_batch() creates a batch in INIT state
    and starts creating jobs
- while it's in progress, a query_batches() RPC enumerates batches
- while the query_batches() is in progress,
    the submit_batch() finishes and changes the state to IN_PROGRESS
- the query_batches() finishes and rewrites the batch in INIT state

Solution: query_batches() doesn't update batches in INIT state

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/html/inc/submit_util.inc
+++ b/html/inc/submit_util.inc
@@ -96,6 +96,11 @@ function delete_remote_submit_user($user) {
 // TODO: update est_completion_time
 //
 function get_batch_params($batch, $wus) {
+    if ($batch->state == BATCH_STATE_INIT) {
+        // a batch in INIT state has no jobs
+        //
+        return $batch;
+    }
     $fp_total = 0;
     $fp_done = 0;
     $completed = true;


lfield:
Is there any reason why the call is done the first place?  The [if statement ](https://github.com/BOINC/boinc/blob/fbc809b0ede32dc247f96331dd61b6bb27156f8a/html/user/submit_rpc_handler.php#L574) could be changed to:

>         if (BATCH_STATE_COMPLETE < $batch->state and  $batch->state < BATCH_STATE_COMPLETE) {

davidpanderson:
?  The function of the RPC is to find the status of batches.

lfield:
I don’t mean the RPC call but the get_batch_params function.

davidpanderson:
That function computes the batch's fraction done,
i.e. what fraction of its jobs have finished.

lfield:
I understand but why call that function at all if the batch is not in the state `BATCH_STATE_IN_PROGRESS`. Why not do the following instead?

> if ($batch->state == BATCH_STATE_IN_PROGRESS) {
>      $batch = get_batch_params($batch, $wus);
> }

lfield:
Sorry clicked on the wrong button.

davidpanderson:
You could do it either way.  This way is fine.

lfield:
ok, I was just considering readability of the code by having the if before the function call verses the abstraction of having it contained with the function.

PR #3077, opened 2019-03-30
update copyright year

JuhaSointusalo:
Update the copyright year in various files to 2019.


Uplinger:
Juha,

What are your thoughts on these files?  There are a bunch listed in this search:
https://github.com/BOINC/boinc/search?q=copyright+California&unscoped_q=copyright+California

There appear to be a lot of instances, I understand some of the ones you pointed towards are not commented out in source code like the majority.  

A few you might need to add if only doing forward facing ones:
doc/manpages/boinc.xml
doc/manpages/boinccmd.xml
doc/manpages/boincmgr.xml

Thanks,
-Keith Uplinger

JuhaSointusalo:
I actually used this command to search for the copyrights:

`grep -ri -e "copyright" -e "(c)" -e "©" -e "Ⓒ" -e "ⓒ" *`

GitHub's search interface seems too limited to me for any real work.

My thinking is that there's two kinds of copyright years, those that describe BOINC as a whole and those that describe individual source code files. I'm no expert on copyrights but I'd imagine you only get a copyright when you create something new or modify something that already was your creation. So, source code files would get a new copyright year when there are edits in the file. But when looking at BOINC as whole there presumably have been changes if we are making a new release then that copyright year can be updated.

So the files I changed should be ones that have BOINC as whole copyright year.

I hadn't really looked inside the man page sources. I somehow assumed they have copyright only as a source file.

If you think the man pages should be updated too I can do that but I think I'll check a DocBook tutorial first to see why the year is so many times in the files.

Uplinger:
I did a similar grep command, but I did mine searching for copyright and california while excluding lines that started with // or # as those would be commented out lines in the source code.

I think since pretty much everything you've changed is in the forward facing and something an end user would see, I would suggest changing those 3 man page xmls as well.  

Beyond that, I would say it's good to merge and should be part of the next client release.  Thanks for your help on this!

JuhaSointusalo:
Man pages added.

I didn't change the Solaris man pages. I think the two man page sets should be merged and I don't think anyone is using the CSW packages anyway so it doesn't matter much that the pages are not up to date. The Solaris pages talk about 5.10 so they are clearly not up to date.

Btw. I updated the list in https://boinc.berkeley.edu/trac/wiki/AdminTasks. Less grepping the next time 😀

Uplinger:
Thanks for your work on this.

PR #3079, opened 2019-03-31
remove gl.py

JuhaSointusalo:
The file isn't referenced anywhere and besides the copyright header it
doesn't have any content either.


davidpanderson:
This file is a header to include in new Python files,
analogous to "gl" for C++ and PHP.
It should be maintained and not removed.

lfield:
As a reason has been given, I will revert and close this PR.

PR #3080, opened 2019-03-31
web: Fix auto-attach process for VBox projects

davidpanderson:
Problem: the new signup/download process (auto-attach) wasn't telling
Mac and Linux users to install VirtualBox for projects that require it.
(These platforms don't have a combined BOINC/VBox installer;
the user has to install VBox separately).

Solution: add links to the VirtualBox download page in appropriate places.

Also: we currently have a <need_vbox/> config option
for projects that require VBox (i.e. all their apps are VM)
Add a <recommend_vbox/> option for projects with some but not all VM apps.

Also: rename download.php to download_software.php.
There was a name conflict with a file from the boinc-site repo.
This caused problems on the BOINC web site, which uses both code bases.
Projects with a link to download.php on their home page
should change this to download_software.php.

### DIFF
--- a/html/inc/bootstrap.inc
+++ b/html/inc/bootstrap.inc
@@ -89,7 +89,7 @@ function navbar_right($user) {
             );
         } else {
             echo sprintf('
-                <li><a href="%screate_account_form.php">Sign Up</a></li>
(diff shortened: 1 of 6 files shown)


lfield:
I am happy to merge if there are no objections.

PR #2208, opened 2017-11-03
add grcpool to account manager list

Holger-Will:
please consider adding grcpool.com to the account manager list. Currently this is done with a manual process using an APK Editor: https://youtu.be/dB-Hmg94E6s

PR #3088, opened 2019-04-04
Added grcpool.com to account manager list

Pythonix:
grcpool.com has more than 10'000 active users and should therefore also be an option in the account manager dropdown menu.

**Release Notes**
- Users can now select grcpool.com as their account manager


AenBleidd:
Duplicate of #2208 
See discussion there for more information

PR #3089, opened 2019-04-05
scheduler: check daily job limit on each job, not just when choosing AV

### LINKED ISSUE #3073: The daily quota is not fullly respected if close to limit.
@davidpanderson I have been using the result of PR #3024 and each day when n_jobs_today is reset to 0, the host will receive more than 1 job even though max_jobs_per_day = 1. This results in n_jobs_today > max_jobs_per_day.  From looking at the code, it seems that the check n_jobs_today < max_jobs_per_day is done quite early on but when it comes to actually selecting the tasks to send, this limit is not considered. Either the selection of tasks should include a check so that it stops when the limit is reached or the number of tasks to be selected is changed based on comparison with the limit.

davidpanderson:
Fixes #3073

### DIFF
--- a/sched/sched_keyword.cpp
+++ b/sched/sched_keyword.cpp
@@ -54,7 +54,7 @@ double keyword_score(int i) {
     USER_KEYWORDS& uk = g_request->user_keywords;
     if (uk.empty()) {
         if (config.debug_keyword) {
-            log_messages.printf(MSG_NORMAL, "user has no keywords; returning 0\n");
+            log_messages.printf(MSG_NORMAL, "[keyword] user has no keywords; returning 0\n");
         }
         return 0;
     }
@@ -66,7 +66,7 @@ double keyword_score(int i) {
         WU_RESULT& wr = ssp->wu_results[i];
         if (empty(wr.workunit.keywords)) {
             if (config.debug_keyword) {
-                log_messages.printf(MSG_NORMAL, "job has no keywords; returning 0\n");
+                log_messages.printf(MSG_NORMAL, "[keyword] job has no keywords; returning 0\n");
             }
             return 0;
         }
@@ -74,7 +74,7 @@ double keyword_score(int i) {
     }
     double s = keyword_score_aux(uk, jk);
     if (config.debug_keyword) {
-        log_messages.printf(MSG_NORMAL, "keyword score: %f\n", s);
+        log_messages.printf(MSG_NORMAL, "[keyword] keyword score: %f\n", s);
     }
     return s;
 }
--- a/sched/sched_score.cpp
+++ b/sched/sched_score.cpp
@@ -272,6 +272,17 @@ void send_work_score_type(int rt) {
             continue;
         }
 
+        // check daily limit for (host, app version)
+        //
+        if (daily_quota_exceeded(job.bavp)) {
+            if (config.debug_quota) {
+                log_messages.printf(MSG_NORMAL,
+                    "[quota] daily host/app version quota now exceeded\n"
+                );
+            }
+            break;
+        }
+
         if (!sema_locked) {
             lock_sema();
             sema_locked = true;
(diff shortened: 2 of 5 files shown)


lfield:
I have spent some time testing this and there are a few issues. The first two seem to be issues with the previous PR #3024. I don't know why I didn't find them when I was testing previously. 

1. There is a missing "where" in the [clause](https://github.com/BOINC/boinc/blob/master/sched/validator.cpp#L197) given to the lookup function. 
1. The function hav.update_validator(hav_orig) should be used instread of [hav.update()](https://github.com/BOINC/boinc/blob/master/sched/validator.cpp#L210)

In this PR the call to daily_quota_exceeded will  always pass as the the number of tasks provided by the for loop are not taken into account. Either the DB needs to be updated on each iteration or a variable is required to store this value and taken into consideration by the function.

davidpanderson:
Fixed now (I think) in this branch and the new dpa_punitive2

lfield:
Still doesn't work. In the DB, max_jobs_per_day=1 and  n_jobs_today=0 so I would expect 1 job but get four. Here are relevant lines from the scheduler log.

> 2019-04-10 10:08:09.2451 [PID=3297865]   Request: [USER#2] [HOST#3641] [IP 128.141.150.54] client 7.12.0
> 2019-04-10 10:08:09.2464 [PID=3297865]    [quota] effective ncpus 1 ngpus 0
> 2019-04-10 10:08:09.2464 [PID=3297865]    [quota] max jobs per RPC: 50
> 2019-04-10 10:08:09.2464 [PID=3297865]    [quota] Overall limits on jobs in progress:
> 2019-04-10 10:08:09.2464 [PID=3297865]    [quota] total: base 50 scaled 50 njobs 0
> 2019-04-10 10:08:09.2464 [PID=3297865]    [quota] CPU: base 50 scaled 50 njobs 0
> 2019-04-10 10:08:09.3737 [PID=3297865]    [quota] reached limit on CPU jobs in progress
> 2019-04-10 10:08:09.3737 [PID=3297865]    [quota] Overall limits on jobs in progress:
> 2019-04-10 10:08:09.3737 [PID=3297865]    [quota] total: base 50 scaled 50 njobs 4
> 2019-04-10 10:08:09.3738 [PID=3297865]    [quota] CPU: base 50 scaled 50 njobs 4
> 2019-04-10 10:08:09.3738 [PID=3297865]    [quota] reached limit on AMD/ATI GPU jobs in progress
> 2019-04-10 10:08:09.3738 [PID=3297865]    [quota] Overall limits on jobs in progress:
> 2019-04-10 10:08:09.3738 [PID=3297865]    [quota] total: base 50 scaled 50 njobs 4
> 2019-04-10 10:08:09.3738 [PID=3297865]    [quota] CPU: base 50 scaled 50 njobs 4
> 2019-04-10 10:08:09.3882 [PID=3297865]    Sending reply to [HOST#3641]: 4 results, delay req 61.00

Please could you do a pull request for [dpa_punitive2](https://github.com/BOINC/boinc/tree/dpa_punitive2), just [click here](https://github.com/BOINC/boinc/compare/dpa_punitive2?expand=1). Just deleted this [comment](https://github.com/BOINC/boinc/compare/dpa_punitive2?expand=1#diff-0f276d25dd9af9c88779ab74a7ae4d48R212) first and I can merge immediately.

davidpanderson:
I did a PR for dpa_punitive2, leaving the valid comment.

davidpanderson:
In my testing this seemed to work.  Remember that the daily limits are per (host, app version).  Maybe it's sending jobs for other app versions.  Set debug_version_select and debug_send.

lfield:
I rebuilt the scheduler from a fresh codebase and it now seems to work.  I just want to leave it running on the dev project overnight before merging.

lfield:
Looks good.

PR #3090, opened 2019-04-05
Fixes uninitialized values of min/max_wu_id and min/max_batch

SETIguy:
I found this problem today while debugging a different server issue, workunits were getting rejected because they were being compared to inititalized values.

Initializing the values solved the problem.

Even though I did a commit range cherry pick, only one commit comment is shown.  Not sure why that is.


davidpanderson:
Compile warnings showed that we needed to use longs instead of ints for IDs.
Eric, let me know if this looks OK, and I'll merge.

SETIguy:
Looks good.  Once the final check finishes it's good to commit.

PR #3093, opened 2019-04-10
Revert "remove gl.py"

lfield:
Reverts BOINC/boinc#3079


ChristianBeer:
I don't see a reason for this. This is basically an empty file that is not referenced in any other file or documentation. It is also placed in the base directory. If it is meant to be included in future python files as part of an API wouldn't this file be better placed inside the API directory? How would a future contributor know about the reason why this file is there and that it should be used?

PR #3094, opened 2019-04-11
Removed "Exit from BOINC Manager", "Shut down connected client" menu entries on Linux

Germano0:
New version of aborted pull request https://github.com/BOINC/boinc/pull/3028

**Description of the Change**
As by default on Linux BOINC runs as a service, the GUI must not stop it.

**Alternate Designs**

**Release Notes**
 Removed "Exit from BOINC Manager", "Shut down connected client" menu entries on Linux

### DIFF
--- a/clientgui/AdvancedFrame.cpp
+++ b/clientgui/AdvancedFrame.cpp
@@ -348,11 +348,15 @@ bool CAdvancedFrame::CreateMenu() {
         _("Select computer...\tCtrl+Shift+I"),
         _("Connect to a BOINC client on another computer")
     );
+
+// As by default on Linux BOINC runs as a service, the GUI must not stop it.
+#ifndef __WXGTK__
     menuFile->Append(
         ID_SHUTDOWNCORECLIENT, 
         _("Shut down connected client..."),
         _("Shut down the currently connected BOINC client")
     );
+#endif
     menuFile->AppendSeparator();
 
     strMenuDescription.Printf(
@@ -367,6 +371,8 @@ bool CAdvancedFrame::CreateMenu() {
         strMenuDescription
     );
 
+// As by default on Linux BOINC runs as a service, the GUI must not stop it
+#ifndef __WXGTK__
     strMenuDescription.Printf(
         _("Exit %s"), 
         pSkinAdvanced->GetApplicationName().c_str()
@@ -387,6 +393,7 @@ bool CAdvancedFrame::CreateMenu() {
         strMenuName,
         strMenuDescription
     );
+#endif
 
 #ifdef __WXMAC__
     // wxWidgets actually puts this in the BOINCManager menu
--- a/clientgui/sg_BoincSimpleFrame.cpp
+++ b/clientgui/sg_BoincSimpleFrame.cpp
@@ -131,6 +131,8 @@ CSimpleFrame::CSimpleFrame(wxString title, wxIconBundle* icons, wxPoint position
         strMenuDescription
     );
 
+// As by default on Linux BOINC runs as a service, the GUI must not stop it.
+#ifndef __WXGTK_
     strMenuDescription.Printf(
         _("Exit %s"),
         pSkinAdvanced->GetApplicationName().c_str()
@@ -146,6 +148,7 @@ CSimpleFrame::CSimpleFrame(wxString title, wxIconBundle* icons, wxPoint position
         strMenuName,
         strMenuDescription
     );
+#endif
 
 #ifdef __WXMAC__
     // wxWidgets actually puts this in the BOINCManager menu


RichardHaselgrove:
I don't think we should forget the https://github.com/BOINC/boinc/pull/3028#issuecomment-470269114 by @JuhaSointusalo in the previous conversation.

Agreed, default (distro built and maintained) Linux versions run as services: but now, if users choose to build and deploy their own versions of the client and manager from source, to run in user mode, they will find that tools which are relevant to their own choices are missing.

lfield:
It you can't start the client from the manager you shouldn't be able to stop it. I think it was a mistake to add this feature in the first place.

RichardHaselgrove:
Unfortunately, I'm not qualified to speak on behalf of Linux users - I'm strictly Windows only. So I've started a consultation thread on the BOINC message board ([Opinions requested from home Linux users](https://boinc.berkeley.edu/forum_thread.php?id=12905)), and I'll pass on any feedback I receive.

JuhaSointusalo:
I thought I had made my opinion clear earlier. I don't approve taking away features from all users just because the features don't work right when BOINC is installed from Fedora repository.

Another issue I have with these changes is that they are one hack on top of another. Neither of you have studied how Manager actually controls the client.

> It you can't start the client from the manager you shouldn't be able to stop it

Users start the Manager which then starts the client.

lfield:
@JuhaSointusalo 
>  Neither of you have studied how Manager actually controls the client.
In addition to studing, we have also documented the [protocol](https://boinc.berkeley.edu/trac/wiki/GuiRpcProtocol).
> I thought I had made my opinion clear earlier. I don't approve taking away features from all users just because the features don't work right when BOINC is installed from Fedora repository.
This is not just about Fedora but all the main Linux distributions that we are supporting. The more I think about this, stopping the client from the manager is not a feature but a bug.
> It you can't start the client from the manager you shouldn't be able to stop it
> Users start the Manager which then starts the client.
Are you sure?

JuhaSointusalo:
> > Neither of you have studied how Manager actually controls the client.
> 
> In addition to studing

The process.

> > Users start the Manager which then starts the client.
> 
> Are you sure?

Yes.

RichardHaselgrove:
I had to study the Windows implementation of this is detail when making a standalone installer for third party applications. That had to detect what mode the client was running in, stop it according to that mode, do the business, and restart it in the same mode - all while leaving the manager open on screen. So, yes, I did study the protocols back then, and I could achieve all that in both User and Service modes - although the tools and procedures were different.

I would say that the bug in this case is that the Manager can't start the Client, in service mode, in Linux. Once that gap is filled, your objection to stopping it falls.

Why there isn't a Linux equivalent of boincsvcctrl.exe, I don't know. Maybe https://github.com/BOINC/boinc/tree/master/clientctrl would help.

RichardHaselgrove:
The relevant lines in Script generated by the HM NIS Edit Script Wizard are:

Stop service:
`nsSCM::Stop /NOUNLOAD "Boinc"`
(that's the NIS Service Control Module)

Stop user:
`nsExec::Exec /TIMEOUT=2000 '"$BOINC_INSTALLDIR\boinccmd.exe" --quit'`

Start user:
`nsExec::Exec /TIMEOUT=100 '"$BOINC_INSTALLDIRboinc.exe" --detach_phase_two'`
(which is basically what the Manager does, although it uses ""C:\BOINC\boinc.exe" --redirectio --launched_by_manager ")

Start service:
`nsExec::Exec /TIMEOUT=100 '"$BOINC_INSTALLDIRboincsvcctrl.exe" --start'`

So it does vary according to mode, but we should have the full set available.

CharlieFenton:
>I would say that the bug in this case is that the Manager can't start the Client, in service mode, in Linux. Once that gap is filled, your objection to stopping it falls.

I’m not an expert on Linux, but on the Mac (Unix) the Manager does start the client as a service if the user has set it up to run as a service. It can also stop the client even if running as a service.

My implementation on the Mac is: exiting the Manager stops the client if and only if it started the client. Note that this is different than the Windows implementation.

Germano0:
> Users start the Manager which then starts the client.

I am already working on a patch to prevent the Manager from starting the Client.
If the user opens the the Manager and the Client for any reason is not already running, the Manager will start the Client as normal user, rather than the confined user used when Client runs as systemd service.
Unfortunately the patch I elaborated does not work and I am still working with GDB to find out where the problem is.
The problem is that I am a bit confused by some ifdef/endif blocks of code, that is also the reason why I started the mailing list discussion ["Question about preprocessor macros" (clientgui/BOINCClientManager.cpp) ](https://groups.google.com/a/ssl.berkeley.edu/forum/#!topic/boinc_dev/_Xn7sc9E-mo)

lfield:
I think that the difference in opinions stems from the overloading of the code to support different models. One is the system service model where the lifecycle of the client is managed by the OS and the GUI is a window into that service. The other is an application model where the client is coupled to the GUI and therefore the clients lifecycle is managed by the GUI. There are most probably other modules. The general issue we face is that the GUI flips between different models depending on the status rather than design and hence leads to inconsistencies. My view for Linux is that the focus should be on supporting the system service model.

RichardHaselgrove:
I agree with the analysis and your distinction between the two models. But I don't necessarily agree with the conclusion.

I think this sort of discussion belongs at the design level, rather than the coding level. Perhaps we could schedule a discussion for the next contributor call?

lfield:
From a practical perspective, we are only supporting the service model. The recent versions of the Linux client have been pushed to the mainstream distributions and are intended to run as a service. The tar distribution is 6 years old.

RichardHaselgrove:
But that neglects the open-source, build-it-yourself, concepts that underpin the whole project. Anybody can download current sources from GitHub. This PR effectively removes user controls from any build where `#ifdef __WXGTK__` applies (I know the logic is reversed - adds controls if ndef): open source builders will have to be aware of that if they are planning their work to be run in User mode.

lfield:
This is an orthogonal argument. There is nothing to stop someone from building it themselves and running it as a service. What I think you mean is people running the BOINC client in a terminal. This is still supported. However as the client lifecycle is manual, the GUI should not manage client lifecycle operations. The use case you actually are referring to is where the GUI is managing the lifecycle of the client.

RichardHaselgrove:
Yes, I agree. But in my (Windows) experience, users use the Manager to manage the client, including all aspects of the lifecycle. For example, certain choices made in cc_config.xml only take effect at the next client restart ('Read config files' is insufficient). They take effect at the next computer reboot, of course, but some users take pride in running Linux continuously for hundreds of days between reboots. If the only component requiring the restart is the BOINC client, being able to manage that aspect of the package through the Manager would seem to be appropriate., even if service management tools are also available.

lfield:
Requiring a machine reboot to reconfigure suggests this is a system service however, this discussion is about Linux so let's focus on that. From what I understand the point you make above is that the GUI should be able to instigate a client restart. This is not necessarily the same as stop/start. In addition the assumption made by the GUI is that it should act upon processes where it should be calling:

`sudo systemctl restart boinc-client`

Stopping the process (which I don't think allowed anyway due to different users) would result in systemctl seeing the process as crashed and starting would create process under the GUI user not the service user account with the state files most probably being located in a different place. 

My point is that using the GUI following the service model works. Using the GUI following the application model may work but is fragile and but together it is broken.

RichardHaselgrove:
While we've been talking, I've activated a new Windows 10 machine I'm planning to use in the summer, and tested our (mine and Juha's) memories of service mode behaviour.

Machine is a desktop-equivalent ultraportable (full Intel i5), running Windows 10 Pro build 1809. BOINC is 7.14.2

Behaviour is still as we remember. In service mode, the BOINC client starts as the machine boots - before user logon (I'm going to change that to 'Automatic - delayed start', because without it BOINC is running before WiFi connects, and the first few network connections fail). If the client isn't running when the user logs on (say the service mode is set to Manual launch), then opening the Manager starts the service. This uses boincsvcctrl.exe, which in turn invokes the Windows 'User Access Control' dialog - the security equivalent of your 'sudo'. If the client is already running, the Manager connects to the running instance without a security check.

Once the session has been established, the Manager can 'snooze' the service from the system tray icon. The Manager can also be minimised, or it can be exited completely - with a choice of stopping the service, or not, at the same time.

I think that's the full set of options we've discussed, and they all work. I think Rom Walton left us with a full set of working tools, partly in the Client/Manager and partly in the Windows installer. It appears that user-mode development for Linux was paused at v7.2.42 (28 Feb 2014 - the last one recommended on the 'download_all' page), and has been allowed to wither on the vine since then.

The discussion we're having - notwithstanding any personal views we may have - is whether the BOINC user community should have access to a complete user-mode environment for Linux. My consultation yesterday had a limited response (just one reply so far), so I've re-posted it on the rather busier SETI board at https://setiathome.berkeley.edu/forum_thread.php?id=84128. Looks like we're getting more input there.

davidpanderson:
When you exit the Manager, it should ask whether you want to exit the client too.
Same on all platforms.

Ferroin:
@davidpanderson The problem with that on Linux is that it's very likely to not do what you actually expect it to, because most Linux users are not running BOINC manually.  The trivial example is that you're running BOINC under a service manager like systemd or monit, in which case _it will be automatically restarted_ when the BOINC Manager shuts down the client using RPC calls or by sending SIGTERM.  Starting the client with the Manager doesn't work sanely on Linux in any case other than manual handling of the client either, you usually end up with it starting up as if there was no previous BOINC data on the system (it doesn't wipe anything, it just doesn't use the standard data location, so it creates a 'new' instance in your home directory).

On Windows it's never really an issue, because all the permissions 'just work out' 99% of the time, and the Manager is able to sanely and securely manage the client running as a system service.  IOW, even in the insanely unlikely event that the Manager doesn't do the right thing when shutting down the client on Windows, it doesn't break anything.

KeithMyers:
Since I have lately been compiling my own client and Managers from github branches, I would be vehemently opposed to a unilateral decision to remove a choice I have come to depend on in the Manager, the dialog of whether to stop the client and exit all tasks or exit the Manager and leave all tasks running in the client.

tckosvic:
I find now that closing boinc manager shuts down the client. I find this need to keep the manager running very annoying. I currently use openSUSE 15.2.
Before I upgraded to 15.2, while running 15.,1 the boinc manager could be started and stopped and the client would keep running. I dont know what version of boinc manager was running under 15.1 but i would like to return to it. I accrued 1.5 million statistics points running 15.1 and starting boinc manager countless times.

The boinc manager with the annoying behavior is 7.16.6 version. In the file tab it says "shut down connected client" with no button for yes or no.

I would like to revert back to allow boinc client to keep running when manager it stopped. I think services should only be turned off by OS system root purposefully and a manager turning off a service client is wrong to me.

I saw some discussion about this here and thought it a good place to post.

tom kosvic

aaronpuchert:
> Stopping the process (which I don't think allowed anyway due to different users)

Agreed, this would go against Linux' service model.

> would result in systemctl seeing the process as crashed

Not quite, systemd can distinguish a crashed process from a process that has exited regularly, see the docs for [Restart](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=) and about [process exit codes](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Process%20exit%20codes).

> This uses boincsvcctrl.exe, which in turn invokes the Windows 'User Access Control' dialog - the security equivalent of your 'sudo'.

Not quite, the equivalent would be `xdg-su`, an unmaintained part of [xdg-utils](https://www.freedesktop.org/wiki/Software/xdg-utils/) ([source](https://cgit.freedesktop.org/xdg/xdg-utils/tree/scripts/xdg-su.in)), `sudo` is a command line program.

In any event, starting/stopping a service requires superuser privileges. That's intentional.

PR #3101, opened 2019-04-16
server: add script for purging retired batches

davidpanderson:


lfield:
I was expecting this to be fixed in the `db_purge `script. How come it is separated?

davidpanderson:
Easier to do it in PHP

lfield:
Even if it is a bit harder, I think this should be done in the `db_purge` script. I will provide an alternative PR.

TheAspens:
This was implemented by pull request #3148.  I recommend that we close this pull request since it has been implemented there.

PR #3107, opened 2019-04-20
Tests: add unit testing framework using googletest

ChristianBeer:
**Description of the Change**
* enable coverage reports from gcc and disable optimizations via option to configure
* install googletest library into buildcache
* script to compile and run unit tests using cmake
* first set of unit tests originally contributed by Keith Uplinger

There are several hardcoded paths and assumptions made in order to get this working on Travis CI. New tooling is using cmake for cross platform builds and as such is not easy to use with an autotools based system.
It's not ideal but better than nothing.

**Alternate Designs**
There are other unit testing frameworks out there that could have been used. Googletest satisfies the current needs and should be for the future since it is xUnit compatible.

**Release Notes**
Added unit-testing capability to BOINC to improve source code quality and find bugs earlier.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,16 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+comment:
+  layout: "diff, files"
+  behavior: default
+  require_changes: false  # if true: only post the comment if coverage changes
+  require_base: no        # [yes :: must have a base report to post]
+  require_head: yes       # [yes :: must have a head report to post]
+
+# don't gather statistics for tets and external libraries
+ignore:
+  - tests/.*
+  - zip/.*
+  - 3rdParty/.*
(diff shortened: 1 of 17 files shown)


ChristianBeer:
@lfield @Uplinger This is it. Everything that is needed for proper unit testing. I'm not sure why the codecov bot report appears twice but that might change once there is a baseline to diff against. If not then I'm going to send an email to them.

Uplinger:
Closing to restart code checks.

Uplinger:
@ChristianBeer I thought maybe there was an issue with travis failing, and attempted to retrigger the build.  However it failed at the same spot.  I'm not sure why this would be the case, because it does not appear that your code changes actually touch this.  The file it apparently failed on hasn't been changed in master in 8 months.

Uplinger:
I have discusses this with @ChristianBeer and others on the Contributors Call.  Since the code that is being updated doesn't involve the failure, we are merging it.  Thanks

PR #3110, opened 2019-04-26
[Android] Remove XML CDATA tag from Event Log messages

### LINKED ISSUE #2751: 7.14.1 for Android: Event Log is filled with <![CDATA[>entries.
Event Log is filled with all these <![CDATA[>entries.
(The forum here drops everything behind a larger than sign, so log removed)

AenBleidd:
Fixes #2751

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/Message.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/Message.java
@@ -64,6 +64,8 @@ public class Message implements Serializable, Parcelable {
         seqno = in.readInt();
         timestamp = in.readLong();
         body = in.readString();
+        body = body.replace("<![CDATA[", "");
+        body = body.replace("]]>", "");
     }
 
     public static final Parcelable.Creator<Message> CREATOR = new Parcelable.Creator<Message>() {


AenBleidd:
CI fail is not related to this particular PR

AenBleidd:
I'm afraid I won't find this thread but it was decided some time ago that all event messages should be covered with CDATA tag because RPC protocol sends data in XML format and these event messages could contain data that could be parsed wrong by XML parser if it will be in plain text. This change was done on client side and I believe everyone just forget about Android client that is why we got this behavior now

adamradocz:
@brevilo would you merge this PR?

ChristianBeer:
The commit that introduced the CDATA elements is 1fe02c9d4fe815a69096e5cfc8e8985aa2308890 and there was a fix to the Manager some time later. This makes sense and yes the CI failure is not related to this PR it is a general failure of the integration tests.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/Message.java
```diff
@@ -64,6 +64,8 @@ private Message(Parcel in) {
         seqno = in.readInt();
         timestamp = in.readLong();
         body = in.readString();
+        body = body.replace("<![CDATA[\n", "");
```
brevilo: Strictly speaking this removes more than the [client change](https://github.com/BOINC/boinc/commit/1fe02c9d4fe815a69096e5cfc8e8985aa2308890#diff-254fd9eea50fb57d93c683e033cd3ad8R220) added, namely the line breaks. Is this intentional?

AenBleidd: Yes, otherwise before and after every message in event log will be empty lines

brevilo: Ok, but that's how the client is written. I don't think we should fix things locally as this would be inconsistent. If that those extra line breaks aren't wanted then this should be fixed (glabally) in the client.

AenBleidd: OK, no problem. Removed line breaks removes.

PR #3118, opened 2019-05-03
[Android] Change the Desktop Environment from Mate to Lubuntu in Vagrant

adamradocz:
**Description of the Change**
Lubuntu Desktop Environment is lighter than the Mate DE, which is very beneficial in a virtual environments situation.
Lubuntu uses 194 MB RAM and 12,47 GB disk space, while Mate uses 435 MB RAM and 13,95 GB disk space. Plus with the new DE the installation process is also faster.

**Release Notes**
N/A


adamradocz:
I did all of the PRs in the last couple of days with this new Desktop Environment, it's fine. Not so different from Mate.

PR #3120, opened 2019-05-03
[Android] Update the Android Gradle plugin to version 3.4.0 and Gradl…

adamradocz:
…e to version 5.1.1.

**Description of the Change**
To take advantage of the latest features, improvements, and security fixes, it is strongly recommended to update the Android Gradle plugin to version 3.4.0 and Gradle to version 5.1.1. [Release notes.](https://developer.android.com/studio/releases/gradle-plugin)

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/build.gradle
+++ b/android/BOINC/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.3.0'
+        classpath 'com.android.tools.build:gradle:3.4.0'
     }
 }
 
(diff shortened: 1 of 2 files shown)


AenBleidd:
This update brings a lot of deprecated functions that would prevent us from updating Gradle to version 6.0.
I believe all of them should be fixed but at this moment there is no real reason to be updated to every upcoming Gradle release.
This update was not so necessary either but ok.

adamradocz:
R8 enabled by default, I think it's worth to upgrade. Plus I like to upgrade things to the latest version anyway, because of its improvements.

AenBleidd:
Upgrade is great when you have proper testing. I believe you heard about Regression Testing. It is really necessary when you upgrade smth. Currently we have no Android testing at all and there are a lot of other issues that should be fixed to make new release possible, and, unfortunately, neither SDK upgrade no toolset upgrade will fix them.

PR #3121, opened 2019-05-03
[Android] Add guide to Vagrant Readme on how to push to GitHub without password using ssh-key

adamradocz:
**Description of the Change**
Added a guide on how to set up the git and GitHub to push without a password prompt.

**Alternate Designs**
These type of guides should be placed in the Wiki, but we don't have one.

**Release Notes**
N/A


AenBleidd:
Looks fine. Thanks

PR #3122, opened 2019-05-03
[Android] Use apply() instead of commit on shared preferences. Wherea…

adamradocz:
…s commit blocks and writes its data to persistent storage immediately, apply will handle it in the background.

**Release Notes**
Increase responsiveness in the Preferences menu.


adamradocz:
I used the IDE to fix this issue. Honestly, I have no idea why the others are not marked and changed during the analysis. If the others don't throw warning messages by the analyzer, I'd leave them as is.

AenBleidd:
But if you go in IDE through the file AppPreferences.java you will find the same functions with the same behavior that are left unchanged. It looks strange for me. If tool fixes one issue and don't fix another similar one there is two possible options:
- tool works incorrectly
- tool knows smth we don't.
In both cases we should ensue that changes we are committing are complete and correct

adamradocz:
You were right. The IDE shows as a warning, I missed it.

PR #3127, opened 2019-05-04
[Android] Fix code style issues

adamradocz:
**Description of the Change**
- Remove unnecessary interface modifiers.
- Use the standars CharSet object. The constant charset String literal can be replaced with a predefined Charset object. This may work a little bit faster, because charset lookup becomes unnecessary.

**Release Notes**
N/A


AenBleidd:
Looks fine. Thanks!

PR #3128, opened 2019-05-04
[Android] Suppress WakeLock warning message. To run BOINC in the back…

adamradocz:
…ground indefinitely, wakeLock has to be used without a timeout, so it's fine as it is.

**Release Notes**
N/A


AenBleidd:
Looks fine. Thanks!

PR #3131, opened 2019-05-06
web: if view empty thread, mark it as read

davidpanderson:
Fixes problem reported by Jord where if you view a thread with no posts
(e.g. the only post got moderated)
the thread continues to be shown as unread.


TheAspens:
Looks good.  Thanks for the fix.

PR #3159, opened 2019-05-24
[Android] Remove unreachable code

AenBleidd:
Fixing warning found by static analyzer:
`V6003 The use of 'if (A) {...} else if (A) {...}' pattern was detected. There is a probability of logical error presence.`

This piece of code duplicates the one already presented above where it logically more correct.

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/ProjectConfigReplyParser.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/ProjectConfigReplyParser.java
@@ -154,9 +154,6 @@ public class ProjectConfigReplyParser extends BaseParser {
                     else if(localName.equalsIgnoreCase("error_num")) { // reply is not present yet
                         mProjectConfig.error_num = Integer.parseInt(mCurrentElement.toString());
                     }
-                    else if(localName.equalsIgnoreCase("client_account_creation_disabled")) {
-                        mProjectConfig.clientAccountCreationDisabled = true;
-                    }
                     else if(localName.equalsIgnoreCase("account_manager")) {
                         mProjectConfig.accountManager = true;
                     }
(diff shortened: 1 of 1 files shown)


android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/ProjectConfigReplyParser.java
```diff
@@ -151,10 +151,6 @@ else if(localName.equalsIgnoreCase("platform") &&
                         platformName = "";
                         platformPlanClass = "";
                     }
-                    else if(localName.equalsIgnoreCase("platform") && withinPlatforms) { // finish platforms
-                        withinPlatforms = false;
-                        mProjectConfig.platforms = mPlatforms;
-                    }
```
brevilo: Don't you have to retain the two statements contained in the then-clause you removed? Shouldn't they be added to original if-clause, say, following line 152?

brevilo: IOW, you could simply have removed the following instead, no?
```
                    }
                    else if(localName.equalsIgnoreCase("platform") && withinPlatforms) { // finish platforms
```

AenBleidd: No, these lines are located here:
https://github.com/BOINC/boinc/blob/8624ed384ef0e23900b647f33ff5a1a7d7626ae3/android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/ProjectConfigReplyParser.java#L92-L95

AenBleidd: Here we're parsing several <platform> tags inside <platforms> tag. If we merge these lines we will have an logic error parsing only one <platform> instead of all,

brevilo: Ok, just wanted to ensure you didn't miss them by accident. Thanks.

AenBleidd:
> My only disappointment is we still don't use a standard JSON or XML library. We shouldn't have had an issue like this.

I believe some day someone find enough time to rewrite existing parser using 3rd party XML parsing library.

@adamradocz, if you're ok please merge this. Thanks

PR #3160, opened 2019-05-24
client: clear the vm_extensions_disabled flag on startup.

### LINKED ISSUE #1460: Client stuck thinking VM extensions are disabled
In some cases, the client (tested on Win/7.6.22 and Linux/7.2.47) can get stuck thinking VM extensions are disabled. There may be more ways, but one way to reproduce is:
- Ensure VM extensions are off in your BIOS
- Attempt to request some jobs from a project which has VM jobs that need extensions (you'll rightly get none)
- Go turn VM extensions on in your BIOS
- Come back to BOINC and try to request jobs again, you'll still get none

This is actually pretty common. Since VM extensions are often off by default, people come in and try to run your jobs, they don't work, then they go and read the website and find out they need VM extensions, but they turn them on and it still doesn't work. 

The [workaround](http://www.cosmologyathome.org/faq.php#i-enabled-vt-xamd-v-but-jobs-say-scheduler-wait-please-upgrade-boinc) btw is the user has to edit the `client_config.xml` file and delete `<p_vm_extensions_disabled>` by hand.

davidpanderson:
Fixes #1460

### DIFF
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -624,6 +624,13 @@ int CLIENT_STATE::init() {
     // domain_name for Android
     //
     host_info.get_host_info(true);
+
+    // clear the VM extensions disabled flag.
+    // It's possible that the user enabled them since the last VM failure,
+    // or that the last failure was specious.
+    //
+    host_info.p_vm_extensions_disabled = false;
+
     set_ncpus();
     show_host_info();
 


marius311:
Just checking, so the behavior after this is  that every time you restart BOINC, you'll get at least one VM job even if you don't have extensions, it'll fail, and the you'll get no more VM jobs until you reboot? This should definitely be an improvement over what is currently happening.

davidpanderson:
That's correct.

PR #3168, opened 2019-05-29
Quality: fix small issues

ChristianBeer:
**Description of the Change**
Fixes uninitialized variables in several components

**Release Notes**
N/A

### DIFF
--- a/samples/vboxwrapper/vbox_vboxmanage.cpp
+++ b/samples/vboxwrapper/vbox_vboxmanage.cpp
@@ -801,9 +801,11 @@ namespace vboxmanage {
                 return deregister_vm(false);
             } else {
                 command  = "closemedium dvd \"" + virtual_machine_slot_directory + "/" + iso_image_filename + "\" ";
+                // coverity[CHECKED_RETURN]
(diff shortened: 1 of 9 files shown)


samples/vboxwrapper/vbox_vboxmanage.cpp
```diff
@@ -801,9 +801,11 @@ namespace vboxmanage {
                 return deregister_vm(false);
             } else {
                 command  = "closemedium dvd \"" + virtual_machine_slot_directory + "/" + iso_image_filename + "\" ";
+                // coverity[CHECKED_RETURN]
```
AenBleidd: @ChristianBeer,  Personally I don't like stuff like this. Are these checks so critical and important that we need to say to coverity that we are ok with this? As far as I know, we don't have a mission to achieve 0 messages from any static analyzer, so why do we need this?

ChristianBeer: You are right we don't have a mission but I'm trying to find the low hanging fruits in the reports so that more serious defects are more easily identified. I'm usually trying to fix the root cause of a report but sometimes it's easier to let the analyzer ignore some specific checks and also let future developers know that the choice to do so was intentional.

PR #3174, opened 2019-06-04
Android version plan class fix

nilsjh:
Fixes #

**Description of the Change**
Detect Android major version as described in issue:  #3172  

Kudos to @JamesMolson 

**Release Notes**
Correct detection of of Android client version from the scheduler in cases where only the Android major version is read (e.g. 9.)

PR #3175, opened 2019-06-04
scheduler: parse Android version strings of the form "(Android 9)"

### LINKED ISSUE #3172: Android version parsing
https://github.com/BOINC/boinc/blob/master/sched/plan_class_spec.cpp#L63-L79

The above code looks for:
(Android x.y.z
or
(Android x.y

but not
(Android x

My phone reports itself as (Android 9), so the function just returns 0.
Something like
```
n = sscanf(p, "%d", &maj);
if (n == 1) {
return maj*10000;
}
```
should be added before the return 0;

davidpanderson:
(existing code only worked with a.b.c and a.b)

Fixes #3172 

Without this change, Android apps that require a min version won't get sent to devices that report their version as a single number.

### DIFF
--- a/sched/plan_class_spec.cpp
+++ b/sched/plan_class_spec.cpp
@@ -60,7 +60,7 @@ static double os_version_num(HOST h) {
     return 0;
 }
 
-// parse "Android 4.3.1" or "Android 4.3"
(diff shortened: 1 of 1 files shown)


adamradocz:
Duplicate #3174

lfield:
This one includes comment change.

PR #3191, opened 2019-06-21
Client: support GUI RPCs via HTTP POST requests,

### LINKED ISSUE #3190: Enable GUI RPC via HTTP
Goal: expose GUI RPCs via HTTP, thereby making it easier to implement GUIs in Javascript (for example, in Electron).

The client already has code for parsing HTTP request headers and sending reply headers.  This works for unauthenticated operations like get_state.  It doesn't work for authenticated operations because the authentication scheme (auth1/auth2) assumes that a sequence of RPCs arrive on a single TCP connection, and HTTP uses one connection per operation (at least, the basic JS APIs like XMLHttpRequest do).  So we need an authentication scheme that spans connections.

My proposal:
Add a new get_auth_id() RPC.  This returns an integer "authentication ID", to be passed in subsequent requests.

The BOINC client stores a list of these; each is associated with a sequence number of the last request.

In subsequent GUI RPCs, the GUI includes an HTTP header item consisting of
(authentication ID, SHA1(password + seqno) and increments the seqno.

To authenticate a request, the BOINC client
1) checks that the seqno is > previous seqno
2) computes SHA1(password + seqno) and checks it against the request

This protocol (I think) prevents spoofing even if bad guys read network traffic.

davidpanderson:
to facilitate Javascript-based GUIs.

Authentication (for operations that require it) is done using
special HTTP header fields.
At some point I'll check in a JS module for this.

Fixes #3190

### DIFF
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -1314,13 +1314,16 @@ bool valid_auth(int id, long seqno, char* hash, char* request) {
         AUTH_INFO& ai = auth_infos[i];
         if (ai.id != id) continue;
         if (seqno <= ai.seqno) return false;
-        ai.seqno = seqno;
         int n = request?(int)strlen(request):0;
-        snprintf(buf, sizeof(buf), "%d%s%s", seqno, gstate.gui_rpcs.password, ai.salt);
+        snprintf(buf, sizeof(buf), "%ld%s%s", seqno, gstate.gui_rpcs.password, ai.salt);
         md5_block((const unsigned char*)buf, (int)strlen(buf), my_hash,
             (const unsigned char*)request, n
         );
-        if (strcmp(hash, my_hash)) return false;
+        if (strcmp(hash, my_hash)) {
+            msg_printf(0, MSG_INFO, "got invalid GUI RPC request");
+            return false;
+        }
+        ai.seqno = seqno;   // bump seqno only if valid request
         return true;
     }
     return false;
@@ -1356,8 +1359,8 @@ static bool authenticated_request(char* buf) {
     if (!p) return false;
     n = sscanf(p+strlen("Auth-Hash: "), "%64s", auth_hash);
     if (n != 1) return false;
-    char* request = strstr(buf, "\n\n");
-    if (request) request += 2;
+    char* request = strstr(buf, "\r\n\r\n");
+    if (request) request += 4;
     return valid_auth(auth_id, auth_seqno, auth_hash, request);
 }
 
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -1276,6 +1276,8 @@ static void handle_get_daily_xfer_history(GUI_RPC_CONN& grc) {
     daily_xfer_history.write_xml(grc.mfout);
 }
 
+// see if we got a complete HTTP request
+//
 static bool complete_post_request(char* buf) {
     if (strncmp(buf, "POST", 4)) return false;
     char* p = strstr(buf, "Content-Length: ");
@@ -1287,6 +1289,69 @@ static bool complete_post_request(char* buf) {
     p += 4;
     if ((int)strlen(p) < n) return false;
     return true;
+}
+
+// We use a different authentication scheme for HTTP because
+// each request has its own connection.
+// Send clients an "authentication ID".
+// Each request has (in HTTP header vars) the ID,
+// a sequence number, and a hash of the seq# and the password.
+//
+struct AUTH_INFO {
+    int id;
+    long seqno;
+};
+
+vector<AUTH_INFO> auth_infos;
+
+// check HTTP authentication info
+//
+bool valid_auth(int id, long seqno, char* hash) {
+    char buf[256], my_hash[256];
+    //printf("valid_auth: id %d seqno %ld hash %s\n", id, seqno, hash);
+    for (unsigned int i=0; i<auth_infos.size(); i++) {
+        AUTH_INFO& ai = auth_infos[i];
+        if (ai.id != id) continue;
+        if (seqno <= ai.seqno) return false;
+        ai.seqno = seqno;
+        snprintf(buf, sizeof(buf), "%d%s", seqno, gstate.gui_rpcs.password);
+        md5_block((const unsigned char*)buf, (int)strlen(buf), my_hash);
+        if (strcmp(hash, my_hash)) return false;
+        return true;
+    }
+    return false;
+}
+
+// create a new authentication ID
+//
+void handle_get_auth_id(MIOFILE& fout) {
+    static int id=0;
+    AUTH_INFO ai;
+    ai.id = id++;
+    ai.seqno = 0;
+    auth_infos.push_back(ai);
+    fout.printf("<auth_id>%d</auth_id>\n", ai.id);
+}
+
+// see if the HTTP request has valid authentication info
+//
+static bool authenticated_request(char* buf) {
+    int auth_id;
+    long auth_seqno;
+    char auth_hash[256];
+    char* p = strstr(buf, "Auth-ID: ");
+    if (!p) return false;
+    int n = sscanf(p+strlen("Auth-ID: "), "%d", &auth_id);
+    if (n != 1) return false;
+    p = strstr(buf, "Auth-Seqno: ");
+    if (!p) return false;
+    n = sscanf(p+strlen("Auth-Seqno: "), "%ld", &auth_seqno);
+    if (n != 1) return false;
+    p = strstr(buf, "Auth-Hash: ");
+    if (!p) return false;
+    n = sscanf(p+strlen("Auth-Hash: "), "%64s", auth_hash);
+    if (n != 1) return false;
+    return valid_auth(auth_id, auth_seqno, auth_hash);
 }
 
 static void handle_set_language(GUI_RPC_CONN& grc) {
@@ -1563,6 +1628,10 @@ int GUI_RPC_CONN::handle_rpc() {
     bool http_request;
     if (complete_post_request(request_msg)) {
         http_request = true;
+        if (authenticated_request(request_msg)) {
+            got_auth1 = got_auth2 = true;
+            auth_needed = false;
+        }
     } else {
         p = strchr(request_msg, 3);
         if (p) {
@@ -1604,6 +1673,8 @@ int GUI_RPC_CONN::handle_rpc() {
             retval = handle_auth2(request_msg, mfout);
             got_auth2 = true;
         }
+    } else if (match_req(request_msg, "get_auth_id")) {
+        handle_get_auth_id(mfout);
     } else if (auth_needed && !is_local) {
         auth_failure(mfout);
         if (sent_unauthorized) {
@@ -1613,6 +1684,8 @@ int GUI_RPC_CONN::handle_rpc() {
     } else {
         retval = handle_rpc_aux(*this);
     }
+
+#define XML_HEADER "<?xml version=\"1.0\" encoding=\"ISO-8859-1\" ?>\n"
 
     mfout.printf("</boinc_gui_rpc_reply>\n\003");
     mout.get_buf(p, n);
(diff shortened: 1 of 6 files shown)


TheAspens:
**Item 1**
Right now someone listening to the wire can watch the communications between client and the GUI. The values produced by the hash snprintf(buf, sizeof(buf), "%d%s", seqno, gstate.gui_rpcs.password); will always be the same sequence of values. As a result, someone watching simply needs to watch the initial call to get_auth_id and the following messages to build up a list of valid hashes to send. Then they can replay the call to get_auth_id and send the sequence of messages and make any calls they want.

Recommendation: I would recommend that get_auth_id returns both the auth_id and an auth_string where auth_string is a long string of random characters that is unique for each auth_id. The auth_string should then be incorporated into the hash generated by snprintf(buf, sizeof(buf), "%d%s", seqno, gstate.gui_rpcs.password); perhaps something like snprintf(buf, sizeof(buf), "%d%s%s", seqno, gstate.gui_rpcs.password, auth_string); This will make each hash unique to only one particular auth_id and thus cannot be reused by another client.

**Item 2**
Right now if someone is able to intercept the communications and then pass them to the client, then they have ability to replace one of the calls with their own call This allows the man in the middle to attach the client to a new project and thus run arbitrary code on the client.

Recommendation: I would recommend that the hash not be sent but instead be used as a key to sign the message. This will ensure that a man in the middle cannot alter the message.

**Item 3**
Once significant development occurs around this interface, it will be hard to maintain compatibility with that development when changes are required.

Recommendation: As a result an API version number should be added to the messages as part of this in order to allow for easier backwards compatibility when things change in the API. The version number should be present on inbound requests and sent in replies.  The existing API should be assumed to be version 1 and things using the modified API should be version 2.  The version number will be important since clients should also start developing knowing that the API response could change and handle it accordingly.

TheAspens:
**Item 4**
Since this GUI is designed to run on a local machine, I do not think that we should be fetching resources from the web (otherwise the GUI won't work if the user is not connected to the internet).  I believe that electron has a way to keep resources local and that mechanism should be used.

**Item 5**
This should be built as a single page application using a modern component based javascript framework such as React.js or Vue.js.  It is dramatically easier to develop sophisticated UI's using these frameworks vs using raw html/css/js.  Since this is new development, we need to use the modern techniques for this development.

AenBleidd:
Closing and reopening to retrigger CI builds

davidpanderson:
Kevin, thanks for your feedback.
I implemented your items 1 and 2; please review.

Item 3: there's already an "exchange_versions" RPC that can be used for this purpose.
It returns the version # of the client (major/minor/release) which should suffice.

Item 4: I just copied and modified one of the Electron sample apps.
Yes, one should bundle the CSS instead of downloading it.

Item 5: The sample RPC code can be used with any Javascript-based system.
It's not specific to Electron.

TheAspens:
David, I'm reviewing item #1 and #2 now.  An initial quick comment is that in `complete_post_request` you use \r\n\r\n to find the start of the body and in `authenticated_request `you use \n\n to find it.  I suspect one of those is wrong.

client/gui_rpc_server_ops.cpp
```diff
@@ -1289,6 +1291,76 @@ static bool complete_post_request(char* buf) {
...
+bool valid_auth(int id, long seqno, char* hash, char* request) {
+    char buf[256], my_hash[256];
+    //printf("valid_auth: id %d seqno %ld hash %s\n", id, seqno, hash);
+    for (unsigned int i=0; i<auth_infos.size(); i++) {
+        AUTH_INFO& ai = auth_infos[i];
+        if (ai.id != id) continue;
+        if (seqno <= ai.seqno) return false;
+        ai.seqno = seqno;
+        int n = request?(int)strlen(request):0;
+        snprintf(buf, sizeof(buf), "%d%s%s", seqno, gstate.gui_rpcs.password, ai.salt);
```
TheAspens: "%d%s%s" should be "%ld%s%s"

TheAspens:
I need to stop testing for now.  One thing I noticed is that the seqno advances on incorrect and correct requests.  This could cause an attacker to submit invalid requests with just the id and seqno because the server would increment but the client would not.  This could be resolved by only incrementing seqno if the request is valid.

TheAspens:
Last comment before I go for now. I'm running into issues getting the hashes to match.  Have you been able to get this working with the javascript code making calls into the client?

FYI - thanks for implement #1 and #2.

davidpanderson:
Thanks for catching those various things; I fixed them.
Yes, I'm seeing matching hashes now (on Windows; hopefully no platform dependencies).

TheAspens:
I was able to test Item #1 and Item #2 this morning and they appear to work like they should.  Thanks for adding those.

Some comments on the remaining items.

**Item #3:**   
Since this is stateless, API version numbers would need to be sent on each request.  Additionally, the GUI needs to be able to tell the core client what version of the API it is trying to use.  An example of why this is important is that if the authentication hash scheme changes in the future, then it is a messy situation if we don't know what version of the API the GUI is trying to use.  However, if we have that api version then we can handle it gracefully.  

This is just like when the core client makes a call to the scheduler and it tells the scheduler what version of the core client it is so that the scheduler can handle things differently based on that information.  

The API version number can be as simple as that used with the core client now (i.e. BOINC 7.14.2).  How to encode the api version number is the subject of much debate but here is one article that is useful: https://www.troyhunt.com/your-api-versioning-is-wrong-which-is/

**Item #4:**
My concern were these calls included on the index.html:
```
<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css">
<script src="https://code.jquery.com/jquery-3.2.1.js"></script>
<script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.1.4/Chart.bundle.min.js"></script>
```

**Item #5:**
React.js and Vue.js should be able to use your js api library.  My concern was again more about the index.html.  As development begins on the new UI, these component based frameworks need to be considered as they allow for dramatically more sophisticated UI's with much less headache/development effort then just using basic html/css/js/jquery development on its own.  

It sounds like index.html is really purely for demo and initial proof of concept

davidpanderson:
A GUI can get the client's version# and error out if it's one that it can't communicate with.
It's sufficient to do this on startup of the GUI.

TheAspens:
The issue is that a newer version of the client may have been released and installed.  The GUI will only know that it is a talking to a newer version of the client.  The GUI wouldn't know if the new features in the new version of client change the interface or expected behavior or not. 

However, if the client knows what version of the API the GUI is trying to use then it can handle the situation gracefully.

client/gui_rpc_server_ops.cpp
```diff
@@ -1289,6 +1291,79 @@ static bool complete_post_request(char* buf) {
...
+    p = strstr(buf, "Auth-Seqno: ");
+    if (!p) return false;
+    n = sscanf(p+strlen("Auth-Seqno: "), "%ld", &auth_seqno);
+    if (n != 1) return false;
+    p = strstr(buf, "Auth-Hash: ");
+    if (!p) return false;
+    n = sscanf(p+strlen("Auth-Hash: "), "%64s", auth_hash);
+    if (n != 1) return false;
+    char* request = strstr(buf, "\r\n\r\n");
+    if (request) request += 4;
```
AenBleidd: Can `"\r\n\r\n"` be moved to constant and it size be added to `request` variable instead of using magic numbers?
If for some reasons `"\r\n\r\n"` will be changed to. for example, `"\r\n"` it's very easy to forget to change magic number `4` to `2`

TheAspens:
@AenBleidd @lfield @SETIguy - what are your thoughts on the API version?  I feel it is important but as @davidpanderson indicated above, he doesn't think it is necessary.  Can you (and anyone else) chime in with your opinion?

TheAspens:
Actually - I'd like any/all of the @BOINC/committer's to chime in with their thoughts on the API version discussion above.

adamradocz:
As [Troy Hunt](https://www.troyhunt.com/your-api-versioning-is-wrong-which-is/) sais: 

> Software evolves, APIs must be versioned

tristanolive:
If the goal is to support arbitrary, third party GUIs by RPC, then there should be API versioning. This is nice for both long term compatibility and rapid new development, as functions can be added or changed in subsequent API versions without messing up earlier versions. A single version of the client could support multiple API versions to maximize compatibility with third party user interfaces. Also, there would very likely be multiple versions of the client with no changes made to this API, so it would be much simpler for a third party to just specify an API version rather than have to do a new release in order to support the latest client release (i.e. nothing has changed around GUI RPCs since version 7.16.2, but my GUI can't know in advance if it will work with 7.16.3; what it can know is that if 7.16.3 supports API v1, all is well).

Now, there is potentially an argument to be made that the first API version does not need to specify a version, as it is the default. It would have to remain the default after future version are released, which is probably ok. Are there any drawbacks to punting on versioning until there is a second version?

lfield:
@TheAspens I have seen many projects try to version APIs for noble reasons but very few actually manage to implement the logic to make it work. The only rule that I think needs following is that API changes should always be backwards compatible. Documenting the APIs is also important so versioning would also imply versioning the documentation. Hence why it is important for the documentation also to be in git.

lfield:
I have tried this and got the simple example to work. My main comment would be we are trying to implement our own authentication scheme. There must be a standard solution for this.

I then tried _<get_auth_id/>_ from my own js code in a Web browser. First it failed due to Cross-Origin Resource Sharing protection but once I disabled this the js failed due to invalid XML. There is no closing tag. 

One thing that I have been thinking about is how similar what we would like to achieve is to the CUPS printing system on Linux. The design is [described here](https://www.cups.org/doc/spec-design.html#TABLE1). Maybe this is a direction we could consider? It would not require the GUI RPC though.

AenBleidd:
I have never work with web rpc but I worked with webapi a bit. I don't know whether it is applicable in this case but usually web request looks like this:
`http://api.com/v1/login`
In this case when some parameters changed\added\deleted or when some other logic changed it is easy to support several versions of api in order to let both old and new clients work with webserver.
Again, I'm not sure about this particular case because I'm not an expert in this field but I'd like to have version support.

Rytiss:
I think versioning is only useful when doing complete overhauls. For gradual improvements, that tend to happen in BOINC, it would be enough to document changes, especially considering we have server releases now (e.g. `supported from client v 7.14.x and server v 1.0.3`).

adamradocz:
@TheAspens's link about versioning is very good, plus here is another one: [https://dzone.com/articles/best-practice-api-versioning-for-http-rest-interfa](https://dzone.com/articles/best-practice-api-versioning-for-http-rest-interfa)

Version number in the URL is usually not preferred.

TheAspens:
> Now, there is potentially an argument to be made that the first API version does not need to specify a version, as it is the default. It would have to remain the default after future version are released, which is probably ok. Are there any drawbacks to punting on versioning until there is a second version?

@davidpanderson indicated that the HTTP mechanism was completely broken prior to these fixes so this effectively represents the first version.  I think this point is valid, so defering this discussion until there is a breaking change (which would be the first time the api version # would really be required)

TheAspens:
All my items are resolved and I'm ready to merge.   However I saw two notes from above that need to be addressed

@AenBleidd's comment about `'\r\n\r\n'` and `4` should be named constants (since they are repeated) should be implemented

@lfield's comment 

> invalid XML. There is no closing tag.

TheAspens:
For @lfield's comment - I see full XML but ending with an unprintable character at the end.  See

```
<boinc_gui_rpc_reply>
<auth_id>0</auth_id>
<auth_salt>523ddb2c70f4127f534ccc136a5bb385</auth_salt>
</boinc_gui_rpc_reply>

```
That character is the "end of text" character (octal \003) generated here: https://github.com/BOINC/boinc/blob/d5c9a7b362de8305cb6363e490a438300b9767e6/client/gui_rpc_server_ops.cpp#L1700

Is that final character necessary?  

@lfield can you test removing that character and see if that fixes the issue?

davidpanderson:
I made the HTTP header delimited into a constant,
and removed the \003 reply delimiter for HTTP.
I'm not seeing any incomplete reply messages.

lfield:
This seems to work for me now. I can call it from Python and have been trying from JavaScript. One way around the CORS issue would be if the client could sever the Web UI files directly. How easy would it be to bolt something onto this port so that a HTTP GET would return an index.hml page from disk?

davidpanderson:
That's a good idea.  The client could serve HTML files in the data directory using GET requests.  I'll do that in a separate PR.

TheAspens:
Thanks for the addition items David.  I think all open items are resolved  (and discussion about the additional suggestions from Laurence will be in a separate PR) so I will merge.

PR #3202, opened 2019-06-30
[Client] Ignore 'skip_cpu_benckmark' flag if requested via RPC

### LINKED ISSUE #2358: Manual CPU benchmark should override <skip_cpu_benchmarks>
Even when <skip_cpu_benchmarks> is set in the cc_config.xml file, I feel that when I press Run CPU benchmarks in BOINC Manager, that it should do the benchmarks. Now it skips them no matter what. 

17/02/2018 15:37:25 |  | [benchmark] start_cpu_benchmarks(): Skipping CPU benchmarks

AenBleidd:
If CPU benchmark is requested via RPC by user, it should always run even if 'skip_cpu_bechmark' is set.

This fixes #2358

### DIFF
--- a/client/client_state.h
+++ b/client/client_state.h
@@ -380,7 +380,7 @@ struct CLIENT_STATE {
 
     void check_if_need_benchmarks();
     bool can_run_cpu_benchmarks();
-    void start_cpu_benchmarks();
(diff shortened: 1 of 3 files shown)


TheAspens:
Looks good.  Thanks.

PR #3211, opened 2019-07-05
make_project: default to https:// in project URL

davidpanderson:


marius311:
Just my 2 cents, but to me it seems like the order most people build a project are 1) build and test project locally 2) make it public and buy domain / set up https. Since running the make project script happens in step 1 before there's https, I think this PR probably complicates things for the typical use case.

lfield:
I think encouraging people to use https by default is good practice. You can use the [lets encrypt](https://httpd.apache.org/docs/trunk/mod/mod_md.html) Apache module or a self signed certificate.

adamradocz:
@marius311 I understand your point, on the other hand creating a script to generate a self-signed certificate for local use is not so hard. It can be developed for your docker project within a couple of minutes.

davidpanderson:
Is it possible for make_project to know whether the host's web server supports HTTPS?

tristanolive:
I think the script would have to try to access the URL via HTTPS. With the requests library, that might look something like:
```python
import requests
...
options.use_https = False
...
try:
    r = requests.get('https://%s/'%options.project_host, verify=False)
    r.raise_for_status()
    options.use_https = True
except requests.exceptions.HTTPError as e: 
    print("Unable to detect HTTPS support", e)
```

Uplinger:
After discussion on the Contributors Call, moving forward with guiding others towards using https by default is a positive thing.  I have tested this for both http and https implementations.  There is an option in the make_file that allows for http to be used still.  

--url_base           default: https://PROJECT_HOST/

As for checking if https is available wouldn't work since apache might not be set up entirely for the system or apache is disabled temporarily.

For getting a certificate or creating one, you can refer to these two pages for help.
https://hostadvice.com/how-to/configure-apache-with-tls-ssl-certificate-on-ubuntu-18/
and
https://letsencrypt.org/getting-started/

mielouk:
Thank you for making transport encryption for all projects mandatory. I had an[ instance ](https://github.com/BOINC/boinc/issues/3352) where there was a strong suspicion that some malicious entity intervened to hack my computer via unencrypted BOINC traffic. Thx again!

PR #3214, opened 2019-07-11
Client: support HTTP GET of files over GUI RPC connections

davidpanderson:
If the client receives a GET HTTP connection,
it looks for a file of that name (default index.html) in the BOINC data directory
(only .html files, no directory traversal) and returns it.
The goal is to get rid of cross-site constraints in Web-based GUIs.


lfield:
I have just tested this and have a few comments.

1. Should we at least for now only enable this if a configuration flag is set in the client configuration?
1. Should we also allow javascript (*.js) files?
1. If a not supported extension is tried, the connection is reset, it should probably fail with HTTP 403?
1. Should we use a sub directory of the data directory so a not to pollute the root?

Ferroin:
My thoughts as an end user:

1. Yes, with the caveat that I'd keep it as a configuration flag long-term.  Users who have some different GUI and are security minded will probably want to disable this.
2. Yes, and probably CSS files too.  Provided that the client doesn't overwrite any of the files after the first startup, this would provide an easy way for users to create a custom web UI for the app.
3. Yes, but unless the implementation supports serving multiple requests over the same connection, it should still reset the connection after sending the response.
4. Yes, but if this is done you can probably drop the file type limitation and allow multiple directory levels provided you don't follow symbolic links.

davidpanderson:
I made the suggested changes.

lfield:
Thanks. I have tested the changes and they work as advertised. I just have a comment on the sub directories. The code does not enforce the use a specific sub-directory nor allow the configuration of a specific directory. Do we want to do this? Sub directories do already work for example:
http://localhost:31416/manager/index.html

However http://localhost:31416/manager/ does not resolve to http://localhost:31416/manager/index.html

This work correctly for http://localhost:31416 which resolves to http://localhost:31416/index.html

davidpanderson:
I got rid of the default to index.html.
I don't think we need any more configuration options.

lfield:
Am happy to merge.

PR #3179, opened 2019-06-04
[WIP] Assorted fixes

JamesMolson:
Do not merge yet, still in progress.

This contains fixes for FreeBSD, NetBSD, MacOS, SunOS, and probably more.

Issue #3166, opened 2019-05-28
mac build + unix makefiles


JamesMolson:
Before @CharlieFenton et al say, I know this is not the supported method.

Some notes on the suggested xcode method, it does not build boinc_api_fortran.o and I would prefer to use gcc since our work needs a fortran compiler, and also we can use the same compiler across all platforms. Note - we only build the core libraries and not the client, etc.

Problem 1:
```
  OBJC     libboinc_graphics2_la-macglutfix.lo
macglutfix.m:34:9: fatal error: 'MultiGPUMig.h' file not found
#import "MultiGPUMig.h"
        ^~~~~~~~~~~~~~~
1 error generated. 
```

Fixed by running the api folder:
```
mig  -header MultiGPUMig.h -sheader MultiGPUMigServer.h MultiGPUMig.defs
```

Problem 2:
```
  CXX      libboinc_la-procinfo_mac.lo
procinfo_mac.cpp: In function 'int procinfo_setup(PROC_MAP&)':
procinfo_mac.cpp:125:14: error: 'strchr' was not declared in this scope
  125 |         lf = strchr(p.command, '\n');
      |              ^~~~~~
procinfo_mac.cpp:37:1: note: 'strchr' is defined in header '<cstring>'; did you forget to '#include <cstring>'?
   36 | #include "mac_branding.h"
  +++ |+#include <cstring>
   37 | 
procinfo_mac.cpp:130:42: error: 'strcasestr' was not declared in this scope
  130 |         p.is_boinc_app = (p.id == pid || strcasestr(p.command, "boinc"));
      |                                          ^~~~~~~~~~
make[2]: *** [libboinc_la-procinfo_mac.lo] Error 1
```

Fixed by including the above mentioned header.

Problem 3:

```
  CXX      libboinc_la-mac_spawn.lo
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/CoreGraphics.framework/Headers/CGContext.h:21,
                 from /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/CoreGraphics.framework/Headers/CGBitmapContext.h:9,
                 from /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/CoreGraphics.framework/Headers/CoreGraphics.h:11,
[...]

PR #3217, opened 2019-07-11
Add potentially missing mac header.

lfield:
From PR #3179

### DIFF
--- a/lib/procinfo_mac.cpp
+++ b/lib/procinfo_mac.cpp
@@ -19,6 +19,7 @@
 
 #include "config.h"
 #include <cstdio>
+#include <cstring>
 
 #include <ctype.h>
 #include <sys/types.h>


AenBleidd:
I don't like 'potential' fixes. According to out CI mac build is successful without this fix. Is it really necessary?

lfield:
@CharlieFenton While this is not the official way to build the MAC client, @JamesMolson is building the libraries on MAC and other platforms for his application. In general this seems to be working for him and he is suggesting this change so that this use case willa work for him and others. I do not see why this change should not be considered.

CharlieFenton:
As long as someone is willing to keep the make files up to date and test that the results work properly, that is fine, especially since I have phased out of working on BOINC. The graphics library, screensaver and graphics apps are the parts I suspect could be the most difficult to build using make files. Note that MultiGPUMig.defs must be built first, using the MIG (Mach Interface Generator) compiler not clang.

CharlieFenton:
I might add, just for the sake of clarity, that the Mac does not install the GNU GCC compiler, but uses the Apple Clang compiler, On Macs with the developer tools (Xcode) installed, gcc is a symbolic link to clang.  In his PR #3166 @JamesMolson [was able to build with Apple Clang using make files](https://github.com/BOINC/boinc/issues/3166#issuecomment-498655363)

TheAspens:
On the contributor call today we agreed that it was reasonable to support make to help build the research applications.  Thus I am asking @ShanghaiTimes to take a look and test this change if he is able.

ShanghaiTimes:
On the first appearance, I agree with Charlie.  Mac does not install gcc and in fact, is moving even further away from having third party programs and programming options on board. So others wishing to simplify the process on Mac by using just XCode have enough to do even now making sure the steps as outlined in the HOW-TO files are followed. 
I've just opened XCode and tested the build again of BOINC_Client and it's built and run cleanly. I haven't done this for months and months now, so a little surprised that it still built ok.
This is XCode Version 10.3 on Mohave, and I have everything installed according to the howto document that I recently updated.
I'd prefer to await XCode 11's release before proceeding too much further because that's sure to change a lot of things. Even more things will be removed, and probably added.
In short - I agree with Charlie. I reject this change on the same grounds.
Building headers and adding things to XCode is a pathway to ruin. Caveat Utilitor

This is the output of the build and run just now.
`20-Sep-2019 08:06:46 [---]    (to change preferences, visit a project web site or select Preferences in the Manager)
20-Sep-2019 08:06:46 [---] Setting up project and slot directories
20-Sep-2019 08:06:46 [---] Checking active tasks
20-Sep-2019 08:06:46 [---] Setting up GUI RPC socket
20-Sep-2019 08:06:46 [---] Checking presence of 0 project files
20-Sep-2019 08:06:46 [---] This computer is not attached to any projects
20-Sep-2019 08:06:46 Initialization completed
20-Sep-2019 08:06:46 [---] Suspending GPU computation - computer is in use`

ChristianBeer:
As this is rejected by two developers I'm closing the PR. If it should be added nevertheless please open a new PR with steps to reproduce a failure. We could then add a CI job for this special use case.

PR #3243, opened 2019-08-06
Add Discord badge

AenBleidd:

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # Social
 
-[![Slack](https://img.shields.io/badge/slack-boincworkspace-blue?logo=slack)](https://join.slack.com/t/boincworkspace/shared_invite/enQtNzA3MTQ4NDA0Njc4LTM4NTEyZTY1MWEwZjAyNTVmOTg3NDA2NjVjZDQzMWQ3NmFjYTc4MTNkNTEyNjRkOTQ4MWU5Nzk3NzRhNTg0NDI)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,7 +4,7 @@
 
 # Social
 
-[![Slack](https://img.shields.io/badge/slack-boincworkspace-blue?logo=slack)](https://join.slack.com/t/boincworkspace/shared_invite/enQtNzA3MTQ4NDA0Njc4LTM4NTEyZTY1MWEwZjAyNTVmOTg3NDA2NjVjZDQzMWQ3NmFjYTc4MTNkNTEyNjRkOTQ4MWU5Nzk3NzRhNTg0NDI)
+[![Slack](https://img.shields.io/badge/slack-boincworkspace-blue?logo=slack)](https://join.slack.com/t/boincworkspace/shared_invite/enQtNzA3MTQ4NDA0Njc4LTM4NTEyZTY1MWEwZjAyNTVmOTg3NDA2NjVjZDQzMWQ3NmFjYTc4MTNkNTEyNjRkOTQ4MWU5Nzk3NzRhNTg0NDI) [![Discord](https://img.shields.io/badge/discord-boincworkspace-blue?logo=discord)](https://discord.gg/kebpfnr)
```
RoboticMind: Shouldn't the name be different than boincworkspace? The discord's called BOINCNetwork so shouldn't that be the name?

AenBleidd: You're definitely right! Copy-paste error, my bad. Thank you for the catch, @RoboticMind. Fixed

RoboticMind:
Maybe also make the color difference of the box so you can find it a bit easier and so the logo look as off? Maybe something like https://img.shields.io/badge/discord-boincnetwork-darkgreen?logo=discord instead?

AenBleidd:
I think it's better to keep them the same color for next reasons:
- color should mean smth (green is good, yellow is not good but still not bad, red is bad in case of build status), in this case blue means nothing so it's perfectly well for something neutral (social networks), so as far as both links are related to social network it could be better to keep them one color
- I don't want main file to look like colorful parrot
- these links are not something that will be used often, actually, it's one-time link to use but you will see it every time when opening this repo, so they should not catch your eye with different colors.

RoboticMind:
Maybe a light blue for would work then? Just want something that doesn't have the exact same color as the Discord logo.

AenBleidd:
Dark blue image
Blue image
Light blue image

If you want another shade of Blue, Dark Blue is better as for me, Light Blue is too pale

AenBleidd:
image image

But having them the same color looks also fine and better (from my prospective). And they are not looks similar

PR #3248, opened 2019-08-12
Remove boincsynergy.com from the stats site list

### LINKED ISSUE #3244: statistics site boincsynergy.com obsolete?
The link to http://www.boincsynergy.com/stats/index.php from the BOINC statistics page leads to an ISP URL, so it might be that this site is no longer maintained. 

**Steps To Reproduce**
1. Look at the stats.php of any BOINC project site, e.g. https://www.cosmologyathome.org/stats.php

2. Click on the link and you will notice  that the link is dead.

**Expected behavior**
Only valid links should be provided via the stats.php page shipped with the BOINC server code.

As a fix, simply edit html/inc/stats_sites.inc and remove the lines:

```
     array("http://www.boincsynergy.com/stats/index.php",
         "BOINC Statistics for the WORLD!",
         "developed by Zain Upton"     
),
```
Should be done upstream with the web server code once we are sure that this site will not be resurrected.

Rytiss:
According to archive.org, the site is down at least since February 2019, so removing it from the list.

Fixes #3244

### DIFF
--- a/html/inc/stats_sites.inc
+++ b/html/inc/stats_sites.inc
@@ -31,9 +31,6 @@ $cpid_stats_sites = array(
     array("BOINCstats",
         "http://boincstats.com/en/stats/-1/user/detail/%s"
     ),
-    array("BOINC Statistics for the WORLD!",
-        "http://www.boincsynergy.com/stats/boinc-individual.php?cpid=%s"
-    ),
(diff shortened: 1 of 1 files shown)


TheAspens:
Thanks for the change.

PR #3255, opened 2019-08-23
Allow the validator to flag compared results as `runtime_outlier`

amereghe:
Fixes # https://github.com/BOINC/boinc/issues/3256

**Description of the Change**

Give the possibility to the validator to set `result.runtime_outlier=true` all the compared results based on the content of the results after validation. This can be achieved via removing tagging the second result in `compare_results` as `const`.

**Alternate Designs**

Setting `result.runtime_outlier=true` could be performed outside of `files_match`, but this would have implied either to store the value of the flag in a temporary variable and use it somewhere else, or parse again the validation file.

**Release Notes**

Give the possibility to the validator to set `result.runtime_outlier=true` all the compared results based on the content of the results after validation.

### DIFF
--- a/sched/sample_bitwise_validator.cpp
+++ b/sched/sample_bitwise_validator.cpp
@@ -65,7 +65,7 @@ void validate_handler_usage() {
 }
 
 
-bool files_match(FILE_CKSUM_LIST& f1, FILE_CKSUM_LIST& f2) {
+bool files_match(RESULT &r1, FILE_CKSUM_LIST& f1, RESULT &r2, FILE_CKSUM_LIST& f2) {
     if (f1.files.size() != f2.files.size()) return false;
     for (unsigned int i=0; i<f1.files.size(); i++) {
         if (f1.files[i] != f2.files[i]) return false;
@@ -113,14 +113,14 @@ int init_result(RESULT& result, void*& data) {
 }
 
 int compare_results(
-    RESULT & /*r1*/, void* data1,
-    RESULT const& /*r2*/, void* data2,
+    RESULT & r1, void* data1,
+    RESULT & r2, void* data2,
     bool& match
 ) {
     FILE_CKSUM_LIST* f1 = (FILE_CKSUM_LIST*) data1;
     FILE_CKSUM_LIST* f2 = (FILE_CKSUM_LIST*) data2;
 
-    match = files_match(*f1, *f2);
+    match = files_match(r1, *f1, r2, *f2);
     return 0;
 }
 
--- a/sched/sample_substr_validator.cpp
+++ b/sched/sample_substr_validator.cpp
@@ -76,7 +76,7 @@ int init_result(RESULT& r, void*&) {
     return 0;
 }
 
-int compare_results(RESULT&, void*, RESULT const&, void*, bool& match) {
+int compare_results(RESULT&, void*, RESULT &, void*, bool& match) {
     match = true;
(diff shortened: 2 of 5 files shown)


TheAspens:
@davidpanderson - can you take a look at this proposed change?

davidpanderson:
I don't understand the need for this.
In the current design, you can flag a job as a runtime outlier in init_result().
Why would you need to look at two jobs to decide if they're runtime outliers?

amereghe:
Thanks, @davidpanderson , for looking into this.

As I said in issue https://github.com/BOINC/boinc/issues/3256, a SixTrack simulation which does not follow the fpops estimate can be perfectly fine; an early end of the simulation may not necessarily be a sign that the input is malformed or the result is malicious. In fact, if the beam gets unstable and is regularly lost, the simulation stops earlier (the aim of our simulations is exactly to catch the boundary between stable and unstable motion of the accelerator beam). This can also be the case for a portion of the tracked beam, with the same early end.

Hence, we need to validate the results first, to be sure that the early stop is a real one (i.e. a genuine result of the simulation) and not malicious; only at this stage, we can flag the concerned tasks as runtime outliers, such that the estimation of the actual CPU speed of the hosts is not  jeopardised.

RichardHaselgrove:
@amereghe: I think you misunderstand the concept and purpose of "runtime outlier".

It was introduced because SETI@Home has exactly the same problem as you: a perfectly valid and successful result can have a very short runtime, because RFI interference in the source data can overflow the designed signal space.

If a succession of these overflow task runtime records is allowed to accumulate in the app_version and host_app_version tables, runtime estimates for the host and project as a whole become distorted: in extreme cases, a project can start aborting perfectly normal tasks because the runtime estimate becomes so low that 'runtime estimate exceeded' becomes a normal outcome.

The 'runtime outlier' flag, as the name suggests, is concerned with keeping the project runtime estimates within normal values, and nothing else. It's an important safeguard.

amereghe:
Thanks, @RichardHaselgrove, for the reply, but that's exactly what we need, i.e. to keep

> the project runtime estimates within normal values

for exactly the same reasons that you report. If we flag a valid result which lasted too short as runtime outlier, we discard its figures for updating runtime estimates.

For instance, as a volunteer, the estimated runtime of tasks on my desktop pc were getting shorter and shorter. With such a change, they got back to more reasonable numbers.

RichardHaselgrove:
But that's how the flag was originally designed. @davidpanderson can confirm, but I believe that tasks reported by the client as an error are already excluded from the runtime estimates: there's no need for a second bite of the cherry after validation. Please indicate where you believe these "malicious" (your term) values are coming from.

amereghe:
I agree with you that error results (invalid) do not need the flagging. Indeed, I am talking about valid results (i.e. not malicious) which finish early

davidpanderson:
The logical place to check for runtime outliers (e.g. jobs that finished early) is init_result(),
not compare_results().
It's fine to mark invalid results as runtime outliers; they'll get ignored anyway.

amereghe:
So, @davidpanderson , what you suggest is to read the validation file in init_result() to verify that the simulation finished early. It seems to me a bit redundant - we parse the file already in compare_results() and get the appropriate info there

ChristianBeer:
@amereghe This PR has a conflict with current master. Please adjust the code or close the PR if not needed anymore.

Issue #3241, opened 2019-08-02
Finish migration to python3


TheAspens:
Pull Request #2477 modified the scripts necessary so that creating and updating a project worked with python3 instead of python2.  I believe the following additional python scripts need to be updated to finish the migration to python3:

- ./stripchart/samples/parse_config
- ./tools/parse_config
- ./tools/dbcheck_files_exist
- ./tools/check_project
- ./tools/appmgr
- ./sched/start

davidpanderson:
Fixed in #4043, which works with both 2 and 3

PR #2477, opened 2018-04-20
try to make python scripts work with python 3

davidpanderson:
Changes:
print()
except XXX as e (rather than except XXX: e)
from Boinc import XXX rather than import XXX
    (I'm not sure how it worked before)

This gets as far as importing the MySQLdb module, which doesn't work with python 3.
Need to use mysqlclient instead:
https://pypi.org/project/mysqlclient/

PR #3259, opened 2019-08-26
Python2 compatible changes on print() and Exception()

smoe:
Hello,

Please grant this patch some scrutiny that is meant to address (bits of) https://github.com/BOINC/boinc/issues/3241 and may continue work in https://github.com/BOINC/boinc/pull/2477 .

The 'from __future__ import print_function' likely preserves compatibility with Python2. I suggest to add this until the transition to Python3 is completed.

Best,
Steffen

### DIFF
--- a/py/Boinc/add_util.py
+++ b/py/Boinc/add_util.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+from __future__ import print_function
 
 # $Id$
 
@@ -130,7 +131,7 @@ def translate_database_arg(database_table, arg, value):
     if len(results) > 1:
         print('Too many %ss match "%s": '%(arg,value), sys.stderr)
         for result in results:
-            print ('   '+result.name, sys.stderr)
+            print('   '+result.name, sys.stderr)
         raise SystemExit
     return results[0]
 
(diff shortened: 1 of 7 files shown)


LocutusOfBorg:
Awesome work! We should probably use some python tool to check if anything is missing, but I really like the changes!

LocutusOfBorg:
Can you please also fix this?
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=903762

smoe:
> Can you please also fix this?
> https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=903762

Was there a second exception that I have missed? Otherwise that fix is in.

TheAspens:
@smoe - just to confirm.  The intent is that this will allow support for both python 2.3 and 3?
If so I'll set up a vm to test this on centos 6 (installing python 2.3 only) and centos 8 (installing python 3.0 only).

smoe:
I am very confident wrt the print instructions.  The exceptions may need a bit of extra love.

TheAspens:
@smoe - PR #4043 should allow support of both python2 and python3.  Is there anything from this PR that should be kept or should it be closed?

smoe:
Not having the opportunity to confirm this myself, trusting upstream though. Many thanks for the support of Python3.

PR #3261, opened 2019-08-28
Fixes a bug reg. "ncpus" in cc_config.xml in comb. with "Read config files".

cwallbaum:
**Description of the Change**
https://boinc.berkeley.edu/wiki/Client_configuration states that you can put "-1" into the ncpus tag to use all cores. But in this case, "Read config files" doesn't work, so you have to restart the client.

This patch fixes the behaviour by introducing an additional variable which stores the original value.

**Release Notes**
If you set ncpus in cc_config.xml to "-1", you no longer have to restart the client; "Read config files" now works in this case.

### DIFF
--- a/client/cpu_sched.cpp
+++ b/client/cpu_sched.cpp
@@ -1627,7 +1627,10 @@ ACTIVE_TASK* CLIENT_STATE::get_task(RESULT* rp) {
 void CLIENT_STATE::set_ncpus() {
     int ncpus_old = ncpus;
 
-    if (cc_config.ncpus>0) {
+    if (cc_config.ncpus == -1) {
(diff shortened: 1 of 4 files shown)


davidpanderson:
Thanks.  I did the same thing (slightly simpler) in #3268.
Please test that.

Note: the wiki documentation was wrong.  0 means use all CPUs.
(-1 works too).

Note 2: computing preferences (not this feature) should be used to limit CPU usage.

TheAspens:
@cwallbaum - I'm going to close this in favor of #3268 please take a look at the discussion there about the difference between the two versions.

Thank you for opening this pull request!  I look forward to seeing more from you in the future!

Issue #3260, opened 2019-08-27
The Fedora + Debian/Ubuntu packages better lose their patches


smoe:
So far so obvious, there should not be any need for a distro's packages to have any patches since all changes are either:
 * not necessary -> don't patch
 * of general interest -> have it upstream
 * an indication of a deficiency to configure/install BOINC -> improve upstream

I would like to leave this issue here as an anchor and over time revisit all patches that have accumulated downstream over the past decade for [Debian](https://salsa.debian.org/pkg-boinc-team/boinc/tree/master/debian/patches) and [Fedora](https://src.fedoraproject.org/rpms/boinc-client/tree/master).

RichardHaselgrove:
Could we also please consider the equivalent Fedora patch list: https://src.fedoraproject.org/rpms/boinc-client/tree/master

We were directed to that list during last week's contributor call, because it currently contains patches which were rejected by 'upstream' BOINC, but considered appropriate for the Fedora target audience. They perhaps fall into your third category.

Germano0:
I am the author of Fedora/EPEL patches. @RichardHaselgrove has well described why we ship them

smoe:
With PR #3285 there shall now be a bit of a break as I get the impression that besides some rather easy ones there are already enough discussion points from which us Debian folks can learn about your preferred form and content of patches. Also, the imho most important bits you have seen, @LocutusOfBorg may find some more. The branches from which the respective pull requests are correspond to the file names in debian/patches.

vwbusguy:
Ah, this seems to explain why my BOINC manager quit starting BOINC in Fedora.  I now just run boinc_client every time I log into Gnome and then have boinc manager connect to localhost where this all used to just happen automatically before by having boinc-manager as a startup app in gnome.

smoe:
> Ah, this seems to explain why my BOINC manager quit starting BOINC in Fedora. 
@vwbusguy , I fail to grasp what exactly makes your boincmgr unhappy.

vwbusguy:
I just grabbed the source there and removed two patches from the spec and rebuilt - the first that prevents idle detection and the other that prevents the boinc manager from starting the boinc_client and now everything is happy again.

With respect to the FP package maintainer, I'm willing to setup a copr repository with these patches removed if it might be helpful for others that prefer the upstream capabilities.

AenBleidd:
@vwbusguy, just FY, there were a lot of hot discussions about this functionality:
#3028
#2969
#3094
#3105

[...]

PR #3262, opened 2019-08-29
Revisited Debian patch 'more_maxpathlen.patch'

smoe:
This patch is about 7 years old in Debian. I don't recally the exact original. Typically the compiler warns about something or say, the (admittedly, mostly irrelelvant) Hurd platform does not have a limit in the maxpathlen and so one has a look into that.

The patch is about using MAXPATHLEN both to limit the number of bytes copied into a char[] and to prepare the size of the char[]s for it. When there are MAXPATHLEN characters allowed, a buffer needs MAXPATHLEN+1 characters max because of the terminating 0. snprintf adds the terminating 0 itself and the "n" in snprintf helps to avoid buffer overruns.

When looking at this code, I wish the boinc_getcwd would be allowed to fail. It happens all the time to me that I have removed a directory that I was still in with another shell. And the same could well happen with some magic performed on the /var partition or whatever magic all these clouds and containers may now be preparing to come up with. This has not been addressed in this patch but should possibly be addressed as a separate issue.

### DIFF
--- a/api/graphics2_util.cpp
+++ b/api/graphics2_util.cpp
@@ -38,9 +38,9 @@
 
 #ifdef __EMX__
 static key_t get_shmem_name(const char* prog_name) {
-    char cwd[MAXPATHLEN], path[MAXPATHLEN];
+    char cwd[MAXPATHLEN+1], path[MAXPATHLEN+1];
     boinc_getcwd(cwd);
-    sprintf(path, "%s/init_data.xml", cwd);
+    snprintf(path, sizeof(path), "%s/init_data.xml", cwd);
     return ftok(path, 2);
 }
 #else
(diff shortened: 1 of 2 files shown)


LocutusOfBorg:
thanks for this, it might solve even security issues or possible buffer overflows!

api/graphics2_util.cpp
```diff
@@ -50,14 +50,14 @@ static void get_shmem_name(const char* prog_name, char* shmem_name) {
     APP_INIT_DATA aid;
     int retval = boinc_get_init_data(aid);
     if (retval) aid.slot = 0;
-    sprintf(shmem_name, "boinc_%s_%d", prog_name, aid.slot);
+    snprintf(shmem_name, MAXPATHLEN+1, "boinc_%s_%d", prog_name, aid.slot);
```
AenBleidd: This code is not safe. I'd suggest to change the function signature and add `shmem_name_size` input parameter at the end

ChristianBeer: This change makes it more secure than it is now. Do you think changing the signature is really needed? It might break something for science apps (I haven't checked usage of `get_shmem_name`) that the projects have to deal with. The changes proposed in this PR on the other hand are sane.

AenBleidd: Yeah, you're probably right. Changing signature without having API versioning is a bad idea.
Also I have to admit that this API is not safe at all and should be revisited.

AenBleidd (approved):

PR #3266, opened 2019-08-31
client: don't allow ".." or "%" in logical names

davidpanderson:
This fixes a vulnerability where hacked projects could
write files outside the BOINC data directory.

Also: error-check parsing of file refs.

This doesn't handle Rytis's requirement; we can do that separately.

### DIFF
--- a/client/client_types.cpp
+++ b/client/client_types.cpp
@@ -107,7 +107,12 @@ int parse_project_files(XML_PARSER& xp, vector<FILE_REF>& project_files) {
         if (xp.match_tag("file_ref")) {
             FILE_REF file_ref;
             retval = file_ref.parse(xp);
-            if (!retval) {
+            if (retval) {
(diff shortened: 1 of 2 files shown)


tristanolive:
Tested and verified that including ".." in the path or "%" (e.g. "%2e%2e/") results in an error:

> 9/6/2019 3:58:48 PM |  | can't parse file_ref in workunit: file name is empty or has '..'
> 9/6/2019 3:58:48 PM | test | [error] Can't parse workunit in scheduler reply: file name is empty or has '..'
> 9/6/2019 3:58:48 PM | test | Scheduler request completed: got 1 new tasks
> 9/6/2019 3:58:48 PM | test | [error] State file error: missing task test_29028_1567799889.612093
> 9/6/2019 3:58:48 PM | test | [error] Can't handle task test_29028_1567799889.612093_0 in scheduler reply

PR #3268, opened 2019-09-03
client: make "reread config files" work for ncpus

davidpanderson:
Before, if you changed cc_config.xml ncpus, you had to restart the client.

Note: don't use cc_config.xml to set your CPU usage.  Use computing preferences.

### DIFF
--- a/client/cpu_sched.cpp
+++ b/client/cpu_sched.cpp
@@ -1627,13 +1627,23 @@ ACTIVE_TASK* CLIENT_STATE::get_task(RESULT* rp) {
 void CLIENT_STATE::set_ncpus() {
     int ncpus_old = ncpus;
 
+    // config file can say to act like host has N CPUs
+    //
+    static bool first = true;
+    static int original_p_ncpus;
+    if (first) {
+        original_p_ncpus = host_info.p_ncpus;
+        first = false;
+    }
     if (cc_config.ncpus>0) {
         ncpus = cc_config.ncpus;
-        host_info.p_ncpus = ncpus;
-    } else if (host_info.p_ncpus>0) {
-        ncpus = host_info.p_ncpus;
+        host_info.p_ncpus = ncpus;  // use this in scheduler requests
     } else {
-        ncpus = 1;
+        host_info.p_ncpus = original_p_ncpus;
+        ncpus = host_info.p_ncpus;
+    }
+    if (ncpus <= 0) {
+        ncpus = 1;      // shouldn't happen
     }
 
     if (global_prefs.max_ncpus_pct) {


client/cpu_sched.cpp
```diff
@@ -1627,13 +1627,23 @@ ACTIVE_TASK* CLIENT_STATE::get_task(RESULT* rp) {
...
     int ncpus_old = ncpus;
 
+    // config file can say to act like host has N CPUs
+    //
+    static bool first = true;
+    static int original_p_ncpus;
+    if (first) {
+        original_p_ncpus = host_info.p_ncpus;
+        first = false;
+    }
```
TheAspens: I believe that these lines would be better implemented as simply:

`static int original_p_ncpus = host_info.p_ncpus;`

If I understand things correctly, a static function variable is allocated when the program starts but it is initialized the first time control passes through its initialization block.  See https://stackoverflow.com/questions/55510/when-do-function-level-static-variables-get-allocated-initialized for a discussion about this.

TheAspens:
@davidpanderson - can you clarify you thinking about why you prefer this approach rather than #3261?  After reviewing both of these changes, I see it as follows.

In this change, you are storing the actual measured value of the number of cpus of the host in a concealed variable local to specific function.  This localizes the storage of that information in a location separate from the host_info structure that exists to store information about the host.

In #3261 the actual measured value of the number of cpus of the host is stored on the host_info structure which is where I would expect to go to find information about the host.

davidpanderson:
The config.xml ncpus mechanism is to make it look (for debugging purposes)
as if the host has N CPUs.
So it makes sense for this mechanism to store the old value for undo
purposes.

TheAspens:
Ah - I see.  If I understand correctly, what you are saying is that the config.xml ncpus mechanism should only be used for debugging purposes - never for actual use.  And when it is being used, BOINC should behave as if that value from config.xml ncpus is the actual value in all cases.  As a result, there is no reason why host_info structure should store both the "in effect" value and the "original".

davidpanderson:
There was also some confusion because the wiki doc said that
0 means zero CPU, and -1 means all CPUs.
This was wrong.  You can't have zero CPUs.
Anything nonpositive means all CPUs.

TheAspens:
@davidpanderson - did you see my comment about how I think that this:

```
static bool first = true;
static int original_p_ncpus;
if (first) {
    original_p_ncpus = host_info.p_ncpus;
    first = false;
}
```

could be replaced with this:
```
static int original_p_ncpus = host_info.p_ncpus;
```

based on how static initialization works (see here for a good discussion: https://stackoverflow.com/questions/55510/when-do-function-level-static-variables-get-allocated-initialized)

davidpanderson:
It could be, but I think it's less clear.

TheAspens:
One last question:

Can this method be called from multiple threads?  If so, this code is not thread safe and we should use the the static initializer as that is thread safe.  If this method cannot be called from multiple threads, then that doesn't matter.

davidpanderson:
it can't be called from multiple threads

TheAspens:
Sounds good.  Thanks for the changes.

cwallbaum:
Your patch works as expected, thanks.

PR #3269, opened 2019-09-04
Fix broken master after 4af1ef748db93bec1fbe10cd4ea7a862cb222f86

AenBleidd:
Remove deleted files tools/db_query and tool/watch_tcp from tools/Makefile.am


TheAspens:
Thanks for the quick fix.

PR #3273, opened 2019-09-05
scheduler: Fix incorrect clamping of host' fraction_available

stream1972:
**Description of the Change**
Code around this point suggests that clamping should be done to FRAC_MIN define, but it's finally set to another, hardcoded and unacceptable small, value. 

**Release Notes**
N/A


TheAspens:
Yup makes sense.  Thanks for the contribution.

PR #3274, opened 2019-09-05
wu_check: Skip WUs without input files

stream1972:
**Description of the Change**
This tool cannot handle workunits without input files (e.g. if task parameters are set from command line). Accesses uninitialized memory, trying to open a file with a garbage name, prints a garbage on screen. In 'repair' mode, it will erroneously kill these completely normal workunits.

The patch silently skips such types of workunits.

**Release Notes**
N/A

### DIFF
--- a/sched/wu_check.cpp
+++ b/sched/wu_check.cpp
@@ -71,7 +71,9 @@ int handle_result(DB_RESULT& result) {
         );
         return 1;
     }
-    get_file_path(wu, path);
+    if (get_file_path(wu, path)) { // no input files
+        return 0;
+    }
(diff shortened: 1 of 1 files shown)


Rytiss:
This does not follow the coding style as described in the [wiki](https://boinc.berkeley.edu/trac/wiki/CodingStyle#braces), namely - it should either be a one-liner or have curly braces (my preference is the latter).

The patch content itself seems valid.

davidpanderson:
Need braces.  Other than that, fine.

stream1972:
Braces style fixed

PR #3276, opened 2019-09-10
Declaring iterators where they are used, not "globally"

smoe:
This is one of the ancient patches in Debian. When substituting
g++ with another compiler, here it was clang, one often sees new
warnings. No idea what the exact complaint was, but presumably
it was noted that "j" is used multiple times and in very different
contexts.

This patch declares "j" multiple times to indicate that there is
no information flow between the different parts of the code. The
human reader of the code may appreciate this more than the compiler.


AenBleidd:
Looks fine, thanks.
Merging...

PR #3277, opened 2019-09-10
cppcheck: avoid chance to read from uninitialised data

smoe:
This was once found by cppcheck and lives for a long time in
the Debian distribution. The decrypt_public function should set the data
attribute, so this is of no functional danger. But as much
as this irritated cppcheck, a human reader is confused at first
sight. Also, at some point the function called may be happy to
know that it is not overwriting a string or the well-defined
string helps some debug output.

Most notably, with this patch, the final copy will always copy
reasonable data, even if the decrypt_public function changes its return
codes.

Anchor: https://github.com/BOINC/boinc/pull/3276

### DIFF
--- a/lib/crypt.cpp
+++ b/lib/crypt.cpp
@@ -349,6 +349,7 @@ int check_file_signature(
     char clear_buf[MD5_LEN];
     int n, retval;
     DATA_BLOCK clear_signature;
+    clear_buf[0]=0;
 
     n = (int)strlen(md5_buf);
     clear_signature.data = (unsigned char*)clear_buf;


lib/crypt.cpp
```diff
@@ -349,6 +349,7 @@ int check_file_signature(
     char clear_buf[MD5_LEN];
     int n, retval;
     DATA_BLOCK clear_signature;
+    clear_buf[0]=0;
```
AenBleidd: It is an output variable.
If OpenSSL reads smth from output variable - it'd weird.
I'm not sure this is necessary here, for me looks like false positive detection

smoe: It is about the line that assigns a value to "answer" in https://github.com/BOINC/boinc/blob/fe1ddbafa7735160e5bd3587618aedbc527e957b/lib/crypt.cpp#L365
I know that a value should be assigned if retval is 0, but cppcheck does not know, so, yes, it is false positive, in a way, that is as long as you trust that function you called. I would nonetheless prefer to have it a bit more robust and see the patch in. Also, maybe you want to set the answer variable when bailing out early.

AenBleidd (approved):

PR #3279, opened 2019-09-10
cppcheck: independent iterators declared independently

smoe:
Another very old patch with Debian, again one that originated from invoking cppcheck on the source tree.

The exact cppcheck warning I don't recall. I would have expected the iterator to know about how to properly terminate itself before going through the next iteration. The indentiation of the code is disturbed by the extra blocks to minimize the size of the patch. Alternatively, "it" could be renamed or the indentiation be updated.

Personally, I do not like this patch too much since the exact need to have to declarations of the iterator is not immediately apparent. A comment would be nice at least. But I very much like not to declare an iterator at the beginning of a function that is only needed in a branch of the code. So, I expect a bit of a discussion about this patch and confidently expect this code to improve over it.

### DIFF
--- a/client/rrsim_test.cpp
+++ b/client/rrsim_test.cpp
@@ -184,7 +184,6 @@ bool CLIENT_STATE::rr_simulation() {
     vector<RESULT*> active;
     unsigned int i;
     double x;
-    vector<RESULT*>::iterator it;
     bool rval = false;
 
     if (log_flags.rr_simulation) {
@@ -287,7 +286,9 @@ bool CLIENT_STATE::rr_simulation() {
         int last_active_size = active.size();
         int last_proj_active_size = pbest->active.size();
 
-        // remove *rpbest from active set,
+        {
+        vector<RESULT*>::iterator it;
+	// remove *rpbest from active set,
         // and adjust CPU time left for other results
         //
         it = active.begin();
@@ -301,7 +302,10 @@ bool CLIENT_STATE::rr_simulation() {
                 ++it;
             }
         }
+	}
 
+	{
+        vector<RESULT*>::iterator it;
         // remove *rpbest from its project's active set
         //
         it = pbest->active.begin();
@@ -313,6 +317,7 @@ bool CLIENT_STATE::rr_simulation() {
                 ++it;
             }
         }
+	}
 
         // If project has more results, add one to active set.
         //


client/rrsim_test.cpp
```diff
@@ -287,7 +286,9 @@ bool CLIENT_STATE::rr_simulation() {
         int last_active_size = active.size();
         int last_proj_active_size = pbest->active.size();
 
-        // remove *rpbest from active set,
+        {
```
AenBleidd: Extra scopes are not necessary here.
Either reuse `it` variable or refactor this code to move these two cycles into separate functions.
Also please adjust aligning.

serval2412:
Running cppcheck git updated today with enable=all gives:
Checking client/rrsim_test.cpp ...
client/rrsim_test.cpp:42:5: warning: Member variable 'PROJECT::cpu_shortfall' is not initialized in the constructor. [uninitMemberVar]
    PROJECT(char* n, double rs) {
    ^
client/rrsim_test.cpp:42:5: warning: Member variable 'PROJECT::rrsim_proc_rate' is not initialized in the constructor. [uninitMemberVar]
    PROJECT(char* n, double rs) {
    ^
client/rrsim_test.cpp:42:5: warning: Member variable 'PROJECT::rr_sim_deadlines_missed' is not initialized in the constructor. [uninitMemberVar]
    PROJECT(char* n, double rs) {
    ^
client/rrsim_test.cpp:65:5: warning: Member variable 'RESULT::rrsim_finish_delay' is not initialized in the constructor. [uninitMemberVar]
    RESULT(PROJECT* p, char* n, double e, double rd) {
    ^
client/rrsim_test.cpp:65:5: warning: Member variable 'RESULT::rrsim_cpu_left' is not initialized in the constructor. [uninitMemberVar]
    RESULT(PROJECT* p, char* n, double e, double rd) {
    ^
client/rrsim_test.cpp:65:5: warning: Member variable 'RESULT::rr_sim_misses_deadline' is not initialized in the constructor. [uninitMemberVar]
    RESULT(PROJECT* p, char* n, double e, double rd) {
    ^
client/rrsim_test.cpp:65:5: warning: Member variable 'RESULT::last_rr_sim_missed_deadline' is not initialized in the constructor. [uninitMemberVar]
    RESULT(PROJECT* p, char* n, double e, double rd) {
    ^
nofile:0:0: information: Cppcheck cannot find all the include files (use --check-config for details) [missingIncludeSystem]
So nothing about iterator pb.

I don't like too the fact that vars are at beginning of the function whereas it's a c++ file.
About the fact to declare this var only when using, I agree but since there's the big "while (active.size())" (line 249), I wonder if it would have some runtime cost.

client/rrsim_test.cpp
smoe: If you do not see anything you want to change in the current BOINC tree then please this PR.

smoe: "close" was missing.

AenBleidd: My comment is still valid

PR #3280, opened 2019-09-10
cppcheck: React to failing memory reallocation

smoe:
This is another one of the concerns that cppcheck raised. If the the memory that "we" own is not extendable, then this is likely an out of memory situation to which BOINC should possibly react nicely as nicely as it can. That patch meant to prevent a memory leak in low-memory situations. It does not look nice, admittedly. Maybe someone else has a constructive idea about it.

### DIFF
--- a/lib/unix_util.cpp
+++ b/lib/unix_util.cpp
@@ -80,7 +80,13 @@ int setenv(const char *name, const char *value, int overwrite) {
             }
             if (i!=envstrings.end()) {
                 // we allocated this string.  Reallocate it.
+		char *b=buf;
                 buf=(char *)realloc(buf,strlen(name)+strlen(value)+2);
+		if (!buf) {
+		    free(b);
+		    errno=ENOMEM;
+		    return -1;
+		}
                 *i=buf;
             } else {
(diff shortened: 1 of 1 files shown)


lib/unix_util.cpp
```diff
@@ -80,7 +80,13 @@ int setenv(const char *name, const char *value, int overwrite) {
             }
             if (i!=envstrings.end()) {
                 // we allocated this string.  Reallocate it.
+		char *b=buf;
                 buf=(char *)realloc(buf,strlen(name)+strlen(value)+2);
+		if (!buf) {
+		    free(b);
+		    errno=ENOMEM;
+		    return -1;
```
brevilo: While I agree with the overall idea/implementation, these two lines look redundant to me (lines 98-101).

brevilo: Unless one favors early returns over redundant code, which I don't 😏

smoe: I fail to see anything redundant. An sprintf to NULL is an uncontrolled abort. And that should be avoided.

AenBleidd:
Closing this as invalid since the buffer is already checked later

PR #3281, opened 2019-09-10
Allow compilation of BOINC when there is no MAX_PATH(length)

smoe:
Linux has a maximal length of paths. Other platforms, notably HURD,
are unlimited in the length of their paths and filenames.
This patch (and there have been others before) allow to compile
and use BOINC on such technically advanced(?) OSs by simply imposing
the same maximal path length as for Linux.

### DIFF
--- a/clientgui/BOINCGUIApp.cpp
+++ b/clientgui/BOINCGUIApp.cpp
@@ -49,6 +49,9 @@
 #include "procinfo.h"
 #include "sg_BoincSimpleFrame.h"
 
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
 
 bool s_bSkipExitConfirmation = false;
 


davidpanderson:
I'm not sure why PATH_MAX is used in that one place.
It should be MAXPATHLEN, which is used everywhere else.

smoe:
I just did a "grep -r" on PATH_MAX. It appears frequently throughout the source tree with various meanings - like
```
./zip/unzip/unzpriv.h:#  define PATH_MAX      36
```
Is it a windows thing, possibly? This seems suggested by
```
./zip/zip/win32/osdep.h:#  include <stdio.h>    /* PATH_MAX is defined here */
```
There is also MAX_PATH very often in the code.  And so is MAX_PATH_SIZE. No exact idea if we gain much by cleaning this up since the individual files seem pretty much consistent in themselves. It is more than what was intended by this pull request :-)

mcandre:
Good changes. I would also recommend generating MAX_PATH like this, to account for more different kinds of OS's and environments:

https://github.com/mcandre/tonixxx/blob/master/examples/fewer/lib/fewer.h#L20-L35

AenBleidd:
`zip` is a third-party dependency, and is panned for removal from the BOINC codebase.
Closing this as not needed.

PR #3282, opened 2019-09-10
adds error message in case of dir open failure

smoe:
Trivial thing. Should also help to get better reports by users when they experience something exceptional.

### DIFF
--- a/lib/filesys.cpp
+++ b/lib/filesys.cpp
@@ -154,7 +154,12 @@ DIRREF dir_open(const char* p) {
     dirp->handle = INVALID_HANDLE_VALUE;
 #else
     dirp = opendir(p);
-    if (!dirp) return NULL;
+    if (!dirp) {
+        char b[MAXPATHLEN+1];
+        boinc_getcwd(b);
+        fprintf(stderr,"dir_open: Could not open directory '%s' from '%s'.\n",p,b);
+        return NULL;
+    }
 #endif
     return dirp;
 }
@@ -682,7 +687,7 @@ static int boinc_rename_aux(const char* old, const char* newf) {
     //
     int retval = rename(old, newf);
     if (retval) {
-        char buf[MAXPATHLEN+MAXPATHLEN];
+        char buf[MAXPATHLEN+MAXPATHLEN+1+7];
         sprintf(buf, "mv \"%s\" \"%s\"", old, newf);
 #ifdef __APPLE__
         // system() is deprecated in Mac OS 10.10.


lib/filesys.cpp
```diff
@@ -682,7 +687,7 @@ static int boinc_rename_aux(const char* old, const char* newf) {
     //
     int retval = rename(old, newf);
     if (retval) {
-        char buf[MAXPATHLEN+MAXPATHLEN];
+        char buf[MAXPATHLEN+MAXPATHLEN+1+7];
```
Rytiss: Why is this `+1+7`? "mv", 2 spaces and 4 quotes are 8 bytes, shouldn't this be `+1+8`?

davidpanderson:
Current directory is always the BOINC data directory

davidpanderson:
The convention in BOINC is that low-level functions just return error codes;
higher-level functions can print error messages if they want.
Also, on Win stderr doesn't go to the event log so most users don't see it.
Error messages should be output using msg_printf()

lib/filesys.cpp
smoe: Both our maths is a bit beyond me, I must admit. There was a time in my life when I kept adding 17 without counting - because [17 being the first/smallest/least random number](https://www.quora.com/Why-is-17-considered-the-least-random-number-in-the-world-of-computer-programming). So, we should add the possibility that it was just a typo.

Good ole memories aside, yes, 1+2+6 sounds about right.

smoe:
I am not sure that I want the message to appear on the event log. But maybe that would be a good thing, indeed.

smoe:
I had a look. The higher-level function cannot distinguish between a low memory and the dir opening error. As such I would very much prefer to see an error reported in the opening case. Such low in memory, there may be little point to try to write anything anywhere, but for the failed open, knowing this would save so much time both on the user's side and on the developers' sides, I am up for it. BOINC should stop either way if that NULL is returned, so, for such fatal messages I suggest to make an exception and write the error message for a lower-level function.

davidpanderson:
Error messages should be written by higher levels, not here.

RichardHaselgrove:
An error message should never mislead. If you can't be certain of the reason for the failure, say so.

smoe:
@RichardHaselgrove 's comment refers to the error message that would be given on the higher level as preferred by @davidpanderson - but for that, the higher-level routine would need to know about the possible causes of failure. The lower level routine can be exact. I would just not attempt to do anything that requests more memory if out of memory.

AenBleidd:
Closing this as not needed.

Issue #1175, opened 2015-02-04
md5_file: Too many open files


romw:
**Reported by smoe on 8 Jan 42640604 02:59 UTC**
I found the boinc-client to have stopped for no apparent reason. It was working only with a local self-built SETI client. I had seen this once a long time before, though, back then with the WCG.

From stderrdae.txt:

No protocol specified[protocol specified[[BR]([BR]]No)]No protocol specified[protocol specified[[BR]([BR]]No)]...

dir_open: Could not open directory 'slots/0'.[Could not open directory 'slots/18'.[[BR]([BR]]dir_open:)]dir_open: Could not open directory 'slots/17'.[Could not open directory 'slots/12'.[[BR]([BR]]dir_open:)]dir_open: Could not open directory 'slots/7'.[Could not open directory 'slots/4'.[[BR]([BR]]dir_open:)]dir_open: Could not open directory 'slots/22'.[Could not open directory 'slots/19'.[[BR]([BR]]dir_open:)]dir_open: Could not open directory 'slots/9'.[Could not open directory 'slots/16'.[[BR]([BR]]dir_open:)]dir_open: Could not open directory 'slots/14'.[Could not open directory 'slots/20'.[[BR]([BR]]dir_open:)]dir_open: Could not open directory 'slots/8'.[Could not open directory 'slots/3'.[[BR]([BR]]dir_open:)]dir_open: Could not open directory 'slots/23'.[Could not open directory 'slots/11'.[[BR]([BR]]dir_open:)]...

[...]

PR #3285, opened 2019-09-10
Closing files after opening them.

smoe:
This patch is with Debian for a really long time, likely to have its origins also in the cppcheck and/or clang runs. Just like PR https://github.com/BOINC/boinc/pull/3284, this also addresses the reported leaking of file descriptors (https://github.com/BOINC/boinc/issues/1175).

### DIFF
--- a/lib/crypt_prog.cpp
+++ b/lib/crypt_prog.cpp
@@ -168,8 +168,8 @@ int main(int argc, char** argv) {
         if (!fpub) die("fopen");
         print_key_hex(fpriv, (KEY*)&private_key, sizeof(private_key));
         print_key_hex(fpub, (KEY*)&public_key, sizeof(public_key));
-	fclose(fpriv);
-	fclose(fpub);
+        fclose(fpriv);
+        fclose(fpub);
 
     } else if (!strcmp(argv[1], "-sign")) {
         if (argc < 4) {
@@ -196,7 +196,7 @@ int main(int argc, char** argv) {
         if (retval) die("scan_key_hex\n");
         generate_signature(argv[2], cbuf, private_key);
         puts(cbuf);
-	fclose(fpriv);
+        fclose(fpriv);
     } else if (!strcmp(argv[1], "-verify")) {
         if (argc < 5) {
             usage();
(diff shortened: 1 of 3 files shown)


davidpanderson:
Closing files before exiting a program isn't needed.

The semantics of MFILE are that if you open() you have to close().
No need to do it in the destructor
(this is used only in cs_statefile.cpp)

smoe:
> Closing files before exiting a program isn't needed.

Well, the computer does not break, but BOINC is running for a long time and there has been that report on a file descriptor leakage for other installations. If the file does not need to be reopened a few seconds later - just close it and free the resources.

> The semantics of MFILE are that if you open() you have to close().
> No need to do it in the destructor
> (this is used only in cs_statefile.cpp)

I disagree. A destructor encountering the left-open file should close it. If the semantics of that class suggest otherwise, it should still close it and send a request to stderr to get this fixed. If there is a forwarding of responsibilities to close that file by some copy operator of some sort, then this needs to be catered for.

Germano0:
In my opinion we should also check the return value of `fclose` and `close` functions

brevilo:
I agree with the PRs motivation and intentions but haven't yet verified the logic/code paths. What needs fixing either way is the indentation that currently a bit off.

smoe:
There is more to the PR than the closing of files. @davidpanderson is however right that the file-closing concerns are indeed driven by short-lived applications. Feel free to cut that. However, from my POV I do not care if it is a long-running process or a short-running one. Did not even look - otherwise I may not even have submitted it to avoid discussions. I always close what is no longer needed. And free what is no longer needed. ASAP, "P" as in "there is no semantics associated with 'having a lock on a file'". That way the code is also allowed to move into other parts of BOINC. There is no unnecessary context required to understand that part of the code. And I admittely sense some satisfaction over that elegance.

ChristianBeer:
If you fix the indentation this is ready to be merged. My POV is that closing a file even before ending a short-lived program shows that the developer intentionally thought about where to close the file. The same goes for MFILE. What if a developer forgot about the semantics and didn't close. The destructor would take care of that.

smoe:
The indenting should be fine now - typical tabs vs blanks. Isn't there possibly a way to have the auto-indented?
Anyway - let me try to rebase this branch prior to merging anything because of the small commits - seems like I should have done that locally.

PR #3303, opened 2019-09-24
client: on startup, check for active task finish files

davidpanderson:
Should fix problem where jobs finish while client is shutting down, as described in #3300

### DIFF
--- a/client/app.h
+++ b/client/app.h
@@ -276,7 +276,7 @@ struct ACTIVE_TASK {
     void get_graphics_msg();
     double est_dur();
     int read_stderr_file();
-    bool finish_file_present();
+    bool finish_file_present(int&);
     bool temporary_exit_file_present(double&, char*, bool&);
     void init_app_init_data(APP_INIT_DATA&);
     int write_app_init_file(APP_INIT_DATA&);
@@ -328,6 +328,7 @@ public:
     void free_mem();
     bool slot_taken(int);
     void get_memory_usage();
+    void check_for_finished_jobs();
 
     void process_control_poll();
     void request_reread_prefs(PROJECT*);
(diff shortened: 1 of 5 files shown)


AenBleidd:
Maybe it's better to change function to return error code instead of adding it as an output parameter?

davidpanderson:
The return value is whether a finish file is present

AenBleidd:
Reasonable. Agreed

TheAspens:
@AenBleidd - were you going to continue this review and merge?  If not let me know and I can do it.

AenBleidd:
@TheAspens, generally I'm ok with the code. But I didn't test it. So please merge it if you believe it's fine and no additional verification needed (or done by you or someone else)

RichardHaselgrove:
I was looking at this one too. It puts us into a bit of a bind with the 'feature freeze' for 7.16

It's part of (an extension of) the #3017 / #3019 pairing, which was merged before the release branch was split (back in March, even before the first abortive split). Like @AenBleidd, the code looks OK, but it's a hard problem to generate a conclusive test for in live running. And we really don't want to delay 7.16 any longer, given the slow rate of testing responses apart from the platform-specific issues with Mac and Android. I'll leave it to you: we've lived with the problem for many years, we can leave it on the slate for the next release if necessary.

TheAspens:
I working on reproducing it with a build from master right now.  Once I can reliably reproduce, then I will build from this branch and repeat the test.  I don't think it will be too hard (famous last words).

TheAspens:
I was able to test by downloading a "uppercase" workunit from a test project and then exiting the client in the middle of the task.  I then completed the task standalone and started the client again.  The version in master restarted the task from scratch (since I don't think uppercase has any checkpointing) whereas the code in this branch saw the finish file and immediately uploaded and reported the task.

This looks good to me.  Thanks @davidpanderson  for the fix.  I'm merging.

PR #3307, opened 2019-09-25
Wrapper: Monitor status of child task process even if parent exits

tristanolive:
**Description of the Change**
On Windows, CreateProcess() is used to launch tasks, but this on its own does not handle child processes; if the parent task process exits, the workunit will be terminated. If <wait_for_children> is set in the job file, attach the task process to a job object instead, which can then be monitored to determine when all child processes are finished.

**Alternate Designs**

**Release Notes**

Add <wait_for_children> option for tasks in job.xml

### DIFF
--- a/lib/proc_control.cpp
+++ b/lib/proc_control.cpp
@@ -277,6 +277,33 @@ void suspend_or_resume_process(int pid, bool resume) {
 #endif
 }
 
+#ifdef _WIN32
+
+// Handling of job objects
+//
+
+void get_job_object_processes(HANDLE job_handle, vector<int>& pids) {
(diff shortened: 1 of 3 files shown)


AenBleidd:
Closing and reopening to rerun CI builds

tristanolive:
Closing and reopening again to get CI builds to run.

davidpanderson:
Why is this needed?  Normally programs that create children wait for them to finish.

ChristianBeer:
@tristanolive @Rytiss This PR has a conflict with current master. Could you please check if the changes are still needed and adjust the code accordingly?

Rytiss:
I believe the changes are still needed. @tristanolive - can you mod the code so that it does not conflict?

PR #3313, opened 2019-10-01
client: remove HTTP header from HTTP GUI RPC request before parsing

davidpanderson:
Fixes a problem reported by Germano Massullo, where issuing a GUI RPC using curl didn't work.


Germano0:
Patch successfully tested. It fixes https://groups.google.com/a/ssl.berkeley.edu/forum/#!topic/boinc_dev/Ps270gU9xJc

PR #3314, opened 2019-10-01
client: fix bug that caused jobs to run after client exit.

davidpanderson:
Take out #2871.
That change was based on the incorrect assumption
that a suspended process doesn't handle quit messages.

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**


RichardHaselgrove:
Merging to re-synchronise master with 7.16 release branch after testing.

PR #3315, opened 2019-10-01
remote job submission: add an API function for setting RPC timeout

davidpanderson:
boinc_set_timeout($x): set timeout for job submission RPCs to x seconds


Rytiss:
df8f370 is not in the scope of this PR. However, it will become a no-op if #3314 is merged first.

davidpanderson:
Oops!  how did that happen.

AenBleidd:
Clean commit was made via #3379

PR #3319, opened 2019-10-04
client: don't ignore new poll intervals for RSS feeds.

davidpanderson:
When projects changed their poll intervals for RSS feeds (such as notices) these changes were being ignored by the client.


TheAspens:
Looks good.  Thanks for the fix.

PR #3322, opened 2019-10-08
rpc query job status

hzclarksm:
Fixes #

**Description of the Change**

Combine the rpc:query_batches and rpc:query_batch procedures into a single rpc:query_batches_status procedure.  The combination eliminates redundant updating of workunit status in the database.  When querying on the order of 100K workunits the time to complete the query was reduced by  factor of 10.

**Alternate Designs**

**Release Notes**

Added streamlined procedure for querying job status through the submit_rpc.


hzclarksm:
Only the third commit is relevant to the request.

davidpanderson:
show_job_status() looks like it does more or less the same thing as the existing show_job_details().  Why a new function?

hzclarksm:
show_job_details() was the starting point for show_job_status().
show_job_status() has the advantage of scanning many less workunits at the expense of reporting less detail about an error.  Completed (done or error) batches are not scanned at all and scanning stops when a single inprogress workunit is found.

The bigger savings no doubt result from eliminating a second get_batch_params() in query_batch.

davidpanderson:
Please reread my comment.  The two functions each process one WU and enumerate its results.  They do more or less the same thing.  Only one of them should exist.

hzclarksm:
Where I used the term workunits  I should have used results.   show_job_status() will scan fewer results.
show_job_status() in large part is show_job_details() with get_cpu_time = NO.  I think that there are efficiency gains in show_job_status() that could be applied to show_job_details().

html/user/submit_rpc_handler.php
```diff
@@ -644,6 +644,32 @@ function show_job_details($wu) {
...
 
+function show_job_status($wu) {
+    if       ($wu->error_mask) {
+        echo "            <status>error</status>\n";
+    } elseif ($wu->canonical_resultid) {
+        echo "            <status>done</status>\n";
+    } else {
+        $results = BoincResult::enum("server_state","workunitid=$wu->id");
+        $in_progress = 0;
+        foreach ($results as $r) {
```
Rytiss: There is no need to iterate here. This shows the code really is copy-paste from the function above.

This could be done in SQL, "workunitid=$wu->id AND server_state=".RESULT_SERVER_STATE_IN_PROGRESS - and then check if the array size is > 0.

ChristianBeer:
@hzclarksm This PR seems to be stale. If you still want this feature added please review the comments and create a new PR.

PR #3329, opened 2019-10-11
installer: Add the upgrade code used by the WCG 7.14.2 version of the…

TheAspens:
WCG accidentally changed the upgrade code when they released their 7.14.2 version (WCG is releasing a 7.14.3 version with a corrected upgrade code).  

This pull request adds that bad upgrade code to the list of related upgrade codes so that the WCG 7.14.2 version can be uninstalled by BOINC versions of the installer and provide a clean upgrade experience.

### DIFF
--- a/win_build/installerv2/BOINC.ism
+++ b/win_build/installerv2/BOINC.ism
@@ -4568,7 +4568,7 @@ VwBlAGIAAQBXAEUAQgB4ADgANgA=
 		<row><td>RebootYesNo</td><td>Yes</td><td/></row>
 		<row><td>ReinstallModeText</td><td>omus</td><td/></row>
 		<row><td>RestartManagerOption</td><td>CloseRestart</td><td/></row>
-		<row><td>SecureCustomProperties</td><td>ACTPROP_BC961760_07FB_4754_A277_8D6943B569CC;ACTPROP_E913E54D_5080_42EC_A312_B21948BA1C02;INSTALLDIR;SUPPORTDIR;ENABLEPROTECTEDAPPLICATIONEXECUTION3;LAUNCHPROGRAM;ISACTIONPROP1</td><td/></row>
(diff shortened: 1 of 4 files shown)


tristanolive:
I understand this is a friendly thing to do for any user who would upgrade in the future from WCG 7.14.2 to a higher version of BOINC, and in my opinion, it only adds a small bit of extra complication to what is already a very complicated set of files. I don't like that it is added to all general BOINC installers, in principle, as it feels messy to me, but I can appreciate the value in it.

I don't mind merging, but don't think this is a solution we would want to follow in general when installer upgrade codes are changed. Any @BOINC/committer have a thought one way or the other?

TheAspens:
The error is that we accidentally changed the upgrade code on our 7.14.2 version (which the public has already installed).  The WCG 7.14.3 version which we are working on releasing (which will include the security patch) will go back to using the same upgrade code as BOINC so that this issue doesn't happen again.  The core issue is that the WCG 7.14.2 client should never have had a changed upgrade code.  This is a mistake we made and have documented test cases to make sure that we don't do this again in the future.

This change allows the BOINC installer to see that WCG 7.14.2 is the same application as BOINC so that it will invoke the uninstaller for the WCG 7.14.2 client before installing the BOINC version (i.e. so that it does the same thing with the WCG 7.14.2 that it does with all other versions).

I am not aware of any other way to resolve this issue.

tristanolive:
I don't think there is another way to resolve this issue in code. The alternative is for users of WCG 7.14.2 who want to install BOINC 7.16+ to visit support forums (or better yet a FAQ) to figure out why their upgrade is failing.

So I fully understand the reason for doing this in your case and would like to do the same thing for GridRepublic in that situation! I just question:

1. whether it taints the core BOINC code base; and
2. whether we care as a community.

My inclination is that yes, it does taint the code base, but that we should be supportive of WCG and allow it. It's tricky, because if there were a dozen little "pet project" type of account managers that were doing this, we might not want to allow it. So this is why I think it would be good to hear from others to get some consensus on it.

(If no one responds with concerns by tomorrow, I'll push it through).

Issue #3236, opened 2019-07-28
[Android][AM] No error message on wrong credentials


AenBleidd:
When attaching to AccountManager from Android BOINC Manager with wrong credentials no error message shown to user. It could be found in log but for the user it's not so obvious, it looks like that Android Manager just initializes the Account Manager.

**Solution**

Error message should be shown to user that credentials are wrong.

PR #3336, opened 2019-10-16
[Android] Credential check for account manager login

okan35:
Fixes #[3236](https://github.com/BOINC/boinc/issues/3236)

**Description of the Change**
Added a message if user does not exist or password in account manager login screen.

**Alternate Designs**
No alternatives for now.

**Release Notes**
With this change user will know if he/she entered wrong password or username before this was not the case on login screen for account manager.


android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
```diff
@@ -1180,7 +1180,12 @@ public int addAcctMgrErrorNum(String url, String userName, String pwd)
                 throws RemoteException {
             AcctMgrRPCReply acctMgr = clientInterface.addAcctMgr(url, userName, pwd);
             if (acctMgr != null) {
-                return acctMgr.error_num;
+                if(acctMgr.messages.get(0).equals(getString(R.string.rpcreason_user_pass_problem))) {
```
AenBleidd: Here will be an issue if user has localized (non-English) Android client. This statement is valid for English Android client only.
To show to the user localized error message is a good idea, but it requires changes on server side. As far as we don't know what server will sends us as a message in response and there is no error code returned with message, it is safer to show to the user just exact the same text that contains in response without any comparisons.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.java
```diff
@@ -1180,7 +1180,12 @@ public int addAcctMgrErrorNum(String url, String userName, String pwd)
                 throws RemoteException {
             AcctMgrRPCReply acctMgr = clientInterface.addAcctMgr(url, userName, pwd);
             if (acctMgr != null) {
-                return acctMgr.error_num;
+                if(acctMgr.messages.get(0).equals("User not found or password wrong.")) {
```
AenBleidd: Unfortunately, this will not work either: as I said previously, we don't know what message will be sent from the server.
According to my observations, there is next situation on wrong credentials (random user/email + random password).
- BAM! - No error code sent
BAM
- GRCPool - No error code sent
GRCPool
- Grid Republic - Error code sent
GridRepublic
- Science United - Error code sent
ScienceUnited

So having this information next algorythm should be used:
- if there is an error code different from 0 - Show error message from response
- if there is an error code 0 and message from response is not empty - Show error message from response
- if there is an error code 0 and message from response is empty - Consider that everything went fine and proceed

PR #3345, opened 2019-10-22
Fix boincsim build broken after 401dccb793635ac4b42c6d069d399dadd07b…

AenBleidd:
…a459


davidpanderson:
Thanks; I fixed it in slightly different way in #3349.

PR #3349, opened 2019-10-23
client simulator: fix build

davidpanderson:
Also fix compile warning


AenBleidd:
Yeah, this solution is better.
Thanks!

PR #3351, opened 2019-10-25
[Manager] Fix connection to second local client instance

AenBleidd:

### DIFF
--- a/clientgui/MainDocument.cpp
+++ b/clientgui/MainDocument.cpp
@@ -583,6 +583,7 @@ int CMainDocument::OnPoll() {
     int iRetVal = 0;
     wxString hostName = wxGetApp().GetClientHostNameArg();
     wxString password = wxGetApp().GetClientPasswordArg();
+    bool isHostnamePasswordSet = wxGetApp().IsHostnamePasswordSet();
     int portNum = wxGetApp().GetClientRPCPortArg();
 
     wxASSERT(wxDynamicCast(m_pClientManager, CBOINCClientManager));
@@ -592,7 +593,7 @@ int CMainDocument::OnPoll() {
         m_bClientStartCheckCompleted = true;
 
         if (IsComputerNameLocal(hostName)) {
-            if (wxGetApp().IsAnotherInstanceRunning()) {
+            if (wxGetApp().IsAnotherInstanceRunning() && !isHostnamePasswordSet) {
                 if (!pFrame->SelectComputer(hostName, portNum, password, true)) {
                     s_bSkipExitConfirmation = true;
                     wxCommandEvent event;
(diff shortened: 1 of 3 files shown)


RichardHaselgrove:
I've tested that the presenting problem is fixed. I need to look back into why we made the previous change, and check that solution is still valid - then I'll post a formal review.

RichardHaselgrove:
Actually, there is one side-effect. I also use (under Windows) the Manager file menu function 'New BOINC Manager window...' (which is an alternative route into using multiple managers). In this use case, the intention is to use the new (additional) window to monitor a different, remote, computer.

Previously the new window opened blank, with the 'Select Computer' dialog showing. Now, it opens with the local client displayed, and I have to make a second mouse movement to 'Select Computer'. Ooh, the hardship! Better that both use-cases still work, with a little extra effort, than that one can't be used. But if....

AenBleidd:
Thank you. I'll fix this too

AenBleidd:
I made a 'blind' fix but didn't verify it yet. Plan to verify it on Monday.

AenBleidd:
@RichardHaselgrove, should be fixed now. Please, verify it again.

Germano0:
Hi Vitalii, I think it would be great to have a full description of the problem that this patch is fixing, so that the testing process can be more accurate.
Thank you

RichardHaselgrove:
@Germano0: It arises from the email I sent to the boinc_alpha email list at 21 Oct at 20:12 (UTC+1). Details are there.

clientgui/MainDocument.cpp
```diff
@@ -583,6 +583,7 @@ int CMainDocument::OnPoll() {
     int iRetVal = 0;
     wxString hostName = wxGetApp().GetClientHostNameArg();
     wxString password = wxGetApp().GetClientPasswordArg();
+	bool isHostnamePasswordSet = wxGetApp().IsHostnamePasswordSet();
```
TheAspens: @AenBleidd - can you fix the indentation here?

AenBleidd:
Damn, I made a force-push of incorrect commit. Will fix it later with the indentation fix requested above

AenBleidd:
Damn, I made a force-push of incorrect commit. Will fix it later with the indentation fix requested above

AenBleidd:
Ok, now it's fine.
FYI: https://objectpartners.com/2014/02/11/recovering-a-commit-from-githubs-reflog/

AenBleidd:
No, in [this](https://groups.google.com/a/ssl.berkeley.edu/forum/#!topic/boinc_alpha/aH_izVE7kRg)

RichardHaselgrove:
> Is it this discussion?

@AenBleidd has the correct thread.

TheAspens:
Based on the discussion on the contributor call this morning and the code and testing it looks good.  @AenBleidd thanks for the fix!

PR #3367, opened 2019-11-07
Make the electron-based GUI work again.

davidpanderson:
RPC requests need to be enclosed in <boinc_gui_rpc_request>.
I'm not sure how it ever worked.

Also: improve GUI RPC error reporting in the client.


AenBleidd:
Was not able to verify this.
Got next error:
image
Old version (from master) doesn't work either

davidpanderson:
This works for me, whether the client is built from this branch or from master.  Make sure you copy window.js from boinc/samples/electron_gui to electron/simple-samples/activity-monitor.

AenBleidd:
Yeah, looks like Electron GUI can't deal with 7.14.2 release version. Tested on latest master branch. Looks fine for me.

PR #3368, opened 2019-11-08
Mac: Fix computation of non_BOINC CPU usage

CharlieFenton:
Fixes a long-standing bug which caused non-BOINC CPU usage to always be computed as zero on Macs.

### DIFF
--- a/lib/procinfo_mac.cpp
+++ b/lib/procinfo_mac.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2018 University of California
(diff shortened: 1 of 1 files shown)


Germano0:
Are those
https://github.com/BOINC/boinc/blob/master/lib/mac/mac_branding.cpp#L46-L52
the names of official BOINC client and custom BOINC clients built by some BOINC projects?

CharlieFenton:
> Are those
https://github.com/BOINC/boinc/blob/master/lib/mac/mac_branding.cpp#L46-L52
the names of official BOINC client and custom BOINC clients built by some BOINC projects?

Yes, that is correct.

PR #3369, opened 2019-11-08
mac: Fixes for screensaver under OS 10.15 Catalina

CharlieFenton:
Under OS 10.15 Catalina:
Problems:
* Screensavers can't launch _setuid_ / _setgid_ executables like `gfx_switcher` or `ps`
 * Screensavers can't launch executables downloaded directly from the Internet unless they have first been vetted by the user via _GateKeeper_, but there is no provision for _GateKeeper_ to offer users that option for bare executables like project graphics apps.

Solution:
 * Instead of launching graphics apps directly via `gfx_switcher`, the BOINC screensaver coordinator writes a file _BOINCSSHelper.txt_ containing the information about what to launch. The file is detected by a LaunchAgent which then launches `gfx_switcher` for us. `gfx_switcher` then deletes the _BOINCSSHelper.txt_ file, and writes a file _BOINCGfxPid.txt_ containing the graphics app's process ID (PID). The BOINC screensaver coordinator waits for the _BOINCGfxPid.txt_ file to exist and reads the PID from it. When the graphics app exits, `gfx_switcher` deletes the _BOINCGfxPid.txt_ file, signaling the BOINC screensaver coordinator that the graphics app has exited. This is rather convoluted, but it works well with no noticeable degradation of performance.
 * Instead of killing graphics apps via `gfx_switcher`, the BOINC screensaver coordinator sends an RPC to the client asking the client to kill them via `switcher`. (Since the graphics app is running as user _boinc_project_, only user _boinc_project_ can send a kill signal to it.) Since `switcher` (not `gfx_switcher`) is called from the client rather than from the screensaver, there is no restriction on launching _setuid_ / _setgid_ apps like `switcher`. 
 * Note that, since this RPC requires the BOINC screensaver coordinator to read the _gui_rpc_auth.cfg_ file, only users authorized to manage BOINC (via BOINC Manager) can run the BOINC screensaver to display graphics apps. (The installer added those users to groups _boinc_master_ and _boinc_project_.)
 * Though we have confirmed these changes work on OS 10.13 High Sierra, we don't use them there (yet). To do: test if we can use these new methods on all our supported versions of OS X (OS < 10.13) to simplify this code. Since I don't have Macs which can run older versions of OS X, I have not been able to test this.

Problem:
 * Apple's ScreenSaverEngine doesn't always call `stopAnimation` before exiting

Solution:
 * BOINC screensaver coordinator launches a new helper app `gfx_cleanup` via `popen()`, communicating via a pipe. When it launches a graphics app it passes that app's process ID (PID) to `gfx_cleanup`. When the graphics app exits, the BOINC screensaver coordinator sends a zero PID to `gfx_cleanup`.  The `gfx_cleanup` helper app detects when the BOINC screensaver coordinator itself exits; if the most recent PID was non-zero it sends an RPC to the client asking the client to kill the graphics app via `switcher`. 
 * In OS 10.13 High Sierra and later, we must use mach-o communication and `IOSurface` to share the graphics app's GPU buffer with our screensaver, so the screensaver can display the graphics. Since the graphics app does not draw directly to the screen, it's own window is all white (blank.) To avoid an annoying momentary white flash when the screensaver quits before the `gfx_cleanup` helper app kills it, `gfx_cleanup` covers the screen temporarily with an all-black window.

Problem:
 * For compatibility with OS 10.13 High Sierra, we required all project graphics apps to be updated to use mach-o communication and `IOSurface`, by linking to version 7.9.2 or later of _libboinc_graphics2.a_. As a temporary transitional measure, we used the `CGWindowList` method to display any project graphics apps which had not yet been updated. This no longer works on OS 10.15, probably because `CGWindowListCopyWindowInfo` and `CGWindowListCreateImage` can no longer copy windows between user _boinc_project_ and the user running the screensaver, due to security concerns with allowing anyone to view another user's windows. (The `IOSurface` method is safer because an app must have specific code to allow sharing its windows with the other app.)

Solution:
 * We can no longer support project graphics apps which have not been updated by linking to version 7.9.2 or later of _libboinc_graphics2.a_. This library has been available since February 2018. The `CGWindowList` method resulted in reduced quality due to a low frame rate and was never intended to be a long-term solution. All projects should have re-linked their graphics app by now (though I suspect that not all have done so.)  

Problem:
 * Apple's ScreenSaverEngine always passes true for `isPreview` argument

 Solution:
* Under OS 10.15 or later, assume it is not a preview if the window width or height is greater than 500.

Problem:
* Under OS 10.15, OpenGL apps use doubled window dimensions on Retina displays (2 pixels per point), but earlier versions of OS X do not.

Solution:
* Under OS 10.15 or later, adjust the drawing routines based on the results of the `[NSScreen backingScaleFactor]` which gives the number of pixels per point on the screen. This works for both Retina and conventional monitors. To do: check that later versions of OS X still require these adjustments; if not, then modify the code to do them only when appropriate.

Problem:
* Screensaver output files _stdoutscr.txt_ and _stderrscr.txt_ are put in an obscure sandboxed directory

Solution:
* The installer creates a symbolic link "_~/Library/Application Support/BOINC/ScreenSaver logs_" to "_~/Library/Containers/com.apple.ScreenSaver.Engine.legacyScreenSaver/Data/Library/Application Support/BOINC/_" so users can find these files by looking in the directory which held them in older versions of OS X and use that link. (The older directory "_~/Library/Application Support/BOINC/_" still contains other BOINC logs.)

**Release Notes**
Fixes the BOINC screensaver for Mac OS 10.15 Catalina.

### DIFF
--- a/api/graphics2_unix.cpp
+++ b/api/graphics2_unix.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2017 University of California
+// Copyright (C) 2019 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -46,8 +46,10 @@ static int win=0;
 static int checkparentcounter=0;
 
 #ifdef __APPLE__
-
 static bool need_show = false;
+
+// Set to true to draw both directly to window and also to offscreen buffer
+bool debugSharedOffscreenBuffer = false;
 #endif
 
 bool fullscreen;
@@ -119,7 +121,7 @@ static void maybe_render() {
 
     if (throttled_app_render(new_width, new_height, dtime())) {
 #ifdef __APPLE__
-        if (UseSharedOffscreenBuffer()) {
+        if (UseSharedOffscreenBuffer() && !debugSharedOffscreenBuffer) {
             return; // Don't waste cycles drawing to hidden window on screen
         }
 #endif
--- a/api/macglutfix.m
+++ b/api/macglutfix.m
@@ -1,6 +1,6 @@
 // Berkeley Open Infrastructure for Network Computing
 // http://boinc.berkeley.edu
-// Copyright (C) 2017 University of California
+// Copyright (C) 2019 University of California
 //
 // This is free software; you can redistribute it and/or
 // modify it under the terms of the GNU Lesser General Public
@@ -285,7 +285,6 @@ kern_return_t _MGSDisplayFrame(mach_port_t server_port, int32_t frame_index, uin
 	{
 		if(clientPortNames[i])
 		{
-            // print_to_log_file("BOINCSCR: about to call _MGCDisplayFrame  with iosurface_port %d, IOSurfaceGetID %d and frameIndex %d", (int)iosurface_port, IOSurfaceGetID(ioSurfaceBuffers[index]), (int)index);
 			_MGCDisplayFrame(clientPortNames[i], index, iosurface_port);
 		}
 	}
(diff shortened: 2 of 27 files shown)


CharlieFenton:
Problem:
`Xcode 11` apparently builds OpenGL / GLUT applications so that they use doubled window dimensions on Retina displays (2 pixels per point), but earlier versions of `Xcode` do not.

Solution:
Changed code in `MacPassOffscreenBufferToScreenSaver()` in _api/macglutfix.m_ to set width and height of `IOSurfaceBuffers` based on width and height of OpenGL `viewport` instead of width and height of window / NSView.

Mac BOINC graphics apps linked with version 7.9.2 or later of libboinc_graphics2.a and built with Xcode 10 or earlier on OS 10.14.6 or earlier need no further work; they should work on all current and past versions of OS X, including OS 10.15.1 Catalina. But Mac BOINC graphics apps built with Xcode 11 or later and / or OS X 10.15 or later must be linked with the new libboinc_graphics2.a provided by this PR to work properly on all current and past versions of OS X.

PR #3372, opened 2019-11-13
Drupal: Set consistent defaults in flag_abuse module

drshawnkwang:
Part of https://dev.gridrepublic.org/browse/DBOINCP-298

**Description of the Change**
Removed flag default file as it seems to create problems with boinc_standard Features.

**Release Notes**
N/A


Germano0:
> Part of https://dev.gridrepublic.org/browse/DBOINCP-298

The URL is not accessible without an account

AenBleidd:
@Germano0, it's an internal ticket tracking system.
Generally you don't need an access to it

tristanolive:
@Germano0 the ticket system used to be public, but there were some security concerns with JIRA in the past and it was restricted. Since the BOINC-Drupal project is not under active development from many people in the BOINC community, there would be reluctance to opening that ticket system back up to the world; but if you want an account to see the discussion / context, I can set that up for you, no problem.

tristanolive:
As for the PR, thanks for the revised approach, Shawn. I think this should help to ensure everything works as expected when a Drupal site is initialized.

PR #3379, opened 2019-11-19
remote job submission: add an API function for setting RPC timeout

AenBleidd:
boinc_set_timeout($x): set timeout for job submission RPCs to x seconds

### DIFF
--- a/html/inc/submit.inc
+++ b/html/inc/submit.inc
@@ -29,6 +29,8 @@
 //      boinc_query_job(): get details of a job
 //      boinc_retire_batch(): retire a batch; delete output files
 //      boinc_submit_batch(): submit a batch
+//
+//      boinc_set_timeout($x): set RPC timeout to X seconds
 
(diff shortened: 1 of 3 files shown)


AenBleidd:
Just a clean copy of #3315 with no unrelated commit

TheAspens:
@lfield - are you able to review this PR?  I believe that you use the remote submission API and would be in a good position to review.  Let me know - thanks!

lfield:
A question and a comment before I test tomorrow. Is this also needed for the C++ code? I think the Python global should also be rpc_timeout.

davidpanderson:
I changed the variable name.  If anyone asks for this feature in the C++ API I'll add it.

lfield:
I testing both the python and php. Python returns:
`ssl.SSLError: ('The read operation timed out',)`
PH returns:
`Error: HTTP error`
As far as I understand both are behaving as designed.

PR #3380, opened 2019-11-20
client: fix bug that broke HTTP GUI RPC for authenticated ops

davidpanderson:
electron sample GUI: add support for starting the client,
and for attaching a project

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -1388,15 +1388,15 @@ static bool authenticated_request(char* buf) {
     int auth_id;
     long auth_seqno;
     char auth_hash[256];
-    char* p = strstr(buf, "Auth-ID: ");
+    char* p = strcasestr(buf, "Auth-ID: ");
     if (!p) return false;
     int n = sscanf(p+strlen("Auth-ID: "), "%d", &auth_id);
     if (n != 1) return false;
-    p = strstr(buf, "Auth-Seqno: ");
+    p = strcasestr(buf, "Auth-Seqno: ");
     if (!p) return false;
     n = sscanf(p+strlen("Auth-Seqno: "), "%ld", &auth_seqno);
     if (n != 1) return false;
-    p = strstr(buf, "Auth-Hash: ");
+    p = strcasestr(buf, "Auth-Hash: ");
     if (!p) return false;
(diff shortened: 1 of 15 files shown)


samples/electron_gui/window.js
```diff
@@ -44,103 +49,207 @@ function gui_rpc(request) {
...
+
+// Start the BOINC client.
+// On Win, the client will find its data directory in the registry.
+// May need to set current dir on other platforms.
+// If there's already a client running,
+// the new one will detect this and exit.
+//
+function start_client() {
+	const child = execFile(
+		'c:/Users/David/Documents/BOINC_git/boinc/win_build/Build/x64/Debug/boinc.exe',
```
AenBleidd: I'm afraid, this will run on your PC only. Could you please change it?

client/gui_rpc_server_ops.cpp
```diff
@@ -1650,10 +1649,17 @@ static bool is_http_post_request(char* buf) {
...
     if ((int)strlen(p) < n) return false;
-    strcpy_overlap(buf, p);
     return true;
 }
 
+// remove HTTP header from request
+//
+static void strip_http_header(char* buf) {
+    char* p = strstr(buf, "\r\n\r\n");
+    p += 4;
```
AenBleidd: Dangerous arithmetic: if substring is not found, return value will be NULL, and this code will lead to segfault

samples/electron_gui/window.js
davidpanderson: This is demo code.  It's not intended to run anywhere other than my PC.

client/gui_rpc_server_ops.cpp
davidpanderson: This is called only if is_http_post_request() return true,
so we already know the string is there.

samples/electron_gui/window.js
AenBleidd: But you share this code with other. It means that anyone who downloads this should be able to run it without changing manually any file

AenBleidd: But you share this code with other. It means that anyone who downloads this should be able to run it without changing manually any file

davidpanderson: It's example code, not intended to work out of the box

PR #3381, opened 2019-11-22
client: add the acct_mgr_info authenticator response

HiMrHu:
We use our account manager and use HTTP for additional communication, so the authenticator is needed for response,

### DIFF
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -664,6 +664,7 @@ static void handle_acct_mgr_info(GUI_RPC_CONN& grc) {
         || strlen(gstate.acct_mgr_info.authenticator)
     ) {
         grc.mfout.printf("   <have_credentials/>\n");
+	grc.mfout.printf("   <authenticator>%s</authenticator>\n", gstate.acct_mgr_info.authenticator);
     }
 
     if (gstate.acct_mgr_info.cookie_required) {
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -664,7 +664,7 @@ static void handle_acct_mgr_info(GUI_RPC_CONN& grc) {
         || strlen(gstate.acct_mgr_info.authenticator)
     ) {
         grc.mfout.printf("   <have_credentials/>\n");
-	grc.mfout.printf("   <authenticator>%s</authenticator>\n", gstate.acct_mgr_info.authenticator);
+        grc.mfout.printf("   <authenticator>%s</authenticator>\n", gstate.acct_mgr_info.authenticator);
     }
 
     if (gstate.acct_mgr_info.cookie_required) {


TheAspens:
Can you explain further the use case for returning the authenticator from the core client in the RPC for account manage information?  Thanks!

client/gui_rpc_server_ops.cpp
```diff
@@ -664,6 +664,7 @@ static void handle_acct_mgr_info(GUI_RPC_CONN& grc) {
         || strlen(gstate.acct_mgr_info.authenticator)
     ) {
         grc.mfout.printf("   <have_credentials/>\n");
+	grc.mfout.printf("   <authenticator>%s</authenticator>\n", gstate.acct_mgr_info.authenticator);
```
TheAspens: Can you replace the tab with spaces so that it properly indents?  Thanks!

TheAspens:
Just noticed that this was a request to merge into David's dpa_gui_rpc_js branch.  I recommend fixing the tab vs spaces issue I mention but otherwise I leave the review up to @davidpanderson

davidpanderson:
What's the use case (in detail)?
Normally all network communication (e.g. AM RPCs) is done by the client.

HiMrHu:
Sometimes our account manager server requires the client to upload the client state, such as task state, and send it through another HTTP request. The account manager server needs the authenticator

client/gui_rpc_server_ops.cpp
HiMrHu: Sorry, it's my fault

HiMrHu:
First, I reading BOINC/acct_mgr_reply.xml file get authenticator, And then I want to get this through the RPC response。

davidpanderson:
A better way to do this:
An AM reply can include a flag <send_task_state/>.
If set, the next AM request will include a list of tasks and their states.
Would that handle your use case?

It's better for network requests to come from the client, not the GUI.
The client knows about HTTP proxies.

HiMrHu:
let me try this.

davidpanderson:
I added the feature above:
If an AM reply contains <send_tasks_active/> or <send_tasks_all/>,
the next AM request will contain a list of tasks, same as in a GUI RPC get_tasks reply.

This is in the dpa_gui_rpc_js branch.

HiMrHu:
Thank you, professor




------------------&nbsp;原始邮件&nbsp;------------------
发件人: "David Anderson"<notifications@github.com&gt;; 
发送时间: 2019年11月25日(星期一) 下午4:46
收件人: "BOINC/boinc"<boinc@noreply.github.com&gt;; 
抄送: "胡先生"<454053152@qq.com&gt;; "State change"<state_change@noreply.github.com&gt;; 
主题: Re: [BOINC/boinc] client: add the acct_mgr_info authenticator response (#3381)




I added the feature above:
 If an AM reply contains <send_tasks_active/&gt; or <send_tasks_all/&gt;,
 the next AM request will contain a list of tasks, same as in a GUI RPC get_tasks reply.
 
This is in the dpa_gui_rpc_js branch.

Issue #3386, opened 2019-11-27
MAX_JOBS_PREF and MAX_CPUS_PREF not behaving as expect. 


lfield:
This is related to the MAX_JOBS_PREF and MAX_CPUS_PREF project preferences as added in this [commit](https://github.com/BOINC/boinc/commit/fb5b9d).

The current the behavior for a multi-threaded application is as follows (threads are cpus for the VM apps):
```
Max 1 CPU, Max 1 Job  = >  1 single threaded job
Max 2 CPU, Max 1 Job => 2 threaded job 
Max 1 CPU, Max 2 Job  => 1 single threaded job
Max 2 CPU, Max 2 Job => 2 x 2 threaded jobs
```
In practice Max CPUs is used by users to set the number of threads and hence CPUs to be used by a VM, hence Max 1 CPU, Max 2 Job  => 1 single threaded job does not function as wished. In this case they would like to run two single CPU jobs.

lfield:
Removing this [if statement](https://github.com/BOINC/boinc/blob/master/sched/sched_send.cpp#L148) partially resolves the issue, however the Max # CPU preference cannot be used to limit the number CPUs used by the VM. This means the VM will have the maximum number of CPUs given by _g_wreq->effective_ncpus_.

lfield:
The number of CPUs used by a VM can be controlled if the if statement from sched_send is added to plan_class.cpp after the [max_threads](https://github.com/BOINC/boinc/blob/master/sched/plan_class_spec.cpp#L923) check. The remaining issue is that in one request if 3 ncpus are available and max_cpu = 2, two 2 CPU jobs are returned where as it should be one 2 CPU and one 1 CPU. If the requests are made separately the correct response is given.

PR #3388, opened 2019-12-02
Moved max_cpus check

lfield:
Moving the max_cpus check to address #3386.

The max_cpus preference is used to set the maximum number of CPUs (threads) to be used by a VM application.


AenBleidd:
@davidpanderson, could you please review this? Thanks in advance

ChristianBeer:
This also removes the max_cpus check for single threaded applications. Which will then affect how much work is send to the client. If a user has set `project_prefs.max_cpus` to 1 intentionally on his quadcore he/she will get 4 times the amount of work send with this change as compared to now.

PR #3400, opened 2019-12-13
[Manager] Fix Manager crash in APP_VERSION::clear()

AenBleidd:
In 219a540 a circular dependency was introduced.
This commit fixes it in a similar way from the commit mentioned above for another structures.

### DIFF
--- a/lib/gui_rpc_client.h
+++ b/lib/gui_rpc_client.h
@@ -222,6 +222,7 @@ struct APP_VERSION {
     PROJECT* project;
 
     APP_VERSION();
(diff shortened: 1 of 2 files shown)


CharlieFenton:
I have tested this change on my Mac and the BOINC Manager no longer crashes. Thank you for the fix. Should i merge it or wait for David's review?

CharlieFenton:
I am about to create another PR that includes other changes to these files, and it would be better if this was merged first. So I would very much like to merge tonight.

PR #3402, opened 2019-12-13
Scheduler (client and server): message tweaks

### LINKED ISSUE #3375: Revisit 'project requested delay' messages
As the author of #3360, please may I ask for a reconsideration of the solutions implemented in #3370 and #3371? As these issues / PRs are all now closed, I think it's cleaner to re-gather the threads in a single place.

#3370: The new 'delay' value shown to the user is the 'time left' in the backoff from the previous RPC. But it isn't.

There are three RPCs in play: previous, current, next. Because we are in the middle of the current RPC, the server will request the full min_sendwork_interval before the next RPC - the amount of time used up in the interval between previous and current RPCs is now ancient history and should be disregarded.

#3371: Because the user would now be seeing the full min_sendwork_interval in the 'too recent' message, why not display it as a normal (non-debug) message in the scheduler reply, and avoid the confusion arising in the first place?

For information, I think the problem arises in two places:
1) Impatient users over-clicking the 'update' button when work is scarce. That's their own silly fault, and we don't need to do anything except tell them what the delay should have been.
2) When performing maintenance chores which require a BOINC restart or computer reboot. The current server delay is non-persistent across restarts, and if the reboot takes less than 5 minutes (SETI) or 1 hour (CPDN), the client may request new work before the curfew has ended.

[this issue fao @delta1512, @Rytiss, @AenBleidd]

RichardHaselgrove:
client - show requested delay always (why hide it?)
server - ignore residual delay from previous RPC

Fixes #3375

### DIFF
--- a/sched/handle_request.cpp
+++ b/sched/handle_request.cpp
@@ -1416,13 +1416,12 @@ void process_request(char* code_sign_key) {
             if (config.min_sendwork_interval) {
                 double diff = dtime() - last_rpc_time;
                 if (diff < config.min_sendwork_interval) {
-                    double t_left = config.min_sendwork_interval - diff;
                     ok_to_send_work = false;
                     log_messages.printf(MSG_NORMAL,
-                        "Not sending work - last request too recent, please wait: %f\n", t_left
+                        "Not sending work - last request too recent, please wait: %f\n", config.min_sendwork_interval
                     );
                     sprintf(buf,
-                        "Not sending work - last request too recent, please wait: %d sec", (int)t_left
+                        "Not sending work - last request too recent, please wait: %d sec", (int)config.min_sendwork_interval
                     );
                     g_reply->insert_message(buf, "low");
 
--- a/client/cs_scheduler.cpp
+++ b/client/cs_scheduler.cpp
@@ -662,9 +662,9 @@ int CLIENT_STATE::handle_scheduler_reply(
         notices.remove_notices(project, REMOVE_SCHEDULER_MSG);
     }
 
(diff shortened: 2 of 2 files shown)


sched/handle_request.cpp
```diff
@@ -1416,13 +1416,12 @@ void process_request(char* code_sign_key) {
             if (config.min_sendwork_interval) {
                 double diff = dtime() - last_rpc_time;
                 if (diff < config.min_sendwork_interval) {
-                    double t_left = config.min_sendwork_interval - diff;
                     ok_to_send_work = false;
                     log_messages.printf(MSG_NORMAL,
-                        "Not sending work - last request too recent, please wait: %f\n", t_left
+                        "Not sending work - last request too recent, please wait: %f\n", config.min_sendwork_interval
```
AenBleidd: This means that user will always see the, let's say, 'default' delay interval even if there is less time left before he can do next request.
E.g.
There is a default delay of 60 sec.
User waits 59 seconds, make another request and see the message that he should wait for 60 seconds more while in fact he had to wait only 1 second before doing next request. Isn't this misleading a little bit?

RichardHaselgrove: > he had to wait only 1 second

That's the problem - he didn't. The 1-second message came in an RPC, and the sending of that RPC re-started the timer. He **does** have to wait the full 60 seconds.

AenBleidd: Ok, I see. No questions then

AenBleidd:
I'm ok with the change and plan to merge in in 4 days (on Friday) if there will be no objections before

PR #3403, opened 2019-12-13
Added Estimated Completion column to Advanced view > Tasks

### LINKED ISSUE #2115: Fields To Show Task Estimated Completion & Relative To Deadline
It would be really useful if two extra columns were available on the Advanced View --> Tasks pane (I'm using the Windows BOINC Manager 7.8.2). These would help visualising if a task was likely to be completed before it's deadline and help users micro-manage priorities to ensure they are, if and as required. It would be especially useful for the larger, longer-running tasks.

# New Field 1: Estimated Completion

This would show the estimated date/time that a task will complete. It would be calculated as `current_datetime + remaining_estimated`

# New Field 2: Relative To Deadline

I'm not sure what the best name for this would be, but basically it would show the difference (in `days hours:minutes:seconds`) between the deadline and the "estimated completion" (as described above). This would be shown as negative if the task is expected to complete early or positive if the task is expected to overrun. This column should be sortable so you can see the tasks most likely to overrun at the top of the list and the tasks with most time in hand at the bottom - or vice-versa.

## Example

If Task1 is expected to complete on 15th September at 13:00 and the deadline is 17th September at 14:30, then this field would show something like `-2d 01:30:00` (i.e. it is expected to complete 2 days 1 hour and 30 minutes before the deadline - plenty of time to spare, no problem).

If Task2 was expected to complete at on 19th September at 14:00 but the deadline was the same day at 10:00, then it would show something like `+ 04:00:00` (i.e. it's expected to overrun by 4 hours - might want to take action to try to get this to complete before the deadline).

These new fields could look something like this (based on current time being 12/09/2017 23:00 and ignoring most existing fields for the sake of the example):

Task | Deadline | Remaining | Estimated Completion | Relative To Deadline
-----|----------|------------|------------------------|----------------------
Task1 | 17/09/2017 14:30:00 | 2d 14:00:00 | 15/09/2017 13:00:00 | -2d 01:30:00
Task2 | 19/09/2017 10:00:00 | 6d 15:00:00 | 19/09/2017 14:00:00 | +0d 04:00:00

ethanteng:
Fixes #2115

**Description of the Change**
Added a new column -- "Estimated Completion" -- to the Advanced > Tasks screen to show the calculated date and time by which a task should complete. It is positioned immediately next to the existing "Deadline" column to make it easier for users to visually assess which task(s) may not complete by the deadline.

**Alternate Designs**
I opted to design "estimated completion" as a derived field. It is calculated as "current date and time" + "remaining time". Therefore, I opted to *not* add a new saved field to the data model for it.

Also, the original issue (#2115) had specified another new field to be added, to explicitly calculate and display the delta between the estimated completion date/time and the deadline date/time. However, it is my opinion that this additional field is not necessary and that having the estimated completion date/time displayed right next to the deadline date/time is enough information for a user to assess whether a task is likely to complete by the deadline, or not. Therefore, I opted to not implement this part of the issue for now.

**Release Notes**
Added "Estimated Completion" as a new column in Advanced view > Tasks to show the date and time by which a task should complete.

### DIFF
--- a/clientgui/ViewWork.cpp
+++ b/clientgui/ViewWork.cpp
@@ -1218,6 +1218,7 @@ void CViewWork::GetDocTimeToCompletion(wxInt32 item, double& fBuffer) const {
     }
 }
 
+
 void CViewWork::GetDocReportDeadline(wxInt32 item, time_t& time) const {
     RESULT*        result = wxGetApp().GetDocument()->result(item);
 
--- a/clientgui/ViewWork.cpp
+++ b/clientgui/ViewWork.cpp
@@ -767,16 +767,18 @@ wxString CViewWork::OnListGetItemText(long item, long column) const {
                 strBuffer = work->m_strStatus;
                 break;
             case COLUMN_ESTIMATEDCOMPLETION:
-                if (work->m_fCPUTime > 0) {
+                if ((work->m_fCPUTime > 0) && (work->m_strStatus.IsSameAs("Running"))) {
                     wxDateTime now = wxDateTime::Now();
                     wxTimeSpan time_to_completion = convert_to_timespan(work->m_fTimeToCompletion);
                     wxDateTime estimated_completion = now.Add(time_to_completion);
                     FormatDateTime(estimated_completion.GetTicks(), strBuffer);
+                    // Only display the Estimated Completion time if the task has
+                    // been started and is currently running.
                 } else {
                     strBuffer = FormatTime(0);
-                    // If the task has not started (CPUTime <= 0), passing
-                    // 0 to FormatTime will return "---" as the estimated
-                    // completion time
+                    // If the task has not started (CPUTime <= 0) or is not currently
+                    // running, pass 0 to FormatTime to display "---" as the Estimated
+                    // Completion time
                 }
                 break;
         }
--- a/clientgui/ViewWork.cpp
+++ b/clientgui/ViewWork.cpp
@@ -767,7 +767,8 @@ wxString CViewWork::OnListGetItemText(long item, long column) const {
                 strBuffer = work->m_strStatus;
                 break;
             case COLUMN_ESTIMATEDCOMPLETION:
-                if ((work->m_fCPUTime > 0) && (work->m_strStatus.IsSameAs("Running"))) {
+                //if ((work->m_fCPUTime > 0) && (work->m_strStatus.IsSameAs("Running"))) {
+                if ((work->m_fCPUTime > 0) && (work->m_strStatus.Contains("Running"))) {
                     wxDateTime now = wxDateTime::Now();
                     wxTimeSpan time_to_completion = convert_to_timespan(work->m_fTimeToCompletion);
                     wxDateTime estimated_completion = now.Add(time_to_completion);
--- a/clientgui/ViewWork.cpp
+++ b/clientgui/ViewWork.cpp
@@ -276,7 +276,7 @@ CViewWork::CViewWork(wxNotebook* pNotebook) :
     m_aStdColNameOrder->Insert(_("Progress"), COLUMN_PROGRESS);
     m_aStdColNameOrder->Insert(_("Status"), COLUMN_STATUS);
     m_aStdColNameOrder->Insert(_("Elapsed"), COLUMN_CPUTIME);
-    m_aStdColNameOrder->Insert(_("Remaining"), COLUMN_TOCOMPLETION);
+    m_aStdColNameOrder->Insert(_("Remaining (estimated)"), COLUMN_TOCOMPLETION);
     m_aStdColNameOrder->Insert(_("Estimated Completion"), COLUMN_ESTIMATEDCOMPLETION);
     m_aStdColNameOrder->Insert(_("Deadline"), COLUMN_REPORTDEADLINE);
     m_aStdColNameOrder->Insert(_("Application"), COLUMN_APPLICATION);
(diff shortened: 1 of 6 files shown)


clientgui/ViewWork.cpp
```diff
@@ -1184,6 +1211,7 @@ void CViewWork::GetDocTimeToCompletion(wxInt32 item, double& fBuffer) const {
     }
 }
 
+
```
AenBleidd: Please remove extra blank line

ethanteng: Done

clientgui/ViewWork.cpp
```diff
@@ -261,7 +276,8 @@ CViewWork::CViewWork(wxNotebook* pNotebook) :
     m_aStdColNameOrder->Insert(_("Progress"), COLUMN_PROGRESS);
     m_aStdColNameOrder->Insert(_("Status"), COLUMN_STATUS);
     m_aStdColNameOrder->Insert(_("Elapsed"), COLUMN_CPUTIME);
-    m_aStdColNameOrder->Insert(_("Remaining (estimated)"), COLUMN_TOCOMPLETION);
+    m_aStdColNameOrder->Insert(_("Remaining"), COLUMN_TOCOMPLETION);
```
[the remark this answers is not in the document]
ethanteng: I added the "(estimated)" part of the column name back.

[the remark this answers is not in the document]
ethanteng: To your other point, I'm making a change so that the Estimated Completion column shows "---" if the Elapsed column is also "---" (i.e. elapsed CPU time is <= 0).

ethanteng:
@AenBleidd I made the changes you requested and addressed your concern that Estimated Completion should be empty if the job has not yet started.

clientgui/ViewWork.cpp
```diff
@@ -745,6 +766,19 @@ wxString CViewWork::OnListGetItemText(long item, long column) const {
...
                 strBuffer = work->m_strStatus;
                 break;
+            case COLUMN_ESTIMATEDCOMPLETION:
+                if (work->m_fCPUTime > 0) {
+                    wxDateTime now = wxDateTime::Now();
+                    wxTimeSpan time_to_completion = convert_to_timespan(work->m_fTimeToCompletion);
+                    wxDateTime estimated_completion = now.Add(time_to_completion);
+                    FormatDateTime(estimated_completion.GetTicks(), strBuffer);
+                } else {
+                    strBuffer = FormatTime(0);
```
AenBleidd: What if the task was started but then paused? It still will show increased time each second I believe.

RichardHaselgrove: What does BoincTasks say in that scenario? I think it would be helpful if both tools followed the same rules.

ethanteng: @AenBleidd wouldn't that be the expected behavior? If I pause a task, I would personally expect the Estimated Completion to continue to increase until I un-pause the task.

@RichardHaselgrove I'm not sure as I haven't used BoincTasks before. But, I can install it and take a look.

AenBleidd: The column name is 'Estimated Completion' but you can't estimate it for paused task. It doesn't matter whether user on scheduler paused the task, if you will just show some datetime in future that is increased every second - it means no sense because it says actually nothing about the real estimated completion time. If you want, you cat try to examine the scheduler code, find its queue and using this information try to predict when every task will start and then make the final calculation for estimated completion. But it's complex, maybe even too complex. So for the tasks that are paused or not started (let's say, the tasks that are not currently running) I'd just show empty field or dash symbol within this column. But it's up to you which path to use

ethanteng: OK, thanks for the feedback! That makes sense. And I agree, examining the scheduler code and using that to make a calculation for estimated completion is probably overkill for this purpose. I'll just go ahead and show the same 3 dashes (---) when a task is not currently running.

AenBleidd:
I had all my tasks paused. I launch newly built Manager with your changes. I see new column. I unpaused tasks execution but new column still shows dashes:
image

davidpanderson:
I don't think the client should show "estimated completion time",
because it doesn't know what it is.
A job can be preempted by another job at any time;
a variety of other factor can cause a job not to run.
The client can estimate how much runtime remains for a given job,
but not when it will be completed.

ethanteng:
@AenBleidd I implemented a fix which should address the behavior you observed. The logic for determining when to display the Estimated Completion was erroneously too restrictive before.

clientgui/ViewWork.cpp
```diff
@@ -745,6 +766,22 @@ wxString CViewWork::OnListGetItemText(long item, long column) const {
             case COLUMN_STATUS:
                 strBuffer = work->m_strStatus;
                 break;
+            case COLUMN_ESTIMATEDCOMPLETION:
+                //if ((work->m_fCPUTime > 0) && (work->m_strStatus.IsSameAs("Running"))) {
+                if ((work->m_fCPUTime > 0) && (work->m_strStatus.Contains("Running"))) {
```
AenBleidd: This check will work for English locale only.
See https://github.com/BOINC/boinc/blob/66c367e11e05712bdc04b7eeabaf522e5f64fe9f/clientgui/MainDocument.cpp#L2553 for reference

ChristianBeer: This requested change needs to be addresses in order to move this PR forward.

PR #3410, opened 2019-12-24
[WIP] User option to exclude a task from being scheduled to run if it is simulated to likely miss its deadline

### LINKED ISSUE #2957: Option to automatically suspend tasks if estimated time > deadline + X hours
**Describe the problem**
As per title

I would like to have an option to check every e.g. 15 minutes, if estimated time > deadline + X hours, abort the task automatically.
So that I do not do this manually, nor do I complete the task but not receive any credit.

**Additional context**
Option should be disabled by default.
I have a low Nvidia card running GPUGRID tasks, and sometimes estimated time > deadline.
Similar to https://github.com/BOINC/boinc/issues/2115

ethanteng:
Fixes #2957 

**Description of the Change**
Whenever BOINC main() polls to generate the list of tasks to run - and the order in which to run them - any tasks that are at risk of slipping past its deadline (as simulated by BOINC) is excluded.

This is a WIP and simply adds the logic to exclude a task from being scheduled if the result is simulated to likely slip past the deadline. I will work on the user configuration option next, e.g. "Exclude jobs if it is at risk of missing its deadline (Y/N)"?

**Alternate Designs**
The original Feature Request was to add an option to explicitly **abort** a task. In my view, this is a cleaner solution to the problem the Feature Request is looking to solve. It relies on the existing, underlying logic for scheduling, prioritizing, and clean-up of tasks, i.e. nothing is explicitly aborted, only excluded from the run list. Thus, the only change would be in the logic applied in cpu_sched->can_schedule().

**Release Notes**
Add user-configurable option to exclude tasks that are likely to miss their deadline from being scheduled to run.

### DIFF
--- a/client/cpu_sched.cpp
+++ b/client/cpu_sched.cpp
@@ -146,6 +146,8 @@ struct PROC_RESOURCES {
                 atp->needs_shmem = false;
             }
         }
+        if (rp->rr_sim_misses_deadline) return false;
+            // Don't schedule if the result is simulated to miss its deadline
         if (rp->schedule_backoff > gstate.now) return false;
         if (rp->uses_gpu()) {
             if (gpu_suspend_reason) return false;


client/cpu_sched.cpp
```diff
@@ -146,6 +146,8 @@ struct PROC_RESOURCES {
                 atp->needs_shmem = false;
             }
         }
+        if (rp->rr_sim_misses_deadline) return false;
+            // Don't schedule if the result is simulated to miss its deadline
```
AenBleidd: So, such tasks that could possibly not meet the deadline will never have a chance to run?
I think this is not nice behavior, especially for those projects who has sometimes some very small tasks after the big one.

davidpanderson:
I'm not convinced this is a desirable feature.  In any case it should not be implemented at such a low level.

CharlieFenton:
I don't know if this still true, but I seem to remember that the estimates for how long a task will run are often very wrong, especially when a project application is first deployed. I believe the server uses actual run times to refine its predictions for subsequent tasks it sends later. 

I believe there are also some projects where it is impossible to accurately predict run time because of variations in the input data. So I wonder if this would prevent tasks from running that would actually finish on time, thus defeating BOINC's method for refining its estimates.

I also have a concern that advanced users may be a tiny minority of BOINC users; most just want to install BOINC and let it run by itself. I worry that as we add more options for advanced users, the complexity will scare away more users. I have long felt that the path to getting more volunteer users is to simplify BOINC by eliminating options, rather than adding more. )I feel the Simple View is not enough to take care of this, as there are too many things you can do only with the Advanced View.)

AenBleidd:
This behavior definitely should not be automatic, and the original ticket should be implemented in a different way.
Closing this without merging.

PR #3418, opened 2020-01-07
[Android] Make BOINC always run foreground service

AenBleidd:
On the latest Android SDKs there are a lot of restrictions for background processes.
So if the service is run in background, it could be killed too early and BOINC will not start.
Removed setting to show notification when running only because now it makes no sense since Notification should be shown always when foreground service is running.

### DIFF
--- a/android/BOINC/app/src/main/aidl/edu/berkeley/boinc/client/IMonitor.aidl
+++ b/android/BOINC/app/src/main/aidl/edu/berkeley/boinc/client/IMonitor.aidl
@@ -97,11 +97,9 @@ List<ImageWrapper> getSlideshowForProject(in String url);   // clientStatus.getS
 ////// app preference ////////////////////////////////////////////
 void setAutostart(in boolean isAutoStart);          // Monitor.getAppPrefs().setAutostart(boolean);
 void setShowNotificationForNotices(in boolean isShow);   // Monitor.getAppPrefs().setShowNotificationForNotices(boolean);
-void setShowNotificationDuringSuspend(in boolean isShow);   // Monitor.getAppPrefs().setShowNotificationDuringSuspend(boolean);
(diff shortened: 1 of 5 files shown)


davidpanderson:
I don't understand the Android client; if this works it has my blessing.

Question: we rely on the GUI to inform the client about critical things like battery level (via GUI RPCs).  Can we guarantee that the GUI will always be running if the client is running?  If not, we need to have the client suspend processing if it hasn't heard from the GUI in a while.

AenBleidd:
We don't rely on GUI itself. Android have GUI and Monitor. The Monitor itself is a service (like daemon on linux) but has a little bit different behaviour. The Monitor itself won't run on system start-up, it relies on system events that fire it to start. Also GUI can start Monitor too. The problem is that with the current design, Monitor is a background service, It was fine several years ago. But now Google implemented more aggressive policies on background services prevent them from running a lot in background. So if someone need to have a service that run instantly, they need to make in foreground (and visible to the user, because every foreground service must have a Notification). Current design of BOINC Manager for Android converts background service to foreground when running project applications. New design will always run foreground service (and user will always see a Notification that BOINC Manager is launched, when BOINC run some tasks, user additionally will see the exact tasks that are executing currently).
With this new design it is less likely that foreground service will be killed and thus left client run forever without any information about the environment.
But this need to be tested additionally.
Current release (7.16.3) has a lot of issues, some of them intercept with each other making hard to test main application workflow. 
So I ask you to announce a Android only 7.16.4 release to Google Play Store beta channel. This will help us to involve many people (e.g. from GRC community) with many different device configurations to test new design.

AenBleidd:
@adamradocz, main goal is to have foreground service only instead of switching between background and foreground modes instantly on events.
So this particular pull request contains two changes:
1 - Make Monitor to be foreground service always (ref. https://developer.android.com/about/versions/oreo/background.html )
2 - Remove the setting that is obsolete because of the change in point 1.

PR #3422, opened 2020-01-15
Fix error messages in account creation RPC

### LINKED ISSUE #3420: Change request for change of wording in the BOINC Manager GUI, currently: 'Account creation is disabled'
Change requested:
Is it possible that in the BOINC Manager 'Add account' popup a link to the project website or a statement saying something like: 'Account creation through the client is disabled, please first create an account through the <project website> and then try to reattach to the project' could be added to this'. The current statement is: 'Account creation is disabled'.

Context:
In climateprediction.net we now have account creation disabled through the BOINC Manager. When new volunteers download the BOINC client on to their machines and then try to attach to the climateprediction.net using the 'Add account' popup and if they don't have an existing account, they are met with the message 'Account creation is disabled' with no link or explanation of how to create a new account. I think we are currently losing a lot of new volunteers through this.

Screenshot from 2020-01-08 14-58-08

davidpanderson:
The create-account RPC returned "Account creation disabled"
even if account creation on the web site is allowed.
Fix this.

Fixes #3420

### DIFF
--- a/html/user/create_account.php
+++ b/html/user/create_account.php
@@ -34,16 +34,20 @@ if ($retval) xml_error($retval);
 
 $config = get_config();
 if (parse_bool($config, "disable_account_creation")) {
-    xml_error(ERR_ACCT_CREATION_DISABLED);
+    xml_error(-1, "The project is not accepting new accounts.");
 }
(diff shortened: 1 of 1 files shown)


TheAspens:
Looks good.  Thanks for doing the update.

PR #3426, opened 2020-01-23
Mac: allow 32-bit applications only if OS X version < 10.15

CharlieFenton:
**Description of the Change**
Don't allow 32-bit project applications if running on Mac OS 10.15 or later

**Alternate Designs**
none

**Release Notes**
N/A


TheAspens:
Thanks for the fix!

PR #3432, opened 2020-01-27
client, Win: don't truncate CPU features

davidpanderson:
Make the size of p_features a #define; set it to 1024


AenBleidd:
all_projects_list.xml changes is not related to this PR.
@davidpanderson, if you want, I can split this PR into two that could be reviewed and merged separately

davidpanderson:
Sure, if you want.

AenBleidd:
This PR is splitted into two:
- #3437 
- #3438

Issue #3442, opened 2020-01-28
7.16.4, remove cc_config.xml results in "Application uses missing ATI GPU" errors


Ageless93:
So in order to test the <max_stderr_file_size> problem I had elsewhere, I removed my cc_config.xml file and restarted BOINC. It immediately forgot all about the anonymous platform AMD GPU applications I use to run for Seti@Home and the default AMD GPU applications in Seti Beta. 

```
28-Jan-2020 19:47:04 [---] cc_config.xml not found - using defaults
28-Jan-2020 19:47:04 [---] Starting BOINC client version 7.16.4 for windows_x86_64
28-Jan-2020 19:47:04 [---] Libraries: libcurl/7.47.1 OpenSSL/1.0.2s zlib/1.2.8
28-Jan-2020 19:47:04 [---] Data directory: G:\BOINC
28-Jan-2020 19:47:04 [---] Running under account elst9
28-Jan-2020 19:47:04 [---] OpenCL: AMD/ATI GPU 0: AMD Radeon RX 5700 XT (driver version 3004.8 (PAL,LC), device version OpenCL 2.0 AMD-APP (3004.8), 8176MB, 8176MB available, 4877 GFLOPS peak)
28-Jan-2020 19:47:04 [SETI@home] Found app_info.xml; using anonymous platform
28-Jan-2020 19:47:04 [---] app version refers to missing GPU type ATI
28-Jan-2020 19:47:04 [---] app version refers to missing GPU type ATI
...
28-Jan-2020 19:47:04 [---] app version refers to missing GPU type ATI
28-Jan-2020 19:47:04 [---] app version refers to missing GPU type ATI
28-Jan-2020 19:47:04 [SETI@home] Missing coprocessor for task 22ja20ab.24388.22153.9.36.234_0
28-Jan-2020 19:47:04 [SETI@home] Missing coprocessor for task 22ja20ab.24388.22153.9.36.239_1
...
28-Jan-2020 19:47:04 [SETI@home] Missing coprocessor for task blc35_2bit_guppi_58692_01957_HIP80644_0118.27983.818.21.44.60.vlar_0
28-Jan-2020 19:47:04 [---] app version refers to missing GPU type ATI
28-Jan-2020 19:47:04 [SETI@home Beta Test] Application uses missing ATI GPU
28-Jan-2020 19:47:04 [---] app version refers to missing GPU type ATI
[...]

PR #3443, opened 2020-01-28
client: fix bug in GPU detection

davidpanderson:
PR #3364 changed the way we clear structures.
This introduced a bug: HOST_INFO::clear_host_info()
intentionally didn't clear HOST_INFO::coprocs.
But it was replaced with HOST_INFO::clear(), which did.
This caused the client to lose GPU info.

Fix: restore HOST_INFO::clear_host_info(), and add a comment
to avoid future errors like this.

Also add some comments in GPU detection,
which is woefully lacking in them.

### DIFF
--- a/client/boinc_cmd.cpp
+++ b/client/boinc_cmd.cpp
@@ -426,7 +426,7 @@ int main(int argc, char** argv) {
         }
     } else if (!strcmp(cmd, "--set_host_info")) {
         HOST_INFO h;
-        h.clear();
+        h.clear_host_info();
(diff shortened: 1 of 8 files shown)


davidpanderson:
Fixes (I hope) #3442

AenBleidd:
I don't like this solution because it have a lot of variables manually cleared. And in case of adding new variable it's very easy to forget to add all initialization/clean/etc functions. But in order to do this properly, it's necessary to reimplement the architecture we have and it's too much work.
So I'm ok with the current implementation and going to merge it.

Issue #3247, opened 2019-08-10
Manager: writes garbage to cc_config.xml


RichardHaselgrove:
**Describe the bug**
When the 'Event log options...' dialog is used to modify the active set of Event Log flags, garbage is written to cc_config.xml.

**Steps To Reproduce**
1. Open Event Log Options dialog, change a setting, apply or save your changes
2. Inspect the contents of cc_config.xml visually
3. (optional) Shut down and restart the BOINC client...

**Expected behavior**
Existing option settings are left unchanged'

**Screenshots**
Original entries in cc_config.xml:
```
    <options>
        <abort_jobs_on_exit>0</abort_jobs_on_exit>
        <allow_multiple_clients>1</allow_multiple_clients>
        <allow_remote_gui_rpc>0</allow_remote_gui_rpc>
        <disallow_attach>0</disallow_attach>
```

New entries in cc_config.xml:
```
    <options>
        <abort_jobs_on_exit>1069290768</abort_jobs_on_exit>
        <allow_gui_rpc_get>0</allow_gui_rpc_get>
        <disallow_attach>0</disallow_attach>
```

**System Information (please complete the following information):**
 - OS: Windows, Linux (Mac untested)
 - BOINC Version: 7.16.1 (new in this version)

**Additional context**
Possibly came about as a side effect of #3214 (added <allow_gui_rpc_get>), but I haven't tracked it down yet - I'll look over the weekend.

With a value >0 in abort_jobs_on_exit - guess what happens on exit! That's the blocker.

Problem was noticed as a result of using the **release** version of 7.16.1 from @LocutusOfBorg PPA, verified with self-build version of boincmgr for Windows (no test version available yet).

RichardHaselgrove:
Further testing reinforces my belief that this is a Manager problem. Sending the 'update log flags' RPC remotely from an older version Manager doesn't corrupt the cc_config.xml file. But #3214 didn't modify the manager, and the changes to  lib/cc_config.cpp look benign. Still investigating.

RichardHaselgrove:
I've been trying to trace back this problem, to find out for certain when and where it started. Bear in mind that (a) Appveyor builds when the PR is created, not when it is merged, and (b) Appveyor wasn't working for a long time. No artifacts were produced for PR 3214.

So I've tested the artifacts for PR 3230 (built July 20, before 3214 was merged) and PR 3233 (built July 26).

[...]

PR #3445, opened 2020-01-29
client: make max_std(err/out)_file_size double

### LINKED ISSUE #3435: Unrecognized tag in cc_config.xml: <max_stderr_file_size>
Started with 7.16.3 apparently, only noticed it first time in 7.16.4
BOINC starts and I immediately get Unrecognized tag in cc_config.xml: <max_stderr_file_size>

I've got <max_stderr_file_size set to 18GB, or 180000000000 bytes. It's not working, at 2MB my stderrdae.txt file renamed into *.old and made a new file. 
The cc_config.xml file is the one made by BOINC, when I adjusted the Event Log Flags menu option. 
I've opened it in Notepad++ and saved it again in ANSI format. That didn't make a difference, the error is still there. 

Sent my cc_config.xml file to David.

AenBleidd:
File sizes should always be double, not int

This fixes #3435

### DIFF
--- a/client/log_flags.cpp
+++ b/client/log_flags.cpp
@@ -419,8 +419,8 @@ int CC_CONFIG::parse_options_client(XML_PARSER& xp) {
         if (xp.parse_int("max_event_log_lines", max_event_log_lines)) continue;
         if (xp.parse_int("max_file_xfers", max_file_xfers)) continue;
         if (xp.parse_int("max_file_xfers_per_project", max_file_xfers_per_project)) continue;
-        if (xp.parse_int("max_stderr_file_size", max_stderr_file_size)) continue;
-        if (xp.parse_int("max_stdout_file_size", max_stdout_file_size)) continue;
+        if (xp.parse_double("max_stderr_file_size", max_stderr_file_size)) continue;
+        if (xp.parse_double("max_stdout_file_size", max_stdout_file_size)) continue;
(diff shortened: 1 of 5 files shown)


RichardHaselgrove:
Appveyor failed and the branch was inaccessible, so I reverted to manual copying of the changes and a home build.

As I half-expected, I now see
```
        <max_stderr_file_size>0.000000</max_stderr_file_size>
        <max_stdout_file_size>52428800.000000</max_stdout_file_size>
```
in cc_config.xml following an Event Log flag edit.

Call me a pedant, but what's this obsession with using floating point arithmetic everywhere? What's the purpose? Is there any language efficiency in recording file sizes to the microbyte? I ran an editor over my client_state.xml file, and it found 1,609 instances of ".000000".

It struck me particularly when I got caught by the 'stray comma' bug (#3247). How could we be writing "1069290768" into a variable which should be designated 'boolean'?

tristanolive:
It's the default when using the %f format in printf() for there to be six digits of precision, regardless of whether the number involved requires it. Combing through code to replace occurrences of %f with %f.0 would remove the decimal precision altogether. I'm going to go ahead and merge this PR, though, as that sort of output cleanup/reformatting is a bit of a different task.

davidpanderson:
we use doubles for file size because they have big mantissas, not because they can represent fractions

PR #3449, opened 2020-02-05
scheduler: fix crashing bug in anonymous platform case.

davidpanderson:
My change a while back to enforce job limits correctly
didn't work with anonymous platform requests.

Also fixed wrong message and int/double format error when
reporting last request too soon.


RichardHaselgrove:
I'm slightly nervous about this. I've tested it at LHC-dev, but I think @lfield was hoping for a second test site before committing to a replacement Server Stable' release. SETI Beta was still exhibiting the problem this morning, but I think it hasn't been updated yet. The SETI Main project seems - knock on wood - to be back to running as normal, so I've written to @SETIguy to ask if he could facilitate a test.

PR #3455, opened 2020-02-07
Drupal: Add privacy description statement for preferences page

drshawnkwang:
Fixes https://dev.gridrepublic.org/browse/DBOINCP-521

**Description of the Change**
Add description of the exporting content consent question, with link to privacy policy. There are also tw-admin menu options to control this feature.

**Release Notes**
N/A


drshawnkwang:
@tristanolive - This is ready for review.

tristanolive:
I stumbled over the preg_match() call for a minute, expecting it was going to do something with matches as is often the case, but then realized that wouldn't make sense anyway and it's just looking for a string in another string -- pretty straightforward and works fine.

PR #3460, opened 2020-02-12
Web: fix ReCaptcha problem

davidpanderson:
I have no idea why, but out of nowhere ReCaptcha stopped working on both SU and S@h.
Changing an arg to the recaptcha constructor
(from a Stack Overflow entry from 2015) fixed the problem.


AenBleidd:
Clean PR is created: #3461

PR #3461, opened 2020-02-12
Web: fix ReCaptcha problem

AenBleidd:
I have no idea why, but out of nowhere ReCaptcha stopped working on both SU and S@h.
Changing an arg to the recaptcha constructor
(from a Stack Overflow entry from 2015) fixed the problem.

### DIFF
--- a/html/inc/recaptchalib.php
+++ b/html/inc/recaptchalib.php
@@ -45,7 +45,7 @@ function boinc_recaptcha_get_html($publickey) {
 function boinc_recaptcha_isValidated($privatekey) {
     if ($privatekey) {
         // tells ReCaptcha to use fsockopen() instead of get_file_contents()
-        $recaptcha = new \ReCaptcha\ReCaptcha($privatekey, new \ReCaptcha\RequestMethod\SocketPost());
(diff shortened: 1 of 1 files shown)


html/inc/recaptchalib.php
```diff
@@ -45,7 +45,7 @@ function boinc_recaptcha_get_html($publickey) {
 function boinc_recaptcha_isValidated($privatekey) {
     if ($privatekey) {
         // tells ReCaptcha to use fsockopen() instead of get_file_contents()
-        $recaptcha = new \ReCaptcha\ReCaptcha($privatekey, new \ReCaptcha\RequestMethod\SocketPost());
+        $recaptcha = new \ReCaptcha\ReCaptcha($privatekey, new \ReCaptcha\RequestMethod\CurlPost());
```
AenBleidd: It is mentioned [here](https://github.com/google/recaptcha/blob/master/src/ReCaptcha/RequestMethod/CurlPost.php) that `this requires the cURL extension to be enabled in PHP`.
I'm not sure but I believe this extension should be either checked or at least mentioned in documentation if its not enabled by default

PR #3466, opened 2020-02-19
Use EnvironmentFile /etc/default/boinc-client for systemd startup

### LINKED ISSUE #3393: [Explained]systemd service not passing arguments to client
Arguments to a service, such as the boinc client, need to be put into an environmental file.  That file exists, but is not identified in the service.

in /lib/systemd/system/boinc-client.service make the following change:

from
```
WorkingDirectory=/var/lib/boinc
ExecStart=/usr/bin/boinc
```

to 
```
WorkingDirectory=/var/lib/boinc
ExecStart=/usr/bin/boinc $BOINC_OPTS
EnvironmentFile=/etc/default/boinc-client
```

the script /etc/init.d/boinc-client attempts to pass $BOINC_OPTS but fails.  The spec
https://wiki.archlinux.org/index.php/Systemd
calls for arguments to be in an environmental file.
If that init.d file is depreciated (so I have heard) then the /etc/default/boinc-client should not be depreciated as it is useful for passing arguments and should be modified to only have something like the following
```
# Here you can specify additional options to pass to the BOINC core client.
# Type 'boinc --help' or 'man boinc' for a full summary of allowed options.
#BOINC_OPTS="--allow_remote_gui_rpc"
BOINC_OPTS=""
```

Right now it has a lot of stuff used by the init.d script but it worked fine when I ran
```
systemctl start boinc-client
```
and the client accepted the arguments I had put it.

yuri-sevatz:
Fixes #3393 

**Description of the Change**
Allow customizing $BOINC_OPTS in systemd unit files

**Alternate Designs**
N/A - Trivial change

**Release Notes**
Allow systemd startup to be customized via /etc/systemd/system unit file


yuri-sevatz:
Saw this mentioned in #3393 

One issue here is this: client/scripts/Makefile.am:
```
    elif [ -d /etc/default ] ; then \
      $(INSTALL) -d $(DESTDIR)$(sysconfdir)/default ; \
      $(INSTALL_DATA) $(srcdir)/boinc-client.conf $(DESTDIR)$(sysconfdir)/default/boinc-client ; \
    else \
      $(INSTALL) -d $(DESTDIR)$(sysconfdir) ; \
      $(INSTALL_DATA) $(srcdir)/boinc-client.conf $(DESTDIR)$(sysconfdir)/boinc-client.conf ; \
    fi
```

I think the serivce.in file will have to be customized based on the detected install path.

yuri-sevatz:
This will probably go better to deal with different .conf file paths, without specifying EnvironmentFile.

The usual way is to create a file in /etc/systemd/system/boinc-service.d/
```
# cat /etc/systemd/system/boinc-client.service.d/00gentoo.conf 
[Service]
Environment="BOINC_OPTS=--allow_remote_gui_rpc"
```

Flowdalic:
I don't think there is something fundamentally wrong with this change. But ideally daemons, like boinc, are started with a single argument: the path to the daemons configuration file. And all daemon specific configuration, like `--allow_remote_gui_rpc` show, go there. Boinc has cc_config.xml for that.

Adding stuff via systemd's environment variable mechanism seems like a detour, that can (and IMHO should) be avoided.

yuri-sevatz:
I'm closing this PR since it seems like this isn't the preferred way.

https://github.com/BOINC/boinc/issues/3393 will need a new solution involving cc_config.xml

PR #3467, opened 2020-02-24
 web: if fail to connect to DB, say so rather than "project stopped".

davidpanderson:


AenBleidd:
Clean #3468 is created

PR #3472, opened 2020-02-25
[lib][mac] Add missing files to Makefile.am

brevilo:
* Should have been part of #2597

**Description of the Change**
Add missing files to Makefile.am, required for building `libboinc` on macOS using `make`.

**Release Notes**
n/a

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -96,6 +96,7 @@ if OS_DARWIN
 mac_sources = \
     procinfo_mac.cpp \
     mac/mac_backtrace.cpp \
+    mac/mac_branding.cpp \
     mac/mac_spawn.cpp \
     mac/QBacktrace.c \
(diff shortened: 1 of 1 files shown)


AenBleidd:
@brevilo, what about #3218?
It contains similar changes but with 2 more files. What do you think about these files?

brevilo:
You're right, I missed those files when I had a look at #3218 yesterday. I haven't come across a requirement for `QTaskMemory.*` but I'm only building the lib/api stuff. Since I presume @lfield has a good reason for his PR and the Makefiles are notoriously lacking, I'm all for merging his PR. Thus I'm closing mine in favor of his.

CharlieFenton:
The files whose names begin with Q in _lib/mac/_ are used to generate backtrace information on the Mac when there is a crash to assist in debugging the crash.

PR #3481, opened 2020-02-28
Remove remnants to subversion

smoe:
I had a problem running that code on "ash" on OpenWRT. No idea what could possibly be wrong about it, but since there is no need for that piece of code, it should just go IMHO.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -35,10 +35,6 @@ AC_CONFIG_SRCDIR(lib/shmem.cpp)
 AC_REVISION([$Revision$])
 REV=`echo '$Revision$' | awk "{print $2}"`
 RDATE=`date '+%Y.%m.%d'`
-if test -d .svn ; then
-  REV=`svn info | grep Revision | awk '{print $2}'`
-  RDATE=`svn info | grep "Last Changed Date:" | awk '{print $4}' | sed 's/-/./g'`
-fi
 AC_SUBST(REV)
 AC_SUBST(RDATE)
 


configure.ac
```diff
@@ -35,10 +35,6 @@ AC_CONFIG_SRCDIR(lib/shmem.cpp)
 AC_REVISION([$Revision$])
 REV=`echo '$Revision$' | awk "{print $2}"`
 RDATE=`date '+%Y.%m.%d'`
-if test -d .svn ; then
```
AenBleidd: Is this full block is really necessary?
https://github.com/BOINC/boinc/blob/ec2f4ee669e2b43bbfea41a763ab7b4c5e83d7d8/configure.ac#L35-L43
I'd delete it completely

brevilo: Why? Even without SVN `REV` and `RDATE` are set. Sure, if those two aren't used anywhere then the block can as well go.

AenBleidd: As far as I see, these two variables are not used anymore

brevilo: You sure? Aren't they used in every `Makefile.in`?

brevilo: They seem to be used in some [old `*.in` files](https://github.com/BOINC/boinc/search?q=RDATE&unscoped_q=RDATE). Are those still used/needed. Shouldn't they be generated dynamically anyway (via `automake`)?

AenBleidd: Search shows no usages:
image
image

AenBleidd: Hm, files you mentioned just wa not shown in my search.
I'm not sure whether these files are still used

brevilo: Yep, GitHub's search has proven to be unreliable over the past weeks 😒

brevilo: Yep, GitHub's search has proven to be unreliable over the past weeks 😒

Issue #2964, opened 2019-01-12
config.h is not installed


AenBleidd:
In #1624 @tomasbrod mentioned:

> I am affected by this issue. config.h is not installed. Almost all header files installed in /usr/include include that "config.h" file making them useless without the missing header.

tomasbrod:
I think this issue was not reported earlier is because building out-of-tree apps is not supported and all apps that are built out-of-tree are using internal boinc tree copy (with config.h).

AenBleidd:
Walking around I found this thread:
https://lists.gnu.org/archive/html/autoconf/2013-07/msg00043.html
Looks like fixing this is not so trivial

tomasbrod:
I thought the `#include <boinc/config.h>` is unambiguous. Maybe the problem mentioned was caused by installing config.h to the root inclusion namespace.

brevilo:
While it might not be trivial, it still seems [straight-forward](https://www.gnu.org/software/autoconf-archive/ax_prefix_config_h.html).

brevilo:
What I don't understand is this: most people who build BOINC should be projects admins/developers. I presume that none of those effectively install BOINC (via `make install`) in the standard `prefix`, eg. `/usr/local`. Project admins rather compile BOINC (server) and copy the binaries to their production server. Project developers compile BOINC (api/libs etc.) in a dedicated build environment as a build dependency of their science apps and screensavers. In both cases BOINC should be rather isolated which would mean installing `config.h` in the configured `prefix` would mean no harm.

I also checked the depencies BOINC itself requires: they don't install `config.h` themselves so there won't be a conflict there either.

That said, not installing `config.h` seems trying to solve a non-existent problem, while causing issues for those having to compile/use BOINC as build dependency. Therefore I do see two ways forward: install `config.h` and see if it **really** causes problems (I doubt it) or use `AX_PREFIX_CONFIG_H` as mentioned in my previous comment.

CharlieFenton:
The issue is ensuring that the client and manager built by Berkeley and World Community Grid for distribution to the public can be installed on a range of Macintosh computers. Project applications are also distributed to the public and must run on a range of computers. As I wrote in my comments to PR #3471:

If you use configure to create the config.h file, the generated code may not work on older versions of OS X, and may not work at all (as implied by PR #3471.) The Xcode build uses a fixed version of config.h at clientgui/mac/config.h, and so should the make file.

[...]

PR #3493, opened 2020-03-06
Install project_specific_defines.h if configured (required by config.h)

### LINKED ISSUE #3471: config.h depends on project_specific_defines.h which is never installed
**Describe the bug**
`config.h` [depends](https://github.com/BOINC/boinc/blob/cbcaa407db44627bb0ed8e8526a848d141a426b2/configure.ac#L657) on `project_specific_defines.h` which is never installed in the configured `--prefix` path.

**Steps To Reproduce**
1. `configure` and `make` BOINC using an install `PREFIX`
2.  Build software against BOINC, e.g. using the BOINC API

**Expected behavior**
A successful build

**System Information**
 - OS: macOS
 - BOINC Version: master

brevilo:
Fixes #3471

**Description of the Change**
Installs `project_specific_defines.h` in `PREFIX/include/boinc` if `--enable-install-headers` is `configure`d. It's `include`d by `config.h` which in turn gets used when building science apps or screensavers. Right now `config.h` isn't installed, but that's a separate issue dealt with in #2964.

**Release Notes**
n/a

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -58,7 +58,8 @@ EXTRA_DIST =                                    \
 if INSTALL_HEADERS
 pkginclude_HEADERS = \
     version.h \
-    svn_version.h
+    svn_version.h \
+    project_specific_defines.h
 endif
 
 # svn_version.h should always be rebuilt.


brevilo:
Hey @AenBleidd or @lfield, could you have a look at this one? Thanks

AenBleidd:
Ideally, svn_version.h file should be refactored and renamed because now we're using git and it is a little bit confusing. But this is not in the scope of this ticket, so I'm ok.

brevilo:
Thanks!

PR #3496, opened 2020-03-10
Update URLs of projects to use HTTPS

illia-v:
I've started using Science United in my BOINC client recently and noticed that some projects still use HTTP, although they seem to support HTTPS.
I checked URLs of that projects and updated them to use HTTPS where it is possible.

There are a couple of projects which does not support HTTPS:
* Enigma@Home;
* Gerasim@Home;
* SRBase;
* CAS@home;
* Radioactive@Home.

And there were a couple of URLs that were moved permanently (HTTP 301), I did not update them though.
I remember these two ones:
* https://einstein.phys.uwm.edu/ -> https://einsteinathome.org/uk/home;
* https://denis.usj.es/denisathome/ -> https://denis.usj.es/en_US/.

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/win_build/installerv2/redist/all_projects_list.xml
+++ b/win_build/installerv2/redist/all_projects_list.xml
@@ -3,7 +3,7 @@
     <project>
         <name>DENIS@Home</name>
         <id>4</id>
-        <url>http://denis.usj.es/denisathome/</url>
+        <url>https://denis.usj.es/denisathome/</url>
         <web_url>https://denis.usj.es/denisathome/</web_url>
         <general_area>Biology and Medicine</general_area>
         <specific_area>Medical physiology</specific_area>
@@ -23,8 +23,8 @@
     <project>
         <name>RNA World</name>
         <id>5</id>
-        <url>http://www.rnaworld.de/rnaworld/</url>
-        <web_url>http://www.rnaworld.de/rnaworld/</web_url>
+        <url>https://www.rnaworld.de/rnaworld/</url>
+        <web_url>https://www.rnaworld.de/rnaworld/</web_url>
         <general_area>Biology and Medicine</general_area>
         <specific_area>Molecular biology</specific_area>
         <description><![CDATA[RNA World seeks to identify, analyze, structurally predict and design RNA molecules on the basis of established bioinformatics software.]]></description>
@@ -46,8 +46,8 @@
     <project>
         <name>GPUGrid.net</name>
         <id>6</id>
-        <url>http://www.gpugrid.net/</url>
-        <web_url>http://www.gpugrid.net/</web_url>
+        <url>https://www.gpugrid.net/</url>
+        <web_url>https://www.gpugrid.net/</web_url>
         <general_area>Biology and Medicine</general_area>
         <specific_area>Molecular simulations of proteins</specific_area>
         <description><![CDATA[GPUGrid.net opens novel computational scenarios by the first full-atom molecular dynamics code (CellMD) specially optimized to run on NVIDIA GPUs. New biomedical applications suddenly become possible giving a new role to computational biology for biomedical research.]]></description>
@@ -71,8 +71,8 @@
     <project>
         <name>Rosetta@home</name>
         <id>7</id>
-        <url>http://boinc.bakerlab.org/rosetta/</url>
-        <web_url>http://boinc.bakerlab.org/rosetta/</web_url>
+        <url>https://boinc.bakerlab.org/rosetta/</url>
+        <web_url>https://boinc.bakerlab.org/rosetta/</web_url>
         <general_area>Biology and Medicine</general_area>
         <specific_area>Biology</specific_area>
         <description><![CDATA[Determine the 3-dimensional shapes of proteins in research that may ultimately lead to finding cures for some major human diseases. By running Rosetta@home you will help us speed up and extend our research in ways we couldn't possibly attempt without your help. You will also be helping our efforts at designing new proteins to fight diseases such as HIV, malaria, cancer, and Alzheimer's]]></description>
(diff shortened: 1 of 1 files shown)


RichardHaselgrove:
Did we ever resolve #2642, 'Make attach process the same for Client and Manager'? As I remember it, Account Managers (like Science United) bypass the `get_project_config.php` step, and hence have to use the master url for attaching.

brevilo:
>     * https://einstein.phys.uwm.edu/ -> https://einsteinathome.org/uk/home;

Should this be migrated then please omit any language specific paths and simply use: https://einsteinathome.org/

brevilo:
What was the problem again with changing those URLs? I remember some discussion (over the years) that changing the official project URLs in the BOINC client can cause issues? Which ones? Is that still the case?

**If not**, than the E@H URL can definitely be changed.

AenBleidd:
Original file all_projects_list.xml that is downloaded by client is located outside of this repository. So this commit will change nothing. David should update this file manually on the server

brevilo:
Nevertheless there should be a canonical version of this list, with its changes tracked. Thus I think it's still a good idea to do that here. Heck, the file could even be auto-deployed "on the server".

AenBleidd:
Yes, and it even was discussed once couple of years ago. I think we could raise this question again

RichardHaselgrove:
Perhaps via #2696.

mcandre:
Looks good to me.

Could there be an option for users who wish to explicitly disable HTTPS? This can be helpful in some complicated network environments.

AenBleidd:
@mcandre, take a look at my comment above

brevilo:
Why was this closed? If that's because of f0833f5f3cbd04c4585e2fd11f254f4cf0cd6d23 (PR #3689) I dare to say that it doesn't do what this this PR was about. In terms of HTTPS only Rosetta was updated...

AenBleidd:
@brevilo, because this should be done by @davidpanderson on his side and then backported here. Otherwise we will just have conflicting changes

RichardHaselgrove:
And in that context, "being done" means a proper, comprehensive, plan to ensure compatibility and maintainability between

* Manager, boinccmd, and account managers
* new installations ('winbuild' ??!!)
* 14-day refresh for existing, working, clients.

I liked @fsbruva's comment in #3180 yesterday: "plan the work, work the plan".

brevilo:
> @brevilo, because this should be done by @davidpanderson on his side and then backported here. Otherwise we will just have conflicting changes

Well, it apparently wasn't. The approach to a "conflict" like this shouldn't be to close this PR because it's now conflicting. Since David's PR got merged first, Illia now needs to update his feature branch by merging master, resolving any conflicts locally. Thereby this PR will get the necessary updates to be merge-able again. That's how git is supposed to be used in a fork/PR model.

AenBleidd:
> The approach to a "conflict" like this shouldn't be to close this PR because it's now conflicting.

Now. This PR was closed because even if it will be merged, it will have  no effect because the file that could be accessed from here https://boinc.berkeley.edu/projects.php will be still not changed.

Correct workflow is next: David fixes this on his side, then create a PR with changes. Because only following this workflow you'll have the same file both in git and on web-site.

@davidpanderson, could you please take a look at the changes in this PR and if you find them relevant, please do all necessary changes on website. Thanks in advance

brevilo:
> David fixes this on his side, then create a PR with changes. Because only following this workflow you'll have the same file both in git and on web-site.

I can only strongly repeat myself here: this needs [to change](https://github.com/BOINC/boinc/pull/3496#issuecomment-599480729) or it will continue to confuse people and eventually cause update inconsistencies.

> and if you find them relevant

Frankly, this isn't a question, is it?

RichardHaselgrove:
Having been through exactly this circle when I was acting as release manager for v7.10, I don't think it's as easy as that. Because of all the inconsistencies (project website, project BOINC server, master url, get_project_config.php usage, client and BOINC website updates), I suspect that this particular PR may be insufficiently nuanced - I fell into that trap. It needs a proper **design**, and then implementers prepared to work inside that design.

brevilo:
While you're right, I think this PR clearly shows where the problems are and that informs how they should be fixed. Honestly, I neither thinks that's hard or even complex. All I'm saying is this:

* there needs to be a *canonical* list which is tracked in this repo
* project admins create PRs against that list to update it
* that list is used for client builds *and* https://boinc.berkeley.edu/projects.php

What am I missing?

RichardHaselgrove:
Well, we made some progress during my tenure - also in #2643 - with allowing separate 'web' and 'attach' urls. But this PR specifically mentions Science United in the rationale for the change, and we still haven't implemented get_project_config for account managers.

For your project, https://einsteinathome.org/get_project_config.php returns

`<master_url>http://einstein.phys.uwm.edu/</master_url>`: this PR would conflict with that.

AenBleidd:
> I can only strongly repeat myself here: this needs to change or it will continue to confuse people and eventually cause update inconsistencies.

I completely agree with you: #2696
But for now we have another workflow and if we will not follow it, we will not achieve the result we're expecting for

AenBleidd:
> For your project, https://einsteinathome.org/get_project_config.php returns
`<master_url>http://einstein.phys.uwm.edu/</master_url>`: this PR would conflict with that.

This is why I said 'if relevant'

brevilo:
Sigh, got it. The very reason we don't touch our master URL (not even the damn protocol!) is that we consider BOINC to be too fragile in that context and we don't want to risk unforeseen consequences. I guess that's a chicken-n-egg problem unless BOINC can ensure it's robust against `master_url` changes on the server side.

RichardHaselgrove:
With apologies to @illia-v, this is an example of trying to make a well-intentioned improvement to an ill-designed system.

Get the underlying design right (robust) first: then we can build on that.

illia-v:
> With apologies to @illia-v, this is an example of trying to make a well-intentioned improvement to an ill-designed system.
> 
> Get the underlying design right (robust) first: then we can build on that.

No worries. I am happy that this PR started the conversation

davidpanderson:
Since BOINC is ill-designed, I suggest that you develop your own system for volunteer computing.  The results would be entertaining.  Oh, and do take your time.

PR #3499, opened 2020-03-12
Upgrade from legacy Android Support library to AndroidX library.

Isira-Seneviratne:
Fixes #

**Description of the Change**
Upgrade from the legacy Android support library to the AndroidX library, as the support library is no longer maintained and the AndroidX library is designed to be its replacement.

**Release Notes**
N/A


AenBleidd:
Could you please resolve conflicts here? Then I can make a review

PR #3503, opened 2020-03-13
[Android] Fix WSL compatibility and improvement to script

truboxl:
Fixes #

**Description of the Change**

Path does not parse correctly for `build_all.sh` in WSL. Use double quotes.
Eliminate the need to redownload the NDK and source code for OpenSSL and cURL for every new build.

**Alternate Designs**

Virtualbox VM is slower, heavier, bandwidth intensive and time consuming just to set up one. WSL is much lighter and support is improving with the use of Visual Studio Code. Using Windows version of Android Studio should be possible. WSL needs to install the same dependencies listed in Vagrantfile. WSL also needs python2 to deal with `make_standalone_toolchain.py`.

ATM requires copying of entire repo to a readable path after binary compilation as Android Studio does not allow reading project files over WSL network share. However some files may not be copied properly because of symlink files. Certain case sensitive files (mainly Android toolchain files) will overwrite each other when copied. The app can still be compiled in Android Studio.

Or just map it to a network drive. https://github.com/microsoft/WSL/issues/3854#issuecomment-465886991
However that causes IOExecptionError at Gradle.

**Release Notes**

N/A

### DIFF
--- a/android/buildAndroidBOINC-CI.sh
+++ b/android/buildAndroidBOINC-CI.sh
@@ -112,6 +112,10 @@ if [ "${doclean}" = "yes" ]; then
     echo "cleaning build dir"
     rm -rf "${BUILD_DIR}"
     mkdir -p "${BUILD_DIR}"
+    echo "cleaning downloaded cache files"
+    rm -f /tmp/ndk.zip
+    rm -f /tmp/openssl.tgz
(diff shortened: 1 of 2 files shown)


AenBleidd:
Thanks for the fix

PR #3506, opened 2020-03-16
Convert uses of concatenated string resources to string resources with placeholders.

Isira-Seneviratne:
Fixes #

**Description of the Change**
Convert string resources which are concatenated with other values to string resources with placeholders. This reduces the number of string resources needed in the Android app and allows it to be localized more easily.

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/BOINCActivity.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/BOINCActivity.java
@@ -315,8 +315,8 @@ public class BOINCActivity extends AppCompatActivity {
                     Button returnB = dialog.findViewById(R.id.returnB);
                     TextView tvVersion = dialog.findViewById(R.id.version);
                     try {
-                        tvVersion.setText(getString(R.string.about_version) + " "
-                                          + getPackageManager().getPackageInfo(getPackageName(), 0).versionName);
+                        tvVersion.setText(getString(R.string.about_version,
+                                                    getPackageManager().getPackageInfo(getPackageName(), 0).versionName));
                     }
                     catch(NameNotFoundException e) {
                         if(Logging.WARNING) {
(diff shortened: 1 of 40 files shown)


AenBleidd:
Please remove all values-*/strings.xml files from this commit because localization on this project is made by Transifex, so all changes in localized files will be overwritten.
And don't split commits to every file, they should be splitted logically, not by file.
Thanks in advance :)

Isira-Seneviratne:
> 
> 
> Please remove all values-*/strings.xml files from this commit because localization on this project is made by Transifex, so all changes in localized files will be overwritten.
So I should just leave the code changes?
> And don't split commits to every file, they should be splitted logically, not by file.
I did it because you suggested making the commits atomic.
> Thanks in advance :)
No problem :)

AenBleidd:
> So I should only alter values/strings.xml?

Yes.

> I did it because you suggested making the commits atomic.

Yes, but atomic means that they contain only minimal logically connected changes. E.g., it should be possible to make a cherry-pick of any commit to different branch and everything should work as expected and no unnecessary changes will be merged with main changes.
I hope you understand my point :)

Isira-Seneviratne:
> 
> 
> > So I should only alter values/strings.xml?
> 
> Yes.
> 
> > I did it because you suggested making the commits atomic.
> 
> Yes, but atomic means that they contain only minimal logically connected changes. E.g., it should be possible to make a cherry-pick of any commit to different branch and everything should work as expected and no unnecessary changes will be merged with main changes.
> I hope you understand my point :)

Yeah, thanks!

AenBleidd:
Could you please fix build errors? Just remove translated values that are not used anymore but do not add new values to the translation files. It could be done in one commit.
Sorry for misleading you before .
Thanks in advance

Isira-Seneviratne:
Ah, it's no issue.

PR #3508, opened 2020-03-18
[TravisCI] Adjust script according to updated syntax

AenBleidd:
* deprecated key sudo (The key `sudo` has no effect anymore.)
* missing os, using the default linux
* key matrix is an alias for jobs, using jobs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+os: linux
 compiler:
   - gcc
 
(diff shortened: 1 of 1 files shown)


AenBleidd:
@BOINC/contributors, please review and merge anyone. It is a very minor change in Travis CI configuration file. Basically, as soon as build is green, this means that change is fine :)

PR #3530, opened 2020-03-26
[WIP] track dependencies as git submodules

### LINKED ISSUE #3529: Copy C/C++ library dependency code into boinc source tree
**Describe the problem**

The build system is complex and fragile, expecting contributors to manage dependencies manually. This is especially problematic when any one of the dependent libraries could be offline or accidentally introduce breaking changes.

Copy dependency library source code into the boinc source tree, such as git submodules or ordinary files. Then configure the build system to reference these local files. So the build behaves more like "Just Works" out of the box.

Once this is done, then contributors will be able to onboard onto boinc development much faster!

mcandre:
Fixes #3529

**Description of the Change**

Copy dependencies into boinc source tree as git submodules, in order to automate more of the developer environment setup process.

Pros: Dramatically simplifies and organizes dependency management.

Cons: Some CI systems and network environments have difficulty using git submodules.

TODO:

We still need to configure the build system to look in these directories for the dependencies. The current setup expects the dependency directories to appear alongside the top-level boinc directory, whereas the new structure would nest dependencies inside boinc/vendor/<lib>.

We should also update the build documentation, to reflect the fact that the dependencies are managed using the git submodule system, as opposed to manual file and folder munging.

**Alternate Designs**

* We could use a prominent, cross-platform C/C++ package manager, such as vcpkg.

Pros: Free, open source, reasonably mature.

Cons: Some limited support for more obscure platforms such as Illumos, Haiku, DragonflyBSD.

* We could also copy the source trees more directly, using archive files and cryptographic checksums to validate contents.

Pros: More resilient to external outages.

Cons: More cumbersome to manage than git submodules. Binary blobs make the source tree larger. Involves a little shell script glue code to validate and extract dependencies during the build process.

**Release Notes**

Track C/C++ library dependencies as git submodules

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,21 @@
+[submodule "vendor/wxWidgets"]
+	path = vendor/wxWidgets
+	url = https://github.com/wxWidgets/wxWidgets.git
+[submodule "vendor/sqlite"]
+	path = vendor/sqlite
+	url = https://github.com/sqlite/sqlite.git
+[submodule "vendor/curl"]
+	path = vendor/curl
+	url = https://github.com/curl/curl.git
+[submodule "vendor/c-ares"]
+	path = vendor/c-ares
+	url = https://github.com/c-ares/c-ares.git
+[submodule "vendor/openssl"]
+	path = vendor/openssl
+	url = https://github.com/openssl/openssl.git
+[submodule "vendor/freetype"]
+	path = vendor/freetype
+	url = https://github.com/aseprite/freetype2.git
+[submodule "vendor/ftgl"]
+	path = vendor/ftgl
+	url = https://github.com/ulrichard/ftgl.git


CharlieFenton:
A simpler solution would be to create scripts based on 3rdParty/buildMacDependencies.sh. This script automatically downloads and builds all the dependencies listed in mac_build/dependencyNames.sh for the Mac and is used for creating dependency cache used by Travis CI.

It calls downloads all the source files and then calls various scripts to build each dependency. Updating to newer versions of dependencies merely involves updating the dependencyNames.sh file.

A similar set of scripts should be created for development and release builds of the Mac and , by extension, for all platforms.

AenBleidd:
Existing scripts with manually set versions on 3rdParty dependencies is fine enough and are not required to be replaced with git submodules

mcandre:
Manually set versions may result in broken packages.

PR #3539, opened 2020-03-28
db_purge: fix truncated mod_time data (field too small for timestamp)

stream1972:
**Description of the Change**

Full timestamp is 19 characters long, so it need a 20-bytes buffer.
This is minor issue - this field is used only by db_purge. It causes db_purge output of this field to be truncated.

**Alternate Designs**

**Release Notes**

N/A

### DIFF
--- a/db/boinc_db_types.h
+++ b/db/boinc_db_types.h
@@ -489,7 +489,7 @@ struct WORKUNIT {
         // without consensus (i.e. WU is nondeterministic)
     char result_template_file[64];
     int priority;
-    char mod_time[16];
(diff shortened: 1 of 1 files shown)


TheAspens:
@stream1972 - the change looks good to me - but I think that the same change should be made to RESULT.mod_time for the same reason.  Can you add that to your pull request as well?  Thanks!

stream1972:
@TheAspens Patch already includes changes for both structures.

TheAspens:
Sigh - my apologies for missing that.  

Looks good.  Merging.

PR #3543, opened 2020-03-29
Added a warning when the manager detects a second copy of itself

lockywolf:
**Description of the Change**
Adds a warning when manager's second instance is running.

**Alternate Designs**
Seriously?

**Release Notes**
"N/A"

### DIFF
--- a/clientgui/BOINCGUIApp.cpp
+++ b/clientgui/BOINCGUIApp.cpp
@@ -411,6 +411,7 @@ bool CBOINCGUIApp::OnInit() {
     // Detect if BOINC Manager is already running, if so, bring it into the
     // foreground and then exit.
     if (DetectDuplicateInstance()) {
+      printf( "Another instance of BOINC Manager already running.\n");
             return false;
     }
 


clientgui/BOINCGUIApp.cpp
```diff
@@ -411,6 +411,7 @@ bool CBOINCGUIApp::OnInit() {
     // Detect if BOINC Manager is already running, if so, bring it into the
     // foreground and then exit.
     if (DetectDuplicateInstance()) {
+      printf( "\nAnother instance of BOINC Manager already running.");
```
AenBleidd: 1 - There should be no newline at the beginning of message, but should be added to the end instead.
2 - Since Manager is a GUI application, I believe, it's reasonable to show GUI message, do you agree @lockywolf?

.gitignore
```diff
@@ -216,3 +216,7 @@ vda/vdad
 
 # OS-specific files
 .DS_Store
+
```
AenBleidd: These changes are not related to this PR. Please remove them

clientgui/BOINCGUIApp.cpp
lockywolf: 1 - Fixed.
2 - Well, sure it would be nice, but I found this issue while debugging why the gui manager doesn't start on my machine, and this is the intended application of this commit. A gui message would have a different scope (as in "I started the manager and forgot about it"). I don't see how these two things exclude each other. 
3 - I also removed the emacs-related entries in .gitignore.

AenBleidd: 2 - You're definitely right. But if console message works for you, this doesn't mean that it works for everyone. E.g. Windows users never run GUI from command line. And since our application is cross platform, it makes sense to make it as general as possible for everyone including non-experiences users on non-linux OSs. 

And another question that is not in the scope of this ticket but closely related to it: why BOINC Manager doesn't become top-most application when second copy is launched. Could you please debug it? I checked on Windows, and it works as designed.

Thanks

lockywolf: Well, I don't see how providing a console message on Windows would harm the users. Windows user _may_ launch an app from PowerShell, and see the message, and if she doesn't, it doesn't matter. 

>>why BOINC Manager doesn't become top-most application when second copy is launched

Because there is no second copy. The "multiple copies detector" is malfunctioning.

RichardHaselgrove: I've seen occasional problems through the development of Windows 10, where the initial copy is launched and resides - hidden, invisible, unnoticed - in the notification area. The 'restore to screen' action after a second copy is launched didn't always work. In testing, please look for boincmgr(.exe) in the task monitor for your operating system, just in case.

AenBleidd: > Well, I don't see how providing a console message on Windows would harm the users. Windows user _may_ launch an app from PowerShell, and see the message, and if she doesn't, it doesn't matter.
> 
> > > why BOINC Manager doesn't become top-most application when second copy is launched
> 
> Because there is no second copy. The "multiple copies detector" is malfunctioning.

I didn't say it would harm anyone. I said that it is not actually obvious for anyone to detect problems running GUI application from command line.

AenBleidd: And if there was no second copy, then it's better to solve the original issue with not correct function that should detect the second copy. Otherwise, your message in console will confuse the people who actually checked that there is no other copy, but there is said that it is.

lockywolf: >>Otherwise, your message in console will confuse the people who actually checked that there is no other copy, but there is said that it is.

That is the main point. If the app says that there is a second copy, but there is no second copy, it means that further debugging is required.  That's exactly why this message is useful. Silently quitting implies that "everything is working as expected", so the user would never even understand that this is not the expected behaviour.

>>not actually obvious for anyone to detect problems running GUI application from command line

It's hard for me to imagine what can be less obvious than an app which produces zero feedback whatsoever. By having a console message a user has at least a small but non-zero chance of finding a problem without spending a weekend in GDB.

AenBleidd: > It's hard for me to imagine what can be less obvious than an app which produces zero feedback whatsoever. By having a console message a user has at least a small but non-zero chance of finding a problem without spending a weekend in GDB.

User will be confused that Manager tells about the second instance that is not exists.

You tried to fix aftermath instead of the reason and thus make user to confuse. There is not so much people that could debug an application in any way, many of them just believe what they see.
So if it's told that there is another instance launched, they will try to find it.
If you want this to be `debug` message in some way, then you need to change the message, if you don't want to fix the original issue, of course

lockywolf: Well, suggestions welcome on how to make the message clearer. It could be something like 

>Warning: I detected a second instance of boinc manager gui already running, please use it. If this is not the case, you need to debug the issue. Exiting.

And since the assumption is that Windows users are unlikely to ever see this message, and that UNIX users are usually fine with seeing warnings, this may be better.

RichardHaselgrove: Is the problem better described as 'False detection - no second copy', or 'Real detection - failed to bring to foreground'?

lockywolf: >>Is the problem better described as 'False detection - no second copy', or 'Real detection - failed to bring to foreground'?
Could be any of those. This not about fixing the ultimate problem, just about not hiding something malfunctioning under the carpet. I can write these two cases in the warning message.

PR #3547, opened 2020-03-31
Use network callback implementation to determine network connectivity on Android Lollipop and higher.

Isira-Seneviratne:
Fixes #

**Description of the Change**
The NetworkInfo class that is currently used in the Android app, together with its associated methods (e.g. getActiveNetworkInfo() in ConnectivityManager), have been deprecated starting with API level 29 (Android Q), so this provides a supported mechanism of detecting network changes that works on API level 21 (Lollipop) and higher. The deprecated version is still used on API levels 19 and 20 (KitKat).

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/ManualUrlInputFragment.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/ManualUrlInputFragment.java
@@ -20,14 +20,21 @@
 package edu.berkeley.boinc.attach;
 
 import edu.berkeley.boinc.R;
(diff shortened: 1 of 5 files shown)


adamradocz:
If the `getActiveNetwork` is a better approach, you can use the deprecated version on API levels 19-22, and use the new version from API level 23.

Isira-Seneviratne:
> 
> 
> If the `getActiveNetwork` is a better approach, you can use the deprecated version on API levels 19-22, and use the new version from API level 23.

The recommended approach in the Javadoc for NetworkInfo was to override the NetworkCallback class.

PR #3548, opened 2020-03-31
Use getActiveNetwork() to retrieve current network on Android Marshmallow and later.

Isira-Seneviratne:
Fixes #

**Description of the Change**
The NetworkInfo class that is currently used in the Android app, together with its associated methods (e.g. getActiveNetworkInfo() in ConnectivityManager), have been deprecated starting with API level 29 (Android Q), so this provides a supported mechanism of detecting network changes that works on API level 23 (Marshmallow) and higher. The deprecated version is still used on API levels 19 through 22 (KitKat and Lollipop).

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/ManualUrlInputFragment.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/ManualUrlInputFragment.java
@@ -22,11 +22,13 @@ package edu.berkeley.boinc.attach;
 import edu.berkeley.boinc.R;
 import edu.berkeley.boinc.utils.*;
 
+import android.app.Activity;
 import android.app.Dialog;
 import android.content.Context;
 import android.content.Intent;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
+import android.os.Build;
 import android.os.Bundle;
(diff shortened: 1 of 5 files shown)


android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/ManualUrlInputFragment.java
```diff
@@ -82,11 +84,21 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
...
     // note: available internet does not imply connection to project server
     // is possible!
-    private Boolean checkDeviceOnline() {
-        ConnectivityManager connectivityManager =
-                (ConnectivityManager) getActivity().getSystemService(Context.CONNECTIVITY_SERVICE);
-        NetworkInfo activeNetworkInfo = connectivityManager.getActiveNetworkInfo();
-        Boolean online = activeNetworkInfo != null && activeNetworkInfo.isConnectedOrConnecting();
+    private boolean checkDeviceOnline() {
+        final Activity activity = getActivity();
+        assert activity != null;
```
AenBleidd, line 89: Isn't it necessary to catch this AssertionError somewhere outside?

PR #3553, opened 2020-04-03
web: fix warning when user has no projects yet

davidpanderson:


TheAspens:
Looks good - thanks for the fix.

PR #3554, opened 2020-04-04
Rewrite MessagesParser to extend BaseParser.

Isira-Seneviratne:
**Description of the Change**
Rewrite the MessagesParser class to extend the BaseParser class and use its characters() method implementation so that it matches the other parser classes.

**Release Notes**
N/A


android/BOINC/app/build.gradle
```diff
@@ -61,11 +61,11 @@ android {
 
     buildTypes {
         release {
-            minifyEnabled false
+            minifyEnabled true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.txt'
         }
         debug {
-            minifyEnabled false
+            minifyEnabled true
```
AenBleidd, line 68: I think this should be disabled for debug

android/BOINC/app/build.gradle
```diff
@@ -61,11 +61,11 @@ android {
 
     buildTypes {
         release {
-            minifyEnabled false
```
AenBleidd, line 64: This change doesn't relate to this PR. Please move to a separate pull request

android/BOINC/build.gradle
```diff
@@ -10,7 +10,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.6.1'
```
AenBleidd, line 13: This change doesn't relate to this PR either. Please move to a separate one. Thanks

PR #3568, opened 2020-04-06
Convert remaining RPC data classes to Kotlin.

Isira-Seneviratne:
Fixes #

**Description of the Change**
* Convert any unconverted RPC data classes to Kotlin.
* Convert some data classes used in the adapters to Kotlin.
* Remove Lombok from the dependencies list, as it is no longer required.

**Release Notes**
N/A


android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/DeviceStatusData.kt
```diff
@@ -16,14 +16,14 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
  */
-package edu.berkeley.boinc.rpc;
```
AenBleidd, line 19: Change a copyright year please

PR #3573, opened 2020-04-06
Drupal: Add CKEditor module and library

drshawnkwang:
Part of https://dev.gridrepublic.org/browse/DBOINCP-523

**Description of the Change**
Replace tinyMCE with CKEditor module and library.

**Release Notes**
N/A


AenBleidd:
@drshawnkwang, is it possible to use [git submodules](https://git-scm.com/docs/git-submodule) instead of copy-paste updated 3rdparty code?

brevilo:
submodules: I agree with Shawn here. Technically it's of course possible but right now I don't think it would be worth the effort.

PR #3578, opened 2020-04-07
Add extra freetype lib only when required

smoe:
The BOINC manager needs to deal with fonts, but nothing else, really.  To always include it may have unexpected side-effects, as reported for the CLI-only build for OpenWrt: https://github.com/openwrt/packages/pull/11768#issuecomment-610513948

**Description of the Change**
BOINC_EXTRA_LIBS are grasping all libraries of a list manually aggregated in configure.ac. Since the freetype library is only dealing with fonts, and fonts are used only the GUI, these should not be linked against if only building the command-line interface.

Admittedly, I am not really liking this patch but to do more may require a bit of an extra discussion. The BOINC_EXTRA_LIBS shall possibly be divided up - one set of libs for the command line, one for the GUI, and one that the two share.

**Release Notes**
N/A

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -506,8 +506,10 @@ SAH_CHECK_LIB([dl], [dlopen],
     [BOINC_EXTRA_LIBS="${BOINC_EXTRA_LIBS} ${sah_lib_last}"])
 SAH_CHECK_LIB([nsl], [gethostbyname],
     [BOINC_EXTRA_LIBS="${BOINC_EXTRA_LIBS} ${sah_lib_last}"])
+if test "${enable_manager}" = yes ; then
 SAH_CHECK_LIB([freetype], [fopen],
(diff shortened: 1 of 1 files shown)


CharlieFenton:
For platforms that have a BOINC screensaver I believe that the default BOINC screensaver boincscr also uses freetype.

AenBleidd:
On Windows these 3 projects are linking with freetype. And there is no Manager among them:
image

I believe in linux situation should be pretty much the same

smoe:
If not building the manager, the compilation works. I cannot tell what configure options trigger the building of these "slideshow" or "graphics" apps.

smoe:
I just saw the "Waiting for Response"-flag. That is not a response from me, is it?

AenBleidd:
> I just saw the "Waiting for Response"-flag. That is not a response from me, is it?

It's exactly for you. You fix is not correct in this PR

smoe:
Shall we close this PR, then?

PR #3584, opened 2020-04-09
Web: fix some PHP warnings

davidpanderson:
There were lots of PHP warnings in the apache error log.
Fix most of these.


TheAspens:
Looks good.  Thanks for the cleanup.

PR #3600, opened 2020-04-14
Add check for "-L/..." in "$SSL_LIBS"

### LINKED ISSUE #3591: libcurl link test may fail with "cannot find -lssl"
**Describe the bug**
See https://github.com/BOINC/boinc/blob/master/m4/libcurl.m4#L109

              if test "x`echo \""$LIBCURL"\" | grep ssl`" = x ; then

It checks for `curl-config --features` which outputs `SSL`
Unfortunately `grep ssl` doesn't take into account capitalise `SSL` which causes `SSL_LIBS` to be not included in libcurl test, causing the linker to have `cannot find -lssl` and `cannot find -lcrypto`

See also https://curl.haxx.se/libcurl/using/curl-config.html

> How do I know if libcurl was built with SSL support?
> 
> $ curl-config --feature | grep SSL

Edit: Corrected the variables. Fun fact `curl-config --feature` and `curl-config --features` have the same output.

truboxl:
Checking for `-lssl` in `$LIBCURL` may not be enough to find where is the SSL lib
The test may fail with `cannot find -lssl` and `cannot find -lcrypto`
Add additional check if `$SSL_LIBS` can provide the SSL lib before moving on

Fixes #3591

**Description of the Change**
In short libcurl link test may fail with `cannot find -lssl` and `cannot find -lcrypto`
`$SSL_LIBS` may have the proper `-L/...` so we add check for that

Note that it is impossible to fail `-lssl` test before and after this patch, the end result is libcurl assumes to have SSL support anyway.

**Alternate Designs**
If there's other ways, please comment.

**Release Notes**
N/A

### DIFF
--- a/m4/libcurl.m4
+++ b/m4/libcurl.m4
@@ -110,6 +110,11 @@ AC_DEFUN([LIBCURL_CHECK_CONFIG],
                 LIBCURL="${LIBCURL} ${SSL_LIBS}"
               fi
 
+              # fix cannot find -lssl
+              if test "x`echo \""$SSL_LIBS"\" | grep '\-L/'`" != 'x' ; then
+                LIBCURL="${LIBCURL} ${SSL_LIBS}"
+              fi
+
               # This is so silly, but Apple actually has a bug in their
 	      # curl-config script.  Fixed in Tiger, but there are still
 	      # lots of Panther installs around.


AenBleidd:
Closed and reopened to retrigger CI

PR #3601, opened 2020-04-14
[AppVeyor] Fix Windows CI build

AenBleidd:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -62,7 +62,7 @@ environment:
   matrix:
     - TOOLCHAIN_VERSION: 12.0
       solution_name: win_build\boinc_vs2013.sln
-      depends_git_path: https://boinc.berkeley.edu/git/boinc_depends_win_vs2013.git
+      depends_zip_path: https://boinc.berkeley.edu/dl/boinc_depends/boinc_depends_win_vs2013.zip
       depends_path: C:\projects\boinc_depends_win_vs2013
   BINTRAY_API_KEY:
     secure: kZI9k0Kh2bFSCbXfkz+J16fGNAee1ToRMl10D8QPQsKpC2PqhF/uVMpd6gRC+OSI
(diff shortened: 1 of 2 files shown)


AenBleidd:
@BOINC/contributors please review and merge

RichardHaselgrove:
I though I'd left a comment about updating the trac documentation, but it hasn't appeared. @davidpanderson has updated https://boinc.berkeley.edu/trac/wiki/AdminDepLibs, but there's a dangling link for "All dependency stores can be found here." in https://boinc.berkeley.edu/trac/wiki/CompileClient#Windows (points to https://boinc.berkeley.edu/gitweb/, gives 404 error).

This PR has the feel of a temporary kludge to get us running again - no bad thing - but we've had an unresolved conversation about dependencies running for a while.

AenBleidd:
I made a transition to VS2019 with vcpkg for 3rdparty dependencies. While it is not tested, I didn't add it s a default way of building BOINC for Windows.
So yes, this is currently a temporary fix, but soon it will be changed to a newer one

RichardHaselgrove:
I have tested the new dependency download with VS2013 - and it works, at least for the basic client and manager.

It's not just a move, but a significant update for files like libeay32.dll, which was causing problems. We ought to notify developers about that.

adamradocz:
Maybe I'm too radical, but really should we support older IDEs then the current one?
If it builds on older IDE it's a plus, but IMO it shouldn't be a requirement.

RichardHaselgrove:
It looks like the Windows v7.16.5 I downloaded and installed on 3 March was compiled with VS2010. I like to keep my private builds for testing as close as possible to the final release candidates, but I think I missed the boat on VS2010 downloads when I set up my current build machine.

AenBleidd:
@RichardHaselgrove, since it is just a CI build that is not releasable for now - no need to notify anyone.

@adamradocz, VS2019 build is not tested. Actually, it contains some issues that should be fixed first before complete move to VS2019. Because we have no integration tests and unit tests (ok, we have couple of unit tests but this is definitely not enough) we can't just drop old IDE build support.

RichardHaselgrove:
@AenBleidd  - That wasn't my point. The VS2013 build process is quite well documented in trac, and I'm sure I'm far from the only user to have downloaded the old dependencies, to enable home building and testing. Any of the PRs I've submitted have been tested that way first, and I sometimes run a special build I've made with (or without) particular features for months, until there's something else to test. New new dependencies are relevant to all private builders, not just CI.

PR #3615, opened 2020-04-18
[Android] Add missing annotations.

Isira-Seneviratne:
**Description of the Change**
Add missing annotations that help with documentation or are used in superclass methods, such as Override and NonNull.

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/BOINCActivity.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/BOINCActivity.java
@@ -1,7 +1,7 @@
 /*
  * This file is part of BOINC.
  * http://boinc.berkeley.edu
- * Copyright (C) 2012 University of California
+ * Copyright (C) 2020 University of California
  *
  * BOINC is free software; you can redistribute it and/or modify it
  * under the terms of the GNU Lesser General Public License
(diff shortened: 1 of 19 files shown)


android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/SelectionListActivity.java
```diff
@@ -31,12 +31,18 @@
 import android.os.Bundle;
 import android.os.IBinder;
 import android.os.RemoteException;
-import androidx.fragment.app.FragmentActivity;
```
AenBleidd, line 34: Please update copyright year

android/BOINC/app/src/main/java/edu/berkeley/boinc/BOINCActivity.java
```diff
@@ -508,7 +511,7 @@ protected Boolean doInBackground(Integer... params) {
 
         @Override
         protected void onPostExecute(Boolean success) {
-            if(success) {
+            if(Boolean.TRUE.equals(success)) {
```
AenBleidd, line 514: Is there a better way to compare Boolean value? Maybe `Boolean` type could be changed to `boolean`?

Isira-Seneviratne: Java's generics system doesn't support the primitive types yet:

[JEP 218](http://openjdk.java.net/jeps/218)

Issue #3587, opened 2020-04-11
Feature request: Change device name on windows client. 


talregev:
**Describe the problem**
A clear and concise description of the problem you're looking to solve.
-->Cannot change or see device name on windows client. (maybe also on mac and linux client).

**Describe the solution you'd like**
-->Please add an option in windows client to change device name, like in android client.

**Additional context**
I am using windows 10 64bit
7.16.5(x64) boinc version.

adamradocz:
On PC BOINC uses the PC name.

talregev:
Why? 
Can you change that? It annoying.

talregev:
> On PC BOINC uses the PC name.

By default it will take computer name, as it doing for years.
But also let user switch the name like you let it in android client. 
Thank you for your help.

adamradocz:
Why do you want a different name than your PC's name? It identifies clearly. If it doesn't, probably you should change the PC name.

Ageless93:
Go to Settings > System > About and select the Rename this PC button.

talregev:
> Why do you want a different name than your PC's name? It identifies clearly. If it doesn't, probably you should change the PC name.

Why you argue with the user? I am using this client on my computer, valuntery give computing time. I see great feature you implemented in android client. I understand all the system already can handle this feature, and it simple task to ask. 
It really annoying me that I cannot give cool name to my pc name. 
No I won't change computer name. It for other things, and also require restart. So?
Do you really want to keep argue about and sting me about this? (Yes I know how to change computer name).
I hope other users will see my issue and vote up for this.

RichardHaselgrove:
If you know your own computer name, that should be enough for identification on a project's website.

Would you really find it significantly cooler if you (and you alone) could see a special BOINC computer name on the website? Computer names are not displayed to other users.

talregev:
> Computer names are not displayed to other users.

It for me only. That what I want. like android device.

talregev:
I am a programer myself. I can help in coding. really. just don't sting me. And help me to help you to do better, cooler and nicer computing distribute client.

Ageless93:
BOINC is open source. If you know how to add the code, you can download the source code, change the code to your liking, compile it and run it that way.

Ageless93:
And before you ask where to get the source code: https://boinc.berkeley.edu/trac/wiki/SourceCodeGit
And how to compile it: https://boinc.berkeley.edu/trac/wiki/CompileClient

talregev:
[...]

PR #3618, opened 2020-04-19
[WIP]Add Device Name to advanced computation option tab.

talregev:
Add Device Name Gui.
TODO:
Add device name to asyncRPC.
#3587


AenBleidd:
Another approach is implemented in #3620.
Closing this as not necessary.
As it was already discussed, UI setting is not necessary for this.

PR #3620, opened 2020-04-20
client: allow specifying device name in cc_config.xml

### LINKED ISSUE #3587: Feature request: Change device name on windows client.

davidpanderson:
If cc_config.xml contains <device_name>x</device_name>,
report that to projects rather than the name returned by gethostname()

Fixes #3587

### DIFF
--- a/client/hostinfo_network.cpp
+++ b/client/hostinfo_network.cpp
@@ -96,6 +96,9 @@ int HOST_INFO::get_local_network_info() {
         inet_ntop(AF_INET6, (void*)(&sin->sin6_addr), ip_addr, 256);
     }
 #endif
+    if (!cc_config.device_name.empty()) {
+        strncpy(domain_name, cc_config.device_name.c_str(), sizeof(domain_name)-1);
+    }
     return 0;
 }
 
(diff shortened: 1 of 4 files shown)


adamradocz:
Closes #3620.

client/hostinfo_network.cpp
```diff
@@ -96,6 +96,9 @@ int HOST_INFO::get_local_network_info() {
         inet_ntop(AF_INET6, (void*)(&sin->sin6_addr), ip_addr, 256);
     }
 #endif
+    if (!cc_config.device_name.empty()) {
+        strncpy(domain_name, cc_config.device_name.c_str(), sizeof(domain_name)-1);
```
AenBleidd, line 100: Maybe it's better to use `safe_strcpy` instead?

davidpanderson: Indeed it is.

AenBleidd:
Looks nice for me. Thanks!

PR #3628, opened 2020-04-22
Use new TaskDescription constructor on Android Pie and later.

Isira-Seneviratne:
**Description of the Change**
Use the new ActivityManager.TaskDescription constructor, which accepts a String and a resource ID, on Android Pie and later as the older constructor is now deprecated.

**Release Notes**
N/A


adamradocz:
@Isira-Seneviratne, you're on fire. :) Thanks for polishing the Android Boinc.

PR #3631, opened 2020-04-23
Mac: Improved screensaver logic for OS 10.15 Catalina

CharlieFenton:
Fixes #

**Description of the Change**
Significantly cleaner implementation of BOINC screensaver for Mac OS 10.15 Catalina than my original #3369

**Alternate Designs**
The previous implementation required installing a LaunchAgent for each user. The BOINC Screensaver Coordinator would write a data file to trigger the LaunchAgent. The LaunchAgent launched _gfx_switcher_. This data file contained information on which project graphics app to run. It was read by _gfx_switcher_ which then launched the requested project graphics app, deleted the data file, and wrote another data file with the graphic app's process ID. The Screensaver Coordinator read the process ID from the file. When the graphics app exited, _gfx_switcher_ deleted the process ID file to signal the Screensaver Coordinator. 

Please see my comments in #3369 for more details on the problems adapting BOINC's screensaver to run under OS 10.15 Catalina, and a fuller explanation of the method employed.

This new implementation eliminates both the need for a LaunchAgent and the rather ugly former method of communicating between the modules by writing and deleting disk files.

**Release Notes**
A much cleaner and more efficient implementation of BOINC's screensaver support on Mac OS 10.15.

### DIFF
--- a/client/boinc_cmd.cpp
+++ b/client/boinc_cmd.cpp
@@ -96,7 +96,8 @@ Commands:\n\
  --run_graphics_app id op         run, test or stop graphics app\n\
    op = run | runfullscreen | stop | test\n\
    id = slot # for run or runfullscreen, process ID for stop or test\n\
-   --set_gpu_mode mode duration       set GPU run mode for given duration\n\
+   id = -1 for default screensaver (boincscr)\n\
+ --set_gpu_mode mode duration       set GPU run mode for given duration\n\
    mode = always | auto | never\n\
  --set_host_info product_name\n\
  --set_network_mode mode duration   set network mode for given duration\n\
@@ -176,6 +177,8 @@ int main(int argc, char** argv) {
 
 #ifdef _WIN32
     chdir_to_data_dir();
+#elif defined(__APPLE__)
+    chdir("/Library/Application Support/BOINC Data");
 #endif
     safe_strcpy(passwd_buf, "");
     read_gui_rpc_password(passwd_buf);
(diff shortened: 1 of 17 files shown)


CharlieFenton:
I accidentally pushed my changes to Master, so I reverted them there. Please merge this PR after it passes the checks.

CharlieFenton:
For more details on the logic flow of this implementation, please see the comments at the start of _clientscr/gfx_switcher_.

CharlieFenton:
I don't understand why Travis-CI is failing linking switcher:
> undefined reference to 'callPosixSpawn(char const*) 

I have no trouble building it, and have confirmed that mac_spawn.cpp is one of the sources in the switcher target in the Xcode project. Is it trying to build using make files? As I have said repeatedly, I do not support building BOINC on the Mac using make files; if people want to build with configure / make then it is their responsibility to maintain the config and make files. I don't know how to do that.

CharlieFenton:
It should be noted that switcher is a Mac-specific executable. It is used only by Mac (Darwin) builds of BOINC.

truboxl:
> It should be noted that switcher is a Mac-specific executable. It is used only by Mac (Darwin) builds of BOINC.

So that's where `switcher` is used.
If true, Linux (or any other OS) should disable making `switcher` binary.
Directly running `switcher` in Ubuntu:
```
$ /usr/lib/boinc-client/switcher
Process creation ((null)) failed: Bad address (errno = -1)
```
No point of packaging for other OS either...

CharlieFenton:
It switches operations from the logged in user's account to the boinc_project account as part of BOINC's security implementation on the Mac, for example to change project files to user and group boinc_project after the client has downloaded them. 

An explanation of how it is used on the Mac can be found [here](http://boinc.berkeley.edu/trac/wiki/SandboxUser) and [here](http://boinc.berkeley.edu/sandbox_design.php). It's use is associated with (enabled by) the global variable _g_use_sandbox_, and is invoked by code in _client/sandbox.cpp_. 

I am not aware of Linux implementations setting _g_use_sandbox_ to true or using _switcher_. But since I have primarily focused on the Mac implementation of BOINC, with some work on the MS Windows version, some Linux BOINC developers may have used it without my knowledge.

PR #3633, opened 2020-04-23
LIB: Fix UB when printf-like format specifiers don't match argument types

dimhotepus:
**Description of the Change**
Using format specifier that doesn't match argument type in `printf` functions family is undefined behavior (UB). 
* Change format specifiers to the correct ones for passed into `printf` family arguments.
* Extend printed 64 bit values to 16 hex digits width with leading zeros.
* Print correct, not truncated to 4 bytes values of crash info params and process dumps on x86-64.

**Release Notes**
* Windows crash info's now contain correct values of cpu registries, stack params and addresses involved for x86-64.
* Windows process information dumps now contain correct values of I/O and memory usage counters for x86-64.

### DIFF
--- a/lib/boinc_win.h
+++ b/lib/boinc_win.h
@@ -277,6 +277,7 @@ typedef LPCSTR PCTSTR, LPCTSTR, PCUTSTR, LPCUTSTR;
 #include <cassert>
 #include <cctype>
 #include <cerrno>
+#include <cinttypes>
 #include <cmath>
 #include <csetjmp>
 #include <csignal>
@@ -291,6 +292,7 @@ typedef LPCSTR PCTSTR, LPCTSTR, PCUTSTR, LPCUTSTR;
 #include <assert.h>
 #include <ctype.h>
 #include <errno.h>
+#include <inttypes.h>
 #include <math.h>
 #include <setjmp.h>
 #include <signal.h>
--- a/lib/diagnostics.cpp
+++ b/lib/diagnostics.cpp
@@ -122,7 +122,7 @@ void boinc_catch_signal_invalid_parameter(
 ) {
 	fprintf(
 		stderr,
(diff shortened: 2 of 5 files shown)


dimhotepus:
May be we should update appveyor build scripts to use vs 2010 image and tools?

Now build is performed with vs 2013 image.

AenBleidd:
This PR is failed to build with VS2010:
https://ci.appveyor.com/project/BOINC/boinc/builds/33474716/job/h72l18pvylw2w4c6

davidpanderson:
Let's not do this.

dimhotepus:
Sorry, was too busy last month. 

What i'm going to do is to define portable `size_t` `printf` format specifier between main compilers (looking at you, MinGW) this weekend. 

It's kind of not so easy job to do, so if you against this pull request, please, tell me, ok?
@AenBleidd, @davidpanderson

AenBleidd:
Personally I'm not against it. But there is a way to use more common format specifiers for almost every type

davidpanderson:
Would this fix any bugs?  If not, let's not do it.

dimhotepus:
> Personally I'm not against it. But there is a way to use more common format specifiers for almost every type

Cast values to 64 bit upper equivalent (`size_t` => `unsigned long long` | `%llu`) and use related format specifiers?
Unfortunately, MinGW is still a problem here, need either to define `__USE_MINGW_ANSI_STDIO 1` (intrusive, can't ensure smth isn't broke) or still define portable format specifier for 64 bit upper equivalent type.

dimhotepus:
> Would this fix any bugs? If not, let's not do it.

Windows crash info contains stripped values of cpu registries, stack params and addresses involved for x86-64, also process information dumps can contain incorrect values of I/O and memory usage counters for x86-64. 

Not a critical bugs, if we can live with them.

AenBleidd:
The same changes were done in a different way in another PR. Closing this.

PR #3637, opened 2020-04-24
ClientGui: Fix potential buffer overflow in lstrcpyn

dimhotepus:
**Description of the Change**
lstrcpyn takes size of buffer in *characters*, not in bytes as last argument. If Unicode version of the function is used, `sizeof(buffer)` returns twice as much buffer size than expected by the function. And it may lead to out of buffer write.

See https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-lstrcpynw#security-warning: 
"If the buffer pointed to by lpString1 is not large enough to contain the copied string, a buffer overrun can occur. When copying an entire string, note that sizeof returns the number of bytes. For example, if lpString1 points to a buffer szString1 which is declared as TCHAR szString[100], then sizeof(szString1) gives the size of the buffer in bytes rather than WCHAR, which could lead to a buffer overflow for the Unicode version of the function.
...
Using sizeof(szString1)/sizeof(szString1[0]) gives the proper size of the buffer."

Even more, "If the function fails, the return value is `NULL` and lpString1 may not be null-terminated.".

**Release Notes**
Fixed potential buffer overflow in Taskbar icon tip and balloons.

### DIFF
--- a/clientgui/msw/taskbarex.cpp
+++ b/clientgui/msw/taskbarex.cpp
@@ -131,7 +131,7 @@ bool wxTaskBarIconEx::SetIcon(const wxIcon& icon, const wxString& message)
 
     if (!message.empty()) {
         notifyData.uFlags       |= NIF_TIP;
-        lstrcpyn(notifyData.szTip, message.c_str(), sizeof(notifyData.szTip));
+        lstrcpyn(notifyData.szTip, message.c_str(), ARRAYSIZE(notifyData.szTip));
     }
 
     UpdateIcon();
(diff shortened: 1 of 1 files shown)


AenBleidd:
Is this MS VS 2010 compatible?

dimhotepus:
@aenbleidd Yes, it is

AenBleidd:
This PR looks ok for me.
@davidpanderson, please merge it if you're ok with it too.

davidpanderson:
Thanks.

PR #3642, opened 2020-04-25
[Android] Fix build warnings.

Isira-Seneviratne:
**Description of the Change**
Fix build warnings by adding the generic parameter to the `ArrayAdapter` declaration in `AcctMgrFragment` and disabling R8 in `gradle.properties`.

**Release Notes**
N/A


android/BOINC/app/proguard-rules.txt
```diff
@@ -23,3 +23,6 @@
 # Useful for debugging
 -printusage build/usage.txt     # Prints unused APIs removed by ProGuard
 -printmapping build/mapping.txt # Prints mappings of class names to their obfuscated names
+
+# Hides warning about missing class java.lang.instrument.ClassFileTransformer
+-dontwarn java.lang.instrument.ClassFileTransformer
```
AenBleidd, line 28: Fix warning and hide warning is not the same. I'd prefer to have it in a log if it can't be fixed for some reason

Isira-Seneviratne: I couldn't find a cause for that particular issue.

However, I found the way to actually fix it.

PR #3643, opened 2020-04-25
client: initialize log flags to tasks, sched_ops, file_xfer

davidpanderson:


RichardHaselgrove:
I have tested this: it compiles under VS2010, and it compiles and produces the correct run-time outcome under VS2013. I approve the 62222e20c6bc09023a8fd478748cb42e04409ed7 commit for merging, but I'm nervous about the second commit.

PR #3646, opened 2020-04-26
[Android] Replace PNGs with vector drawables.

Isira-Seneviratne:
**Description of the Change**
Replace the PNGs currently used for icons with similar vector drawables and SVGs. This eliminates the need for separate resource sets with each resolution type.

**Release Notes**
Replace the icons used in the app with Material icons.


adamradocz:
Shouldn't the file type be SVG instead of XML?

Isira-Seneviratne:
On Sun, Apr 26, 2020, 8:06 PM Adam Radocz <notifications@github.com> wrote:

> Shouldn't the file type be SVG instead of XML?
>

Android Studio requires SVGs to be imported as vector drawables.

adamradocz:
I see Android Studio uses XML file to define the vector drawable.

AenBleidd:
@Isira-Seneviratne, please fix merge conflict

PR #3651, opened 2020-04-27
added install target, $(CXX)

smoe:
I just had a look around - please take this PR just as an anchor to discuss what should happen in general with "make install" when components are configured that are more assigned to the server side of boinc - like these scientific/sample apps.

### DIFF
--- a/samples/sleeper/Makefile
+++ b/samples/sleeper/Makefile
@@ -4,6 +4,7 @@ BOINC_DIR = ../..
 BOINC_API_DIR = $(BOINC_DIR)/api
 BOINC_LIB_DIR = $(BOINC_DIR)/lib
 
+CXX ?= g++
 CXXFLAGS += -g \
     -I$(BOINC_DIR) \
     -I$(BOINC_LIB_DIR) \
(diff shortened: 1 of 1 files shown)


samples/sleeper/Makefile
```diff
@@ -16,6 +17,9 @@ PROGS = sleeper
 
 all: $(PROGS)
 
+install: all
+	@echo "--- The example app $(PROGS) is not installed ---"
```
AenBleidd, line 21: Such kind of message will mislead the user, (s)he will decide that smth wrong happened during installation and it should be fixed somehow.

AenBleidd:
According to your comment, converting this PR to draft.

smoe:
I have removed the message. Since regular users don't see this, I don't ultimately share your concern. Important to me is only that it compiles - which it does not without the install target.

smoe:
I once went through this for the BOINC server package on Debian, or the [boinc-app-examples](https://packages.debian.org/experimental/boinc-app-examples). I suggest to install to /usr/lib/boinc/apps. Maybe you bring this up at the next boinc contributors meeting.

AenBleidd:
Looks fine, thanks. Merging

PR #3652, opened 2020-04-28
[Android] Add Robolectric tests.

Isira-Seneviratne:
**Description of the Change**
Add tests that use the Robolectric framework, to eliminate the need to set up an Android emulator for testing certain functionality.

**Release Notes**
N/A

**Note**
JaCoCo code coverage does not currently work for Robolectric tests: https://github.com/robolectric/robolectric/issues/3023

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,10 @@ jobs:
             - android-29
             - extra-google-m2repository
             - extra-android-m2repository
+      dist: xenial
       env: BOINC_TYPE=manager-android
+      jdk:
+        - openjdk9
 
 before_install:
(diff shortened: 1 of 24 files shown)


.travis.yml
```diff
@@ -58,7 +58,10 @@ jobs:
             - android-29
             - extra-google-m2repository
             - extra-android-m2repository
+      dist: xenial
```
AenBleidd, line 61: I believe we should try to move all builds to a newer distro. Could you please move this to root?

Isira-Seneviratne: Sure.

Isira-Seneviratne:
This line should be of interest:

`./integration_test/installTestSuite.sh: line 113: ansible-playbook: command not found`

PR #3658, opened 2020-04-29
[Windows] Device name in Windows gui client

### LINKED ISSUE #3656: Feature request: Change device name on windows client via GUI
**Describe the problem**
-->Cannot change or see device name on windows client via GUI. (maybe also on mac and linux client).

-->Please add an option in windows client to change device name in GUI, **like in android client**.

**Additional context**
I am using windows 10 64bit
7.16.5(x64) boinc version. 

Hi boinc community,
I re-open this feature request because I want to make my voice heard.
This is the previous tiket I open: #3587
You are open source code community, but you don't seems to be an open community. 
You are not open to new ideas, to new features, and not for new members.
In the open source community the number of user that active is already low, and from them 
the number the active developer is more small. Please keep them and not let them leave.

Don't change my idea, you can suggest better place to put it in the gui, or improve my code quality. 
I suggest this idea because this issue is very annoying  from me. I want to be able change the device name from the gui like android device. 
I understand that you don't agree with this idea, and it not very important for you, and not all of the users will want it. But I want it, and I already Implement it.
So please keep this issue open that it will show me that you open for my new ideas even you don't agree with them. 

I already implement this feature. Soon I will publish a pull request.
Here a glimpse how it look like: 
In advance view:
advDeviceName

talregev:
Fixes #3656

### DIFF
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -632,6 +632,7 @@ int CLIENT_STATE::init() {
     //
     host_info.p_vm_extensions_disabled = false;
 
+    set_device_name();
     set_ncpus();
     show_host_info();
 
--- a/client/client_state.h
+++ b/client/client_state.h
@@ -347,6 +347,8 @@ struct CLIENT_STATE {
         // - any result op is done via RPC (suspend/resume)
     void set_ncpus();
 
+    void set_device_name();
+
 // --------------- cs_account.cpp:
     int add_project(
         const char* master_url, const char* authenticator,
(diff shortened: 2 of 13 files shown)


truboxl:
Regarding my last https://github.com/BOINC/boinc/issues/3587#issuecomment-616129710, because not all platforms have GUI managers, the client should by itself have the ability to change device name, at least when talking to project servers.

GUI options can still be a necessity on Windows where users don't normally run `boinc.exe` with text editors editing `cc_config.xml` (that file needs to be stored in a hidden ProgramData folder anyway) and instead rely on `boincmgr.exe`. It's worth reconsider this.

talregev:
> It's worth reconsider this.

Thank you.

talregev:
@truboxl can you comment it also in my issue, that more people will see?
#3656

PR #3664, opened 2020-05-01
[Android] Replace StringReader with CharSequenceReader in RpcClient.

Isira-Seneviratne:
**Description of the Change**
Use Apache Commons IO's `CharSequenceReader` instead of `StringReader`, as the former supports all classes that implement the `CharSequence` interface.

**Release Notes**
N/A


adamradocz:
Java works out of the box. I don't see the benefits of moving to Apache Commons IO's.

Isira-Seneviratne:
I thought it might be useful to directly read from the `StringBuilder` without having to convert it to a `String` first.

adamradocz:
Oh, OK. I thought the `toString()` just redundant.

PR #3665, opened 2020-05-01
[Android][ARM] build fix and use NEON in benchmark

### LINKED ISSUE #3657: General improvement for ARM platform
**Describe the problem**
Split off from https://github.com/BOINC/boinc/issues/3596#issuecomment-616013774
Maybe will find some time for PR, here are just some initial plans

Currently benchmark optimisation has been limited to specifically 32bit ARM Android platform. Suggest update the sources to allow all ARM platform (arm and aarch64 / arm64) to benefit from the same optimisation.

https://github.com/BOINC/boinc/commit/7b2ca9e787a204f2a57f390bc7249bb7f9997fea will be a starting point.

EDIT: see linked PR for reference. Note that the linked PR is outdated. aarch64 checking code can be removed since its assumed to have NEON always.

**Additional context**
This commit https://github.com/BOINC/boinc/commit/e3fa4ce5567bf4ef36d62325121f9e962912039e seems to suggest not to use NEON in aarch64. Why though? [All ARMv8 devices (and by extension aarch64) ship with NEON.](https://developer.android.com/ndk/guides/cpu-arm-neon) Yes there are 32bit ARMv8 out there.

Android NDK 21 also compiles with NEON (see above link) and [PIE](https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md#additional-required-arguments) by default. Maybe time to drop some workaround like Cortex-A8 (Galaxy S) as I don't see them capable running Android 4.1.

https://github.com/BOINC/boinc/issues/1702

https://github.com/magnumripper/JohnTheRipper/issues/1998

EDIT: Changed some parts

truboxl:
Fixes #3657

**Description of the Change**
1. client/Makefile.am
Change `OS_ARM_LINUX` to `OS_LINUX_ARM` and add `OS_LINUX_AARCH64`
Add `EXTRA_boinc_client_DEPENDENCIES =` to ensure `libwhet*.a` are built before linking with boinc_client. This fixes the issue with Android NDK on MSYS2 on Windows where generated `boinc/client/Makefile` yields:
```
all-am: Makefile $(PROGRAMS) $(LIBRARIES) all-local
```
instead of the output from a typical Linux environment
```
all-am: Makefile $(LIBRARIES) $(PROGRAMS) all-local
```

2. client/cpu_benchmark.h
Rename parts to suit all ARM platforms

1. client/cs_benchmark.cpp
Rename parts to suit all ARM platforms, add aarch64 part

1. client/whetstone.cpp
Rename parts to suit all ARM platforms, namespace selectively controlled by `-DARM_NEON` and `-DARM_VFP`

1. configure.ac
Change `OS_ARM_LINUX` to `OS_LINUX_ARM` and add `OS_LINUX_AARCH64`
Allow cross compile and native compile

1. m4/boinc_platform.m4
Fix indent and hardcode Android platform string
Sometimes clients appear as `$ARCH-unknown-linux-android*` instead of `$ARCH-android-linux-gnu`. This tries to solve this without having to change the entire BOINC platform name for Android sake and avoid Android clients receiving tasks meant for `$ARCH-unknown-linux-gnu` or `$ARCH-pc-linux-gnu`. Passing `--host` should be sufficient to override now.
https://boinc.berkeley.edu/trac/wiki/BoincPlatforms

**Alternate Designs**
Feedbacks welcome.

**Release Notes**
Client: CPU benchmark now use NEON if available on all ARM platforms (incl. 64bit Android)


RichardHaselgrove:
Could we take a moment the rethink the purpose of this code, please?

The desktop BOINC clients (Linux/Mac/Windows) use Dhrystone and Whetstone benchmark code (dhrystone.cpp and whetstone.cpp). There has been some debate in the user community as to the relevance of these routines in the modern scientific computing world, given the easy availability of optimising hardware (SIMD extensions etc) and optimising compilers.

It's been suggested to me (and I'll try to get a quote) that the current Android benchmarks bear a different relationship to the real-life performance of BOINC project applications from the desktop code. And there's another problem with the speed calculation of GPU applications, where we make no attempt to measure the actual speed of the device, but instead use a rule-of-thumb multiplier of the theoretical peak speed.

a-priori speed estimates play an important role in the linked areas of runtime estimation and cobblestone credit. It's important that we adopt a consistent policy in these areas.

truboxl:
The purpose of this PR is "if I see unused code and there's untapped potential, I will make a PR". Of course, removing them is OK if it doesn't break anything. Generally this is OLD code that has been lying there for years and I am not adding anything to it. I doubt there's any devices still using this to their potential. And even if there's one, the current projects mandates PIE which will render them useless as they can't run it. I also want to get rid of codes to make sure they still can be compiled and run under modern hardware.

I also believe that the desktop counterpart clients are compiled with their respective instructions. To be consistent means they have to drop their optimisations as well to match the current state of un-accelerated mobile platform.

I also think it's unfair to compare the mobile with desktop. Running the same code unoptimised on different sets of hardware and call it one is better than the other makes no sense. It is entirely up to how projects compiled their programs and make them work. At the end of the day, the benchmark is a reference. Whether the projects use it as part of their credit system is up to them.

AenBleidd:
There are a lot of changes. I need to review this first

truboxl:
```
Making all in client
  CXX      libwhetneon_a-whetstone.o
In file included from /usr/include/features.h:448,
                 from /usr/include/arm-linux-gnueabihf/c++/8/bits/os_defines.h:39,
                 from /usr/include/arm-linux-gnueabihf/c++/8/bits/c++config.h:508,
                 from /usr/include/c++/8/cstdlib:41,
                 from /usr/include/c++/8/stdlib.h:36,
                 from whetstone.cpp:29:
/usr/include/arm-linux-gnueabihf/gnu/stubs.h:7:11: fatal error: gnu/stubs-soft.h: No such file or directory
 # include <gnu/stubs-soft.h>
           ^~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [Makefile:897: libwhetneon_a-whetstone.o] Error 1
make[1]: *** [Makefile:649: all-recursive] Error 1
make: *** [Makefile:553: all] Error 2
```
Interesting. Looks like I missed doing testing on actual Raspberry Pi 3B+.
Installing [`libc6-dev-armel-cross`](https://packages.debian.org/search?searchon=contents&keywords=gnu%2Fstubs-soft.h&mode=path&suite=stable&arch=armhf) doesn't solve it as the file lives in
`/usr/arm-linux-gnueabi/include/gnu/stubs-soft.h` different from above.
May have to separate ARM 32bit further with hardfloat (armv8) (RPI) and softfloat (armv7) (really old hardware?). Changing to draft now.

I also want to request others to test this with their ARM devices, Raspberry Pi, ODROID what now so it doesn't break them.

truboxl:
Looks like the real cause of failure to build Android ARM build is not because NDK (Windows) or MSYS2 but Automake 1.16.1.

The new Automake is going to break the build on Linux (eg: Ubuntu 20.04) if the patch is not applied (the `EXTRA_boinc_client_DEPENDENCIES =` part).

truboxl:
`OS_LINUX_ARMHF` is used to determine `arm-unknown-linux-gnueabihf` (Debian family, Raspbian) and `armv7hl-redhat-linux-gnueabi` (Redhat family)

Generally the triplets are assumed to be armv7 on Debian and Redhat but Raspbian use
`arm-unknown-linux-gnueabihf` as armv6. So dropping `-march=armv7-a` for compatibility, however this does mean different performance depending on distro and their choice of `-march`.

32bit ARM Android will always use `OS_LINUX_ARM` for compatibility as it uses
 `arm-linux-androideabi` or `armv7a-linux-androideabi` or some other variant
that can be commonly traced to `arm*linux-android*`

References:
https://wiki.debian.org/RaspberryPi#Raspberry_Pi_issues
https://fedoraproject.org/wiki/Architectures/ARM/Raspberry_Pi#What_about_support_for_the_Raspberry_Pi_Models_A.2FA.2B.2C_B.2FB.2B_.28generation_1.29.2C_Zero.2FZeroW_and_Compute_Module.3F
https://wiki.debian.org/ArmPorts
https://wiki.debian.org/ArmHardFloatPort/VfpComparison
https://lwn.net/Articles/524945/ (who started it all for Redhat)
https://bugzilla.redhat.com/show_bug.cgi?id=1134914#c9
https://bugzilla.redhat.com/show_bug.cgi?id=1720734 (still ongoing)
https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md
https://developer.android.com/ndk/guides/other_build_systems

truboxl:
Ready for another round of review

truboxl:
I am still unhappy how this improvement is limited to Linux based OS. Will rework again down to using the preprocessor macro to identify architecture difference rather than using triplets.

truboxl:
The intention was to have the code be reused where it is no longer useful on its intended platform (ARM Android). However this may open up some unintended use and abuse which maybe in favour of the ARM platform as @RichardHaselgrove suggested. After a long pondering, I think this isn't really my intention so marking this PR as closed.

PR #3667, opened 2020-05-01
Mac fix gfx apps on xcode11

CharlieFenton:
his adds to my previous fixes (PR #3369) for the Macintosh BOINC screensaver running under OS 10.15 Catalina when built using Xcode 11. That PR solved problems with displaying graphics applications in the screensaver on Retina displays, but overlooked the same problem displaying project graphics apps in a window when invoked by the "Show graphics" button in the BOINC Manager.

This PR adds a fix for that issue. It also updates the graphics sample code in samples/example_app.

None of these problems occurred when building with earlier versions of Xcode or on versions of MacOS before OS 10.15 Catalina. These changes, along with my earlier ones, preserve proper operation when built with earlier Xcode or run on earlier versions of MacOS.

OpenGL / GLUT apps built under Xcode 11 apparently use window dimensions based on the number of backing store pixels. That is, they double the window dimensions for Retina displays (which have 2X2 pixels per point.) But OpenGL apps built under earlier versions of Xcode don't. Catalina assumes OpenGL apps work as built under Xcode 11, so it displays older builds at half width and height, unless we compensate in our code.

This PR is part of my attempt to ensure that BOINC graphics apps built on all versions of Xcode work properly on different versions of OS X.

### DIFF
--- a/api/macglutfix.m
+++ b/api/macglutfix.m
@@ -1,6 +1,6 @@
 // Berkeley Open Infrastructure for Network Computing
 // http://boinc.berkeley.edu
-// Copyright (C) 2019 University of California
+// Copyright (C) 2020 University of California
 //
 // This is free software; you can redistribute it and/or
 // modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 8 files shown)


CharlieFenton:
This applies the same fixes to master which PR #3666 applies to the clent_release/7/7.14 branch.

This PR is ready to be merged as soon as all checks are successful.

PR #3672, opened 2020-05-02
Server programs: try to print error message if can't connect to DB

davidpanderson:


db/db_base.cpp
```diff
@@ -85,10 +85,14 @@ int DB_CONN::open(
     // CLIENT_FOUND_ROWS means that the # of affected rows for an update
     // is the # matched by the where, rather than the # actually changed
     //
-    mysql = mysql_real_connect(
+    MYSQL* mysql2 = mysql_real_connect(
```
AenBleidd, line 88: What is the reason of adding new variable here? In case of unsuccessful connect, `mysql` variable will still be uninitialized.

davidpanderson: If the connect failed, mysql will be NULL.  mysql_errno() and mysql_error() require a non-NULL argument.

AenBleidd: But will `mysql` will contain relevant error information then?

davidpanderson: Yes.  It will contain info about the fail connect.

PR #3673, opened 2020-05-02
Prefer curl-config from --with-libcurl

truboxl:
**Description of the Change**
If "--with-libcurl" is passed, prefer using "curl-config" from directory instead of system provided

**Alternate Designs**
Really makes no sense not to do this

**Release Notes**
N/A


AenBleidd:
Looks nice. Thank you.

PR #3676, opened 2020-05-02
Eliminate some compile warnings in vda/.

### LINKED ISSUE #3675: ssim.cpp: Using non-virtual destructor compiler warning
No idea about what our exact class hierarchy looks like but ... could we just have a virtual destructor?
```
ssim.cpp: In member function ‘virtual void SIM_HOST::handle()’:
ssim.cpp:422:16: warning: deleting object of polymorphic class type ‘CHUNK_ON_HOST’ which has non-virtual destructor might cause undefined behavior [-Wdelete-non-virtual-dtor]
  422 |         delete c;
      |                ^
```

davidpanderson:
Added virtual destructors.
Fixed some sprintf warnings but not all; that seems like an infinite rat-hole.
Fixes #3675

### DIFF
--- a/vda/des.h
+++ b/vda/des.h
@@ -33,6 +33,7 @@ using std::vector;
 struct EVENT {
     double t;
     virtual void handle(){}
+    virtual ~EVENT(){}
 };
 
 inline bool compare(EVENT* e1, EVENT* e2) {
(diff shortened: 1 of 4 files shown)


AenBleidd:
I'll take care about Android failure that is not related to this PR.

PR #3677, opened 2020-05-03
[Android] Fix translation build errors.

Isira-Seneviratne:
**Description of the Change**
Fix build errors caused by Slovak strings that are not present in the default strings.xml file.

**Release Notes**
N/A


Isira-Seneviratne:
@davidpanderson you might want to make changes to this

PR #3680, opened 2020-05-04
[Android] Modify app to make use of dependency injection via Dagger.

Isira-Seneviratne:
**Description of the Change**
Make use of [Dagger's](https://dagger.dev/) dependency injection functionality to initialize some variables that would otherwise be initialized using their object constructors. Using dependency injection provides the following advantages:

- Code reusability
- Ease of testing
- Ease of refactoring

Useful information about Dagger can be found [here](https://developer.android.com/training/dependency-injection/dagger-android) and [here](https://codelabs.developers.google.com/codelabs/android-dagger).

**Release Notes**
N/A


AenBleidd:
@Isira-Seneviratne, please fix merge conflicts

PR #3681, opened 2020-05-04
Client: fix build error on FreeBSD (from Jason Darfus)

davidpanderson:


client/client_state.cpp
```diff
@@ -431,7 +431,7 @@ static void set_client_priority() {
 #ifdef __linux__
     char buf[1024];
     snprintf(buf, sizeof(buf), "ionice -c 3 -p %d", getpid());
-    system(buf);
+    if (!system(buf)) {}
```
AenBleidd, line 434: What's the purpose of this change? It basically changes nothing until we want to put smth between curly braces

davidpanderson: It removes a compiler warning

PR #3682, opened 2020-05-04
hostinfo_unix: Enable FreeBSD CPU family/model/stepping detection.

fsbruva:
**Description of the Change**
Use native FreeBSD CPUID tools to derive family, model and stepping info. Tested on FreeBSD 12.1.

**Alternate Designs**
1. dmidecode: GNU tool creates unnecessary external dependency, and has trouble running in some VMs.
2. cpuinfo: FreeBSD has a kernel linux emulation capability, which allows a linprocfs to be mounted, containing cpuinfo. However, the implementation requires particular kernel knobs, and sometimes produces erroneous data (e.g., nonsense flags like 3dnow on Intel CPU).
3. dmesg: CPU data is visible and easily parsed from dmesg, but sane security settings generally prevent access to kernel message buffer from within FreeBSD jail.

**Release Notes**
FreeBSD now detects CPU family, model and stepping data.

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -718,9 +718,11 @@ static void parse_cpuinfo_linux(HOST_INFO& host) {
 #include <sys/types.h>
 #include <sys/cdefs.h>
 #include <machine/cpufunc.h>
+#include <machine/specialreg.h>
 
 void use_cpuid(HOST_INFO& host) {
     u_int p[4];
+    u_int cpu_id;
     int hasMMX, hasSSE, hasSSE2, hasSSE3, has3DNow, has3DNowExt, hasAVX;
     char capabilities[256];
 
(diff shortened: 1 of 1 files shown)


AenBleidd:
@fsbruva, why did you close it?

fsbruva:
@AenBleidd , I realized one PR with 3-4 commits probably made more sense than each commit having its own PR. Or would you rather I 

Scope of work - There are numerous things wrong with the FreeBSD output:

> Processor: 4 amd64 Intel(R) Core(TM) i3-4340 CPU @ 3.60GHzIntel(R) Core(TM) i3-4340 CPU @ 3.60GHz [] [sse sse2 pni mmx avx]
Processor features:

So, after I submitted stepping detection PR, I also fixed the duplicate model information, and moving the feature flags into host.p_features. I'll fix the vendor_id and submit a unified PR. Or would you rather the PR be atomic to the commit?

AenBleidd:
PRs should be atomic. But if you have some general improvements or fixes that are closely connected (I believe it is the case), it's better to combine them in one PR

fsbruva:
The PR would generally be: "Improve FreeBSD CPU detection & reporting format".  I'll submit a single PR, and if it needs to be broken apart, I can accommodate.

PR #3684, opened 2020-05-05
Improve FreeBSD CPU detection & reporting format

fsbruva:
**Description of the Change**
Use native FreeBSD CPUID tools to detect vendor, family, model, and stepping info. Rectify duplicate reporting of CPU model, and use host.p_features for CPU flags instead of host.p_model. Tested on FreeBSD 12.1.

Original output:
> Processor: Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz [Family 6 Model 158 Stepping 13]Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz [Family 6 Model 158 Stepping 13] [] [sse sse2 pni mmx]
Processor features:

Output after applying PR:
>Processor: 4 GenuineIntel Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz [Family 6 Model 158 Stepping 13]
Processor features: sse sse2 pni mmx avx

**Alternate Designs**
1. dmidecode: GNU tool creates unnecessary external dependency, and has trouble running in some VMs.
2. cpuinfo: FreeBSD has a kernel linux emulation capability, which allows a linprocfs (Linux procfs) to be mounted, containing cpuinfo. However, the implementation requires particular kernel knobs, and sometimes produces erroneous data (e.g., nonsense flags like 3dnow on Intel CPU).
3. dmesg: CPU data is visible and easily parsed from dmesg, but sane security settings generally prevent access to kernel message buffer from within FreeBSD jails.

**Release Notes**
FreeBSD better detects and reports CPU vendor, family, model and stepping data.

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -718,9 +718,12 @@ static void parse_cpuinfo_linux(HOST_INFO& host) {
 #include <sys/types.h>
 #include <sys/cdefs.h>
 #include <machine/cpufunc.h>
+#include <machine/specialreg.h>
 
 void use_cpuid(HOST_INFO& host) {
     u_int p[4];
+    u_int cpu_id;
+    char vendor[13];
     int hasMMX, hasSSE, hasSSE2, hasSSE3, has3DNow, has3DNowExt, hasAVX;
     char capabilities[256];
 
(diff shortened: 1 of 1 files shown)


fsbruva:
Just noticed that I have a duplicate declaration of cpu_id. Will fix, and rebase, to avoid git log pollution.

fsbruva:
@AenBleidd I am guessing that this PR requires review by someone other than yourself prior to merging? If FreeBSD build stability is a concern - would there be merit in adding FreeBSD CI tasks to streamline review? Cirrus-CI can do FreeBSD builds, and I could develop a PR to add that.

PR #3685, opened 2020-05-05
Add doc on building BOINC Android on WSL

### LINKED ISSUE #3507: [Android] Building on Windows
**Describe the problem**

With #3503 merged, I can confirm the Android client can be compiled under Windows using WSL. APK can also be built and installed using Windows version of Android Studio. It will be nice to have documentation in using this approach. This should open up more people to try develop for the Android version of BOINC without having to go through fixing VM or Oracle licenses issues.

I am currently using Ubuntu 18.04 as WSL distro to build using the command:

    sudo apt-get update && \
    sudo apt-get install ubuntu-make git automake libtool pkg-config unzip python -y && \
    git clone https://github.com/boinc/boinc.git ~/boinc && \
    cd ~/boinc/android/ && \
    ./build_all.sh

After building, copy the whole BOINC repo to a path Android Studio can read:

    cp -frv ~/boinc /mnt/c/<AnyReadablePath>/boinc

Then run Android Studio and open project file in `C:\<AnyReadablePath>\boinc\android\BOINC` to proceed development for Android version of BOINC or build the APK.

**Additional context**

1. While Vagrant is nice, I think BOINC should leverage the lighter WSL for building and not just running BOINC apps.
2. Don't bother using msys2 / cygwin + NDK (Windows) as it will fail at OpenSSL compilation due to building for the wrong architecture / format using the same set of scripts.
3. While using this approach, Windows Defender Real Time Protection needs to be temporarily disabled to speed up compilation times and fix some weird issues (fork error). Adding process or path to Exclusion whitelist is good too.
4. WSL acts exactly like a UWP app and can be easily Reset so you don't need to uninstall, download, reinstall again to fix problems.
5. apt-get may pause at the Restart Service stage, selecting yes or no, and it will proceed as usual.

truboxl:
Fixes #3507 

**Description of the Change**
Add initial version of doc and appropriate feedback for users compiling BOINC Android on WSL

**Alternate Designs**
After much testing, WSL is still faster than MSYS2 in significant ways. It should complement Vagrant method if users can't accept Oracle Virtualbox Extension licenses. If there's plan to move BOINC to use CMake (follow Linux side of plan), we may revise adding it if it proves faster.

**Release Notes**
N/A


android/WSL.README.md
```diff
@@ -0,0 +1,76 @@
...
+
+    You should see this once you are done compiling:
+
+        ===== BOINC for all platforms build done =====
+
+    If you ever had problem at this point, try `git clean -fxd ..` and run again
+
+1. Copy the main BOINC repo to Android Studio readable path. Replace `<pathtofolder>` with appropriate location.
+
+        $ cp -fr ~/boinc /mnt/c/<pathtofolder>/boinc
```
AenBleidd, line 48: You should warn users here that after moving files to Windows accessible location there will be a mess with line-endings (if they want to continue development on Windows, line-ending should be adjusted manually). As an alternative ,yo can suggest users not to move the whole repository to Windows accessible location but only assets files (from `android\BOINC\app\src\main\assets` folder)

android/WSL.README.md
```diff
@@ -0,0 +1,76 @@
...
+
+Import project from `C:\<pathtofolder>\boinc\android\BOINC`
+
+After Gradle sync completed, you can now start development or compile the APK.
+
+## Tips
+
+1. For compatibility and speed reasons, you may want to temporarily disable Windows Defender Real Time Protection.
+
+        Start Menu > search Virus & Threat Protection > Manage settings > toggle Real-time protection to Off
```
AenBleidd, line 60: I'd never suggest to users to disable any kind of endpoint security applications. It's dangerous for them. And also it creates a feeling that smth is wrong with the application (maybe it's malicious but just tries to mimic to a nive application?).

android/WSL.README.md
```diff
@@ -0,0 +1,76 @@
+# Building BOINC Android client on Windows (WSL)
+
+## Goal
+
+Allow Windows users to use built-in WSL for setting up for BOINC Android development
+
+## Limitations
+
+* Windows 10 only
```
AenBleidd, line 9: - User must be running Windows 10 version 1607 (the Anniversary update) or above.
- WSL will only run on 64-bit versions of Windows 10. 32-bit versions are not supported.
- Also, if I remember correctly, only Windows 10 Pro or similar supports WSL

android/WSL.README.md
```diff
@@ -0,0 +1,76 @@
...
+* Windows Defender may affect compile times (see [Tips#1](#Tips))
+* fork/exec error on WSL (see [Tips#1](#Tips))
+
+## Prerequisites
+
+* Windows 10
```
AenBleidd, line 15: The same here

android/WSL.README.md
truboxl: https://github.com/microsoft/WSL/issues/2469

android/WSL.README.md
truboxl: This could interfere with APK build because it relies on git tags, I see what I can do

android/WSL.README.md
truboxl: Should I change to add folder exclusion instead?

AenBleidd: There is no official answer from Microsoft to disable real-time protection or exclude a folder.
The issue you mentioned is just a user finding, the same as a workaround to it. You can put a troubleshooting section that will lead to this issue instead of giving direct recommendations to decrease users' security

android/WSL.README.md
AenBleidd: Both systems can checkout the same commit. Basically, you can run vscode locally with connection to WSL, install gradle plugin and work with Java code without Android Studio

android/WSL.README.md
truboxl: Ok, actually Home can run WSL
https://ubuntu.com/wsl
https://devblogs.microsoft.com/commandline/wsl-2-post-build-faq/

AenBleidd: WSL2, not 1sr generation of WSL. As far as I know, WSL2 isn't released on stable channel yet

truboxl: Should I mention the further distinction of WSL1 and WSL2? I don't think it should be covered here as it's out of scope already. The point is WSL is available regardless of editions, only the build version and bitness matters

android/WSL.README.md
truboxl: Can we settle on whether we want to use Android Studio or VS Code + various plugins to develop for Android? I don't intend to introduce another IDE just for WSL. My current workflow is commit changes on WSL and test the code in Android Studio

android/WSL.README.md
AenBleidd: Currently WSL can't be installed on Windows 10 Home. I suppose this month Microsoft will do a major update and WSL 2 will be officially released but I didn't see the exact date of this

android/WSL.README.md
AenBleidd: Again, it's better to have two different source code bases, one inside WSL, another outside of it.
On WSL you can build native binaries and move only them to Windows and continue development on another source base with Android Studio

truboxl: Ok that's acceptable

android/WSL.README.md
truboxl: Windows 10 Home can run WSL1 (and soon WSL2)
https://en.wikipedia.org/wiki/Windows_10_editions#Comparison_chart
VirtualBox_windows 10 home_06_05_2020_20_43_18

android/WSL.README.md
AenBleidd: https://scotch.io/bar-talk/trying-the-new-wsl-2-its-fast-windows-subsystem-for-linux

Here's an info that if you have your files on VHD in WSL2 (not accessible outside of WSL) then the performance is super high.

So you probably can recommend this but it's up to you

talregev:
@truboxl @AenBleidd 
What is missing for this doc?

truboxl:
@talregev I think I have different opinions on best practices of using WSL1 to build the Android native client and submission of commits at an acceptable speed, error free and convenience. That said I will come back revisit again when WSL2 officially launch.

I am curious what is your set up of doing the Android side of development? How do you use Android Studio or another IDE? How do you build the native client on Windows?

talregev:
> I am curious what is your set up of doing the Android side of development?

I am using Window 10.
> How do you use Android Studio or another IDE?

I am using android studio. 
> How do you build the native client on Windows?

WSL
> How do you do commits on Windows?

Smartgit.

You?

talregev:
@truboxl WSL2 is official with may 2020 update.
Can you update the doc?

talregev:
@truboxl In wsl2 there is a memory issue.
Please also add this solution. In the example there is 6GB. But I was using 3GB running together ubuntu 18.04 and 20.04 and it was great performance.
https://github.com/microsoft/WSL/issues/4166#issuecomment-526725261

talregev:
@truboxl Can you add precondition for install the scripts?
From here:
https://boinc.berkeley.edu/trac/wiki/SoftwarePrereqsUnix
You should install this commans before install the android scripts: 

### Ubuntu 20.04
```bash
sudo apt install git python make m4 pkg-config dh-autoreconf libmysqlclient-dev libssl-dev python3-mysqldb libnotify-dev
```
### Ubuntu 18.04
```bash
sudo apt install git python make m4 pkg-config dh-autoreconf libmysqlclient-dev libssl-dev python-mysqldb libnotify-dev unzip
```

truboxl:
I follow these:
https://github.com/BOINC/boinc/blob/master/android/Vagrantfile#L119
https://boinc.berkeley.edu/trac/wiki/AndroidBuildClient#BuildtheClient
https://github.com/BOINC/boinc/blob/master/client/android/AndroidBuild.sh
and... https://boinc.berkeley.edu/trac/wiki/SoftwarePrereqsUnix table (client column)

> python

That's for building standalone toolchain from NDK which is unnecessary. It can be dropped in the future and use the tools directly from the NDK.

> dh-autoreconf

autoconf is sufficient. Unless you want to build `./configure`? That's also covered by `./_autosetup`. https://developer.fedoraproject.org/tech/languages/c/autotools.html

> libmysqlclient-dev python3-mysqldb

That's for building BOINC Server

> libnotify-dev

That's for building BOINC Manager for Linux

> libssl-dev

That's disto-specific architecture-dependent OpenSSL headers and libraries for Linux. It cannot be used on Android. We already download and build from sources and use that instead.

PR #3688, opened 2020-05-06
Updated Rosetta@home URL to use HTTPS

### LINKED ISSUE #3687: Rosetta@home now use HTTPS, Boinc list of projects doesn't
See https://boinc.bakerlab.org/rosetta/forum_thread.php?id=13894#95682 for more information

unguul:
Fixes #3687

**Description of the Change**

As per https://boinc.bakerlab.org/rosetta/forum_thread.php?id=13894#95682, Rosetta have updated their URL to use HTTPS. Probably a good idea update Boinc too.

P.S. I looked for a Linux variant of this list (`win_build` is for Windows only?) but could not find one so I'm guessing there isn't one.

**Alternate Designs**
N/A
**Release Notes**
Updated Rosetta@home URL to use HTTPS

### DIFF
--- a/win_build/installerv2/redist/all_projects_list.xml
+++ b/win_build/installerv2/redist/all_projects_list.xml
@@ -71,8 +71,8 @@
     <project>
         <name>Rosetta@home</name>
         <id>7</id>
-        <url>http://boinc.bakerlab.org/rosetta/</url>
-        <web_url>http://boinc.bakerlab.org/rosetta/</web_url>
+        <url>https://boinc.bakerlab.org/rosetta/</url>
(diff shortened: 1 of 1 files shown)


unguul:
Closing this as it will not have any affect as per https://github.com/BOINC/boinc/issues/3687#issuecomment-624842279

RichardHaselgrove:
There is no need to change all_project_list.xml for **clients** to attach direct to **projects**, because the get_project_config.php process is supplying the correct (new) master url for rosetta.

But all account managers will need to be updated, as discussed in #2642.

PR #3693, opened 2020-05-08
Mac: clean up gfx changes from PR #3666 (7.14 branch)

CharlieFenton:
Optimizes handling of graphics applications on Retina displays when built on Xcode 11 and run on OS 10.15 Catalina. My earlier code assumed that the problem (which I consider a bug in Apple's code) will continue on future versions of Xcode and OS X, and applied the workaround whenever built on Xcode 11 or later and run on OS 10.15 or later. This new approach is my attempt to future-proof the graphics apps and should work correctly whether or not Apple fixes the bug in the future.

Retina displays have 2X2 pixels per point. When OpenGL / GLUT apps built using Xcode 11 are run on a Retina display under OS 10.15, they fail to adjust their pixel dimensions to double the window size, and so fill only 1/4 of the window (display at half width and height.) They do correct this if resized resized by calls to glutReshapeWindow(), glutFullScreen(). etc. 

However, they work correctly when run on earlier  versions of OS X. OpenGL / GLUT apps built using earlier versions of Xcode do not have this problem on OS 10.15. 

We work around this by calling glutReshapeWindow() twice, first adding one to the width and then restoring the window's original size. This is imperceptible to the user, but transparently fixes the problem when necessary without ultimately changing the window's size, and does no harm when not necessary.

This PR also fixes an issue writing log files from the default BOINC screensaver graphics app boincscr on Macs and updates the samples/example_app project on Macs.

### DIFF
--- a/api/macglutfix.m
+++ b/api/macglutfix.m
@@ -60,14 +60,15 @@ int compareOSVersionTo(int toMajor, int toMinor);
 #define SAVERDELAY 30
 
 void MacGLUTFix(bool isScreenSaver) {
-    static bool done = false;
+    static int count = 0;
     static NSMenu * emptyMenu;
     NSOpenGLContext * myContext = nil;
     NSView *myView = nil;
     NSWindow* myWindow = nil;
+    static int requestedWidth, requestedHeight;
 
-    if (done) return;
-    done = true;
+    if (count > 1) return;   // Do this only twice
+    count++;
     
     if (! boinc_is_standalone()) {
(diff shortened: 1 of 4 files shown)


CharlieFenton:
This applies the same fixes to master which PR #3666 applies to the clent_release/7/7.14 branch. The reason for updating the clent_release/7/7.14 branch with this PR is that it is still being used by some projects for their branded versions of BOINC.

This PR is ready to be merged.

PR #3697, opened 2020-05-09
[Android] Add extension property to determine if the device is online.

Isira-Seneviratne:
**Description of the Change**
Add an extension property for `ConnectivityManager` to determine if the device is currently online and alter the existing code to make use of this. This is so that the mechanism to determine if the device is online can also be used in any future code.

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/.idea/misc.xml
+++ b/android/BOINC/.idea/misc.xml
@@ -39,5 +39,5 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="9" project-jdk-type="JavaSDK" />
 </project>
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/utils/BOINCUtils.kt
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/utils/BOINCUtils.kt
@@ -21,9 +21,20 @@
 package edu.berkeley.boinc.utils
 
 import android.content.Context
+import android.net.ConnectivityManager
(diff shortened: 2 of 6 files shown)


android/BOINC/.idea/misc.xml
```diff
@@ -39,5 +39,5 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="9" project-jdk-type="JavaSDK" />
```
adamradocz, line 42: Are you sure `project-jdk-name="9"`?

Isira-Seneviratne: Yeah, JDK 9 is required by Robolectric to run tests against Android SDK 29.

The Travis configuration has also been updated to use JDK 9 and therefore Xenial instead of Trusty.

android/BOINC/app/src/main/java/edu/berkeley/boinc/utils/BOINCUtils.kt
```diff
@@ -21,10 +21,16 @@
...
 
 import android.content.Context
+import android.net.ConnectivityManager
+import android.os.Build
 import edu.berkeley.boinc.R
 import java.io.IOException
 import java.io.Reader
 
+val ConnectivityManager.isOnline: Boolean
+    get() = if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M)
```
AenBleidd, line 31: What result will be in case when `Build.VERSION.SDK_INT >= Build.VERSION_CODES.M` ?

Isira-Seneviratne: There's an else case for that

Issue #3679, opened 2020-05-04
GPU Exclusive Application leaves CPU threads idle - Work Fetch does not get work


JacobWKlein:
**Describe the bug**
When I have a GPU Exclusive Application setup, and that process is running, then Work Fetch does not get work for all of my CPUs.

**Steps To Reproduce**
1. Setup a GPU Exclusive Application. Mine are "FahCore_21.exe" and "FahCore_22.exe".
2. Make sure that you have work for your GPU(s), that is downloaded and ready to run, but is not running because of the GPU Exclusive Application. My PC has 3 GPUs that are not being used by BOINC because of the GPU Exclusive Application, even though I have plenty of GPU work to run on them. Each GPU Work Item is set to use "1 CPU + 1 NVIDIA GPU".
3. (Probably required) For the projects that can possibly get CPU work, set them for Resource Share 0, so they are backup projects that will only get work when resources are idle.
4. (Optional) Set BOINC CPU Usage % to something that you can easily monitor. I have intentionally set mine for 50%, on an 8-core-16-thread PC.
5. BUG: BOINC Only runs 5 CPU threads, and does not fetch work for the other 3 CPUs.

**Expected behavior**
Work Fetch should see nidle of 3, to get work to use the CPUs, when the GPUs are not usable due to GPU Exclusive application in use.

**System Information**
 - OS: Windows 10 x64
 - BOINC Version: 7.16.5 x64

**Additional context**
Here's a work fetch log, when BOINC had the GPU Exclusive Application running and was only using 5 CPUs, despite being configured for 50% to use 8 CPUs .... and work fetch did not get work!

```
5/3/2020 11:40:23 PM |  | choose_project(): 1588563623.786711
5/3/2020 11:40:23 PM |  | [work_fetch] ------- start work fetch state -------
5/3/2020 11:40:23 PM |  | [work_fetch] target work buffer: 172800.00 + 43200.00 sec
5/3/2020 11:40:23 PM |  | [work_fetch] --- project states ---
5/3/2020 11:40:23 PM | https://boinc.drugdiscoveryathome.com/ | [work_fetch] REC 0.000 prio -0.000 can't request work: master URL fetch pending (82483.75 sec)
5/3/2020 11:40:23 PM | Acoustics@home | [work_fetch] REC 0.000 prio -1000.000 can't request work: "no new tasks" requested via Manager
[...]

PR #3703, opened 2020-05-11
client: avoid CPU starvation when GPU computing is suspended

davidpanderson:
Suppose we have 1 CPU and have (only) a GPU job that takes 1 CPU.
If GPU computing is suspended we won't fetch GPU work,
because the RR simulation doesn't take "suspended" into account.

Solution: in the RR simulation, if GPU computing is suspended,
act like GPU jobs don't use any CPU.
Seems like this should work though there may be unintended consequences.

Addresses #3679

### DIFF
--- a/client/rr_sim.cpp
+++ b/client/rr_sim.cpp
@@ -97,10 +97,17 @@ struct RR_SIM {
     inline void activate(RESULT* rp) {
         PROJECT* p = rp->project;
         active.push_back(rp);
-        rsc_work_fetch[0].sim_nused += rp->avp->avg_ncpus;
-        p->rsc_pwf[0].sim_nused += rp->avp->avg_ncpus;
-
         int rt = rp->avp->gpu_usage.rsc_type;
+
+        // if this is a GPU app and GPU computing is suspended,
+        // don't count its CPU usage.
+        // That way we'll fetch more CPU work if needed.
+        //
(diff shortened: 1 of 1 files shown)


AenBleidd:
@davidpanderson, does this fix real client behavior? For me it looks like just fix in simulator, if I'm not mistaken, of course

davidpanderson:
Different simulator.  The client does a simulation (rr_sim.cpp) to estimate job completion times and device utilization.

JacobWKlein:
Mmmmm... This change looks potentially delicious to me.

davidpanderson:
Glad to hear it.  Are you able to build the client and test it?

JacobWKlein:
Well, I've not tried in several years. If it is buildable on Windows 10 x64, using Visual Studio 2019 x64, and I had a "coach" like @AenBleidd for an hour, I'd be willing to give it a shot. Should I try?

AenBleidd:
@JacobWKlein, you can use binaries built with VS2013 from here: https://bintray.com/beta/#/boinc/boinc-ci/pull-requests/PR3703_2020-05-11_a327a0e2?tab=files

davidpanderson:
It is in fact buildable on Win 10, VS 2019.  Open win_build/boinc_vs2019.sln
But easier to get binaries and Vitalii suggests.

PR #3704, opened 2020-05-11
Tal r isi sen bug fix master

talregev:
Fix many issues that application crash in the master.


talregev:
I think it not so good. So I am closing this pr.

PR #3709, opened 2020-05-13
Client: don't allow empty GUI RPC password

davidpanderson:
If the GUI RPC file exists and is empty,
create a random password and write it to the file.

Running with an empty password is a security risk.
It means anyone on the same machine can control the client,
e.g. attach it to an arbitrary project.

### DIFF
--- a/client/boinc_cmd.cpp
+++ b/client/boinc_cmd.cpp
@@ -174,6 +174,7 @@ int main(int argc, char** argv) {
     char passwd_buf[256], hostname_buf[256], *hostname=0;
     char* passwd = passwd_buf, *p, *q;
     bool unix_domain = false;
+    string msg;
 
 #ifdef _WIN32
     chdir_to_data_dir();
@@ -181,7 +182,11 @@ int main(int argc, char** argv) {
     chdir("/Library/Application Support/BOINC Data");
 #endif
     safe_strcpy(passwd_buf, "");
-    read_gui_rpc_password(passwd_buf);
+    retval = read_gui_rpc_password(passwd_buf, msg);
+    if (retval) {
+        fprintf(stderr, "Can't get RPC password: %s\n", msg.c_str());
+        fprintf(stderr, "Only operations not requiring authorization will be allowed.\n");
+    }
 
 #if defined(_WIN32) && defined(USE_WINSOCK)
     WSADATA wsdata;
--- a/client/boinc_log.cpp
+++ b/client/boinc_log.cpp
@@ -15,10 +15,9 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 
-// boinclog: command-line interface to a BOINC client,
-// using GUI RPCs.
+// boinc_log: show stream of event-log messages from a client
 //
-// usage: boinccmd [--host hostname] [--passwd passwd] command
+// usage: boinc_log [--host hostname] [--passwd passwd] 
 
 #if defined(_WIN32) && !defined(__STDWX_H__) && !defined(_BOINC_WIN_) && !defined(_AFX_STDAFX_H_)
 #include "boinc_win.h"
@@ -176,7 +175,8 @@ int main(int argc, char** argv) {
 #endif
     }
 
-    read_gui_rpc_password(passwd_buf);
+    std::string msg;
+    read_gui_rpc_password(passwd_buf, msg);
 
     retval = rpc.init(hostname, port);
     if (retval) {
--- a/client/gui_rpc_server.cpp
+++ b/client/gui_rpc_server.cpp
@@ -120,7 +120,7 @@ bool GUI_RPC_CONN_SET::recent_rpc_needs_network(double interval) {
 }
 
 // read the GUI RPC password from gui_rpc_auth.cfg;
-// create one if missing.
+// create one if missing or empty.
 //
 void GUI_RPC_CONN_SET::get_password() {
     int retval;
@@ -132,39 +132,43 @@ void GUI_RPC_CONN_SET::get_password() {
             strip_whitespace(password);
         }
         fclose(f);
-        if (!strlen(password)) {
-            msg_printf(NULL, MSG_INFO,
-                "gui_rpc_auth.cfg is empty - no GUI RPC password protection"
-            );
+        if (strlen(password)) {
+            return;
         }
-        return;
+
+        // File is empty; don't allow this.
+        // Fall through and create a password.
+        //
+        msg_printf(NULL, MSG_INFO,
+            "%s is empty - assigning new GUI RPC password", GUI_RPC_PASSWD_FILE
+        );
     }
 
-    // if no password file, make a random password
+    // make a random password
     //
     make_secure_random_string(password);
 
     // try to write it to the file.
-    // if fail, just return
+    // if fail, just return; we're still password-protected
     //
     f = fopen(GUI_RPC_PASSWD_FILE, "w");
     if (!f) {
         msg_printf(NULL, MSG_USER_ALERT,
-            "Can't open gui_rpc_auth.cfg - fix permissions"
+            "Can't open %s - fix permissions", GUI_RPC_PASSWD_FILE
         );
     } else {
         retval = fputs(password, f);
         fclose(f);
         if (retval == EOF) {
             msg_printf(NULL, MSG_USER_ALERT,
-                "Can't write gui_rpc_auth.cfg - fix permissions"
+                "Can't write %s - fix permissions", GUI_RPC_PASSWD_FILE
             );
         }
     }
 #ifndef _WIN32
-    // if someone can read the password,
+    // Make sure the password file is not world-read or write.
+    // If someone can read or set the password,
     // they can cause code to execute as this user.
-    // So better protect it.
     //
     if (g_use_sandbox) {
         // Allow group access so authorized administrator can modify it
--- a/client/gui_rpc_server.cpp
+++ b/client/gui_rpc_server.cpp
@@ -165,14 +165,25 @@ void GUI_RPC_CONN_SET::get_password() {
             );
         }
     }
-#ifndef _WIN32
-    // Make sure the password file is not world-read or write.
+#ifdef _WIN32
+#elif defined(__APPLE__)
+    // Mac: Make sure the password file is not world-read or write.
     // If someone can read or set the password,
     // they can cause code to execute as this user.
     //
     if (g_use_sandbox) {
         // Allow group access so authorized administrator can modify it
         chmod(GUI_RPC_PASSWD_FILE, S_IRUSR|S_IWUSR | S_IRGRP | S_IWGRP);
+    } else {
+        chmod(GUI_RPC_PASSWD_FILE, S_IRUSR|S_IWUSR);
+    }
+#else
+    // general case: allow group read if group is "boinc"
+    //
+    gid_t gid = getgid();
+    struct group *g = getgrgid(gid);
+    if (g && !strcmp(g->gr_name, "boinc")) {
+        chmod(GUI_RPC_PASSWD_FILE, S_IRUSR|S_IWUSR | S_IRGRP);
     } else {
         chmod(GUI_RPC_PASSWD_FILE, S_IRUSR|S_IWUSR);
     }
--- a/clientgui/BOINCBaseFrame.cpp
+++ b/clientgui/BOINCBaseFrame.cpp
@@ -500,19 +500,10 @@ void CBOINCBaseFrame::ShowConnectionBadPasswordAlert(
         if (password_msg.empty()) {
             password_msg = "Invalid client RPC password.  Try reinstalling BOINC.";
         }
-        ShowAlert(
-            strDialogTitle,
-            password_msg,
-            wxOK | wxICON_ERROR
-        );
     } else {
-        ShowAlert(
-            strDialogTitle,
-            _("The password you have provided is incorrect, please try again."),
-            wxOK | wxICON_ERROR
-        );
+        password_msg = "Invalid client RPC password.  Try reinstalling BOINC.";
     }
-
+    wxMessageBox(wxString(password_msg), strDialogTitle, wxOK | wxICON_ERROR);
     wxLogTrace(wxT("Function Start/End"), wxT("CBOINCBaseFrame::ShowConnectionBadPasswordAlert - Function End"));
 }
 
@@ -623,12 +614,7 @@ void CBOINCBaseFrame::ShowDaemonStartFailedAlert() {
     );
 #endif
 
-    ShowAlert(
-        strDialogTitle,
-        strDialogMessage,
-        wxOK | wxICON_ERROR
-    );
-
+    wxMessageBox(strDialogMessage, strDialogTitle, wxOK | wxICON_ERROR);
     wxLogTrace(wxT("Function Start/End"), wxT("CBOINCBaseFrame::ShowDaemonStartFailedAlert - Function End"));
 }
 
@@ -684,12 +670,7 @@ void CBOINCBaseFrame::ShowNotCurrentlyConnectedAlert() {
         pSkinAdvanced->GetApplicationShortName().c_str(),
         pSkinAdvanced->GetApplicationShortName().c_str()
     );
-    ShowAlert(
-        strDialogTitle,
-        strDialogMessage,
-        wxOK | wxICON_ERROR
-    );
-
+    wxMessageBox(strDialogMessage, strDialogTitle, wxOK | wxICON_ERROR);
     wxLogTrace(wxT("Function Start/End"), wxT("CBOINCBaseFrame::ShowNotCurrentlyConnectedAlert - Function End"));
 }
 
--- a/clientgui/BOINCBaseFrame.cpp
+++ b/clientgui/BOINCBaseFrame.cpp
@@ -476,7 +476,9 @@ bool CBOINCBaseFrame::SelectComputer(wxString& hostName, int& portNum, wxString&
 }
 
 
-void CBOINCBaseFrame::ShowConnectionBadPasswordAlert( bool bUsedDefaultPassword, int iReadGUIRPCAuthFailure ) {
+void CBOINCBaseFrame::ShowConnectionBadPasswordAlert(
+    bool bUsedDefaultPassword, int iReadGUIRPCAuthFailure, std::string password_msg
+) {
     CSkinAdvanced*      pSkinAdvanced = wxGetApp().GetSkinManager()->GetAdvanced();
     wxString            strDialogTitle = wxEmptyString;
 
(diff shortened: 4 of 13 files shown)


RichardHaselgrove:
@davidpanderson - can I ask you to liaise closely with Linux package maintainers like @Germano0 and @LocutusOfBorg in developing this idea further?

My experience is with Linux Mint (based on Ubuntu 18.04) and Gianfranco's PPA distribution - other experiences may vary.

The installation roadmap is:
```
Binaries:    /usr/bin
Data:        /var/lib/boinc-client
Preferences: /etc/boinc-client
```
There is a softlink in the Data directory to each of the files in the Preferences directory, including gui_rpc_auth.cfg

Under this configuration, BOINC Manager **will not function** on localhost when gui_rpc_auth.cfg is non-empty: the Manager launches, but cannot connect to the Client. This is a password problem: I have worked round it by adding `--password=password` to the command line in my desktop launcher for the Manager.

The Windows Manager has the ability to detect the local machine gui_rpc_auth.cfg password when invoked on the local machine: the password only has to to be provided manually when using [Controlling BOINC remotely](https://boinc.berkeley.edu/wiki/Controlling_BOINC_remotely). Either the BOINC code, or the installation package, will have to emulate the Windows behaviour for this change to be acceptable to users.

Germano0:
The behaviour on Fedora / CentOS / RHEL is that BOINC client the first time is started as service, automatically creates a `/var/lib/boinc/gui_rpc_auth.cfg` file with a password. Then if the system administrator wants to remove that password, can simply run `# echo > /var/lib/boinc/gui_rpc_auth.cfg` and restart BOINC client.
I think that users must be able choose to remove the password so I disagree with this pull request

RichardHaselgrove:
Further analysis: The Mac and Windows Managers have the capability to determine the BOINC data directory (and hence the location of gui_rpc_auth.cfg) through

`void CBOINCGUIApp::DetectDataDirectory()`
(https://github.com/BOINC/boinc/blob/master/clientgui/BOINCGUIApp.cpp#L796 ff)

There is no matching capability for the Linux Manager.

TheAspens:
There are three issues tied up here making this a bit of a mess:

1) Locating the data directory on Linux
2) Securing access to BOINC
3) How the manager should behave on Linux if it cannot find the gui_rpc_auth.cfg

I would suggest the following as the fix for all three:

1) /etc/boinc-client/config.properties should be used on all Linux platforms and contain the following:
`data_dir=/var/lib/boinc` or `data_dir=/var/lib/boinc-client` depending on where the data directory actually is.  It should not contain the files (cc_config.xml, global_prefs_override.xml, gui_rpc_auth_cfg or remote_hosts.cfg) that are located there on my rasberry pi.

2) The package managers should not create gui_rpc_auth.cfg files as part of the installation.  BOINC creates them when the core client first runs.  Additionally, it is important to be secure so if the file is an empty string, then the core client should replace it with a random password as if the file was not present.

3) boincmgr and boinccmd on Linux should attempt to read the gui_rpc_auth.cfg in the following locations: 
- from the command line
- in the current directory
- in the directory designated by the config file /etc/boinc-client/config.properties

If it cannot find or cannot read the gui_rpc_auth.cfg file, it should report a user friendly message to the user stating that it cannot connect to the core client. (The goal would be for this message to provide enough information to resolve the issue without having to look up information on the internet).  A message like "The user knreed needs to be added to the boinc group in order to read the access file /var/lib/boinc-client/gui_rpc_auth.cfg" if the file was found but could not be read.  A message like "Unable to locate the gui_prc_auth.cfg file in order to connect to the core client." if the file cannot be found.  Once the message was exited the manager would be in a disconnected state. (Note that the message text needs editing).

The suggestions above need some tweaks, but I believe that this approach would improve usability of the manager on Linux as well as improve the security on Linux as is the intent of David's changes.

RichardHaselgrove:
I think there may be a (positive, beneficial) reason for placing the user-editable configuration files separately in /etc/boinc-client. On my older machine (only), I find that I can edit these files using my normal desktop user account and the softlinks in /var/lib/boinc-client: all other text files in /var/lib/boinc-client are read-only. On my newer machine, even the softlinked files are read-only. I've done more testing and experimenting on the older machine, so I may at some point have added my user account to a boinc group, or something: I don't remember, but it's useful. @Germano0 may be able to suggest a reason for how the difference might have come about.

It might be helpful to readers to have a reference copy of boinc-client.service, as placed in /etc/systemd/system/multi-user.target.wants by @LocutusOfBorg's PPA.

```
[Unit]
Description=Berkeley Open Infrastructure Network Computing Client
Documentation=man:boinc(1)
After=network-online.target

[Service]
Type=simple
ProtectHome=true
PrivateTmp=true
ProtectSystem=strict
ProtectControlGroups=true
ReadWritePaths=-/var/lib/boinc -/etc/boinc-client
Nice=10
User=boinc
WorkingDirectory=/var/lib/boinc
ExecStart=/usr/bin/boinc
ExecReload=/usr/bin/boinccmd --read_cc_config
ExecStopPost=/bin/rm -f lockfile
IOSchedulingClass=idle
# The following options prevent setuid root as they imply NoNewPrivileges=true
# Since Atlas requires setuid root, they break Atlas
# In order to improve security, if you're not using Atlas,
# Add these options to the [Service] section of an override file using
# sudo systemctl edit boinc-client.service
#NoNewPrivileges=true
#ProtectKernelModules=true
#ProtectKernelTunables=true
#RestrictRealtime=true
#RestrictAddressFamilies=AF_INET AF_INET6 AF_UNIX
#RestrictNamespaces=true
#PrivateUsers=true
#CapabilityBoundingSet=
#MemoryDenyWriteExecute=true

[Install]
WantedBy=multi-user.target
```

RichardHaselgrove:
Ah - SUDO has now granted all users (there's only me) read-write access to /etc/boinc-client and contents. The two machines now behave the same: user files are editable, boinc files are read-only.

TheAspens:
@RichardHaselgrove - the reason you couldn't edit the files is due to the ownership and the permissions assigned to the files not because of the location of the files.

On my rasberry pi, the files in /etc/boinc-client were owned by root and had a group of boinc.  The files permissions were set so that only the owner could read and write them while the group only had permission to read them (there might have been one file that let the group read and write - i've changed the permissions on my local machine so I don't know for sure).

I think that the proper way to handle access controls on Linux is that users who need to manage BOINC should be added to the boinc group `usermod -a -G boinc knreed` (for example).

"User editable" files should be set with group read/write and files that users really shouldn't be editing should be set to group read only (or no group permission at all).  

As I understand it, many linux distributions now have a nice UX that allows for application installation.  I believe that logic can be added to this interface so that the user doing the installation can be automatically added to the boinc group.  (note that I do not believe that this can be done via command line installation.  i.e. `sudo yum install boinc-client boinc-manager`)

If the user who installed the software was automatically added to the boinc group during installation and the other steps that I recommend above are implemented, and if the boinc core client is automatically started after installation, then the user will be able to immediately launch the manager and starting interacting without any further action on their part.

TheAspens:
> I think that users must be able choose to remove the password so I disagree with this pull request

The software should not allow users to use an insecure configuration.  This means that it should never allow unauthenticated access to control the operations of the core client.  

For example, with some of the changes in the 7.16 client, the core client can handle HTTP requests.  This means that you have to worry about the user browsing to a website on that computer and the website hosting malicious code that will initiate a connection to the client from that code that attaches the client to a malicious project.  If the password is empty then this becomes an easy path of attack.

RichardHaselgrove:
@TheAspens - close, but not exact. For my PPA installation, I have

/etc/boinc-client - Owner root, Group root, Others access files (that's the one I tweaked)
/var/lib/boinc-client - Owner 'boinc - BOINC core client', Group boinc, Others access files.

But in outline, I agree completely with your approach.

Mint is one of the Linux distros targeted at migrating Windows users, with a fully graphical interface. I can install software with a Synaptic Package Manager (it prompts for a sudo admin password), I can run a File Browser with either normal or administrator privileges, and so on. It works well for the home owner-administrator setting, which I've commented on before - we can't be calling on Corporate HQ to get us out of a mess. And there are more of us around since the rather raw introduction of Windows 10.

Germano0:
> For example, with some of the changes in the 7.16 client, the core client can handle HTTP requests. This means that you have to worry about the user browsing to a website on that computer and the website hosting malicious code that will initiate a connection to the client from that code that attaches the client to a malicious project. If the password is empty then this becomes an easy path of attack.

Ok this makes sense but I am also concerned about the average user that will have to enter a password everytime he wants for example to pause the client

RichardHaselgrove:
I think that the intended behaviour is that the client will fill in its own password if you are trying to pause localhost: the user will only have to supply the password to pause a remote machine over the network.

I think that's a reasonable balance between usability and security (if an attacker has got access to your machine as if it's localhost - either in person or via SSH - you've got a lot more to worry about than being attached to a random BOINC project).

davidpanderson:
I added code to implement Kevin's model

TheAspens:
I've tested this on Linux and here is what I found:

What works as expected:
- If the core client finds an empty gui_prc_auth.cfg file, it creates a random password and writes it to the file
- If the boinccmd tool cannot find the gui_rpc_auth.cfg file it writes the following to stderr 
```
Can't get RPC password: gui_rpc_auth.cfg not found.  Try reinstalling BOINC.
Only operations not requiring authorization will be allowed.
```
- If the boinccmd tool can find the gui_rpc_auth.cfg file for the /etc/boinc-client/config.properties file but it cannot read the file it writes: 
```
Can't get RPC password: /var/lib/boinc/gui_rpc_auth.cfg exists but can't be read.  Check the file permissions.
Only operations not requiring authorization will be allowed.
```
- If the boinccmd tool is able to read both files, its able to run all commands correctly
- If the boincmgr is able to read both files, it connects automatically

What doesn't work as expected:
- If the boincmgr is not able to read or find the gui_rpc_auth.cfg file, it opens in a disconnected state without any information to the user about how to fix the problem.  Expected Behavior: It should provide the same level of messaging as the boinccmd tool as an error message displayed in the GUI.


This is a huge step forward.  I am very happy to be able to run "boincmgr" from my command line have it open and connect instead of having to type boincmgr -d /var/lib/boinc every time.  I think once we can get the error message to display as expected this will be ready.  Thanks for doing this work - I think it will make a lot of Linux users happy.

davidpanderson:
I fixed those problems and a few more; please try again.

TheAspens:
That worked like a charm.  I'm ready to approve this pull request.   Since I'm not able to test on Mac or Windows I'm going to leave this open until Wednesday to see if someone is able to test on those platforms.  Otherwise I'll merge on Wednesday.

Note - This will definitely be a nice add on Linux.  I had gotten used to not running the manager b/c typing boincmgr -d /var/lib/boinc all the time was a bit annoying, but with these changes I've added it to my "Favorites" and have found myself using it every couple of days to check on things.   Thanks for implementing these changes.

RichardHaselgrove:
Tested on Windows 7/64, using the Appveyor artifacts. New Manager connected to running client, with existing password - success. Deleted qui_rpc_auth.cfg, launched new client by running new Manager - client created new file/random password, started, Manager connected - success.

Old Manager used for monitoring upstairs disconnected - showed message 'password incorrect' (first time that message has been true, in my experience!).

Changed password back to normal, read config files - Manager upstairs failed to connect.
Restarted client, Manager upstairs does now connect.

Would it be possible/easy/desirable for 'read config files' to activate a new password immediately? Otherwise, we need to document that the client restart is needed.

I'll do the same test on Mint as soon as I can remind myself how to locate the Linux build (@AenBleidd did tell me, but I've lost the note...)

AenBleidd:
@RichardHaselgrove,
https://bintray.com/beta/#/boinc/boinc-ci/pull-requests/PR3709_2020-05-18_745e5056?tab=files

RichardHaselgrove:
@AenBleidd - got 'em, ta.

RichardHaselgrove:
OK, repeated same tests in Linux Mint.

1) New manager wouldn't launch from desktop - silent fail. Started it from terminal instead, got a missing library error on libpng12.0. Downloaded new copy, Firefox offered to run installer, installer said same version is already installed, but now Manager runs. Shrug.

2) Manually placed new client in var/lib/boinc-client. Deleted both previous copies of gui_rpc_auth.cfg (softlink and real). Started with sudo systemctl start boinc-client. Fans speed up - it's running. gui_rpc_auth.cfg is back in var/lib/boinc-client, as 32-byte file. Start Manager from desktop. Get GUI error message 'File var/lib/boinc-client found, but can't be read'. Fix file access rights as superuser, Manager finally opens properly.

3) Manager upstairs reports bad password (true - it was running on my normal one) and disconnects. Superuser reverts random password in new file, restarts service, and we're back in business both upstairs and downstairs. I now have three preference files in one folder, one in a different folder. As myself, I can only edit the softlinked three. I can live with that - I don't change this password very often - but I don't think it's ideal. Perhaps @Germano0 can tweak.

Germano0:
I tested the patch on BOINC running as a service, and everytime I open the manager I get the message
`gui_rpc_auth.cfg not found. Try reinstalling BOINC`
After that I am able to connect to the client by clicking `File->Select computer`, and inserting localhost (or 127.0.0.1) and the gui_rpc_auth.cfg password.
The error appears regardless if gui_rpc_auth.cfg contains a password or not (and if not, the client creates one at the first client start).
The error persists even setting the whole `/var/lib/boinc` with 777 permissions.

RichardHaselgrove:
@Germano0 - did you create the /etc/boinc-client/config.properties redirect (data folder location) that @TheAspens suggested?

I'm going to spend some time today trying to explore the file permissions that led to my 'file exists, but can't be read' error message yesterday. I'll be using my second, less experimental, host for the purpose.

Germano0:
> Germano0 - did you create the /etc/boinc-client/config.properties redirect (data folder location) that TheAspens suggested?

No I did not. I will try as soon as possible

RichardHaselgrove:
Well, this is getting more and more confusing. I'll try to document the complexities as best I can. This was the starting position on the upstairs (production) machine - the one I do little experimenting on.
```
richard@host:~$ groups
richard adm cdrom sudo dip plugdev lpadmin sambashare
```
(so I'm not in the boinc group - I am downstairs, and this may be significant)

```
richard@host:/etc/boinc-client$ ls -l gui_rpc_auth.cfg
-rw-r--rw- 1 root boinc 9 Feb  3 14:43 gui_rpc_auth.cfg
```
(the PPA created this, but I added the second w)

```
richard@host:/var/lib/boinc-client$ ls -l gui_rpc_auth.cfg
lrwxrwxrwx 1 root root 34 Apr  2 15:35 gui_rpc_auth.cfg -> /etc/boinc-client/gui_rpc_auth.cfg
```
(PPA, unchanged)

So, I closed down the PPA client. Renamed the binaries, in case I needed them again. Using File Browser in administrative mode, dragged the test binaries out of the .7z files and dropped them in /usr/bin. Used admin terminal to set chmod +x on each, just in case. Deleted all copies and links for gui_rpc_auth.cfg. Client started just fine and created the file, with random password. Manager gave me the same grief about libpng12-0 not being installed. Installed it, shouldn't be a problem with the package install.

And I couldn't get the Manager to see the cfg file. No way. Just

Screenshot at 2020-05-19 12-43-44

The file the Client created was
```
richard@host:/var/lib/boinc-client$ ls -l gui_rpc_auth.cfg
-rw------- 1 boinc boinc 32 May 19 12:16 gui_rpc_auth.cfg
```
I went all the way up to 
`-rw-r--r-- 1 boinc boinc 32 May 19 12:16 gui_rpc_auth.cfg`
but it still couldn't be seen.

Then I tried adding myself to the boinc group, with this result:
```
sudo usermod -a -G boinc richard
root@host:/home/richard# adduser richard boinc
The user `richard' is already a member of `boinc'.
richard@host:/$ groups
richard adm cdrom sudo dip plugdev lpadmin sambashare
```
Am I, or aren't I?

So I did a clean reboot of the machine, checked I had Manager access with a password on the command line, and went out for a walk and a think.

Trying again after I got back, and it works perfectly - launches from desktop launcher, no additional password needed. I can only surmise that the combination of adding myself to the boinc group, and rebooting the machine, did the trick - but I'm running out of machines to test on.

Germano0:
I tried adding `/etc/boinc-client/config.properties` and I opening the manager I got the error message `gui_rpc_auth.cfg exists but can't be read. Check the file permissions.`
Trying to put 777 permissions on `/var/lib/boinc` fixed the problem. Obviously this cannot be a solution.
On Fedora / CentOS / RHEL `/var/lib/boinc` is owned by `boinc` user and the manager runs with "regular user" account so it cannot access directly to `/var/lib/boinc` data. Only BOINC client can access to that folder since it's running as a service and as `boinc` user

TheAspens:
I recommend that you use the following settings:
```
chmod 755 /etc/boinc-client
chmod 644 /etc/boinc-client/config.properties
chown -R boinc:boinc /etc/boinc-client

chmod 750 /var/lib/boinc-client
chmod 640 /var/lib/boinc-client/gui_rpc_auth.cfg
chown boinc:boinc /var/lib/boinc-client
chown boinc:boinc /var/lib/boinc-client/gui_rpc_auth.cfg
```
The "regular user" needs to be added to the boinc group in order to manage the boinc client.  This also means that only users in the boinc group can manage the boinc client.  This is an appropriate situation to be in from a security perspective.

TheAspens:
@David - can you change things so that when the core client creates/alters the gui_rpc_auth.cfg file it sets permissions to read/write for user, read for group and nothing for other?

TheAspens:
@davidpanderson - Can you change the message 

> gui_rpc_auth.cfg exists but can't be read. Check the file permissions.

to

> gui_rpc_auth.cfg exists but can't be read. Check the file permissions. You may need to be added to the boinc group.

Or something that indicates that the preferred option is for the user to add themselves to the group rather than change the permissions of the file.

davidpanderson:
Problem is, if someone runs a BOINC client, it's controllable by anyone in their group,
which may not be what they want.
I suggest that when the client creates the password file:
- if the current group is "boinc", make the file group readable
- otherwise don't
(and leave things as they are on Win and Mac)

TheAspens:
@davidpanderson - good point.  I think your proposal is the way to go.

TheAspens:
Tested and this works as designed.  Thanks for the change!

Ageless93:
Off late we have people on the BOINC forums who install BOINC on their new Linux systems and immediately get shut out because BOINC Manager cannot connect to the client due to gui_rpc_auth.cfg existing but that it cannot be read, mostly due to file permissions. It appears the file is made by root and is read-only. A user cannot easily edit the file to add his own password, or run BOINC Manager natively without a lot of legwork. 

The latest threads (https://boinc.berkeley.edu/dev/forum_thread.php?id=14023 and https://boinc.berkeley.edu/dev/forum_thread.php?id=14013) have all the Linux gurus try to help, but it's getting a bit out of hand. It should be easier to have BOINC Manager connect to the BOINC client without having to rewrite every file's permissions. Because this can cost a lot of potential people who don't want to ask at the forums how to make this work, but instead just uninstall and tell their friends not to bother, because it isn't easily installed and used.

Vulpine05:
Not that there needs to be another example, but I'm having this problem with 18.04.5 as well, as discussed here:  [https://boinc.berkeley.edu/forum_thread.php?id=13946#100574](https://boinc.berkeley.edu/forum_thread.php?id=13946#100574).  I'm running Linux as a free OS to use to run BOINC, I am also too inexperienced to navigate Linux to troubleshoot, nor have the time to (unfortunately).

standard7452:
idk if this has been all sorted out, but i got this to work on F33 by following the steps in this:  https://boinc.berkeley.edu/wiki/Installing_BOINC_on_Fedora

PR #3714, opened 2020-05-14
[Android] Add debug for android scripts

talregev:
Add debug for android scripts
Environment variable BOINC_DEBUG=ON
Add CFLAGS and CXXFLAGS can added from outside.

### DIFF
--- a/android/build_boinc_arm.sh
+++ b/android/build_boinc_arm.sh
@@ -28,8 +28,13 @@ export LD=arm-linux-androideabi-ld
 export CFLAGS="--sysroot=$TCSYSROOT -DANDROID -DDECLARE_TIMEZONE -Wall -I$TCINCLUDES/include -O3 -fomit-frame-pointer -fPIE -march=armv7-a -D__ANDROID_API__=19 "$CFLAGS
 export CXXFLAGS="--sysroot=$TCSYSROOT -DANDROID -Wall -I$TCINCLUDES/include -funroll-loops -fexceptions -O3 -fomit-frame-pointer -fPIE -march=armv7-a -D__ANDROID_API__=19 "$CXXFLAGS
 export LDFLAGS="-L$TCSYSROOT/usr/lib -L$TCINCLUDES/lib -llog -fPIE -pie -latomic -static-libstdc++ -march=armv7-a -Wl,--fix-cortex-a8"
-export GDB_CFLAGS="--sysroot=$TCSYSROOT -Wall -g -I$TCINCLUDES/include"
 export PKG_CONFIG_SYSROOT_DIR="$TCSYSROOT"
+
+if [ ! "$DEBUG" = "ON" ]; then
+    export GDB_CFLAGS="--sysroot=$TCSYSROOT -Wall -g -I$TCINCLUDES/include"
+else
+    export GDB_CFLAGS=""
+fi
 
 # Prepare android toolchain and environment
 ./build_androidtc_arm.sh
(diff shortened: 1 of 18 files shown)


adamradocz:
Using only `DEBUG` instead of `BOINC_DEBUG` wouldn't be a better choice?

talregev:
I am not sure, Deubg is so general, so maybe it will do side effects for things we  don't know.

android/build_boinc_arm.sh
```diff
@@ -25,8 +25,8 @@ export PATH="$TCBINARIES:$TCINCLUDES/bin:$PATH"
 export CC=arm-linux-androideabi-clang
 export CXX=arm-linux-androideabi-clang++
 export LD=arm-linux-androideabi-ld
-export CFLAGS="--sysroot=$TCSYSROOT -DANDROID -DDECLARE_TIMEZONE -Wall -I$TCINCLUDES/include -O3 -fomit-frame-pointer -fPIE -march=armv7-a -D__ANDROID_API__=19"
-export CXXFLAGS="--sysroot=$TCSYSROOT -DANDROID -Wall -I$TCINCLUDES/include -funroll-loops -fexceptions -O3 -fomit-frame-pointer -fPIE -march=armv7-a -D__ANDROID_API__=19"
+export CFLAGS="--sysroot=$TCSYSROOT -DANDROID -DDECLARE_TIMEZONE -Wall -I$TCINCLUDES/include -O3 -fomit-frame-pointer -fPIE -march=armv7-a -D__ANDROID_API__=19 "$CFLAGS
```
AenBleidd, line 28: This change adds no support of Debug build. You can leave here an ability to add parameters from the outside but also please override this value with `-g` parameter for Debug builds.

android/build_boinc_arm.sh
```diff
@@ -51,7 +51,9 @@ if [ -n "$COMPILEBOINC" ]; then
 
     echo "Stripping Binaries"
     cd stage/usr/local/bin
-    arm-linux-androideabi-strip *
+    if [ ! "$BOINC_DEBUG" = "ON" ]; then
```
AenBleidd, line 54: Let's use general `DEBUG`. I see no reason why it should have some side effects.

android/build_boinc_arm.sh
talregev: I am not sure I understand

talregev:
> These two comments are applicable to all other scripts.
> Also, if we are building Debug, it should be better to build openssl and curl in Debug mode too.
> So please adjust corresponding scripts too.

The openssl and curl already build in debug mode.

talregev:
From my test all the binaries are build in debug mode. Boinc, openssl, and curl.
It strip off the debug symbol by strip command.
When the Debug flag is on, it not doing the strip and it stay at debug.
I am not passing the -g throw outside CFLAGS and CXXFLGAS because the -g flag already in GDB_CFLAGS in all scripts (boinc, openssl curl) in all platforms.

talregev:
So if you allow to add external CFLAGS and CXXFLAGS in this PR,
All I need to do is to change BOINC_DEBUG to DEBUG.

android/build_boinc_arm.sh
AenBleidd: Apply then ability to add parameters from the outside to all other scripts

truboxl:
I think GDB_CFLAGS was not used at all, if ever
So I think can remove them and directly integrate them into CFLAGS and/or CXXFLAGS based on DEBUG script flag

Please test if it ever works with LLDB, I can't seem to get it work

talregev:
I compile the boinc with GDB_CFLGAS empty and without the strip and I get a debug information.
I don't know how to compile in release mode without the strip.
@AenBleidd  What do you want to do about this?

PR #3719, opened 2020-05-15
[Android] Fix some issues with the app.

Isira-Seneviratne:
**Description of the Change**
This fixes an issue with the `NavDrawerItem` class that prevented projects that have been added from being displayed in the navigation drawer.

In addition, an exception will no longer be thrown when parsing `AppVersion`.

**Release Notes**
Added projects will now be displayed properly in the navigation drawer.


android/BOINC/app/src/main/java/edu/berkeley/boinc/adapter/NavDrawerListAdapter.java
```diff
@@ -195,20 +188,21 @@ public Bitmap getProjectIconForMasterUrl(String masterUrl) {
      * @param projects Project list
      * @return Returns number of project items in nav bar after adding
      */
-    public int compareAndAddProjects(List<Project> projects) {
+    public Integer compareAndAddProjects(List<Project> projects) {
```
AenBleidd, line 191: `int` could be used instead of `Integer`

PR #3721, opened 2020-05-16
Manager: change http to https in simple view background

davidpanderson:


RichardHaselgrove:
Simple, safe, worthwhile. Merging to reduce the backlog count.

PR #3724, opened 2020-05-16
Update gitignore

talregev:
Update gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,9 +129,6 @@ locale/*/*.flag
 3rdParty/linux/
 3rdParty/mac/
 3rdParty/buildCache/
-3rdParty/android*
-3rdParty/curl-*
-3rdParty/openssl-*
 
 .libs/
 svn_version.h
--- a/.gitignore
+++ b/.gitignore
@@ -129,6 +129,9 @@ locale/*/*.flag
 3rdParty/linux/
 3rdParty/mac/
 3rdParty/buildCache/
+3rdParty/android*
+3rdParty/curl-*
+3rdParty/openssl-*
 
 .libs/
 svn_version.h
@@ -213,6 +216,12 @@ vda/ssim
 vda/vda
 vda/vdad
 /win_build/*.sdf
+/win_build/*.xml
+/win_build/notices/*.xml
+/win_build/*.txt
+/win_build/*.cfg
+/win_build/*log
+
 
 # OS-specific files
 .DS_Store


.gitignore
```diff
@@ -129,6 +129,9 @@ locale/*/*.flag
 3rdParty/linux/
 3rdParty/mac/
 3rdParty/buildCache/
+3rdParty/android*
```
AenBleidd, line 132: build dependencies for android should be inside `3rdParty/buildCache` directory

.gitignore
```diff
@@ -213,6 +216,12 @@ vda/ssim
 vda/vda
 vda/vdad
 /win_build/*.sdf
+/win_build/*.xml
```
AenBleidd, line 219: `/win_build/` should not be used as a DATA directory

talregev: You compile the boic.exe to win_build. 
When you debug it will create data file to there.
So this is the default folder as a data.
Developer should not see this extra files in the git when he push PR.
So if it is not data folder as you say, you shouldn't add xml file in a default. If want in the future to add an xml to win_build with other reason, you can still do it with git add force.
Then you know what are you doing.

.gitignore
talregev: But it not because the android scripts.
It create two folders that start with android:
3rdParty\android\.wget-hsts
3rdParty\android-ndk-r18b

AenBleidd: Android scripts by default use `prefix` that is targeted to buildcache folder. If it's not - then scripts should be fixed

talregev: I think it not. That why it also install the other folders openssl-* and curl-*
And I think It good idea that the prefix should be 3rdParty/buildCache/

.gitignore
AenBleidd: Just pass `datadir` parameter with value as a command line parameter to BOINC Manager or `dir` with value to BOINC client

talregev: I don't think is a good idea that every time I start a new debug I should set something.
When I push play on the debug, it should work.
Maybe to create a default folder inside boinc for data, then change it for release. 
Not sure how to do it.

AenBleidd: You can set commandline parameters in this window (if you're using VS) and it will be saved
image

talregev: > You can set commandline parameters in this window (if you're using VS) and it will be saved

This is the things I am trying to avoid. Setting every time I am switching or change something in the branch.

AenBleidd: This setting will be saved even when you switch branches

talregev: If you don't agree with me, you can close this PR.
But this is what is how you should work, easy and without any other settings.
If you can do it without intefeare the users  and other developer and make life easier then you can do it.

talregev:
@AenBleidd On android we agree, on windows we will never agree. Close this PR.

PR #3732, opened 2020-05-17
Nicer look template.

talregev:
Nicer look template.
I think it easier to read this template especially in the feature request template when you have only txt.
It easier to read the user txt within template txt.  

Here an example for feature request look like: 
Feature request

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,22 +4,29 @@ about: Create a report to help us improve
 
 ---
 
-**Describe the bug**
-A clear and concise description of what the bug is.
+> **Describe the bug**
+> A clear and concise description of what the bug is.
 
-**Steps To Reproduce**
+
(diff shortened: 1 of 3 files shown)


AenBleidd:
I don't like the idea of quoting. Comments are supposed to removed when creating an issue because it's just a hint to reporter.
Maybe, you can use [this](https://github.com/microsoft/vcpkg) repo as an inspiration?

talregev:
> Maybe, you can use [this](https://github.com/microsoft/vcpkg) repo as an inspiration?

If you like this issues in their repo, feel free to take.
I did it just a suggestion. 
Because if you are a dev that want to read the bug issue / feature request very quickly without the repeating txt that make hard to understand what the user want.
quoting is dimmer the repeating txt and also keep the structure of the issue.
If you know other trick to it without quoting, let me know.
If you don't like this suggestion, you can close it.

AenBleidd:
Current template is fine enough

PR #3737, opened 2020-05-18
[Android] Fix account manager  authentication

talregev:
Fix  account  manager authentication.


talregev:
Push force on the way ...

PR #3740, opened 2020-05-19
Fix boinc platform strings for Android

truboxl:
**Description of the Change**
Some compiled clients may appear as `$ARCH-unknown-linux-android` instead of
`$ARCH-android-linux-gnu` to project servers
This is because of Android NDK specifies `$ARCH-linux-android` triplets and
users forget to specify `--with-boinc-platform`

Fix this in `boinc_platform.m4` so that Android boinc platform strings changes are not needed across all projects

boinc_alt_platform is excluded to prepare for 64bit only Android devices
Users can add back requesting 32bit tasks with `--with-boinc-alt-platform` during compile or `<alt_platform>` during runtime

**Alternate Designs**
Feedbacks are welcome.

**Release Notes**
N/A

### DIFF
--- a/m4/boinc_platform.m4
+++ b/m4/boinc_platform.m4
@@ -53,12 +53,14 @@ AC_DEFUN([BOINC_PLATFORM],[
         ;;
       aarch64*linux-android)
         boinc_platform="aarch64-android-linux-gnu"
+        boinc_alt_platform="arm-android-linux-gnu"
         ;;
       arm*linux-android*)
         boinc_platform="arm-android-linux-gnu"
         ;;
       x86_64*linux-android)
         boinc_platform="x86_64-android-linux-gnu"
+        boinc_alt_platform="x86-android-linux-gnu"
         ;;
       i686*linux-android)
         boinc_platform="x86-android-linux-gnu"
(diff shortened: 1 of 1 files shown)


AenBleidd:
I'm not sure whether these lines are correct or not. 
@davidpanderson, could you please take a look at it?
Thanks in advance

davidpanderson:
The official list of platform names is here:
https://boinc.berkeley.edu/trac/wiki/BoincPlatforms
These are all consistent with that.

AenBleidd:
@truboxl, could you please add similar logic to define `boinc_alt_platform`?

truboxl:
I think the intention is not to add them to prepare for 64bit only systems. If the devs think it should be included then ok I guess I will add them, Once these added I think can remove `--with-boinc-platform` and etc from build scripts...

AenBleidd:
You still need them in build scripts when cross-compiling for different platforms but if you build it for your native platform - then yes, no need to specify this parameter in build script

truboxl:
Ah yes you are right
I totally forgot needing to pass `--host=$ARCH-linux-android`
using that will automatically picks up the respective platform strings without needing overrides `--with-boinc-platform` and etc

This is inline with https://developer.android.com/ndk/guides/standalone_toolchain#projects_using_autoconf and https://github.com/curl/curl/blob/master/docs/INSTALL.md#android

PR #3753, opened 2020-05-22
 Client: let a project master URL change from http: to https: without involving the user.

### LINKED ISSUE #3744: "This project is using an old URL." notices persist in BOINC client.
When a project changes its URL, typically from http: to https:, you get a notice informing you that you should detach from the project and then add the project using the new URL.  This is shown in the "Notices" tab in the BOINC manager.  The exact message is "This project is using an old URL. When convenient, remove the project, then add <URL>".

When you detach and re-add the project, the message still is shown in the BOINC manager.  Closing the BOINC manager doesn't clear the message.  Closing the BOINC client will clear the message.

**Steps To Reproduce**
1. Have a message on a project saying that you need to change the URL.  If you don't currently have such a message, you *might* be able to create this scenario by attaching with http: instead of https:.  I'm not really certain if that will work, and it might depend on the apache configuration on the server.  
2. When you have the message in your notices, detach from the project.
3.  Add the project, using the new, correct, URL.
4. Observe that the notice about having a new URL is still present in the BOINC manager.

**Expected behavior**
After attaching with the correct URL, the notice should no longer show in the BOINC manager, or be reported by the BOINC client to an alternate manager such as BOINCTasks.

**System Information**
 - OS:  Windows 10
 - BOINC Version: 7.14.2

davidpanderson:
Fixes #3744

### DIFF
--- a/client/app_control.cpp
+++ b/client/app_control.cpp
@@ -69,6 +69,7 @@ using std::vector;
 #include "shmem.h"
 #include "str_replace.h"
 #include "str_util.h"
+#include "url.h"
 #include "util.h"
 
 #include "client_msgs.h"
@@ -1633,7 +1634,7 @@ void ACTIVE_TASK::read_task_state_file() {
         );
         return;
     }
-    if (strcmp(s, result->project->master_url)) {
+    if (!urls_match(s, result->project->master_url)) {
         msg_printf(wup->project, MSG_INTERNAL_ERROR,
             "wrong project URL in task state file"
         );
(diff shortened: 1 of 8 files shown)


squaregoldfish:
I believe this won't handle the URLs in other parts of BOINC's config. For example, I have the following in `cc_config.xml`:

```xml
<exclude_gpu>
    <url>http://www.gpugrid.net/</url>
    <device_num>1</device_num>
</exclude_gpu>
```
which wouldn't be picked up. When the master URL for the project changes, this directive will no longer be in effect.

davidpanderson:
Good point; I'll fix this.

PR #3777, opened 2020-05-27
Add subdirectory support to update_versions

Rytiss:
**Description of the Change**
BOINC client correctly handles subdirectories when downloading application files, however, update_versions expects directories to have a signature files, which is impossible.

This change skips signature check for directories and independently processes files in subdirectories denoted in the version.xml file. Files deployed to the download dir preserve directory structure.

**Release Notes**
Add subdirectory support for application versions.

### DIFF
--- a/tools/update_versions
+++ b/tools/update_versions
@@ -285,6 +285,9 @@ function process_files($a, $v, $p, $fds) {
     while ($f = readdir_aux($d)) {
         if ($f == "version.xml") continue;
         if (strstr($f, ".sig") == ".sig") continue;
+        if (is_dir('apps/'.$a.'/'.$v.'/'.$p.'/'.$f)) { // Skip folder structure, it will get processed later
+            continue;
+        }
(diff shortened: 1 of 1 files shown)


davidpanderson:
This can be simplified a bit.
In process_files(), skip directories as you're doing.
Then add a loop over $fds, for elements with / in the physical name,
and call process_files() for each of them.
That way you don't need to extra code in process_version().

Either way, this should be documented in the wiki.
It should be pointed out that files in subdirectories
but not listed in version.xml won't get added to the app version.

Rytiss:
Not sure I understand your comment: process_files() loops over the files in the directory, not the version.xml file, so there still needs to be a separate loop to process version.xml. It ends up being a matter of where the code to do that resides: inside process_files() or process_version(). I'm fine either way, so it can be moved into process_files() if that is what you want.

I'll update the documentation when the PR is merged so as not to have documentation for features that don't exist.

davidpanderson:
Ignore my comment - this is fine as is.

PR #3790, opened 2020-05-31
[Android] Initialize more variables using dependency injection.

Isira-Seneviratne:
**Description of the Change**
Initialize more class variables, such as `Monitor`'s `clientStatus`, `deviceStatus` and `noticeNotification` variables, using dependency injection, as this removes the need to implement the singleton pattern for these variables separately.

**Release Notes**
N/A


AenBleidd:
Please resolve merge conflicts

PR #3799, opened 2020-06-04
[Android] Fix crash on API levels < 21.

Isira-Seneviratne:
**Description of the Change**
Fix a crash that occurs on API levels lower than 21 (Lollipop) due to the use of vector drawables in the app.

**Release Notes**
Fix the app crashing on Android versions older than Lollipop.


AenBleidd:
Please resolve merge conflicts

PR #3804, opened 2020-06-05
switcher: build only for macOS

### LINKED ISSUE #3751: Drop building switcher binary everywhere in Makefiles
Probably a very low priority for cleaning up the Makefiles.
Need checking with projects.
Probably need a Mac to test...

Previous discussion:
https://github.com/BOINC/boinc/pull/2780
https://github.com/BOINC/boinc/issues?q=mac+switcher

truboxl:
Fixes #3751

**Description of the Change**
Clean up build files.

**Alternate Designs**
N/A

**Release Notes**
switcher executable is removed for all OS except macOS as it is only used on macOS.
Details on switcher usage on macOS can be found at https://boinc.berkeley.edu/sandbox_design.php

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1049,12 +1049,15 @@ if echo $host_os | grep '^darwin' >/dev/null ; then
   AS_IF([test "${enable_client}" = "yes" || test "${enable_manager}" = "yes"],
         [AC_MSG_WARN([
 ===============================================================================
-WARNING: Building BOINC for macOS through Autotools may not be secure.
+WARNING: Building BOINC client and manager for macOS through Autotools is
+         deprecated. See PR#3804.
 
-    Please refer boinc/mac_build/HowToBuildBOINC_XCode.rtf on how to build
-    BOINC for macOS.
+    Please refer to the docs below on how to build BOINC for macOS.
 
-    To continue building other BOINC components, please run with the option:
+    boinc/mac_build/HowToBuildBOINC_XCode.rtf
+    boinc/mac_build/HowToBuildBOINC_XCode.pdf
+
+    To disable this warning, please run with the option:
 
     ./configure --disable-client --disable-manager
 
--- a/configure.ac
+++ b/configure.ac
@@ -1046,6 +1046,21 @@ if echo $host_os | grep '^darwin' >/dev/null ; then
   if test `uname -r | sed 's/\.//g'` -lt 800 ; then
     AC_DEFINE_UNQUOTED(DARWIN_10_3, [1],[Define to 1 if compiling under OS X 10.3 or earlier])
   fi
+  AS_IF([test "${enable_client}" = "yes" || test "${enable_manager}" = "yes"],
+        [AC_MSG_WARN([
+===============================================================================
+WARNING: Building BOINC for macOS through Autotools may not be secure.
+
+    Please refer boinc/mac_build/HowToBuildBOINC_XCode.rtf on how to build
+    BOINC for macOS.
+
+    To continue building other BOINC components, please run with the option:
+
+    ./configure --disable-client --disable-manager
+
+===============================================================================
+        ])
+  ])
 fi
 
 AM_CONDITIONAL(OS_LINUX, [echo $host_os | grep '^linux' > /dev/null])
--- a/client/Makefile.am
+++ b/client/Makefile.am
@@ -25,7 +25,7 @@ if OS_DARWIN
 endif
 
 bin_PROGRAMS = boinc_client boinccmd
-if !OS_WIN32
+if OS_DARWIN
 bin_PROGRAMS += switcher
 endif
 


talregev:
Can we do it with CMake?
I want to start a process to convert compilation of Boinc with CMake.
Would you like to join my effort?

truboxl:
@talregev #1882

talregev:
Thanks for the Issue, 
But my question is still valid. Do you want to join the Cmake effort?
Maybe you have Cmake knowledge.

truboxl:
@talregev I can try. But I maybe pretty much a noob though.

talregev:
I don't afraid from a little knowledge, you can always learn from the internet.
If you have a little time and willing to help, jut write add me in discord.

configure.ac
```diff
@@ -1037,6 +1037,21 @@ if echo $host_os | grep '^darwin' >/dev/null ; then
   if test `uname -r | sed 's/\.//g'` -lt 800 ; then
     AC_DEFINE_UNQUOTED(DARWIN_10_3, [1],[Define to 1 if compiling under OS X 10.3 or earlier])
   fi
+  AS_IF([test "${enable_client}" = "yes" || test "${enable_manager}" = "yes"],
```
AenBleidd, line 1040: It is supported. It's still, let's say. not official, but there were put a lot of effort to add ability to build BOINC with Autotools. Also it's built now on Travis CI exactly using Autotools

truboxl: Unfortunately I don't have a Mac to test this. Its unlikely to work given how #2780 made the changes to build the missing components to "supposedly" work under Linux which wasn't meant to use there. When I make this commit, I am basing around https://github.com/BOINC/boinc/pull/2780#issuecomment-433253013

truboxl:
I think its fine now. Only client and manager part are blocked from building using Autotools for macOS. Building other components for macOS should still work using Autotools.

I use `--host=x86_64-apple-darwin` to test.

AenBleidd:
I strongly don't agree with this PR.
@brevilo, @CharlieFenton, @lfield, what's your opinion about this?

CharlieFenton:
I decline to get involved. My long time record of not supporting the use of Autotools, configure and make files for building BOINC on the Mac is well documented, though there are a number of other BOINC developers (including @brevilo) who have expressed strong feelings otherwise.

I have very limited understanding of the internals and syntax of these methods, so I don't know whether the proposed changes actually accomplish what is claimed. But if they do actually limit the use of Autotools, configure and make files on Macintosh to allow only building the BOINC libraries, then that seems to me to be an improvement.

truboxl:
I have removed the aggressive warnings. I will wait for the discussion from June to finalize then.

CharlieFenton:
Please leave in the aggressive warnings. While building the Mac client and manager from _autotools_ / _make_ may be possible, using those builds presents a security vulnerability, because they run without the sandbox security built into BOINC on the Mac when built according to the official [directions](https://github.com/BOINC/boinc/raw/master/mac_build/HowToBuildBOINC_XCode.pdf). Among 
* The client and manager will not be built with `SANDBOX` defined.
* The client, manager and BOINC data need to have special permissions set. This is done by the _PostInstall_ code in the  Mac installer , which is not built by any make files.
* The installer needs to be assembled from the various components by the script _mac_installer/release_boinc.sh_.

As far as I know, only one project (Einstein@home) is using _autotools_ / _make_ to generate Mac code for BOINC, and they are only building the BOINC libraries to link with their project applications. No one should have any reason to build the Mac client or manager that way. I have provided a different method to do it all from the command line for those so inclined, and it is described in the above referenced build directions.

truboxl:
As per CharlieFenton's advice, I add back the warning. I change the warning to be less aggressive, leaving the option to build macOS version of the client and manager through Autotools still possible, albeit the resulting build maybe insecure.

CharlieFenton:
> option to build macOS version of the client and manager through Autotools still possible, albeit the resulting build maybe insecure.

Actually, there are more shortcomings than that. For example, the version number won't be displayed correctly; please see [this comment](https://github.com/BOINC/boinc/pull/3296#issuecomment-533519975) and [this one](https://github.com/BOINC/boinc/pull/3296#issuecomment-533734492).

truboxl:
So there's people that did try to improve Autotools situation on macOS but mysteriously abandoned anyway. If using Makefiles is so problematic, untested, broken on macOS as proven by the unmerged Closed PRs, and even outright insecure, it's safe to say this method should have been blocked and removed.

Unfortunately, the project does not seem to allow removing codes and supports as many platforms as possible. Whether or not the contributors come back to do testing is questionable when the maintainer cannot test it themselves. I end up with situation like #3260 and make my own patches that probably will never see the light in upstream.

I want to make it clear again, this PR only affects building client and manager portion for macOS (and switcher for non-macOS platform which is a broken stub) through Autotools. A simple warning to direct future macOS contributors to use the recommended method of building the C&M.

ChristianBeer:
Close and reopen to trigger CI build

AenBleidd:
Closed in favor of #4854 that is more clean and easy solution

PR #3807, opened 2020-06-06
[Android] [Scripts] Remove exit from script

talregev:
Remove exit from script:
I think it not good thing that there is an `exit` in the end of the script. It never has, and also it will be exit if you in another window.


AenBleidd:
Thanks for the fix. Sorry I missed this in my another review

truboxl:
Duly noted, sorry for changing the behaviour of the script

PR #3808, opened 2020-06-07
Delete TODO and Checkin Notes Files

### LINKED ISSUE #3204: TODO_OLD and TODO - Safe to delete?
Starting with the file TODO_OLD, it looks like it is really outdated, being last updated in 2002. The TODO file looks like it hasn't been updated since 2007. I assume that since those files are that old, they aren't really needed to be referenced from. Even if for some reason we might need it later, we'll still have the files in the git history.

RoboticMind:
Fixes #3204

**Description of the Change**
Deletes 13 to 5 year old files with outdated todo information and old updates. These don't need to be in the repository since if they are ever needed again they can be obtained from git history. 

**Alternate Designs**
The files could have been moved to a directory for outdated files, but this still clutters the repository with information and clutters results when trying to search or git grep through the repo. It also still has these unneeded files taking up space in the repository.

**Release Notes**
N/A


CharlieFenton:
I **very** strongly object to deleting the old checkin_notes files. Much or all of this information is not available anywhere else, as it was several generations of source control systems before the switch to GIT. I have often found them very useful in refreshing my memory of why we did things a certain way, and for reviewing older changes. I would not want to have to search for them in the GIT history in order to access them.

Regarding the TODO files, I don't know whether or not they contain still relevant issues (some of which may have been overlooked or forgotten.) Before deleting them, they should be reviewed and new GITHub issues created for any items that are still relevant.

13 years may seem "old" to those who have been involved with BOINC for only a few years, but not to me, as one of the original developers, and as someone who has been writing computer code for 55 years. I was disappointed when we stopped using them, a I feel that in many cases they offer better information than the commit message in GIT. And the early source control systems offered little if any opportunity to attach comments to the commits (which at the time were called checkins.)

The "clutter" these few files add to the repository is inconsequential compared to all the other files present in the 411 MB repository for GIT master (on my Mac).

A better way to reduce the size of the basic repository might be to separate out areas that have little overlap, such as perhaps splitting off the Android and / or Java source files from the Linux/Mac/Windows C++ files.

> clutters results when trying to search or git grep through the repo

I consider their appearing in searches a **good** thing. A major reason for using a modern source control system like GIT is to preserve the development history.

davidpanderson:
These files stay.

adamradocz:
I agree with @RoboticMind that we need to do something with these files. What about the alternate design options? Moving them in a separate folder sounds a viable option.

AenBleidd:
I agree with David and Charlie here.
If we remove these files - it will it reduce the size of the repo since these files will stay in .git folder anyway.
Moving them to a separate folder will not help when using grep.
My suggestion is next: if these files annoy you-all just ignore them in a tool you're using for search.
Or start using Windos or Mac: then you will work with project structure that doesn't contain all these files.
These files are useful. I also sometimes search for some info there.
Regarding Jav files and Android, it's closely connected to native binaries written in c++ (client) and moving them to separate repository will require additional steps to synchronize releases. I'd left them also here because it's easier to work with them.

RoboticMind:
I would take moving these to a folder for old information over leaving them here because the other benefit is getting rid of 16 files in the root directory that are completely irrelevant when you are trying to understand the structure of the repo.

AenBleidd:
The structure of the repo has nothing common with these files that have self-descriptive names

RoboticMind:
Yes my point is that it doesn't have anything to do with the structure, and it takes up visual space that just distracts. To be more concrete, in my fork I went from having to scroll in the file explorer to not having to scroll at all. The checkin files have similar names and especially when scrolling, this similar name draws attention. These were some of the only files I really noticed when looking at the repo and they were files that I didn't need to look at. 

Furthermore, it decreases the shock factor of looking at this repo. The root directory is huge and it's something that will turn people away as they'll just scroll through it and get not even glance at any files and go away. It's like looking in a messy room. People will just open the door, see the mess and not come back. Sure they could maybe try and understand the messy room, but will they? A cleaner room makes people consider going in. I can tell you that I haven't looked at this project in a year largely from this because I don't even know where to begin since looking at so many files and folder makes me want to give up. The issue is that you can't see all the people who've just given up since they won't generally try and ask about how things are; they'll just leave.

AenBleidd:
People are coming and leaving everyday, and it's ok.
As I remember, you like comparison to linux.
In case of linux, you can't just clone the repo and start contributing. You need to learn for some time how the existing community works and why repo has its current structure.
Here we have the same situation: if you don't understand the structure and you don't want to spend some time to learn how this project organized and you don't even ask people to help you - then it's just a waste of time both for you and for us: we can't be nice for everyone.
We have two people that started to work on this project recently without any complain about these files. So the real reason is not with these files but with just people who don't want to learn and to be involved in the existing community.

CharlieFenton:
> getting rid of 16 files in the root directory that are completely irrelevant when you are trying to understand the structure of the repo.

These files are part of the documentation. Documentation is never irrelevant. Reading them might even help you understand the software.

As @AenBleidd explained, BOINC is a rather complex set of software, and yes, it has a fairly steep learning curve. Many of our contributors have gone through that; I'm afraid there are not many shortcuts. But asking questions in GitHub or on the developer email list is one of them. Reading the documentation is another. https://boinc.berkeley.edu/trac/wiki/TitleIndex and https://boinc.berkeley.edu/wiki/Special:AllPages are good places to start. Other BOINC developers may have additional suggestions.

Making 6 documentation files that are important to many of the developers less easily accessible won't make a significant dent in the learning curve.

I will add one more thought. We very much appreciate all volunteer help. But we do not appreciate people with little BOINC experience who can't take no for an answer and keep arguing with those of us who have been involved for years . You are wasting our time and our limited resources.

PR #3811, opened 2020-06-07
[Android][Compile][Scripts] Compile with `vfpv3-d16` flag

talregev:
Compile with `vfpv3-d16` flag.
To support running boinc cpp running on old devices with android api 19. (Android 4.4 Kitkat)

### DIFF
--- a/client/cpu_benchmark.h
+++ b/client/cpu_benchmark.h
@@ -32,9 +32,9 @@ extern bool benchmark_time_to_stop(int which);
 // separate different compilations of whetstone.cpp which will utilize
 // various ARM fp features ie neon, vfp, or "normal"
 #ifdef ANDROID
-#ifdef ANDROID_NEON
+#ifdef ANDROID_NEON1
 // add CXXFLAGS/CFLAGS for gcc:  -DANDROID_NEON -mfloat-abi=softfp -mfpu=neon
-#include <arm_neon.h>
+    #include <arm_neon.h>
 #endif // ANDROID_NEON
 
 namespace android_neon {
--- a/client/cs_benchmark.cpp
+++ b/client/cs_benchmark.cpp
@@ -183,12 +183,18 @@ int cpu_benchmarks(BENCHMARK_DESC* bdp) {
     //
     if (strstr(gstate.host_info.p_features, " neon ")) { 
         // have ARM neon FP capabilities
-        retval = android_neon::whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
+        #if defined(ANDROID_NEON1)
+            retval = android_neon::whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
+        #endif
     } else if (strstr(gstate.host_info.p_features, " vfp ")) { 
         // have ARM vfp FP capabilities
-        retval = android_vfp::whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
+        #if defined(ANDROID_VFP)
+            retval = android_vfp::whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
+        #endif
     } else { // just run normal test
-        retval = whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
+        #if !defined(ANDROID_NEON1) && !defined(ANDROID_VFP)
+            retval = whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
+        #endif
     }
 #else
     retval = whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
--- a/client/whetstone.cpp
+++ b/client/whetstone.cpp
@@ -41,7 +41,7 @@
 #endif
 
 #ifdef ANDROID
-#ifdef ANDROID_NEON
+#ifdef ANDROID_NEON1
     namespace android_neon {
 #else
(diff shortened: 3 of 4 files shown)


client/cpu_benchmark.h
```diff
@@ -32,9 +32,9 @@ extern bool benchmark_time_to_stop(int which);
 // separate different compilations of whetstone.cpp which will utilize
 // various ARM fp features ie neon, vfp, or "normal"
 #ifdef ANDROID
-#ifdef ANDROID_NEON
+#ifdef ANDROID_NEON1
```
AenBleidd, line 35: Where did you get this macro? I believe it's a wrong one ane using it you just disable a part of functionality that should be there

client/cpu_benchmark.h
```diff
@@ -32,9 +32,9 @@ extern bool benchmark_time_to_stop(int which);
...
 // add CXXFLAGS/CFLAGS for gcc:  -DANDROID_NEON -mfloat-abi=softfp -mfpu=neon
-#include <arm_neon.h>
+    #include <arm_neon.h>
```
AenBleidd, line 37: Wrong indent

client/cs_benchmark.cpp
```diff
@@ -183,12 +183,18 @@ int cpu_benchmarks(BENCHMARK_DESC* bdp) {
     //
     if (strstr(gstate.host_info.p_features, " neon ")) { 
         // have ARM neon FP capabilities
-        retval = android_neon::whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
+        #if defined(ANDROID_NEON1)
```
AenBleidd, line 186: This is redundant here since if platform has `neon` support it should test it.

client/cs_benchmark.cpp
```diff
@@ -183,12 +183,18 @@ int cpu_benchmarks(BENCHMARK_DESC* bdp) {
...
+            retval = android_neon::whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
+        #endif
     } else if (strstr(gstate.host_info.p_features, " vfp ")) { 
         // have ARM vfp FP capabilities
-        retval = android_vfp::whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
+        #if defined(ANDROID_VFP)
```
AenBleidd, line 191: This is redundant here since if platform has `vfp` support it should test it.

client/cs_benchmark.cpp
```diff
@@ -183,12 +183,18 @@ int cpu_benchmarks(BENCHMARK_DESC* bdp) {
...
+            retval = android_vfp::whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
+        #endif
     } else { // just run normal test
-        retval = whetstone(host_info.p_fpops, fp_time, MIN_CPU_TIME);
+        #if !defined(ANDROID_NEON1) && !defined(ANDROID_VFP)
```
AenBleidd, line 195: This is redundant here since platform has no `neon` or `vfp` support.

client/whetstone.cpp
```diff
@@ -41,7 +41,7 @@
 #endif
 
 #ifdef ANDROID
-#ifdef ANDROID_NEON
+#ifdef ANDROID_NEON1
```
AenBleidd, line 44: Wrong macro again

client/whetstone.cpp
```diff
@@ -290,6 +290,6 @@ int whetstone(double& flops, double& cpu_time, double min_cpu_time) {
     return 0;
 }
 
-#if defined(ANDROID_NEON) || defined(ANDROID_VFP)
+#if defined(ANDROID_NEON1) || defined(ANDROID_VFP)
```
AenBleidd, line 293: Same here

talregev:
@AenBleidd 
I change the macro name to ANDROID_NEON1 because it enable the ANDROID_NEON when I add these flags and I don't want it. Even when I put -DANDROID_NEON=OFF.
How it turn it off without change the name?

talregev:
@AenBleidd 
> This is redundant here since if platform has neon support it should test it.
> This is redundant here since platform has no neon or vfp support.

When I leave it intact it not compile, it looking for other implementation of the function.
But we compile it only once with one of the option:
- neon
- vfp
- none

In the same binary there is only one option, but in run time, you asking to call one of them. 
That why the compiler is tell you that it error. and it right.
Then if you only have one of the function compiled, this all section of the code is redundant and that why I suggest to remove it.
Unless you  can share with your wisdom and find other solution for this compilation error.

talregev:
Will you have time to compile it yourself and see the errors I am talking about?

truboxl:
@RichardHaselgrove you are absolutely right, maybe this code should be removed

talregev:
Maybe not. I am doing discussion with @AenBleidd about this.
Do you have a knowledge in this auto make tool?
Do you want to join the discussion?

talregev:
I understand it now, And I want to do the things like it was before?
Can you help?

truboxl:
In client/Makefile.am it currently uses -mfpu=vfp

Is this currently broken on Android 4.4?

talregev:
Is an neon optimization Issue. 
@truboxl Do you mind to talk with me in Discord?

talregev:
This is not good PR. I made a new one.

PR #3812, opened 2020-06-07
Move Todo and Checkins to a New Folder

RoboticMind:
Fixes #

**Description of the Change**
Moves the TODO and Checkin files to a folder called old_files. This helps clean up the root directory and make it so you can scroll less

**Alternate Designs**
Deleting the files was not seen as a viable option so moving the files to a new location is used instead.

**Release Notes**
N/A


AenBleidd:
@RoboticMind, maybe I could give you some ticket to work on instead of just moving files from one folder to another?

PR #3821, opened 2020-06-09
[Android] Remove neon optimization from arm.

talregev:
Remove neon optimization from arm, to support old devices (android 4.4 Kitkat, api 19)
Our arm compiler compile the code with neon optimization by default:
```bash
./boinc/3rdParty/buildCache/android-tc/arm/bin/arm-linux-androideabi-clang -E -dM - < /dev/null | grep -i neon
```
```c++
#define ARM_NEON 1
#define ARM_NEON_FP 0x4
#define ARM_NEON 1
```
This information brought to me by @truboxl.
I also verify it in different way.
The global flag: ` -mfpu=vfpv3-d16` disable neon optimization.
There was conflict with neon benchmark module that I solve.

Running on AVD 4.4:
running

### DIFF
--- a/android/build_boinc_arm.sh
+++ b/android/build_boinc_arm.sh
@@ -25,8 +25,8 @@ export PATH="$TCBINARIES:$TCINCLUDES/bin:$PATH"
 export CC=arm-linux-androideabi-clang
 export CXX=arm-linux-androideabi-clang++
 export LD=arm-linux-androideabi-ld
-export CFLAGS="--sysroot=$TCSYSROOT -DANDROID -DDECLARE_TIMEZONE -Wall -I$TCINCLUDES/include -O3 -fomit-frame-pointer -fPIE -march=armv7-a -D__ANDROID_API__=19"
-export CXXFLAGS="--sysroot=$TCSYSROOT -DANDROID -Wall -I$TCINCLUDES/include -funroll-loops -fexceptions -O3 -fomit-frame-pointer -fPIE -march=armv7-a -D__ANDROID_API__=19"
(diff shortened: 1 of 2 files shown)


talregev:
@truboxl

truboxl:
Note that client/Makefile.am isn't just used by Android, but also Linux, FreeBSD, Windows, maybe macOS (if there is), and not just clang but also gcc, then there's handling of building for different architectures too

So likely will fail Travis CI

talregev:
I will edit client/Makefile.am that it will add this flags only for arm android. there already if condition for it.

truboxl:
Its ugly, but it works.
Doesn't break on my 32bit Raspberry Pi too with no added `*FLAGS`.

However those want to specify `-mfpu=vfpv3-d16` for other ARM combinations may not get applied. They should don't use `-mfpu=vfpv3-d16` and use `-mfpu=vfp` or use aarch64 or patch themselves or add nothing special at all. We can't really cover all grounds at this point. And really its often questionable to add those optimisation beyond `-O3`.

I know there's a cleaner way to do this out there but for now this is what we come up with.
(maybe dropping selective benchmark entirely and use NEON all the way)

For a bit more context on this PR, it tries to **really** fix #1702 and replaces #1830. As talregev said, the default compiler compiles with NEON support by default (without any `-mfpu` flags) and this can break on non-NEON devices.

talregev:
I did it ugly because of the auto make tool.
I don't know how to solve it in a cleaner way. You can suggest another PR that solve this problem, I will not offended.
I choose -mfpu=vfpv3-d16 for two strong reasons:
From this link:
https://wiki.debian.org/ArmHardFloatPort/VfpComparison
> VFPv3-D16 is the common denominator of the processors to support here (therefore the recommended build option is -mfpu=vfpv3-d16)
Second reason that our curl and openssl arm already have this global flag in the script.

talregev:
@truboxl How should I talk with you online? Should I tag you here every time I want to talk with you in discord?

PR #3826, opened 2020-06-11
client/app interface: pass priority-related user preferences to wrappers

### LINKED ISSUE #3764: process priority flags don't work for wrapper tasks
**Describe the bug**
setting process_priority_special doesn't seem to work on the wrapper tasks at GPUGRID. setting this flag to '2', the tasks still run at nice 19. 

**Steps To Reproduce**
1. set the process_priority and process_priority_special flags in the cc_config.xml
2. observe no affect on wrapper gpu app tasks (GPUGRID)

**Expected behavior**
with process_priority_special, I would expect that the nice values for the wrapper tasks would be reflective of my setting. setting a value of '2' should implement nice 0, etc. 

**System Information**
 - OS: Ubuntu 18.04 & 20.04
 - BOINC Version: 7.17.0

The only project I run that uses wrappers is GPUGRID which uses a wrapper for their GPU applications, I don't know about other projects.

davidpanderson:
Fixes #3764

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/html/inc/submit.inc
+++ b/html/inc/submit.inc
@@ -1,7 +1,7 @@
 <?php
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2013 University of California
+// Copyright (C) 2018 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -29,9 +29,12 @@
 //      boinc_query_job(): get details of a job
 //      boinc_retire_batch(): retire a batch; delete output files
 //      boinc_submit_batch(): submit a batch
+//
+//  See https://boinc.berkeley.edu/trac/wiki/RemoteJobs#PHPinterface
 
-//// Implementation stuff follows
 
+// Convert a request message from PHP object to XML string
+//
 function req_to_xml($req, $op) {
     if (!isset($req->batch_name)) {
         $req->batch_name = "batch_".time();
@@ -52,6 +55,14 @@ function req_to_xml($req, $op) {
     }
     if ((isset($req->app_version_num)) && ($req->app_version_num)) {
         $x .= "    <app_version_num>$req->app_version_num</app_version_num>
+";
+    }
+    if (!empty($req->allocation_priority)) {
+        $x .= "    <allocation_priority/>
+";
+    }
+    if (isset($req->priority)) {
+        $x .= "    <priority>$req->priority</priority>
 ";
     }
     foreach ($req->jobs as $job) {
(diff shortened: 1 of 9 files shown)


AenBleidd:
@davidpanderson, I believe this commit  bed2cd45dbd59f12a310e83e48a49815b6f70ecd is not related to this particular PR.

RichardHaselgrove:
I see two different problems with the 'primary purpose' section of this PR (f257fbc1e48beb373b79d09aa75eed843514937e).

**1. Design**

The need identified in #3764 was to adjust the priority of a single GPU app. It was the GPU nature of the app which was primary: the fact that it had been deployed via a wrapper was an incidental impediment.

Because cc_config.xml is global across all projects attached to the client, a 'process_priority_special' designed for that GPU app will also be applied to any CPU wrapper apps that may have been issued by another project. It would have been more appropriate to design this into the 'Project-level configuration' area, perhaps even down to the app_version level.

**2. Implementation**

Wrapper app priorities in the client are set (as in this PR) at https://github.com/BOINC/boinc/blob/master/lib/proc_control.cpp#L264 ff, using definitions contained in https://github.com/BOINC/boinc/blob/master/lib/common_defs.h#L81. This defines a base-1 list of process priorities:
```
1: lowest (Win: IDLE; Unix: 19)
2: low (Win: BELOW_NORMAL; Unix: 10
3: normal (Win NORMAL; Unix: 0)
4: high (Win: ABOVE_NORMAL; Unix: -10)
5: highest (Win: HIGH; Unix: -16)
```
However, for native (non-wrapper) apps, the client uses different process control logic, found at https://github.com/BOINC/boinc/blob/master/client/app_start.cpp#L508 ff, which hard-codes a base-0 list of process priorities:
```
    case 0: return IDLE_PRIORITY_CLASS;
    case 1: return BELOW_NORMAL_PRIORITY_CLASS;
    case 2: return NORMAL_PRIORITY_CLASS;
    case 3: return ABOVE_NORMAL_PRIORITY_CLASS;
    case 4: return HIGH_PRIORITY_CLASS;
    case 5: return REALTIME_PRIORITY_CLASS;
```
The implementation in cc_config.xml will use the same user-selected values against the two different process priority lists, and the results will differ from their expectations in at least one of the two cases.

For this reason, I am reverting the change made this morning to the User Manual (which appears to have been prompted by this confusion over process priority lists): it is important (by definition) that the User Manual describes the **current** behaviour of the public-release client, not the future behaviour of a putative revised client.

RichardHaselgrove:
@davidpanderson: Please cast your eye back over your two commits:

317dc915b0e9b6bc4d512fe6a84579213fe6b614 (13 Oct 2014, for wrapper apps)
fcafba0a874839f8889214067daf9dbcba396252 (29 Sep 2015, for 'native' apps)

I think these two, taken together, are where the confusion has arisen.

For the record, I still feel that the User Manual, relating to current clients, should reflect the 2015 numbering for native apps.

davidpanderson:
Oops!  somehow this PR combined two unrelated things.  I split them into separate PRs.

PR #3835, opened 2020-06-12
Windows: fix many VS2019 compile warnings.  No functional changes.

davidpanderson:


lib/mfile.cpp
```diff
@@ -106,7 +106,7 @@ size_t MFILE::write(const void *ptr, size_t size, size_t nitems) {
     buf = (char *)realloc_aux( buf, len+(size*nitems)+1 );
     if (!buf) {
         fprintf(stderr,
-            "ERROR: realloc() failed in MFILE::write(); len %d size %lu nitems %lu\n",
+            "ERROR: realloc() failed in MFILE::write(); len %d size %zu nitems %zu\n",
```
dimhotepus, line 109: Do we still support MinGW as compiler? If so, it doesn't support `%zu` format specifier for `size_t`, at least without any changes: 
https://sourceforge.net/p/mingw-w64/wiki2/gnu%20printf/
https://stackoverflow.com/questions/44382862/how-to-printf-a-size-t-without-warning-in-mingw-w64-gcc-7-1

AenBleidd: We do

PR #3836, opened 2020-06-12
[Android][Scripts] Add android 4.1 api 16 support in cpp compilation.

talregev:
Add android 4.1 api 16 support in cpp compilation.
I will check on AVD 4.1 (api 16) and I will past here a screenshots.

### DIFF
--- a/android/build_androidtc_arm.sh
+++ b/android/build_androidtc_arm.sh
@@ -12,5 +12,5 @@ export ANDROID_TC="${ANDROID_TC:-$HOME/android-tc}"
 export ANDROID_TC_ARM="${ANDROID_TC_ARM:-$ANDROID_TC/arm}"
 
 if [ ! -d "$ANDROID_TC_ARM/arm-linux-androideabi" ]; then
-    "$NDK_ROOT/build/tools/make-standalone-toolchain.sh" --verbose --platform=android-19 --arch=arm --stl=libc++ --install-dir="$ANDROID_TC_ARM" "$@"
+    "$NDK_ROOT/build/tools/make-standalone-toolchain.sh" --verbose --platform=android-16 --arch=arm --stl=libc++ --install-dir="$ANDROID_TC_ARM" "$@"
 fi
--- a/android/build_boinc_arm.sh
+++ b/android/build_boinc_arm.sh
@@ -26,8 +26,8 @@ export PATH="$TCBINARIES:$TCINCLUDES/bin:$PATH"
 export CC=arm-linux-androideabi-clang
 export CXX=arm-linux-androideabi-clang++
 export LD=arm-linux-androideabi-ld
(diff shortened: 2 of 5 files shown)


davidpanderson:
Why is the change to filesys.cpp needed?

talregev:
Because in the ndk api 16 you have both `<sys/statvfs.h>` and `<sys/statfs.h>`
And the function for initate the state inside `<sys/statvfs.h>` is define if you have api 19 or aboove
` #if __ANDROID_API__ >= 19`
So the other one that fit is inside `<sys/statfs.h>`
So hot the file `filesys.cpp` is written, I need to undefined `HAVE_SYS_STATVFS_H`
that in that if else it will take the `HAVE_SYS_STATVS_H`

talregev:
Screenshot that is working on AVD 4.1
Capture

talregev:
@davidpanderson I am not sure why windows compilation is failed. I didn't touche the windows at all.

talregev:
Should I rebase with the master? it will fix it?

AenBleidd:
@talregev, it failed because of the timeout on CI.
During last couple of days I saw this issue several times.
I believe I need to implement cache mechanism on CI. I'll do this today.

truboxl:
This code seems to cover all Android version though...
Why not use #if __ANDROID_API__ lesser or greater than version number here like you have suggested?
Also don't forget Android 4.1 has i686 too...

talregev:
@truboxl It not for all android code. the x64 bit will use the HAVE_SYS_STATVFS_H the one is undefined. 
I will add x86 api 16 support after this code review will accepted.

talregev:
@AenBleidd 
After @truboxl  note. I can change the If to:
`#if defined(ANDROID) && __ANDROID_API__ < 19`
Is it better?

talregev:
I will change it.

AenBleidd:
@talregev, yes

adamradocz:
Why do we bother supporting Android 4.1?

talregev:
The manger also will support api 16.

adamradocz:
I appreciate your work, but I don't see any benefits. Devices with Android 4.1 usually very old and barely can compute anything because they have very low RAM and slow CPU.

truboxl:
Actually looking back (I didn't look closely), the change involves libs which are used by projects to create their science apps. Is it ok to do this? statvfs calls are used on WCG's MCM which causes them to recommend users to have at least Android 4.4 because statvfs calls didn't appear before 4.4 (API 19) as per https://android.googlesource.com/platform/bionic/+/master/libc/libc.map.txt

Older devices who tried to request and run the tasks will result in error tasks stating
```
<![CDATA[
<message>
process exited with code 1 (0x1, -255)
</message>
<stderr_txt>
CANNOT LINK EXECUTABLE: cannot locate symbol "statvfs" referenced by "../../projects/www.worldcommunitygrid.org/wcgrid_mcm1_map_7.53_arm-android-linux-gnu"...

</stderr_txt>
]]>
```

If our goal is to support running BOINC clients on older devices all the way to 4.1, its already achieved. There should be some API to stop projects sending out incompatible tasks by Android API version like how some projects send out PIE and non PIE version of tasks, and NEON and non NEON version of tasks, etc.
https://github.com/BOINC/boinc/blob/59c476125369c3d33855b170f06ab4a435f4a056/win_build/installerv2/redist/all_projects_list.xml#L214-L217

PR #3837, opened 2020-06-12
[Android] Reintroduce support for API level 16 and higher.

Isira-Seneviratne:
**Description of the Change**
Make the changes necessary to support API level 16 (4.1 Jelly Bean) and higher.

**Release Notes**
The app now supports Android 4.1 and higher instead of only 4.4 and higher.


talregev:
Verify it working on AVD 4.1

PR #3838, opened 2020-06-13
Add Error On Sqlite3 Check Fail

RoboticMind:
**Description of the Change**
Currently if configure.ac doesn't find sqlite and needs it, it doesn't do anything. Then when you build BOINC it fails several minutes in. This makes it so that it errors while configuring after failing to find sqlite.

**Alternate Designs**
I don't know automake well enough to know of alternative ways to implement this

**Release Notes**
N/A

### DIFF
--- a/m4/sqlite3.m4
+++ b/m4/sqlite3.m4
@@ -47,6 +47,12 @@ if test "x$have_sqlite3" = "xyes"; then
     AC_DEFINE([HAVE_SQLITE3], [1], [Define to 1 if you have sqlite3.])
     AC_SUBST(SQLITE3_LIBS)
     AC_SUBST(SQLITE3_CPPFLAGS)
+else
+  AC_MSG_ERROR([
+================================================================================
+ERROR: could not find sqlite3
+================================================================================
+  ])
 fi
 
 LIBS=$LIBS_save


talregev:
@RoboticMind Do you know Cmake?

RoboticMind:
Somewhat. Is there a place where cmake is used in the Linux builds? I wouldn't be able to test very well if there's something I need to change for Windows or MacOS

talregev:
I want to start a process to compile boinc with Cmake. Will you have time to help?

talregev:
let's start only on Linux and WSL.

AenBleidd:
I'd say, do this for linux and only. We can still build BOINC for Windows using mingw, but it's not an official way for building it. Of course, it would be nice to keep it but not replace VS2019 projects with cmake

talregev:
With Cmake you can generate a visual studio solutions that compile your code.

talregev:
But not go too far with this dream. let's start on linux.

AenBleidd:
I know. And also I know that now VS could run cmake projects without converting them.
But solution for Windows is just perfect enough and there is no need to convert it to another tool.
BTW, for cmake you can look at vcpkg to manage dependencies (we are also using it now on VS2019)

talregev:
@RoboticMind Let me know if you want to help. You can contact me also in discord.

AenBleidd:
sqlite dependency is removed. this fix is not needed anymore.
@RoboticMind, thank you for your contribution anyway

PR #3840, opened 2020-06-13
[Android] Fix debug build crashing on Android versions < Q.

Isira-Seneviratne:
**Description of the Change**
Fix a crash that occurs on startup in the debug build on Android versions older than Q. This does not affect the release build: https://stackoverflow.com/questions/9126567/method-not-found-using-digestutils-in-android.

**Release Notes**
N/A


AenBleidd:
I believe this was already fixed in #3837
Please check and close if true otherwise resolve the conflict

PR #3849, opened 2020-06-18
[Android] Use the Jetpack navigation component.

Isira-Seneviratne:
**Description of the Change**
* Rewrite the app to make use of the Jetpack navigation component.
* Add a "View details" option to the project menu to allow users to view the project details, like with the old navigation drawer.

**Release Notes**
* The app now conforms to Material Design guidelines in the navigation drawer.
* A "View details" option has been added to the project menu.

**Screenshots**
Screenshot_2020-06-21-10-09-20-281_edu berkeley boinc
Screenshot_2020-06-21-10-09-26-190_edu berkeley boinc
Screenshot_2020-06-21-10-09-32-965_edu berkeley boinc
Screenshot_2020-06-21-10-09-40-047_edu berkeley boinc
Screenshot_2020-06-21-10-11-23-886_edu berkeley boinc


AenBleidd:
Maybe provided screenshot is outdated, but according to it I can say that 'Add project' menu has no indent. As soon it is a logical submenu of 'Projects' it should have indend.
This is how it looks like on existing version:
image
Please fix this or attach new screenshot if it was already fixed

Isira-Seneviratne:
I tried indenting it, but it didn't work.

AenBleidd:
Hm, could this be then done as a submenu that is just expanded by default?
It's really nice to have this 'Add project' item along with every attached project as a subitems of Projects item.

Isira-Seneviratne:
> 
> 
> Hm, could this be then done as a submenu that is just expanded by default?

I tried setting it as a submenu, but it resulted in the Projects item being unselectable due to it being displayed as a category.

Maybe it could be removed entirely? It could be implemented as a floating action button instead, or alternatively the user could navigate to the Projects fragment and then add a new project from there.

> It's really nice to have this 'Add project' item along with every attached project as a subitems of Projects item.

Unfortunately, it doesn't appear to be possible to add all the attached projects. I tried that and ran into an issue where the same project would get continuously added.

AenBleidd:
This should be done in some way. I can't accept such obvious regression

Isira-Seneviratne:
> 
> 
> This should be done in some way. I can't accept such obvious regression

How about a floating action button?

AenBleidd:
@Isira-Seneviratne, how this UI element looks like?

AenBleidd:
But you still need to put the list of projects in the menu so adding this component is not a solution

Isira-Seneviratne:
> 
> 
> But you still need to put the list of projects in the menu so adding this component is not a solution

I guess I'll close this PR, then.

PR #3852, opened 2020-06-19
[Android] Set small notification icons to white for android 4.4 and below.

talregev:
Set small notification icons to white for android 4.4 and below,
to be the same as android 5 and above.


talregev:
Some screenshots:
not1
not2

Crunch on AVD 4.1:
crunch

talregev:
Notice notification on AVD 4.1
not3
not4

PR #3861, opened 2020-06-23
[Android] Use RecyclerView in TasksFragment.

Isira-Seneviratne:
**Description of the Change**
Use the `RecyclerView` widget in TasksFragment instead of `ListView`, as it is more flexible than `ListView`.

**Release Notes**
N/A

**Screenshots**
Screenshot_2020-07-02-04-46-42-039_edu berkeley boinc


Isira-Seneviratne:
> 
> 
> On the screenshot I see no cancel button

Updated the screenshot.

PR #3880, opened 2020-07-03
[Android][Scripts] Upgrade android scripts to ndk r-21d

talregev:
Important: Please delete the old ndk zip in tmp folder before you try local:
`rm -rf /tmp/ndk.zip`

Upgrade android scripts to ndk r-21d.
Remove toolchain. deprecated in ndk21.

### DIFF
--- a/android/buildAndroidBOINC-CI.sh
+++ b/android/buildAndroidBOINC-CI.sh
@@ -134,29 +134,21 @@ fi
 
 export COMPILEOPENSSL="no"
 export COMPILECURL="no"
-export NDK_FLAGFILE="$PREFIX/NDK-${NDK_VERSION}_done"
+export NDK_FLAGFILE="$PREFIX/NDK-${NDK_VERSION}-${arch}_done"
 export CURL_FLAGFILE="$PREFIX/curl-${CURL_VERSION}-${arch}_done"
 export OPENSSL_FLAGFILE="$PREFIX/openssl-${OPENSSL_VERSION}-${arch}_done"
 export CREATE_NDK_FOLDER=${CREATE_NDK_FOLDER:-"no"}
 
-createNDKFolder()
-{
+if [ ! -e "${NDK_FLAGFILE}" ]; then
+    rm -rf $OPENSSL_FLAGFILE
+    rm -rf $CURL_FLAGFILE
     if [ $CREATE_NDK_FOLDER = "no" ]; then
         rm -rf "$BUILD_DIR/android-ndk-r${NDK_VERSION}"
         wget -c --no-verbose -O /tmp/ndk.zip https://dl.google.com/android/repository/android-ndk-r${NDK_VERSION}-linux-x86_64.zip
-        unzip -qq /tmp/ndk.zip -d $BUILD_DIR
+        unzip -qq /tmp/ndk_${NDK_VERSION}.zip -d $BUILD_DIR
         export CREATE_NDK_FOLDER="yes"
     fi
-}
-
-if [ ci = "yes" ]; then
-    createNDKFolder
-else
-    if [ ! -e "${NDK_FLAGFILE}" ]; then
-        export CREATE_NDK_FOLDER="no"
-        createNDKFolder
-        touch "${NDK_FLAGFILE}"
-    fi
+    touch "${NDK_FLAGFILE}"
 fi
 
 export NDK_ROOT=$BUILD_DIR/android-ndk-r${NDK_VERSION}
--- a/android/buildAndroidBOINC-CI.sh
+++ b/android/buildAndroidBOINC-CI.sh
@@ -153,6 +153,7 @@ if [ $ci = "yes" ]; then
     createNDKFolder
 else
     if [ ! -e "${NDK_FLAGFILE}" ]; then
+        export CREATED_NDK_FOLDER="no"
         createNDKFolder
         touch "${NDK_FLAGFILE}"
     fi
--- a/android/buildAndroidBOINC-CI.sh
+++ b/android/buildAndroidBOINC-CI.sh
@@ -27,7 +27,6 @@ export ANDROID_HOME=$HOME/Android/Sdk
 export NDK_ROOT=$HOME/Android/Ndk
 export ANDROID_TC=$HOME/Android/Toolchains
 
-
 # checks if a given path is canonical (absolute and does not contain relative links)
 # from http://unix.stackexchange.com/a/256437
 isPathCanonical() {
@@ -114,7 +113,7 @@ fi
 
 mkdir -p "${PREFIX}"
 mkdir -p "${BUILD_DIR}"
-
+echo $PREFIX
 if [ "${doclean}" = "yes" ]; then
     echo "cleaning cache"
     rm -rf "${PREFIX}"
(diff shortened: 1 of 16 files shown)


android/buildAndroidBOINC-CI.sh
```diff
@@ -129,18 +134,31 @@ fi
...
 
-if [ ! -e "${NDK_FLAGFILE}" ]; then
-    rm -rf "$BUILD_DIR/android-ndk-r${NDK_VERSION}"
-    rm -rf "${PREFIX}/${arch}"
-    rm -f "${CURL_FLAGFILE}" "${OPENSSL_FLAGFILE}"
-    wget -c --no-verbose -O /tmp/ndk.zip https://dl.google.com/android/repository/android-ndk-r${NDK_VERSION}-linux-x86_64.zip
-    unzip -qq /tmp/ndk.zip -d $BUILD_DIR
-    touch "${NDK_FLAGFILE}"
+if [ ci = "yes" ]; then
+    createNDKFolder
```
AenBleidd, line 153: Looks like every time on CI cache folder will be deleted before each run

android/build_boinc_arm.sh
```diff
@@ -15,25 +15,23 @@ VERBOSE="${VERBOSE:-no}"
...
+export NDK_ROOT=${NDK_ROOT:-$HOME/Android/Ndk}
 export ANDROID_TC="${ANDROID_TC:-$HOME/android-tc}"
 export ANDROIDTC="${ANDROID_TC_ARM:-$ANDROID_TC/arm}"
-export TCBINARIES="$ANDROIDTC/bin"
+export TOOLCHAINROOT="$NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64/"
+export TCBINARIES="$TOOLCHAINROOT/bin"
 export TCINCLUDES="$ANDROIDTC/arm-linux-androideabi"
-export TCSYSROOT="$ANDROIDTC/sysroot"
-export STDCPPTC="$TCINCLUDES/lib/armv7-a/libstdc++.a"
+export TCSYSROOT="$TOOLCHAINROOT/sysroot"
```
AenBleidd, line 24: Before this change `TCSYSROOT` target was folder with cache. After this change everything will be installed outside of cache directory

android/travis_build_all.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+set -e
+
+#
+# See: https://boinc.berkeley.edu/trac/wiki/AndroidBuildClient
+#
+
+# Script to compile everything BOINC needs for Android
+
+./buildAndroidBOINC-CI.sh --cache_dir "$ANDROID_TC" --build_dir "$BUILD_DIR" --silent --ci --arch arm
```
AenBleidd, line 10: `ANDROID_TC` and `BUILD_DIR` are undefined here

android/buildAndroidBOINC-CI.sh
talregev: No. The whole run in the CI it will create only once.
Inside the function there is a check that insure it.

android/build_boinc_arm.sh
talregev: Before the change TCINCLUDES is the target inside the cache folder.
After the change, also TCINCLUDES is the target inside the cache folder.

android/travis_build_all.sh
talregev: Yes. it like was inside the travis. and it like the build_all.sh.
Just recently I define BUILD_DIR in build_all.sh to be the cache folder. but it not good for the CI.
Inside buildAndroidBOINC-CI.sh it define the ANDROID_TC and BUILD_DIR like always.

talregev: If you look on the build_all.sh script. you will see it copy past from there, just without the BUILD_DIR that i define and also more parameters for the ci.

android/buildAndroidBOINC-CI.sh
AenBleidd: What check? `CREATE_NDK_FOLDER` ?
If yes - then between two runs (e.g. two commits) this folder will be rectreated again. Because on every run it will check that it is on ci, `CREATE_NDK_FOLDER` by default is `no` because no information about the state of this variable was saved from the previous run, and cache will be deleted

android/build_boinc_arm.sh
AenBleidd: I'm talking here about `TCSYSROOT`

android/travis_build_all.sh
AenBleidd: I mean, the real command that will be run is next:
`./buildAndroidBOINC-CI.sh --cache_dir "" --build_dir "" --silent --ci --arch arm`
because these variables are not defined

android/build_boinc_arm.sh
talregev: TCSYSROOT is not a target, never was.

android/travis_build_all.sh
talregev: Yes. I understand. It the same without run with these params: `--cache_dir` `--build_dir`.
You want me to remove it.
I just wanted to be the same like `build_all.sh` script.

android/buildAndroidBOINC-CI.sh
talregev: This is not create the standalone tool like it was before. 
Before the change it was downloading the all ndk folder very run just for compile boinc.  
After the change it do the same. It not cache the tool chain, because there isn't standalone tool chain to cache like before.
It use the ndk folder to  compile without create any tool chain standalone tool.

android/travis_build_all.sh
talregev: If in the future we will want to define `ANDROID_TC `or `BUILD_DIR` in the script, it will be more easy.

android/build_boinc_arm.sh
talregev: Everything in this change install inside the cache folder, in separate folder for each platform according to `TCINCLUDES ` like before this change.

android/buildAndroidBOINC-CI.sh
talregev: Every run it download the ndk folder once. It don't save it between runs.

AenBleidd: And every run we will rebuild openssl and curl.
This need to be fixed in some way

android/build_boinc_arm.sh
AenBleidd: Did you check this? I'm pretty much sure that there will be nothing inside cache directory after `make install`

talregev: I will check it again.

talregev: run travis_build_all.sh in my wsl2 after delete cache folder:
travis_build_all

Here is a result from openssl arm just finish:
libcrypto


I can wait and show you all the content of the folder when it finish if you want to.

android/buildAndroidBOINC-CI.sh
talregev: No, we don't need to rebuild openssl and curl because it save in the cache. 
It just download the ndk folder for boinc compilation, compile boinc with openssl and curl from the cache.

talregev: In the next version of the ndk, when they reduce the size of the ndk zip file lower then 1G , I will cache the ndk folder too.

PR #3881, opened 2020-07-03
[Android] Alter some strings to use placeholders.

Isira-Seneviratne:
**Description of the Change**
Make use of placeholders in some string resources.

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/.idea/misc.xml
+++ b/android/BOINC/.idea/misc.xml
@@ -39,5 +39,5 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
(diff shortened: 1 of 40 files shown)


Ageless93:
One request, if there are new strings that these will be pushed to Transifex ahead of full testing the new Android client, so that us translators can translate these to our languages and these can be included in the new version.

Isira-Seneviratne:
> 
> 
> One request, if there are new strings that these will be pushed to Transifex ahead of full testing the new Android client, so that us translators can translate these to our languages and these can be included in the new version.

There aren't any new strings, but some existing ones have been modified.

Ageless93:
Even if existing ones have been modified, they need to be translated again as the meaning may have changed.

Isira-Seneviratne:
> 
> 
> Even if existing ones have been modified, they need to be translated again as the meaning may have changed.

So how should the changes be pushed to Transifex?

AenBleidd:
In suitable time I'll ask David to do this

PR #3889, opened 2020-07-04
[Android][Scripts] Remove neon optimization from openssl

talregev:
Remove neon optimization from openssl.
Check inside build_all.sh and travis_build_all.sh for libs if there is neon optimization. 

ARM compiler is compile by default with neon optimization. 
My change is add CFLAGS to openssl. and with these flags it cancel neon optimization.

### DIFF
--- a/android/build_all.sh
+++ b/android/build_all.sh
@@ -17,7 +17,7 @@ cd android/
 ./buildAndroidBOINC-CI.sh --cache_dir "$ANDROID_TC" --build_dir "$BUILD_DIR" --arch x86
 ./buildAndroidBOINC-CI.sh --cache_dir "$ANDROID_TC" --build_dir "$BUILD_DIR" --arch x86_64
 
-array_libs=("libcrypto.a" "libssl.a" "libcurl.a")
+declare -a array_libs=("libcrypto.a" "libssl.a" "libcurl.a")
 
 for i in "${array_libs[@]}"; do
     if [[ $(readelf -A $(find $ANDROID_TC/arm  -name "$i") | grep -i neon | head -c1 | wc -c) -ne 0 ]]; then
--- a/android/build_all.sh
+++ b/android/build_all.sh
@@ -17,8 +17,14 @@ cd android/
 ./buildAndroidBOINC-CI.sh --cache_dir "$ANDROID_TC" --build_dir "$BUILD_DIR" --arch x86
 ./buildAndroidBOINC-CI.sh --cache_dir "$ANDROID_TC" --build_dir "$BUILD_DIR" --arch x86_64
 
+if [ ! -d ../3rdParty/buildCache/android-tc/arm ]; then
+    echo You run this script from diffrent folder: $PWD
+    echo Please run it from boinc/android
+    exit 1
+fi
+
 for i in "libcrypto.a" "libssl.a" "libcurl.a"; do
-    if [[ $(readelf -A $(find $ANDROID_TC/arm  -name "$i") | grep -i neon | head -c1 | wc -c) -ne 0 ]]; then
+    if [ $(readelf -A $(find ../3rdParty/buildCache/android-tc/arm  -name "$i") | grep -i neon | head -c1 | wc -c) -ne 0 ]; then
         echo "$i" is with neon optimization
         exit 1
     fi
--- a/android/build_all.sh
+++ b/android/build_all.sh
@@ -23,7 +23,9 @@ if [ ! -d ../3rdParty/buildCache/android-tc/arm ]; then
     exit 1
 fi
 
-for i in "libcrypto.a" "libssl.a" "libcurl.a"; do
+list_libs="libcrypto.a libssl.a libcurl.a"
+
+for i in $list_libs; do
     if [ $(readelf -A $(find ../3rdParty/buildCache/android-tc/arm  -name "$i") | grep -i neon | head -c1 | wc -c) -ne 0 ]; then
         echo "$i" is with neon optimization
         exit 1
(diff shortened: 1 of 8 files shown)


android/build_openssl_arm.sh
```diff
@@ -44,6 +44,9 @@ if [ "$COMPILEOPENSSL" = "yes" ]; then
     fi
     if [ -n "$CONFIGURE" ]; then
         ./Configure linux-generic32 no-shared no-dso -DL_ENDIAN --openssldir="$TCINCLUDES" 1>$STDOUT_TARGET
+        #override flags in Makefile
```
AenBleidd, line 47: ad blank after `#`

android/clear_environment.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+set -e
+
+# Script to clear environment BOINC for Android
```
AenBleidd, line 4: Verify that this script run from the correct folder, otherwise fail

android/travis_build_all.sh
```diff
@@ -12,4 +12,15 @@ set -e
 ./buildAndroidBOINC-CI.sh --cache_dir "$ANDROID_TC" --build_dir "$BUILD_DIR" --silent --ci --arch x86
 ./buildAndroidBOINC-CI.sh --cache_dir "$ANDROID_TC" --build_dir "$BUILD_DIR" --silent --ci --arch x86_64
 
+echo ANDROID_TC=$ANDROID_TC
```
AenBleidd, line 15: Remove this

android/clear_environment.sh
talregev: Can you suggest how to do it?

AenBleidd: https://github.com/BOINC/boinc/blob/005ace950810d8c467c6c06a7dca66e368506c10/3rdParty/buildLinuxDependencies.sh#L44-L48

talregev: Thanks! that help! :)

android/build_all.sh
```diff
@@ -17,4 +17,19 @@ cd android/
...
+    echo You run this script from diffrent folder: $PWD
+    echo Please run it from boinc/android
+    exit 1
+fi
+
+list_libs="libcrypto.a libssl.a libcurl.a"
+
+for i in $list_libs; do
+    if [ $(readelf -A $(find ../3rdParty/buildCache/android-tc/arm  -name "$i") | grep -i neon | head -c1 | wc -c) -ne 0 ]; then
+        echo "$i" is with neon optimization
```
AenBleidd, line 30: `"$i" is with neon optimization` => '[ERROR] "$i" contains neon optimization'

android/clear_environment.sh
```diff
@@ -0,0 +1,16 @@
...
+
+if [ ! -d ../../boinc/android ]] ; then
+    echo You run this script from diffrent folder: $PWD
+    echo Please run it from boinc/android
+    exit 1
+fi
+
+rm -rf ../3rdParty/buildCache/
```
AenBleidd, line 12: What is the purpose to remove everything from this folder? It could contain cache for different architectures.
I'd remove only ${PREFIX} folder but it should be already cleaned up from `android/buildAndroidBOINC-CI.sh` script

android/travis_build_all.sh
```diff
@@ -12,4 +12,19 @@ set -e
...
 
+if [ ! -d ../3rdParty/buildCache/android-tc/arm ]; then
+    echo You run this script from diffrent folder: $PWD
+    echo Please run it from boinc/android
+    exit 1
+fi
+
+list_libs="libcrypto.a libssl.a libcurl.a"
+
+for i in $list_libs; do
```
AenBleidd, line 23: Since this particular piece of code is needed in two separate scripts, maybe it's reasonable to move it to `android/buildAndroidBOINC-CI.sh`?

android/build_all.sh
```diff
@@ -17,4 +17,19 @@ cd android/
...
+if [ ! -d ../3rdParty/buildCache/android-tc/arm ]; then
+    echo You run this script from diffrent folder: $PWD
+    echo Please run it from boinc/android
+    exit 1
+fi
+
+list_libs="libcrypto.a libssl.a libcurl.a"
+
+for i in $list_libs; do
+    if [ $(readelf -A $(find ../3rdParty/buildCache/android-tc/arm  -name "$i") | grep -i neon | head -c1 | wc -c) -ne 0 ]; then
```
AenBleidd, line 29: This path could be undefined if `--cache_dir` is used to target any other folder

talregev: I can add another check.
If the ANDROID_TC is define, you it. otherwise you this.
That ok?

talregev: Fix it.

AenBleidd: No, you should ${PREFIX} always because this could be any location. If this path is set by default this doesn't mean that it will never change in future

talregev: ${PREFIX} is not define here. ${PREFIX} is define by this $ANDROID_TC in buildAndroidBOINC-CI.sh
I think if I use here a ${PREFIX} it always be undefined. 
Because  buildAndroidBOINC-CI.sh is doing this line: `export ANDROID_TC=$PREFIX`
And $ANDROID_TC here undefined. 
So it better like I am doing.

android/clear_environment.sh
AenBleidd: I'm afraid, this is still unresolved:
- here we have path to cache that could be changed by parameter
- in case if this folder contains cache of other architectures, it will be removed too, and I don't think that it's good

AenBleidd: Maybe the content of this file could be safely moved to `android/buildAndroidBOINC-CI.sh` ? then it can safely use set paths.

talregev: This is a local script. It resolve many problem in one click.
I think it not should merge with the clean.
Just leave it intact like it, and if someone has a problem, it will help it to fix it.

talregev: i can remove it from do clean from android/buildAndroidBOINC-CI.sh

AenBleidd:
MacOS build fail is not related to this PR

PR #3895, opened 2020-07-09
[linux][cuda] Fix CUDA version number truncated

AenBleidd:
Change minor version to 99 if actual minor version is > 99

This fixes #3893

Signed-off-by: Vitalii Koshura <lestat.de.lionkur@gmail.com>

PR #3896, opened 2020-07-10
When parsing NVIDIA driver version, max minor version with 99.

davidpanderson:
simpler version of #3895


AenBleidd:
Cleaner PR #3897 is made

PR #3899, opened 2020-07-11
[Android][Scripts] armv6 support

talregev:
armv6 support

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ jobs:
 
 before_install:
    - if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then ( sudo apt-get -qq update ) fi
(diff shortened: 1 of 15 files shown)


talregev:
armv6

talregev:
Screenshot_2020-07-11-08-38-45
Screenshot_2020-07-11-08-39-38
Screenshot_2020-07-11-08-40-40

AenBleidd:
Please remove this commit: 16134748ac4cecb3ea87787e638a7df3d2003d62

talregev:
@AenBleidd Removed.

truboxl:
Here's my take on this.

Generally speaking if no project wants to add support for Android armv6, having a BOINC client for it is a waste of time.
Also all, Android 32bit has an expire date in approx. 18 years from now aka Year 2038 bug.
https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md

Its unlikely Google will fix this and even so, OEM will not go back creating updates for version 6.x and below Android phones.

talregev:
@AenBleidd Fix conflicts with master.

AenBleidd:
After careful consideration I decided to close this PR.
Main reason: we want to target newer devices more (the same as Projects do). There are too few ARMv6 based devices nowadays and they are too slow in comparison with newest ARM CPUs.

PR #3900, opened 2020-07-11
Client: if AM reply includes a project we're attached to under a different account,  honor the params in the AM reply, e.g resource share.

### LINKED ISSUE #3877: Attaching to a Project using an Account Manager, does not use custom Resource Share from Account Manager Host settings
**Describe the bug**
Attaching to a Project using an Account Manager, does not use custom Resource Share from Account Manager Host settings

**Steps To Reproduce**
I have an Account Manager (AM) setup, and configured for my hosts.
I also have custom Resource share values, for each of the projects for each of the hosts.
If the Host is attaching to the project from the AM, it does not use the custom Resource Share.
This often catches me by surprise, because it's not using the settings that I carefully defined!

As a workaround, I have to change the AM host project resource share to some other value, sync, change it back, sync.
A needless wasteful pain.

Reproduced using BAM!, 7/1/2020, on BOINC 7.16.7.

**Expected behavior**
When attaching via AM, the custom Resource Share from the AM Host settings should be respected.

davidpanderson:
Also: fix WxWidgets runtime warnings about spacers.

Possibly fixes #3877

### DIFF
--- a/client/acct_mgr.cpp
+++ b/client/acct_mgr.cpp
@@ -672,12 +672,11 @@ void ACCT_MGR_OP::handle_reply(int http_op_retval) {
                                     "Received new authenticator from account manager"
                                 );
                             } else {
-                                // otherwise skip this update
+                                // otherwise keep using the old one
                                 //
                                 msg_printf(pp, MSG_INFO,
                                     "Already attached to a different account"
                                 );
-                                continue;
                             }
                         }
                     }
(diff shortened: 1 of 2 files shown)


AenBleidd:
@JacobWKlein, is it possible for you to verify this fix?

JacobWKlein:
Any chance that we can get a "preview release" of BOINC with the fix, for me to Test? That's how I prefer to test these things, and I'm not setup to compile my own.

RichardHaselgrove:
For Windows, look at the line for 'AppVeyor' in the list of checks below. Click 'details' in the right-hand column.

There will be a list of job names below. I personally prefer VS2013, version 12.0, x64 build. Check the version carefully - I don't think you can rely on them always being in the same order.

Click on the one you feel is most appropriate. Double-check that the build log looks plausible. Above the log. click 'artifacts' (far right). Three downloads are listed: win-client, win-manager, win-apps. You want the first of those for this test, but grab whatever you want.

Stop the current boinc, backup the existing program files, replace them with the (extracted!) download files, and restart BOINC for testing.

Those instructions assume that you have MS Visual C runtime libraries for VS 2013 already installed (most Windows installations have those by now). Ask if problems.

PR #3901, opened 2020-07-11
Client: when computing disk usage, don't include VM shared directories

davidpanderson:
These directories are inside the VM, so we've already counted them.
Also, traversing them can be very slow if the dir is big (e.g. nanoHUB@home)

### DIFF
--- a/lib/filesys.cpp
+++ b/lib/filesys.cpp
@@ -449,11 +449,26 @@ int clean_out_dir(const char* dirpath) {
     return 0;
 }
 
+// check whether the dir is probably the shared dir for a VM job.
+// We don't want to count these in disk usage because the storage
+// is inside the VM, so it's already been counted
+//
+static bool is_vm_shared_dir(const char* dirpath) {
+    if (!strstr(dirpath, "slots/")) return false;
+    const char* p = strstr(dirpath, "shared");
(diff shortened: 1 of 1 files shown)


davidpanderson:
The Travis CI build is failing on fetching sqlite.
This can be removed; sqlite isn't used anymore.

AenBleidd:
@davidpanderson, I see sqlite3 usage still in browser.cpp
Should these functions be deleted from this class (or maybe even the whole class)?

AenBleidd:
I found this branch: https://github.com/BOINC/boinc/compare/dpa_remove_sqlite

davidpanderson:
I must have forgotten to make a PR for that.

davidpanderson:
That file (browser.cpp) should be removed.

AenBleidd:
@davidpanderson, it's already removed in a different PR

davidpanderson:
Actually, it turns out that shared/ resides in the host FS,
But apparently if the host (i.e. BOINC client) accesses it while the VM is running,
there's a locking mechanism or something that makes it very slow.
So let's not merge this for now.

AenBleidd:
I'm changing this PR to draft then. Feel free to change it back when you think it's ready

PR #3915, opened 2020-07-14
Manager: remove cookie-based autoattach schemes

davidpanderson:
A long time ago we tried auto-attach schemes (project and AM)
where web sites sent cookies with account info,
and the manager looked for these.
We dropped this idea because browsers kept changing how they stored cookies,
and we couldn't keep up.

Remove this code.  Also, we no longer need sqlite anywhere.


AenBleidd:
@CharlieFenton, could you please take a look at this?

CharlieFenton:
@AenBleidd Please let me fix the Mac build; it is not a good idea to edit the boinc.xcodeproj files with a text editor. And it is not a good idea to edit the associated build scripts if you are not completely familiar with them. I will check later tonight to see whether I have write access to the dpa_remove_sqlite branch.

AenBleidd:
@CharlieFenton, sure, feel free to fix it by yourself. But according to CI, my removal is quite clean and has no side-effects.

AenBleidd:
Thank you, @CharlieFenton!

CharlieFenton:
@AenBleidd It still linking to sqlite3, which worked because sqlite3 is still in Travis CI's dependency cache. When I temporarily removed sqlite3 on my Mac, the linker only gave a warning rather than a hard error, because there are no actual calls to that library. I will now reset the dependency cache (for this branch only) as a further test. This should also be done in the master branch when merging this PR.

CharlieFenton:
I found the following files still refer to sqlite (none apply to the Mac):
locale/update_templates.sh
xcompile/xcompile.MinGW32_on_cygwin.sh
win_build/load_dependencies.bat
clientgui/makefile.am
m4/sqlite3.m4

Also, I don't know whether the Travis CI cache needs to be rebuilt for other platforms to remove those sqlite libraries.

Finally, after rebuilding the cache, it is important to revert .travis.yml so it doesn't rebuild the cache every time Travis CI runs. I will do this after the current checks ar complete.

AenBleidd:
I'll check these files

RichardHaselgrove:
The current Windows installer (v7.16.7) deposits a file in the BOINC program directory:

sqlite3.dll, version3.10.2, datestamp 18 December 2016

Once all references have been deleted, and all checks are complete, this file should be removed from the installation list.

AenBleidd:
- locale/update_templates.sh : is needed to work with Transifex (not a part of build process of BOINC), so it should be still here
- xcompile/xcompile.MinGW32_on_cygwin.sh : it is not used anywhere but I removed sqlite from it
- win_build/load_dependencies.bat : I removed sqlite in this script
- clientgui/makefile.am : Cleaned-up
- m4/sqlite3.m4 : Removed

Once I test this on my branch, I'll make a commit to this PR

talregev:
Why you don't merge this PR?

AenBleidd:
> Why you don't merge this PR?

I'll answer you privately

TheAspens:
I'm still doing some testing on this however, I noted that project_init.h still has a field setup_cookie that I don't think is needed anymore.  The only other place it appears to be referenced in the code is int gui_prc_server_ops.cpp  handle_get_project_init_status().  I believe that it can be removed from these locations.

TheAspens:
Looks like I started posting right after you merged.  I'm going to go ahead and put my comments here once I finish testing and if needed another PR can be opened.

TheAspens:
As far as I can tell - the code for handling a project_init.xml file  in the startup directory is broken.

If I create a file like as defined here: https://boinc.berkeley.edu/wiki/Initialization_files_for_custom_installers#project_init.xml and set the URL for WCG and use my authenticator, then when I start up I get an ugly screen like this: 

image

My understanding of the previous use of the file is the following (memories fade, so this is what I think I remember):

**Scenario 1**:  If the project_init.xml was populated with just a URL, this would trigger the cookie search mechanism and if found, complete the attach process to the project.  If the cookie wasn't found, then I believe it would take you to the following page in the attach process: 
image


**Scenario 2**:  If the project_init.xml was popoulated with a URL and a valid authenticator (or weak authenticator) then it would automatically attach the user to the project using the given account.

**Scenario 3**:  An older version  [CACreateProjectInitFile.cpp](https://github.com/BOINC/boinc/blob/8408b989cd3c3797188879c9ff46de7b53bb1d55/clientsetup/win/CACreateProjectInitFile.cpp)  would read command line parameters passed into the windows installer as described here: https://boinc.berkeley.edu/wiki/Creating_custom_installers .   The command line arguments PROJINIT_URL, PROJINIT_AUTH and PROJINIT_TEAMNAME were copied by the custom action and written out as the project_init.xml file.  This allowed for automating the install through mass deployments techniques where the project and account could be passed in via the command line of the windows installer.

Unfortunately, none of these work anymore.  I don't know what part this code had in breaking them but if we are cleaning up the cookie mechanism this should be cleaned up as well.  Ideally scenario 3 should be fixed as that was a really useful mechanism for automated installation scenarios.

AenBleidd:
Looks like I see the reason of this, I'll fix it on weekend

davidpanderson:
Kevin: are the problems with project_init.xml a result of this PR?

PR #3920, opened 2020-07-16
GUI RPC: bind to INADDR_ANY if a remote host is actually configured

Flowdalic:
Instead of binding to INADDR_ANY as soon as remote_hosts.cfg exists,
only do so if the file actually contains at least one host. Prior to
this change, boinc client would bind to INADDR_ANY even if
--allow_remote_gui_rpc was *not* given and the remote_hosts.cfg
does *not* contain any hosts (but exists). This behavior is not
sensible, because in that case no client would be allowed to connect
eventually.

Furthermore, many distributions ship an example remote_hosts.cfg as
part of their boinc client package. And the previous behavior caused
boinc client to bind on INADDR_ANY, opening a remotely accessible port
per default.

To tighten security further, boinc client will now only
bind to localhost even if remote-hosts.cfg exists, but there are no
remote hosts configured witin that file.

**Release Notes**
Only listen on INADDR_ANY if either --allow_remote_gui_rpc is given, or if there is a least one host configured in remote_hosts.cfg. Otherwise, listen only on localhost. Previously, boinc would listen on INADDR_ANY as soon as remote_hosts.cfg exists, even if it does not contain any hosts.

### DIFF
--- a/client/gui_rpc_server.cpp
+++ b/client/gui_rpc_server.cpp
@@ -98,7 +98,7 @@ GUI_RPC_CONN::~GUI_RPC_CONN() {
 }
 
 GUI_RPC_CONN_SET::GUI_RPC_CONN_SET() {
-    remote_hosts_file_exists = false;
+    remote_hosts_configured = false;
     lsock = -1;
     time_of_last_rpc_needing_network = 0;
     safe_strcpy(password,"");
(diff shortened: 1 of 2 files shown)


davidpanderson:
You could remove the init to false, and then say
remote_hosts_configured = !allowed_remote_ip_addresses.empty();

Flowdalic:
> You could remove the init to false, and then say
> remote_hosts_configured = !allowed_remote_ip_addresses.empty();

Ha, believe it or not, but I pushed that change while you wrote that comment. :)

Flowdalic:
And another force-push to remote a stray empty line.

Flowdalic:
It appears [travis failing](https://travis-ci.org/github/BOINC/boinc/jobs/708647691#L115) is unrelated to this change.

PR #3924, opened 2020-07-16
[Android] Use Okio to handle RPC.

Isira-Seneviratne:
**Description of the Change**
Use Okio's `BufferedSource` and `BufferedSink` classes to handle RPC communication instead of using the raw `InputStream` and `OutputStreamWriter`, due to providing many advantages over the standard Java IO APIs: https://jakewharton.com/a-few-ok-libraries/.

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
@@ -31,8 +31,6 @@ import org.xml.sax.helpers.DefaultHandler;
 
 import java.io.BufferedReader;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStreamWriter;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
@@ -44,7 +42,10 @@ import edu.berkeley.boinc.utils.BOINCDefs;
 import edu.berkeley.boinc.utils.BOINCUtils;
 import edu.berkeley.boinc.utils.Logging;
 import kotlin.text.Charsets;
+import okio.BufferedSink;
+import okio.BufferedSource;
 import okio.ByteString;
+import okio.Okio;
 
 import static org.apache.commons.lang3.BooleanUtils.toInteger;
 
@@ -84,9 +85,9 @@ public class RpcClient {
     public static final int MGR_SYNC = 31;
 
     private LocalSocket mSocket;
-    private OutputStreamWriter mOutput;
-    private InputStream mInput;
-    private byte[] mReadBuffer = new byte[READ_BUF_SIZE];
+    private BufferedSource socketSource;
+    private BufferedSink socketSink;
+    private final byte[] mReadBuffer = new byte[READ_BUF_SIZE];
     protected StringBuilder mResult = new StringBuilder(RESULT_BUILDER_INIT_SIZE);
     protected StringBuilder mRequest = new StringBuilder(REQUEST_BUILDER_INIT_SIZE);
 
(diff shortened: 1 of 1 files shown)


android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
```diff
@@ -316,12 +317,12 @@ protected void sendRequest(String request) throws IOException {
...
         if (Logging.RPC_DATA && Logging.DEBUG)
             Log.d(Logging.TAG, "Sending request: \n" + request);
-        if (mOutput == null)
+        if (socketSink == null)
             return;
-        mOutput.write("<boinc_gui_rpc_request>\n");
-        mOutput.write(request);
-        mOutput.write("</boinc_gui_rpc_request>\n\003");
-        mOutput.flush();
+        socketSink.writeString("<boinc_gui_rpc_request>\n", Charsets.ISO_8859_1);
```
AenBleidd, line 322: Is there a way to set global or default charset in Okio and not set it on every write?

talregev: Why there was `'\003'` char? Is needed now?

AenBleidd: It's still here (two lines above)

talregev: Why we need this char?

davidpanderson:
The \003 is a delimiter for request messages.

talregev:
> The \003 is a delimiter for request messages.

Why delimiter is needed? One request is not one command?

RichardHaselgrove:
The XML request closure will be the logical end of (multiple line) RPC data.

But the \003 delimiter will be an 'end of transmission' signal to the communications layer to say that the message is complete and can be passed to the logic handler for processing.

talregev:
Thank you for your details answer.
For me, I don't like the `\003` is a delimiter.  I guess it matter of taste. 
So for my taste, you can choose delimiter other then spacial chars, and also implement it without end of transmission delimiter?
But it not so important. And I wanted to understand.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
Isira-Seneviratne: > 
> 
> Is there a way to set global or default charset in Okio and not set it on every write?

Okio uses UTF-8 by default.

talregev: You can do it like this:
https://javarevisited.blogspot.com/2012/01/get-set-default-character-encoding.html

Isira-Seneviratne: > 
> 
> You can do it like this:
> https://javarevisited.blogspot.com/2012/01/get-set-default-character-encoding.html

Thanks for the suggestion, but I don't think such a change is necessary. The charset only needs to be specified for 3 lines of code.

RichardHaselgrove:
Just had a PM from Fred / efmer of BoincStats and BAM! asking if this could be added - told him it looks like it's already done, and suggested @AenBleidd as a co-ordination point. I guess we ought to link you up and start testing.

RichardHaselgrove:
Oops, my mistake. Travelling, and been out all day in the sun. Yes - the report relates to BoincTasks, but all else remains.

PR #3931, opened 2020-07-19
[Android] Change default of ram max used idle frac to 0.9 like pc.

talregev:
Change default of ram max used idle frac to 0.9 like pc.

### DIFF
--- a/lib/prefs.cpp
+++ b/lib/prefs.cpp
@@ -233,9 +233,6 @@ void GLOBAL_PREFS::defaults() {
     network_wifi_only = false;
 #endif
     ram_max_used_busy_frac = 0.5;
-#ifdef ANDROID
-    ram_max_used_idle_frac = 0.9;
-#else
     ram_max_used_idle_frac = 0.9;
 #endif
     run_gpu_if_user_active = false;
--- a/lib/prefs.cpp
+++ b/lib/prefs.cpp
@@ -234,7 +234,7 @@ void GLOBAL_PREFS::defaults() {
 #endif
     ram_max_used_busy_frac = 0.5;
 #ifdef ANDROID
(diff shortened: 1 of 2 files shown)


lib/prefs.cpp
```diff
@@ -234,7 +234,7 @@ void GLOBAL_PREFS::defaults() {
 #endif
     ram_max_used_busy_frac = 0.5;
 #ifdef ANDROID
-    ram_max_used_idle_frac = 0.5;
+    ram_max_used_idle_frac = 0.9;
```
adamradocz, line 237: I think the code would be simpler if you just simply delete the android section.

talregev: @adamradocz 
Yes, I agree. But I was waiting to hear the opinion of @AenBleidd about changing the value itself.
If he agree, I will change the code to be more simpler. 

@adamradocz Thank you for your time to review my PRs.

AenBleidd:
Android devices (here I talk about smartphones only) usually have no so much RAM. If set RAM usage by BOINC to 90% will likely remove system UI processes from RAM to swap that will lead to lugging UI when turning screen on and unblocking the phone. If user is experienced and they know what they are doing - they could easily set more comfortable RAM usage for them.
I tend to left this settings as-is and not merge this PR.
@talregev, if you have strong arguments about this change - feel free to put them here.

talregev:
@AenBleidd I don't have strong argument about that. I will close this PR.

adamradocz:
This is idle setting, so I assume when the screen is turned on again, the tasks stop and the user won't notice anything about it, maybe a little lag. But the extra RAM could make a significant impact in terms of computing.
BTW, I don't remember this option is editable in the advanced menu?

talregev:
It editable and user can change this setting.

PR #3939, opened 2020-07-22
[WIP] improve accounting of non-BOINC CPU usage

davidpanderson:
The current method doesn't count short-lived processes.


AenBleidd:
This was already implemented in a more accurate way via #4859

PR #3946, opened 2020-07-23
[Android] Remove warnings in compilation time.

talregev:
Remove:
* `w: boinc\android\BOINC\app\src\main\java\edu\berkeley\boinc\attach\BatchProcessingActivity.kt: (137, 29): 'FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET: Int' is deprecated. Deprecated in Java`
Solution from:
https://stackoverflow.com/questions/33179042/i-tried-using-this-flag-in-setflags-from-tutorial-but-its-deprecated-what-do-i
* `w: boinc\android\BOINC\app\src\main\java\edu\berkeley\boinc\client\Monitor.kt: (340, 25): Condition 'state == null' is always 'false'` warning.
Add check to state if it null. if yes, write to log. and continue as before this change.

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.kt
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.kt
@@ -331,17 +331,17 @@ class Monitor : LifecycleService() {
                 val transfers = clientInterface.fileTransfers
                 val acctMgrInfo = clientInterface.acctMgrInfo
                 val newNotices = clientInterface.getNotices(clientStatus.mostRecentNoticeSeqNo)
+                var nullValues = ""
+                if (state == null) {
+                    nullValues += "state "
+                    Log.e(Logging.TAG, "readClientStatus(): connection problem, null: " + nullValues);
+                }
                 if (allNotNull(status, state, state!!.hostInfo, acctMgrInfo)) {
                     clientStatus.setClientStatus(status, state.results, state.projects,
                             transfers, state.hostInfo, acctMgrInfo,
                             newNotices)
                 } else {
-                    var nullValues = ""
-                    if (state == null) {
-                        nullValues += "state "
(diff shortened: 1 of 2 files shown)


adamradocz:
I don't think suppressing messages is a good idea.

talregev:
> I don't think suppressing messages is a good idea.

Suppressing msg is a bad idea.
If you will notice, you will see that is under if of version. old version use old constant, new version of code use new constant. 
They have the same value. in run time it not different. Also we know that this old deprecate code on old if section, so we don't need this warning, so we suppress it.

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.kt
```diff
@@ -331,17 +331,17 @@ class Monitor : LifecycleService() {
...
                             transfers, state.hostInfo, acctMgrInfo,
                             newNotices)
                 } else {
-                    var nullValues = ""
-                    if (state == null) {
-                        nullValues += "state "
-                    } else {
-                        if (state.hostInfo == null) nullValues += "state.host_info "
-                    }
+                    if (state.hostInfo == null) nullValues += "state.host_info "
```
AenBleidd, line 344: This will fail if `state == null`

talregev: No. It will fall before. (That why it have the warning). At this point of the code the state will never be null.
if state will be null. it will never enter inside the if. look here:
https://github.com/BOINC/boinc/pull/3946/files#diff-e5abe666500a14899c926f7fb6a9a9bbL334
`if (allNotNull(status, state, state!!.hostInfo, acctMgrInfo)) {`  we put inside `state!!.hostInfo`.
We can change the function that will not get the state.info, and take it from inside. (and then check if state is null inside).

PR #3948, opened 2020-07-24
Pass user priority config items to wrappers

davidpanderson:
The user can specify 3 priority-related prefs in cc_config.h:
bool no_priority_change
int process_priority
int process_priority_special

Wrappers should honor these prefs,
but currently they don't have access to them.
Pass them in the APP_INIT_DATA structure.

Modify wrapper.cpp to use these prefs if they exist;
this overrides the priority specified in the job file, if any.

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/lib/app_ipc.cpp
+++ b/lib/app_ipc.cpp
@@ -113,6 +113,9 @@ void APP_INIT_DATA::copy(const APP_INIT_DATA& a) {
         project_preferences = NULL;
     }
     vbox_window                 = a.vbox_window;
+    no_priority_change          = a.no_priority_change;
+    process_priority            = a.process_priority;
+    process_priority_special    = a.process_priority_special;
     app_files                   = a.app_files;
 }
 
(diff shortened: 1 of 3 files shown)


RichardHaselgrove:
Subject to verification, it seems to me that this PR suffers the same defects as its predecessor #3826. The 'wrapper' and 'normal' process priority values are drawn from two different, and incompatible, lists - base 1 and base 0 respectively. I repeat my comment from the previous PR:

@davidpanderson: Please cast your eye back over your two commits:

317dc91 (13 Oct 2014, for wrapper apps)
fcafba0 (29 Sep 2015, for 'native' apps)

I think these two, taken together, are where the confusion has arisen.

For the record, I still feel that the User Manual, relating to current clients, should reflect the 2015 numbering for native apps.

RichardHaselgrove:
To put it another way:

image

PR #3952, opened 2020-08-01
Added EVT_MENU_OPEN to the event table for Windows - Fixes #3934

### LINKED ISSUE #3934: [Manager] 'Event log options' menu is active when no client connected
**Describe the bug**
'Event log options' menu is active when no client connected and show dialog with default options enabled

**Steps To Reproduce**
1. Run BOINC Manager
2. Shutdown connected client
3. Do not connect again
4. Press 'Event log options' menu

**Expected behavior**
No dialog is shown

**System Information**
 - OS: Windows x64
 - BOINC Version: 7.16.7

delta1513:
Fixes #3934 

**Description of the Change**

Removed the pre-processor directives that stopped the loading of the EVT_MENU_OPEN event. This means that `CAdvancedFrame::OnMenuOpening()` will be executed as an event as expected on all platforms and hence prevent particular menu functionality from being accessed when the client is shut down.

Changes affect the BOINC manager.

**Alternate Designs**

Adding appropriate code to shutting down and then connecting to a client was one alternative.

**Release Notes**

Options are now greyed-out when the BOINC client is shutdown on Microsoft Windows.

### DIFF
--- a/clientgui/AdvancedFrame.cpp
+++ b/clientgui/AdvancedFrame.cpp
@@ -1111,6 +1111,17 @@ void CAdvancedFrame::OnMenuOpening( wxMenuEvent &event) {
     if (exitItem) {
         exitItem->Enable(true);
     }
+
+    // Specific menu items to keep enabled always (Switch to simple view and "other options")
+    wxMenuItem* simpleViewItem = menu->FindChildItem(ID_CHANGEGUI, NULL);
+    if (simpleViewItem) {
+        simpleViewItem->Enable(true);
+    }
+
+    wxMenuItem* otherOptionsItem = menu->FindChildItem(ID_OPTIONS, NULL);
+    if (otherOptionsItem) {
+        otherOptionsItem->Enable(true);
+    }
     
     wxLogTrace(wxT("Function Start/End"), wxT("CAdvancedFrame::OnMenuOpening - Function End"));
 }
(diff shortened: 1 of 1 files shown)


RichardHaselgrove:
You'll need to check the VS2019 build failure, but I've taken the VS2013 artifact for a brief test.

1) When no client is running, the "Simple View" commands for
- change skin
- switch to advanced view
- Options|Other options...

are available, as before.

2) In "Advanced View" with no client, even the 'switch to simple view' and 'other options' commands are greyed out: I think they should be available - the first as the reciprocal of the simple view command, and the second for the reason I suggested in #3187.

While testing, I had 'Run the client?' deselected in Other options: something kept over-riding that and starting the client anyway. I'll explore that another time and report back.

Apart from the two missing active menu items and the apparent client restart bug,  this does what it says on the tin.

AenBleidd:
@RichardHaselgrove, VS2019 build failure is not related to this PR. I'll fix it separately.
@delta1513, please take a look at comment above and fix Manager behavior accordingly.
Thanks you

RichardHaselgrove:
Tested under Windows 7/64 and Windows 10/64 - new code runs fine. Looked at the Linux Manager - behaves the same as @CharlieFenton describes for Mac in https://github.com/BOINC/boinc/issues/3934#issuecomment-660981056 ff. The extra two permanently-active submenu items should help those platforms too. I've also reviewed the new code - should be free of unwanted side-effects. Merging, with thanks to @delta1513.

PR #3954, opened 2020-08-02
Manager: clean up "Welcome page"

davidpanderson:
When a client and manager start up and there are no projects attached,
the client checks for a project_info.xml and passes its into to the manager.
If the file has both a project URL and an account key,
the manager tells the client to attach that project, and that's it.
If the file has URL but no account key, the manager puts up a "Welcome page".
The idea is to notify the user of what they're about to sign up for.
If the user clicks Next, they go to the name/password dialog.

Both of these cases work now.
The <name> element in project_info.xml is required.
I had to do a bit more work to make things work right when the user cancels
out of the welcome page, then adds a project the usual way.

This part of Manager is convoluted.
Big chunks of it should be deleted, the Welcome page among them.
Here I took the path of least resistance.

### DIFF
--- a/3rdParty/vcpkg_ports/ports/rappture/CMakeLists.txt
+++ b/3rdParty/vcpkg_ports/ports/rappture/CMakeLists.txt
@@ -91,8 +91,13 @@ add_library(rappture ${SRC_RAPPTURE_CORE})
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src/core/)
 
-find_package(EXPAT REQUIRED)
-target_link_libraries(rappture PRIVATE EXPAT::EXPAT)
+find_package(expat CONFIG REQUIRED)
+
+if(WIN32 AND NOT MINGW)
+    target_link_libraries(rappture PRIVATE expat::libexpat)
+else()
+    target_link_libraries(rappture PRIVATE expat::expat)
+endif()
 
 install(TARGETS rappture
             RUNTIME DESTINATION bin
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -767,6 +767,12 @@ int CLIENT_STATE::init() {
     process_autologin(true);
     acct_mgr_info.init();
     project_init.init();
+    // if project_init.xml specifies an account, attach
+    //
+    if (strlen(project_init.url) && strlen(project_init.account_key)) {
+        add_project(project_init.url, project_init.account_key, project_init.name, false);
+        project_init.remove();
+    }
 
     log_show_projects();    // this must follow acct_mgr_info.init()
 
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -762,15 +762,11 @@ static void handle_get_project_init_status(GUI_RPC_CONN& grc) {
         "<get_project_init_status>\n"
         "    <url>%s</url>\n"
         "    <name>%s</name>\n"
-        "    <team_name>%s</team_name>\n"
-        "    <setup_cookie>%s</setup_cookie>\n"
         "    %s\n"
         "    %s\n"
         "</get_project_init_status>\n",
         gstate.project_init.url,
         gstate.project_init.name,
-        gstate.project_init.team_name,
-        gstate.project_init.setup_cookie,
         strlen(gstate.project_init.account_key)?"<has_account_key/>":"",
         gstate.project_init.embedded?"<embedded/>":""
     );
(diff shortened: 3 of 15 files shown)


RichardHaselgrove:
Which 'project url' are we expecting to see passed on the command line? One supplied by the user, one supplied by an Account Manager, or a master url parsed from get_project_config?

All of these get out of step from time to time, depending on the speed of updating of all_projects_list and Account Managers. Any unverified url should be passed through get_project_config, so that each project has final control over their own master url at the instant of joining.

TheAspens:
@davidpanderson - I got two compile errors with this:

- 'team_name': is not a member of 'PROJECT_INIT'	boinc	C:\Users\KevinReed\boinc\client\gui_rpc_server_ops.cpp	772	
- 'setup_cookie': is not a member of 'PROJECT_INIT'	boinc	C:\Users\KevinReed\boinc\client\gui_rpc_server_ops.cpp	773

TheAspens:
Note that those compile errors are why the Travis CI job failed.

davidpanderson:
Oops!  Please update and try again.

AenBleidd:
Retriggering Ci...

TheAspens:
I was able to build the client and test the installer using Visual Studio 2013.  I found that passing a command line configuration of:
`/qn PROJINIT_URL=http://www.worldcommunitygrid.org PROJINIT_AUTH=712563_basdfase14a29easqqqsdfc3cb5ea0c` would correctly create the project_init.xml file in the BOINC data directory during installation.

The existence of project_init.xml file with  a URL would trigger the attach screen to pop up with this image:
image

which unfortunately is not correct for a couple of reasons:

1. The URL http://www.worldcommunitygrid.org is the URL in the all_projects_list.xml file so it should have been recognized as a project by BOINC

1. The project_init.xml also contained the account_key tag with a valid weak authenticator.  As a result, the dialogue should have automatically progressed to the final "you are attached to World Community Grid" screen without any user interaction.

A couple of thoughts.  Either 
A) this needs to restore the previous behavior as described above or 
B) Perhaps the logic to handle reading the project_init.xml file should be moved into the core client.  If did this, then on startup of the core client, it would check if the project_init.xml is found.  If it finds it, it will attempt an attach using the project URL and authenticator provided .

If we did B we would lose the case where a project can create a installer where the project_init.xml is created as part of the installer and so after installation it starts the autoattach to that project.  I think we are probably ok losing this use case because the new auto-attach mechanism should be used instead of this use case (in my opinion).

However, with B we would still have the ability to let automated installation tools easily perform a mass installation by passing the project URL and authenticator via command lines through the installer.   It would actually be nicer in my opinion, because it would work even in a manager free installation scenario (i.e. if they want to install it to run as a service without anyone ever seeing the manager).

davidpanderson:
You also need to specify the project name as well.

Are you using the right code?  I took out the "This project is not registered... stuff.

TheAspens:
Sorry about that - I did build from the wrong branch.  I had issues getting my environment setup on Windows 10 and that distracted me from that important bit.  However, I have been able to build using the 2013 community edition on Windows 10 from the dpa_proj_init branch now and got the following outcomes:

- Passing command line parameters into the installer correctly creates the project_init.xml file.  The file it created looks like this (although the actual file has a real and valid account key):
```
<project_init>
  <url>https://www.worldcommunitygrid.org</url>
  <name>https://www.worldcommunitygrid.org</name>
  <account_key>76863_asdfasdfasdfasdfasf</account_key>
  <embedded>0</embedded>
</project_init>
```
- I then launched the manager and got the following:
image
image

Since the account_key was provided in the project_init.xml file, it should have moved directly to completing the attach process without any user interaction required.  Unfortunately it appears that it is not using the account key at all.

TheAspens:
Ok - slightly more testing.  Note:  I deleted my data directory before this test.

The email address shown in the screen shot above is from a previous registration (i.e. it was pulled from the registry).  I went ahead and signed in using the prefilled email address "knreed@us.ibm.com" and its password.  However, instead of attaching to the account associated with the email and password I provided, it attached using the weak authenticator that was stored in the project_init.xml file which is associated with a different account.

davidpanderson:
I fixed the above problem.

TheAspens:
@davidpanderson - works correctly now.  Thanks for the fix.

I'm ready to merge this pull request however, the title still has WIP in it.  Is there other work in this pull request pending or can that be removed?

thanks!

TheAspens:
I spoke with @davidpanderson and he said that the WIP can be removed.  As a result I have done so and I'm now merging this PR.

PR #3968, opened 2020-08-10
[Manager] Stopped shutdown client dialog from showing when told not to

### LINKED ISSUE #3933: [Manager] Don't show dialog 'Shut down connected client' option not applied if pressed when no client connected
**Describe the bug**
Option changes its state when no client connected

**Steps To Reproduce**
1. Shutdown connected using corresponding menu option
2. Check there a checkbox 'Don't show this dialog again'
3. Press cancel when asked to connect to client
4. Try to 'Shutdown connected client' again
5. Repeat 3-4 couple of times

**Expected behavior**
Dialog is never shown again

**System Information**
 - OS: Windows x64
 - BOINC Version: 7.16.7

delta1513:
Fixes #3933

**Description of the Change**

The "shutdown connected client" modal will not show ever again if you check the "do not show this dialog again" box provided.

This fix only fixes the non-persistence issue and does not allow the user to make the modal appear once again and hence reverse their check of the box.

**Alternate Designs**

For future changes, people may want to be able to re-enable the modal and a setting being placed in the options was considered but not followed through.

**Release Notes**

The shutdown client dialog will no longer appear after checking the "don't show this dialog again" checkbox

### DIFF
--- a/clientgui/AdvancedFrame.cpp
+++ b/clientgui/AdvancedFrame.cpp
@@ -1312,6 +1312,7 @@ void CAdvancedFrame::OnClientShutdown(wxCommandEvent& WXUNUSED(event)) {
     CMainDocument*     pDoc = wxGetApp().GetDocument();
     CSkinAdvanced*     pSkinAdvanced = wxGetApp().GetSkinManager()->GetAdvanced();
     int                showDialog = wxGetApp().GetBOINCMGRDisplayShutdownConnectedClientMessage();
+    int                doShutdownClient = 0;
     CDlgGenericMessage dlg(this);
     wxString           strDialogTitle = wxEmptyString;
     wxString           strDialogMessage = wxEmptyString;
@@ -1347,13 +1348,17 @@ void CAdvancedFrame::OnClientShutdown(wxCommandEvent& WXUNUSED(event)) {
         dlg.m_DialogMessage->SetLabel(strDialogMessage);
         dlg.Fit();
         dlg.Centre();
+
+        if (wxID_OK == dlg.ShowModal()) {
+            wxGetApp().SetBOINCMGRDisplayShutdownConnectedClientMessage(!dlg.m_DialogDisableMessage->GetValue());
+            doShutdownClient = 1;
+        }
     }
 
-    if (!showDialog || wxID_OK == dlg.ShowModal()) {
-        wxGetApp().SetBOINCMGRDisplayShutdownConnectedClientMessage(!dlg.m_DialogDisableMessage->GetValue());
+    if (!showDialog || doShutdownClient) {
         pDoc->CoreClientQuit();
         pDoc->ForceDisconnect();
-        
+
         // Since the core cliet we were connected to just shutdown, prompt for a new one.
         ProcessEvent(evtSelectNewComputer);
     }
--- a/clientgui/AdvancedFrame.cpp
+++ b/clientgui/AdvancedFrame.cpp
@@ -1112,15 +1112,24 @@ void CAdvancedFrame::OnMenuOpening( wxMenuEvent &event) {
         exitItem->Enable(true);
     }
 
-    // Specific menu items to keep enabled always (Switch to simple view and "other options")
+    // Specific menu items to keep enabled always
+    // View->Simple view...
     wxMenuItem* simpleViewItem = menu->FindChildItem(ID_CHANGEGUI, NULL);
     if (simpleViewItem) {
         simpleViewItem->Enable(true);
     }
 
+    // Options->Other options...
     wxMenuItem* otherOptionsItem = menu->FindChildItem(ID_OPTIONS, NULL);
     if (otherOptionsItem) {
         otherOptionsItem->Enable(true);
+    }
+
+    // Specific menu items to enable based on connected client
+    // File->Shutdown connected client...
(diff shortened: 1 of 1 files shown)


RichardHaselgrove:
I've taken the VS2013 Windows build of this artifact for a test drive.

The behaviour is different from that requested in #3933: this behaviour is not conditional on the client being in a connected state. But it does tackle the secondary issue: the checkbox 'Don't show this dialog again' now works.

But I don't think it's a good idea. In principle, I don't like 'trapdoor' functions in user interfaces. Once the checkbox has been activated, there is no way to restore the default UI behaviour except by hacking the registry or hidden configuration file. Users (of all software, not just BOINC) have been known to explore unknown controls just to 'see what happens': they have also been known to hit the wrong option with a tired mouse or touchscreen. They should be able to recover from that.

Instead, the UI on selecting menu option 'Shut down connected client...' immediately displays the selection dialog for a new connection. If you realise this was a mistake and click 'cancel', the original connection has already been shut down without confirmation.

Given that we have representations from the Linux repo maintainers suggesting that users should **never** be able to shut down the client (they have removed this menu item from their distributions), we are pulling in two different directions at once: I think we should pause this until we've thought through all the implications of what is being suggested.

AenBleidd:
> But I don't think it's a good idea. In principle, I don't like 'trapdoor' functions in user interfaces. Once the checkbox has been activated, there is no way to restore the default UI behaviour except by hacking the registry or hidden configuration file. Users (of all software, not just BOINC) have been known to explore unknown controls just to 'see what happens': they have also been known to hit the wrong option with a tired mouse or touchscreen. They should be able to recover from that.

It's definitely not the topic of this PR because we have this behavior everywhere in BOINC Manager. So it should be the same here too.

> They should be able to recover from that.

I agree, but then it should be some option or action (restore default settings that will restore everything).

> Instead, the UI on selecting menu option 'Shut down connected client...' immediately displays the selection dialog for a new connection. If you realise this was a mistake and click 'cancel', the original connection has already been shut down without confirmation.

I agree, there should be a confirmation dialog first, but again, it's out-of-scope of this particular PR.
It's a nice feature request instead. @RichardHaselgrove, could you please create a separate ticket for this?

@delta1513, could you please fix this PR according to the first point Richard mentioned?

RichardHaselgrove:
I'd like to get some views from participating users before trying to formulate a different solution. What are we trying to achieve?

We have a confirmation dialog already: it had a dysfunctional "Don't show again" checkbox, but now it functions as intended - and we lose the confirmation dialog when it's checked. We can't have it both ways.

We _**could**_ simply remove the checkbox, so the dialog is shown unconditionally.

We _**could**_ hook the dialog up to the connection status, so it is shown (always) when there is an active connected client to be shut down, but not otherwise.

Or we _**could**_ invent a whole new process: **detach** from the current client (leaving it running), and select a different client to monitor/control. I do this now with the top file menu item 'New BOINC Manager window'.

Can anybody make a case for any of those (or any other variant) that catches all the perceived problems with this menu item?

AenBleidd:
I see it next:
1. no client connected: menu is disabled, no dialog shown
2. client is connected
- if checkbox 'Don't show this dialog again' is not checked: show dialog
- otherwise: silently shutdown connected client with no dialog shown

Here we could add confirmation dialog whether user really wants to shutdown client, but this should not be done in this PR

RichardHaselgrove:
> Here we could add confirmation dialog ...

But we can't show a confirmation dialog after the user has told us not to - they're the same dialog!

Ageless93:
Completely aside from a confirmation dialog or not, to reset the confirmation dialog you can expand the "Show exit dialog" option in Other options in BOINC Manager. Now it's one checkbox as it is only for the exit dialog on the Manager. We can make it a drop down option, where you choose which one to reset.

AenBleidd:
> > Here we could add confirmation dialog ...
> 
> But we can't show a confirmation dialog after the user has told us not to - they're the same dialog!

You're completely right, sorry

AenBleidd:
@RichardHaselgrove, could you please retest this? Thanks in advance

RichardHaselgrove:
Done. Tested with win-manager_PR3968_2020-08-16_040c37fe_x64, so with 040c37fe6532ead62f9f9a959736674de4c66d15

This does what was intended (disables menu item when client not connected), even when the client departure was a simulated crash not involving Manager interaction. `(boinccmd --quit)`

But we still have the trapdoor function on "Don't show this dialog again". If checkbox checked, Client running, menu active - client shut down even on cancel of only dialog shown - the one to start a different client. No confirmation, and no way to re-activate confirmation without hacking the registry.

Could we add that other item to the 'Other Options' dialog?

Enable Manager exit dialog? (existing)
Enable Client shutdown dialog? (new)

hooking them to `DisplayShutdownClientDialog` and `DisplayShutdownConnectedClientDialog` respectively.

[I'm scratching my head over `DisplayExitWarning` - it's in my Registry, but can't be found via a Github code search. From an old version, perhaps - unfortunately the Registry doesn't keep datestamps)

AenBleidd:
@RichardHaselgrove, thank you for testing this.

> Could we add that other item to the 'Other Options' dialog?

Yes! Could you please create a ticket for this?

RichardHaselgrove:
I think that sort of 'duplicate and adapt' should be within my skill set, so I'll try to go direct for a PR. Got an appointment this morning, but I'd hope to do it by tonight/tomorrow

PR #3970, opened 2020-08-13
Manager: change  "Show graphics" button to "Stop graphics" when gfx running

CharlieFenton:
This is an important feature in preparation for the imminent release of MacOS 11 Big Sur, and complements PR #3964 by providing another workaround. As explained in that PR:
> When a user presses the Show graphics button in the _BOINC Manager_, the project graphics application is run for the selected task. On the Macintosh, the graphics app runs as a free standing process, without a functioning menu bar, so the usual way to shut it is to use the red close button at the top left corner of the application's window. But under MacOS 11 Big Sur, that button is disabled.

Until projects relink their graphics apps with the new code from #3964, when a user clicks the _Show graphics_ button, the only way to stop the graphics app will be by a control-click (or right-click) on the graphics app's icon in the Dock and selecting `Quit`, which is not obvious,

This PR adds a more convenient and clearer alternative. The _Show graphics_ button label changes to _Stop graphics_ while the graphics app is running. Clicking on it stops the graphics app. This is a useful addition to the GUI even when the close button / close box is available.

This PR also fixes a related issue. Some projects have not provided a 64 bit (_x86_64_) Macintosh graphics application. They only have a 32 bit (_i686_) version, which cannot run on Macs with MacOS 10.15 or later. So nothing happens when the user clicks on the _Show graphics_ button. This PR adds code which checks each graphics app to determine whether it's architecture is compatible with the CPU hardware and OS version, and disables the _Show graphics_ button for incompatible graphics apps.

Because Apple has announced it will release _Apple Silicon_ (_arm64_) Macs very soon, this code also adds compatibility checks for _x86_64_ and _arm64_ graphics apps on _x86_64_ and _arm64_ CPUs.

### DIFF
--- a/client/app.h
+++ b/client/app.h
@@ -182,11 +182,6 @@ struct ACTIVE_TASK {
         return _task_state;
     }
 
-#if (defined (__APPLE__) && (defined(__i386__) || defined(__x86_64__)))
-    // PowerPC apps emulated on i386 Macs crash if running graphics
-    int powerpc_emulated_on_i386;
-    int is_native_i386_app(char*);
-#endif
     int request_reread_prefs();
     int request_reread_app_info();
     int link_user_files();
--- a/client/app_start.cpp
+++ b/client/app_start.cpp
@@ -55,13 +55,6 @@
 #include <process.h>
 #endif
 
-#if (defined (__APPLE__) && (defined(__i386__) || defined(__x86_64__)))
-#include <mach-o/loader.h>
-#include <mach-o/fat.h>
-#include <mach/machine.h>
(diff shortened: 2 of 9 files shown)


CharlieFenton:
Closing and reopening to trigger CI Travis build

CharlieFenton:
Please merge as soon as possible. BOINC needs significant fixes to run on MacOS 11 Big Sur, which Apple will be releasing to the public very soon (currently at Beta 4). So we need to release a new version of BOINC for the Mac, including this PR,  preferably before Apple's public release of MacOS 11 Big Sur.

Issue #3669, opened 2020-05-01
Compile android ndk on windows.


talregev:
This template is for miscellaneous issues not covered by the other issue categories.
I am using windows 10. Download msys64 for support sh and more linux command.
I modified the script and compile, but I have some issue, and I have questions:

My current issue, when it try to compile boinc for arm, it didn't recognize 'pkg-config'. but in the visual studio is download it and compile it. in  $ boinc\3rdParty\Windows\vcpkg. 

1. Why in the script it say no for compile openssl and curl? 
2. How you compile android ndk on windows?
3. Can we transform these scripts to cmake scripts? 
Android studio will compile and download all relevant 3rdparty, like openssl and curl and dpkg. All compilation will be inside android studio no matter your os.
4. Can you publish compiled ndk binaries for android studio? (arm, arm64, x86, x64 platform) for 7.6.15 or 7.6.16?

https://developer.android.com/ndk/guides
https://developer.android.com/studio/projects/add-native-code

talregev:
We can keep boinc with current build system. 
We can convert the openssl and curl and pkg-config scripts that will download and recognized by boinc build system using cmake.
If I will give it the correct pkg-config folder, it will work?

AenBleidd:
@talregev, I suggest you to instal WSL and compile BOINC for Android there

talregev:
I will try. any pre install need inside wsl?

AenBleidd:
Yes. Mainly all necessary packages that is required to build on linux

AenBleidd:
> 1. Why in the script it say no for compile openssl and curl? 

It is done to optimize CI build that uses cache. By default, openssl and curl are building.

> 2. How you compile android ndk on windows?

Personally I use preconfigured linux VM

> 3. Can we transform these scripts to cmake scripts? 
> Android studio will compile and download all relevant 3rdparty, like openssl and curl and dpkg. All compilation will be inside android studio no matter your os.

I see no reason for this. You have to transform linux version then too

> 4. Can you publish compiled ndk binaries for android studio? (arm, arm64, x86, x64 platform) for 7.6.15 or 7.6.16?

By default they are published here: https://boinc.berkeley.edu/dl/

talregev:
It not working on wsl.
![wsl](https://user-images.githubusercontent.com/7043539/80833326-fe20ce80-8bf6-11ea-983a-565889782121.PNG)

talregev:
[...]

PR #3980, opened 2020-08-18
Mac: restore updated error message lost by my commit fcbdcc75b7

CharlieFenton:
Mac: restore updated error message lost by my commit fcbdcc75b7 (which was from PR# 3631)

In my PR #3369, I had changed this error message from  `Screenaver of application <name>  is not compatible with this version of OS X. `
to `Could not connect to screensaver of application  <name>.`

Somehow that part of PR #3669 was lost / reversed in my more recent PR #3631. This corrects that.


CharlieFenton:
I want to include this in my upcoming Mac release because the _not compatible_ message is misleading. Failure to get the screensaver working can be, and often is, due to any of a number of transient factors.

RichardHaselgrove:
I see the changes and the previous reversion - good enough for me, although I haven't explored the wider context. Merging.

PR #3986, opened 2020-08-27
Update switcher.cpp to include mac_spawn only if the system is APPLE

LocutusOfBorg:
this fixes a build failure on Linux/Debian systems where mac/* directory is stripped from the source tarball.

### DIFF
--- a/client/switcher.cpp
+++ b/client/switcher.cpp
@@ -34,7 +34,9 @@
 #include "app_ipc.h"
 #include "filesys.h"
 #include "str_replace.h"
+#ifdef __APPLE__
 #include "mac_spawn.h"
+#endif
 
 using std::strcpy;
 


CharlieFenton:
This should not be needed. Switcher is not used by any platform other than Apple, so it should not be built for any other platform.

CharlieFenton:
The correct fix for this is PR #3804. Please revert this change (PR #3986) and merge PR #3804, with has been held up by @AenBleidd since June 5, even though the author fixed the problem. As far as I know, make files are used on Macs only to build BOINC libraries, never the client (_boinc_), manager (_BOINCManager_), screensaver controller (_BOINCSaver.saver_) or default screensaver graphics (_boincscr_).

CharlieFenton:
An alternative fix might be to change line 28 of _client/makefile.am_ from this:
`if !OS_WIN32`
to this:
`if OS_DARWIN`

LocutusOfBorg:
@CharlieFenton thanks for the new pull request!

I'm even happier to not build that program at all under linux if not needed

CharlieFenton:
> Please revert this change

I've decided that we might as well leave this change alone, since there are other instances of #ifdef __APPLE__ in this source file. But I still recommend adoption of PR #3804.

PR #3988, opened 2020-08-28
client and wrapper: fix process priority mismatch

davidpanderson:
It turns out we have two different encodings of process priority:
1) specified in cc_config.xml and used by the client: 0 (low) to 4 (high)
2) specified in job.xml and used by the wrapper: 1 (low) to 5 (high).

This didn't cause any problems until recently when I added code
to pass the cc_config.xml info to the wrapper;
it was interpreting it on the 1-5 scale.

Fix: have the wrapper convert it (add one).

Also: I forgot to have the client actually put the priority into
in the app_init_data.xml file.

### DIFF
--- a/samples/wrapper/wrapper.cpp
+++ b/samples/wrapper/wrapper.cpp
@@ -747,7 +747,10 @@ int TASK::run(int argct, char** argvt) {
         priority_val = 0;
     } else {
         if (aid.process_priority > 0) {
-            priority_val = process_priority_value(aid.process_priority);
+            // priority coming from the client is on scale where 0 is idle.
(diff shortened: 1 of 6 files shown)


samples/wrapper/wrapper.cpp
```diff
@@ -747,7 +747,10 @@ int TASK::run(int argct, char** argvt) {
         priority_val = 0;
     } else {
         if (aid.process_priority > 0) {
```
RichardHaselgrove, line 749: Please also change this test to 

`if (aid.process_priority > CONFIG_PRIORITY_UNSPECIFIED) {`

PR #3989, opened 2020-08-28
Revert "Update switcher.cpp to include mac_spawn only if the system is APPLE"

davidpanderson:
Reverts BOINC/boinc#3986


CharlieFenton:
I've decided that we might as well leave this alone, since there are other instances of `#ifdef __APPLE__` in this source file.

PR #3999, opened 2020-08-31
wrapper: interpret priority value in cc_config.xml correctly

davidpanderson:


RichardHaselgrove:
@davidpanderson - please also consider my note at https://github.com/BOINC/boinc/pull/3988#discussion_r478877592

PR #4003, opened 2020-09-01
[Android] Use more extensions.

Isira-Seneviratne:
**Description of the Change**
* Make use of more extension functions provided in Jetpack, such as `String.toUri()`.
* Add an extension function to convert a `long` to a `LocalDateTime`.

**Alternate Designs**
N/A

**Release Notes**
N/A

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectDetailsFragment.kt
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectDetailsFragment.kt
@@ -24,7 +24,6 @@ import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
 import android.content.IntentFilter
-import android.graphics.Bitmap
 import android.graphics.Point
 import android.os.Bundle
 import android.os.RemoteException
@@ -34,6 +33,7 @@ import android.util.Log
 import android.view.*
 import android.widget.Button
 import android.widget.TextView
+import androidx.core.graphics.scale
 import androidx.core.net.toUri
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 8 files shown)


android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectDetailsFragment.kt
```diff
@@ -364,9 +364,7 @@ class ProjectDetailsFragment : Fragment() {
                 val slideshowBinding = ProjectDetailsSlideshowImageLayoutBinding.inflate(layoutInflater)
                 var bitmap = image.image!!
                 if (scaleImages(bitmap.height, bitmap.width)) {
-                    bitmap = Bitmap.createScaledBitmap(image.image!!,
-                            image.image!!.width * 2,
-                            image.image!!.height * 2, false)
+                    bitmap = bitmap.scale(bitmap.width * 2, bitmap.height * 2, filter = false)
```
AenBleidd, line 367: Why do we do upscale here?

android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/glide/ScaleBitmapBy2.kt
```diff
@@ -19,13 +19,14 @@
 package edu.berkeley.boinc.attach.glide
 
 import android.graphics.Bitmap
+import androidx.core.graphics.scale
 import com.bumptech.glide.load.engine.bitmap_recycle.BitmapPool
 import com.bumptech.glide.load.resource.bitmap.BitmapTransformation
 import java.security.MessageDigest
 
 class ScaleBitmapBy2 : BitmapTransformation() {
```
AenBleidd, line 27: Why do we need this class at all? Is it possible to remove it and update package with better graphics?

AenBleidd (changes requested):

android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/glide/ScaleBitmapBy2.kt
Isira-Seneviratne: It's used in Glide to transform the downloaded image before displaying it.

android/BOINC/app/src/main/java/edu/berkeley/boinc/ProjectDetailsFragment.kt
Isira-Seneviratne: I believe the original developer did it to make the slideshow images more visible.

PR #4008, opened 2020-09-06
[VS2019] Remove custom vcpkg rappture port

AenBleidd:


AenBleidd:
@RichardHaselgrove, while you're here, could you please review and merge?
Changes are minimal: we remove our own vcpkg rappture port in favor of official one from Microsoft vcpkg repo (yes, it was also implemented by me, but anyway).
Build is green so this means that change works as designed.
Thanks in advance.

RichardHaselgrove:
I'm not an expert in this area, and I don't have a VS2019 compiler on this laptop... But as you say, it's a clean remove of the vcpkg port overlay: the other files are already deleted. Merging - we should have plenty of time to revert before the next release branch is needed, so we can revert if needed.

Vulpine05:
I'm by no means an expert, but I did install VS2019, so if you need someone to test compiling, let me know.  I haven't gotten it to work yet, but that could be due to my lack of experience/knowledge.

PR #4026, opened 2020-09-21
Add a project-customizable hook to result pages

Rytiss:
**Description of the Change**
Workunit pages already have a hook to display project-specific data for a workunit, but result pages lack such functionality. This change adds a similar hook to the result pages.

**Release Notes**
Add a hook to easily specify project-specific information in the result pages.


davidpanderson:
Thanks.  I updated the doc.

PR #4031, opened 2020-09-25
client/scheduler: update NVidia detection again

RichardHaselgrove:
for new RTX 30x0 range released September 2020

**Description of the Change**

GFLOPS peak is under-reported by a factor of 2 on the new cards, compared with other reporting tools

**Alternate Designs**

I've been asked if we could use NvAPI_GPU_GetGpuCoreCount() instead  (available since 2014), but rejected as a more complex modification which might break backwards compatibility.

**Release Notes**
Correct detection of NVidia RTX30x0 range GPUs


ChristianBeer:
Close and reopen to trigger CI and provide current build artifacts

AenBleidd (approved):

PR #4037, opened 2020-09-27
Avoid using boinc_db global.

### LINKED ISSUE #4032: Can't use new database connection in create_work function
**Describe the bug**
The following line of code in tools/backend_lib.cpp hard codes the use of boinc_db database connection defined in boinc_db.cpp. This prevents an application from using its own database connection for creating workunits via create_work function. Furthermore, the application may crash when boinc_db.mysql is not initialized.

backend_lib.cpp:388             
`wu.id = boinc_db.insert_id()`

The line should be replaced with the following:

backend_lib.cpp:388             
`wu.id = wu.db->insert_id();`

**Steps To Reproduce**
1. Define a database connection for inserting information about workunits into the mysql database. Database connection must be initialized

```
DB_CONN dbconn;
dbconn.open(db_name, db_host, db_user, db_passwd);
```

2. Pass the database connection to a workunit (wu)

`DB_WORKUNIT wu = DB_WORKUNIT(&dbconn);`

3. Pass the wu to create_work function 

```
create_work(
        wu,
        in_template,
        path,
        config.project_path(path),
        infiles,
        2,
        config
    ); 
```

**Expected behavior**
create_work function should update the mysql database using dbconn and not boinc_db database connection. 

**Screenshots**
NA

**System Information**
 - OS: ubuntu 18.04
 - BOINC Version: 7.17.0

**Additional context**
NA

tomasbrod:
Fixes #4032 (maybe)

**Description of the Change**
Use db pointer from DAL to invoke db operation instead of accessing global var.

### DIFF
--- a/sched/credit.cpp
+++ b/sched/credit.cpp
@@ -1106,7 +1106,7 @@ int write_modified_app_versions(vector<DB_APP_VERSION_VAL>& app_versions) {
             if (retval) {
                 break;
             }
-            if (boinc_db.affected_rows() == 1) {
+            if (av.db->affected_rows() == 1) {
                 break;
             }
             retval = av.lookup_id(av.id);
--- a/tools/backend_lib.cpp
+++ b/tools/backend_lib.cpp
@@ -385,7 +385,7 @@ int create_work2(
             );
             return retval;
         }
-        wu.id = boinc_db.insert_id();
+        wu.id = wu.db->insert_id();
     }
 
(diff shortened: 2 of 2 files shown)


davidpanderson:
I don't understand the use case.
All the BOINC programs use boinc_db, i.e. they start with boinc_db.open(....)
Is there a reason why you can't or don't want to do this?

Issue #3778, opened 2020-05-27
Client: project-wide file upload backoff is too aggressive ...


RichardHaselgrove:
**Describe the bug**
Project-wide file upload backoff is too aggressive ...
... when multiple files need to be uploaded for a single task.

GPUGrid (as an example) uploads six files per task, and occasionally suffers network congestion/connection problems.

The client tries to upload each new file at least once, and applies a separate backoff to each file if it fails - starting small for each file, but increasing exponentially (within limits) if successive retries for each file fail.

The project-wide backoff is cumulative - by the time the sixth file is reached, the backoff can already be well over an hour, as it was in this case (I retried the uploads manually):

27/05/2020 09:12:02 | GPUGRID | Sending scheduler request: To fetch work.
27/05/2020 09:12:02 | GPUGRID | Requesting new tasks for NVIDIA GPU
27/05/2020 09:12:02 | GPUGRID | [sched_op] CPU work request: 0.00 seconds; 0.00 devices
27/05/2020 09:12:02 | GPUGRID | [sched_op] NVIDIA GPU work request: 14638.31 seconds; 0.00 devices
27/05/2020 09:12:02 | GPUGRID | [sched_op] Intel GPU work request: 0.00 seconds; 0.00 devices
27/05/2020 09:12:04 | GPUGRID | Scheduler request completed: got 0 new tasks
27/05/2020 09:12:04 | GPUGRID | [sched_op] Server version 613
27/05/2020 09:12:04 | GPUGRID | No tasks sent
27/05/2020 09:12:04 | GPUGRID | This computer has reached a limit on tasks in progress
27/05/2020 09:12:04 | GPUGRID | Project requested delay of 31 seconds
27/05/2020 09:12:04 | GPUGRID | [sched_op] Deferring communication for 00:00:31
27/05/2020 09:12:04 | GPUGRID | [sched_op] Reason: requested by project
27/05/2020 09:54:19 | GPUGRID | Computation for task 1a5cA00_413_3-TONI_MDADex2sa-3-50-RND5174_0 finished
27/05/2020 09:54:19 | GPUGRID | Starting task 1c7sA02_348_1-TONI_MDADex2sc-3-50-RND8505_3
27/05/2020 09:54:19 | GPUGRID | [cpu_sched] Starting task 1c7sA02_348_1-TONI_MDADex2sc-3-50-RND8505_3 using acemd3 version 210 (cuda101) in slot 3
[...]

Issue #3359, opened 2019-10-30
Manager: missing/inconsistent keyboard handlers


RichardHaselgrove:
**Describe the bug**
Keyboard shortcuts for moving between controls on the Manager's 'Advanced View' display are broken. User report from user with broken pointing device, in https://boinc.berkeley.edu/forum_thread.php?id=13193

**Steps To Reproduce**
Under Windows -
The TAB key moves focus forward through all active controls on a page (command buttons, list controls). It does **not** move focus to the tab label.
Shift-TAB moves focus back through all controls, **including** the tab label - but doesn't continue to re-start the loop.
Ctrl-TAB and Ctrl-Shift-TAB cycle though the notebook pages. If the movement was started with the focus on a tab label, it moves through all pages. If the movement was started with the focus on a different control, it reaches the Notices page and stops there. There's no way off the Notices page (unless it has focus) from the keyboard, except by going via the View menu.

Under Linux -
TAB and Shift-TAB cycle through all controls, including the tab label, and continue on to the next control in the loop.
Ctrl-TAB does nothing.

**System Information**
 - OS: Windows 7 and 10: Linux Mint, Cinnamon desktop: I don't have a Mac
 - BOINC Version: up to and including v7.16.3

**Additional context**
This is poor design for users without the manual dexterity to use a precision pointing device. It also hinders users with a broken or missing pointing device.

winkies:
@RichardHaselgrove Based on same logic as CTRL+A shortcut, I added the CTRL+TAB shortcut on advance view of boinc manager.
ATM, it's only available on my fork : see the [Draft PR](https://github.com/winkies/boinc/pull/1).

CharlieFenton:
Please see my objection to adding the CTRL+TAB shortcut in [this comment](https://github.com/BOINC/boinc/pull/4146#issuecomment-759150837) to #4146

CharlieFenton:
This PR is not necessary or appropriate.

The ability to move among controls within the Advanced View are already present, at least for the Projects and Tasks tabs on MS Windows and Macs. The existing built-in and, I believe standard, controls for moving among controls within BOINC Manager are Tab and Shift-Tab. I believe these are implemented by wxWidgets itself, or perhaps by the Operating System, not by BOINC code. 

As far as I can tell, there is only one problem associated with Tab and Shift-Tab: They don't work until an item in the list of tasks / projects has been selected. This is what needs to be fixed. We do not need to add even more keyboard shortcuts.

[...]

PR #4041, opened 2020-09-29
Add CTRL+A shortcut

### LINKED ISSUE #2155: CTRL+A shortcut doesn't work
The CTRL+A shortcut doesn't work on the Projects and Tasks tab. Assuming also this is the case on the Transfers tab.
The CTRL+Left Click and the SHIFT+UP/DOWN/HOME/END shortcuts are working.

winkies:
Fixes #2155

**Description of the Change**
Define shortcut on accelerator table and bind it with event table. 

**Alternate Designs**
Tested only on Linux environment. It should work as well on Windows but **it should be tested on MacOS X**.

**Release Notes**
Add CTRL+A shortcut for advance view of boinc manager.

### DIFF
--- a/clientgui/AdvancedFrame.cpp
+++ b/clientgui/AdvancedFrame.cpp
@@ -204,6 +204,7 @@ BEGIN_EVENT_TABLE (CAdvancedFrame, CBOINCBaseFrame)
     EVT_TIMER(ID_REFRESHSTATETIMER, CAdvancedFrame::OnRefreshState)
     EVT_TIMER(ID_FRAMERENDERTIMER, CAdvancedFrame::OnFrameRender)
     EVT_NOTEBOOK_PAGE_CHANGED(ID_FRAMENOTEBOOK, CAdvancedFrame::OnNotebookSelectionChanged)
+    EVT_MENU(ID_SELECTALL, CAdvancedFrame::OnSelectAll)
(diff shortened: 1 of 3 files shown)


RichardHaselgrove:
Tested on Windows: it does work on the Tasks tab, which is where it will be most useful for me (see #3778).

I rarely need to operate on all tasks across all projects, but I might need to operate on all tasks from a single project. It would be helpful to have a command button "Show [tasks from] only this project", analogous to the button at the foot of the Event Log window - the two controls would work well together.

AenBleidd:
@RichardHaselgrove, sound like a new feature request ;)

winkies:
@RichardHaselgrove Glad it's useful :)

I see what you need : it doesn't seem to be too much difficult. 
If maintainer (@AenBleidd ) approve your feature request, I could implement it in another one pull request ;)

RichardHaselgrove:
The test for me of how useful a feature is: how many machines do I install in on, on the first day of testing.

Three hosts so far, monitoring eight clients between them!

So, yes, I propose "Show [tasks from] only this project" as a feature request, and I look forward to testing the resulting PR!

RichardHaselgrove:
@winkies: not wanting to overwhelm you, but if you're comfortable working with the keyboard handler in wxWidgets, may I draw your attention to my #3359 - that one might be up your street too.

AenBleidd (approved):

PR #4043, opened 2020-09-30
Change python scripts to use python3.

davidpanderson:
You have to install mysqlclient for python3;
that has the same interface as mysqldb, which we used with python2.

The scripts don't work with python2.

### DIFF
--- a/tools/check_project
+++ b/tools/check_project
@@ -1,6 +1,5 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 
-# $Id$
 '''
 Checks project permissions and php modules.
 
--- a/tools/check_project
+++ b/tools/check_project
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # $Id$
 '''
@@ -26,15 +26,15 @@ def check_user(user):
    try:
       mypwdb = pwd.getpwnam(user)
    except KeyError as e:
-      print "ERROR: ", e
-      print "ERROR: Specified apache user (%s) not found, Aborting"%(user)
+      print("ERROR: ", e)
+      print("ERROR: Specified apache user (%s) not found, Aborting"%(user))
       sys.exit(1)
    #try
 # check_user
 
 def check_dir(dir):
    if not os.path.isdir(dir):
-      print "ERROR: project directory %s does not exist, Aborting"%(dir)
+      print("ERROR: project directory %s does not exist, Aborting"%(dir))
       sys.exit(1)
    #try
    return dir
@@ -64,10 +64,10 @@ def determineapacheuser(user):
       if len(out1)==1:
          return out1.pop()
 
-      print "WARNING: Script could not determine which user runs apache Web server. Aborting"
-      print "         You should specify the apache user with the -a/--apache_user flag"
-      print "         Try running the following command with apache Web server running:"
-      print "           ps -ef | egrep '(httpd|apache2|apache)' | grep -v `whoami` | grep -v root"
+      print("WARNING: Script could not determine which user runs apache Web server. Aborting")
+      print("         You should specify the apache user with the -a/--apache_user flag")
+      print("         Try running the following command with apache Web server running:")
+      print("           ps -ef | egrep '(httpd|apache2|apache)' | grep -v `whoami` | grep -v root")
       sys.exit(1)
    else:
       check_user(user)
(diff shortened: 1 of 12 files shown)


davidpanderson:
BTW, this is essentially the same as #3259, with a couple of extra things.

davidpanderson:
I tested this (make_project and start) with python 2.7.17 and 3.6.9

AenBleidd:
These scripts looks suspicious for me (and contain a lot of broken and outdated stuff). I ran pylint on it (to check both python2 and python3 correctness) and it showed me a lot of errors. I'll create a PR with CI to check this and probably help fix these scripts.

AenBleidd:
@davidpanderson, here's an example of pylint (for python2, for python3 it looks similar): https://travis-ci.org/github/BOINC/boinc/jobs/732352359

CI job is not finished yet but you can see errors already (those one that starts with `E:`).
I check one of them and it looks valid. Not sure about all of them.
I could take a look later but if you briefly look at errors (don't looks at warning and notices, they are not important right now but maybe you will find some of them valuable and useful) they looks pretty much interesting.

Please let me know what do you think about it and whether they are valid and should be fixed.

Later I plan to finish these CI jobs and fill them with all valuable Python scripts.

davidpanderson:
pylint's complaints are not related to this PR; please move to new issue if you want
(although cleaning up the python code is a very low priority as far as I'm concerned).
The goal of this PR is just to make the scripts work with both python 2 and 3.

tools/check_project
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
AenBleidd, line 1: I'm not sure this will run with python2

tools/upgrade
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
AenBleidd, line 1: The same is here

AenBleidd (commented):
Two of four files (that have no *.py extension) have changed `python` to `python3`.
I'm not sure these scripts will run with python2. In any case, all four should be unified.
Pylint found a lot of errors but they are pretty much similar for both python2 and python3 and not related to this particular PR.
They should be definitely fixed but in a separate PRs.

tools/upgrade
davidpanderson: Oops!  I forgot to change that back after testing.  The scripts work with either Python now.

TheAspens:
Greetings - I started working on testing this out first with a python3 on Ubuntu.  I got the following error message while trying to run the make_project script:

```
knreed@DESKTOP-THCN98H:~/boinc/tools$ ./make_project --delete_prev_inst --project_root /boinc --url_base http://localhost --db_name BOINC --db_user boinc  --db_passwd abc123 WSLBOINC 'BOINC Server running on WSL'
Creating project 'BOINC Server running on WSL' (short name 'WSLBOINC'):
   PROJECT_ROOT = /boinc/
   PROJECT_HOST = DESKTOP-THCN98H
       URL_BASE = http://localhost/
  HTML_USER_URL = http://localhost/WSLBOINC/
   HTML_OPS_URL = http://localhost/WSLBOINC_ops/
        KEY_DIR = /boinc/keys/
        DB_NAME = BOINC
        DB_HOST =

Delete /boinc/? [y/N] Traceback (most recent call last):
  File "./make_project", line 207, in <module>
    if not query_noyes('Delete %s?'%options.project_root):
  File "/home/knreed/boinc/py/Boinc/setup_project.py", line 713, in query_noyes
    return tools.query_noyes(str)
  File "/home/knreed/boinc/py/Boinc/tools.py", line 79, in query_noyes
    return input().strip().lower().startswith('y')
UnboundLocalError: local variable 'input' referenced before assignment
```

which I am guessing is because of this:
```
def query_noyes(str):
    '''Query user; default No'''
    print (str, "[y/N] ", end="")
    try: input = raw_input
    except NameError: pass
    return input().strip().lower().startswith('y')
```
Which appears to occur due to the reason cited in this stackoverflow: https://stackoverflow.com/questions/20873285/unboundlocalerror-local-variable-input-referenced-before-assignment

davidpanderson:
Fixed; please try again.

TheAspens:
I tested creating a new project on Red Hat Enterprise Linux 7 (which has python 2 at /usr/bin/python) and Ubuntu 20.04 (via WSL) which has python 3 at /usr/bin/python.  Both were successful and I was able to get the sample app installed and get the daemons running.

I think this looks good.  Thanks for making python 3 a viable option!

PR #4051, opened 2020-10-08
[Android] Remove old sched method in client

### LINKED ISSUE #2549: [Android] Boinc doesn't utilize all the cores
I have a LeEco Le Max 2 phone with LineageOS 15.1 (Android 8.1). It has Snapdragon 820 SOC with 4 cores. The power management is set to performance.

I set the boinc to use all the 4 cores, and if I check the CPU utilization with the Kernel Auditor, it shows that only 1 CPU is used. I've tried multiple scenarios with different options like use 1, 2 or 3 cores, but I got the same results.

screenshot_kernel_adiutor_20180605-155029

truboxl:
Fix #2549 

**Description of the Change**
Remove old Android scheduling method and use Linux scheduling policies instead

**Alternate Designs**
Newer NDK have `SCHED_IDLE` defined and we should use that instead.

I am also not a fan of directly writing to files instead of using system calls.

Direct access to `/dev/cpuctl` directory is not available Android 10 anyway without root.

`set_sched_policy()` can be used to update this code but its considered legacy, see https://source.android.com/devices/tech/perf/cgroups

This is a cgroup feature used by Android system internally and should not be used by app.
see `PCY` sections:
https://stackoverflow.com/questions/10051152/does-any-one-know-the-top-command-s-results-meaning-in-android
https://stackoverflow.com/questions/32701187/adb-shell-top-shows-my-application-in-foreground-but-actually-its-in-background
https://android.googlesource.com/platform/external/toybox/+/master/toys/posix/ps.c

**Release Notes**
Client: Remove old Android scheduling method

### DIFF
--- a/client/app_start.cpp
+++ b/client/app_start.cpp
@@ -1078,18 +1078,6 @@ int ACTIVE_TASK::start(bool test) {
                 perror("setpriority");
             }
 #endif
-#ifdef ANDROID
-            // Android has its own notion of background scheduling
-            if (!high_priority) {
-                FILE* f = fopen("/dev/cpuctl/apps/bg_non_interactive/tasks", "w");
-                if (!f) {
-                    msg_printf(NULL, MSG_INFO, "Can't open /dev/cpuctl/apps/bg_non_interactive/tasks");
-                } else {
-                    fprintf(f, "%d", getpid());
-                    fclose(f);
-                }
-            }
-#endif
 #if HAVE_SCHED_SETSCHEDULER && defined(SCHED_IDLE) && defined (__linux__)
             if (!high_priority) {
                 struct sched_param sp;


truboxl:
May or may not fix linked issues:
https://github.com/BOINC/boinc/issues/2549 (see JacobWKlein's comment)

truboxl:
I am changing this PR to full removal of this code

`/dev/cpuctl/bg_non_interactive/` has been removed in Android init.rc
https://android.googlesource.com/platform/system/core/+/a27a91a5fd23730c9ef0af751cddf119d869e029%5E%21/
https://android.googlesource.com/platform/system/core/+/master/rootdir/init.rc#301

It is based on cgroup feature used by Android which means it is supposed to be used at kernel / OEM level
The fact that the equivalent `set_sched_policy()` mentioned only at `source.android.com`
is information for ROM builders and kernel developers, not app developers

It is also noted that in the commit message that it is unreliable which may cause such different behaviour
in different devices since init.rc is customised by different manufacturers, heck may even freeze the device
Users also cannot control it unless have root access

I think the change happened in Android 8.0+ since that's where I find `bg_non_interactive` missing
But some OEMs may or may not include this change so this does not mean all 8.0+ do not have issues
Android 10 just block access to `/dev/cpuctl` for apps anyway
https://android.googlesource.com/platform/system/core/+/android-8.0.0_r37/rootdir/init.rc#145
https://android.googlesource.com/platform/system/core/+/android-7.1.1_r61/rootdir/init.rc#139

It would be better off to remove this code that uses non-user-changeable parameters
(settings in `/dev/cpuctl/bg_non_interactive` are access denied anyway in newer Android)
and have something equivalent in the GUI that can be controlled by the user if possible
(switch between fg and bg)

or just rely on Linux scheduling policies
and use settings that controls number of CPU cores and CPU time used
which work 100% more reliably

talregev:
Can you align with the master that it will do 23 checks?

PR #4053, opened 2020-10-13
[Android] Update Android Gradle plugin to 4.1.

Isira-Seneviratne:
**Description of the Change**
Update the Android Gradle plugin from 4.1 RC2 to 4.1 stable.

**Release Notes**
N/A


AenBleidd:
Thanks for the update

PR #3956, opened 2020-08-02
Mac: Automatically build Universal x86_64 / arm64 binaries

CharlieFenton:
Mac: Automatically build:
* Universal x86_64 / arm64 binaries on Xcode versions that support building both architectures
* x86_64-only binaries on _Xcode_ versions that support only x86_64
* arm64-only binaries on _Xcode_ versions that support only arm64.

CAUTION # 1: I don't have arm64 macintosh hardware so this has been tested only on an x86_64 system.
CAUTION # 2: This was developed on _MacOS 11.0 beta 3_ and _Xcode 12.0 beta 2_, not final release builds
CAUTION # 3: At this time _OpenSSL_ does not yet provide a way to build for MacOS with arm64 architecture, so I implemented a temporary workaround, which may or may not work properly on arm64 Macs.

Fixes #3827 

Apple has announced that they will be transitioning to all arm64 ("Apple Silicon") Macs over the next two years, with the first ones to be released at the end of 2020. Apple Silicon Macs will initially include an emulator or virtual machine capable of running x86_64 code, but these will run less efficiently than native arm64 binaries and this capability will be phased out over time.

At this time, some (but not all) builds of _Xcode 12_ support building binaries for both Intel x86_64 Macintoshes and Apple Silicon (arm64) Macs. While these will automatically build _Xcode_ based projects as Universal Binaries containing both architectures, BOINC's third party dependent libraries use _configure_ and _make_ and so require additional effort to produce Universal Binaries. And BOINC code itself required a few modifications to be compatible with arm64 Macs.

This PR does the following:

* Updates the build scripts for the dependent libraries _c-ares_, _curl_, _freestyle_, _ftgl_, _openssl_ and _wxWidgets_ to automatically build for x86-64 architecture, arm64 architecture, or both, depending on the capabilities of the _Xcode_ version being used.

* Eliminates obsolete build settings in the _Xcode_ project which were incompatible with building for arm64.

* Adds arm64-apple-darwin as a valid BOINC platform in addition to i686-apple-darwin and x86_64-apple-darwin. (Note that the 32-bit i686-apple-darwin platform is only available on versions of MacOS prior to _OS 10.15_.)

* Updates the release script release_boinc.sh to name the directories containing the installer and associated files boinc_x.y.z_macOSX_x86_64boinc_x.y.z_macOSX_arm64 or  boinc_x.y.z_macOSX_universal as appropriate. It similarly updates the release_brand.sh script to label its products with the appropriate architecture.

[...]

PR #4062, opened 2020-10-20
Update OpenSSL build script to build arm64 and x86_64 Universal binary

CharlieFenton:
Adapts code I found in the OpenSSL GitHub project to correctly build OpenSSL for BOINC as a Universal 2 binary containing code for both Intel (x86_64) and the coming new Apple Silicon (arm64) Macs. This is a more reliable implementation the the one I made as part of #3956, since it patches an openSSL configuration file to match a change by an OpenSSL project developer.


CharlieFenton:
This has been tested and is ready to be merged. 

 I built BOINC 7.16.11 for Apple Silicon (arm64) Macs using the OpenSSL library built using the code in this PR. @AustinConlon has tested it and confirmed that it successfully communicates with BOINC projects.

PR #4066, opened 2020-10-23
client: disable GET feature of GUI RPC

davidpanderson:
This feature isn't being used and is a security risk.


TheAspens:
This looks correct.   I will hold off merging this over the weekend in case anyone wants to comment.  Thanks for making the change.

AenBleidd:
Personally I'd completely remove it from the code as well. In any case it will be saved in git history and could be reverted if necessary.

Rytiss:
Why are there code changes dealing with path handling, and all of it later gets disabled? I agree with @AenBleidd about removing it completely.

davidpanderson:
This feature may be used in the future.  It's easier to remove #if 0 than to scan git history.

PR #4069, opened 2020-10-27
Support plan classes with negative host_summary_regex matching

Rytiss:
**Description of the Change**
Currently host_summary_regex field in the plan class definition only allows to send work to hosts matching a certain spec. This change allows to specify a negative regex to send work to hosts NOT matching the specified regex.

Use case: send work only to hosts without Virtualbox.

**Release Notes**
Support plan classes with negative host_summary_regex matching.

### DIFF
--- a/sched/plan_class_spec.cpp
+++ b/sched/plan_class_spec.cpp
@@ -290,6 +290,17 @@ bool PLAN_CLASS_SPEC::check(
         }
         return false;
     }
+    if (have_host_summary_regex_negative
+        && (regexec(&(host_summary_regex_negative), g_reply->host.serialnum, 0, NULL, 0) == 0)
+    ) {
+        if (config.debug_version_select) {
+            log_messages.printf(MSG_NORMAL,
+                "[version] plan_class_spec: host summary '%s' matched negative regexp\n",
+                g_reply->host.serialnum
+            );
+        }
+        return false;
+    }
 
     // OS version
     //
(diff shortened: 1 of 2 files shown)


Rytiss:
Note: needs doc updates at https://boinc.berkeley.edu/trac/wiki/AppPlanSpec but I seem to have misplaced my password, and password reset also cannot find my account.

bema-aei:
The current 'master' version of the plan_class_spec code differs a bit from what I sent David originally. In our (Einstein@Home) code I do the following to achieve negated regex (and better logging that may help with debugging):
- The originally parsed string is kept in a separate variable (usually tag_string, here host_summary_string). This also allows for more explicit logging.
- in parse(), if the string starts with an `!`, the regex (here host_summary_regex) is compiled without it (i.e. from `&(host_summary_string[1])` or `host_summary_string+1`, i.e. skipping the `!`.
- in check(), the truth value of the regex match is compared to the first character of the string being `!`
Here's an example (`cpu_model`)

in PLAN_CLASS_SPEC:
```
  regex_t cpu_model_regex; /* a regexp to match the device name */
  char   cpu_model_string[256];
```
in parse() (this is an old parser, but you get th eidea):
```
                if(xp.parse_str(tag, "cpu_model", pc.cpu_model_string, sizeof(pc.cpu_model_string))) {
                  char*c=pc.cpu_model_string;
                  if (*c == '!') {
                    c++;
                  }
                  if ( regcomp(&(pc.cpu_model_regex), c, REG_EXTENDED|REG_NOSUB) ) {
                    log_messages.printf(MSG_CRITICAL, "BAD REGEXP: %s\n", pc.cpu_model_string);
                    return ERR_XML_PARSE;
                  }
                  continue;
                }
```
in check()
```
  // CPU model (regexp match)                                                                                                                                                                                                                                                                                                                                            
  if((strlen(cpu_model_string) > 0) &&
     ((regexec(&(cpu_model_regex), sreq.host.p_model, 0, NULL, 0) == 0) == (*cpu_model_string == '!'))) {
    if (config.debug_version_select)
      log_messages.printf(MSG_NORMAL,
                          "[version] CPU model '%s' doesn't match '%s'\n",
                          sreq.host.p_model, cpu_model_string);
    return false;
  }
```

davidpanderson:
I created a PR #4390 which allows negation of any of the regular expressions in a plan class spec (OS, CPU vendor, etc.)

Issue #3715, opened 2020-05-14
Linux Client: Multiple bugs in idle detection


RichardHaselgrove:
**Describe the bug**
Not my personal observation, but drawing attention to a detailed error report and workround in
https://boinc.berkeley.edu/forum_thread.php?id=13673
Multiple users reporting.

**Steps To Reproduce**
1. Problem reported in v7.16.6 release
2. Problem resolved by downgrading to v7.16.3

**Expected behavior**
Return to previous behavior - suspend when user is active, resume computation when user is inactive.

**System Information**
 - OS: Ubuntu/Kubuntu 20.04 (change from behavior in 19.10)
 - BOINC Version: 7.16.6 (problem exists), 7.16.3 (problem resolved)

**Additional context**
We should have a change list somewhere so that we can identify the exact changes between v7.16.3 and v7.16.6, where this might have been introduced.

Germano0:
This is no longer a problem on Fedora / CentOS / RHEL due my patch
https://src.fedoraproject.org/rpms/boinc-client/blob/master/f/disable_idle_time_detection.patch

RichardHaselgrove:
Unfortunately, simply smothering the bug kills the intended behaviour (noise reduction, freeing resources for paying work) too.

mpsdskd:
Then I could also just disable idle detection in the GUI and have the same effect?

> This is no longer a problem on Fedora / CentOS / RHEL due my patch
> https://src.fedoraproject.org/rpms/boinc-client/blob/master/f/disable_idle_time_detection.patch
> Note: I have tested (again) BOINC idle time detection before deciding to remove all the code

RichardHaselgrove:
Better that we find what stopped it working and restore the intended behaviour.

Germano0:
Actually I was remembering it wrong
https://bugzilla.redhat.com/show_bug.cgi?id=1822723#c2

mpsdskd:
In the boinc forum thread, one user also references the following _folding@home_ forum threads:

https://foldingforum.org/viewtopic.php?f=89&t=33351
https://foldingforum.org/viewtopic.php?f=89&t=33898

In the first one there is a workaround script, which controls F@H using xprintidle - might also work for boinc, but is also at best a temporary workaround.

Agbar:
[...]

PR #4071, opened 2020-10-28
client: allow an empty GUI RPC password, but generate alert message

davidpanderson:
boinccmd: show alert messages after attach RPCs

PR #3709 disallowed empty GUI RPC password files.
This increased security on shared machines.
But it meant that on Linux, after installing BOINC as a package,
the user had to locate and change the protection
and/or the ownership of the password file, which is undesirable.

This change allows empty password files but tells the user
that they should think about the security implications.
With the Manager this is delivered as a notice.
With boinccmd the message is written to stderr after an attach operation.

### DIFF
--- a/lib/gui_rpc_client.cpp
+++ b/lib/gui_rpc_client.cpp
@@ -478,16 +478,10 @@ int read_gui_rpc_password(char* buf, string& msg) {
         return ERR_FOPEN;
 #endif
     }
-    char* p = fgets(buf, 256, f);
-    if (p) {
-        // trim CR
-        //
-        int n = (int)strlen(buf);
-        if (n && buf[n-1]=='\n') {
-            buf[n-1] = 0;
-        }
+    buf[0] = 0;
+    if (fgets(buf, 256, f)) {
+        strip_whitespace(buf);
     }
     fclose(f);
     return 0;
 }
-
--- a/lib/gui_rpc_client.cpp
+++ b/lib/gui_rpc_client.cpp
@@ -455,8 +455,13 @@ int read_gui_rpc_password(char* buf, string& msg) {
                 return ERR_FOPEN;
             }
         } else {
-            sprintf(msg_buf, "%s not found.  Try reinstalling BOINC.",
-                GUI_RPC_PASSWD_FILE
+            char buf2[MAXPATHLEN];
+            if (!getcwd(buf2, MAXPATHLEN)) {
+                strcpy(buf2, "");
+            }
+            sprintf(msg_buf, "No BOINC data directory was specified, and %s was not found in the current directory (%s).  See https://boinc.berkeley.edu/gui_rpc.php for more information.",
+                GUI_RPC_PASSWD_FILE,
+                buf2
             );
             msg = msg_buf;
             return ERR_FOPEN;
--- a/lib/gui_rpc_client.cpp
+++ b/lib/gui_rpc_client.cpp
@@ -401,15 +401,22 @@ int RPC::parse_reply() {
 
 // Look for a GUI RPC password file and read it.
 // If fail, return a prescriptive message.
-// Win/Mac: look in current dir.
-// Linux: also look in a directory specified in
-// /etc/boinc-client/config.properties
+// Win/Mac: look in
+//  - current dir
+// Linux: look in:
+//  - current dir
+//  - a directory specified in /etc/boinc-client/config.properties
+//  - /var/lib/boinc-client
 //
+// Note: the Manager (on all platforms) has a -datadir cmdline option.
+// If present, it chdirs to that directory.
+
 int read_gui_rpc_password(char* buf, string& msg) {
     char msg_buf[1024];
     FILE* f = fopen(GUI_RPC_PASSWD_FILE, "r");
     if (!f) {
 #if defined(__linux__)
+        char path[MAXPATHLEN];
         if (errno == EACCES) {
             sprintf(msg_buf,
                 "%s exists but can't be read.  Check the file permissions.",
@@ -418,9 +425,12 @@ int read_gui_rpc_password(char* buf, string& msg) {
             msg = msg_buf;
             return ERR_FOPEN;
         }
+
+        // look for config file
+        //
         FILE* g = fopen(LINUX_CONFIG_FILE, "r");
         if (g) {
-            char buf2[MAXPATHLEN], path[MAXPATHLEN];
+            char buf2[MAXPATHLEN];
             char *p = 0;
             while (fgets(buf2, MAXPATHLEN, g)) {
                 strip_whitespace(buf2);
@@ -455,16 +465,30 @@ int read_gui_rpc_password(char* buf, string& msg) {
                 return ERR_FOPEN;
             }
         } else {
-            char buf2[MAXPATHLEN];
-            if (!getcwd(buf2, MAXPATHLEN)) {
-                strcpy(buf2, "");
+            // no config file; look in default data dir
+            //
+            sprintf(path, "%s/%s", LINUX_DEFAULT_DATA_DIR, GUI_RPC_PASSWD_FILE);
+            f = fopen(path, "r");
+            if (!f) {
+                if (errno == EACCES) {
+                    sprintf(msg_buf,
+                        "%s exists but can't be read.  Check the file permissions.",
+                        path
+                    );
+                    msg = msg_buf;
+                    return ERR_FOPEN;
+                }
+                char buf2[MAXPATHLEN];
+                if (!getcwd(buf2, MAXPATHLEN)) {
+                    strcpy(buf2, "");
+                }
+                sprintf(msg_buf, "No BOINC data directory was specified, and %s was not found in the current directory (%s).  See https://boinc.berkeley.edu/gui_rpc.php for more information.",
+                    GUI_RPC_PASSWD_FILE,
+                    buf2
+                );
+                msg = msg_buf;
+                return ERR_FOPEN;
             }
-            sprintf(msg_buf, "No BOINC data directory was specified, and %s was not found in the current directory (%s).  See https://boinc.berkeley.edu/gui_rpc.php for more information.",
-                GUI_RPC_PASSWD_FILE,
-                buf2
-            );
-            msg = msg_buf;
-            return ERR_FOPEN;
         }
 #else
         // non-Linux
--- a/lib/gui_rpc_client.cpp
+++ b/lib/gui_rpc_client.cpp
@@ -416,11 +416,12 @@ int read_gui_rpc_password(char* buf, string& msg) {
     FILE* f = fopen(GUI_RPC_PASSWD_FILE, "r");
     if (!f) {
 #if defined(__linux__)
+#define HELP_URL "https://boinc.berkeley.edu/gui_rpc.php"
         char path[MAXPATHLEN];
         if (errno == EACCES) {
             sprintf(msg_buf,
-                "%s exists but can't be read.  Check the file permissions.",
-                GUI_RPC_PASSWD_FILE
+                "%s exists but can't be read.  See %s",
+                GUI_RPC_PASSWD_FILE, HELP_URL
             );
             msg = msg_buf;
             return ERR_FOPEN;
@@ -445,12 +446,12 @@ int read_gui_rpc_password(char* buf, string& msg) {
                 if (!f) {
                     if (errno == EACCES) {
                         sprintf(msg_buf,
-                            "%s exists but can't be read.  Check the file permissions.",
-                            path
+                            "%s exists but can't be read.  See %s",
+                            path, HELP_URL
                         );
                     } else {
-                        sprintf(msg_buf, "%s not found.  Try reinstalling BOINC.",
-                            path
+                        sprintf(msg_buf, "%s not found.  See %s",
+                            path, HELP_URL
                         );
                     }
                     msg = msg_buf;
@@ -458,8 +459,8 @@ int read_gui_rpc_password(char* buf, string& msg) {
                 }
             } else {
                 sprintf(msg_buf,
-                    "No data_dir= found in %s.  Try reinstalling BOINC.",
-                    LINUX_CONFIG_FILE
+                    "No data_dir= found in %s.  See %s",
+                    LINUX_CONFIG_FILE, HELP_URL
                 );
                 msg = msg_buf;
                 return ERR_FOPEN;
@@ -472,19 +473,14 @@ int read_gui_rpc_password(char* buf, string& msg) {
             if (!f) {
                 if (errno == EACCES) {
                     sprintf(msg_buf,
-                        "%s exists but can't be read.  Check the file permissions.",
-                        path
+                        "%s exists but can't be read.  See %s",
+                        path, HELP_URL
                     );
                     msg = msg_buf;
                     return ERR_FOPEN;
                 }
-                char buf2[MAXPATHLEN];
-                if (!getcwd(buf2, MAXPATHLEN)) {
-                    strcpy(buf2, "");
-                }
-                sprintf(msg_buf, "No BOINC data directory was specified, and %s was not found in the current directory (%s).  See https://boinc.berkeley.edu/gui_rpc.php for more information.",
-                    GUI_RPC_PASSWD_FILE,
-                    buf2
+                sprintf(msg_buf, "%s not found.  See %s",
+                    GUI_RPC_PASSWD_FILE, HELP_URL
                 );
                 msg = msg_buf;
                 return ERR_FOPEN;
--- a/client/acct_mgr.cpp
+++ b/client/acct_mgr.cpp
@@ -56,7 +56,7 @@ static const char *run_mode_name[] = {"", "always", "auto", "never"};
 int ACCT_MGR_OP::do_rpc(ACCT_MGR_INFO& _ami, bool _via_gui) {
     int retval;
     unsigned int i;
-    char buf[256];
+    char buf[1024];
 
     ami = _ami;
 
--- a/client/boinc_cmd.cpp
+++ b/client/boinc_cmd.cpp
@@ -163,6 +163,30 @@ void acct_mgr_do_rpc(
     }
 }
 
+// Get messages from client, and show any that are USER_ALERT priority.
+// Intended use: show user that GUI RPCs are not password-protected.
+// For now, do this after attach to project or AM
+//
+void show_alerts(RPC_CLIENT &rpc) {
+    MESSAGES messages;
+    int retval = rpc.get_messages(0, messages);
+    if (retval) {
+        fprintf(stderr, "Can't get alerts from client: %s\n",
+            boincerror(retval)
+        );
+        return;
+    }
+    for (unsigned int j=0; j<messages.messages.size(); j++) {
+        MESSAGE& md = *messages.messages[j];
+        if (md.priority != MSG_USER_ALERT) continue;
+        if (!md.project.empty()) continue;
+        strip_whitespace(md.body);
+        fprintf(stderr, "\nAlert from client: %s\n",
+            md.body.c_str()
+        );
+    }
+}
+
 int main(int argc, char** argv) {
     RPC_CLIENT rpc;
     int i, retval, port=0;
@@ -382,6 +406,7 @@ int main(int argc, char** argv) {
         canonicalize_master_url(url, sizeof(url));
         char* auth = next_arg(argc, argv, i);
         retval = rpc.project_attach(url, auth, "");
+        show_alerts(rpc);
     } else if (!strcmp(cmd, "--file_transfer")) {
         FILE_TRANSFER ft;
 
@@ -529,6 +554,7 @@ int main(int argc, char** argv) {
             char* am_name = next_arg(argc, argv, i);
             char* am_passwd = next_arg(argc, argv, i);
             acct_mgr_do_rpc(rpc, am_url, am_name, am_passwd);
+            show_alerts(rpc);
         } else if (!strcmp(op, "info")) {
             ACCT_MGR_INFO ami;
             retval = rpc.acct_mgr_info(ami);
--- a/client/boinc_cmd.cpp
+++ b/client/boinc_cmd.cpp
@@ -193,7 +193,7 @@ int main(int argc, char** argv) {
     MESSAGES messages;
     NOTICES notices;
     char passwd_buf[256], hostname_buf[256], *hostname=0;
-    char* passwd = passwd_buf, *p, *q;
+    char* passwd=0, *p, *q;
     bool unix_domain = false;
     string msg;
 
@@ -202,12 +202,6 @@ int main(int argc, char** argv) {
 #elif defined(__APPLE__)
     chdir("/Library/Application Support/BOINC Data");
 #endif
-    safe_strcpy(passwd_buf, "");
-    retval = read_gui_rpc_password(passwd_buf, msg);
-    if (retval) {
-        fprintf(stderr, "Can't get RPC password: %s\n", msg.c_str());
-        fprintf(stderr, "Only operations not requiring authorization will be allowed.\n");
-    }
 
 #if defined(_WIN32) && defined(USE_WINSOCK)
     WSADATA wsdata;
@@ -217,6 +211,11 @@ int main(int argc, char** argv) {
         exit(1);
     }
 #endif
+
+    // parse command line.
+    // TODO: do this the right way.
+    // shouldn't require args to be in a particular order.
+
     if (argc < 2) usage();
     i = 1;
     if (!strcmp(argv[i], "--help")) usage();
@@ -268,6 +267,16 @@ int main(int argc, char** argv) {
         i++;
     }
     if (i == argc) usage();
+
+    if (!passwd) {
+        passwd = passwd_buf;
+        safe_strcpy(passwd_buf, "");
+        retval = read_gui_rpc_password(passwd_buf, msg);
+        if (retval) {
+            fprintf(stderr, "Can't get RPC password: %s\n", msg.c_str());
+            fprintf(stderr, "Only operations not requiring authorization will be allowed.\n");
+        }
+    }
 
     // change the following to debug GUI RPC's asynchronous connection mechanism
     //
--- a/client/gui_rpc_server.cpp
+++ b/client/gui_rpc_server.cpp
@@ -120,7 +120,7 @@ bool GUI_RPC_CONN_SET::recent_rpc_needs_network(double interval) {
 }
 
 // read the GUI RPC password from gui_rpc_auth.cfg;
-// create one if missing or empty.
+// create one if missing
 //
 void GUI_RPC_CONN_SET::get_password() {
     int retval;
@@ -132,16 +132,15 @@ void GUI_RPC_CONN_SET::get_password() {
             strip_whitespace(password);
         }
         fclose(f);
-        if (strlen(password)) {
-            return;
-        }
 
-        // File is empty; don't allow this.
-        // Fall through and create a password.
+        // if password is empty, allow it but issue a warning
         //
-        msg_printf(NULL, MSG_INFO,
-            "%s is empty - assigning new GUI RPC password", GUI_RPC_PASSWD_FILE
-        );
+        if (!strlen(password)) {
+            msg_printf(NULL, MSG_USER_ALERT,
+                "Warning: GUI RPC password is empty.  BOINC can be controlled by any user on this computer.  See https://boinc.berkeley.edu/gui_rpc_passwd.php for more information."
+            );
+        }
+        return;
     }
 
     // make a random password
--- a/client/switcher.cpp
+++ b/client/switcher.cpp
@@ -48,14 +48,15 @@ int main(int /*argc*/, char** argv) {
     APP_INIT_DATA   aid;
     FILE            *f;
     int             retval = -1;
-    int             i;
     char            libpath[8192];
     char            newlibs[256];
     char            *projectDirName;
     const char      *screensaverLoginUser = NULL;
+#ifdef __APPLE__
+    int             i;
     bool            launching_gfx=false;
     char            current_dir[MAXPATHLEN];
-
+#endif
 
     strcpy(boinc_project_user_name, "boinc_project");
     strcpy(boinc_project_group_name, "boinc_project");
--- a/lib/common_defs.h
+++ b/lib/common_defs.h
@@ -376,7 +376,14 @@ struct DEVICE_STATUS {
 #else
 #define DEFAULT_SS_EXECUTABLE       "boincscr"
 #endif
+
 #define LINUX_CONFIG_FILE           "/etc/boinc-client/config.properties"
+
+// Used by Manager and boinccmd to locate the data dir.
+// You can define this in "configure" if you want.
+//
+#ifndef LINUX_DEFAULT_DATA_DIR
 #define LINUX_DEFAULT_DATA_DIR      "/var/lib/boinc-client"
+#endif
 
 #endif
--- a/lib/common_defs.h
+++ b/lib/common_defs.h
@@ -377,5 +377,6 @@ struct DEVICE_STATUS {
 #define DEFAULT_SS_EXECUTABLE       "boincscr"
 #endif
 #define LINUX_CONFIG_FILE           "/etc/boinc-client/config.properties"
+#define LINUX_DEFAULT_DATA_DIR      "/var/lib/boinc-client"
 
 #endif


Germano0:
I made some tests. My gui_rpc_auth.cfg is empty. The default access permission to the file are

```
# ls -lah /var/lib/boinc/gui_rpc_auth.cfg 
-rw-------. 1 boinc boinc 1 26 set 23.14 /var/lib/boinc/gui_rpc_auth.cfg
```
but I also tried rwrwrw and the results are the same.

**Command line interface**
```
$ boinccmd --get_tasks
Can't get RPC password: gui_rpc_auth.cfg not found.  Try reinstalling BOINC.
Only operations not requiring authorization will be allowed.

======== Tasks ========
[ CUT ]
```
**boinc-manager**
In Alerts(Messages) tab:
```
Warning: GUI RPC password is empty. BOINC can be controlled by any user on this computer. See https://boinc.berkeley.edu/gui_rpc_passwd.php for more information.
```

So while boinc-manager says that the password file is empty, boinccmd instead always says that the file is missing

TheAspens:
@Germano0 - did you have a file at /etc/boinc-client/config.properties that contained an entry "data-dir" that pointed to the data directory?

TheAspens:
@davidpanderson  - I'm getting some odd behavior.  

In all cases, I have the following:

```
[knreed@oc4182665077 boinc]$ ls -lh /etc/boinc-client/config.properties 
-rw-r--r--. 1 boinc boinc 24 Oct 30 11:11 /etc/boinc-client/config.properties
[knreed@oc4182665077 boinc]$ cat /etc/boinc-client/config.properties 
data_dir=/var/lib/boinc
```

and each of the first 4 cases below, the core client started and logged a warning "Warning: GUI RPC password is empty.  BOINC can be controlled by any user on this computer.  See https://boinc.berkeley.edu/gui_rpc_passwd.php for more information." and in the 5th case, no message was logged.  This part is correct.

**Case 1: An empty gui_rpc_auth.cfg file as would be created using touch**
```
[knreed@oc4182665077 boinc]$ ls -lh /var/lib/boinc/gui_rpc_auth.cfg 
-rw-r-----. 1 boinc boinc 0 Oct 30 11:33 /var/lib/boinc/gui_rpc_auth.cfg
```
then the command `boinccmd --project http://www.worldcommunitygrid.org update` works without any output (and the client does an update)

the boincmgr opens with an error "BOINC Manager - Connection Error "Invalid client RPC password.  Try reinstalling BOINC"

**I don't think this is the intended behavior.**  I believe that the boinc manager should have connected fine, but then gave a notice warning about the empty password

**Case 2: An empty gui_rpc_auth.cfg file (but 1 byte file) as would be created using vi and then saving the file**
```
[knreed@oc4182665077 boinc]$ ls -lh /var/lib/boinc/gui_rpc_auth.cfg 
-rw-r-----. 1 boinc boinc 1 Oct 30 11:25 /var/lib/boinc/gui_rpc_auth.cfg
```
then the command `boinccmd --project http://www.worldcommunitygrid.org update` works without any output (and the client does an update)

the boincmgr opens and gives a "Notice from BOINC" in the notices tab warning of the empty password.

**This appears to work as it should**

**Case 3: A gui_rpc_auth.cfg file with a single space**
```
[knreed@oc4182665077 boinc]$ ls -lh /var/lib/boinc/gui_rpc_auth.cfg 
-rw-r-----. 1 boinc boinc 2 Oct 30 11:28 /var/lib/boinc/gui_rpc_auth.cfg
```
then the command `boinccmd --project http://www.worldcommunitygrid.org update`  fails with output `Authorization failure: -155`

the boincmgr opens with an error "BOINC Manager - Connection Error "Invalid client RPC password.  Try reinstalling BOINC"

**I do not believe this is the intended behavior.** I believe that this should have had the same outcome as in case 2 so something is wrong  here

**Case 4: A gui_rpc_auth.cfg with a single newline**
```
[knreed@oc4182665077 boinc]$ ls -lh /var/lib/boinc/gui_rpc_auth.cfg 
-rw-r-----. 1 boinc boinc 2 Oct 30 11:31 /var/lib/boinc/gui_rpc_auth.cfg
```

then the command `boinccmd --project http://www.worldcommunitygrid.org update` works without any output (and the client does an update)

the boincmgr opens and gives a "Notice from BOINC" in the notices tab warning of the empty password.

so this appears to work as it should

**Case 5: The gui_rpc_auth.cfg has a random string**
```
[knreed@oc4182665077 boinc]$ ls -lh /var/lib/boinc/gui_rpc_auth.cfg 
-rw-r-----. 1 boinc boinc 22 Oct 30 11:40 /var/lib/boinc/gui_rpc_auth.cfg
```

The core client starts up and does not log any messages about an empty file.

then the command `boinccmd --project http://www.worldcommunitygrid.org update` works without any output (and the client does an update)

the boincmgr opens and does not give a a message about password

so this appears to work as it should

**Conclusion**
something is wrong in cases 1 and 3

RichardHaselgrove:
@TheAspens - you asked in https://github.com/BOINC/boinc/pull/4071#issuecomment-719651991 "did you have a file at /etc/boinc-client/config.properties that contained an entry "data_dir" that pointed to the data directory?"

It was your suggestion in https://github.com/BOINC/boinc/pull/3709#issuecomment-628343486 that this was a way out of the dilemma that we had created for ourselves. But I don't see a crucial implementation detail - who is responsible for creating and deploying this file?

My Linux (Mint) machines have BOINC deployed via @LocutusOfBorg's PPA. They are currently running the v7.16.11 release versions of all BOINC programs, deployed via that PPA. There is no more recent version waiting for installation.

Yet my machines do **NOT** have that config.properties file. Did something drop through the cracks here?

TheAspens:
@RichardHaselgrove  - it has to be the package maintainers.  They actually have a lot of additional files that they deploy - for example the file /usr/lib/systemd/system/boinc-client.service which they deploy to run boinc under systemd (which I think are well done).

The package maintainers are also the ones who established the difference between /var/lib/boinc and /var/lib/boinc-client as the different data directories between the two major flavors of linux.  As a result, that difference in how the packages are created and deployed are directly impacting how this file should be configured and thus should be specific to the deployment package and maintained by the package maintainers.

TheAspens:
I should also mention that we did email the known package maintainers about this back in May.

RichardHaselgrove:
But @Germano0 is one of the package maintainers, and he didn't seem to be aware of it.

Germano0:
> The package maintainers are also the ones who established the difference between /var/lib/boinc and /var/lib/boinc-client as the different data directories between the two major flavors of linux

The decision of using `/var/lib/boinc` instead of `/var/lib/boinc-client` is a thing that we inherited from people that introduced boinc-client into Fedora repositories and are no longer package maintainers. Changing it would broke the settings of all BOINC users in Fedora / EPEL.

Concerning `/etc/boinc-client/config.properties`, I just made an update (boinc-client-7.16.11-2) with this new file https://bodhi.fedoraproject.org/updates/?packages=boinc-client

RichardHaselgrove:
@Germano0 - I believe @LocutusOfBorg's work is derived from that file? I'll watch out for that update and report back.

Germano0:
> @Germano0 - I believe @LocutusOfBorg's work is derived from that file? I'll watch out for that update and report back.

I don't know

davidpanderson:
Kevin, I committed a change that (I think) fixes the problems you saw.

smoe:
I had a first somewhat deeper look into this. But I do not get what you were doing. Can we have that configuration file you came up with only as an option instead of a requirement, please? Much like sshd has default settings and just works? On line https://github.com/BOINC/boinc/blob/4bf7d903bbbfc9bfadc528e6801c797618b71a95/lib/gui_rpc_client.cpp#L449 then please test if /var/lib/boinc-client (or boinc on RPM systems) exists and then just use that as a default and bail out only if that is not existing. If I add that out-of-the-blue configuration file then we configure the same thing at two different locations
```
$ grep -i dir /etc/default/boinc-client
# This is the data directory of the BOINC core client.
BOINC_DIR="/var/lib/boinc-client"
```
and you may not be surprised that I do not like this too much. I presume that entry should then go, right?

To be slightly more constructive I herewith offer to add that pesky configuration file, if in exchange (and I had also asked in May) you release a version 7.6.16, such that we can get back in sync with our version numbering after the extracurricular 7.6.15 of yours in February. Please.

smoe:
I added the commit of this PR to https://salsa.debian.org/pkg-boinc-team/boinc, together with a config.properties file. boinccmd and boincmgr work, boinctui does not - unauthorized access to localhost. Could this be related? I saw the message warning about an empty password in the Boinc Manager. Pending a comment from @LocutusOfBorg this version should then soon find its way to the Debian servers.

RichardHaselgrove:
I see that a new version of the @LocutusOfBorg PPA has enqueued itself in my disto's pending updates. It's a little late to do that sort of maintenance today: I'll do it in the morning, and report back.

Changelog: 
> boinc (7.16.15+dfsg.is.7.16.11+dfsg+202011100036~ubuntu18.04.1) bionic; urgency=low
> 
> * Auto build.

lib/gui_rpc_client.cpp
```diff
@@ -455,16 +465,30 @@ int read_gui_rpc_password(char* buf, string& msg) {
...
-            if (!getcwd(buf2, MAXPATHLEN)) {
-                strcpy(buf2, "");
+            // no config file; look in default data dir
+            //
+            sprintf(path, "%s/%s", LINUX_DEFAULT_DATA_DIR, GUI_RPC_PASSWD_FILE);
+            f = fopen(path, "r");
+            if (!f) {
+                if (errno == EACCES) {
+                    sprintf(msg_buf,
+                        "%s exists but can't be read.  Check the file permissions.",
```
smoe, line 475: Please add another "in %s" and ", LINUX_DEFAULT_DATA_DIR". Otherwise very nice, many thanks!

RichardHaselgrove:
**Summary**
@LocutusOfBorg's PPA now deploys the config.properties file. From the timings, I now presume this came from https://github.com/BOINC/boinc/pull/4071#issuecomment-724369436
This change is insufficient, because the general unprivileged user still does not have read permission on  gui_rpc_auth.cfg
We therefore need a recovery pathway for users who installed after [28 May](https://github.com/BOINC/boinc/pull/3709#event-3384063889) - this includes the v7.16.11 release.

**Full test report**
I come into the category of owner-operator-administrator of a home computer. Ever since my first venture into Linux, 18 months ago, my Linux machines have run under an ordinary user account, but I have the ability to enter administrative mode via a password when needed. So I see both sides of the security coin.

I ran into the permissions problem immediately. I wanted to set a GUI RPC password, so I could remotely monitor the machines: but with that in place, I couldn't run the Manager locally. So I developed two workrounds:

1) Place the password in the launch command for BOINC Manager (insecure - visible in plain text for anyone)
2) Add my personal account to the boinc group, **and reboot**

For today's PPA test, I reverted back to workround (1), so I could test both modes quickly. The situation remains as described: gui_rpc_auth.cfg can be found, but cannot be read without one or other of the workrounds.

In fact, under this PPA installation framework, the gui_rpc_auth.cfg file in /var/lib/boinc-client/ is a symlink: the real password file can be found in /etc/boinc-client/ - but as things stand, neither can be opened without one of the workrounds.

This may be the result of too many disjointed tests having been run on this machine. It looks as if the PPA's payload remains unchanged, at v7.16.11 release (no pre-release notice, no version change detected by the client when I installed today). So we are not performing a true end-to-end package test: I doubt any code changes committed since 30 August have been included. We will need to assemble a complete package, combining both BOINC code and deployment tools, and test it on a clean machine, before we can claim to have cracked the problem. And as I said in the summary, we will have to devise an escape patch for users who had an unwanted and uncontrolled password foisted on them by #3709.

And as I said in the very first line of the very first comment on that PR,

> can I ask you to liaise closely with Linux package maintainers like @Germano0 and @LocutusOfBorg in developing this idea further?

(and add @smoe to that list - sorry, Steffen). The messaging clearly didn't work last time.

smoe:
Thank you for your report, @RichardHaselgrove . I have a bit of a communication breakdown with Locutus and do not know if this pull request also found entry in his PPA package. Somehow too much, including mails from github, get sorted into my spam folder. I have not uploaded to Debian's server, yet, since I ran into a packaging failure, i.e. the systemd files are gone - very unrelated. The package works for me without them, but these kind of things Locutus knows. boinccmd and boinc-manager worked for me in total, also adding projects, which did not work with the stock Debian 7.16.11 version before. boinctui fails to connect, still, though. And I really started to love that.

Concerning links to us packaging-helpers - I cordially invite to get an account on salsa.debian.org and to then press the "join team" button on https://salsa.debian.org/pkg-boinc-team . 

Concerning permissions and passwords, only the administrator sets them. Yes, a group boinc is fine and if the config file is not installed with that group setting then this is a bug. Otherwise, if a regular user wants full control, that user shall compile their BOINC on their own or use a prepared binary and start boinc in a screen session or a VM. Sidenote, as you may kind of already guess, I still fail to understand what you need that configuration file for ;)

RichardHaselgrove:
@smoe - thanks for the kind words.

> what you need that configuration file for ;)

Which configuration file did you mean? config.properties? I'm not sure either - my distro / PPA didn't need it either. It already knew where to look for gui_rpc, just blocked by the permission problem. I'll defer to @TheAspens on that one.

The other points are off-topic for this PR. Would it be OK if I email you direct on the address you use for boinc_alpha?

LocutusOfBorg:
hello, sorry for the late answer.
I was already working on this issue, and came on the very same permission problem you are mentioning.
I didn't find any answer to the issue, so I didn't commit/push (something I'm doing shortly now)

for now, I did craft a patch like this
```
Description: Explain to the user how to fix the boinc startup issue
Author: Gianfranco Costamagna <locutusofborg@debian.org>
Last-Update: 2020-09-18

--- boinc-7.16.15+dfsg.is.7.16.11+dfsg.orig/lib/gui_rpc_client.cpp
+++ boinc-7.16.15+dfsg.is.7.16.11+dfsg/lib/gui_rpc_client.cpp
@@ -415,7 +415,7 @@ int read_gui_rpc_password(char* buf, str
 #if defined(__linux__)
         if (errno == EACCES) {
             sprintf(msg_buf,
-                "%s exists but can't be read.  Check the file permissions.",
+                "%s exists but can't be read.  Check the file permissions, or ask the system administrator to add the user in the 'boinc' group.",
                 GUI_RPC_PASSWD_FILE
             );
             msg = msg_buf;
```

Just to inform the end user that he should be allowed by the system administrator to run the manager daemon.

Not sure if this is enough user friendly or not.

Concerning the ppa it just automagically build what its in debian git repository, so each time somebody does a git push on salsa, a new ppa version comes out with the change built.
https://code.launchpad.net/~costamagnagianfranco/+archive/ubuntu/boinc

I git pushed all my changes

G.

smoe:
boinctui works on that machine when started with sudo. As regular user it does not start anywhere, not  even from within the data_dir.
@LocutusOfBorg , is your user a member of the boinc group? 
I was a bit surprised but I had to add this IIRC.

LocutusOfBorg:
yes, this is why the patch I added explictly says that, to add the user to boinc group to be able to manage the read of the file...

smoe:
> yes, this is why the patch I added explictly says that, to add the user to boinc group to be able to manage the read of the file...

And if the file is just readable to world? That location is not a secret after all.

smoe:
@davidpanderson When specifying the --dir on the command line then this has precedence over that static file, right?

AenBleidd:
> I see that a new version of the @LocutusOfBorg PPA has enqueued itself in my disto's pending updates. It's a little late to do that sort of maintenance today: I'll do it in the morning, and report back.

If you want to do any testing - do it with vanilla build from official BOINC repo: linux distros like 'debian' and some other contain custom patches that could affect any functionality of BOINC software. Of course you can use prebuilt binaries provided in any PPA but without vanilla build such testing will be not complete

RichardHaselgrove:
@AenBleidd: But we are struggling with a change (new file config.properties) which I persuaded @TheAspens to explicitly confirm has to form part of the distribution management.

For the Manager to obtain the client password, two problems have to be solved: it has to know **where** it is, and it has to be able to **read** it. The problems are different, but related: and they both relate to deployment and security choices made by the packager. Raw binaries are not the problem.

AenBleidd:
I agree but what I mean is that you can't fully rely on **non-vanilla** build to check changes that are done to **vanilla build**.
They should be checked both because of the custom patches we (as BOINC) do not control.

RichardHaselgrove:
OK, I have a second machine to test this morning. I will update from the PPA to get the operating environment up-to-date, and then I will overwrite the PPA binaries with the latest Travis artifacts to include a11bb44f131be94c55f78f6b71513722ed2d88ed. That'll be the most complete test I can think of.

To be honest, I haven't seen any code changes which would enable the Manager to read a non-readable file (permissions on gui_rpc_auth.cfg). But that's what testing's for! :)

LocutusOfBorg:
@RichardHaselgrove I think, we should "just" add a message to the user explaining why its important to ask to be added to boinc group.

I also think, if anybody wants to be able to run boinc manager without being in that group, can ask the system administrator to provide the password, and store it under ~/.boinc.

The manager, before trying to look on config.properties might want to test a local configuration file, in case the system has lots of users but only some of them can contact the client API. (and adding the user to boinc group might be an overkill)

LocutusOfBorg:
probably, if the "local configuration file" change is accepted upstream, we can change this
```
+        ch_stat 0644 root:boinc "$CONF_DIR/config.properties"
```
to something more secure like
```
+        ch_stat 0660 root:boinc "$CONF_DIR/config.properties"
```

RichardHaselgrove:
OK, stage 1 testing complete. I have:
Disabled workround(2) - removed myself from the boinc group - so I can see what's going on.
Updated the machine fully, and restarted. I get

Screenshot at 2020-11-12 10-07-40
Note the wording. That comes from https://github.com/BOINC/boinc/pull/4071#issuecomment-725583609.

Re-enabled workround(1), which confirms that the client is running properly.
I need to go out briefly, while the machine gets on with some work. I'll do the second stage test later.

@LocutusOfBorg - I see you've posted while I was drafting. It would be helpful for self-administrators if the message also indicated **how** to do this - but I'd still prefer it to be automated "in the right circumstances" (to avoid annoying corporate users).

RichardHaselgrove:
OK, I've located and downloaded [client/manager]_PR4071_2020-11-11_a11bb44f.7z, and I also see I have yet another PPA enqueued. But I'm running out of time before my next (short) appointment. I'll test both later.

Germano0:
Test made on a new machine using lastest [4071 patch](https://src.fedoraproject.org/rpms/boinc-client/blob/f33/f/4071.patch)
Build [boinc-client-7.16.11-3.fc33](https://bodhi.fedoraproject.org/updates/FEDORA-2020-df51b3e029)

```
# ls -lah /etc/boinc-client
totale 4,0K
drwxr-xr-x. 1 root root   34 12 nov 09.54 .
drwxr-xr-x. 1 root root 4,7K 12 nov 11.57 ..
-rw-r--r--. 1 root root   25 11 nov 21.53 config.properties
```

```
# ls -lah /var/lib/boinc
totale 96K
drwxr-xr-x. 1 boinc boinc  460 12 nov 15.11 .
drwxr-xr-x. 1 root  root   748 12 nov 09.55 ..
-rw-r--r--. 1 boinc boinc  51K 12 nov 09.55 all_projects_list.xml
-rw-r--r--. 1 boinc boinc 3,4K 12 nov 15.11 client_state_prev.xml
-rw-r--r--. 1 boinc boinc 3,4K 12 nov 15.11 client_state.xml
-rw-r--r--. 1 boinc boinc  300 12 nov 11.50 coproc_info.xml
-rw-r--r--. 1 boinc boinc  119 12 nov 15.11 daily_xfer_history.xml
-rw-r--r--. 1 boinc boinc 8,9K 12 nov 11.51 get_current_version.xml
-rw-r--r--. 1 boinc boinc 1,4K 12 nov 11.51 global_prefs.xml
-rw-r-----. 1 boinc boinc   32 12 nov 09.55 gui_rpc_auth.cfg
-rw-r--r--. 1 boinc boinc    0 12 nov 11.50 lockfile
drwxrwx--x. 1 boinc boinc   18 12 nov 15.10 notices
drwxrwx--x. 1 boinc boinc    0 12 nov 15.11 projects
drwxrwx--x. 1 boinc boinc    2 12 nov 11.51 slots
-rw-r--r--. 1 boinc boinc    0 12 nov 09.55 stderrgpudetect.txt
-rw-r--r--. 1 boinc boinc  204 12 nov 11.50 stdoutgpudetect.txt
-rw-r--r--. 1 boinc boinc  958 12 nov 13.24 time_stats_log
```

```
# cat /etc/boinc-client/config.properties 
data_dir=/var/lib/boinc
```

```
# cat /var/lib/boinc/gui_rpc_auth.cfg 
7189be8ab3965ba4ea5a1000328107ed
```

# user A, in boinc group
## boinccmd
```
$ boinccmd --passwd 7189be8ab3965ba4ea5a1000328107ed --project_attach https://lhcathome.cern.ch/lhcathome key_censored
```

project attached even if client password was not provided

## boinc-manager
after opening it, it connects to the client without and user action required. No password has been provided

# user B, not in boinc group

## boinccmd

### test 1
```
$ boinccmd --passwd 7189be8ab3965ba4ea5a1000328107ed --project_attach https://lhcathome.cern.ch/lhcathome key_censored
Can't get RPC password: /var/lib/boinc/gui_rpc_auth.cfg exists but can't be read.  Check the file permissions.
Only operations not requiring authorization will be allowed.
```

despite the error, the project has been be attached anyways.

### test 2
Removed the project, and tried with no passoword

```
$ boinccmd --project_attach https://lhcathome.cern.ch/lhcathome key_censored
Can't get RPC password: /var/lib/boinc/gui_rpc_auth.cfg exists but can't be read.  Check the file permissions.
Only operations not requiring authorization will be allowed.
Operation failed: authentication error
```

No project attached as expected

### test 3

Tried with a wrong client password
```
$ boinccmd --passwd s7189be8ab3965ba4ea5a1000328107ed --project_attach https://lhcathome.cern.ch/lhcathome key_censored
Can't get RPC password: /var/lib/boinc/gui_rpc_auth.cfg exists but can't be read.  Check the file permissions.
Only operations not requiring authorization will be allowed.
Authorization failure: -155
```

No project attached as expected


## boinc-manager
After opening it, you will get error
```
/var/lib/boinc/gui_rpc_auth.cfg exists but can't be read.  Check the file permissions.
```

then you can select a computer from File->Select computer, enter localhost and the password and manage the client without problems

# My personal conclusions
Since:
- if an user is in boinc group, he can manage the client without providing any password;
- if an user is not in boinc group, he must provide the password contained in `/var/lib/boinc/gui_rpc_auth.cfg` in order to manage the client password;

the admin could simply add the user to boinc group instead of having to use the password contained in `/var/lib/boinc/gui_rpc_auth.cfg`

smoe:
> I agree but what I mean is that you can't fully rely on **non-vanilla** build to check changes that are done to **vanilla build**.
> They should be checked both because of the custom patches we (as BOINC) do not control.

I happily see https://github.com/BOINC/boinc/issues/3260 revived.

smoe:
Anybody having success with boinctui?

RichardHaselgrove:
I've now tested both the new PPA, and the Travis artifacts representing the test build of this PR to a11bb44f131be94c55f78f6b71513722ed2d88ed (i.e. from vanilla BOINC code).

The PPA didn't behave any differently from the report I've already made.

With a password in gui_rpc_auth.cfg, but no password supplied, the vanilla message is 

Screenshot at 2020-11-12 18-58-07
('Pre-release' and different style confirms the change of Manager)

The command-line workround allows the Manager to connect to the client, as does the 'select computer' menu item.

An empty gui_rpc_auth.cfg is allowed, but generates a warning notice - as per the subject of this PR.

But if gui_rpc_auth.cfg is entirely missing, it is both created and populated with a 32-byte random string. I'm not sure if this was the intended final behaviour.

I haven't separately tested boinccmd, but my observations are consistent with @Germano0's description. I still think Windows refugees would expect to find themselves as user A (in boinc group) following installation, and wouldn't know how to proceed if they found themselves in the position of user B.

smoe:
@davidpanderson When specifying --dir on the command line then this has precedence over the information from that static file, right?

**TL;DR: New (-2) Debian version works after reboot.**

I just updated to the latest version in Debian and, well, somehow there now is a regression from what I think to have worked for me before and what works in Ubuntu from what Richard described. No password is set.

This should be the situation on a regular Debian system before the update:

```
$ boinccmd --project http://www.worldcommunitygrid.org/ suspend
Can't get RPC password: gui_rpc_auth.cfg not found.  Try reinstalling BOINC.
Only operations not requiring authorization will be allowed.
Operation failed: authentication error
```
Updating.
```
$ sudo apt install boinc-client
...
The following additional packages will be installed:
  libboinc7
Suggested packages:
  boinc-manager boinc-client-opencl boinc-client-nvidia-cuda
The following packages will be upgraded:
  boinc-client libboinc7
2 upgraded, 0 newly installed, 0 to remove and 132 not upgraded.
Need to get 959 kB of archives.
After this operation, 1,024 B disk space will be freed.
Do you want to continue? [Y/n]
Get:1 http://deb.debian.org/debian unstable/main amd64 boinc-client amd64 7.16.15+dfsg.is.7.16.11+dfsg-2 [514 kB]
Get:2 http://deb.debian.org/debian unstable/main amd64 libboinc7 amd64 7.16.15+dfsg.is.7.16.11+dfsg-2 [445 kB]
Fetched 959 kB in 0s (3,725 kB/s)
Preconfiguring packages ...
(Reading database ... 263055 files and directories currently installed.)
Preparing to unpack .../boinc-client_7.16.15+dfsg.is.7.16.11+dfsg-2_amd64.deb ...
Unpacking boinc-client (7.16.15+dfsg.is.7.16.11+dfsg-2) over (7.16.15+dfsg.is.7.16.11+dfsg-1) ...
Preparing to unpack .../libboinc7_7.16.15+dfsg.is.7.16.11+dfsg-2_amd64.deb ...
Unpacking libboinc7:amd64 (7.16.15+dfsg.is.7.16.11+dfsg-2) over (7.16.15+dfsg.is.7.16.11+dfsg-1) ...
Setting up libboinc7:amd64 (7.16.15+dfsg.is.7.16.11+dfsg-2) ...
Setting up boinc-client (7.16.15+dfsg.is.7.16.11+dfsg-2) ...
Installing new version of config file /etc/default/boinc-client ...
Processing triggers for systemd (246.6-2) ...
Processing triggers for man-db (2.9.3-2) ...
Processing triggers for libc-bin (2.31-4) ...
```
Trying again
```
$ boinccmd --project http://www.worldcommunitygrid.org/ suspend
Operation failed: authentication error
```
Some error messages disappeared but config is not working. 
```
$ groups|cut -f1 -d' '
boinc
$ sudo boinccmd --project http://www.worldcommunitygrid.org/ suspend
Operation failed: authentication error
$ cat /etc/boinc-client/config.properties
data_dir=/var/lib/boinc-client
$ ls -l /etc/boinc-client/config.properties
-rw-r--r-- 1 root boinc 31 Nov  9 23:38 /etc/boinc-client/config.properties
$ cat /var/lib/boinc-client/gui_rpc_auth.cfg

$
```
This cost me about an hour trying. stopping and starting the boinc demon. I eventually rebooted and then it just worked. Also boinctui is just fine now.

RichardHaselgrove:
Ironically, I've just seen the new PPA and installed it, so I should have a copy of your work. It was the only update since yesterday, so I simply stopped the client, let it install, and started the client - no reboot.

While the client was stopped I deleted gui_rpc_auth.cfg and client.properties - the two files we're testing.
Installing the PPA re-created gui_rpc_auth.cfg as a broken softlink to /etc/boinc-client - no target. I created the target as an empty text file, and the softlink worked again.

After starting the client, the Manager connected normally, with no workrounds. The empty gui_rpc_auth.cfg was accepted, as intended by this PR.

I then stopped the client, added my usual password in gui_rpc_auth.cfg, and started the client again. The Manager attached with the current workround(1). Now, it also reconnects **without** the workround, which I'm a little suspicious about - it may because  the Debian manager only allows 'Close Window', not a full exit.

I need to check that, and at some point I need to re-join the boinc group - when a re-boot is convenient. (I'm not currently listed as a member in the /etc/group file). But apart from the broken softlink, it looks good.

smoe:
The new version of OpenWrt will come with a BOINC package (https://github.com/openwrt/packages/tree/master/net/boinc/patches) - joint work with @N30dG. @TheAspens  - WCG works fine on these ARM machines, would be nice if you would support the platform arm-openwrt-linux-muslgnueabi . Some projects do not distributed static binaries and do not go nicel with the musl C library of OpenWrt, which is why we decided not to pose as a regular Linux.

To save the flash drive, the default Linux machine on that platform is on /opt/boinc. The routers plug in an exterrnal USB drive and everything happens on there, preferably no snapshots since running 24/7 anyway. Now, I really do not want to have anything on the internal flash memory - the fat routers have only 128MB or 256MB, the small ones just 4MB or 32MB - and would hence not install anything to /etc as you may guess. Since @AenBleidd outed himself as not much of a fan of distributions preparing patches, can we have the Linux default data directory as a configuration parameter, then? Or do we get an excempt for that?

AenBleidd:
> The new version of OpenWrt will come with a BOINC package (https://github.com/openwrt/packages/tree/master/net/boinc/patches) - joint work with @N30dG. @TheAspens - WCG works fine on these ARM machines, would be nice if you would support the platform arm-openwrt-linux-muslgnueabi . Some projects do not distributed static binaries and do not go nicel with the musl C library of OpenWrt, which is why we decided not to pose as a regular Linux.
> 
> To save the flash drive, the default Linux machine on that platform is on /opt/boinc. The routers plug in an exterrnal USB drive and everything happens on there, preferably no snapshots since running 24/7 anyway. Now, I really do not want to have anything on the internal flash memory - the fat routers have only 128MB or 256MB, the small ones just 4MB or 32MB - and would hence not install anything to /etc as you may guess. Since @AenBleidd outed himself as not much of a fan of distributions preparing patches, can we have the Linux default data directory as a configuration parameter, then? Or do we get an excempt for that?

Not sure all this applies to this thread. For me it's better to move OpenWrt topic to a separate ticket and don't mix things.

Yes, I'm not a fan of distributions patches because it's hard to support them. Whenever possible, I prefer to have this patches merged upstream. For this particular case, I see three patches and one the last one was proposed to upstream (at least I don't remember other two).

Regarding default data directory, it could be configured in the installer, and linux distributions have already these configuration IIRC.

The only thing I'm worry about is performance: I don't know how default installation will be configured but if BOINC will use too much resources, I'm afraid, this will lead to he situation when doesn't have enough CPU time to process incoming packages. So I'd first do a huge test under extreme load to adjust default settings.

RichardHaselgrove:
Guys, guys - the position of the data directory, and specifically whether it's static or configurable, is precisely why we're in this PR. The Manager has to know where the data is, even if that position is only determined at installation time. And the installer has to leave a record of that position for the Manager to use. This has been universal in the Mac and Windows versions of BOINC since the directories were split: Linux is playing (belated) catch-up.

Wearing my Systems Analyst hat, that's the end of the story. Coding it is an implementation detail. I'm personally agnostic about who does it. But because we have one BOINC (well, two with OpenWrt / WCG), and several dozen versions of Linux, it might seem better for BOINC to do it.

But whoever does it, has to take ownership for the whole design / coding / documentation / maintenance cycle. We are in this mess because #3709 was coded and released without a proper design phase.

RichardHaselgrove:
> it also reconnects without the workround, which I'm a little suspicious about

While I'm here, I've just done a full reboot (cold start) of today's testbed. The Manager found the password for the client without any workrounds. I don't know how it does it - we should probably document that somewhere - but it resolves my suspicions. We're good to go.

smoe:
> The only thing I'm worry about is performance: I don't know how default installation will be configured but if BOINC will use too much resources, I'm afraid, this will lead to he situation when doesn't have enough CPU time to process incoming packages. So I'd first do a huge test under extreme load to adjust default settings.

Newer routers come with a 256-1024 MB of RAM and 2-4 cores of which we suggest to leave one spare for what routers should be routing. We have done some testing on https://openwrt.org/toh/netgear/r7000 and https://openwrt.org/toh/linksys/linksys_wrt3200acm. Should be fine.

AenBleidd:
@RichardHaselgrove, I agree with you with one little 'but': this particular PR has nothing to deal with OpenWrt installation because I hardly believe that router will have BOINC Manager installed (at least if I understood all these messages correctly).

smoe:
> Regarding default data directory, it could be configured in the installer, and linux distributions have already these configuration IIRC.

Yes, this would be nice. I suggest to put an #ifndef around the LINUX_DEFAULT_DATA_DIR definition on https://github.com/BOINC/boinc/blob/a11bb44f131be94c55f78f6b71513722ed2d88ed/lib/common_defs.h#L380 so this could be a mere -D flag during compilation.

TheAspens:
> @davidpanderson When specifying the --dir on the command line then this has precedence over that static file, right?

short answer: yes
long answer: David added these comments that describes the precedent:
// Linux: look in:
//  - current dir
//  - a directory specified in /etc/boinc-client/config.properties
//  - /var/lib/boinc-client
// Note: the Manager (on all platforms) has a -datadir cmdline option.
// If present, it chdirs to that directory.

TheAspens:
I just retested the current version of it now and it looks pretty good.  

I see three open items in this thread.

1) If the current user doesn't have read access to the gui_rpc_auth.config and if they provided the password on the command like in a command then they get an unnecessary message as follows:
```
[knreed@oc4182665077 ~]$ boinccmd --passwd bd103a105b4d8313686178ad562fdee1 --project http://www.worldcommunitygrid.org update
Can't get RPC password: /var/lib/boinc/gui_rpc_auth.cfg exists but can't be read.  Check the file permissions.
Only operations not requiring authorization will be allowed.
```
The statement "Only operations not requiring authorization will be allowed." is not true if the password provided on the command line is valid.  @davidpanderson  - can we suppress this message if the password is passed in on the command line?

2) @LocutusOfBorg suggested the following change to the message output (it is prescriptive which is helpful):
```
--- boinc-7.16.15+dfsg.is.7.16.11+dfsg.orig/lib/gui_rpc_client.cpp
+++ boinc-7.16.15+dfsg.is.7.16.11+dfsg/lib/gui_rpc_client.cpp
@@ -415,7 +415,7 @@ int read_gui_rpc_password(char* buf, str
 #if defined(__linux__)
         if (errno == EACCES) {
             sprintf(msg_buf,
-                "%s exists but can't be read.  Check the file permissions.",
+                "%s exists but can't be read.  Check the file permissions, or ask the system administrator to add the user in the 'boinc' group.",
                 GUI_RPC_PASSWD_FILE
             );
             msg = msg_buf;
```
@davidpanderson - what do you think about changing the message as proposed above?

3. @smoe proposed using a command line option in the compiler to set the value of the constant "LINUX_DEFAULT_DATA_DIR" this was proposed as alternative to using the /etc/boinc-client/config.properties file to override the location of the data directory.  My opinion on this is that making it compilation dependent means that a version of the binary compiled by one person might not work as expected as a version compiled by a different person which would be confusing.  Additionally, from a support perspective it is easy to say "Check in /var/lib/boinc-client or in the directory specified in the file /etc/boinc-client/config.properties" vs having to ask them what platform they are on and then possibly having to gain access to that platform to figure out where the data directory is located.  Having said that I would appreciate it if @smoe , @LocutusOfBorg  and @Germano0  could agree and confirm in this thread how they want this handled.  Tx.

smoe:
>   @smoe proposed using a command line option in the compiler to set the value of the constant "LINUX_DEFAULT_DATA_DIR" this was proposed as alternative to using the /etc/boinc-client/config.properties file to override the location of the data directory.  My opinion on this is that making it compilation dependent means that a version of the binary compiled by one person might not work as expected as a version compiled by a different person which would be confusing.  Additionally, from a support perspective it is easy to say "Check in /var/lib/boinc-client or in the directory specified in the file /etc/boinc-client/config.properties" vs having to ask them what platform they are on and then possibly having to gain access to that platform to figure out where the data directory is located.  Having said that I would appreciate it if @smoe , @LocutusOfBorg  and @Germano0  could agree and confirm in this thread how they want this handled.  Tx.

If doubting the datadir's whereabouts I suggest to ask users to look in the logfile. Otherwise one is in trouble if the datadir was changed and the original location not removed. Debian installs to /var/lib/boinc-client and there is a symlink from /var/lib/boinc to /var/lib/boinc-client for inter-distributional compatibility or the folks who don't realize that the boinc-client from a System's perspective is a demon, and boinccmd and boincmgr are the real clients. As of today, you have no problem whatsoever to point the user to that directory. WIth that config.properties file - have fun.

For OpenWrt there is a need for the extra mount point and there is only one user, the admin. I did not follow your concern about needing to know about the distribution. You should always know about the distribution when there is a bug report. Today you know that /var/lib/boinc always works. But what the config.properties file is introducing now are infinitely many possible location for the data directory. And even with that file you still have no confidence about the directory that the boinc-client reads from since you do not know if the boinc-client was restarted since the file was last changed. And the permission to restart the machine may not be with the individual who had been added to the boinc group. I don't need to tell you that this is a recipe for desaster.

If the compilation flag is a problem, then I suggest that there is a verbose mode to the clients that inform about the directory that is looked in. So, to answer the question, I kindly ask for every distribution to have complete freedom to place that boinc data directory. The config.properties file shall be optional. And the tools shall tell where they are looking - the boinc-client in the logfile and boinccmd and boincmgr on stdout/the about box(?). I am not happy with that config.properties file, so I want this to be optional, which now seems to be the case. And I stopped being happy with a single user "boinc" but I need to think more about this.

The config.properties file I am confident is not the right way to address the prolem that you are having. It will make everything worse. The above suggested status messages may be sufficient and render config.properties obsolete.

RichardHaselgrove:
I note that the work we've done so far on this issue isn't supported by the most recent test build for #3715 provided by @LocutusOfBorg - that's fine, it would be unwise to try and test too many things at once. But I sense we're reaching the phase where we need to consolidate all the bug fixes for a full release.

Having said that, may I draw attention to two new issues affecting the same build? (!)

#4100 (bug - can't reopen Manager if Event Log visible at close)
#4104 (enhancement - relocate data directory, may impact this PR)

original-birdman:
> I see three open items in this thread.
> 
>     1. If the current user doesn't have read access to the gui_rpc_auth.config and if they provided the password on the command like in a command then they get an unnecessary message as follows:

It might just be me, but I'm unsure of the relevance of read access to this file by a user.

I use `boincmgr` to communicate with _remote_ systems, so being able to read the local file isn't meaningful, as it's a remote file that contain any password.

(And, FWIW, two of those remote systems are Asus routers running Entware  - I have my own BOINC build).

AenBleidd:
@original-birdman, if I understand correctly, all this thread is about Managing local client (that is running under 'boinc' account while Manager is running under users account and thus has no access to the file that could be read fine by the client).
When Managing remote system, Manager doesn't read local file with password because he has nothing to do with it

original-birdman:
> When Managing remote system, Manager doesn't read local file with password because he has nothing to do with it

It is, however, the same file and hence anything done to it's usage _might_ affect remote clients too, so that needs to be taken into account.

RichardHaselgrove:
When the Manager is accessing the client on the remote machine, it has no way of knowing the password in use by the client on that machine. The password, for security, should always be supplied by the user.

It is when the client and the manager are running on the **same** machine that read permissions come into play. The client reads the file on startup, and listens for this password in rpcs: if the manager can read **the same file on the same machine**, it can supply the password the client is expecting, and connect without that manual authentication step. That's how the Mac and Windows integrated systems work: I would like the Linux **local machine** installations to behave the same way.

original-birdman:
> It is when the client and the manager are running on the **same** machine that read permissions come into play.

That's just one specific case in what is a generic mechanism.

> The client reads the file on startup, and listens for this password in rpcs: if the manager can read **the same file on the same machine**, it can supply the password the client is expecting, and connect without that manual authentication step. That's how the Mac and Windows integrated systems work: I would like the Linux **local machine** installations to behave the same way.

There is no such thing as "local machine" in this set-up.  There is a client talking to a server, which may or may not be on the same system.
What you are suggesting seems to mean that when I talk to a _remote_ system I'll send a "local" password if I can read it.

RichardHaselgrove:
That wasn't the intention. With the default settings, the Manager will attempt to connect to localhost on launch: it is this scenario (yes, a special case) that I was hoping to automate with a local read.

If you wish the Manager to monitor a remote client, you have to supply the remote name or IP address. You can supply that, either on the command line, or after launch in the GUI. In either case, you should need to supply the remote password alongside the remote identifier. The local password should only be used to authenticate an implicit localhost connection.

smoe:
I have two machines on which I do not get boinctui or the boinc-manager to run because of authorization errors - invalid RPC password it tells me. My BOINC version is 7.16.17.

```
$ boinccmd --project_attach https://www.sidock.si/sidock  419_myweakaccountkey
Operation failed: authentication error
```
I tried reinstallations.

`boinccmd --get_state` does not give any warning. This is the second machine this happened to me. config.properties is set (also tried without). May I ask to be a bit more verbose in the error messages?

dchmelik:
Works on server/desktop Unix/GNU/Linux but not Android Linux; I blanked Android gui_rpc.auth.cfg then just couldn't connect from desktop... they're behind a firewall and going nowhere so it's okay.

PR #4077, opened 2020-10-31
Boinc android pid

efmer:
The PID name is incorrect and shouldn't include the directory.
The BOINC client MUST close down at this point.


AenBleidd (changes requested):
I'm afraid, this PR contains a lot of commits that are not related to this particular bugfix

PR #4092, opened 2020-11-11
[WIP] Manager and web: don't pass authenticator to account_finish.php

davidpanderson:
If you attach to a project using the Manager, as a "new user",
the Manager finishes the process by opening a browser window
to account_finish.php on that project,
which asks you for your name, country, and (optionally) postal code.
It passes the authenticator to this script;
this could be viewed as a security risk.

I changed things so that:
- the Manager doesn't pass the authenticator
- account_finish.php asks you to log in (with email/passwd).

Compatibility issues:
Old manager, new project: no problem.  User will see login form.
New manager, old project: user will see confusing "no such user" message

### DIFF
--- a/clientgui/WizardAttach.cpp
+++ b/clientgui/WizardAttach.cpp
@@ -613,7 +613,7 @@ void CWizardAttach::OnFinished( wxWizardExEvent& event ) {
     } else {
         // Attached to a project
         if (GetAccountCreatedSuccessfully() && GetAttachedToProjectSuccessfully()) {
-            wxLaunchDefaultBrowser(GetProjectURL() + wxT("account_finish.php?auth=") + GetProjectAuthenticator());
+            wxLaunchDefaultBrowser(GetProjectURL() + wxT("account_finish.php"));
         }
     }
 
--- a/html/user/account_finish.php
+++ b/html/user/account_finish.php
@@ -24,13 +24,7 @@ require_once('../inc/boinc_db.inc');
 require_once('../inc/util.inc');
 require_once('../inc/countries.inc');
 
-check_get_args(array("auth"));
-
-$auth = get_str("auth");
-$user = BoincUser::lookup_auth($auth);
-if (!$user) {
-    error_page("no such account");
-}
+$user = get_logged_in_user();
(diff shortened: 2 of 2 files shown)


davidpanderson:
I'm not sure this is worth doing.  If a hacker could intercept the authenticator, they can intercept the email/password, and use that to get the authenticator.

TheAspens:
> I'm not sure this is worth doing. If a hacker could intercept the authenticator, they can intercept the email/password, and use that to get the authenticator.

The risk is not the authenticator (or email/password) being intercepted.  The risk is that the PHP script treats the authenticator as a full authentication token (i.e. grants full access to the users account).  Thus if someone obtains someone else's authenticator, then they can use the account_finish.php to gain full access to the account.

davidpanderson:
The login script (login_action.php) lets you use the authenticator in place of password.  This serves two purposes: a) if a user has forgotten their password and their email addr no longer works, they can still log in; b) it lets project admins log in as a user, e.g. to debug problems.

AenBleidd:
@davidpanderson, @TheAspens
Hm, what about changing the logic of all this?
- leave old behavior as-is and implement new changes for new version of server software (I believe there is some versioning, if not - it's a good time to add it)
- use authenticator to finalize account creation but mix it with some salt and use it only once?
- remove an ability to login with authenticator only (generate some one-time recovery codes if user forgot his email/password
- remove an ability for project admins to login with authenticator as a users because it's a potential security vulnerability: no one except user can login as a user.

davidpanderson:
I couldn't reproduce this; "new user" attach on Windows worked OK.

davidbolvansky:
My steps:
1. Enter project URL: https://live.fitcrack.cz/fitcrack
2. Then:
image
3. Then Finish
image
4. Chrome opens with URL: https://live.fitcrack.cz/fitcrack/account_finish.php?auth=
image

Windows 8.1, Boinc 7.20.2

davidbolvansky:
> I couldn't reproduce this; "new user" attach on Windows worked OK.

Ah yeah, debugged it - this is related to special characters. Created: https://github.com/BOINC/boinc/issues/5024

PR #4097, opened 2020-11-14
Test Github Actions for BOINC

ChristianBeer:
Work in Progress.

@AenBleidd please comment and maybe look into the Windows related todos. I would like to keep development of this in my personal repo for now until it is ready. I gave you write access. You can use my cb_action_test branch.

ToDo List:

- fix OSX build dependencies (needs newer openssl and probably curl too)
- Windows: entangle vcpkg from solution and provide script to run before msbuild
- Windows: look into using the run-vcpkg action (https://github.com/marketplace/actions/run-vcpkg)
- All: add proper caching using https://github.com/marketplace/actions/cache
- All: search bintray action or modify scripts to deploay all artifacts to bintray
- All: add bintray secrets
- Linux: add coverity scan workflow (on schedule)
- All: add weekly schedule and deploy to bintray
- Linux: add workflow or job to run unit and integration tests

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -18,8 +18,13 @@ jobs:
           sudo apt-get -qq update
           sudo apt-get install -y freeglut3-dev libxmu-dev libxi-dev libfcgi-dev libxss-dev libnotify-dev libxcb-util0-dev libgtk2.0-dev libwebkitgtk-dev mingw-w64 binutils-mingw-w64-i686 binutils-mingw-w64-x86-64 gcc-mingw-w64 gcc-mingw-w64-i686 gcc-mingw-w64-x86-64 g++-mingw-w64 g++-mingw-w64-i686 g++-mingw-w64-x86-64 p7zip-full
 
+      - name: automake
+        if: ${{ success() }}
+        run: ./_autosetup
+
       - name: configure libs and make
-        run: cd $GITHUB_WORKSPACE && ./configure --disable-server --disable-client --disable-manager && make
+        if: ${{ success() }}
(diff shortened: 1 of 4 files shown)


CharlieFenton:
@ChristianBeer Please ensure that you are working from the current MacOS dependency build scripts in GIT master, which I have recently updated so they can build Universal2 libraries containing both Intel and Apple Silicon code. Apple is now shipping Apple Silicon Macs.

These scripts automatically determine whether the version of Xcode being used supports building both Intel and Apple Silicon binaries. They build only Intel libraries on older versions of Xcode and Universal2 libraries when supported. I believe Xcode 12.2 or later is required to build Universal2 binaries on Intel Macs, and requires macOS Catalina 10.15.4 or later.

Please note also that if you change the versions of the dependencies used by BOINC on the Mac, you must also change the paths in the relevant build settings in the Xcode project. Please be sure to also change the BOINC version number when changing dependency versions. I have very intentionally hard-coded these to guarantee that any version of BOINC must always be built with the same versions of the dependencies, ensuring a high degree of reproducibility.

ChristianBeer:
@CharlieFenton I'm aware of the implications when changing the Mac depenencies. GitHub Actions currently offers MacOS 10.15 with several xcode versions. The first problem I ran into was that openssl wouldn't compile because it tried to use a deprecated and removed perl module which was fixed in a later version of openssl. Now curl complains about something but I didn't investigate further. I'll keep this in my personal fork until the problems are sorted out and the changes needed to build scripts and dependencies are well known.

I'm going to lock this PR as it was created by accident. Please post follow up discussions here: https://github.com/ChristianBeer/boinc/pull/1

Issue #2600, opened 2018-07-12
curl/curl.h


ritua2:
The "curl.h" file is being included in lib/remote_submit.cpp. However, we do not see this file in the "curl" directory. Is this file not required anymore?

Thanks,
Ritu

JuhaSointusalo:
`curl.h` comes from libcurl development files. On debians the package you want is `libcurl4-openssl-dev` and on redhats `curl-devel`.

JuhaSointusalo:
To avoid any potential misunderstandings. You don't need to copy `curl.h` to BOINC source tree. `configure` should find it after you have installed the development package (which means you need to re-run `configure` after installing the package).

`curl` directory's purpose is to hold the certificate bundle that's used in Windows and Mac clients.

ritua2:
Thanks a lot Juha! That is very helpful!


Best Regards,

Ritu


________________________________

Uplinger:
This issue appears to be closed, why is it still open?

Thanks,
-Keith

PR #4106, opened 2020-11-21
Add check for libcurl when building for server

### LINKED ISSUE #3501: Make blows up with missing curl/curl.h
**Describe the bug**
While compiling server with `./configure --disable-client --disable-manager --enable-fcgi && make` on a system that does not have libcurl installed, configure does not catch the missing dependency and the build blows up.

```
  CXX      remote_submit.o
../lib/remote_submit.cpp:24:10: fatal error: curl/curl.h: No such file or directory
 #include <curl/curl.h>
          ^~~~~~~~~~~~~
compilation terminated.
make[2]: *** [Makefile:820: remote_submit.o] Error 1
```

**Steps To Reproduce**
```
git clone https://github.com/BOINC/boinc.git boinc-git
cd boinc-git
git checkout tags/server_release/1.2/1.2.1 -b server_release/1.2.1
./_autosetup
./configure --disable-client --disable-manager --enable-fcgi
make
```

**Expected behavior**
Configure should notify about the missing lib and abort.

truboxl:
Fixes #3501 and #2600

**Description of the Change**
Add check for libcurl when building for server
Block early rather than fail after a long time of compile when libcurl is not supplied

`lib/remote_submit.cpp` needs `curl/curl.h`
https://github.com/BOINC/boinc/blob/faf272ff362cd9fc22bba5d2cbd4f57ff374364c/lib/remote_submit.cpp#L24

`lib/remote_submit.cpp` is built when building in `tools` folder
https://github.com/BOINC/boinc/blob/faf272ff362cd9fc22bba5d2cbd4f57ff374364c/tools/Makefile.am#L61

which is only built when ENABLE_SERVER
https://github.com/BOINC/boinc/blob/faf272ff362cd9fc22bba5d2cbd4f57ff374364c/Makefile.am#L25-L28

ENABLE_SERVER is checked here
https://github.com/BOINC/boinc/blob/faf272ff362cd9fc22bba5d2cbd4f57ff374364c/configure.ac#L1065

**Alternate Designs**
N/A

**Release Notes**
N/A

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -451,7 +451,7 @@ AS_IF([test "x${enable_client}" = "xyes" -o "x${enable_server}" = "xyes"],
 dnl ---------- libcurl (m4/libcurl.m4) ------------------------------
 dnl curl is needed for client
 
-AS_IF([test "x${enable_client}" = "xyes"],
+AS_IF([test "x${enable_client}" = "xyes" -o "x${enable_server}" = "xyes"],
   [
     LIBCURL_CHECK_CONFIG([yes], [7.17.1], [haveCurl=yes], [haveCurl=no])
 


AenBleidd:
@truboxl, lgtm. Thanks

PR #4113, opened 2020-11-26
client: show IPV6 address correctly

### LINKED ISSUE #3340: Wrong cast when reporting non-allowed rpc access
For some time, reports from new installations of Linux show non-allowed access from ip addresses starting with 2.  2.x.x.x is typically a French address.
reference issue #3246.  I could not find an issue like I put together here but I might have missed it.

In actuality, the 2 is the identifier AF_INET due to wrong cast.
In program gui_rpc_server.cpp function "show_connect_error" the following lines of code:
```
#ifdef _WIN32
    sockaddr_in* sin = (sockaddr_in*)&s;
    safe_strcpy(buf, inet_ntoa(sin->sin_addr));
#else
    inet_ntop(s.ss_family, &s, buf, 256);
#endif
```
The reference s.ss_family causes the "2" to be picked up.  The actual ip address starts at s.ss_data which is not defined AFAICT

I ran a test from 192.168.1.241 and attempted to access the client.  The remote_host.cfg file has to be empty as that causes the problem somehow.
=====================bug================
18-Oct-2019 14:03:49 [---] Setting up GUI RPC socket
18-Oct-2019 14:03:49 [---] Checking presence of 0 project files
18-Oct-2019 14:03:49 [---] This computer is not attached to any projects
18-Oct-2019 14:03:49 Initialization completed
18-Oct-2019 14:04:14 [---] GUI RPC request from non-allowed address 2.0.233.131
(note that the above can change but the 2.0 will not as that is the value AF_INET)
If you dump out the hex values of the buffer "buf" one gets the following

18-Oct-2019 15:36:37 [---] 
debug2 02 00 e9 83 c0 a8 01 f1 00 00 00 00 00 00 00 00
----------------^^ this is 233
------------------^^  this is 131
the 2.0.233.131 came from the buffer "buf" 

however, the computer that triggered the access was 192.168.1.241 and that value is also in that buffer but not at the location the coder expected 
debug2 02 00 e9 83 c0 a8 01 f1 00 00 00 00 00 00 00 00
----------------------^^ this is 192
--------------------------^^ this is 168
----------------------------^^ this is the 1
-------------------------------^^ this is the 241

A "fix" is to use the same code as the win32.  I took a guess and changed
```
#ifdef _WIN32
    sockaddr_in* sin = (sockaddr_in*)&s;
    safe_strcpy(buf, inet_ntoa(sin->sin_addr));
#else
    inet_ntop(s.ss_family, &s, buf, 256);
#endif
```
to 

```
#ifdef _WIN32
    sockaddr_in* sin = (sockaddr_in*)&s;
    safe_strcpy(buf, inet_ntoa(sin->sin_addr));
#else
    sockaddr_in* sin = (sockaddr_in*)&s;
    safe_strcpy(buf, inet_ntoa(sin->sin_addr));
#endif
```

the above ran correctly
File tempfix.png shows the correct ip address and the hex dump of the buffer
file tempfix1.png shows the code that dumped out the buffer

temp_fix

temp_fix1

[EDIT]  Not necessary for remote_access.cfg to be actually empty or null for the bug to show up.

davidpanderson:
Fixes #3340

This is the same as #3342 but with an additional check for unknown address type

### DIFF
--- a/client/gui_rpc_server.cpp
+++ b/client/gui_rpc_server.cpp
@@ -391,7 +391,15 @@ static void show_connect_error(sockaddr_storage& s) {
     sockaddr_in* sin = (sockaddr_in*)&s;
     safe_strcpy(buf, inet_ntoa(sin->sin_addr));
 #else
-    inet_ntop(s.ss_family, &s, buf, 256);
+    if (s.ss_family == AF_INET) {
+        sockaddr_in* sin = (sockaddr_in*)&s;
+        inet_ntop(AF_INET, (void*)(&sin->sin_addr), buf, 256);
+    } else if (s.ss_family == AF_INET6) {
+        sockaddr_in6* sin = (sockaddr_in6*)&s;
+        inet_ntop(AF_INET6, (void*)(&sin->sin6_addr), buf, 256);
+    } else {
+        sprintf(buf, "Unknown address family %d", s.ss_family);
+    }
 #endif
     msg_printf(NULL, MSG_INFO,
         "GUI RPC request from non-allowed address %s",


rlpm:
Thanks for fixing this!

PR #4118, opened 2020-12-01
client: default to just checking first X server

BryanQuigley:
**Description of the Change**
default to just checking first X server idle usage
You can go back to checking all using the already existing
check_all_logins boinc option. This gives us a nicer default
but still lets anyone who depends on the old functionality to
get it.

Also removed a related comment that is no longer true from a previous commit.

Resolves: #2256 

**Alternate Designs**
There have been a few.. but the most recent before trying this was to get the total number of users from utmp and then only check that number of X servers.  That led me to discover that the utmp bit I wanted to extend never ran - because of check_all_logins being false - which led me to this solution.

**Release Notes**
BOINC users who want idle detection against more than 1 XServer must specify --check_all_logins to do so now.

### DIFF
--- a/client/cs_cmdline.cpp
+++ b/client/cs_cmdline.cpp
@@ -50,7 +50,7 @@ static void print_options(char* prog) {
         "    --allow_remote_gui_rpc         allow remote GUI RPC connections\n"
         "    --allow_multiple_clients       allow >1 instances per host\n"
         "    --attach_project <URL> <key>   attach to a project\n"
-        "    --check_all_logins             for idle detection, check remote logins too\n"
+        "    --check_all_logins             for idle detection, check remote and more than 1 local login too\n"
         "    --daemon                       run as daemon (Unix)\n"
         "    --detach_console               detach from console (Windows)\n"
         "    --detach_project <URL>         detach from a project\n"
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1840,11 +1840,7 @@ const vector<string> X_display_values_initialize() {
     // According to "man Xserver", each local Xserver will have a socket file
     // at /tmp/.X11-unix/Xn, where "n" is the display number (0, 1, 2, etc).
     // We will parse this directory for currently open Xservers and attempt
-    // to ultimately query them for their idle time. If we can't open this
-    // directory, or the display_values vector is otherwise empty, then a
-    // static list of guesses for open display servers is utilized instead
-    // (DISPLAY values ":{0..6}") that will attempt connections to the first
-    // seven open Xservers.
+    // to ultimately query them for their idle time. 
     //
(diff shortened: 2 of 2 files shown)


BryanQuigley:
One previous failed attempt to fix this was #3153.

BryanQuigley:
@SETIguy does this cover your concerns from #3153 ?

ChristianBeer:
Close and reopen to trigger CI

ChristianBeer:
@RichardHaselgrove 
In order to get the build artifacts you need to click on the "Checks" Tab above and then on the white space just right of "Linux". This is actually a bit tricky, as I just found out. When clicking on the name of a workflow you expand the list of the jobs within this workflow. But the list of artifacts is only visible when clicking on the white space. This page will always have the artifacts no matter who created the PR.
There is an alternative page that also stores build artifacts: https://bintray.com/beta/#/boinc/boinc-ci/pull-requests?tab=overview but this page only contains pull requests that originate from within the BOINC/boinc repository (as is usually done by regular Committers).

I'm planing on writing an email to the dev and alpha list about this soonish(TM).

RichardHaselgrove:
@ChristianBeer 
It actually worked quite easily for me. Using Linux Firefox: clicking on Linux, on the exclamation bubble, or anywhere in between, all opened https://github.com/BOINC/boinc/actions/runs/484893675 and displayed the artifacts at the bottom of the page.

I now have the 7.17.0 pre-release client running, and the matching 'manager-without-webview' active (nice to see the full menus back: version with webview wouldn't run on this system).

But the Event Log reports
```
Thu 14 Jan 2021 10:40:41 GMT |  | [idle_detection] Error (2) opening /tmp/.X11-unix/.
Thu 14 Jan 2021 10:40:41 GMT |  | [idle_detection] Could not connect to any DISPLAYs. XSS idle determination impossible.
```
so I don't think this fix has worked as intended.

BryanQuigley:
@RichardHaselgrove 

That looks unrelated to this change.  Do you have a directory  /tmp/.X11-unix/ and PrivateTmp=true disabled?

Unfortunately in my most recent tests I'm failing to get X11 detection to work at all.  I'm running Debian unstable and I'm not sure what's changed.

RichardHaselgrove:
@BryanQuigley 

1) I do have /tmp/.X11-unix/, and it contains the single file 'X0'.
2) I have PrivateTmp=true **enabled** via an override file, because of the conflict described in #4125

I must confess I was short of time when I last commented, and didn't check the contents of your commit. From the title of this PR I'd assumed (my bad) that you had hard-coded an X0 fallback test, to replace the X0-X7 tests removed in #3549.

BryanQuigley:
I'm dropping this - I couldn't get X detection to work as I expect enough to make this patch make sense.  (And I'm just building without xss for more personal use).

PR #4120, opened 2020-12-06
Test base64 library

hectorespert:
Fixes #

**Description of the Change**
Add test class for base64 library using Google test suite.

**Alternate Designs**

**Release Notes**
N/A


ChristianBeer:
Was this test triggered by a bug in the base64 function? Would it be poosible for you to add more testcases. Using strings with spaces, umlauts (äöüß), diacritics and newlines would make the test more broad and eventually cover the whole file. I'm not sure it would be good to test the failure cases as those seem to throw exceptions whereas the rest of BOINC doesn't use exceptions.

AenBleidd:
It would be also nice to add this test to VS2019 project  https://github.com/BOINC/boinc/blob/master/win_build/unittests_vs2019.vcxproj

hectorespert:
> It would be also nice to add this test to VS2019 project https://github.com/BOINC/boinc/blob/master/win_build/unittests_vs2019.vcxproj

Added

hectorespert:
> Was this test triggered by a bug in the base64 function? 

No, it was only to add code coverage in these funtions.

> Would it be possible for you to add more testcases. Using strings with spaces, umlauts (äöüß), diacritics and newlines would make the test more broad and eventually cover the whole file. I'm not sure it would be good to test the failure cases as those seem to throw exceptions whereas the rest of BOINC doesn't use exceptions.

I added more testcases to cover your examples.

ChristianBeer:
The changes are fine with me. Thanks for the contribution. Do you plan on adding more test cases in the near future?

hectorespert:
> Do you plan on adding more test cases in the near future?

Not for base64 library. But I have a some tests prepared to test utils library.

PR #4121, opened 2020-12-07
Determine gtkver for wxWidgets more correctly

makise-homura:
Fixes #

**Description of the Change**
On several platforms (e.g. [Elbrus]()) `wx-config --selected-config` produces a value that don't start with `gtk*`. For example, it may look like `e2k-mcst-linux-gnu-gtk2-unicode-3.0`. So I've expanded detection procedure a bit to allow these configurations to be treated correctly, e.g. `boinc-wxwidgets.m4` should react properly not just to `gtk-*`, `gtk2-*`, and `gtk3-*`, but also to `*-gtk-*`, `*-gtk2-*`, and `*-gtk3-*`.

**Alternate Designs**
I have no idea how to do it better, it seems that expanding the check in `boinc-wxwidgets.m4` might be the only reasonable option. No `configure` parameters can solve this problem, as far as I guessed.

**Release Notes**
Determine GTK version for wxWidgets more correctly


ChristianBeer:
Thanks for this enhancement.

PR #4126, opened 2020-12-12
hostinfo_unix: Ignore tty(S|ACM) devices in TTY idle time calculation

Flowdalic:
The atime (access time) of device nodes whose name starts with
/dev/ttyS (serial port) or /dev/ttyACM (serial USB) should not be used
for TTY idle time calculation. It is perfectly possible that they are
used without any user being active.

Previously BOINC would not run on some of our systems, because those systems also had a monitoring device connected via a serial port. Since this monitoring device is periodically checked, BOINC would assume that a user is active. I think it is sensible to exclude `ttyS` and `ttyACM` (and probably also `ttyUSB` devices. But those I couldn't test as we do not have one of those) from TTY idle time calculation.

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1567,9 +1567,19 @@ inline long device_idle_time(const char *device) {
 static const struct dir_tty_dev {
     const char *dir;
     const char *dev;
+    const vector<string> ignore_list;
+
+    bool should_ignore(const string &devname) const {
+        for (const string &ignore : ignore_list) {
+            if (devname.rfind(ignore, 0) == 0) return true;
+        }
+        return false;
+    }
 } tty_patterns[] = {
 #ifdef unix
-    { "/dev", "tty" },
+    { "/dev", "tty",
+      {"ttyS", "ttyACM"},
+    },
     { "/dev", "pty" },
     { "/dev/pts", NULL },
 #endif
@@ -1596,6 +1606,11 @@ vector<string> get_tty_list() {
             //
             if (tty_patterns[i].dev) {
                 if ((strstr(devname, tty_patterns[i].dev) != devname)) continue;
+
+                // Ignore some devices. This could be, for example,
+                // ttyS* (serial port) or devACM* (serial USB) devices
+                // which may be used even without a user being active.
+                if (tty_patterns[i].should_ignore(devname)) continue;
             }
 
             sprintf(fullname, "%s/%s", tty_patterns[i].dir, devname);


AenBleidd:
@Flowdalic, please remove e37ec47 because we're not support c++17 for all platforms and cfba160 as not related to this PR

Flowdalic:
> @Flowdalic, please remove [e37ec47](https://github.com/BOINC/boinc/commit/e37ec474e34e4658e1e289e1e640be644f898ba3) because we're not support c++17 for all platforms

I did so, but I am still battling issues caused by the build not even supporting C++11. May I ask which platforms prevent the upgrade to C++17 (or C++11)? At least C++11 should be widely available by now and makes the code a lot easier to read and write. I could live without C++17, but C++11 would be really nice to have.

AenBleidd:
@Flowdalic, currently Windows build is the last that is not support C++11 or later. We have planned move to c++11 (or later) but since it's not decided yet, we should still not use any C++11 features. It shouldn't be a big deal to not use any c++11 features in the code because your change isn't big

Flowdalic:
> It shouldn't be a big deal to not use any c++11 features in the code because your change isn't big

Sure, if you are experienced with pre-C++11 coding. But right now I have to find out how to replace my modern idioms with pre-C++11 ones. And additionally, it appears I can only become aware of used C++11 idioms in my code via BOINC's CI, as the build may succeeds locally but fails in the CI because of some CI jobs not supporting C++11. I am not sure, but this, in turn, is probably because BOINC's build system does not specify a C++ standard version (e.g. `--std=c++03`).

I do not know how much work it would be to switch BOINC to C++11, but I would suggest that you consider it, because frankly, as "drive-by spare-time contributor" I am not really motivated to code pre-C++11.

AenBleidd:
> I do not know how much work it would be to switch BOINC to C++11, but I would suggest that you consider it, because frankly, as "drive-by spare-time contributor" I am not really motivated to code pre-C++11.

And this definitely will be done soon

ChristianBeer:
Close and reopen to trigger Github Actions build

ChristianBeer:
Close and reopen to trigger Github Actions build and creation of artifacts

Flowdalic:
I have force pushed this PR back to https://github.com/BOINC/boinc/commit/8a8088ac93f231a625be77724c6517246eb8db0c, i.e. the initial version of this PR. This commit uses a "fancy" for-each loop to iterate over a `std::vector`, which was previously not possible to use in BOINC, since BOINC used an older C++ standard where for-each loops where not possible.

AenBleidd:
@BOINC/committer, please don't merge it right now

client/hostinfo_unix.cpp
```diff
@@ -1567,9 +1567,19 @@ inline long device_idle_time(const char *device) {
 static const struct dir_tty_dev {
     const char *dir;
     const char *dev;
+    const vector<string> ignore_list;
+
+    bool should_ignore(std::string devname) const {
```
AenBleidd, line 1572: `std::string devname` -> `const std::string &devname`

client/hostinfo_unix.cpp
```diff
@@ -1567,9 +1567,19 @@ inline long device_idle_time(const char *device) {
...
+        for (std::string ignore : ignore_list) {
```
AenBleidd, line 1573: `std::string ignore` -> `const std::string &ignore`

AenBleidd (changes requested):

client/hostinfo_unix.cpp
```diff
@@ -1567,9 +1567,19 @@ inline long device_idle_time(const char *device) {
 static const struct dir_tty_dev {
     const char *dir;
     const char *dev;
+    const vector<string> ignore_list;
+
+    bool should_ignore(const std::string &devname) const {
+        for (auto ignore : ignore_list) {
```
AenBleidd, line 1573: Yeah, I also like `auto` but please set here explicitly `const std::string &ignore`

client/hostinfo_unix.cpp
```diff
@@ -1567,9 +1567,19 @@ inline long device_idle_time(const char *device) {
 static const struct dir_tty_dev {
     const char *dir;
     const char *dev;
+    const vector<string> ignore_list;
+
+    bool should_ignore(const std::string &devname) const {
```
AenBleidd, line 1572: Sorry, my bad: `std::` prefix could be removed here since we're defining `using std::string;` above

AenBleidd (changes requested):

client/hostinfo_unix.cpp
Flowdalic: Out of curiosity (and cause I would prefer auto here): Why do you want to be explicit in this case about the type?

AenBleidd: There are couple of reasons:
1 - The code in our codebase is pretty old and written in one style so it's easier to read the code when even new code uses the same code style as the old one. In this particular case it's not a complicated template of something like this that is easier to write using `auto` and, by the way, even with `auto` you're missing `const` and `&`
2 - It's easier to do a code review in browser because you don't need to look what is the type is hidden behind `auto`

AenBleidd (approved):

AenBleidd:

RichardHaselgrove:
@AenBleidd - not quite sure what you want me to test? I've installed 
`linux_client_ed70369c091923addf5a8d64bed92c0417e91e07`
and set PrivateTmp=true again (the conflicting project has no work at the moment). Idle detection works as intended.

This machine has mouse and keyboard connected via a single USB connection (KVM switch acts as USB hub). Idle time debug log counts upwards as expected, and restarts computation after I've kept my hand off the mouse for three minutes. I can reset the settings to run always over a remote RPC from upstairs, without interrupting work again.

But I don't have any other TTY or emulated-TTY devices connected, so I can't tell if they're ignored now.

AenBleidd:
> But I don't have any other TTY or emulated-TTY devices connected, so I can't tell if they're ignored now.

At least normal USB mouse is not ignored. That's enough as for me

PR #4129, opened 2020-12-16
Specify that the build is done with C++03

Flowdalic:


AenBleidd:
Why this is necessary?

Flowdalic:
Motivation is https://github.com/BOINC/boinc/pull/4126#issuecomment-746855877

AenBleidd:
Without this instruction it's possible to build with newer language standard if it's possible. You could add this to your branch temporarily to test your code but I can't accept this PR.

AenBleidd:
> But isn't that pointless given that not all platforms of BOINC support a newer standard?

BOINC is build for 4 platforms: Windows, Linux, Android and MacOS. Only Windows builds currently don't support c++11.
Linux release builds are done separately by distro maintainers and we don't want to force them to build all code with c++03.
For Android and MacOS situation is different but also these builds shouldn't be forced to use c++03 for building.
Because we don't want to block some compiler specific optimizations that are available in newer language standards if this doesn't affect Windows builds.
And as I told you before, we plan to move to C++11 (or even C++14 or C++17) soon so there is no real reason to put this limitation even temporary. For such purposes we have our CI that can tell us when some code can't be built for Windows.

Flowdalic:
Thanks I get it now.

> Because we don't want to block some compiler specific optimizations that are available in newer language standards if this doesn't affect Windows builds.

To be honest, I think the advantages of a fixed programming language version outweigh the potential advantages due to compiler optimizations (leave alone that I wonder how many compiler optimizations can only be applied if build with a newer version). I also think that this part of BOINC is probably not really performance critical: It is the BOINC apps that you want to optimize to the fullest. Furthermore, downstream distributions like Debian could easily patch the script.

AenBleidd:
@Flowdalic, I agree with you. But since we will anyway upgrade used C++ standard soon - I see to reason to pin it now.

PR #4130, opened 2020-12-23
[CI] replace Travis and AppVeyor with Github Actions

ChristianBeer:
**Description of the Change**
This consolidates the different CI builds on one platform that provides support for the major three operating systems we build on (Linux, Windows, MacOS). The nomenclature and syntax is a bit different but it has the same functionality.
This commit introduces the Linux and Windows based workflows. All the build artifacts can now be downloaded from [bintray](https://bintray.com/beta/#/boinc/boinc-ci?tab=files). The CI workflows run for every pull request, every merge into the master branch and every Sunday around 12:00 (GMT).
Every first Sunday in a month a cleanup Script is run to delete old artifacts from Bintray. Artifacts from closed/merged pull requests get deleted then. Artifacts from weekly builds and merges into the master branch older than 6 months get deleted too.

Fixes #4123

Co-authored-by: @AenBleidd

**Alternate Designs**
An alternative would have been to apply for a limited amount of credits every month to run Travis-CI. 

**Release Notes**
N/A

PR #4131, opened 2020-12-23
[CI] add OSX workflow

ChristianBeer:
**Description of the Change**
Replaces the previous Travis-CI job for OSX. The MacOS runner on Github Actions has a newer perl version installed which forced me to increase version numbers of dependencies in order to get it working.

@CharlieFenton please review and comment. Do not merge before #4130! Please provide a list of files that are useful for testers so we can add them to the artifacts available on bintray.

Co-authored-by: @AenBleidd

Todo after merge: Add the OSX Badge to Readme and finish deployment for OSX.

**Alternate Designs**
An alternative would have been to apply for a limited amount of credits every month to run Travis-CI.

**Release Notes**
N/A


CharlieFenton:
This will fail unless I change several other files. I am very busy now; I will try to work on this within a few days.

AenBleidd:
@CharlieFenton, why then it didn't fail on CI? What is missed there? Are you talking about runtime instead? Maybe we can help you somehow and change these files for you?

CharlieFenton:
> why then it didn't fail on CI?

Probably because he did not rebuild the dependencies cache, so the cache still has the older version of OpenSSL. I'll try to create a separate PR tonight for the changes necessary to use the new version of OpenSSL.

CharlieFenton:
I just noticed that he is also changing the curl version. I'll see if I can commit my changes to this PR rather than creating a separate one.

AenBleidd:
All dependencies were build. Because all them are build on a new CI system. And yes, you can add a commit here

AenBleidd:
In case of any issues - please send me a patch and I'll apply it to this branch since I have permissions in Christians repo

AenBleidd:
@CharlieFenton, please take a look at the log here: https://github.com/BOINC/boinc/pull/4131/checks?check_run_id=1600283439
For me everything was build with ok and maybe no changes needed?

ChristianBeer:
Building the OSX manager with the CI script behaves a little bit different than building "by hand". This is because the dependency versions are hardcoded in the pbxproj file (look for `SEARCH_PATHS`). In the CI script, I'm overriding those variables and provide the correct path without version numbers.

CharlieFenton:
>  In the CI script, I'm overriding those variables and provide the correct path without version numbers.

@ChristianBeer I ask that you NOT do that. It is very important that:
 * All builds for a given version of BOINC on the Mac use the same versions of the dependencies
 *  Travis-CI fail so it alerts us if there is a mismatch between the dependency versions and the Xcode project
 * The build scripts must be updated to match the version of the dependencies because they contain patches specific to the expected source files.
 
I've updated the comments in dependencyNames.sh to emphasize these points. And I have updated the Xcode project and the build scripts and Xcode prokject.

CharlieFenton:
It was a good idea to update curl and OpenSSL anyway. This is a good time to do that. 

While it might be good to also update WxWidgets, that will be a much bigger job. New versions of WxWidgets have pretty much always broken things that require new patches to WxWidgets or workarounds to BOINC source files.

ChristianBeer:
> > In the CI script, I'm overriding those variables and provide the correct path without version numbers.
> 
> @ChristianBeer I ask that you NOT do that. It is very important that:
> 

We both agreed to this when we added the CI script for Travis some time ago. The dependency scripts already take care of the correct versions for each BOINC version. Putting the dependencies outside the build directory (as is done in the xcode project) is not advisable for the CI run and we also agreed on the current way to override the paths for the CI build but keep the hardcoded paths for the manual build.

As to what to put on bintray: is it correct that all files are copied into `mac_build/build/Deployment`? Can I just archive the content of that directory and upload it to bintray?

CharlieFenton:
@ChristianBeer Sorry, I am now quite confused. Perhaps I don't understand what you meant when you wrote:
> In the CI script, I'm overriding those variables and provide the correct path without version numbers.

My understanding is that when we change the version of a dependency we use, we must tell Travis-CI to rebuild the dependency cache by adding `--clean` to _dependencyNames.sh_ , which is what I did. (Since the cache has been rebuilt, I've now reversed that.) If you are overriding the dependency version number, why is rebuilding the cache necessary whenever we update version numbers in _dependencyNames.sh_? This is what you told me to do in an email on 22 January 2019.

Since you have changed the versions of curl and openSSL in _dependencyNames.sh_, I made the corresponding changes to the build scripts _buildcurl.sh_ and _buildopenssl.sh_ so that the patches performed by those scripts would work correctly, and I updated the Xcode project and the build instructions to match. The comments in _dependencyNames.sh_ have always explained the need for that; I've now modified those comments to more clearly emphasize them.

Whether or not my changes are needed for Travis-CI to work, they are all necessary so that, when this PR is merged, BOINC for the Mac can be built on a Mac. 

In any case, I've stayed up until 5:30am trying to get this working so I would not delay you, because you asked for my review. I'm sorry if it is not what you wanted, but I'm going to bed now.

CharlieFenton:
> As to what to put on bintray: is it correct that all files are copied into mac_build/build/Deployment? Can I just archive the content of that directory and upload it to bintray?

Sorry, I do not know what _bintray_ is. When BOINC is built using the script _BuildMacBOINC.sh_,  xcodebuild writes all build products into `mac_build/build/Deployment`. When built using the Xcode.app user interface, they are built elsewhere but the BOINC libraries (_libboinc.a_, etc.) are copied into that directory. The full path to the built products is always written to the text file `mac_build/Build_Deployment_Dir` (when the client or Manager is built.)

The build products for the dependent libraries (OpenSSl, curl, etc.) are put wherever their authors decided.

ChristianBeer:
> In any case, I've stayed up until 5:30am trying to get this working so I would not delay you, because you asked for my review. I'm sorry if it is not what you wanted, but I'm going to bed now.

I apologize if I was unclear of the timeframe for this PR. I intentionally separated this from the other workflows in order to give you more time to make the neccessary changes. There was no need to rush but thanks for making the changes.

> Sorry, I do not know what _bintray_ is. When BOINC is built using the script _BuildMacBOINC.sh_, xcodebuild writes all build products into `mac_build/build/Deployment`. When built using the Xcode.app user interface, they are built elsewhere but the BOINC libraries (_libboinc.a_, etc.) are copied into that directory. The full path to the built products is always written to the text file `mac_build/Build_Deployment_Dir` (when the client or Manager is built.)

Thanks. That's the information I needed. For the Linux and Windows builds we already gathered the build products and uploaded them to the external service [bintray](https://bintray.com/beta/#/boinc/boinc-ci?tab=packages). People who want to test the changes made in a PR can download those archives and run the programs on their machines.

CharlieFenton:
> People who want to test the changes made in a PR can download those archives and run the programs on their machines.

Thank you for the explanation.

Unfortunately the Mac binaries as generated by the compiler and linker are not terribly useful. It is very difficult to use the Manager, Client or Screensaver in this form. There is another major step, assembling the components and creating the installer by running the script _mac_installer/release_boinc.sh_. 

It is possible to use the components boinc (the Client), switcher, setprojectgrp and boinccmd as provided in our `Unix command-line version` releases on https://boinc.berkeley.edu/download_all.php, and setting the ownership using the script _mac_build/Mac_SA_Secure.sh_. But this is difficult because MacOS blocks execution of files downloaded from the Internet unless they have been properly signed by a licensed Apple Developer and directly approved by Apple in a process Apple calls _Notarization_.

CharlieFenton (approved):
Thanks. Looks good to me now.

CharlieFenton:
One nice thing about the GitHub-Actions build is that it uses _Xcode 12.2_, so it builds Universal2 binaries containing code for both Intel x86_64 and Apple Silicon arm64 CPUs. Travis-CI uses an older version of Xcode and so builds only Intel binaries.

ChristianBeer:
I just looked into signing and notarizing the binaries using Github Actions. And although it is in general possible it's not feasible for BOINC right now. We would need a Mac Developer to provide his/her signing key and notarization credentials that get stored as secrets and would be used to sign and notarize the app bundle for each build automatically.

PR #4139, opened 2021-01-07
client: only report a graphics app if file is there can we can execute it

davidpanderson:
Problem: when parsing an APP_VERSIOn (e.g. in scheduler reply)
we check whether it has a graphics app (i.e. a file named "graphics_app").
If so we could report it to the GUI before the file has been downloaded,
causing an error when "Shot Graphics" is clicked.

Also, on Mac we check whether the graphics app can run on
this CPU, given its header.
But this check was being done during APP_VERSION parsing,
when typically the file doesn't exist yet.

Fix: when parsing APP_VERSION,
store a pointer to the FILE_INFO for the graphics app.
When reporting an ACTIVE_TASK to the GUI,
check whether the file is present and (on Mac) can be executed.
Only then tell the GUI about it.
If the file download failed or the file can't be executed,
zero out the pointer so we don't do the check again.


CharlieFenton:
The code in can_run_on_this_CPU() requires opening the graphics file and reading its Mach headers, so it is a somewhat expensive operation. It is wasteful / inefficient to repeat this for each task that uses the same graphics app.

Ideally, the check should only be made once per graphics app after the graphics app has been downloaded. It appears that the check is repeated for every task. Can you add a persistent flag in the client_state.xml file for each graphics app indicating that it has already been tested?

If that is not possible, please confirm that the method is called no more than once for each worker app.

A less major concern is this: if a task is begun before the graphics app has finished downloading, the "Show Graphics" button will never be enabled for that task because the task has already been identified as an Active Task. Though it's probably not practical to immediately change the status of the button while the task is running, it would be nice to do so when the Tasks window is updated for any reason, such as when the task is suspended and then resumed, tasks are added or removed, etc. The Show Graphics buttons are all updated, based on the Results struct in MainDocument.cpp whenever the Tasks window is updated. The Manager obtains a new get_results rpc from the client at frequent intervals.

The best time to do the test might be when the download of any graphics app finishes. The code could then step through all tasks from that project and fill in the graphics_exec_path for each task.

davidpanderson:
We're checking it once per app version, not per job.

I'm having some build issues related to compareOSVersionTo(), which was defined in two places.
I should have this fixed shortly.

CharlieFenton:
> I'm having some build issues related to compareOSVersionTo(), which was defined in two places.

Be very careful moving those around. I spent a fair amount of time reducing the number of places the same code needs to be repeated. For various reasons, it cannot be in just one source file, unless we created a separate source file for just that one routine. It is used by many modules, including the client, Manager, installer (Installer.app and PostInstall.app), Uninstaller, AddRemoveUser utility, graphics library and the gfx_cleanup screensaver helper executable.

CharlieFenton:
> We're checking it once per app version, not per job.

Thank you. I was misled by this sentence in the original description for this PR:
> When reporting an ACTIVE_TASK to the GUI,
> check whether the file is present and (on Mac) can be executed.

CharlieFenton:
The code for compareOSVersionTo() is repeated in exactly 3 source files: 
* client/gpu_opencl.cpp for all references within the client
* lib/macglutfix.m for the graphics library, where it is called only in that file.
* mac_util.mm for all other modules: Manger, Installer (including the PostInstall helper), Uninstaller, Screensaver (including its helper apps gfx_switcher and gfx_cleanup), the AddRemove User utility and SetUpSecurity utility.

The client cannot link with mac_util.mm due to restrictions on the use of Cocoa (objective-C) methods, which would interfere with our account-based sandbox security on the Mac; I won't go into the details explaining why. Linking the graphics library with mac_util.mm would require also linking with extra Mac Frameworks (collections of libraries), complicating things for project developers.

I strongly urge you not to make any changes in the area. It can too easily break things due to unintended consequences.

CharlieFenton:
@davidpanderson I am now pleading with you. Please leave the implementations of `compareOSVersionTo()` as they were. The changes you are making are far too extensive and are creating the opportunity for unintended consequences.

Please close this PR and start a clean one with only the changes needed for the original purpose: _client_app.cpp_, _client_state.cpp/.h_ and _client_types.cpp/.h_. I would prefer moving `can_run_on_this_CPU()` to _client_types.cpp_ where it is called, but I won't argue If you insist on moving it to hosting_unix.cpp and lib/hostinfo.h. None of the other changes you've been making belong in this PR because they are not required to fix the issue this PR was created for.

davidpanderson:
Replaced by #4141

PR #4140, opened 2021-01-07
Detect GLIBC version by calling gnu_get_libc_version()

CallMeFoxie:
**Description of the Change**
GLIBC version string can be detected easier by calling gnu_get_libc_version() instead of parsing it from ldd output.

Not all systems even have ldd script itself available to begin with as it is not required for normal system runtime. This function was made available around libc 2.1 I think so it is pretty safe nowadays.

**Alternate Designs**
There was also __GLIBC__ and __GLIBC_MINOR__ macros but those are compile time.

**Release Notes**
Detect and report GLIBC version from a glibc function call rather than from parsing ldd output when possible.

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -41,6 +41,10 @@
 #include <cstring>
 #endif
 
+#ifdef __GLIBC__
+#include <gnu/libc-version.h>
+#endif
+
 #if HAVE_XSS
 #include <X11/extensions/scrnsaver.h> //X-based idle detection
 // prevents naming collision between X.h define of Always and boinc's
@@ -1371,6 +1375,12 @@ int HOST_INFO::get_memory_info() {
 // return BOINC_SUCCESS if at least version could be found (extra_info may remain empty)
 // return ERR_NOT_FOUND if ldd couldn't be opened or no version information was found
 //
+#ifdef __GLIBC__
+int get_libc_version(string& version, string& extra_info) {
+    version = string(gnu_get_libc_version());
+    return BOINC_SUCCESS;
+}
+#else
 int get_libc_version(string& version, string& extra_info) {
     char buf[1024] = "";
     string strbuf;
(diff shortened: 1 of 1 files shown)


AenBleidd:
@CallMeFoxie, could you please provide an example output of `gnu_get_libc_version()` after your change?

CallMeFoxie:
it drops the "Ubuntu" or "Debian" part of the version string and reports 
purely the actual libc version, is that a problem?

```

ashley@foxtop:/tmp$ cat foobar.c
#include <gnu/libc-version.h>

#include <iostream>

int main(void) {
     std::string ver;
     ver = std::string(gnu_get_libc_version());
     std::cout << ver << "\n";
}
ashley@foxtop:/tmp$ g++ foobar.c
ashley@foxtop:/tmp$ ./a.out
2.31

ashley@foxtop:/tmp$ ldd --version
ldd (Ubuntu GLIBC 2.31-0ubuntu9.1) 2.31
...

```


the patch is in use here 
https://lhcathome.cern.ch/lhcathome/hosts_user.php?userid=785180 on the 
ARM64 boards.


Ashley

AenBleidd:
Looks like that extra information is lost (in your example it's shown in brackets: `Ubuntu GLIBC 2.31-0ubuntu9.1`). Not sure whether this extra information is really necessary.
@davidpanderson, what do you think?

ChristianBeer:
I implemented this by parsing the ldd output so we are not restricted to glibc distributions. At the time I investigated this it seemed to be the most portable solution. See my original commit: https://github.com/BOINC/boinc/commit/870c04afa42c60ec7d73c906e3dab346093b7426

I liked this solution because it also showed the specific glibc version info from the distribution not just the major and minor version numbers. This might help in troubleshooting on the project side. In the end projects probably just need the plain "2.32" or "2.28" either way. The distribution name is gathered by another function in this file so the extra information is strictly not needed.

I'm fine with the changes in this PR.

truboxl:
Since the OS string still pertains "Linux Ubuntu", projects can use that to identify the OS rather than parsing from libc version string...

AenBleidd (approved):

PR #4144, opened 2021-01-11
Drupal: ignore temporary field

brevilo:
**Description of the Change**
Ignore another form field. Its value is used for temporary purposes only.

**Alternate Designs**
none

**Release Notes**
N/A


tristanolive:
This should be fine. The current password is used in the validation state first, to verify that it is correct, and then again during the update state in cases where the email address is being changed (need current pw in order to generate the new hash). After that, it would be unnecessary.

PR #4145, opened 2021-01-11
[Android] Add Xiaomi specific screens

AenBleidd:
Add Xiaomi specific builds.

### DIFF
--- a/android/BOINC/app/build.gradle
+++ b/android/BOINC/app/build.gradle
@@ -94,6 +94,12 @@ android {
             testCoverageEnabled true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules-debug.txt'
         }
+        xiaomi_release {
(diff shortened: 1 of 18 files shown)


davidpanderson:
I suggest that you rename things (function and files) to remove "Xiaomi".
It's possible that we'll want intro screens for other purposes.

AenBleidd:
@davidpanderson, I made requested changes. I left check for 'xiaomi' build since it's a specific build for them and it has very specific requirements. Further we can extend it if we will have more non-general releases. 
Please review it one more time.

ChristianBeer:
Please also note that Translation need to go into transifex or else they get overwritten next time they are updated. The translations can be added once this updated values/strings.xml is sent to transifex. Which I can do after this is merged.

Isira-Seneviratne (approved):

PR #4146, opened 2021-01-12
Add CTRL+TAB shortcut

### LINKED ISSUE #3359: Manager: missing/inconsistent keyboard handlers

winkies:
Partially Fixes #3359

**Description of the Change**
Define shortcut on accelerator table and bind it with event table.

**Alternate Designs**
Tested only on Linux environment. It should work as well on Windows but **it should be tested on MacOS X**.

**Release Notes**
Add CTRL+TAB shortcut for advance view of boinc manager.

### DIFF
--- a/clientgui/AdvancedFrame.cpp
+++ b/clientgui/AdvancedFrame.cpp
@@ -205,6 +205,7 @@ BEGIN_EVENT_TABLE (CAdvancedFrame, CBOINCBaseFrame)
     EVT_TIMER(ID_FRAMERENDERTIMER, CAdvancedFrame::OnFrameRender)
     EVT_NOTEBOOK_PAGE_CHANGED(ID_FRAMENOTEBOOK, CAdvancedFrame::OnNotebookSelectionChanged)
     EVT_MENU(ID_SELECTALL, CAdvancedFrame::OnSelectAll)
+    EVT_MENU(ID_NEXT_PAGE, CAdvancedFrame::OnNextPage)
     EVT_SIZE(CAdvancedFrame::OnSize)
     EVT_MOVE(CAdvancedFrame::OnMove)
 #ifdef __WXMAC__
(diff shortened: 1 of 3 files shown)


AenBleidd:
@winkies, it is possible to extend this PR to include all changes that are requested in #3359?
Otherwise I'll ask @RichardHaselgrove to split his ticket into two

AenBleidd:
This is not needed on Windows since `Ctrl-Tab` and `Shift-Ctrl-Tab` already work fine without this fix. I don't know about Mac, but for now it looks like that this should be a linux-only fix . Also please add `Shift-Ctrl-Tab`  shortcut to iterate tabs backward (and for linux-only too).

winkies:
@AenBleidd, I'm on `Ctrl+Shift-Tab` shortcut.

About the `Tab` / `Shift-Tab` issue, I think that is more complex than just add shortcut on the advanced frame. 
Moreover, it need Windows environment to reproduce the issue.

RichardHaselgrove:
> This is not needed on Windows...

Yes, it is: Ctrl-tab still gets stuck on the Notices page, as noted in #3359.

CharlieFenton:
I object to this PR. WxWidgets translates _Ctrl_ in accelerators to Cmd (the command key) on Macs. Cmd+Tab is a standard keyboard shortcut on Macs to step through open applications. It is equivalent to Alt+Tab on MDS Windows. The earlier change to add Ctrl+A as a shortcut for Select All is OK because that is the standard meaning of Cmd+A on Macs.

In addition, there are already keyboard shortcuts to select each of the tabs. I see little need for another shortcut to be able to step through the tabs.

This need to avoid conflicting with standard keyboard shortcuts is the reason most BOINC-specific accelerators include the Shift key.

AenBleidd:
So this should be linux-only feature because looks like linux is the only platform that has no these bindings in wxWidget by default

CharlieFenton:
Please also see [this comment](https://github.com/BOINC/boinc/issues/3359#issuecomment-759173518) in #3359.

RichardHaselgrove:
Testing with win_manager_cbd87af9f19501ff306d54aeb44e12249f2a56d6: partial success.

Starting with the focus on the tab label row: Ctrl-Tab and Ctrl-Shift-Tab cycle through all pages (as before).

Starting with the focus on another control in the body of a page (default view), Ctrl-Tab and Ctrl-Shift-Tab cycle through all pages **without** getting stuck on Notices: that's an improvement.

But if you click or tab into the body of the Notices page, there's no way out via the keyboard, except by going through the main menu with accelerator keys. And the accelerators aren't visible until you press 'Alt'.

winkies:
Please see the [lastest status](https://github.com/BOINC/boinc/issues/3359#issuecomment-762508290) about Notices page issue.

IMO, you can split the ticket in two except if someone wants to give it a shot on this.

PR #4149, opened 2021-01-14
[VS2019] Add fallback sign server timestamp.globalsign.com to use it …

AenBleidd:
…for signing when timestamp.verisign.com is not reachable


AenBleidd:
@BOINC/committer, could someone please review and merge this minor enhancement?

PR #4152, opened 2021-01-20
Allow invitation codes for create-account RPCs only

davidpanderson:
If you define INVITE_CODES_RPC, create-account RPCs must include
a matching invitation code.
Record this code in user.signature to keep track
of where the RPCs are coming from.


Rytiss:
User.signature field is already (ab)used for email validation. Fields should not be reused, especially in multiple places where collisions can happen.

davidpanderson:
With this change, user.signature is no longer used for email validation.

PR #4154, opened 2021-01-21
Fix sah_select_bitness.m4

### LINKED ISSUE #853: configure --enable-bitness=32 doesn't work
**Reported by Nicolas on 4 Apr 39293706 04:00 UTC**
I tried to compile BOINC API for 32-bit Linux, being on a 64-bit Linux machine. Using `configure --enable-bitness=32` doesn't work; it compiles 64-bit code anyway (I didn't let `make` run to completion; I aborted it as soon as I saw a few files being compiled without any 32-bit flag).

I found the problem is the `SAH_OPTION_BITNESS` macro (defined in `m4/sah_select_bitness.m4`) adds `-m32` to `CFLAGS`, but not to `CXXFLAGS`, so C code would be compiled as 32-bit but C++ code wouldn't.

Another minor bug: if the macro detects `-m32` works, it prints "Selecting 32 bit model... -m32" but then prints "failed" on the next line anyway. The printing of "failed" is done always, instead of in the 'else' clause of checking if `-m32` works.

Migrated-From: http://boinc.berkeley.edu/trac/ticket/876

truboxl:
Fixes #853

**Description of the Change**
Rework `sah_select_bitness.m4` to:
fix bitness test,
clarify errors and messages,
clean up formatting and use 2 spaces indents,
allow user defined ${OBJDUMP},
change from hardcoded file descriptor to AS_MESSAGE_LOG_FD,
add C++ compiler bitness tests (C and C++ must be same bitness to pass test)

**Alternate Designs**
Better use declarative unit test rather than iterative unit test in the future
Tested so far on Linux and emsdk

**Release Notes**
N/A

### DIFF
--- a/m4/sah_select_bitness.m4
+++ b/m4/sah_select_bitness.m4
@@ -39,25 +39,27 @@ AC_DEFUN([SAH_DEFAULT_BITNESS],[
 
 AC_DEFUN([SAH_SELECT_BITNESS],[
   SAH_DEFAULT_BITNESS
-  AC_LANG_PUSH(C)
-  AC_MSG_CHECKING(Selecting $1 bit model)
-  echo "int main() { return 0; }" >conftest.$ac_ext 
-  if test "$1" != "${COMPILER_MODEL_BITS}"
-  then
-    ${CC} ${CFLAGS} ${CPPFLAGS} -m$1 -c conftest.$ac_ext >&5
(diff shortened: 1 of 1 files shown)


AenBleidd:
@SETIguy, could you please review this? Thanks in advance

AenBleidd (approved):
I'm not a big expert in m4 scripts, bit so far it looks fine for me.

PR #4157, opened 2021-01-23
[Android] Fix Tasks loading

### LINKED ISSUE #2123: Android Boinc shows 0 tasks
If there are too much tasks (more than 100?) are running but the android client shows 0 tasks.

This happens on a Galaxy S5 Neo with Android 6.0.1.

See also https://boinc.berkeley.edu/dev/forum_thread.php?id=11640

AenBleidd:
On modern devices with more than 4 cores or when there is a big tasks cache set could happen a situation when 0 tasks are shown.
This happens because of the limit of 1MB to transfer data between processes.
This commit breaks this flow and instead adds a possibility to load data by chunks.
Also small optimization implemented to update only active tasks every second.
Full update is done every 10 seconds now because most of the time non-active tasks are not updated so often.

This fixes #2123

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/TasksFragment.kt
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/TasksFragment.kt
@@ -1,7 +1,7 @@
 /*
  * This file is part of BOINC.
  * http://boinc.berkeley.edu
- * Copyright (C) 2020 University of California
+ * Copyright (C) 2021 University of California
  *
  * BOINC is free software; you can redistribute it and/or modify it
  * under the terms of the GNU Lesser General Public License
@@ -43,11 +43,13 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import kotlin.collections.ArrayList
 import java.util.*
 
 class TasksFragment : Fragment() {
     private lateinit var recyclerViewAdapter: TaskRecyclerViewAdapter
     private val data: MutableList<TaskData> = ArrayList()
+    private var lastFullUpdateTimeStamp: Long = 0
     private val mClientStatusChangeRec: BroadcastReceiver = object : BroadcastReceiver() {
         override fun onReceive(context: Context, intent: Intent) {
             if (Logging.VERBOSE) {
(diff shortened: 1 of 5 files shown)


Isira-Seneviratne (approved):

AenBleidd:
@BOINC/committer, please merge

android/BOINC/app/src/main/java/edu/berkeley/boinc/TasksFragment.kt
```diff
@@ -89,29 +91,78 @@ class TasksFragment : Fragment() {
         super.onPause()
     }
 
-    private fun loadData() {
-        // try to get current client status from monitor
-        val tmpA = try {
-            BOINCActivity.monitor!!.tasks
+    private fun loadTasks(start: Int, count: Int, isActive: Boolean): MutableList<Result>? {
+        return try {
+            BOINCActivity.monitor!!.getTasks(start, count, isActive)
```
Isira-Seneviratne, line 96:
```suggestion
        return BOINCActivity.monitor?.getTasks(start, count, isActive) ?: emptyList()
```

android/BOINC/app/src/main/java/edu/berkeley/boinc/TasksFragment.kt
```diff
@@ -89,29 +91,78 @@ class TasksFragment : Fragment() {
         super.onPause()
     }
 
-    private fun loadData() {
-        // try to get current client status from monitor
-        val tmpA = try {
-            BOINCActivity.monitor!!.tasks
+    private fun loadTasks(start: Int, count: Int, isActive: Boolean): MutableList<Result>? {
```
Isira-Seneviratne, line 94:
```suggestion
    private fun loadTasks(start: Int, count: Int, isActive: Boolean): MutableList<Result> {
```

android/BOINC/app/src/main/java/edu/berkeley/boinc/TasksFragment.kt
AenBleidd: Fixed

android/BOINC/app/src/main/java/edu/berkeley/boinc/TasksFragment.kt
AenBleidd: Fixed

PR #4159, opened 2021-01-27
Enhance clean commands for build unix

### LINKED ISSUE #1328: use standard targets for unix build process
**Reported by ChristianB on 18 Jan 44414593 10:40 UTC**
Following the GNU convention for [makefile targets](http://www.gnu.org/prep/standards/html_node/Standard-Targets.html#Standard-Targets) I ran some tests with a fresh copy of the boinc-v2 source. I identified
several files that can be ignored and get deleted by clean/distclean and files that should be deleted by clean/distclean but are not.

Here is a short summary of the convention:

Files created by ./_autosetup need to be deleted in case of a system
upgrade because most likely the tools tested are updated. Those files
are not deleted via Makefile because it is a step before Makefile's exist.
Maybe adding an option to _autosetup that clears the files before
running the tests could help in this case.

What's important for building is that after running "make clean" we
should have the same files as after running "./configure" and after
running "make distclean" we should have the same files as after
"./_autosetup".

The attached textfile contains all those files in gitignore style and
surely some of them should go into .gitignore.

Could someone who wrote the _autosetup and/or build stuff enhance clean
and distclean please?

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1357

winkies:
Fixes #1328 

**Description of the Change**
Add files for `clean` and `distclean` commands.
Define `maintainer-clean` command for the cleaning of `_autosetup` command. 

**Alternate Designs**
Use `git clean -fd` command but not integrate in build process.

**Release Notes**
Enhance clean commands for build unix.


ChristianBeer:
I did some tests and found that when running `maintainer-clean` the files `version.h` and `version.h.in` get deleted. But `_autosetup` needs the latter to succeed. When restoring the file everything works again. Is this intended?

Error: `configure.ac:1294: error: required file 'version.h.in' not found`

AenBleidd:
Not sure about `version.h.in` file but `version.h` must not be deleted while running clean because it's not autogenerated

winkies:
`version.h` is deleted by running `make distclean` on **master** branch, so I let the default behavior.

`version.h.in` is deleted by `make maintainer-clean`, certainly cause `*.in` in `MAINTAINERCLEANFILES` variable.

winkies:
I replaced `*.in` in `MAINTAINERCLEANFILES` by all files except `version.h.in`. 

So, after `make maintainer-clean`, we can re-run `./_autosetup`.

ChristianBeer:
Looks good now. Thanks for the quick changes. I'm doing a quick final check on my desktop as soon as I have access again.

winkies:
You're welcome. FYI, the file `version.h` is still deleted by `make distclean` but it can be regenerated by `./configure`.

ChristianBeer:
Thanks. Just tested it and works like a charm. Merging now.

Issue #2339, opened 2018-02-05
reference .desktop file


JuhaSointusalo:
Different Linux distros have different ideas of what should be in .desktop file (the file behind the entry in start menu) for Manager. All the different versions could be upstreamed so that there is one reference file.

The most important part of the file is using `Path` to set working directory to BOINC's data directory so that Manager can find `gui_rpc_auth.cfg`. It would be nice if translation of the file was integrated with the rest of the BOINC translation system.

[Arch Linux](https://git.archlinux.org/svntogit/community.git/tree/trunk/boinc.desktop?h=packages/boinc)
[Debian](https://anonscm.debian.org/cgit/pkg-boinc/boinc.git/tree/debian/boinc-manager.desktop)
[Fedora](https://src.fedoraproject.org/rpms/boinc-client/blob/master/f/boinc-manager.desktop)
[Mageia](http://sophie.zarb.org/rpms/a43a82667e3b928558f576b6007cfc5b/files/3)
[Suse](https://build.opensuse.org/package/view_file/network/boinc-client/boinc-gui.desktop?expand=1)

BryanQuigley:
I don't seem to need to set path - will have to test more cases.  Having trouble building boinc atm, but will try to integrate in future.

Currently testing with:
[Desktop Entry]
Version=1.0
Type=Application
Name=BOINC Manager
Exec=boincmgr
Icon=boincmgr
Terminal=false
Keywords=computation;science;
Categories=System;Monitor;GTK;
GenericName=BOINC monitor and control utility
GenericName[cs]=Monitorovací a ovládací nástroj pro BOINC
GenericName[de]=BOINC Überwachungs- und Kontrollprogramm
GenericName[pt]=Monitorização BOINC e utilitário de controlo
GenericName[pt_BR]=Utilitário de controle e monitoramento do BOINC
Comment=Configure or monitor a BOINC core client
Comment[cs]=Monitoruje a nastavuje klienta BOINC
Comment[de]=BOINC Basis Client konfigurieren oder überwachen
Comment[pt]=Configurar ou monitorizar o cliente básico do BOINC
Comment[pt_BR]=Configure ou monitore o cliente básico do BOINC

JuhaSointusalo:
Unless you have empty GUI RPC password (Debian/Ubuntu style) or have linked `gui_rpc_auth.cfg` to your home directory (IIRC, openSUSE style) then I'm pretty sure you need to set `Path`.

[...]

PR #4171, opened 2021-02-07
[Linux] Adds desktop file to launch boinc manager

BryanQuigley:
Also moves the 64x64 boinc png from a directory for this version.
Long term it makes sense to reorg the icons so we can more easily
reference them.

Resolves: #2339

Fixes #

**Release Notes**
BOINC desktop file is now installed when installing the manager


BryanQuigley:
@AenBleidd Are you able to review this? Thanks!

AenBleidd (approved):

AenBleidd:

RichardHaselgrove:
If I understand the purpose of this PR correctly, I think I'll have to wait until the Debian maintainers adopt this file in place of their existing solution, and @LocutusOfBorg's automated build distributes it through his PPA. Unless you, Debian, or Gianfranco can suggest a shortcut?

I'm also worried that the clientgui/res/boinc.desktop file contains hard-coded translations in (very few of) the supported non-English languages. Is that the right way to proceed, rather than Transifex?

BryanQuigley:
If you enable manager and do an install it will now install the desktop file and icon. Debian can simplify their setup around this - but they don't *have* to IMU.

I guess it's unlikely [this](https://www.schauenburg.nl/posts/2021/01/11_translators_-_please_do_not_translate_everything.html) will happen with BOINC, but I'd like to have a better way to avoid it.  (I also have no experience with the translation system here, so pointers welcome!)

LocutusOfBorg:
as long as you merge it I can see if I can drop the debian specific desktop file and use the system one!

RichardHaselgrove:
@AenBleidd - what are the conventions for that? I have the merge button, but am I allowed to press it **before** testing? We can revert if it goes wrong, yer?

LocutusOfBorg:
The desktop file is really similar to the one we have in Debian, so I'll patch it a little bit and just use.

BryanQuigley:
I was trying to make it compatible with all Linux distros, what needs
changes for Debian?

LocutusOfBorg:
```
-Path=/var/lib/boinc
+Path=/var/lib/boinc-client
```
This is something in Debian we have differently, but no real big deal here :)

RichardHaselgrove:
When I install from your PPA to Linux Mint, the system creates both `/var/lib/boinc` and `/var/lib/boinc-client`. The full files are in boinc-client, and client provides a soft-link to them. So both destinations are covered. 

I don't know whether any distributions still use `/var/lib/boinc`, and if so, whether they provide a similar fallback from `/var/lib/boinc-client`.

AenBleidd:
@RichardHaselgrove, you could merge this and test afterwards.
Regarding translations, I'll use existing translation files afterwards and fix them

RichardHaselgrove:
I'll do that later today (after lunch - I have some long-running (8 day) tasks due to finish then).

Meanwhile, preserving the existing Linux Mint/PPA desktop file for later comparison:
```
[Desktop Entry]
Type=Application
Version=1.0
Exec=boincmgr
Path=/var/lib/boinc-client
Icon=boincmgr
Categories=System;Monitor;GTK;
Keywords=computation;science;
Name=BOINC Manager
GenericName=BOINC monitor and control utility
Comment=Configure or monitor a BOINC core client
```
The alternative `Path=/var/lib/boinc` also works.

RichardHaselgrove:
Checking by eye while waiting for my tasks to finish...

The Icon= line in my existing file, and in the original issue #2339, names the icon as boincmgr - which seems logical for an application called boincmgr.

The file created here in 24de0ea29d814625e53df0081c2a7d3e0c0b744b is named simply `boinc`. Are we all sure we want to make this change?

RichardHaselgrove:
Well, I've tested what I can. Removed user from boinc group, installed updates, re-booted, re-started client, deleted launch icon, recreated launch icon with text editor from this commit.

1) Required me to give it trusted status - reasonable for a manual install.
2) BOINC Manager itself couldn't read the password - as usual without group membership.
3) Re-opened with text editor - complained about invalid characters on both open and save. Wouldn't launch app - corrupted file.

Take 2 - create file as before
Grant trusted status
Edit via right-click, properties. Add password.
File can now be edited - no invalid characters. Translation lines have been removed.
BOINC Manager opens normally, displays full contents.

So, at the user level, this change makes no discernable difference - just the expected variation between manual and automatic deployment. I'll merge it, and let the package maintainers loose on it.

PR #4176, opened 2021-02-10
android: new domain name creation method

### LINKED ISSUE #2846: Change the domain name creation process on Android
**Describe the problem**
The Boinc Client generates a domain name like `android_xxxxxxxx` on Android devices. It is really hard to track on the project's website  which device how much point generated when you have multiple devices.

Change the domain name generation procedure to `Manufacturer - Model`, which is more human readable and recognizable then a random string.

**Additional context**
The `boinc/client/hostinfo_network.cpp` file have to be modified.

truboxl:
Fixes #2846 

**Description of the Change**
New domain name will be created based on brand, model, random strings on Android, else fallback to old way if something went awry

**Alternate Designs**
No idea how to do this using Java / Kotlin
I pick brand over manufacturer, since some phones have brand "Nokia" but manufacturer "HMD Global"

**Release Notes**
client: new android domain name creation method

### DIFF
--- a/client/hostinfo_network.cpp
+++ b/client/hostinfo_network.cpp
@@ -82,6 +82,9 @@ int HOST_INFO::get_local_network_info() {
     } else {
         snprintf(domain_name, sizeof(domain_name), "android_%s", buf);
     }
+    if (!cc_config.device_name.empty()) {
+        safe_strcpy(domain_name, cc_config.device_name.c_str());
+    }
     return 0;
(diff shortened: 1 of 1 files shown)


truboxl:
Information for `__system_property_get()`
https://stackoverflow.com/questions/28413530/api-to-get-android-system-properties-is-removed-in-arm64-platforms/28416743#28416743

PR #4207, opened 2021-02-19
Maximum CPU time - spread tasks evenly over timeslot #4181

### LINKED ISSUE #4181: Maximum CPU time - spread tasks evenly over timeslot
**Describe the problem**
When choosing a value of 'Maximum % of CPU time" BOINC will schedule the runtime in seconds.
For example, selecting 25% will make BOINC tasks run for one second, and sleep for 3 seconds.
In a multi core environment this is quite detrimental for performance, power spikes and more.
Also when all the tasks start at once, the computer often is less responsive for one second, and then 3 seconds is very usable.

It would be beneficial for the system to spread the multi-core tasks in time, using slots.
For example, if we have an 8 core system with a 25% usage selected, we have have 8 tasks running:
- First second: run 2 tasks out of the 8 (tasks 1, 2)
- Second second: run next 2 tasks out of the 8 (task 3, 4)
- Third second: run other 2 tasks (task 5, 6)
- Fourth second: run other 2 tasks (tasks 7,8)

This would provide a better user experience:
- Fans will not throttle up and down that aggressively reducing noise
- Power supplies have a less hard time
- More gentle thermal effect on CPU (less expansion/contraction cycles
- Computer is more usable in general
- Simply makes much more sense

**Additional context**
In case you want your computer to consume a steady amount of power this is the only way.

Note: It would be beneficial that CPU % LOAD is separately configurable to GPU % LOAD. Same concept, but time slots also take into account amount of GPUs installed in the system.

cnergyone:
Fixes #4181

**Description of the Change**
Modified the behavior of cpu_usage_limit to run tasks spread over a 100 second window, instead of all at once.

**Alternate Designs**
The old implementation gave a bad user experience, as all tasks where running for a short time, blocking to PC for other processes. This implementation will smooth out the cpu usage over a longer time window. 

**Release Notes**
Changed the cpu_usage_limit throttling behavior to spread out tasks over time.

### DIFF
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -1184,7 +1184,9 @@ void* throttler(void*) {
 #endif
     unsigned int current_tick = 0;
     unsigned int task_count;
-    unsigned int spread;
+    unsigned int stride;
+    unsigned int run_time;
+    float cpu_usage_limit;
 
     // Initialize diagnostics framework for this thread
     //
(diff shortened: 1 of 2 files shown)


davidpanderson:
I don't understand what this does, or what problem it solves; please explain.

cnergyone:
From the ticket:

Describe the problem
When choosing a value of 'Maximum % of CPU time" BOINC will schedule the runtime in seconds.
For example, selecting 25% will make BOINC tasks run for one second, and sleep for 3 seconds.
In a multi core environment this is quite detrimental for performance, power spikes and more.
Also when all the tasks start at once, the computer often is less responsive for one second, and then 3 seconds is very usable.

Describe the solution you'd like
It would be beneficial for the system to spread the multi-core tasks in time, using slots.
For example, if we have an 8 core system with a 25% usage selected, we have have 8 tasks running:

First second: run 2 tasks out of the 8 (tasks 1, 2)
Second second: run next 2 tasks out of the 8 (task 3, 4)
Third second: run other 2 tasks (task 5, 6)
Fourth second: run other 2 tasks (tasks 7,8)
Second example:
In a 100 core system with a 1% CPU load selected:
In the old situation all 100 cores would run for 1 second, completely bogging down the system.
For the next 99 seconds there would be no BOINC tasks running.
In the new situation, every task would run for 1 second in consecutive order.

This would provide a better user experience:

Fans will not throttle up and down that aggressively reducing noise
Power supplies have a less hard time
More gentle thermal effect on CPU (less expansion/contraction cycles
Computer is more usable in general
Simply makes much more sense
Additional context
In case you want your computer to consume a steady amount of power this is the only way.

Note: It would be beneficial that CPU % LOAD is separately configurable to GPU % LOAD. Same concept, but time slots also take into account amount of GPUs installed in the system.

dirspb:
> if we have an 8 core system with a 25% usage selected, we have have 8 tasks running:
> First second: run 2 tasks out of the 8 (tasks 1, 2)
> Second second: run next 2 tasks out of the 8 (task 3, 4)
> Third second: run other 2 tasks (task 5, 6)
> Fourth second: run other 2 tasks (tasks 7,8)
Such behavior can already be achieved by configuring: Use 25% of the CPUs, use 100% of CPU time.

Issue #3822, opened 2020-06-10
[Bug][Windows] Disk tab advanced view in Hebrew show in mirror view. 


talregev:
> **Describe the bug**
> A clear and concise description of what the bug is.

Disk tab advanced view in Hebrew show in mirror view.  Instead Right to left.
> **Steps To Reproduce**

1. Install boinc
2. Switch to Hebrew (I don't know how).
> **Expected behavior**
> A clear and concise description of what you expected to happen.

It show the disk in Hebrew in Right to left in correct way.

> **Screenshots**
> If applicable, add screenshots to help explain your problem.

![Disk-hebrew-bug](https://user-images.githubusercontent.com/7043539/84250012-28c64700-ab14-11ea-9081-159ed38f8a9a.PNG)
> **System Information**

 - OS: Windows 10 build 2004 (Yes with the new May 2020 update).
 - BOINC Version: 7.16.7

> **Additional context**
> Add any other context about the problem here.

Even that you don't understand the Hebrew letters, you can notice the few English letters that they in mirror view. I think this bug for all right left language that boinc support.

AenBleidd:
Closed as a duplicate of #988

Issue #988, opened 2015-02-04
Localization of right-to-left languages


romw:
**Reported by mo v on 7 Nov 40709268 20:26 UTC**
Ido Schwartz, the Boinc Hebrew translator, has noticed problems in the Boinc manager Hebrew translation arising from the fact that this language is written right-to-left. He expects the same problems to arise with all RTL languages eg Arabic. 

In Boinc 6.11.6 and again in 6.11.7 he noticed that:
- the text in the Statistics and Disk tabs is inverted and unreadable
- the graphs are inverted from LTR to RTL 
- in the "Simple View" mode, the "Advanced View", "Preferences", "Pause" and "Messages" buttons do not translate. (I am not sure myself, however, whether this is related to the RTL problem)
- in the "Preferences" window the text is aligned to the left instead to the right

Ido has provided 3 screenshots from 6.11.6 showing the problem. In each case part of the BM is shown in its Hebrew and English versions.  

Statistics tab: ![](http://i15.photobucket.com/albums/a390/Dj_Leg0la5/Statistics.jpg)

Disk: ![](http://i15.photobucket.com/albums/a390/Dj_Leg0la5/Disk.jpg)

Simple View: ![](http://i15.photobucket.com/albums/a390/Dj_Leg0la5/simple.jpg)

A resolution of this problem would help our efforts to make Boinc more accessible world-wide. 

Thank you

Mo

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1011

romw:
**Commented by mo.v on 12 Sep 40710417 08:00 UTC**
Ido also says:


As far as I can guess, the problem is the alignment of the entire UI to the right.
I think the alignment includes everything, even the statistics info pic.
As a result, it should be corrected if there will be a checking whether the UI Interface is RTL or not.
If it does, only the content of the Disk and Statistics Tabs - meaning the generated info pics section only - should be aligned back to the left, or another code that generates those info-pics will be needed, only for the RTL Languages.

ChristianBeer:
Possibly related or duplicate of #911. Needs to be tested with current 7.6.33

AenBleidd:
Still an issue. Please take a look at my comment in #911

talregev:
The problem is more simple.
There is in the code mirror view.
Don't use it.
It not effect the text, and it only in this tab.

AenBleidd:
@talregev, how about fixing it? You're among these rare people who use Hebrew and could really see the problem and verify it ;)

[...]

PR #4210, opened 2021-02-20
Fix Disk tab in advanced view at Hebrew show as mirror view.

talregev:
Fix: https://github.com/BOINC/boinc/issues/3822 https://github.com/BOINC/boinc/issues/988

I try to look on the CPP code, and understand where is the decision is made that the disk tab is mirror when choosing Hebrew.
All other tab is nor mirror, it specific tab, so it made me believe that the decision is in our code.
Any help will be appreciated, hopefully I will crush this bug.

Can you help me find some pointer in the cpp code? Thanks in advance.

### DIFF
--- a/clientgui/ViewStatistics.cpp
+++ b/clientgui/ViewStatistics.cpp
@@ -1945,6 +1945,8 @@ CViewStatistics::CViewStatistics(wxNotebook* pNotebook) :
 	m_PaintStatistics = new CPaintStatistics(this, ID_LIST_STATISTICSVIEW, wxDefaultPosition, wxSize(-1, -1), 0);
 	wxASSERT(m_PaintStatistics);
 
+	m_pTaskPane->SetLayoutDirection(wxLayout_LeftToRight);
+	m_PaintStatistics->SetLayoutDirection(wxLayout_LeftToRight);
     itemFlexGridSizer->Add(m_pTaskPane, 1, wxGROW|wxALL, 1);
     itemFlexGridSizer->Add(m_PaintStatistics, 1, wxGROW|wxALL, 1);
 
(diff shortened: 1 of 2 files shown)


talregev:
I force the layout to be LTR in the disk tab and the statistics tab, I change to Hebrew, but I cannot see the Hebrew words.  
Fail to load somehow the language. I guess it cannot find the language files. 
Also the boinc cannot connect to the internet. I am not sure why. 
Any Hint? 
I compile with visual studio community 2019, on windows 10, running boinc manager inside visual studio.

AenBleidd:
Not sure about the Internet, but if you want locale to be used in Manager, you should copy 'locale' folder from the installation location to your debug executable folder

talregev:
Thanks @AenBleidd It was help!
I have disk tab results: 
The hebrew now is not mirror although all sentence is to the left. but it better then when it mirror. it improvement! 
I cannot verify not the statistics tab because it lacking to connect to the internet.  

Before
After
Hebrew after

talregev:
Also fix the statistics now it look perfect :)
Statistics after

clientgui/ViewStatistics.cpp
```diff
@@ -1945,6 +1945,8 @@ CViewStatistics::CViewStatistics(wxNotebook* pNotebook) :
 	m_PaintStatistics = new CPaintStatistics(this, ID_LIST_STATISTICSVIEW, wxDefaultPosition, wxSize(-1, -1), 0);
 	wxASSERT(m_PaintStatistics);
 
+	m_pTaskPane->SetLayoutDirection(wxLayout_LeftToRight);
```
AenBleidd, line 1948: Could you please fix indents it this whole funtion?

AenBleidd (changes requested):

clientgui/ViewStatistics.cpp
talregev: Yes, I did

PR #4224, opened 2021-02-21
Add finalize target that copy locale folder and ca-bundle.crt 

talregev:
Add finalize target that copy locale folder and ca-bundle.crt  to output project.


AenBleidd (changes requested):
I told you that I'll tell you how to do this properly ;)
This could be done in a more easy way. I'll write you in Discord later today with instructions what to do

win_build/boinc_finalize.targets
```diff
@@ -0,0 +1,10 @@
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+	<Target Name="AfterBuild">
```
AenBleidd, line 2:
```suggestion
	<Target Name="CopyCertificateAndLocaleFiles" DependsOnTargets="Build" AfterTargets="Build">
```

win_build/boinc_finalize.targets
```diff
@@ -0,0 +1,10 @@
...
+		<ItemGroup>
+			<LOCALE Include="..\locale\**\*.*" />
+			<CERT Include="..\curl\ca-bundle.crt" />
+		</ItemGroup>
+		<Copy SourceFiles="@(LOCALE)" DestinationFolder="$(OutDir)\locale\%(RecursiveDir)" SkipUnchangedFiles="true" />
+		<Copy SourceFiles="@(CERT)"   DestinationFolder="$(OutDir)"                        SkipUnchangedFiles="true" />
+	</Target>
+</Project>
```
AenBleidd, line 10: Please add new line

AenBleidd (changes requested):

win_build/boinc_finalize.targets
talregev: I will. All other projects and target doesn't have new line.

AenBleidd: I know but anyway for new files it's better to add it :)

AenBleidd:
image

Next files should not be copied:
- Makefile.am
- update_templates.sh
- update_translations.sh

'templates' should not be copied too.

Also only *.mo files should be copied to language folders .

AenBleidd (approved):
LGTM

Issue #4232, opened 2021-02-22
[Android] Remote monitoring: User should be able to edit gui_rpc_auth.cfg file


AenBleidd:
Part of #1699.

- [x] Add GUI element in Advanced settings to display current content of gui_rpc_auth.cfg file
- [x] If gui_rpc_auth.cfg file doesn't exist or empty - generate new password (32 characters long)
- [x] By default, this password should be hidden (with dots like password)
- [x] When user clicks on this element, dialog box should be displayed with current password unhidden
- [x] In this dialog user should be able to set new password, this password should be written to gui_rpc_auth.cfg file
- [x] Empty password is not allowed, user should be notified about this
- [x] Add description line to inform user: 'BOINC client will be relaunched after password change"

talregev:
> Part of #1699.
> 
> * [x]  Add GUI element in Advanced settings to display current content of gui_rpc_auth.cfg file
> * [ ]  If gui_rpc_auth.cfg file doesn't exist or empty - generate new password (32 characters long)
> * [ ]  By default, this password should be hidden (with dots like password)
> * [ ]  On the right side f the element should be a button 'Show\Hide' password
> * [ ]  If user clicks to show password, it should be unhide
> * [ ]  If user clicks to hide password, it should be hide again
> * [ ]  On the left of 'Show\Hide' password button should be 'Edit' password button
> * [ ]  If user clicks 'Edit' password button, it should be able to set new password, and this password should be written to gui_rpc_auth.cfg file
> * [ ]  If user sets new password while client is already running with remote monitoring on, user should be informed that after new password set, client will be restarted and all running tasks will be interrupted
> * [ ]  If user confirms change password when remote monitoring is on, restart client
> * [ ]  NICE TO HAVE: Add protection to both buttons ('Show' password and 'Edit' password) to confirm act with default device protection mechanism: password or Fingerprint/FaceID if available).

1/11

AenBleidd:
@talregev, You can check point right in the header

talregev:
I don't have write access to this issue

AenBleidd:
Ok, then I'll check them for you ;)

talregev:
I want to update my progress, so I will do it from the comment. not big deal

PR #4234, opened 2021-02-24
[Android] Add Tcp Socket preparation

talregev:
Small task from #4232
Add Tcp Socket preparation.
In This PR it will always connect to unix socket. User cannot control of it.
When it remote off - unix socket, it working great.
~~When I am using Tcp connection:
failed to connect to localhost/127.0.0.1 (port 31416) from /127.0.0.1 (port 46703) after 30000ms: isConnected failed: ECONNREFUSED (Connection refused)~~

~~Not sure why,
Any hint will be appreciated.~~


talregev:
```
2021-02-24 16:35:19.947 29494-29494/edu.berkeley.boinc E/BOINC_GUI: SplashActivity.BroadcastReceiver.onReceive() error: 
    android.os.NetworkOnMainThreadException
at edu.berkeley.boinc.client.IMonitor$Stub$Proxy.runBenchmarks(IMonitor.java:1726)
it throw exception on read exception. how ironic
_reply.readException();  <-  IMonitor.java:1726
```

I read about the exception: `android.os.NetworkOnMainThreadException`
https://stackoverflow.com/questions/6343166/how-to-fix-android-os-networkonmainthreadexception

> This exception is thrown when an application attempts to perform a networking operation on its main thread. 

Some can help me how to run the benchmark in another thread? Thank you.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
```diff
@@ -85,6 +88,8 @@
     public static final int MGR_SYNC = 31;
 
     private LocalSocket mSocket;
+    private Socket mTcpmSocke;
```
AenBleidd, line 91: `mTcpmSocke` -> `mTcpmSocket`

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
```diff
@@ -278,7 +330,11 @@ public synchronized boolean authorize(String password) {
      * @return true if connected to BOINC core client, false if not connected
      */
     public final boolean isConnected() {
-        return (mSocket != null && mSocket.isConnected());
+        if (mIsRemote) {
```
AenBleidd, line 333: I'd check here for bot local are remote connection. What if GUI is already connected to client locally and try to establish new remote connection?

AenBleidd (changes requested):

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
talregev: You mean mTcpSocket. Yes, I will change my typo.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
talregev: You can run the boinc only once. and you can run it two possibility. once you run it lets say with remote, it connect with connect remote different with connect unix and save remote member. The program knows what state of the boinc is run, and no need to guess.

talregev: ok. I push the change.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
```diff
@@ -85,6 +88,8 @@
     public static final int MGR_SYNC = 31;
 
     private LocalSocket mSocket;
+    private Socket mTcpSocke;
```
AenBleidd, line 91: It's still a typo here:
`mTcpSocke` -> `mTcpSocket`

AenBleidd (changes requested):

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
talregev: Fix! :)

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
```diff
@@ -219,12 +263,20 @@ public synchronized void close() {
             if (Logging.WARNING) Log.w(Logging.TAG, "output close failure", e);
         }
         try {
-            mSocket.close();
+            if(mIsRemote){
```
AenBleidd, line 266: This is wrong. When called from `open()` function, (e.g. `mIsRemote == true`), here you'll try to close `mTcpSocket` but it could be wrong because at that moment could be active local connection and thus `mSocket` will still be open and connection is active that is wrong.  I'd suggest to remove `if(mIsRemote)` condition and try to close both sockets (yes, one of them will give error because it's not connected but it's fine and could be just logged as warning)

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
```diff
@@ -163,12 +168,51 @@ private static String modeName(int mode) {
...
+        if (Logging.DEBUG) Log.d(Logging.TAG, "Connected successfully");
+        return true;
+    }
+
     /**
      * Connect to BOINC core client via Unix Domain Socket (abstract, "boinc_socket")
      *
      * @return true for success, false for failure
      */
     public boolean open(String socketAddress) {
```
AenBleidd, line 214: Content of
`public boolean open(String address, int port)`
and
`public boolean open(String socketAddress)`
is almost identical. This is not good.
Try either to merge these functions into one with condition check inside or move the common part to the different function that will be called from these functions.

AenBleidd (changes requested):

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
talregev: When you switch from local socket to tcp socket, you must do it with runClient in monitor.
You cannot be just switch from Local to Tcp, it must be with ordered sequence:
Kill boinc, run it in the correct mode, connect it with the correct mode. All things are made from monitor.  
Decision is made in runClient in monitor and not from RpcClient. When you Kill boinc and run it again with runClient in monitor, it will update the remote by connect to the correct function of open() and update the mIsRemote accordingly.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
talregev: To do function with 3 parameter that not all use (and put null in some of them) and choose from the paramter base on remote is a bad design. That why I am against to merge these functions into one.
I try to implement the other option that is better, with a base class of both tcpSocket and localSocket.
But to switch from localSocket to base class need api 17 (we are on 16).
To switch from tcpSocket to base class need api 19.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
talregev: I remote the mIsRemote

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
talregev: I did it with obj.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
AenBleidd: Idiot-proof design, you remember? ;)

talregev: I don't see as Idiot-proof design, I see that we can determinate the system state in different way. so I remove the mIsRemote.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
talregev: I refactor and did in other way.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
```diff
@@ -219,12 +225,14 @@ public synchronized void close() {
             if (Logging.WARNING) Log.w(Logging.TAG, "output close failure", e);
         }
         try {
-            mSocket.close();
+            if (mTcpSocket != null)  mTcpSocket.close();
```
AenBleidd, line 228: I'd put `mTcpSocket.close()` and `mSocket.close()` into different `try...catch` block because otherwise if first call (in this case `mTcpSocket.close()` fails, `mSocket.close()` won't be called).

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
```diff
@@ -1338,4 +1347,43 @@ public synchronized boolean runBenchmarks() {
             return false;
         }
     }
-}
\ No newline at end of file
+
+    private boolean initBuffersFromSocket()
+    {
+        boolean isLocal = mSocket != null;
```
AenBleidd, line 1353: It's safer to pass `isLocal` as a parameter to this function to delegate this decision to the caller function.

AenBleidd (changes requested):

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
talregev: as the logic now, even in the Idiot-proof design, you cannot have both allocate mTcpSocket and mSocket even when you switch. While you connect to one of them (when you switch), you clean the system.

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
talregev: ok

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/RpcClient.java
talregev: I added.

AenBleidd (approved):

PR #4242, opened 2021-03-03
[WIP][Android] Async bug on network command

talregev:
Fixes #

**Description of the Change**
Convert all async lifecycle tasks in settings to async Task runner

In android 8 and below when you want to change a settings like "cpu usage" or "cpu core" and other settings, it was throw
`android.os.NetworkOnMainThreadException`
You cannot run network operation on main thread.
So lifecylcle on old android is running in main tread.


talregev:
Not relevant anymore. solve it in MonitorAsync PR

PR #4061, opened 2020-10-20
lib: don't include config.h from .h files

davidpanderson:
Fixes #2964

PR #4245, opened 2021-03-04
fix API build for C apps

bema-aei:
**Description of the Change**
Fixes build and install of the API (headers) for building C apps

- there's a declaration of a `bool` variable in `diagnostics.h`, which makes C compilers throw an error. Protect this with `#ifdef __cplusplus`

### DIFF
--- a/lib/diagnostics.h
+++ b/lib/diagnostics.h
@@ -34,7 +34,9 @@
 #include <dlfcn.h>
 #endif
 
+#ifdef __cplusplus
 extern bool main_exited;
+#endif
 
 // some of the Android stuff below causes seg faults on some devices.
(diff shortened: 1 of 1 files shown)


AenBleidd:
config.h should not be installed, see discussion here: #2964

bema-aei:
> config.h should not be installed, see discussion here: #2964

Thanks. I'll investigate what went wrong with my build.

bema-aei:
Thanks for pointing me to issue #2964. Possibly there was an old header lying around. This problem was actually fixed in #4061. The `diagnostics.h` fix is still necessary, though.

AenBleidd (approved):
LGTM

PR #4247, opened 2021-03-04
[Android] Null bug in the drawer

talregev:
Fixes # Null bug in the drawer


talregev:
This fixs are included inside MonitorAsync PR

PR #4260, opened 2021-03-12
Fix gcc compile warnings, and fix code formatting

davidpanderson:


AenBleidd:
I'll fix issues for Android and Mac build. I already know what is the reason of that and will provide a fix soon. All other changes looks good to me, so I'll merge it right after I have a fix

davidpanderson:
I may have just fixed it

AenBleidd:
Ok, let's wait for the build results

AenBleidd:
But I believe the issue (also for Mac) is because we need to enable c++11 build by default

AenBleidd (approved):

CharlieFenton:
> But I believe the issue (also for Mac) is because we need to enable c++11 build by default

I believe  c++11 is already the default on current versions of the Mac compiler

PR #4252, opened 2021-03-09
[Linux] Openclapp in CI

talregev:
Add opencl check to apps

PR #4261, opened 2021-03-12
[WIP] Compile boinc with dynamic libs in ci

talregev:
Dynamic in ci
Investigating issue in Ubuntu and Debian in installation process #4252

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Configure dynamic
         if: ${{ success() && matrix.type == 'client-with-dynamic-lib' }}
-        run: ./configure --enable-apps --enable-client --disable-manager --enable-dynamic-client-linkage  --disable-fcgi --enable-libraries --enable-unicode --enable-shared --enable-dynamic-client-linkage --with-ssl
+        run: ./configure --disable-apps --enable-client --disable-manager --enable-dynamic-client-linkage  --disable-fcgi --enable-libraries --enable-unicode --enable-shared --enable-dynamic-client-linkage --with-ssl
 
       - name: Configure libs
         if: ${{ success() && matrix.type == 'libs' }}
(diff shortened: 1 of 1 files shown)


LocutusOfBorg:
It already failed :)

```
./configure: line 14691: : supported targets:.* elf: command not found
checking whether the gcc linker (/usr/bin/ld -m elf_x86_64) supports shared libraries... no
checking dynamic linker characteristics... GNU/Linux ld.so
checking how to hardcode library paths into programs... unsupported
```

LocutusOfBorg:
Now just wait for the build and you will see that its happily just building static libraries and not shared ones

talregev:
@LocutusOfBorg Do you mind to talk over discord channel? (app or website).

talregev:
@LocutusOfBorg 
I will write here my thoughts:
1. First thing, the problem you show me at opencl PR with your logs from Ubuntu/Debian installation process, doesn't seems to connect to share or statics libs. I think it connected to autoconf tool, sometimes doing problems when switching between configs. When it happen, I usually delete 2 makefiles. one in `boinc` directory, and another in `m4` directory. 
2. When you compile apps, in some apps it doesn't matter if it dynamic or static, the makefile will always compile with the static library (*.a) because the makefile config to do.
3. Even when the boinc creates a static library instead of shared ones (when we put dynamic flags in the config), it not connected to my PR. If it does, it an old problem. So undo my PR will not solve static libs instead of shared ones. (maybe it create both?).

So anyway I think your problem is different, and we taking it to the wrong direction, that why I wanted to talk over discord. 
If you really want to solve your problem on your side, please consider my thoughts.
Also it more convenient to write on my PR, because if I want to do tests, or change things, I can change here.
I prefer to talk over discord, but we can stay here if you prefer so.

talregev:
I propose a fix: #4268.
I am closing this PR

PR #4263, opened 2021-03-14
[Linux] Add rappture app to ci

talregev:
Add rappture app to ci


talregev:
Switch to PR #4290
Compile wrappture for linux and android

PR #4270, opened 2021-03-16
Check for supports shared libraries

talregev:
Check for supports shared libraries.


talregev:
I want to add a a check, when the configure will say:
checking whether the gcc linker (/usr/bin/ld -m elf_x86_64) supports shared libraries... no
The CI will fail. Can you help?
unknown 1

talregev:
This PR is failing because we have the bug that gcc don't supports shared library. 
I fix the bug in other PR. 
This PR should merge only after my fix.

AenBleidd (approved):

PR #4272, opened 2021-03-17
configure.ac, boinc-client and boinc-manager:

LocutusOfBorg:
make the check for new version something configurable via autoconf via switch:
--disable-new-version-check to disable it (something for package maintainers)
--enable-new-version-check to enable it (default)

This way, boinc into package repository won't try to search for new version, avoiding to
call home.

Also, include config.h into version.h, otherwise the new ENABLE_NEW_VERSION_CHECK define
is not taken into account from clientgui graphical manager.

### DIFF
--- a/clientgui/AdvancedFrame.cpp
+++ b/clientgui/AdvancedFrame.cpp
@@ -195,7 +195,9 @@ BEGIN_EVENT_TABLE (CAdvancedFrame, CBOINCBaseFrame)
     EVT_MENU(ID_HELPBOINCMANAGER, CAdvancedFrame::OnHelpBOINC)
     EVT_MENU(ID_HELPBOINCWEBSITE, CAdvancedFrame::OnHelpBOINC)
     EVT_MENU(wxID_ABOUT, CAdvancedFrame::OnHelpAbout)
+#ifdef ENABLE_NEW_VERSION_CHECK
     EVT_MENU(ID_CHECK_VERSION, CAdvancedFrame::OnCheckVersion)
+#endif
     EVT_MENU(ID_REPORT_BUG, CAdvancedFrame::OnReportBug)
     EVT_HELP(wxID_ANY, CAdvancedFrame::OnHelp)
     // Custom Events & Timers
@@ -684,6 +686,7 @@ bool CAdvancedFrame::CreateMenus() {
     );
     menuHelp->AppendSeparator();
 
+#ifdef ENABLE_NEW_VERSION_CHECK
     strMenuName.Printf(
         _("Check for new %s version"),
         pSkinAdvanced->GetApplicationShortName().c_str()
@@ -698,6 +701,7 @@ bool CAdvancedFrame::CreateMenus() {
         strMenuDescription
     );
     menuHelp->AppendSeparator();
+#endif
 
     menuHelp->Append(
         ID_REPORT_BUG,
--- a/version.h
+++ b/version.h
@@ -1,5 +1,6 @@
 /* Platform independent version definitions... */
 
+#include "config.h"
 #ifndef BOINC_VERSION_H
 #define BOINC_VERSION_H
 
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -647,7 +647,9 @@ int CLIENT_STATE::init() {
     // NOTE: this must be called AFTER
     // read_vc_config_file()
     //
+#ifdef ENABLE_NEW_VERSION_CHECK
     newer_version_startup_check();
+#endif
 
     // parse account files again,
     // now that we know the host's venue on each project
--- a/client/scheduler_op.cpp
+++ b/client/scheduler_op.cpp
@@ -134,7 +134,9 @@ int SCHEDULER_OP::init_op_project(PROJECT* p, int r) {
         // and project list
         //
         if (!cc_config.no_info_fetch) {
+#ifdef ENABLE_NEW_VERSION_CHECK
             gstate.new_version_check();
+#endif
             gstate.all_projects_list_check();
         }
     }
--- a/clientgui/sg_BoincSimpleFrame.cpp
+++ b/clientgui/sg_BoincSimpleFrame.cpp
@@ -91,7 +91,9 @@ BEGIN_EVENT_TABLE(CSimpleFrame, CBOINCBaseFrame)
     EVT_MENU(ID_HELPBOINCMANAGER, CSimpleFrame::OnHelpBOINC)
     EVT_MENU(ID_HELPBOINCWEBSITE, CSimpleFrame::OnHelpBOINC)
     EVT_MENU(wxID_ABOUT, CSimpleFrame::OnHelpAbout)
+#ifdef ENABLE_NEW_VERSION_CHECK
     EVT_MENU(ID_CHECK_VERSION, CSimpleFrame::OnCheckVersion)
+#endif
     EVT_MENU(ID_REPORT_BUG, CSimpleFrame::OnReportBug)
     EVT_MENU(ID_EVENTLOG, CSimpleFrame::OnEventLog)
     EVT_MOVE(CSimpleFrame::OnMove)
@@ -318,6 +320,7 @@ bool CSimpleFrame::CreateMenus() {
     );
     menuHelp->AppendSeparator();
 
+#ifdef ENABLE_NEW_VERSION_CHECK
     strMenuName.Printf(
         _("Check for new %s version"),
         pSkinAdvanced->GetApplicationShortName().c_str()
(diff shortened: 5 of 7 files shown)


CharlieFenton:
@LocutusOfBorg As you have this written, Mac and Windows builds will have version check disabled because `ENABLE_NEW_VERSION_CHECK` will be undefined, since Mac and Windows builds do not use the configure / build scripts.

I suggest something like this instead:
`#ifndef DISABLE_NEW_VERSION_CHECK
     newer_version_startup_check();
 #endif
`

davidpanderson:
Why not have this selected by e.g. a config file option?

clientgui/AdvancedFrame.cpp
```diff
@@ -195,7 +195,9 @@ BEGIN_EVENT_TABLE (CAdvancedFrame, CBOINCBaseFrame)
     EVT_MENU(ID_HELPBOINCMANAGER, CAdvancedFrame::OnHelpBOINC)
     EVT_MENU(ID_HELPBOINCWEBSITE, CAdvancedFrame::OnHelpBOINC)
     EVT_MENU(wxID_ABOUT, CAdvancedFrame::OnHelpAbout)
+#ifdef ENABLE_NEW_VERSION_CHECK
```
AenBleidd, line 198: As I told you before, there are two additional OSs we support: Windows and OSX. If you build BOINC Manager for linux to control BOINC client for linux - this works perfect. But if you decide to control BOINC client for Windows or OSX with BOINC Manager for Linux - then you'll have no way to check for a new version of BOINC client for that particular OS. The same if applicable for BOINC Manager for Windows or OSX, I'll have an option to check for a new version that will do basically nothing because this functionality just dropped from the BOINC client for linux.

version.h
```diff
@@ -1,5 +1,6 @@
 /* Platform independent version definitions... */
 
+#include "config.h"
```
AenBleidd, line 3: ' version.h' file could be installed.
'config.h' file must not be installed so this include will just break installed haders

AenBleidd (changes requested):

CharlieFenton:
> Why not have this selected by e.g. a config file option?

I believe that is how @LocutusOfBorg has implemented it:
> make the check for new version something configurable via autoconf via switch

But neither the Mac nor Windows builds of client and Manager use configure / make, which use custom, prebuilt config.h files. Since this change is intended only for Linux, it should be done in such a way that requires no modifications to these prebuilt config.h files or the corresponding build files (namely boinc.xcodeproj for Mac; boinc*.sln and boinc*.vcxproj for Windows.) That is why I suggested [this](https://github.com/BOINC/boinc/pull/4272#issuecomment-800751660).

AenBleidd:
@CharlieFenton, your suggestion will not work in case when Manager on one platform manages remote client on different platform

talregev:
> @CharlieFenton, your suggestion will not work in case when Manager on one platform manages remote client on different platform

@AenBleidd I believe if you compile by design a client with a not update flags, this client will not able to check update. Mangers from any platforms shouldn't able to check the update for this client. 

Any suggestion on this case?

AenBleidd:
> Mangers from any platforms shouldn't able to check the update for this client.

If you take a look at the suggested code, this compilation flag removes menu from Manager. So if I use linux Manager with disabled updated to check for updates of Windows client, I'll have no option to do this manually.
Yes, there is a workaround to restart the client, but it's not what users expects to achieve

talregev:
> > Mangers from any platforms shouldn't able to check the update for this client.
> 
> If you take a look at the suggested code, this compilation flag removes menu from Manager. So if I use linux Manager with disabled updated to check for updates of Windows client, I'll have no option to do this manually.
> Yes, there is a workaround to restart the client, but it's not what users expects to achieve

I agree, I didn't think about this case. In this case, what we want to happen?

AenBleidd:
> Any suggestion on this case?

Use suggestion of David and make this configurable option in cc_config.xml?

AenBleidd:
> I agree, I didn't think about this case. In this case, what we want to happen?

Manager always asks client for the update. Manager doesn't do any kind of update because it could be located on a different machine and have another version (even the latest one but client could be older and could require update).

talregev:
> > I agree, I didn't think about this case. In this case, what we want to happen?
> 
> Manager always asks client for the update. Manager doesn't do any kind of update because it could be located on a different machine and have another version (even the latest one but client could be older and could require update).

So how the manager itself get update?

AenBleidd:
> So how the manager itself get update?

When local client receives an update.
Yes, it's not the best solution but it's better than duplicating version check functionality implemented in client to Manager.
In future it could be redesigned but for now it's good enough

talregev:
> > So how the manager itself get update?
> 
> When local client receives an update.
> Yes, it's not the best solution but it's better than duplicating version check functionality implemented in client to Manager.
> In future it could be redesigned but for now it's good enough

Just for curiosity, 
When the manager doesn't have a local client, and only remote other clients. 
It mean that he cannot be updated?

RichardHaselgrove:
> So how the manager itself get update?

Neither the Client or the Manager ever 'gets an update' - that's been a core principle ever since the beginning.

What it get is 'information about an update' (which should be relevant to the platform issuing the request, where the cross-platform issue comes into play). It's up to the computer user or their system administrator to decide how and when to act on that information.

AenBleidd:
@talregev, as @RichardHaselgrove correctly told, there is no auto-update feature or similar to this. It's just an information that there is an update for client on the platform where client is running.

talregev:
@AenBleidd, @RichardHaselgrove Thank you both. Now I understand it better.

CharlieFenton:
> Use suggestion of David and make this configurable option in cc_config.xml?

Thank you for that clarification. I misunderstood the comment from @davidpanderson; I thought he was referring to the configure file created by autoconf. Making it an option in the cc_config.xml file would make this a run-time decision rather than compile time. I am OK with doing this, but only if the default is the current behavior (enable version check) when the flag is not present in cc_config.xml, for backward compatibility.

> @CharlieFenton, your suggestion will not work in case when Manager on one platform manages remote client on different platform

@AenBleidd Thank you for noticing this; you are correct. And of course a corresponding problem would exist with using the original `#ifdef ENABLE_NEW_VERSION_CHECK` method.

 But please note that for a cc_config.xml flag to accomplish this, the Manager must get the remote client's cc_config data using the `read_cc_config` RPC when the user invokes the _Select Computer_ command in the Manager. I took a look at the code and did not immediately find logic which does this, though I may have missed it.

RichardHaselgrove:
> But please note that for a cc_config.xml flag to accomplish this, the Manager must get the remote client's cc_config data using the read_cc_config RPC when the user invokes the Select Computer command in the Manager. I took a look at the code and did not immediately find logic which does this, though I may have missed it.

I use a Windows Manager to monitor and control a remote Linux client: it appears to me that this already happens when, for example, changing the set of Event Log flags while the client is running.

But I appreciate your caution and warning that this must be properly enabled if the cc_config.xml proposal is taken up.

CharlieFenton:
> the Manager must get the remote client's cc_config data using the `read_cc_config` RPC when the user invokes the _Select Computer_ command in the Manager.

Correction: it only needs to call the `get_cc_config` RPC. But see below for an alternate solution.

> it appears to me that this already happens when, for example, changing the set of Event Log flags while the client is running.

Yes, I believe the _Event Log options_ command will force a new `get_cc_config` RPC. Sorry I was not more clear. I meant that it needs to happen without any further user action when connecting to a remote client using the _Select computer_ command. It's not clear to me that is currently the case.

On the other hand, the only items in the cc_config.xml file which currently affect the Manager appear to be `disallow_attach`, `simple_gui_only` and `max_event_log_lines.` These 3 items are passed to the Manager by the `get_cc_status` RPC, **which is called** when connecting to a remote client using the _Select computer_ command. So a simpler fix would be to add the value of the new `cc_config.disable_version_check` option to the `get_cc_status` RPC, being careful not to cause issues with backward compatibility.

But looking at the code, it appears to me there may be a possible bug in the _Read config files_ command. It updates the information in the client, but it looks like it may not get passed to the Manager. This could be a problem if, for example, you manually edit the cc_config.xml file and then want the client and Manager to have the updated options flags. However, the Manager normally issues a `get_cc_status` RPC once per second, so this is probably only a minor concern.

I will add one more thought. Using the `get_cc_status` RPC to pass those 3 cc_config items to the Manager appears to be a workaround for not calling the `get_cc_config ` RPC in some situations. This should serve as a reminder of the complexity of the interplay between the client and Manager and as a caution to be very careful when making changes in that area.

LocutusOfBorg:
Hello, I am not too confident to implement this feature, the risk of breaking the current software is too high...

talregev:
@LocutusOfBorg Why you afraid of anything? Here we have a group of people that expert in boinc, and will not approve your changes if you break something. Please don't afraid to try if you have time.

AenBleidd:
@talregev, this PR will not work using selected approach. It should be discussed and properly designed to implement needed changes without unwanted side effects

LocutusOfBorg:
new pr #4616 is now opened

PR #4275, opened 2021-03-19
client: Fix invalid XML tags in handle_set_app_config

chausner:
Fixes invalid XML in the RPC response when an error occurs during the `set_app_config` RPC.


AenBleidd (approved):

AenBleidd:
LGTM. Thanks for the fix!

PR #4278, opened 2021-03-22
[Codecov] Fix codecov.yml to wait for all reports before reporting st…

AenBleidd:
…atus

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,8 @@
 codecov:
+  require_ci_to_pass: yes
   notify:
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -16,6 +18,7 @@ comment:
   require_base: no        # [yes :: must have a base report to post]
   require_head: yes       # [yes :: must have a head report to post]
   branch: master
```
thomasrockhu, line 20: This should be

```
  branches:
    - master
```

PR #4291, opened 2021-03-29
Client: fix app_config write from RPC

RichardHaselgrove:
In May 2017, a new RPC was added for reading and writing remote app_config.xml files (commit 5b6f648570cdef551b77bbabf3e69243dac69de4, no associated PR). The file written contained unparsable XML with unrecognised tags. This patch restores compliance with the existing documentation at

https://boinc.berkeley.edu/wiki/Client_configuration#Project-level_configuration
https://scienceunited.org/trac/wiki/ClientAppConfig

Fixes 5b6f648570cdef551b77bbabf3e69243dac69de4

Bug re-discovered during Beta testing of covid-19 app at World Community Grid.

**Release Notes**
Allow edit of app_config.xml files by BoincTasks

### DIFF
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -1195,7 +1195,6 @@ static void handle_get_app_config(GUI_RPC_CONN& grc) {
         return;
     }
     sprintf(path, "%s/%s", p->project_dir(), APP_CONFIG_FILE_NAME);
-    // printf("path: %s\n", path);    // debug only
     int retval = read_file_string(path, s);
     if (retval) {
         grc.mfout.printf("<error>app_config.xml not found</error>\n");
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -1195,7 +1195,7 @@ static void handle_get_app_config(GUI_RPC_CONN& grc) {
         return;
     }
     sprintf(path, "%s/%s", p->project_dir(), APP_CONFIG_FILE_NAME);
-    printf("path: %s\n", path);
+    // printf("path: %s\n", path);    // debug only
     int retval = read_file_string(path, s);
     if (retval) {
         grc.mfout.printf("<error>app_config.xml not found</error>\n");
--- a/lib/cc_config.cpp
+++ b/lib/cc_config.cpp
@@ -864,19 +864,21 @@ int APP_CONFIGS::parse_file(FILE* f, MSG_VEC& mv, LOG_FLAGS& log_flags) {
 
 void APP_CONFIGS::write(MIOFILE& out) {
     out.printf(
-        "   <app_config>\n"
+        "<app_config>\n"
     );
     for (unsigned int i=0; i<app_configs.size(); i++) {
         APP_CONFIG& ac = app_configs[i];
         out.printf(
-            "       <app>\n"
-            "           <name>%s</name>\n"
-            "           <max_concurrent>%d</max_concurrent>\n"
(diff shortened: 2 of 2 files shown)


client/gui_rpc_server_ops.cpp
```diff
@@ -1195,7 +1195,7 @@ static void handle_get_app_config(GUI_RPC_CONN& grc) {
         return;
     }
     sprintf(path, "%s/%s", p->project_dir(), APP_CONFIG_FILE_NAME);
-    printf("path: %s\n", path);
+    // printf("path: %s\n", path);    // debug only
```
AenBleidd, line 1198: Could be completely removed

AenBleidd (changes requested):

client/gui_rpc_server_ops.cpp
RichardHaselgrove: Indeed it could, but I don't like deleting other people's code until they've had a chance to consider why they wrote it (and left it in).

24 hours without comment, and it'll be gone.

AenBleidd: It's definitely a debug left-over, and we can wait for 24 hours :)

Uplinger: I agree with removing the line, git will remember the change

RichardHaselgrove:
While we're here, does any boinc_cmd user have a comment on

https://github.com/BOINC/boinc/blob/master/client/boinc_cmd.cpp#L678 ff:
```
        // for testing purposes only
        //
        APP_CONFIGS ac;
        APP_CONFIG a;
        ac.clear();
        strcpy(a.name, "uppercase");
        a.max_concurrent = 2;
```
It's part of the original commit.

Uplinger:
@RichardHaselgrove I don't see anything about your commit that gives me pause to it.  Seems like it matches exactly what the specification was that was written years ago also.  I'm fine with accepting this change.

Uplinger (changes requested):
I will approve after commented line is removed...

client/gui_rpc_server_ops.cpp
RichardHaselgrove: Line removed as requested.

RichardHaselgrove:
@Uplinger - line removed as requested. Ready for final review.

RichardHaselgrove:
Still ToDo (1) - but in another issue.

I've looked into boinccmd. 
`--get_app_config URL               show app config for given project`
is documented, and works as expected. The corresponding
`--set_app_config URL`
is **not** documented, and remains skeletal. It can command the client to write out an app_config.xml file, but only one containing the hard-coded parameters. I made one like this:
```
<app_config>
    <app>
        <name>dummy_app_name</name>
        <max_concurrent>0</max_concurrent>
        <gpu_versions>
            <gpu_usage>0.000000</gpu_usage>
            <cpu_usage>0.000000</cpu_usage>
        </gpu_versions>
        <fraction_done_exact>0</fraction_done_exact>
        <report_results_immediately>0</report_results_immediately>
    </app>
    <project_max_concurrent>0</project_max_concurrent>
    <report_results_immediately>0</report_results_immediately>
</app_config>
```
which generated
```
30-Mar-2021 13:14:42 [climateprediction.net] gpu_usage must be positive in app_config.xml
30-Mar-2021 13:14:42 [climateprediction.net] Your app_config.xml file refers to an unknown application 'uppercase'.  Known applications: None
```
It's a template, I suppose, but without the `<app_version>` syntax, and without any user data. This would need a lot of work before it's revealed to the public, and I doubt it's worth the effort.

AenBleidd (approved):

PR #4128, opened 2020-12-16
Switch to C++11

Flowdalic:
This PR is, for now, primary to check if travis would fail with C++11 and as basis for discussion. Related to https://github.com/BOINC/boinc/pull/4126#issuecomment-746829778.

PR #4304, opened 2021-03-31
Mac: Fix Mac builds to support C++11

### LINKED ISSUE #3635: On C++11 and up support
Does BOINC support C++11?

For example, i want to fix undefined behavior (which may lead to security issues) when `printf`-like functions get incompatible format specifier and actual parameter type: https://github.com/BOINC/boinc/pull/3633 In order to do that, i either need to include `<cinttypes>` (C++11 thing) for portable format specifiers, or add ugly macroses to duplicate such functionality in BOINC headers.

If we support MinGW builds, things become even more complicated. MinGW uses old msvcrt runtime, and requires special define [__USE_MINGW_ANSI_STDIO=1](https://sourceforge.net/p/mingw-w64/wiki2/gnu%20printf/) to mimic C11 behavior (`%zu` format specifier for `size_t`). This also can lead to changes in how other format specifiers work on MinGW builds.

So possible solution is to
1. Define all needed format specifier macroses manually (including special cases for MinGW) in BOINC headers. Hard to do correctly, but possible (Visual Studio added support of different format specifiers from different versions, MinGW with its own formatters, 32 vs 64 bit types, etc). 

or  

Allow C++11 in gcc builds (as it's done for Windows builds on vs2019) and 
1. Define special portable format specifiers based on `<cinttypes>` for MinGW.
or
2. Define __USE_MINGW_ANSI_STDIO=1 for MinGW builds and use standard format specifiers from `<cinttypes>`. 
or
3. Drop support of MinGW and use VsCode or Community versions of Visual Studio to compile BOINC on Windows. So cross-compiling Windows builds on Linux will not work.

What do you think?

CharlieFenton:
Reference: #3635 #4128 #4207

### DIFF
--- a/client/dhrystone.cpp
+++ b/client/dhrystone.cpp
@@ -94,7 +94,7 @@ int dhrystone(
     double                  benchtime;
     double Dhrystones_Per_Second;
 
-    register unsigned long  Run_Index;
+    REG unsigned long  Run_Index;
 
 
   Next_Ptr_Glob = (Rec_Pointer) malloc (sizeof (Rec_Type));
--- a/client/dhrystone.h
+++ b/client/dhrystone.h
@@ -1,5 +1,9 @@
 
+#ifdef __APPLE__
+#define REG
+#else
 #define REG register
+#endif
 
 typedef enum    {Ident_1, Ident_2, Ident_3, Ident_4, Ident_5} Enumeration;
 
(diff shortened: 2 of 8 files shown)


CharlieFenton:
This PR is now ready to be merged.

AenBleidd (approved):
LGTM. Thanks

PR #4307, opened 2021-04-01
[Android][Script] Changes in android ci script

talregev:
android-tc new location: -> "$cache_dir"/Android/android-tc
Set ANDROID_ARM_NEON from triplet 
Change cache vcpkg of android to $cache_dir/Android/vcpkgcache/
Add more option to clear_environemt
change BUILD_DIR -> 3rdparty/Android
Change vcpkg location aka vcpkd_root env var:  $BUILD_DIR/vcpkg


android/clear_environment.sh
```diff
@@ -10,13 +13,35 @@ if [ ! -d android ]] ; then
...
+    rm -rf $build_dir/android-ndk-*
+elif [ "vcpkg_cache" = "$1" ]; then
+    echo "vcpkg_cache clean"
+    rm -rf $cache_dir/vcpkgcache/
+elif [ "cache" = "$1" ]; then
+    echo "cache dir clean"
+    rm -rf $cache_dir
+elif [ "build" = "$1" ]; then
+    echo "build dir clean"
+    rm -rf $cache_dir
```
AenBleidd, line 33: Should be `$build_dir`

AenBleidd (changes requested):

android/clear_environment.sh
talregev: Yes, you are right. I will fix it. Thank for noticing.

AenBleidd (approved):

Issue #4204, opened 2021-02-19
[Bug report] mali GPU not detected on android boinc 7.16.16


talregev:
Add support in android gpu - mali gpu.
https://developer.arm.com/ip-products/graphics-and-multimedia/mali-gpus/mali-driver-development-kit
And Intel GPU:
https://software.intel.com/content/www/us/en/develop/tools/opencl-sdk.html

I can help to compile mali ddk on android, and I can help to compile opencl-sdk in windows and linux. 

-------------------------------------------------------------
After edit:
mali GPU not support on android boinc 7.16.16

AenBleidd:
BOINC already supports this. If you want to help projects to build their application for these GPUs - please either write on their forums or use mailing list boinc_projects@ssl.berkeley.edu

talregev:
If it already support it don't show any indication that it did. So even If a project have a binary that support it, it has no way to know. 
So I should report it as a bug.

AenBleidd:
Do you have an example when BOINC client doesn't recognize any GPU?

talregev:
Yes. I have many.

Ageless93:
Mine is recognized, and BOINC asks for work for it on a regular basis from all projects, but since none have an application for the Mali GPU no work is gotten. I'm not so sure I even want my Mali GPU to be used without my phone having active cooling, because the GPU gets quite hot when in use (by for example my navigation app, it has overheated my phone already). 

https://i.ibb.co/s1pgbn3/2.jpg

talregev:
I have Mali-G32. Did you install some driver?

AenBleidd:
> I'm not so sure I even want my Mali GPU to be used without my phone having active cooling

You just need a tank with liquid nitrogen where you can put your phone overnight ;)

Ageless93:
I did not install any driver, I have several devices with Mali GPUs that are detected by 7.4.53 without a problem. 
As for cooling, all my other Android devices are actively cooled. They lay on top of a cupboard in my living room with a 200mm case fan at 5V blowing cool air over them. 

That's a luxury I don't have in my car.

RichardHaselgrove:
I'm sure my Samsung tablet had a Mali GPU when I bought it: in fact, it's still visible on my SETI account page https://setiathome.berkeley.edu/show_host_detail.php?hostid=8865273

> Coprocessors	OpenCL GPU Mali-G71 (1853MB) driver: 2.00

[...]

PR #4315, opened 2021-04-08
draft: android: set LD_LIBRARY_PATH for client for OpenCL detection

truboxl:
Draft fix for #4204 

**Description of the Change**
Before:
Screenshot_20210401-020446_BOINC

After:
Screenshot_20210401-021313_BOINC

**Alternate Designs**
Vendor libs are isolated to a different namespace by the Android linker.
Any other proper solutions, which is properly implement a way to access vendor libs for BOINC use,
will be better than this stopgap.
Though likely need to bump minimum API version to 21 to use `android_dlopen_ext()` inside client.

Read more about the cause:
https://developer.android.com/about/versions/nougat/android-7.0-changes#ndk
https://fadeevab.com/android-linker-namespace-security-flaws/

**Release Notes**
android: restore OpenCL detection

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.kt
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/client/Monitor.kt
@@ -537,7 +537,17 @@ class Monitor : LifecycleService() {
             val param = if (remote) "--allow_remote_gui_rpc" else "--gui_rpc_unix_domain"
             val cmd = arrayOf(boincWorkingDir + fileNameClient, "--daemon", param)
             if (Logging.ERROR) Log.w(Logging.TAG, "Launching '${cmd[0]}' from '$boincWorkingDir'")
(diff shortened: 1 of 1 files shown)


talregev:
I PR you a suggested code. I try to do it here, but it not let me.

talregev:
I also test it on my other device. (with my suggested code):
This is a h96 max x3 device with android 9 with mali gpu Mali-G31. armv7 os.
It say operation is not permitted. Do you have a clue? 
image

talregev:
> Actually there's no point setting `LD_LIBRARY_PATH` in older Android release as the older release don't isolate the libs like the recent ones and pretty much just adds a little overhead. I guess I can merge yours since you tested.
> But ultimately I don't think this hack is the correct way. There's a reason Google blocked this behaviour by default. Maybe look into using `android_dlopen_ext()` in the future...

What is the correct way? 
I also think this code is not idle, but it better from what we have.
On all my device it didn't succeeded to recognize even one GPU. 
But if for you it recognize, it improvement.

PR #4316, opened 2021-04-09
[Linux][vcpkg] Compile Linux with vcpkg

talregev:
Compile Linux with vcpkg

boinc client running with the latest libcurl and libssl:
unknown


AenBleidd (approved):
LGTM

PR #4317, opened 2021-04-10
[Android] Upgrade Kotlin backend to 1.4.30 and enable IR compiler

AenBleidd:


talregev (approved):
Compile, run and crunch projects on my oneplus one.

Issue #4327, opened 2021-04-13
Non-BOINC CPU usage time is incorrect


Flowdalic:
BOINC's calculation of non-BOINC CPU usage time is incorrect, as processes which exited within the sample interval (typically 10 seconds) are not taken into consideration. This leads to BOINC underestimating the non-BOINC CPU usage, especially in situations where a lot of short-lived processes are present.

We run BOINC as SCHED_IDLE batch process on our Continuous-Integration (CI) nodes and configured BOINC to suspend once more than 10% CPU time is consumed by other applications. But since short-lived processes are very common when compiling (C/C++) projects, BOINC is constantly underestimating the non-BOINC CPU usage. As result, it often kicks in, while a CI jobs is actually fully utilizing the system. Needles to say, that this behavior is undesirable.

I saw that this shortcoming was already identified in https://github.com/BOINC/boinc/pull/3939. But this PR seems to be stalled. Hence, besides this issue acting as tracking issue for the deficiency, please feel encouraged to continue work on PR #3939 by it. :)

PR #4328, opened 2021-04-13
client: be conservative about non-BOINC CPU usage on underflow

Flowdalic:
Instead of assume that the non-BOINC CPU usage was 0%, assume that it
was 50%. This potentially prevents BOINC from kicking-in in the
presence of short-lived processes (if configured to run only if
non-BOINC CPU usage is below 50%).

The main issue is that BOINC's calculation of non-BOINC CPU usage time
is incorrect, as processes which exited within the sample
interval (typically 10 seconds) are not taken into consideration. This
leads to BOINC underestimating the non-BOINC CPU usage, especially in
situations where a lot of short-lived processes are present. This
commit does not fix that, but mitigates the effects of this. For more
information, see issue #4327.

### DIFF
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -505,9 +505,10 @@ void ACTIVE_TASK_SET::get_memory_usage() {
     double new_cpu_time = pi.user_time + pi.kernel_time;
     if (!first) {
         non_boinc_cpu_usage = (new_cpu_time - last_cpu_time)/(diff*gstate.host_info.p_ncpus);
-        // processes might have exited in the last 10 sec,
-        // causing this to be negative.
-        if (non_boinc_cpu_usage < 0) non_boinc_cpu_usage = 0;
+        // Processes might have exited in the last 10 sec, causing
+        // this to be negative. In that case, be conservative about
+        // the non-BOINC CPU usage and assume it to be 50%.
+        if (non_boinc_cpu_usage < 0) non_boinc_cpu_usage = 0.5;
         if (log_flags.mem_usage_debug) {
             msg_printf(NULL, MSG_INFO,
                 "[mem_usage] non-BOINC CPU usage: %.2f%%", non_boinc_cpu_usage*100


AenBleidd:
I'm afraid such behavior will lead to many false positives that will constantly interrupt running tasks that usually is not good.

Flowdalic:
You are in principle right. However, I am not sure about the "constantly" part. Looking at BOINC's debug output on our nodes, where we ran BOINC as SCHED_IDLE task together with CI jobs, it appears the current value of 0 causes BOINC frequently to start running, when in fact, the system is fully utilized. So it is a judgement call which erronous  behavior is favorable. And I'd argue the one which favors BOINC not running is. Otherwise we have to consider turn of BOINC completely on our nodes, as it does harm our CI jobs.

Flowdalic:
Friendly reminder. Only using this PR in combination with #4336 makes it possible to run BOINC on our CI infrastructure. If those are not applied, then BOINC frequently reports an incorrect low utilization by non-BOINC processes (see #4327). This eventually causes BOINC to start running and impacts our CI jobs to a high degree (even tough the BOINC jobs run with `SCHED_IDLE`). Which makes it impossible to donate our spare CPU cycles to BOINC. :(

I strongly suggest to applying this workaround until the underlying issue is fixed. But given that the fixing attempt is non-trivial, and a incomplete PR which tries to fix it, #3939, did not make any progress in over a year, this seems like a viable interims solution.

Flowdalic:
@AenBleidd Another friendly reminder. Anything I can do to get this merged? Probably make it configurable and default to the current behavior? I'd really like to continue running BOINC in our lab network, but I currently have to maintain a patched version of the debian package containing the patches of #4336 and #4328, to have my co-workers not kill BOINC as background job.

AenBleidd:
@Flowdalicm, TBH, I don't like workarounds. It's better to fix the original method of non-cpu usage calculation (that was started in #3939) than patching with a temporary solution.
@davidpanderson, could you please take a look at this?

Flowdalic:
> TBH, I don't like workarounds. It's better to fix the original method of non-cpu usage calculation (that was started in #3939) than patching with a temporary solution.

I totally agree, but given that #3939 hasn't seen any progress in over a year I don't have high hopes that it will land anytime sone. Meanwhile the workaround was available, solves a real world problem¹, and had even a chance to land in Debian Bullseye, which it now missed. But it's not to late to get included in a Debian backports release. And as soon as the right solution lands, the commits on the workaround can (ideally) simply be reverted.

1: In my experience, BOINC's current behavior is way to aggressive to be run as low priority background batch job. I really want to donate CPU time, especially in those COVID times, but with the defaults, even running BOINC as SCHED_IDLE and which a recent enough Linux Kernel where [SCHED_IDLE does actually something sensible](https://lwn.net/Articles/805317/), BOINC had a noticeable impact on our lab- and CI systems. Only with those two PRs applied BOINC will get out of the way *and* stay out of the way, as soon as any of our higher-priority batch JOBs kicks in. Just relying on SCHED_IDLE is sadly not enough for us, as some of our systems have not enough memory to keep BOINC suspended during other activity (this is also because our other jobs are also memory intensive, leading to swap storms of systemd-oomd kills).

Flowdalic:
Friendly reminder about this. The current method used by BOINC to calculate the non-BOINC CPU usage is just broken. Apparently there are no resources for a proper fix in short term, so please consider merging this PR to increase the acceptance of BOINC side by side to other CPU intensive applications.

AenBleidd:
I still don't like this solution.
I'll plan this issue on the next iteration to solve it properly.

Flowdalic:
> I still don't like this solution.

That's fine, but it is way better than what we currently have. Right now, if `non_boinc_cpu_usage` is negative, then BOINC simply assumes that the `non_boinc_cpu_usage` is 0. But if `non_boinc_cpu_usage` is negative, then all bets are off: You can not make any assumption in this case. But changing the assumption to 50% non-BOINC CPU usage is still better then assuming 0% non-BOINC CPU usage. Assuming 0% means that BOINC will probably start crunching, using CPU, because the user configured it to do so if there is no other process using the CPU. But assumption that there is no other process using the CPU is wrong, especially if the other process is a build job with many short lived processes. Then we see a lot of negative values for `non_boinc_cpu_usage`. Assuming the non-BOINC CPU usage is 50% instead gives the user the option to set the threshold when BOINC should start working above 50% or below. If it is below 50%, then BOINC will occasionally not start when it could (i.e., when the assumption `non_boinc_cpu_usage < 0` → 50% non-BOINC CPU usage was wrong). When then user sets the threshold above 50%, then the same behavior is achieved as if the assumption is 0%. So the proposed change in this PR really does not make things worse for anyone, it only givers users like me to set a reasonable threshold to not affect other jobs on the same system BOINC runs on.

> I'll plan this issue on the next iteration to solve it properly.

And I love a proper solution. But I fear that BOINC in the next Debian stable release will still not ship with any solution at all if this doesn't get merged. Which means I either have to continue building my own BOINC packages or I just stop all BOINC services. And merging this does not prevent you in any way to work on a proper solution.

I sincerely hope that you reconsider merging this.

Flowdalic:
FWIW, we shut down BOINC on our Cluster, where BOINC was running as a SCHED_IDLE batch job if not other jobs where active, basically due to this being not merged and/or otherwise resolved.

davidpanderson:
We measure CPU usage by adding up the usage of current  processes.
This works OK for BOINC processes, which are long-running.
It doesn't work well for other processes, which may come and go.

Instead, we should be using system APIs for getting total CPU usage.
For Win this is GetSystemTimes(); for Linux it's /proc/stat.
I'm sure there's something similar for Mac.
I'll look into this.

davidpanderson:
This is fixed in #4859

PR #4339, opened 2021-04-21
Update all_projects_list.xml - add iThena project

CyberCF:


AenBleidd:
@CyberCF, it doesn't work like this. If you're an owner of this project - write an email to @davidpanderson https://boinc.berkeley.edu/anderson/ and ask him to include your project to the list.
If you're not an owner - ask Project owner to do this instead.

Issue #3765, opened 2020-05-25
Android: Username in Android client may not be the same as used on the forums


Ageless93:
This is with 7.4.53, but may be of importance for those who rewrite the client software for BOINC on Android. 

Over the past week and a half I was trying to get one of my devices attached to Einstein. 
I would fill in the email address I used, and the username as I use it on their forums and website (Jord), password... and all I would get back was "Communication failed". No explanation, nothing. 

In the past I changed my username from Ageless to Jord on most projects, and for all other projects this works fine. Just not for Einstein. Because apparently it holds an internal unique username (maybe because they run Drupal?). So I could add Einstein using my email address, username Ageless and my password. 

I then reported this on their forums. They've of course asked me why I need to give up my username, something I have been wondering about myself for a long time. Maybe someone here can answer that. 

Oliver went and changed my internal username from Ageless to Jord. 
I detached from Einstein and tried reattaching using my email address, username Jord and password. This failed with "Communication failed". I then tried to attach using Ageless as username. That also failed with the same message. 

... but when you look on the Einstein site, your username is in all CAPS. 
So as a last resort I tried email address, username JORD and password and that worked. 

Which makes it interesting for the rewrite. Do you add a check for all CAPS for this project, or do you automate that the username for Einstein is all CAPS? Or do you do away with the requirement for using the username on existing accounts (because they're not used in the WebRPC anyway)?

brevilo:
Let me add a few findings. While Jord certainly just describes what he experienced, I can only say that there must be something else at play here. As far as I can see the following is true:
* Drupal is totally unrelated here as it just passes the RPC calls to BOINC's web API
* Neither [lookup_account](https://boinc.berkeley.edu/trac/wiki/WebRpc#lookup_account), nor [create_account](https://boinc.berkeley.edu/trac/wiki/WebRpc#create_account) make use of the user name to attach to project with an *existing* account. That's solely done by email address and password only.

What are we missing?

If anything, the error/code message returned by the server should be available to the user somewhere. If not in the direct error message, then at least in the log.

Ageless93:
These are all the messages it wrote. Android logs are notoriously inadequate. I removed four lines about Rosetta tasks that were 6 days past the deadline. 

Mon May 25 17:55:26 GMT+02:00 2020|Einstein@Home|New computer location: home
Mon May 25 17:55:26 GMT+02:00 2020|Einstein@Home|Scheduler request completed: got 1 new tasks
[...]

Issue #4196, opened 2021-02-16
Android BOINC 7.16.16 adding Einstein now requires username in all lower chars


Ageless93:
I was adding Einstein, but each time I gave my password BOINC would end with the message "There have been problems attaching scientific projects: Einstein: Could not connect."
Which is not very useful to tell me what went wrong. So I tested my password via the website, could log in fine there with that password. But in my account profile for Einstein my name shows as JORD (all caps). Changing my name in the app to all caps didn't work. Changing my name to jord in all lower case in the app did work. 

This is fundamentally different from the earlier #3765 where I reported that my name required to be in all caps. 
Adding World Community Grid requires my name again to be Jord with a capitalized J. It refused to add the project with jord, all lower case. This is something we cannot explain to people.

RichardHaselgrove:
Couldn't attach to World Community Grid, either - with lower case email address, and either lower or normal case user name.

New account, successfully added Linux (before Android) and Windows (after) clients. Other two OS show 'Terms of Use' before attaching, Android doesn't. Only line in Event Log (detail level 5) is

`Fetching configuration file from http://www.worldcommunitygrid.org/get_project_config.php`

Process:
Open 'attach' wizard
Select WCG
Read project information blurb
Click 'Continue'
**Separately**, select checkbox for project
Click 'Continue' on main wizard page
Supply email, username, password (NB - no question whether 'new' or 'returning' user)
Read hint 1 of 3, click continue
GUI reports

```
There have been problems connecting projects:
World Community Grid
Could not connect
```
Only controls on page are chevron against WCG project, 'Finish' button.
Clicked chevron, brought up name/password form - this time with buttons for 'Register' and 'Sign in'. Signed in, successful. Page header still says 'There have been problems...', but 'could not connect' line has cleared.

Soon after, got 'Welcome' notification from WCG.

AenBleidd:
Error looks not connected to the user name and should be fixed in #4343.
Closing this as invalid

brevilo:
FWIW, the issue title is misleading as a) the issue is mostly dealing with WCG and b) the "username" [isn't used at all](https://github.com/BOINC/boinc/issues/3765#issuecomment-633662239) to sign in a user at E@H.

AenBleidd:
Originally Einstein@home was a reason of this fix ;) WCG was added later. Sorry for misleading title.

PR #4343, opened 2021-04-24
[Android] Fix Einstein@home attach

AenBleidd:
For now Einstein@home is the only project that could return an error in error message instead of error code.
Attach behavior fixed to include this situation in flow and react accordingly

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/ProjectAttachService.kt
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/attach/ProjectAttachService.kt
@@ -359,6 +359,7 @@ class ProjectAttachService : LifecycleService() {
          * @return returns status conflict
          */
         fun lookupAndAttach(forceLookup: Boolean): Int {
+            var isForceLookup = forceLookup
             if (Logging.DEBUG) {
                 Log.d(Logging.TAG, "ProjectAttachWrapper.attach: attempting: $name")
             }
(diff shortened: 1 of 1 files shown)


AenBleidd:
Should fix #3765 and #4196 also

talregev (approved):
I test it on oneplus one device.
Add 3 projects together: WCG, rosseta, and Einstein.
all 3 projects was added after I press continue without errors.

brevilo:
> For now Einstein@home is the only project that could return an error in error message instead of error code.

Which RPC are we talking about exactly?

Thanks

AenBleidd:
@brevilo, not 100% sure, but looks like this: `<create_account/>`

brevilo:
Do you have an example for your claim? We're following the [official RPC spec](https://boinc.berkeley.edu/trac/wiki/WebRpc), including the returned XML document in case of errors (with number and message). Please note that the wiki entry on the return XML is wrong: the error message is found in a `[error_msg](https://github.com/BOINC/boinc/blob/master/html/inc/web_rpc_api.inc#L87)` tag, not `error_string`, but that's unrelated to the issue discussed here.

AenBleidd:
When this RPC is sent with the credentials of existing account, 
`error_num` is `0`,
`error_string` is `-208`
This is what returned by Einstein@home.
Definitely, it should be next:
`error_num` is `-208`
`error_string` is some meaningful message.
But other projects (e.g. Rosetta) in case of existing account just return authenticator (this is a desired behavior, IIRC).

brevilo:
> When this RPC is sent with the credentials of existing account,
> `error_num` is `0`,
> `error_string` is `-208`

I doubt that. When I send a `create_account` RPC I get this:
```
<error>
    <error_num>-208</error_num>
    <error_msg>Account creation is done through our Web site. Please register at https://einsteinathome.org/user/registration</error_msg>
</error>
```
That is perfectly valid in terms of the RPC response formatting, isn't it?

Also, as I just noted: if you're really parsing `error_string` then you're parsing the XML response wrong. The tag is called `error_msg`.

AenBleidd:
@brevilo, excuse me, you're completely right, it's an issue on our side. I'll fix that. Sorry for any inconveniences

brevilo:
No problem. Just wanted to make sure I don't miss a potential issue on our side 👍

PR #4350, opened 2021-04-29
[Test] Test curl-config for vcpkg

talregev:
Test curl-config for vcpkg:
https://github.com/microsoft/vcpkg/pull/17418


AenBleidd:
https://github.com/microsoft/vcpkg/pull/17418 is merged. This test PR is not needed anymore

PR #4354, opened 2021-04-30
GPU list: strip leading "AMD" or "NVIDIA" from model to avoid dups in list

### LINKED ISSUE #4353: Suggestion for better "Top GPU Models" page
**Describe the problem**
NVIDIA & AMD GPU models are reported twice on the **Top GPU Models** page, probably based on how the drivers chose to report the card name over time as driver versions changed, causing functionally duplicate entries in the relative GPU performance statistics page. For example:

**(1.000) NVIDIA GeForce RTX 3090
(0.921) GeForce RTX 3090
(0.871) GeForce RTX 3080
(0.867) NVIDIA GeForce RTX 3080
...**

I am creating this request from my thread [http://www.primegrid.com/forum_thread.php?id=9659](http://www.primegrid.com/forum_thread.php?id=9659), specifically referring to the **Top GPU Models** page on [http://www.primegrid.com/gpu_list.php](http://www.primegrid.com/gpu_list.php). Copy/pasting from that thread:

> ...
> On Windows, the driver comes from NVIDIA. And, all current NVIDIA drivers refer to the cards as "**NVIDIA blahblah**". Yet, the "**NVIDIA GeForce**" vs "**GeForce**" name duplication in the **Top GPU Models** page exists under the Windows columns, therefore that is not due to Linux drivers adding to the mix. Maybe the NVIDIA drivers changed sometime in the recent past to add the "NVIDIA" prefix.
> 
> In any case, the way it is currently itemized feels... ...not meaningful.
> 
> Maybe the code that generates that data that results in this page should ignore any "**NVIDIA **" (and, similarly "**AMD **") prefix in the card name (along with the trailing space!), which would then combine the perf numbers of cards more intuitively? Separating perf numbers along an arbitrary line of when a company decided to include the company name in the driver report seems non-scientific.
> 
> Unless there is a whole another reasoning behind all of this.
> 
> Tuna
> 

**Additional context**
None.

davidpanderson:
fixes #4353

### DIFF
--- a/html/user/gpu_list.php
+++ b/html/user/gpu_list.php
@@ -26,10 +26,11 @@
 
 require_once("../inc/util.inc");
 
-// strip leading AMD or NVIDIA
+// strip leading AMD, NVIDIA, etc.
+// This avoids showing the same model twice
 //
 function strip_vendor($model) {
-    foreach (array("AMD ", "NVIDIA ") as $maker) {
+    foreach (array("AMD ", "NVIDIA ", "ATI ", "Intel(R) ") as $maker) {
         $n = strlen($maker);
         if (substr($model, 0, $n) == $maker) {
             return substr($model, $n);
--- a/html/user/gpu_list.php
+++ b/html/user/gpu_list.php
@@ -26,6 +26,18 @@
 
 require_once("../inc/util.inc");
 
+// strip leading AMD or NVIDIA
+//
+function strip_vendor($model) {
+    foreach (array("AMD ", "NVIDIA ") as $maker) {
+        $n = strlen($maker);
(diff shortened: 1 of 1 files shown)


html/user/gpu_list.php
```diff
@@ -26,6 +26,18 @@
 
 require_once("../inc/util.inc");
 
+// strip leading AMD or NVIDIA
+//
+function strip_vendor($model) {
+    foreach (array("AMD ", "NVIDIA ") as $maker) {
```
AenBleidd, line 32: I'd suggest to add "ATI " and "Inter(R) " to this list

PR #4362, opened 2021-05-09
[Android] Fix an issue when mutex acquisition was checked before real…

### LINKED ISSUE #3334: [Android] 7.16.3, Another volunteer app is already running on this device
I can only start BOINC once on my Huawei P20 lite (Android 9) and look at it, all other times thereafter, whether I try to recall it from memory or start by its icon, I immediately get a popup that another volunteer app is already running on this device and that only one at a time can run. I have to force close BOINC and restart it for it to work as normal for a bit, before the pop up returns.

AenBleidd:
… acquisition happened. This led to a false positive error about another instance of BOINC found.

This fixes #3334.


talregev (approved):
I tested in oneplus one (android 10).
I open another instance while boinc is open. I try to make the:
`Another volunteer app is already running on this device` msg without success.
Also the splash is loading without exception 
Also boinc is crunching.

I approve.

PR #4363, opened 2021-05-10
[Android] Adjust logging levels

AenBleidd:


talregev (changes requested):
We should delete all the `if(Logging.DEBUG)` and similar check.
the log itself check the priority of the log, and if it fit, it print the message. 
If it not happen, we should check why it not like that.
The logger design should support that.

talregev (approved):
@AenBleidd Checked the verbose level.
I tested on oneplus one (android 10) and projects was crunching.
Also see all the debug and verbose msgs in android studio.

I approve.

PR #4369, opened 2021-05-19
Mac: fix build scripts to be compatible with both bash and zsh shells

CharlieFenton:
The default shell under Mac OS 10.15 and MacOS 11 is zsh, but on earlier versions of MacOS the default shell was bash. This PR updates the Mac build scripts to be compatible with both shells.


talregev:
Will you also consider to add vcpkg compilation for mac?
Now boinc windows, linux and android use it.

CharlieFenton:
@talregev Please understand that I retired from software development six years ago, including BOINC. I am now only fixing critical BOINC bugs specific to the Mac platform; I continue to do that as a service to the community, since no other Mac developer has stepped forward. I am doing the absolute minimum necessary to keep BOINC running on the Mac. 

I have always declined to work on supporting any build method on the Mac other than the Xcode project and build scripts that I developed. I have no objection to other people doing so, as long as no one expects me to fix problems with such builds.

I'm sorry, but I know absolutely nothing about vcpkg and have no interest in learning about it. Perhaps you can find another developer who knows the Mac.

talregev:
I understand. If I had a mac, I was doing it myself. 
I hope there is a mac dev in the crowd that will see this msg, and help boinc.

CharlieFenton:
I have finished testing this and it is ready to be merged.

PR #4376, opened 2021-05-25
Fix Xcode build for Apple Silicon M1 Macs

bema-aei:
During the BOINC build with Xcode a utility called "SetVersion" gets built and ran.

However, on a new Apple Silicon / M1 / ARM64 Mac the build can't get beyond running SetVersion, because the built binary can't run. For some reason the arm64 version of the "SetVersion" crashes mysteriously. Actually, the process is killed by signal 9 before it actually runs, possibly it's the loader that gets killed.

However, the x86_64 version of "SetVersion" (that is also built) runs fine even on M1 Macs.

Rosetta2, which allows to run x64_86 on ARM64/M1 Macs, is required when installing Xcode, so an M1 Mac with xcode should always be able to run x86_64 binaries.

Thus, a feasible solution to the problem is to build SetVersion only for x86_64. This patch does this.

### DIFF
--- a/mac_build/boinc.xcodeproj/project.pbxproj
+++ b/mac_build/boinc.xcodeproj/project.pbxproj
@@ -2877,7 +2877,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "# echo \"BuiltProductsDir = ${BUILT_PRODUCTS_DIR}\"\n# echo \"CONFIGURATIONTEMPDIR = ${CONFIGURATION_TEMP_DIR}\"\n# echo \"SRC ROOT = ${SRCROOT}\"\n# echo \"architecture = ${ARCHS}\"\n# echo \"native architecture = ${ARCHS}\"\n\"${BUILT_PRODUCTS_DIR}/SetVersion\"\n";
+			shellScript = "# echo \"BuiltProductsDir = ${BUILT_PRODUCTS_DIR}\"\n# echo \"CONFIGURATIONTEMPDIR = ${CONFIGURATION_TEMP_DIR}\"\n# echo \"SRC ROOT = ${SRCROOT}\"\n# echo \"architecture = ${ARCHS}\"\n# echo \"native architecture = ${ARCHS}\"\n codesign -f -o runtime -s - \"${BUILT_PRODUCTS_DIR}/SetVersion\"\n\"${BUILT_PRODUCTS_DIR}/SetVersion\"\n";
 		};
 		DDC8FB2B1F6D398700BE55B8 /* ShellScript */ = {
 			isa = PBXShellScriptBuildPhase;


bema-aei:
I can confirm that an Xcode build of BOINC on an M1 Mac succeeds with this patch and fails without it.

CharlieFenton:
I cannot help with this because I do not have an Apple Silicon M1 (arm64) Mac. That is why Bernd created this issue.

> Thus, a feasible solution to the problem is to build SetVersion only for x86_64. This patch does this.

That is only a temporary solution, because after a few years Apple will remove the ability to run X86_64 code under Rosetta 2 emulation.

**@bema-aei, have you confirmed that your SetVersion app is signed?**

As I wrote to @bema-aei, I ran across some information that might help understand the SetVersion mystery, though it's unclear whether it applies.

<https://eclecticlight.co/2021/01/27/how-rosetta-complicates-call-chains-on-m1-macs/> says:
There are two ways to force tools to be run in Rosetta translation: set this option in Terminal’s Get Info window to run the whole shell in translation, or use the shell command arch. This has simple syntax:
arch -x86_64|-arm64 commandname
So
arch -x86_64 zsh
will run the zsh shell in translation. This will by default mean that Universal binaries will be run in translation, and won’t be killed in the event that they’re unsigned.

This is of interest for 2 reasons:
[1] It offers a way to run the Intel code in xcodebuild and the entire toolchain it calls under Rosetta emulation on an M1 Mac. Presumably this will cause SetVersion to run as Intel code even if it is built as a fat binary.

[2] It allude to a possible reason SetVersion was killed before running. This web page states this more explicitly:
Should you make the mistake of trying to run an unsigned Universal binary from an arm64 shell, that will be refused with the error message that the process was killed, under ARM-native security rules.

This latter statement is confirmed by Apple's documentation here:
<https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-universal-apps-release-notes > which says:
New in macOS 11 on Macs with Apple silicon, and starting in macOS Big Sur 11 beta 6, the operating system enforces that any executable must be signed before it’s allowed to run. There isn’t a specific identity requirement for this signature: a simple ad-hoc signature is sufficient.
and
This new policy doesn’t apply to translated x86 binaries running under Rosetta 2, nor does it apply to macOS 11 running on Intel-based platforms

Also:
To reduce the impact on existing development workflows, starting with Xcode 12 beta 4, the toolchain will now automatically sign your executables whenever you build from Xcode, or use command-line utilities such as clang(1) or ld(1).


And it does appear that the SetVersion I build with xcodebuild using Xcode 12.2 on MacOS 11.2 is code signed with an ad-hoc signature:
codesign --display --verbose /Volumes/Dev/BOINC_BigSur/boinc_fix_scripts_for_zsh/mac_build/build/Deployment/SetVersion 
Executable=/Volumes/Dev/BOINC_BigSur/boinc_fix_scripts_for_zsh/mac_build/build/Deployment/SetVersion
Identifier=SetVersion-5555494461782be86ee034628697898cd9ecb0e9
Format=Mach-O universal (x86_64 arm64)
CodeDirectory v=20400 size=652 flags=0x2(adhoc) hashes=9+7 location=embedded
Signature=adhoc
Info.plist=not bound
TeamIdentifier=not set
Sealed Resources=none
Internal requirements count=0 size=12

CharlieFenton:
You could also try adding the following in the script in the SetVersion target's Script Phase to ensure it is ad-hoc code signed before execution:
`codesign -f -o runtime -s - "${BUILT_PRODUCTS_DIR}/SetVersion"`
before this line:
`"${BUILT_PRODUCTS_DIR}/SetVersion"`

CharlieFenton:
I recently noticed this comment I wrote some time ago (currently at line 544) in _mac_installer/release_boinc.sh_:

```
# Workaround for code signing problem under Xcode 12.2:
# Code sign each architecture separately then combine into a universal binary
```
  
Unfortunately, I don't remember how I discovered this problem or why it affected only the boinccmd executable. But it may be worth investigating whether this problem could be affecting SetVersion, causing the ad_hoc code signature to apply only to the X86_64 architecture of the fat binary and not to the arm64 portion.

bema-aei:
Indeed this looks like a problem with the signature. When built for x86_64 only, "SetVersion" has a proper signature:
```
codesign --display --verbose mac_build/build/Deployment/SetVersion
Executable=/Users/bema/source/BOINC_dev/boinc/mac_build/build/Deployment/SetVersion
Identifier=SetVersion-55554944df8460240c7d3713a622a001b52b2a7b
Format=Mach-O thin (x86_64)
CodeDirectory v=20400 size=652 flags=0x2(adhoc) hashes=9+7 location=embedded
Signature=adhoc
Info.plist=not bound
TeamIdentifier=not set
Sealed Resources=none
Internal requirements count=0 size=12
``` 
However when the build fails, i.e. without this patch, and "SetVersion" is a "universal" binary, it doesn't appear to be signed at all:
```
file mac_build/build/Deployment/SetVersion
mac_build/build/Deployment/SetVersion: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64:Mach-O 64-bit executable arm64]
mac_build/build/Deployment/SetVersion (for architecture x86_64):	Mach-O 64-bit executable x86_64
mac_build/build/Deployment/SetVersion (for architecture arm64):	Mach-O 64-bit executable arm64

codesign --display --verbose mac_build/build/Deployment/SetVersion
mac_build/build/Deployment/SetVersion: code object is not signed at all
```

bema-aei:
> You could also try adding the following in the script in the SetVersion target's Script Phase to ensure it is ad-hoc code signed before execution:
> `codesign -f -o runtime -s - "${BUILT_PRODUCTS_DIR}/SetVersion"`
> before this line:
> `"${BUILT_PRODUCTS_DIR}/SetVersion"`

This did the trick on my M1 Mac and I also checked on an Intel machine that it does have no negative effects. The branch has been updated to just include the script change.

bema-aei:
The funny thing is that on an Intel Mac with the same OS & Xcode version SetVersion also gets build "universal" with both architectures, and *is* signed automatically:
```
codesign --display --verbose mac_build/build/Deployment/SetVersion
Executable=/Users/bema/source/BOINC_dev/boinc/mac_build/build/Deployment/SetVersion
Identifier=SetVersion-55554944df8460240c7d3713a622a001b52b2a7b
Format=Mach-O universal (x86_64 arm64)
CodeDirectory v=20400 size=652 flags=0x2(adhoc) hashes=9+7 location=embedded
Signature=adhoc
Info.plist=not bound
TeamIdentifier=not set
Sealed Resources=none
Internal requirements count=0 size=12
```
I still don't understand why this doesn't happen on an M1 Mac.

CharlieFenton:
> The funny thing is that on an Intel Mac with the same OS & Xcode version SetVersion also gets build "universal" with both architectures, and is signed automatically:

That explains why I never saw this problem on my Intel Mac.

> I still don't understand why this doesn't happen on an M1 Mac.

It might be a bug with the arm64 version of Xcode. You might want to submit a bug report to Apple.

CharlieFenton (approved):

PR #4385, opened 2021-05-31
[Linux][Manager][Scripts] Compile linux manager with vcpkg

talregev:
Compile Linux manager with vcpkg.

Wait for PRs in vcpkg to be merge: 
* ~~https://github.com/microsoft/vcpkg/pull/17111~~
* ~~https://github.com/microsoft/vcpkg/pull/18580~~
* ~~https://github.com/mesonbuild/meson/issues/6710~~
* ~~https://github.com/microsoft/vcpkg/pull/22201~~
* [x] https://github.com/microsoft/vcpkg/pull/23765
* [x] https://github.com/microsoft/vcpkg/pull/24047
* [x] Change to original vcpkg link.

### DIFF
--- a/linux/update_vcpkg_manager.sh
+++ b/linux/update_vcpkg_manager.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+set -e
+
+if [ ! -d "linux" ]; then
+    echo "start this script in the source root directory"
+    exit 1
+fi
+
+CACHE_DIR="$PWD/3rdParty/buildCache/linux"
+BUILD_DIR="$PWD/3rdParty/linux"
+VCPKG_PORTS="$PWD/3rdParty/vcpkg_ports"
+VCPKG_ROOT="$BUILD_DIR/vcpkg"
+
+if [ ! -d $VCPKG_ROOT ]; then
+    mkdir -p $BUILD_DIR
+    git -C $BUILD_DIR clone https://github.com/microsoft/vcpkg
+fi
+
+git -C $VCPKG_ROOT pull
+$VCPKG_ROOT/bootstrap-vcpkg.sh
+$VCPKG_ROOT/vcpkg install curl[core,openssl] wxwidgets --clean-after-build --overlay-triplets=$VCPKG_PORTS/triplets/ci
+$VCPKG_ROOT/vcpkg install freetype[core,bzip2,png] freeglut ftgl --clean-after-build --overlay-triplets=$VCPKG_PORTS/triplets/ci
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        type: [libs, client, apps, libs-vcpkg, client-vcpkg, apps-vcpkg, libs-cmake, server, manager-with-webview, manager-without-webview, unit-test, integration-test]
+        type: [libs, client, apps, libs-vcpkg, client-vcpkg, apps-vcpkg, libs-cmake, manager-with-webview-vcpkg, server, manager-with-webview, manager-without-webview, unit-test, integration-test]
       fail-fast: false
     steps:
       - uses: actions/checkout@v2
@@ -29,7 +29,7 @@ jobs:
         run: |
           sudo add-apt-repository 'deb http://archive.ubuntu.com/ubuntu bionic main universe'
           sudo apt-get -qq update
-          sudo apt-get install -y libftgl-dev freeglut3-dev libcurl4-openssl-dev libxmu-dev libxi-dev libfcgi-dev libxss-dev libnotify-dev libxcb-util0-dev libgtk2.0-dev libwebkitgtk-dev p7zip-full libxxf86vm-dev ocl-icd-opencl-dev zip
+          sudo apt-get install -y libftgl-dev freeglut3-dev libcurl4-openssl-dev libxmu-dev libxi-dev libfcgi-dev libxss-dev libnotify-dev libxcb-util0-dev libgtk2.0-dev libgtk-3-dev libsecret-1-dev libgcrypt20-dev libsystemd-dev libwebkitgtk-dev p7zip-full libxxf86vm-dev ocl-icd-opencl-dev zip
 
       - name: Install dependencies for integration testing
         if: matrix.type == 'integration-test'
@@ -87,6 +87,10 @@ jobs:
       - name: Configure libs with cmake vcpkg
         if: success() && matrix.type == 'libs-cmake'
         run: linux/ci_configure_libs_cmake.sh
+
+      - name: Configure manager with webview and vcpkg
+        if: success() && matrix.type == 'manager-with-webview-vcpkg'
+        run: linux/ci_configure_manager.sh
 
       - name: Configure server
         if: success() && matrix.type == 'server'
--- a/deploy/prepare_deployment.py
+++ b/deploy/prepare_deployment.py
@@ -209,6 +209,9 @@ def prepare_linux_apps_vcpkg(target_directory):
 def prepare_linux_manager_with_webview(target_directory):
     prepare_7z_archive('linux_manager-with-webview', target_directory, linux_manager_list)
 
+def prepare_linux_manager_with_webview_vcpkg(target_directory):
+    prepare_7z_archive('linux_manager-with-webview-vcpkg', target_directory, linux_manager_list)
+
 def prepare_linux_manager_without_webview(target_directory):
     prepare_7z_archive('linux_manager-without-webview', target_directory, linux_manager_list)
 
@@ -249,6 +252,7 @@ boinc_types = {
     'linux_apps': prepare_linux_apps,
     'linux_apps-vcpkg': prepare_linux_apps_vcpkg,
     'linux_manager-with-webview': prepare_linux_manager_with_webview,
+    'linux_manager-with-webview-vcpkg': prepare_linux_manager_with_webview_vcpkg,
     'linux_manager-without-webview': prepare_linux_manager_without_webview,
     'win_apps-mingw': prepare_win_apps_mingw,
     'win_apps-mingw-vcpkg': prepare_win_apps_mingw_vcpkg,
--- a/linux/ci_configure_manager.sh
+++ b/linux/ci_configure_manager.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+set -e
+
+if [ ! -d "linux" ]; then
+    echo "start this script in the source root directory"
+    exit 1
+fi
+
+CACHE_DIR="$PWD/3rdParty/buildCache/linux"
+BUILD_DIR="$PWD/3rdParty/linux"
+VCPKG_ROOT="$BUILD_DIR/vcpkg"
+export VCPKG_DIR="$VCPKG_ROOT/installed/x64-linux"
+
+linux/update_vcpkg_manager.sh
+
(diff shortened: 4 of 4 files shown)


talregev:
Trying https://github.com/microsoft/vcpkg/pull/18580

CharlieFenton:
@talregev: Does building with vcpkg provide a way to ensure that a particular version of BOINC can always be built with exactly the same versions of all the dependent libraries? In other words, do older versions of dependent libraries used in earlier releases of BOINC always remain available? This is very important to ensure reproducibility when investigating bug reports from the field.

Because this is important, I have carefully written the Xcode project for each release of BOINC on the Mac to fail if one tries to build it again with a different version of any of the dependent libraries than the ones with which the original release was built. And the current scripts to build the dependent libraries for the Mac in GitHub actions ensure that the CI builds use those same library versions.

AenBleidd:
@CharlieFenton, in current design we always use the latest version of dependencies. But it's possible to stick with particular version of any dependency: https://devblogs.microsoft.com/cppblog/take-control-of-your-vcpkg-dependencies-with-versioning-support/

talregev:
> exactly the same versions of all the dependent libraries?

You can configure vcpkg to work like this. 
For windows, android, and linux we always take the latest. for windows is for 2 years now. and everting work great. not sure why mac need to be different. 
Also @AenBleidd can give is opinion here too. He have more experience with vcpkg than I do.

CharlieFenton:
> in current design we always use the latest version of dependencies.

I think that is a big mistake. Any particular version of BOINC should be 100% reproducible. That requires always using the same dependent libraries. As I wrote, this is important for investigating bug reports, among other purposes.

talregev:
> I think that is a big mistake. Any particular version of BOINC should be 100% reproducible. That requires always using the same dependent libraries. As I wrote, this is important for investigating bug reports, among other purposes.

I don't agree with you. 
As I say we already work like this and it not hard for us to find bug inside boinc.
Also we have artifacts that save the boinc version exactly as it compile.

talregev:
> I have carefully written the Xcode project for each release of BOINC on the Mac to fail if one tries to build it again with a different version of any of the dependent libraries than the ones with which the original release was built.

This is a big mistake. harder to dev like this, and if you expect someone will replace or help you in mac dev, this is needed to change.

CharlieFenton:
There can be many reasons why a bug is reproducible when built with one version of a dependent library but not with others. I have found problems in almost every version of WxWidgets, for example, and different versions of wxWidgets have had different problems. There can also be more subtle bugs, such as when a bad pointer in BOINC code causes writing to a location in one of the dependent libraries. Since that memory location is in different parts of the code in different library versions, the symptoms may be different.

AenBleidd:
> As I wrote, this is important for investigating bug reports, among other purposes.

And this is still possible to do. It's not a big deal to build BOINC with any version of dependency from vcpkg (at least if this version exists there). In any case, using vcpkg for linux is optional. For Windows we use now the latest version of dependencies, but it's not released yet. Afterwards we can decide to use some particular version of dependencies for release branch, but for master it's easier to use the latest versions and test with them (and fix if necessary). For Android release we haven't use vcpkg yet for release, but we will see how it goes. If there will be any issues - we can always change our flow :)

CharlieFenton:
> I don't agree with you.

I'm not surprised ;=)

> This is a big mistake. harder to dev like this, and if you expect someone will replace or help you in mac dev, this is needed to change.

It is not harder at all. As long as the file mac_build/dependencyNames.sh is kept in sync with the Xcode project, the build scripts take care of it automatically. 

Over the years, I have found that many versions of the dependent libraries have problems in BOINC and their source files need to be patched before building. (Of course, this is on the Mac; I haven't been involved with setting up dependent libraries on other platforms.) And different versions of the same library have needed different patches. This is another reason I have found it important to ensure consistency in library versions.

I do realize that the entire approach to (and philosophy of) building and distributing software on Linux is very different from that on Windows and Macs. I would say that Linux users and developers have a different culture. as I understand it, a common expectation on Linux is that each user will build applications and libraries themselves for that specific computer. That is the basis for the configure / make approach which originated in UNIX and Linux.

On Windows and Mac, the expectation is that a developer will build the software in such a way that it will be compatible with a range of OS versions, and the computers tend to have fewer variations in their environment. Instead of building using the configure and make approach, developers on these platforms generally use the integrated development environments (IDE) Visual Studio and Xcode. And of course, Microsoft and Apple also have significant cultural and philosophical differences rom each other.

I haven't been paying much attention to how BOINC has been built for MS Windows recently, but in the past Rom Walton prebuilt a set of dependent library for each version of Visual Studio. Anyone building BOINC for Windows used those same prebuilt libraries, which guaranteed that the identical library versions were always used.

CharlieFenton:
> in the past Rom Walton prebuilt a set of dependent library for each version of Visual Studio
 
[Here](https://boinc.berkeley.edu/dl/boinc_depends/)  is the link to those prebuilt libraries.

talregev:
You refuse to do new active dev, and you refuse let other people make changes. Why?
If you have strong opinion on how things need to be done, please join to be active dev.

AenBleidd:
@CharlieFenton, during the last year, I found no issues with the current approach for Windows. Maybe the issue you're talking about is applicable to OSX only (or even for OSX it's not an issue anymore, and you just overcomplicating things, I believe, some fresh look is needed on all this (I'm not try to blame you or smth like this)). At least I saw no OS dependent patches for dependencies fr any platform except OSX. But I believe, in this case it's just wxWidgets that has quite shitty support of OSX.

> but in the past Rom Walton prebuilt a set of dependent library for each version of Visual Studio

And it was a problem: update dependencies in this case is way to harder (and they are too old there). I don't believe someone will decide to build some old version of BOINC for Windows. The latest version you can always build with the latest dependencies as we're doing now. Also, from my experience during last year of extensive usage of vcpkg on my job, I can say that we faced no issues on Windows with vcpkg (at least, no major issues).

Regarding linux, using vcpkg is just another way of BOINC build. I don't believe everyone will switch to use it (remember that linux users usually don't like Microsoft products, even free one ;) ). But from development perspective it's easier, as you don't need to take care about update (especially critical security updates of curl and openssl) of dependencies.

To summarise all above: we're not switching (at least for linux, Android and OSX) the way BOINC built, it's just an alternative method, and it affects build scripts only. For Windows we switched to vcpkg in VS2019 and now it's way more easier to build BOINC.

CharlieFenton:
> we're not switching (at least for linux, Android and OSX) the way BOINC built, it's just an alternative method, and it affects build scripts only.

@AenBleidd Thank you for explaining that. I did not realize it.

> You refuse to do new active dev, and you refuse let other people make changes. Why?
> If you have strong opinion on how things need to be done, please join to be active dev.

@talregev You have a very good point. I apologize. When you have been working on a project as many years as I have, it can be hard to let go. I need to be better at that.

talregev:
Wait for PRs in vcpkg to be merge: 
* [[wxwidgets] Fix usage, add cmake options #17111](https://github.com/microsoft/vcpkg/pull/17111) 
* [[wxwidgets] Fix wx-config --libs output #18580](https://github.com/microsoft/vcpkg/pull/18580)

linux/update_vcpkg_manager.sh
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+CACHE_DIR="$PWD/3rdParty/buildCache/linux"
+BUILD_DIR="$PWD/3rdParty/linux"
+VCPKG_PORTS="$PWD/3rdParty/vcpkg_ports"
+VCPKG_ROOT="$BUILD_DIR/vcpkg"
+
+if [ ! -d $VCPKG_ROOT ]; then
+    mkdir -p $BUILD_DIR
+    git -C $BUILD_DIR clone https://github.com/JackBoosY/vcpkg -b dev/jack/23758
```
AenBleidd, line 16: Build is probably failed because of this custom repo that is probably outdated and doesn't contain the link to the latest release of vcpkg-tool

talregev: I know. I already handle it.

talregev:
Ready for review. (Never too late).

linux/update_vcpkg_manager.sh
```diff
@@ -0,0 +1,22 @@
...
+VCPKG_ROOT="$BUILD_DIR/vcpkg"
+
+if [ ! -d $VCPKG_ROOT ]; then
+    mkdir -p $BUILD_DIR
+    git -C $BUILD_DIR clone https://github.com/microsoft/vcpkg
+fi
+
+git -C $VCPKG_ROOT pull
+$VCPKG_ROOT/bootstrap-vcpkg.sh
+$VCPKG_ROOT/vcpkg install curl[core,openssl] wxwidgets --clean-after-build --overlay-triplets=$VCPKG_PORTS/triplets/ci
```
AenBleidd, line 21: I assume this should use now manifest mode

configure.ac
```diff
@@ -1286,7 +1286,7 @@ else
     CLIENTGUIFLAGS="${CLIENTGUIFLAGS} -DNDEBUG"
 fi
 
-CLIENTGUILIBS="${WX_LIBS}"
+CLIENTGUILIBS="${WX_LIBS} -lX11"
```
AenBleidd, line 1289: Why this is needed here?

linux/update_vcpkg_manager.sh
talregev: We cannot do manifest mode. wxwidgets need to be installed separate because of it will not install with free type and other depends of fglut.

configure.ac
talregev: Because it missing a flags.

AenBleidd: Why this was not needed before? For me it looks like some detection is failing to detect X11 and add this flag automatically.

linux/update_vcpkg_manager.sh
AenBleidd: I don't understand. Could you please rephrase it?

talregev: Please ask on the part you don't understand.

configure.ac
talregev: It not fail to detect anything. The flags is coming from `wx-config`. In the source code version when they don't compile with cmake, the order of the flag is different from the order of the flags that generate when compile in vcpkg. that why I need to add this extra flag.

AenBleidd: Then this should be fixed on wxwidgets port side

linux/update_vcpkg_manager.sh
talregev: Please be more specific that I can answer you.

AenBleidd: Why can't you install wxwidgets and all its dependencies using vcpkg manifest?

talregev: Because `wxwidgets` port will not install together with `freetype`  It need install first wxwidgets then another command to install the rest. it cannot done in manifest mode.
You can read more here: 
https://github.com/microsoft/vcpkg/pull/24047#issuecomment-1098377279

configure.ac
talregev: I removed this flag.

linux/update_vcpkg_manager.sh
AenBleidd: Is there an open ticket for this issue?

talregev: It not an issue.

AenBleidd: If some part of functionality doesn't work as designed - it's an issue

talregev: ok. You can open an issue on vcpkg, and tell them that.

PR #4386, opened 2021-05-31
Apple M1 CPU feature detection

### LINKED ISSUE #4137: Mac: New universal BOINC client breaks CPU feature detection
**Describe the bug**
From Einstein server log posted on [BOINC message board](https://boinc.berkeley.edu/forum_thread.php?id=14047&postid=102391#102391)
```
2021-01-03 19:48:16.5251 [PID=7492 ] [version] Checking plan class 'FGRPSSE'
2021-01-03 19:48:16.5252 [PID=7492 ] [version] CPU [' '] lacks feature ' sse '
```

**Steps To Reproduce**

User reports:
> I've been running einstein@home on my 12Z Mac mini. Using the BOINC universal build, I never get any new wu's. Using a non-universal BOINC does get me FGRPSEE wu's, which in general validate fine. Am I not getting new wu's from the BOINC universal build because I have a 12Z, and not a M1?

**Additional context**
This may require @CharlieFenton and @bema-aei to liaise and tweak the 'universality' of the new Apple silicon handling.

bema-aei:
Fixes #4137 

**Description of the Change**

The MR adds code (to `hostinfp_unix.cpp`) that detects the features of the ARM64 CPU (e.g. neon*) and records these in the features string that is reported to the project scheduler.

Also with the (Mac OSX) client there is a separate helper program "detect_rosetta_cpu" built only for x86_64 architecture. During installation this gets installed in the BOINC Data directory. The native ARM64 OSX client tries to launch this program (in `cs_platforms.cpp`). If it runs successfully, it means that the system can run x86_64 binaries (i.e. the Rosetta 2 translator is installed), and "x86_64-apple-darwin" gets added to the platforms. When the program runs, it leaves a text file in the BOINC Data directory that contains the feature string of the emulated / translated x86_64 CPU (e.g. "SSE*"). If this file exists, it gets read in `hostinfo_unix.cpp`, and the features are added to the CPU features string. This string then is subject to conversion / translation as a native x86_64 feature string (e.g. tolower()).

Additionally the release scripts in `mac_installer/` are updated, e.g. made zsh-compatible.

**Release Notes**
Detect and report the features of a Apple Silicon M1 CPU, including a possibly emulated x86_64 CPU.


bema-aei:
Darn - noticed a bug. The current version will break CPU feature detection on Intel Macs. Will fix this.

CharlieFenton (approved):

CharlieFenton:
The new detect_rosetta_cpu target has a fairly large number of build settings which, though identical to the default project settings, are shown in bold font (in the Xcode IDE) because they have been entered separately for this target. Please change these so they inherit from the project-wide settings. Otherwise, this looks good to me.

bema-aei:
should be fixed by last commit.

PR #4398, opened 2021-06-08
[Android] Run boinc after boot

talregev:
Run boinc after boot
Fix for android 8 and above
Tested on 
- oneplus one android 10. (lineageOS)
- galaxy gio android 4.4. (cyanogenmod)


talregev:
This fix also work on my armv6 (android 4.4)

PR #4401, opened 2021-06-13
[WIP] Mac trigger "Install Rosetta 2?" dialog on Apple Silicon Macs

CharlieFenton:
Apple Silicon (arm64) Macs can run X86_64 executables using Apple's Rosetta 2 x86_64 translation / emulation software. For BOINC projects to take advantage of this, the BOINC client must report the emulated x86_64 CPU features to project servers. The BOINC client launches its detect_rosetta_cpu helper app to determine these features so the client can report them. 

But Rosetta 2 is not installed by default on Apple Silicon Macs. MacOS automatically displays its built-in "Install Rosetta 2?" dialog when a user launches an x86_64-only application if Rosetta 2 has not yet been installed. But detect_rosetta_cpu, which is a bare (command-line) executable rather than a full MacOS application, apparently does not trigger this dialog.

This PR is intended to update the detect_rosetta_cpu helper app to trigger the dialog when appropriate, by embedding an info.plist section in it. We hope that is sufficient to trigger the dialog. If not, we may need to convert the detect_rosetta_cpu command-line executable into a full Mac application bundle.

### DIFF
--- a/client/cs_platforms.cpp
+++ b/client/cs_platforms.cpp
@@ -42,6 +42,7 @@ LPFN_ISWOW64PROCESS fnIsWow64Process;
 
 #ifdef __APPLE__
 #include <sys/sysctl.h>
+#include <Carbon/Carbon.h>
 extern int compareOSVersionTo(int toMajor, int toMinor);
 #endif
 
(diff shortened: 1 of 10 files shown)


CharlieFenton:
closed in favor of #4406

PR #4408, opened 2021-06-20
[Android] Resume computations autostart after reboot after powersourc…

AenBleidd:
…e is connected


talregev (approved):
I test this on oneplus one (android 10) and it working.
I approve.

PR #4421, opened 2021-06-25
[Android] localization update

AenBleidd:


Ageless93:
A couple of things. 
I have just finished translating the 15 strings, but find something missing about the English language.  

1. We're using "Remote Control" to allow this BOINC to be managed remotely. Using remote control will probably translate into many languages as the little physical device you use to control your TV with. I'm wondering, are we managing the client remotely, or controlling it remotely? I know https://boinc.berkeley.edu/wiki/Controlling_BOINC_remotely says the latter, but how can we make this better translatable? I have translated it to manage the client, not control it. 
2. The English uses sentences such as "When changed, BOINC client will be relaunched." and "BOINC client will be relaunched after password change." Shouldn't we add "the" in front of BOINC? Same with "Use account manager to control this device", "an account manager?"
3. Use of full stops. Some sentences end on a full stop, others don't. 
4. Some of the sentences to be translated end on an empty line, such as "Use account manager to control this device
    ". Transifex complains about this when I copy the empty line: The translation ends with an empty line, while it should not.

AenBleidd:
@Ageless93,
> but how can we make this better translatable?
This is a good question. We could literally change 'Remote Control' to 'Controlling BOINC remotely'.
Could you please open a ticket to find a better name for this?

> The English uses sentences such as "When changed, BOINC client will be relaunched." and "BOINC client will be relaunched after password change." Shouldn't we add "the" in front of BOINC? Same with "Use account manager to control this device", "an account manager?"

Unfortunately, I'm not a native speaker. Thanks for these findings, could you please open a PR for this?

> Use of full stops. Some sentences end on a full stop, others don't.

> Some of the sentences to be translated end on an empty line, such as "Use account manager to control this device
". Transifex complains about this when I copy the empty line: The translation ends with an empty line, while it should not.

I'll open a ticket to fix it.

Ageless93:
> > The English uses sentences such as "When changed, BOINC client will be relaunched." and "BOINC client will be relaunched after password change." Shouldn't we add "the" in front of BOINC? Same with "Use account manager to control this device", "an account manager?"
> 
> Unfortunately, I'm not a native speaker. Thanks for these findings, could you please open a PR for this?

You mean a ticket? I don't write code. ;)

AenBleidd:
Ok, let it be a ticket :) If you (or someone else with better (than mine) language knowledge) will give me to fixes to the strings that should be fixed - I'll do a PR.

Ageless93:
I emailed these to you some time ago.

AenBleidd:
@Ageless93, hm, sorry, I thought I replied you back. Tough day, my bad.
Thank you for findings and reporting them to me!

PR #4423, opened 2021-06-28
[Android] Run boinc from read only path (API 29)

talregev:
Run boinc from read only path (API 29)

compile and run and crunch on Oneplus one (android 10)


talregev:
I check again. It compile, and run and crunch on OnePlus one (android 10).

PR #4441, opened 2021-07-05
[Android] Add 'https://' prefix to the url field by default to 'Add p…

### LINKED ISSUE #4436: [Android] When adding an URL by hand, fill in https:// for the user
At the moment when adding an URL by hand, the whole field is empty. Okay, I am not sure whether adding an URL without the https:// prefix will add the https:// prefix automatically, but during the whole adding process it just shows as the URL without the https:// prefix. 

Perhaps we can get a set https:// text before the box in which we fill in the URL, so people don't have to add it by hand and add the prefix so the user doesn't have to do it?

AenBleidd:
…roject by URL' window.

This fixes #4436


talregev (approved):
I tested on OnePlus one, and it add the https:// as expected.

PR #4457, opened 2021-07-22
[Android] Investigate race condition that cause long start on remote

### LINKED ISSUE #4452: [Android][Remote] On remote, long time to boinc client to load.
On remote It take long time to boinc client to load. 
You can notice it if you close boinc and open it again with long stay on slpash activity.

talregev:
Investigate race condition that cause long start on remote.
Fixes #4452


talregev:
Cannot reproduce #4452

PR #4471, opened 2021-07-28
[Android] improve logging system combined

AenBleidd:


talregev (approved):
I compile, run and do some tests on OnePlus One (Android 10).
All my tests are pass, also seems very stable nice GUI.

PR #4493, opened 2021-08-05
[CI] Always upload unit tests results

talregev:
Always upload unit tests results


.github/workflows/windows.yml
```diff
@@ -105,6 +105,7 @@ jobs:
           path: deploy/win_manager/*.7z
 
       - name: Upload Google Tests Results
+        if: always()
```
AenBleidd, line 108: Please move this line after 'uses'

talregev: Fix.

AenBleidd (approved):

PR #4497, opened 2021-08-05
[WIP][Android][Debug] network patch

talregev:
network patch


talregev:
Cannot reproduce #4452

PR #4507, opened 2021-08-12
[Lib] Add remote_submit to lib for linux and android

talregev:
Add remote_submit to lib for linux and android and mingw.

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -66,6 +66,7 @@ generic_sources = \
     proc_control.cpp \
     project_init.cpp \
     proxy_info.cpp \
+    remote_submit.cpp \
     sched_msgs.cpp \
     shmem.cpp \
     str_util.cpp \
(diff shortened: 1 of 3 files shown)


davidpanderson:
The functions in remote_submit.cpp don't belong in the client.

talregev:
ok. I can rename `struct ERROR` -> `struct RS_ERROR` for compile condor example in mingw.   
This is acceptable?

davidpanderson:
What error are you trying to fix?

talregev:
Expected identifier before numeric constant.   
My guess that ERROR is some define in mingw. (maybe also in windows).

talregev:
I am closing this PR. this PR is meant to insert inside the `remote_submit.cpp` inside lib, to give services to apps, that want to submit report. 
I will create another PR for compile more apps (examples) in mingw. 
If we want, we can create public discussion if `remote_submit.cpp` inside lib.

PR #4520, opened 2021-09-03
[Android] Fix an issue with 'Project attach' screen when not all cred…

AenBleidd:
…entials are provided


talregev (approved):
I test it on OnePlus one (Android 10). It running and crunching. 
Also when I put an empty name, it show me toes bar show me the warning and the reason why it not continue.

PR #4529, opened 2021-09-29
Scheduler: allow negation of regular expressions in plan class XML specs (standard)

bema-aei:
This MR is an extension of #4390 with the changes discussed but not (yet) implemented there, i.e. using the standard `(?!regex)` as proposed by @Rytiss instead of the simple but non-standard `!regex` proposed by me and implemented by @davidpanderson . The branch is based on David's `dpa_regex` with an extra commit.

### DIFF
--- a/sched/plan_class_spec.cpp
+++ b/sched/plan_class_spec.cpp
@@ -33,12 +33,16 @@ int REGEX_CLAUSE::init(const char* p) {
     present = true;
     negate = false;
     expr = p;
-    if (*p == '!') {
-        p++;
+    if (!strncmp(p,"(?!",3) && p[strlen(p)-1] == ')') {
+        p += 3;
+        p[strlen(p)-1] = '\0'; // don't parse trailing ')'
         negate = true;
     }
     if (regcomp(&regex, p, REG_EXTENDED|REG_NOSUB) ) {
         return ERR_XML_PARSE;
+    }
+    if (negate) {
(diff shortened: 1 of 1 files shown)


Rytiss:
I think this should work, but I'm not a C++ guy so I won't merge this.

Thanks for taking your time to mod the original PR!

bema-aei:
@Rytiss I think this implementation / modification comes closest to our original intentions of negation of regex. I wanted to have this ready before your original proposal becomes implemented. Thanks @davidpanderson for taking the time for the previous implementation, factoring out these regex as a separate datatype was valuable.

However I didn't test this yet very much, due to lack of time and opportunity, and I didn't look much around in that code, in particular where the C-Strings that are manipulated there originate from. Certainly there is potential for an accidental error (probably off-by-one) in my modification. And we recently came across trouble with storing (`char*`) pointers to C strings converted from C++ strings with `c_str()`. In modern clib versions the lifetime of these converted strings seems different than in previous versions, a pointer to such a string conversion buffer that is stored permanently may give unexpected results after some time, i.e. after other such conversions had happened. Maybe we should use `strdup()` to preserve the original regex string.

AenBleidd:
@bema-aei, I'll check c++ code (without logic behind it)

AenBleidd:
Reference: https://godbolt.org/z/r1szzY9rr

PR #4539, opened 2021-10-02
all master URL update to https:

davidpanderson:
If a scheduler reply has a different master URL than the current one:
- if it's a http: to https: change, do it
- otherwise tell the user the project seems to have changed its master URL,
  and they should detach/reattach.
Remove the nonsensical "your are attached twice..." message.

Note: code you wrote a long time ago sometimes doesn't seem
to make any sense at all.


AenBleidd (approved):

TheAspens:
Thanks for the fix!

PR #4540, opened 2021-10-02
Update ca-bundle.crt

AenBleidd:


AenBleidd:
@davidpanderson, looks like this bundle works fine. Please merge this PR and I'll create a Release branch for Windows only release

RichardHaselgrove:
I notice that there are significant differences between this 'bindle' (??!!) and the 31 May 2020 bundle it will be replacing. To save time, could you perhaps indicate where this version has been sourced from?

AenBleidd:
As it mentioned in the file itself, it's an official Mozilla bundle as before: https://curl.se/docs/caextract.html

RichardHaselgrove:
Ta. I've confirmed that with a fresh download copy from source (identical), and it works with two of the projects which were troubling us on Thursday. I'll merge it.

Issue #4544, opened 2021-10-05
Temporary release problems following certificate expiry


RichardHaselgrove:
I tried downloading and building under VS2019 for the first time today. The client built - with warnings, but it's currently running.

But the Manager failed to build. I can provide error messages tomorrow, but that's not important.

There's a suggestion that we could release a composite installer, with an updated client and ca-certificate bundle, but an older Manager. Could that keep the science on track, and buy time to sort out what's wrong with the Manager build?

AenBleidd:
> If you can build the Manager OK, why isn't there a new release in testing already? What's the hold-up? The silence is deafening.

'I can build Manager' doesn't mean it's ready for release ;) And there is a blocker I'm working on currently: https://github.com/BOINC/boinc/issues/4073

Since it will be done - I believe we could think about new major release for alpha testing.

RichardHaselgrove:
I don't think I'd call that a blocker, compared with all Windows clients being unable to return completed science to multiple projects.

RichardHaselgrove:
The whole principle of the GitHub workflow is that the master branch is maintained in constant readiness, just in case of an eventuality like this. Blockers should be isolated in WIP branches until solved and merged.

I made a temporary working certificate bundle available to active participants on 30 September. Set-and-forget participants need something stronger.

Closing comms for tonight.

CharlieFenton:
> The whole principle of the GitHub workflow is that the master branch is maintained in constant readiness, just in case of an eventuality like this. Blockers should be isolated in WIP branches until solved and merged.

Hopefully the master branch is always ready for building a release **candidate** ready to go through extensive alpha testing before becoming a public release. This is different from being ready for immediate public release, even in an emergency. The latter would require every PR to go through full alpha testing before merging into master.

I agree with @AenBleidd that the correct way to do this is to add the minimal necessary changes  to the exiting release. However, due to the terrible pollution of the current release branch _client_release/7/7.16_ due to people mistakenly merging master into it repeatedly, that will be difficult.

I created a less polluted release branch for Mac releases: _client_release/7.16/7.16.12_branch_ by branching off the  _client_release/7/7.16_ branch at a point before much of the damage was done. You might want to create a new branch from my _client_release/7.16/7.16.12_branch_ as a starting point. **Please do not merge anything into my _client_release/7.16/7.16.12_branch_**!

AenBleidd:
I already made a branch directly from 7.16.11 release tag that was the last Windows release

[...]

PR #4545, opened 2021-10-06
[Windows] Build curl for Windows without openssl.

### LINKED ISSUE #4542: [Windows][Manager] Build curl with schannel
Currently curl is build with openssl that requires ca-bundle.crt to exist.
If curl is build with schallen instead of openssl, it will use WIndows Certificate Store that should simplify future certificate support.

AenBleidd:
Build it with schannel instead that allows curl to use Windows certificate store and doesn't relay on ca-bundle.crt anymore.

This fixes #4542


AenBleidd:
@RichardHaselgrove, could you please test this PR?

RichardHaselgrove:
@AenBleidd - I'll get onto it within the hour. I want to re-run the VS2019 build first, so I can get you the error messages you asked for. I'll be on more familiar territory to do a PR test under VS2013 - would that be acceptable?

AenBleidd:
@RichardHaselgrove,

> I'll be on more familiar territory to do a PR test under VS2013 - would that be acceptable?

No, this is for VS2019 only.
On the next major release for Windows we should get rid of VS2010 and VS2013 solutions and support only one, the latest

AenBleidd:
Btw, you can just use build artifacts from GitHub actions to test this

AenBleidd:
> I want to re-run the VS2019 build first, so I can get you the error messages you asked for.

That would be nice, because as you can see from our CI, it's build with VS2019 without any issues

RichardHaselgrove:
Test machine is booted back into Win 10. I'll post results from the build back in my issue from last night, then try the artifacts from here.

RichardHaselgrove:
Provisional report only. I've reverted to a standard v7.16.11 installation, and verified that it failed to connect with two of the projects that are affected by the current outage - GPUGrid and CPDN.

I then replaced boinc.exe and boincmgr.exe with the artifacts from here: noting that ca-bundle.crt is still deployed by the test build. But I deleted it from the working folder to test - I also deleted the curl and OpenSSL DLLs from the standard installation.

Client attached successfully to GPUGrid.
It failed to attach to CPDN - logged `Fetching configuration file from https://climateprediction.net/get_project_config.php`and nothing else, and reported "The project is currently unavailable". However, I can reach it fine from another machine, so it may be a project problem - they're being badly let down by their hosting University, with multiple network problems.

I'll check it again on Windows 7 - I haven't got any older versions in use at the moment, though I know some volunteers do.

RichardHaselgrove:
Tried it under Windows 7 - same procedure as last time. Got this in the log, which seems about right.
```
06/10/2021 11:02:46 |  | Version change (7.16.20 -> 7.19.0)
06/10/2021 11:02:46 | climateprediction.net | URL https://climateprediction.net/; Computer ID 1460497; resource share 100
06/10/2021 11:02:46 | climateprediction.net | Fetching scheduler list
06/10/2021 11:02:47 | climateprediction.net | Master file download succeeded
```
I'll check up on the 'get_project_cfg' failure direct with the project.

I've allowed production work to resume, and scheduler contacts and downloads seem to be proceeding normally.

CharlieFenton (approved):
I don't have a working MS Windows computer so I can't do any testing, but this looks OK as far as I can tell. I am not very knowledgeable about ssl and encryption, so I don't know whether any code in _lib/crypt.cpp_ depends on the certificate.

AenBleidd:
Works well on Win7 SP1 and higher (have no Vista to check on it also)

CharlieFenton:
@AenBleidd Please excuse my ignorance about _vcpkg_, but in the changed files I do not see anything that builds curl without OpenSSL. according to the _help_ in curl's _configure_ file, this involves building curl with the option 
`--without-ssl           disable OpenSSL`.

My understanding from [this document](https://curl.se/docs/sslcerts.html) is that to use Windows built-in certificate you should build cur with either
`  --with-winssl           enable Windows native SSL/TLS`
or 
 ` --with-schannel         enable Windows native SSL/TLS`

I see nothing in this PR that does any of these things.

To answer your question from PR #4544, the above referenced document says that curl can be told to use MacOS system certificate store by Secure Transport Support by building with this option:
`--with-secure-transport enable Apple OS native SSL/TLS`
but according to [Apple's documentation](https://developer.apple.com/documentation/security/secure_transport?language=objc) Secure Transport is considered legacy (deprecated.)

CharlieFenton:
@AenBleidd Please ignore my previous question. I see there is a reference to channel in your changed _vcpkg_ call.

CharlieFenton:
> So probably we can put back the code that checks for ca-bundle.crt file and instead put some option to use ca-bundle.crt instead of Windows certificate store in case of some errors. What do you think?

Would this not always use the ca-bundle.crt file instead of the Windows certificate store, unless this code is executed only after an error? In any case, I think the alternate approach we have been discussing in PR #4544 might be a better long-term solution than the one in this PR for the reasons I gave [here](https://github.com/BOINC/boinc/issues/4544#issuecomment-937558197).

Ageless93:
> 
> 
> Do we have anyone in the development community who can remember the original rationale for deciding on the current Windows procedure in the first place? As I remember it, it was Rom Walton who did much of the Windows work.

If the question is why ca-bundle.crt is put in the BOINC Programs directory and not the data directory, then looking at https://boinc.berkeley.edu/trac/wiki/ClientSetupLogicWinFileLayout this was the setup for the client version 5, which solely ran from the C:\Program Files\BOINC directory. Those were Windows 95, 98, ME and XP days, maybe even 2000. 

When the client setup was changed to present logic (https://boinc.berkeley.edu/trac/wiki/ClientSetupLogicWinSix, https://boinc.berkeley.edu/trac/wiki/ClientSetupWinSix) the setup of the BOINC Programs directory was left unchanged.

AenBleidd:
> Which apparently isn't followed these days, maybe we should bring it back.

That's exactly what I did in my last PR. I plan to automate this process somehow to always have on master the latest ca-bundle.crt file

AenBleidd:
@davidpanderson, what do you think? Should we go with this solution (use Windows certificate store) or should I implement automatic ca-bundle.crt update?

PR #4549, opened 2021-10-08
web: for News forum, always put oldest post (the news) first

davidpanderson:


RichardHaselgrove:
It makes sense to always have the actual 'News' visible, but could you

1) Warn news posters that they will have to post problem resolutions separately - otherwise they'll be missed
2) Do something to hide/modify the sort order selector on News threads - it has become misleading.

Issue #2352, opened 2018-02-16
Config: "Don't contact ref site" blocks "Get Project List"


RichardHaselgrove:
With the increasing emphasis on having an up-to-date 'all_projects_list.xml' file (especially for autologin), is there any remaining justification for preventing the download of a new file when cc_config.dont_contact_ref_site is set?

https://github.com/BOINC/boinc/blob/master/client/acct_setup.cpp#L254

TheAspens:
Yes.  The usage that I have seen for that flag is set when large users do mass deployments like WCG has worked with various corporations and other entities to do.  There security groups don't want the client talking to anything but the WCG project servers they have set it up to communicate with.  They also generally do not allow the people at the workstations to control the client either.

We could argue that the parameter is named poorly, but the use case above requires that these additional requests that the client makes continue to be disallowed.

RichardHaselgrove:
OK. I've used the flag (for many years) to avoid being rude and pestering Google - I have better ways to diagnose the health of my personal internet connection. I obviously don't think like a large, secure, corporation!

PR #4550, opened 2021-10-08
[WIP] Auto update certificate bundles

### LINKED ISSUE #4544: Temporary release problems following certificate expiry

CharlieFenton:
Fixes #4544

**Description of the Change**
Adds code to the BOINC client to automatically download a fresh _ca-bundle.crt_ file from the BOINC server every 2 weeks. It expects the file to be at
`https://boinc.berkeley.edu/ca-bundle.crt` which is a shortcut for the full URL: 
`https://boinc.berkeley.edu/dl/ca-bundle.crt`

**Note**: Currently, this code will not affect curl on MS Windows computers because as of PR #4545 curl uses _schannel_ instead of _openssl_ and so uses the Windows certificate store rather than _curl/ca-bundle.crt_. If it is desired to return to using _curl/ca-bundle.crt_, this automatic updating can be enabled for MS Windows by changing line 55 of _client/http_curl.cpp_ from 
`#define AUTO_UPDATE_MSW_CERTS 0`
 to 
`#define AUTO_UPDATE_MSW_CERTS 1`

**Alternate Designs**
PR #4545 partly solves this, but only for computers running currently supported versions of MS Windows. (Versions of MS Windows which are no longer supported by Microsoft do not get their certificate store updated automatically, though many can be updated manually by the user.)

Another approach would be to create mini-installers for each supported platform which update a new _ca-bundle.crt_ file from either BOINC, https://curl.se or Mozilla. But these require every BOINC user to be aware when a certificate is about to expire (or has expired) and to manually run the mini-installer.

**To Do:**
A chron job must be added to the BOINC server (currently on isaac) to periodically update the file 
`https://boinc.berkeley.edu/ca-bundle.crt` from either https://curl.se or Mozilla. I suggest doing this about every week or two, since https://curl.se updates it on their web site about every 6 weeks or so. This can be done with the following shell script:
`cd /home/boincadm/boinc-site/dl/`
`curl --remote-name --time-cond cacert.pem https://curl.se/ca/cacert.pem`
`if [  $? -ne 0 ]; then`
      `mv -f cacert.pem ca-bundle.crt`
`fi`

See https://curl.se/docs/caextract.html for more information.

### DIFF
--- a/client/http_curl.cpp
+++ b/client/http_curl.cpp
@@ -495,7 +495,7 @@ int HTTP_OP::libcurl_exec(
     if (! boinc_file_exists(CA_BUNDLE_FILENAME)) {
         // If ca-bundle.crt is not in the BOINC Data directory, 
         // look in the old location in the Program Files directory
-        if (strlen(m_curl_ca_bundle_location) == 0) {
+        if (strlen(curl_MSW_ca_bundle_location) == 0) {
             TCHAR szPath[MAX_PATH-1];
             GetModuleFileName(NULL, szPath, (sizeof(szPath)/sizeof(TCHAR)));
 
@@ -504,14 +504,14 @@ int HTTP_OP::libcurl_exec(
                 szPath[pszProg - szPath + 1] = 0;
 
                 strlcat(
-                    m_curl_ca_bundle_location,
+                    curl_MSW_ca_bundle_location,
                     szPath,
-                    sizeof(m_curl_ca_bundle_location)
+                    sizeof(curl_MSW_ca_bundle_location)
                 );
                 strlcat(
-                    m_curl_ca_bundle_location,
+                    curl_MSW_ca_bundle_location,
                     CA_BUNDLE_FILENAME,
-                    sizeof(m_curl_ca_bundle_location)
+                    sizeof(curl_MSW_ca_bundle_location)
                 );
 
                 if (log_flags.http_debug) {
@@ -519,16 +519,16 @@ int HTTP_OP::libcurl_exec(
                         project,
                         MSG_INFO,
                         "[http] HTTP_OP::libcurl_exec(): ca-bundle '%s'",
-                        m_curl_ca_bundle_location
+                        curl_MSW_ca_bundle_location
                     );
                 }
             }
         }
-        if (boinc_file_exists(m_curl_ca_bundle_location)) {
+        if (boinc_file_exists(curl_MSW_ca_bundle_location)) {
             // call this only if a local copy of ca-bundle.crt exists;
             // otherwise, let's hope that it exists in the default place
             //
-            curl_easy_setopt(curlEasy, CURLOPT_CAINFO, m_curl_ca_bundle_location);
+            curl_easy_setopt(curlEasy, CURLOPT_CAINFO, curl_MSW_ca_bundle_location);
             if (log_flags.http_debug) {
                 msg_printf(
                     project,


CharlieFenton:
The only reason this is marked [WIP] is that it is waiting on implementation of the chron job on the BOINC server as described above. @davidpanderson are you willing to do that? I can probably do it but would be more comfortable if you did.

AenBleidd:
@CharlieFenton, I'd abandon this and instead create a code to update certificates in Windows certificate store (if needed). Since this is needed for Windows only, I'd not return to OpenSSL (or hybrid mode).
Speaking of task to automate update of ca-bundle.crt on git - sure, I'll do this.

CharlieFenton:
>  I'd abandon this and instead create a code to update certificates in Windows certificate store (if needed). Since this is needed for Windows only, I'd not return to OpenSSL (or hybrid mode).

@AenBleidd Why is using _schannel_ and the Windows certificate store, which is only for Windows, preferable to using a common code base for all platforms (_OpenSSL_ and _ca-bundle.crt_)?

RichardHaselgrove:
I think I can answer that one: because the MS Windows platform doesn't have any current access to an auto-updating process. That's because (as we established yesterday) it was designed before the introduction of an accessible Windows certificate store, with Windows Vista.

Windows could continue to use OpenSSL, but only if it was updated to at least a new version with LTS. The current v1.0 doesn't meet that criterion.

But, apart from a tip of the hat to the Open Source community, what benefit would a newer OpenSSL bring over schannel?

CharlieFenton:
> If it is desired to return to using curl/ca-bundle.crt, this automatic updating can be enabled for MS Windows by changing line 55 of client/http_curl.cpp from
#define AUTO_UPDATE_MSW_CERTS 0
to
#define AUTO_UPDATE_MSW_CERTS 1

I have moved this to client/client_state.h line 65 so that it can be tested in `CLIENT_STATE::all_projects_list_check()` to avoid downloading ca-bundle.crt when that file is not used on MS Windows.

CharlieFenton:
@RichardHaselgrove Of course I'm assuming that if we switched BOINC for MS Windows back to _OpenSSL_, it would use the latest version of _OpenSSL_. If we reverted the changes in PR #4545 and just updated the Windows build to use the current _OpenSSL_, my code will keep the certificate store up to date for all versions of Windows. All that is needed is to enable my code for Windows by changing 
_#define AUTO_UPDATE_MSW_CERTS 0_
to
`#define AUTO_UPDATE_MSW_CERTS 1`

With schannel, there needs to be separate code to update the Windows certificate store. The advantage is simpler BOINC using a common code base for all platforms.

RichardHaselgrove:
> CLIENT_STATE::all_projects_list_check()

I have an instant gut worry over that. I run my machines with `<dont_contact_ref_site> 1`, because I have a reliable fibre broadband connection, and pinging Google doesn't give me any additional connection failure diagnostics.

But I've been told - #2352 - that setting that flag also deliberately blocks the auto download of all_projects_list - for the benefit of security-minded administrators who do not want their employees to have open access to the full internet.

By that rule, dont_contact_ref_site should also inhibit auto-download of a public ca-bundle.crt. The corporate system admins of such a site will presumably only permit BOINC use on quasi internal projects like WCG. and will have their own ways of achieving that.

RichardHaselgrove:
> we can't update any file in installation directory because of lack of permissions

**False**. An installer could give itself permission. Take a look at [Lunatics Installer v0.46 released for 64-bit Windows](https://setiathome.berkeley.edu/forum_thread.php?id=85144&postid=2030148#2030148). The download link is to my Google Drive, because I prepared and distributed it.

AenBleidd:
> An installer could give itself permission. 
Installer - yes. BOINC client - no. BOINC client might even run as a service with much less permissions as a regular user.
I mean, auto-update by client might be possible if we put this file to BOINC data directory only. But this could lead to another consequences I already wrote in my previous message.

AenBleidd:
@CharlieFenton,

> But isn't the Windows certificate store in a place where BOINC can't update it? This would mean that on unsupported systems such as Windows 7 the user would need to know to run a special updater installer? 

True. But this might be necessary maybe once in next decade or even rarer.

> It seems better to let BOINC do this automatically.

Again, it's too risky for me. If this file is deleted, corrupted, or anything else, BOINC has no way to recover.
And it would be harder to explain people how to recover their BOINC Data directory (and not just run the installer that will do everything for them).

> We could make the rpc that downloads ca-bundle.crt use http:// instead of https://.

And this is a security hole

AenBleidd:
@CharlieFenton, If http connection is intercepted in any way - you can't verify anything, and at least for two weeks you'll have compromised certificates. And this is too much time.
The real question is next: does this really worth to implement this now?
For me it looks like a waste of time, because if this work on the OS that was not updated last 5 years (and I doubt it's used for BOINC at all) - then we're quite safe with Windows 7/8/8.1/10/11.
Reference: https://gs.statcounter.com/os-version-market-share/windows/desktop/worldwide

RichardHaselgrove:
And in the context we're considering it:

> This site can’t provide a secure connection
> asher.tools sent an invalid response.
> ERR_SSL_PROTOCOL_ERROR

CharlieFenton:
> By that rule, dont_contact_ref_site should also inhibit auto-download of a public ca-bundle.crt. The corporate system admins of such a site will presumably only permit BOINC use on quasi internal projects like WCG. and will have their own ways of achieving that.

@RichardHaselgrove Yes, that is how I programmed it.

CharlieFenton:
I may have a solution to the concerns expressed by @AenBleidd. [This article](https://stackoverflow.com/questions/16291809/programmatically-verify-certificate-chain-using-openssl-api) shows that we can use OpenSSL to verify the certificate bundle. If the verification fails, my code can leave the old certificate in place and (as it already does for other errors) set things up to try again one day later.

PR #4559, opened 2021-10-11
Mac: build curl to use Secure Transport instead of OpenSSL

### LINKED ISSUE #4544: Temporary release problems following certificate expiry

CharlieFenton:
Mac: build curl to use Secure Transport instead of OpenSSL so it uses MacOS maintained certificates instead of ca-bundle.crt

Fixes #4544 for Macintosh

### DIFF
--- a/client/http_curl.cpp
+++ b/client/http_curl.cpp
@@ -483,6 +483,19 @@ int HTTP_OP::libcurl_exec(
     curl_easy_setopt(curlEasy, CURLOPT_SSL_VERIFYPEER, 1L);
     //curl_easy_setopt(curlEasy, CURLOPT_SSL_VERIFYPEER, FALSE);
 
+    // MSW now uses schannel and Mac now uses Secure Transport
+    // so neither uses ca-bundle.crt
+#if (!defined(_WIN32) && !defined(__APPLE__))
+    // if the above is nonzero, you need the following:
+    //
+    if (boinc_file_exists(CA_BUNDLE_FILENAME)) {
+        // call this only if a local copy of ca-bundle.crt exists;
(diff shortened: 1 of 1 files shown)


CharlieFenton:
This PR is ready to be merged once it passes checks

AenBleidd:
@davidpanderson, it's unrelated and now should be fixed.
This is an integration test and it was failing because docker container we use for testing contained outdated image with old certificates. Any further build should be fine

CharlieFenton:
@barton2526 I created an updater for the certificates on older builds the last time we had a problem with exprired certs. It is available on the same download_all page and is the last download in the Macintosh section. It contains only a tiny installer .pkg which places a new certificate in the BOINC Data directory, and scripts to set the proper permissions and ownership of the certificate. Since it contains no code, it will work on any architecture Mac which runs OS X (PowerPC, i686, x86_64, even arm64.) It installs a certificate which won't expire until 2028.

At some point it would be nice to create a newer certificate updater using either AppleScript or a shell script that downloads a current certificate from curl.se as explained [here](https://github.com/BOINC/boinc/issues/4544#issuecomment-938217619), renames it to _ca-bundle.crt_, replaces the one in the BOINC Data directory and sets the permissions and ownership. Maybe you would like to do that!

PR #4563, opened 2021-10-18
Mac: update to latest dependent libraries

CharlieFenton:
Mac: update dependent libraries to latest: c-ares-1.17.2, curl-7.79.1, freetype-2.11.0, openssl-3.0.0
Previously updated to wxWidgets-3.1.5. FTGL version ftgl-2.1.3~rc5 is still the current version.

Note the following about OpenSSL 3.0:
- Some OpenSSL APIs used in BOINC are now deprecated.
- EVP_PKEY_get0_RSA() now returns a pointer of type "const struct rsa_st*" to an immutable value. I made this compatible with both OpenSSL 3.0.0 and earlier versions of OpenSSL by casting the returned value to (rsa_st*) as follows:
`rsa = (rsa_st*)EVP_PKEY_get0_RSA(pubKey);`
Do not try to modify the contents of the returned struct.

wxWidgets 3.1.5 now requires MacOS 10.10 for wxWebView and possibly other APIs used by BOINC.


CharlieFenton:
@AenBleidd: 
I wrote:
>How would you change line 674 to declare the variable rsa as as a const?

Actually, I can do that by changing line 674 in lib/crypt.cpp from `RSA *rsa;` to `const rsa_st *rsa;`.
If I do that, then I can eliminate the cast in line 678: `rsa = (rsa_st*)EVP_PKEY_get0_RSA(pubKey);` but the following 3 lines get _no matching function_ compiler errors:
line 679: `if (!RSA_blinding_on(rsa, c)) {`
line 690: `retval = RSA_verify(NID_md5, md5_md, MD5_DIGEST_LENGTH, sfileMsg, sfsize, rsa);`
line 691: `RSA_blinding_off(rsa)`
because those functions are declared without const in the arguments.

This is under _openssl-3.0.0_.  Under _openssl-1.1.1l_  declaring the variable rsa as as a const will generate those errors plus a similar one for line 678 if it does not have the cast.

The way I have already modified line 678 in _lib/crypt.cpp_ works for both openssl-3.0.0 and earlier versions of openssl. It is effectively unchanged from the code that has been in _lib/crypt.cpp_ for many years, and so is no less safe than what we had before.

PR #4583, opened 2021-11-16
Mac: ensure curl is built without needing libraries not normally available

CharlieFenton:
Curl's `make` builds curl so that it requires a number of libraries if `configure` finds those libraries on the computer used for building curl. But those libraries may have been installed on the build machine even though they are not normally present in that OS. Since the BOINC client is not linked with those libraries, the build will probably fail.

Even if the build succeeds, this situation may cause the built executable to crash on computers other than the build machine. 

This PR modifies the _buildcurl.sh_ script to explicitly tell `configure` not to use libraries which are not normally present on MacOS, even if they are present on the build machine.

This PR also simplifies some redundant code in the _buildcurl.sh_ script, and adds a comment in _3rdParty/buildMacDependencies.sh_ explaining that _3rdParty/buildMacDependencies.sh_ is used only for GitHub Continuous Integration, and anyone building BOINC on a Macintosh should use _mac_build/setupForBOINC.sh_ instead.

### DIFF
--- a/3rdParty/buildMacDependencies.sh
+++ b/3rdParty/buildMacDependencies.sh
@@ -2,7 +2,7 @@
 
 # This file is part of BOINC.
 # http://boinc.berkeley.edu
-# Copyright (C) 2017 University of California
+# Copyright (C) 2021 University of California
 #
 # BOINC is free software; you can redistribute it and/or modify it
 # under the terms of the GNU Lesser General Public License
@@ -17,6 +17,9 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 #
+
+## IMPORTANT: This script is used only by GitHub Continuous Integration. Do not use it for 
+## building BOINC on a Macintosh computer! Use mac_build/setupForBOINC.sh instead.
 
 ## support script to build BOINC dependencies on Macintosh
 ## This script checks if cached versions are available and builts them if not.
--- a/3rdParty/buildMacDependencies.sh
+++ b/3rdParty/buildMacDependencies.sh
@@ -145,7 +145,6 @@ mkdir -p 3rdParty/mac
 mkdir -p "${PREFIX}"
 cd "${ROOTDIR}/3rdParty/mac" || exit 1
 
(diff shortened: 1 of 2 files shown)


AenBleidd:
@CharlieFenton, is it possible to unify build scripts to build BOINC in the same way in CI and locally?
It would greatly simply the things, and will reduce redundancy and users confusion.
Also, it will help us to identify issues with local builds also using just CI (this is an idea behind CI to check that everything works as designed without running and testing it locally).
E.g. on Windows we have it exactly like this.

AenBleidd:
@CharlieFenton, btw, if the PR is not ready to be merged, feel free to convert it to draft while you're working on it and testing smth:
image
For now I dis this for you, feel free to mark is as ready when ready:
image

CharlieFenton:
> is it possible to unify build scripts to build BOINC in the same way in CI and locally?

Not easily, because of the way CI needs to use a cache.  Actually, there is very little redundancy; most of the code is already shared between the CI and local builds. The only difference between the CI and local builds is that the CI builds use _3rdParty/buildMacDependencies.sh_ while local builds use _mac_build/setupForBOINC.sh_; these two scripts call the same 6 build scripts (one for each library), but passing different arguments. 

The other main difference is that _3rdParty/buildMacDependencies.sh_ automates the download of the libraries, since that is necessary for CI. While I've considered adding that functionality to _mac_build/setupForBOINC.sh_, I decided against it to give the user more control.

Christian Beer and I put a lot of effort into having as much common code as possible between the two types of builds.

Using _mac_build/setupForBOINC.sh_ allows me to keep multiple versions of each library on my computer so that I can switch among building and testing different versions of BOINC.

CharlieFenton:
Although I had already tested the new _buildcurl.sh_ script, I forced a cache rebuild on CI to ensure that the CI build works properly, since the local and CI builds use that same script. I have now reverted the temporary change which forced a rebuild of the cache, so this PR is ready to be merged into master.

AenBleidd:
Ok, thanks for the answer. For me scripts look fine, so I'm going to merge them once all CI finish

AenBleidd (approved):

Issue #2917, opened 2018-12-19
manager: support Mojave Dark Mode


JuhaSointusalo:
From https://boinc.berkeley.edu/dev/forum_thread.php?id=12755:

Hi, would be wonderful if you add a dark theme to reflect Mojave color. Moreover doing a dark theme avoid drastic color changes while switching between apps.

From https://boinc.berkeley.edu/forum_thread.php?id=12684:

The BOINC menu bar icon for macOS is really awful in the Dark Mode of Mojave…

![image](https://user-images.githubusercontent.com/15806192/50245558-7eb4a580-03db-11e9-983d-5eeb41c46ae7.png)

Because the current one was drawn for a white background.

![image](https://user-images.githubusercontent.com/15806192/50245577-8a07d100-03db-11e9-9657-d25e0b06d363.png)

Is it possible to have another icon displayed in dark mode, to add support for Mojave?
Or at least please a simple gray color of the icon? It will be more convenient.

JuhaSointusalo:
One of the highlights of wxWidgets 3.1.2 released on December 10, 2018 is

- Initial support for macOS 10.14 and its dark mode.

[wxWidgets ticket #18146](https://trac.wxwidgets.org/ticket/18146) seems to track the progress.

CharlieFenton:
I doubt that wxWidgets' code will modify the Manager's taskbar (menu bar) icons for dark mode automatically. I expect someone will need to create new artwork for the three icons (normal, suspended [snooze] and disconnected) in addition to the three we already have. Note also that skins can and do override the default taskbar icons, so this will need to be done for all skins (such as World Community Grid, GridRepublic, etc.) These icons are built using the wxWidgets cross-platform approach of XPM files, which is very different from the way this would be handled in a conventional Macintosh application, so the standard way of doing this on the Mac does not apply here.

There is code [here](https://stackoverflow.com/questions/51672124/how-can-it-be-detected-dark-mode-on-macos-10-14) for detecting when in Dark Mode. Since this is Cocoa code, it should probably be added to _lib/mac/mac_util.mm_. Finally, the relevant code in _clientgui/BOINCGUIApp.cpp_,  _clientgui/SkinManager.cpp_ and _clientgui/BOINCTaskBar.cpp_ and related header files will need to be updated to get and select the correct icons.

[...]

PR #4586, opened 2021-11-23
Use light/dark menubar icons on macOS

TrevorBurnham:
Re: https://github.com/BOINC/boinc/issues/2917

**Description of the Change**
This branch replaces the menubar icons under macOS with new greyscale versions, to be consistent with design standards for macoS Mojave (10.14) and newer. The new menubar icons are higher-resolution and support light/dark mode. I've also increased the size of the `snooze`/`disconnect` indicators to make them more legible.

## Before

Boinc menu bar - Current (light)

Boinc menu bar - Current (dark)

## After

Boinc menu bar - New (light)

Boinc menu bar - New (dark)

**Alternate Designs**

I considered a few variations before coming up with these, and I've tried to make it relatively easy for others to modify them. Here are the steps (replace `boincdisconnect_macmenubar` with the name of the specific icon you want to replace in each step):

1. Open `boinc_macmenubar.psd` in any Photoshop-compatible image editor
2. Select the layers you want (e.g. "Boinc icon (light mode)" + "Disconnect" to generate the `boincdisconnect_macmenubar` icon)
3. Export as PNG to `clientugi/res/boincdisconnect_macmenubar.png`
4. In that dir, run `magick convert boincdisconnect_macmenubar.png -format xpm boincdisconnect_macmenubar.xpm`
5. Open that `.xpm` file in a text editor and change `static char *boincdisconnect_macmenubar[]` to `static const char *boincdisconnect_macmenubar_xpm[]`

---

**This is WIP because the [`wxSystemAppearance::IsDark()`](https://docs.wxwidgets.org/trunk/classwx_system_appearance.html) call doesn't appear to be returning the correct result. In my tests (under macOS Monterey), it consistently returned `false`. Further, just getting the correct value at initialization time isn't sufficient, because we need to be able to change the icon as the system switches between light/dark modes. wxWidgets has that capability (via the [`wxSysColourChangedEvent`](https://docs.wxwidgets.org/3.0/classwx_sys_colour_changed_event.html) class), but wiring it up is a bit of a challenge.

I'd appreciate some assistance with that. It's possible that binding the color change event will address the issue I'm seeing with `IsDark()`. My guess is that `wxWidgets` initially assumes that `IsDark()` is `false`, then updates that internal value later.


clientgui/SkinManager.cpp
```diff
@@ -668,6 +682,18 @@ bool CSkinAdvanced::InitializeDelayedValidation() {
     m_iconApplicationIcon.SetDefaults(wxT("application"), wxT("boinc"));
     m_iconApplicationDisconnectedIcon.SetDefaults(wxT("application disconnected"), wxT("boincdisconnect"));
     m_iconApplicationSnoozeIcon.SetDefaults(wxT("application snooze"), wxT("boincsnooze"));
+#elif __APPLE__
```
TrevorBurnham, line 685: I'm not sure if this is the canonical way to check that we're running under macOS. If there's a better way, please let me know.

clientgui/SkinManager.cpp
```diff
@@ -32,6 +32,7 @@
 #include "MainDocument.h"
 #include "version.h"
 
+#include <wx/settings.h>
```
TrevorBurnham, line 35: This is to bring in `wxSystemSettings`.

CharlieFenton:
> This is WIP because the wxSystemAppearance::IsDark() call doesn't appear to be returning the correct result. In my tests (under macOS Monterey), it consistently returned false

Because the code in BOINC Manager did not support dark mode and it was unusable in Dark Mode, I disabled Dark mode in [here](https://github.com/BOINC/boinc/pull/3536). You need to remove `NSRequiresAquaSystemAppearance` from _clientgui/mac/templates/Info.plist_ to see how it works in Dark mode.

That said, I personally **strongly** dislike using grayscale BOINC icons in the menu bar. Yes, I know that is Apple's "standard" appearance, but I feel the colored icons look much better and make it easier to see the _paused_ and _stopped_ badges. 

In any case, why did you not invert the color of the _pause_ badge on the dark mode icon? It is quite hard to see with black against a black background.

Before you spend too much time on icons, I suggest you please experiment with enabling dark mode by removing `NSRequiresAquaSystemAppearance` form  _Info.plist_ and see if you can solve the larger problems you will encounter.

> I'm not sure if this is the canonical way to check that we're running under macOS. If there's a better way, please let me know.

You need to use `#elif defined(__APPLE__)`

clientgui/SkinManager.cpp
```diff
@@ -43,10 +44,16 @@
 #include "res/skins/default/graphic/workunit_waiting_image.xpm"
 #include "res/boinc.xpm"
 #include "res/boinc32.xpm"
+#include "res/boinc_macmenubar.xpm"
```
AenBleidd, line 47: Please add `#ifdef` guards to every platform-specific includes

CharlieFenton:
@TrevorBurnham Unfortunately, this PR has major incompatibilities with all branded builds of BOINC (GridRepublic, CharityEngine, etc.) and especially with World Community Grid (see _mac_installer/release_brand.sh_.) It is also incompatible with all custom skins (see [this document](https://boinc.berkeley.edu/trac/wiki/ProjectSkin).)

> That said, I personally strongly dislike using grayscale BOINC icons in the menu bar.

Note that all branded builds of BOINC also use full-color menu bar icons.

> You need to use #elif defined(__APPLE__)

Correction: In BOINC Manager source files, please test whether `__WXMAC__` is defined. In all other crop-platform source files, test whether `__APPLE__` is defined.

CharlieFenton:
@TrevorBurnham I will be unavailable to assist you or answer questions until mid-December. Please put this Pull Request on hold until then. In the meantime, I suggest starting a new PR removing `NSRequiresAquaSystemAppearance` from _clientgui/mac/templates/Info.plist_ and working on getting other aspects of dark mode working, both for Simple View and especially Advanced View. In some cases, it may be as simple as inverting foreground and background colors.

Also, please ensure that all APIs you use are either available back to MacOS 10.10 or are properly guarded. You  might find the BOINC function lib/mac_util/compareOSVersionTo(major, minor) useful for this purpose.

TrevorBurnham:
Thanks for all the quick feedback!

I consulted Apple's Human Interface Guidelines for [Menu Bar Extras](https://developer.apple.com/design/human-interface-guidelines/macos/extensions/menu-bar-extras/) (their term for menu bar icons) and realized that the approach they recommend to create icons that look good across light/dark mode is quite different:

> **Use a template image to represent your menu bar extra.** A template image discards color information and uses a mask to produce the appearance you see onscreen. Template images automatically adapt to the user’s appearance settings, so they look good on both dark and light menu bars, and when your menu bar extra is selected.

So we don't need explicit light/dark mode detection in order to make the icon compliant with Apple's standards. Instead, we need to use a template image, which is an [`NSImage`](https://developer.apple.com/documentation/appkit/nsimage) with [`isTemplate`](https://developer.apple.com/documentation/appkit/nsimage/1520017-istemplate) set to `true`. As explained there:

> Images you mark as template images should consist of only black and clear colors. You can use the alpha channel in the image to adjust the opacity of black content, however.

I'm not sure if wxWidgets supports this, unfortunately, so this may take some effort. Additionally, it's not clear to me how the `snooze` and `disconnect` badges would fit in with the template icon. I've seen some menu bar icons (like Adobe Creative Cloud's) use a red badge on top of their monochromatic icon, but I'm not sure how they're doing it.

At any rate, I think that's the right direction to explore, so I'm going to close this out.

---

In the meantime, I've created a separate PR to add a toggle option for the menu bar icon under macOS: https://github.com/BOINC/boinc/pull/4589 That should mitigate this issue, since users who find the colored menu bar icon jarring would be able to simply hide it.

PR #4589, opened 2021-11-24
Add option to hide menu bar icon (macOS-only)

TrevorBurnham:
**Description of the Change**
This PR allows macOS users to hide the menu bar icon, via a new "Show status in menu bar?" option:

Options menu with 'Show status in menu bar' checkbox

This is consistent with Apple's Human Interface Guidelines for [Menu Bar Extras](https://developer.apple.com/design/human-interface-guidelines/macos/extensions/menu-bar-extras/) (their term for the icons that appear in the upper-right corner of the menu bar):

> **Let people decide whether to enable your menu bar extra.** Users, not apps, should choose when a menu bar extra is added to the menu bar. Typically, this is done by changing a setting in an app’s preferences window.

Adding the option should help to mitigate the aesthetic clash between the current colored menu bar icon and those with Apple's recommended monochromatic designs, especially in dark mode (see #2917).

**Alternate Designs**

I considered different phrasings, but "Show status in menu bar" seems most consistent with Apple's standards, as can be seen in these examples from System Preferences (as of macOS 12.0.1):

Show battery status in menu bar
Show Wi-Fi status in menu bar
Show input menu in menu bar

**Release Notes**

Add option to show/hide the Mac menu bar icon (find it under "Other options...")


clientgui/BOINCTaskBar.cpp
```diff
@@ -412,13 +412,19 @@ wxMenu *CTaskBarIcon::CreatePopupMenu() {
...
-    bool result;
     int err = noErr;
     int w, h, x, y;
 
     if (m_iconType != wxTBI_DOCK) {
-        result = wxGetApp().GetMacDockIcon()->SetIcon(icon);
-        return (result && wxTaskBarIcon::SetIcon(icon));
+        bool displayMenuBarExtra = wxGetApp().GetBOINCMGRDisplayMenuBarExtra();
+        if (displayMenuBarExtra) {
+            bool result;
```
AenBleidd, line 421: No need to create this local variable, `wxGetApp().GetMacDockIcon()->SetIcon(icon)` could be moved into return statement.

clientgui/DlgOptions.h
```diff
@@ -57,6 +57,7 @@
 #define ID_ENABLEEXITMESSAGE 10032
 #define ID_ENABLERUNDAEMON 10033
 #define ID_ENABLESHUTDOWNMESSAGE 10034
+#define ID_ENABLEMENUBAREXTRA 10034
```
AenBleidd, line 60: This must have unique ID

AenBleidd (changes requested):

clientgui/DlgOptions.h
TrevorBurnham: Good eye! c5abf9f

clientgui/BOINCTaskBar.cpp
```diff
@@ -412,13 +412,19 @@ wxMenu *CTaskBarIcon::CreatePopupMenu() {
...
 bool CTaskBarIcon::SetIcon(const wxIcon& icon, const wxString& ) {
     wxImage macIcon;
-    bool result;
     int err = noErr;
     int w, h, x, y;
 
     if (m_iconType != wxTBI_DOCK) {
-        result = wxGetApp().GetMacDockIcon()->SetIcon(icon);
-        return (result && wxTaskBarIcon::SetIcon(icon));
+        bool displayMenuBarExtra = wxGetApp().GetBOINCMGRDisplayMenuBarExtra();
```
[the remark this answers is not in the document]
TrevorBurnham: Done: ae543d9c1d

clientgui/BOINCTaskBar.cpp
TrevorBurnham: Done: ae543d9c1d

clientgui/BOINCTaskBar.cpp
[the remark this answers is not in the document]
TrevorBurnham: Done: a301f2f833

CharlieFenton:
When upgrading from an earlier version of BOINC, or when a user installs BOINC for the first time, there will be no _DisplayMenuBarExtra_  entry in the BOINC Manager preferences file. Please make the default to show the menu bar, so that it will be shown in those situations. Otherwise, new users won't even be aware of its existence. 

Also, I don't consider BOINC's menu bar icon and its menu an _extra_. While its functionality is also available from the Dock icon, the menu bar icon's menu is the easiest and most convenient way to control those Manager functions. Other than the menu bar and Dock icon menus, there is no other way to invoke BOINC's _Snooze_ or _Snooze GPU_. Though Apple may refer to these icons as "menu bar _extras_", I've never heard anyone else call them that. In fact, your mention in this PR is the first I've ever heard that Apple calls them that.

I suggest making the tag in the BOINC Manager preferences file _HideMenuBarIcon_ and the associated variable _m_iHideMenuBarIcon_ (rather than _DisplayMenuBarIcon_ and _m_iDisplayMenuBarExtra_) to make it clear in the code that the default is to show BOINC's menu bar icon.

In other words, the menu bar icon should be displayed unless _HideMenuBarIcon_ has a non-zero value in the BOINC Manager preferences file, and unless _m_iHideMenuBarIcon_ has a non-zero value in the code.

CharlieFenton:
> In other words, the menu bar icon should be displayed unless _HideMenuBarIcon_ has a non-zero value in the BOINC Manager preferences file, and unless _m_iHideMenuBarIcon_ has a non-zero value in the code.

It is fine for the dialog to still label the checkbox "Show status in menu bar." I am asking that the value of _HideMenuBarIcon_ in  the BOINC Manager preferences file should be set to the complement of the checkbox value: zero if the checkbox is set and non-zero if the checkbox is clear. The value of _m_iHideMenuBarIcon_ in the code should match the value of _HideMenuBarIcon_  in  the BOINC Manager preferences file.

TrevorBurnham:
@CharlieFenton Sure thing. I've incorporated the changes you requested.

CharlieFenton (approved):

CharlieFenton:
@TrevorBurnham I apologize for the long delay for reviewing this.I got busy with other things and just now remembered this.  I am merging your ode now. Thank you for your contribution, and I hope you will continue to support BOINC development.

Have you thought any more about adding support for dark mode on MacOS? As I wrote [here](https://github.com/BOINC/boinc/issues/2917#issuecomment-968262721) The main problem will probably be that the main BOINC Advanced View displays use _BOINCListCtrl_, which is derived from wxListCtrl, which is a native control only under MS Windows but not on MacOS or Linux. But BOINC code may mostly just require calls to SetForegroundColour and SetBackgroundColour in various placesto invert the colors if in dark mode. Also, you will need to remove `NSRequiresAquaSystemAppearance` from _clientgui/mac/templates/Info.plist_ as I wrote [here](https://github.com/BOINC/boinc/pull/4586#issuecomment-977207570).

A couple of other areas where we could use your help:
* The accessibility code in _clientgui/mac/MacAccessibility.mm_ uses APIs and techniques deprecated by Apple as of MacOS 10.10. While the Simple View uses only native controls so accessibility is handled automatically, the Advanced View has some custom controls (including _BOINCListCtrl_) which need custom accessibility code. Of course, there is no way to know if or when Apple will actually remove support for these APIs.
* The default screensaver graphics _boincscr_ (Xcode target _ss_app_) and those provided by projects are actually command-line apps which use OpenGL and GLUT, and the screensaver engine (_clientscr/*\.\*_) uses _IOSurfaceBuffer_ to communicate with the graphics apps and display them as a MacOS screensaver. _IOSurfaceBuffer_ only works with OpenGL textures. Unfortunately, Apple deprecated OpenGL and GLUT as of MacOS 10.9, so we may need to find suitable third-party OpenGL and GLUT libraries for future compatibility. Again, there is no way to know if or when Apple will actually remove support for these APIs.

Please let me know if you have any questions.

PR #4590, opened 2021-11-24
[Android] Fix armv6 vcpkg build.

AenBleidd:
Use curl 7.80.0 for non-vcpkg Android build.


talregev (approved):
I am approving.

Issue #192, opened 2015-02-03
Working in background cursor when applications are started


romw:
**Reported by Thyme Lawn on 6 Apr 37370793 13:20 UTC**
When a Windows application starts up the default behaviour is to display the working in background cursor for 2 seconds (see the [STARTUPINFO](http://msdn2.microsoft.com/en-us/library/ms686331.aspx) page on MSDN).

The following patch to app_start.C will prevent the cursor from being changed when BOINC starts a science application.

```
445c445,446
<     //startup_info.cb = sizeof(startup_info);
--
>     startup_info.cb = sizeof(startup_info);
>     startup_info.dwFlags=STARTF_FORCEOFFFEEDBACK;
```

This change has already been applied to the CPDN controller program to prevent the cursor from being changed during post-processing (when a program is started 100+ times and was causing the cursor to flicker in a very noticeable and disconcerting fashion).

romw:
**Commented by davea on 8 Dec 37370981 14:13 UTC**
I checked in the suggested change.  - David

Issue #4117, opened 2020-12-01
Client: loses track of new work requirements


RichardHaselgrove:
**Describe the bug**
The client sometimes gets into an internal state, where it makes repeated requests for extra work which bear little resemblance to real needs.

**Steps To Reproduce**
This bug cannot be reproduced on demand, so I've gathered as much data as I can for this occurrence. 

**System Information**
 - OS: Windows 7/64
 - Hardware: Quad-core CPU (i5), 1x Intel_gpu, 2x NVidia GPU
 - BOINC Version: v7.16.11 release - but has been observed over many years with previous versions

**Additional context**
The particular episode under investigation started on 24-Nov-2020, with the machine making 108 work requests between

24-Nov-2020 09:59:40 [Rosetta@home] [sched_op] CPU work request: 65.56 seconds; 0.00 devices
24-Nov-2020 09:59:42 [Rosetta@home] [sched_op] estimated total CPU task duration: 28803 seconds

and

24-Nov-2020 11:07:56 [Rosetta@home] [sched_op] CPU work request: 4697.21 seconds; 0.00 devices
24-Nov-2020 11:07:58 [Rosetta@home] [sched_op] estimated total CPU task duration: 28803 seconds

[Full list](https://github.com/BOINC/boinc/files/5624833/RS_fetch.txt)

108 tasks, estimated at 8 hours each, would have kept all four CPU cores occupied for 9 full days. Rosetta tasks have a three day deadline: the server should never have allocated so many tasks, and the client should never have requested them. My work fetch settings are 0.25 days + 0.05 days (21600.00 + 4320.00 sec)

I completed as many as possible, but the client aborted a substantial number on the third day for "not started by deadline". At the same time, the machine had also over-fetched task for a second CPU project, NumberFields@Home. **Work requests for GPU tasks, from GPUGrid and Einstein, remained normal. All tasks have been displayed normally on the tasks tab in BOINC Manager.**

In previous years, these sporadic overfetch events have stopped whenever I've attempted to gather logs or other evidence. This time, it appears to be stable, so from a test this evening I've gathered

[Work request summary](https://github.com/BOINC/boinc/files/5624985/requests.txt)
[Complete work fetch log](https://github.com/BOINC/boinc/files/5624987/stdoutdae.txt)
[Sample scheduler request](https://github.com/BOINC/boinc/files/5624996/sched_request_boinc.bakerlab.org_rosetta.txt)
[...]

Issue #3065, opened 2019-03-21
Client: review work request issues following #2918 and #3001


RichardHaselgrove:
#2918 (fix job scheduling bug) has been accepted and merged. But in the process, adjustments were made to work fetch which were less than ideal. Quoting from the PR,

https://github.com/BOINC/boinc/pull/2918#issuecomment-450718595: @davidpanderson wrote

> I'm going to stop here with this PR. Maintaining a full buffer in the presence of max_concurrent would require nontrivial changes to the RR simulation which I might do later.

https://github.com/BOINC/boinc/pull/2918#issuecomment-460804035: @JuhaSointusalo wrote

> Even though I merged this I don't consider the job done yet. As previously discussed in my opinion the work fetch code now needs improving.

- If work fetch is blocked as long as there are max_concurrent or more tasks in buffer then before work fetch resumes there has to be an idle device.
- Those who want to use max_concurrent will now have to choose between using max_concurrent and having work buffered. My gut feeling is that people will choose work buffer and that means they won't be able to upgrade BOINC beyond 7.14 as long as the work fetch works the way it does now.

My view remains that the work fetch ceiling should, as now, reflect the sum of the base and additional work buffers, expressed in wall-clock time but proportional to the number of devices in use (max_concurrent rather than nCPUs).

#3001 (fix possible overflow in peak FLOPS) has also been merged, and also fixed in server release 1.0.4. But in [Message Board conversation](https://boinc.berkeley.edu/forum_thread.php?id=12842&postid=90266#90266) I noticed an anomaly, not previously reported on GitHub.

Work fetch is currently determined by [Client Scheduling October 2010 - Estimated credit](https://boinc.berkeley.edu/trac/wiki/ClientSchedOctTen#Estimatedcredit). I compared two separate work fetch logs, the first with the Peak FLOPS overflow bug, and the second without.

```
				RAC		    REC
Einstein@Home		  18,906.01	142,297,886.589
MilkyWay		   4,143.10	    216,419.883
SETI@home		   6,462.70	605,565,081.292

OpenCL: AMD/ATI GPU 0: AMD Radeon(TM) Vega 8 Graphics (43,980,464 GFLOPS peak)
```
The SETI GPU REC has been blown out of the water by the GFlops error. I don't think we'd thought of that before. If that's the only GPU in the system, it shouldn't make much difference - but it would unbalance the work fetch calculations between CPU, other GPUs, and the Ryzen GPU component.
```
				RAC		    REC
[...]

PR #4592, opened 2021-12-01
client: fix work-fetch logic when max concurrent limits are used

davidpanderson:
The round-robin simulation would stop simulating jobs for a project
once a max concurrent limit (app or project) was reached.
As a result it would decide there was a shortfall,
and keep requesting work up to the limit of 1000 jobs.

To fix this:
1) keep simulating a project after an MCL is reached
2) for each (project, resource) pair, keep track of the latest
    simulation time T when an MCL was reached.
3) for such a project, don't request work for a resource if
    T > now + work buf size

This allows us, e.g., to request GPU jobs from a project
even if its CPU jobs (taking MCL into account) fill the buffer.

This works in the simulation case that showed the problem (#192).

Also: add a bit more logging, and improve names

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/client/rr_sim.cpp
+++ b/client/rr_sim.cpp
@@ -328,9 +328,6 @@ void RR_SIM::pick_jobs_to_run(double reltime) {
                 if (have_max_concurrent) {
                     switch (max_concurrent_exceeded(rp)) {
                     case CONCURRENT_LIMIT_PROJECT:
-                        // no more jobs for this project
-                        //
-                        //rsc_pwf.pending_iter = rsc_pwf.pending.end();
                         rsc_pwf.last_mc_limit_reltime = reltime;
                         p->pwf.at_max_concurrent_limit = true;
                         if (log_flags.rr_simulation) {
--- a/client/rr_sim.cpp
+++ b/client/rr_sim.cpp
@@ -88,11 +88,11 @@ static inline void set_bits(
 // refer to RESULT
 //
 struct RR_SIM {
-    vector<RESULT*> active;
+    vector<RESULT*> active_jobs;
 
     inline void activate(RESULT* rp) {
         PROJECT* p = rp->project;
-        active.push_back(rp);
+        active_jobs.push_back(rp);
         int rt = rp->avp->gpu_usage.rsc_type;
 
         // if this is a GPU app and GPU computing is suspended,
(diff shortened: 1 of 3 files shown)


AenBleidd (approved):
Haven't tested this but the code looks fine for me.

RichardHaselgrove:
Because of my interest in #3065 and #4117, I'll download and test the artifact for this PR in the morning: they may both be fixed by these changes (the description of the logic causing the problem sounds right).

RichardHaselgrove:
First check: I've run a new simulation on the existing files in scenario 186. Last year, I commented that the timeline showed an over-fetch from the Rosetta project (https://github.com/BOINC/boinc/issues/4117#issuecomment-738029477) - "Rosetta started with 2 tasks, downloaded 24 tasks, reported 3 tasks, and finished with 23 tasks".

The new run on the same data shows Rosetta downloading four tasks, reporting three, and finishing with three. That's much more appropriate for the project, and very encouraging.

RichardHaselgrove:
@AenBleidd - I see all the right CI actions for the Windows / Release-x64-build (pull_request), culminating in

> Artifact win_client_x64_4d92df4bd7eb20d650facd36d14857df0589b938 has been successfully uploaded!

but it doesn't say where its been uploaded **to**! Could you remind me, please?

AenBleidd:
@RichardHaselgrove, sure.
When all tasks in CI group (Windows in this case) are finished, at the bottom you can find all artifacts of this group.
For this PR they are here:
https://github.com/BOINC/boinc/actions/runs/1528043392
image

RichardHaselgrove:
Ah. I can access all of those from any of the [OS] Tests Reports details links, but I was put off by the OSX label on all lines. I've got it now - I'll try and set up a local live test a bit later.

RichardHaselgrove:
Live test is running on one machine: no problems observed so far, but I'll keep monitoring it.

I tried, but failed, to re-create the 'infinite fetch' bug with the current v7.16.20 client. I'll have another go tomorrow with a different machine. (I'm running a different mix of projects compared with last year, and some of them have longer run-times, which may account for some of the difficulty)

RichardHaselgrove:
Finally got the second machine (the one I used for #4117) to reproduce the problem I had then - fetching the same amount of work repeatedly each time the server backoff permitted it, and disregarding the work successfully allocated. I confirm that replacing the v7.16.20 client with the artifact from this PR corrects that problem. 

In the absence of any contrary reports, I think we're good to go.

hucker75:
Should this have fixed the problem in Android Boinc?  Or did it only go to Windows?  I'm getting too much work downloaded for Einstein, which is limited to 4 cores by app config, when the other project Universe is unable to give out work.

AenBleidd:
Work fetch logic is the same on all platforms

hucker75:
Then something is up, I'm using the latest Android version from https://boinc.berkeley.edu/download_all.php on an Android 11 Samsung A10.  It's attached to Universe and Einstein.  Recently, Universe had a power failure, so it was only able to get Einstein.  But I've limited Einstein on that phone to 4 of the 8 cores, or it runs out of RAM and crashes.  It downloaded 50 days of work, when I'd requested 0+1, and the deadline was 14 days.

hucker75:
Thanks for letting me know.  Can I get a Beta release of Android anywhere?

AenBleidd:
There is no Beta release currently. I'll let you know when we start it

PR #4600, opened 2021-12-28
Add "Suspend when no mouse/keyboard input in last XX minutes" to local manager

### LINKED ISSUE #4153: Add "Suspend when no mouse/keyboard input in last XX minutes" to local client
**Describe the problem**
Most of the preferences are duplicated from the website to the local program. When setting "Computing preference - Using local overrides", the local preferences do not have the option:
Suspend when no mouse/keyboard input in last XX minutes 
This seems to be the only option that is not in both locations. In this case, even though the system may be using Local overrides, it still goes to the website to get this one preference.

Implement the "Suspend when no mouse/keyboard input in last XX minutes" preference in the local client and do not go to website if local overrides are being used.

**Additional context**

image
image

Vulpine05:
Fixes #4153

**Description of the Change**
Allows the suspend if no recent input to be modified from BOINC Manager.  I have tested this on a Windows 10 system, but no other platforms.  Since this is my first PR for BOINC, any additional testing would be appreciated.  I am not familiar enough with the Android build, so I don't know how these changes may affect that platform, if it even does.

Besides being a feature request, this also fixes a bug.  If a computer has web preferences set to a value >0 for suspend if not recent input, the computer will use this web preference.

Tentative procedure to reproduce bug:
1. Make sure web preferences for the project have "Suspend when no mouse/keyboard input in last:" = 0.
2. Run BOINC.
3. Set local preferences.  I assume the combination of settings is (mostly) irrelevant.
4. Under the project's web preferences, change "Suspend when no mouse/keyboard input in last:" to a nonzero value.  Be sure to change the setting for location the test computer is assigned to (default, home, school, work).  Save settings.
5. Under BOINC Manager->Projects, update the project where the web preferences were updated.
6. Wait the assigned time, and BOINC should suspend after the computer is idle for that period of time.

NOTE:  As I have tried to test this bug, I discovered the <suspend_if_no_recent_input> value would either remain zero, and/or global_prefs.xml would sometimes not update.  I need to do more testing why this is happening, but I don't think it is related to this issue or PR.

**Alternate Designs**
To bypass the bug, the user could either be sure the suspend if no recent input value is zero for the projects on the user account, or they can manually add <suspend_if_no_recent_input>0</suspend_if_no_recent_input> to the global_prefs_override file.  Neither seem as user-friendly.

**Release Notes**
Add "Suspend when no mouse/keyboard input in last XX minutes" to local manager.


CharlieFenton:
Several things:
[1]
> Under the project's web preferences, change "Suspend when no mouse/keyboard input in last:" to a nonzero value. 

Once you set local preferences, web preferences are ignored by design. This is correct and intended behavior, so that some of your computers can have different computing preferences while others follow your global (i.e., web-based) preferences. There is a warning on the dialog telling you this. I'm pretty certain that we don't want this changed.

[2] I don't understand how is your new added preference "Suspend when no mouse/keyboard input in last XX minutes" adds anything different from these 3 existing lines in the Computing Preferences dialog?
```
Suspend when computer is in use
Suspend GPU when computer is in use
'In use' means mouse/keyboard input in last xx minutes
```

Why is 100 minutes not long enough to suspend BOINC after computer use? Note that most operating systems provide a way for applications to mimic computer use (as if the user used the mouse or keyboard) while watching movies, etc.

[3] I wonder if you misunderstood this setting, when you wrote in your PR #6401:

> In my opinion, 1 hour 40 minutes does not seem to be a high enough limit before the computer should suspend. I could see the possibility of someone wanting to let a computer run 2, 4, maybe 8 hours before wanting it to suspend?

The time is not how long you wait for BOINC to suspend. It is the time before BOINC resumes after user activity ceases.

[4] I just looked at PR #6401, and it appears to me that BOINC already does exactly what that PR requests. I believe the poster did not understand the dialog. I have responded there.

davidpanderson:
Charlie, there are two prefs that sound similar but are different:

A) suspend when the computer is in use
B) suspend when the computer *hasn't* been in use for a while
  (e.g. so the computer can hibernate)

Currently B) can be set in web prefs but not local prefs;
that was an oversight, and should be fixed
(which is what this PR does)

Vulpine05:
@CharlieFenton, thank you for the questions.  Responses are below.

> Once you set local preferences, web preferences are ignored by design. This is correct and intended behavior, so that some of your computers can have different computing preferences while others follow your global (i.e., web-based) preferences. There is a warning on the dialog telling you this. I'm pretty certain that we don't want this changed.<

You are correct, we want the local preferences to ignore the web preferences.  However, as I was preparing to fix this issue, I needed to validate that the actual preference (web or local) worked.  As I was testing, I discovered that for this particular preference, changing the web preference would sometimes affect the computer even if it was using local preferences.  What I discovered is that BOINC not only didn't have a means to display this particular preference, but it did not have a means to read, save, or validate the setting.  My assumption is because this was not done at the local level, the web setting is being used instead.

All I wanted to highlight was that I think this PR isn't just a nice feature to have, it fixes a problem that others may encounter if they have a mix of web and local preferences, and want to use the suspend if no recent input feature for web preferences only.

Here is a better procedure to reproduce the problem:
1.  Change computer's preferences to web preferences (I used Einstein@home in this case).
2.  Change the web preferences' "suspended when no mouse/keyboard input in last:" to a nonzero number (I used .15 minutes).  Save preferences.
3.  Update the project in BOINC Manager.
4.  For some reason, I couldn't get BOINC to suspend with this web preference value.  I closed BOINC, and relaunched it a few minutes later.
5.  Observe the computer suspended after leaving idle for the period of time.
6.  Under Computing Preferences, revise to local preferences.  Save changes and accept.
7.  Observe the computer suspend, even though the computer has local preferences and BOINC 7.16.20 does not have a means to modify this preference.
8.  Under the web preferences, change this value back to zero.  Save preferences.
9.  Update the project where web preferences were changed.
10.  Observe that the computer will NOT suspend after the idle time passes, even though the computer is still using local preferences.

As for your other questions, I think @davidpanderson answered this.  Please ignore my Issue #4601.  I set the range on the local preferences from 0 to 100.  I thought I was mimicking what was in the web preferences, but after I checked again, I found the range was 0 to 9999 (I'm going off the top of my head right now).  Regardless, a later commit to this PR corrects it to mimic the same time range as "'In use' means mouse/keyboard input in last xx minutes".

AenBleidd (approved):
Looks good to me. Thanks for the PR.

image

PR #4603, opened 2022-01-13
[vboxwrapper]Enable vboxwrapper to use multiattach mode disk images

computezrmle:
Required in connection with feature request BOINC#4602

**Description of the Change**
Enables vboxwrapper to deal with multiattach mode disk images and differencing images as described in the feature request.
Additional settings need to be done by the project admin who wants to use this functionality.
Also described in the feature request.
The traditional use of disk images is still there and remains the default.

**Alternate Designs**
None.
It's an enhancement based on a previously unused VirtualBox feature.

**Release Notes**
N/A


AenBleidd:
Windows build fail is expected, waiting for the fix to be merged: https://github.com/microsoft/vcpkg/pull/22509

computezrmle:
Who is responsible to trigger manual page changes?


Since the PR passed all automatic test I assume it may be merged into the master some day.
In that case some manual pages will/may also need to be updated.

I'm currently aware of the following pages.

https://boinc.berkeley.edu/trac/wiki/VboxApps#TheVboxjobdescriptionfile
Will need an update since the PR introduces a new tag to be used in vbox_job.xml:

https://boinc.berkeley.edu/trac/wiki/AppVersionNew#Theversiondescriptionfile
May need an update since project admins should not use the <copy_file/> tag if differencing images are used.


Where should the following hints for project admins go to?

When an image file is ready to be send out, the new method can safely be used.
Do not use differencing images while image file creation is in progress since
all changes would go to the differencing image.

AenBleidd:
@computezrmle, did you test this on Windows? Artifacts are available on the CI

computezrmle:
@AenBleidd No I did'nt. I don't have a Windows system where I can test it.

computezrmle:
Question to the workflow experts.
I'd like to add few lines to a file that is part of this PR but not yet merged.
What is the suggested way to do this?

AenBleidd:
Create a new branch from this branch

computezrmle:
Minor changes regarding the benchmark files were planned to be part of a separate PR.
They are now here by accident, sorry.

The changes avoid gcc compiler warnings caused by
- uninitialized variables
- use of C++17

client/dhrystone.cpp
```diff
@@ -90,7 +90,7 @@ int dhrystone(
     unsigned long         Loops;
     DS_DATA dd;
 
-    double                  startclock, endclock;
```
AenBleidd, line 93: This is not related to this PR.
Please create a separate branch and a separate PR with this change.

client/dhrystone.h
```diff
@@ -1,5 +1,5 @@
 
-#ifdef __APPLE__
```
AenBleidd, line 2: What exact warning is fixed by this?
Also, it's not related to this particular PR, so should be moved to another PR.

client/whetstone.cpp
```diff
@@ -98,7 +98,7 @@ int whetstone(double& flops, double& cpu_time, double min_cpu_time) {
     SPDP x,y,z;
     long j,k,l, jjj;
     SPDP e1[4];
-    double startsec, finisec;
```
AenBleidd, line 101: This is not related to this PR.
Please create a separate branch and a separate PR with this change.

AenBleidd (changes requested):

client/dhrystone.h
computezrmle: Reverted back and resubmitted the change via a separate PR.

client/dhrystone.cpp
computezrmle: Reverted back and resubmitted the change via a separate PR.

AenBleidd:
Tested on Windows 10 with VBox 6.1.32 on Rosetta with anonymous platform and manually removed `<copy_file/>` flag from the tasks options.
@davidpanderson, @computezrmle, please don't forget to update the doc.

davidpanderson:
I'm preparing to test this feature.  @computezrmle , can you please tell me how to convert our current minimal Linux VM image,
which is here:
https://boinc.berkeley.edu/dl/vmimage_x64.zip
to a multi-attach image?  Oracle's docs are not clear.

computezrmle:
To see what happens I suggest to do the following steps via the VirtualBox GUI:

1. Extract vmimage_x64.vdi from the zip
2. Create a new VM (Linux guest as it looks like the vdi contains a Linux system)
3. Attach vmimage_x64.vdi to the new VM (this sets the disk mode to "Normal" to stay compatible with VirtualBox 4.3)
4. From the VirtualBox Manager open File -> Tools -> Virtual Media Manager
5. Select the entry for vmimage_x64.vdi and click on the properties icon
6. Change the attribute "Type" to "Multi-attach"

Done

When you start the VM for the first time VirtualBox will open vmimage_x64.vdi read-only (as "parent disk") and it will automatically create a writable differencing vdi file below the VM's standard directory.
That way other VMs can use the same parent disk as each VM gets it's own differencing image.
As long as a differencing image is registered as child of the parent disk the disk mode can't be set back to "Normal".
From the perspective of the guest OS the split into 2 vdi files is completely hidden.


The steps from above can also be executed via vboxmanage.
That's how it is implemented in vboxwrapper and used by ATLAS and CMS from LHC@home.

PR #4609, opened 2022-01-22
[Android] Parse project disk usage

talregev:
Parse project disk usage.
Until now, never use the disk usage, although we get the information from the state xml.

### DIFF
--- a/android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/ProjectsParser.kt
+++ b/android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/ProjectsParser.kt
@@ -165,6 +165,8 @@ class ProjectsParser : BaseParser() {
                     mProject.noCUDAPref = mCurrentElement.toString() != "0"
                 } else if (localName.equals(Project.Fields.NO_ATI_PREF, ignoreCase = true)) {
                     mProject.noATIPref = mCurrentElement.toString() != "0"
+                } else if (localName.equals(Project.Fields.DISK_USAGE, ignoreCase = true)) {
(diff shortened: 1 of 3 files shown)


android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/ProjectsParser.kt
```diff
@@ -165,6 +165,8 @@ class ProjectsParser : BaseParser() {
                     mProject.noCUDAPref = mCurrentElement.toString() != "0"
                 } else if (localName.equals(Project.Fields.NO_ATI_PREF, ignoreCase = true)) {
                     mProject.noATIPref = mCurrentElement.toString() != "0"
+                } else if (localName.equals(Project.Fields.DISK_USAGE, ignoreCase = true)) {
+                    mProject.diskUsage = mCurrentElement.toDouble()
```
AenBleidd, line 169: Please update unit test to include this new line

AenBleidd (changes requested):

android/BOINC/app/src/main/java/edu/berkeley/boinc/rpc/ProjectsParser.kt
talregev: Can you have a link to upper lines that have a unit test?

AenBleidd: Should be this: https://github.com/BOINC/boinc/blob/master/android/BOINC/app/src/test/java/edu/berkeley/boinc/rpc/ProjectInfoParserTest.kt

AenBleidd (approved):

Issue #2662, opened 2018-08-26
Mac screensaver app can not be built on Travis CI


ChristianBeer:
The Xcode ss_app target can not be built on Travis CI because of missing symbols. The error message seems to point to freetype or FTGL or some other missing static library. Since this is not release critical (it only happens on the Travis OSX VM) I just disabled the target but it would be nice if someone can investigate and fix it.

```
Undefined symbols for architecture x86_64:
  "_png_error", referenced from:
      _read_data_from_FT_Stream in libfreetype.a(sfnt.o)
  "_png_create_info_struct", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_set_read_fn", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_read_info", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_set_palette_to_rgb", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_get_error_ptr", referenced from:
      _error_callback in libfreetype.a(sfnt.o)
      _read_data_from_FT_Stream in libfreetype.a(sfnt.o)
  "_png_create_read_struct", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_set_longjmp_fn", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
      _error_callback in libfreetype.a(sfnt.o)
  "_png_set_tRNS_to_alpha", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_set_expand_gray_1_2_4_to_8", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_set_strip_16", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_read_image", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_set_packing", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_set_gray_to_rgb", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_set_interlace_handling", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
  "_png_read_update_info", referenced from:
      _Load_SBit_Png in libfreetype.a(sfnt.o)
[...]

PR #4613, opened 2022-01-23
[CI][OSX] Check boinc binary is created

talregev:
Check boinc osx binary is created in ci

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -35,6 +35,10 @@ jobs:
         if: success()
         run: ./mac_build/buildMacBOINC-CI.sh --no_shared_headers
 
+      - name: test boinc binary created
+        if: success()
+        run: if ! [ -e mac_build/build/Deployment/boinc ]; then exit 1; fi
+
       - name: deploy
         if: ${{ success() }}
         run: echo "Would start deployment now"


AenBleidd:
No need to check the binary. Instead you should make script to fail when build fails.

talregev:
> No need to check the binary. Instead you should make script to fail when build fails.

The build is not fail when the binary is not created.

AenBleidd:
> > No need to check the binary. Instead you should make script to fail when build fails.
> 
> 
> 
> The build is not fail when the binary is not created.

Yes, that's why build script should be fixed instead of flow YML file

talregev:
I will rebase from the master 
@AenBleidd 
Now the issue is fix, can you add this check?

AenBleidd:
There is still an issue with build script somewhere. Once I find some time - I'll make the investigation of it. I still don't like this check because:
- it checks for one binary only, but we are building several ones
- this is like an exclusive verification for OSX only
- this doesn't really fix the issue with the build script
Maybe I'll merge this as a temporary solution but since it's a low prio ticket, I'll keep it here for a little bit more.

talregev:
> There is still an issue with build script somewhere. Once I find some time - I'll make the investigation of it. I still don't like this check because:
> 
> * it checks for one binary only, but we are building several ones
> * this is like an exclusive verification for OSX only
> * this doesn't really fix the issue with the build script
>   Maybe I'll merge this as a temporary solution but since it's a low prio ticket, I'll keep it here for a little bit more.

* I can import the python code of prepare upload , and create python check for all binaries.
* I don't mind to check all os system. I just encounter first time on mac os.
* I agree, it not fix the issue, but it catches the problem on time.
The macos create the binary in the master?

AenBleidd:
> The macos create the binary in the master?

Didn't get your question.

talregev:
In the master, we are create the mac os binary or there is some issue and it still not created?

AenBleidd:
> In the master, we are create the mac os binary or there is some issue and it still not created?

I haven't check logs for a while but that particular issue should be already fixed when I dropped crappy cache

talregev:
> > In the master, we are create the mac os binary or there is some issue and it still not created?
> 
> I haven't check logs for a while but that particular issue should be already fixed when I dropped crappy cache

ok. thank you. I will rebase this pr and we will find out.

AenBleidd:
I can't accept this PR in its current state because there are more than 1 binary that is produced during the build, and if we want to make some kind of a general solution, we should check for all binaries and not the one that was faulty several months ago.
Also, we still have no artifacts uploaded for OSX builds, so this is might be a good point to do that (in a separate PR of course).

And the main point why I don't want to merge it right now is next: we are not using CI to produce release binaries. And if script is silently failing - it might still fail when building locally, and the check in this PR will not help to catch such a situation.
All out other platforms have no such a special check for created binaries and they are failing in their build flows (that are common both for CI and local release build) when smth goes wrong.
And I don't know how to proceed with this because I don't want to add such a check into build script, because this is an indicator of flaky script that should instead fail during the build process if smth went wrong.

So going back to this PR, what exactly we want to achieve here? This will not make build script more stabler. Yes, this might signal about the problem with generated binaries, but then it looks like a temporary solution before we make a proper fix of build script.

Personally for me it's better to solve the original issue itself than try solve aftermaths.

So I will keep it unmerged for now and get back to it later.

talregev:
Can you tell me all the binaries we create on mac os? 
similar to this list? 
https://github.com/BOINC/boinc/blob/master/deploy/prepare_deployment.py#L22

talregev:
This PR don't intend to solve the issue, it intend to detect it, and other issues.
Even after you fix it, the bug can return, or other bug that will not create a binary and it not fails. 
So this is let us know quickly.

AenBleidd:
> Can you tell me all the binaries we create on mac os?

I have no this information. You should find out this by yourself by exploring OSX build scripts and logs.
Please also be aware about this issue: https://github.com/BOINC/boinc/issues/2662

CharlieFenton:
The MacOS BOINC Xcode project is quite complex. It builds 22 binaries: 16 executable applications, one plugin (for the screensaver engine) and 5 libraries. But most of the applications are not usable by themselves. They must be combined in very specific ways and incorporated in the _BOINC_Installer_ (which is one of the 16 applications) using the _mac_installer/release_boinc.sh_ script. For example, the _gfx_switcher_ and _gfx_cleanup_ apps are helper apps for the screensaver engine and are embedded as resource inside the _BOINCSaver_ plugin. The applications _boinccmd_ and _AddRemoveUser_ are utilities which allow users to manage BOINC. The SetVersion application is used during the build process to set the BOINC version number in the metadata for the various components.

This PR does not say the problem you are trying to solve. What exactly are you trying to accomplish with this PR? Are there cases where the CI check says it succeeded when it failed to create a binary? 

The compiler returns an error when it does not generate a binary and CI should recognize the error from the compiler. CI should be using the _mac_build/BuildMacBOINC-CI.sh_ script. @ChristianBeer adapted it from the build script _mac_build/BuildMacBOINC.sh_ which I created to make it easy for people to do their own build. If _mac_build/BuildMacBOINC-CI.sh_ is not returning a non-zero value when there is an error, you should fix the script.

AenBleidd:
@CharlieFenton, thanks for the explanation of the OSX build process.

> If mac_build/BuildMacBOINC-CI.sh is not returning a non-zero value when there is an error, you should fix the script.

And that's exactly what I plan to do next when I clean up my table a little bit.

> It builds 22 binaries: 16 executable applications, one plugin (for the screensaver engine) and 5 libraries.

I believe, we can still upload them as a build artifacts even if they are not generally usable (in any case, that's free for us).
At some point this PR might be a starting point of making another one to upload these binaries to GitHub but I still need to think how to do this the best way.

talregev:
> The compiler returns an error when it does not generate a binary and CI should recognize the error from the compiler.
The script will not fail if there error in the binary file, this is the issue now. Even if you fix the issue, this bug can return again, and that why we need to check if it create the binary always. You can choose any other way to check if there is a binary creation, but not if the compilation script fail or not. That why I implement this simple check.

PR #4615, opened 2022-01-27
Add collect data app

talregev:
Add collect data app

### DIFF
--- a/deploy/prepare_deployment.py
+++ b/deploy/prepare_deployment.py
@@ -27,6 +27,7 @@ linux_client_list = [
 
 linux_apps_list = [
     './samples/condor/boinc_gahp',
+    './samples/collect_data/collect_data',
     './samples/example_app/uc2',
     './samples/example_app/ucn',
     './samples/example_app/uc2_graphics',
@@ -86,6 +87,12 @@ android_apps_list = [
     './samples/condor/android_arm64_boinc_gahp',
     './samples/condor/android_x86_boinc_gahp',
     './samples/condor/android_x86_64_boinc_gahp',
+    # collect_data
+    './samples/collect_data/android_armv6_collect_data',
+    './samples/collect_data/android_arm_collect_data',
+    './samples/collect_data/android_arm64_collect_data',
+    './samples/collect_data/android_x86_collect_data',
+    './samples/collect_data/android_x86_64_collect_data',
     # uc2
     './samples/example_app/android_armv6_uc2',
     './samples/example_app/android_arm_uc2',
@@ -149,6 +156,7 @@ windows_apps_list = [
     './win_build/Build/x64/Release/multi_thread*.exe',
     './win_build/Build/x64/Release/test*.exe',
     './win_build/Build/x64/Release/wrappture*.exe',
+    './win_build/Build/x64/Release/collect_data*.exe',
     './win_build/Build/ARM64/Release/htmlgfx*.exe',
     './win_build/Build/ARM64/Release/wrapper*.exe',
     './win_build/Build/ARM64/Release/vboxwrapper*.exe',
@@ -160,7 +168,8 @@ windows_apps_list = [
     './win_build/Build/ARM64/Release/boinclog.exe',
     './win_build/Build/ARM64/Release/multi_thread*.exe',
     './win_build/Build/ARM64/Release/test*.exe',
-    './win_build/Build/ARM64/Release/wrappture*.exe'
+    './win_build/Build/ARM64/Release/wrappture*.exe',
+    './win_build/Build/ARM64/Release/collect_data*.exe',
 ]
 
 windows_client_list = [
--- a/samples/collect_data/collect_data.cpp
+++ b/samples/collect_data/collect_data.cpp
@@ -1,0 +1,196 @@
+// This file is part of BOINC.
+// http://boinc.berkeley.edu
+// Copyright (C) 2021 University of California
+//
+// BOINC is free software; you can redistribute it and/or modify it
+// under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation,
+// either version 3 of the License, or (at your option) any later version.
+//
+// BOINC is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
+
+// Application for testing non-CPU-intensive features of BOINC.
+// TODO: make it exit after a certain amount of time
+
+#include <sstream>
+#include <fstream>
+#include <iostream>
+#include <string>
+#include <chrono>
+#include <thread>
+#include <sockpp/socket.h>
+#include <sockpp/tcp_connector.h>
+#ifndef WIN32
+    #include <sockpp/unix_connector.h>
+#endif
+
+void my_sleep(std::chrono::nanoseconds nanoseconds) {
+    std::this_thread::sleep_for(nanoseconds);
+}
+
+sockpp::connector* get_connected(sockpp::connector* array_connection[], unsigned int arrSize) {
+    if (NULL == array_connection || 0 == arrSize) {
+        return NULL;
+    }
+    for (unsigned int i = 0; i < arrSize; ++i) {
+        sockpp::connector* conn = array_connection[i];
+        if (conn != NULL && conn->is_connected()) {
+            return conn;
+        }
+    }
+    return NULL;
+}
+
+bool is_connected(sockpp::connector* array_connection[], unsigned int arrSize) {
+    if (NULL == array_connection || 0 == arrSize) {
+        return NULL;
+    }
+    for (unsigned int i = 0; i < arrSize; ++i) {
+        sockpp::connector* conn = array_connection[i];
+        if (conn != NULL && conn->is_connected()) {
+            return true;
+        }
+    }
+    return false;
+}
+
+int main(int, char**) {
+    //boinc_init();
+
+    sockpp::socket_initializer sockInit;
+    sockpp::tcp_connector tcp_conn;
+    sockpp::connector* array_connection[] = { &tcp_conn, NULL };
+#ifndef WIN32
+    sockpp::unix_connector unix_conn;
+    array_connection[1] = &unix_conn;
+#endif
+    sockpp::connector* conn = NULL;
+    const unsigned int arrSize = sizeof(array_connection) / sizeof(array_connection[0]);
+    const bool isDebug = false;
+    const int port = 31416;
+    const int buffer_size = 4096;
+    const int wake_minutes = 30;
+    const int exec_hourly = 5;
+    int suffix_file_idx = 0;
+    const std::string android_address = "edu_berkeley_boinc_client_socket";
+    const std::string unix_address = "boinc_socket";
+    const std::string tcp_address = "localhost";
+    const std::string prefix_file_name = "boinc";
+    const std::chrono::minutes wake_every_minutes = std::chrono::minutes(wake_minutes);
+    const std::chrono::minutes exec_time_hours = std::chrono::minutes(exec_hourly);
+    
+    std::stringstream stream;
+    char buffer[buffer_size];
+    const std::chrono::time_point<std::chrono::system_clock> starttime = std::chrono::system_clock::now();
+    while (true) {
+        int attemped_connections = 0;
+        while (!is_connected(array_connection, arrSize)) {
+            ++attemped_connections;
+            std::cout << "attemped connections: " << attemped_connections << "\n";
+            if (!tcp_conn.connect(sockpp::inet_address(tcp_address, port))) {
+                std::cout << "tcp connection fail:\n";
+                std::cout << tcp_conn.last_error_str() << "\n";
+                tcp_conn.clear();
+            }
+#if! _WIN32
+            if (!is_connected(array_connection, arrSize) && !unix_conn.connect(sockpp::unix_address(unix_address))) {
+                std::cout << "unix connection fail:\n";
+                std::cout << unix_conn.last_error_str() << "\n";
+                unix_conn.clear();                
+            }
+
+            if (!is_connected(array_connection, arrSize) && !unix_conn.connect(sockpp::unix_address(android_address))) {
+                std::cout << "android unix connection fail:\n";
+                std::cout << unix_conn.last_error_str() << "\n";
+                unix_conn.clear();                
+            }
+#endif
+            if (is_connected(array_connection, arrSize)) {
+                conn = get_connected(array_connection, arrSize);
+            }
+            std::cout << "\n";
+            my_sleep(std::chrono::seconds(1));
+        }
+
+        const std::string request = "<boinc_gui_rpc_request><get_state/></boinc_gui_rpc_request>\3";
+        if (isDebug) {
+            std::cout << "before NULL == conn: " << conn << "\n";
+        }
+        if(NULL == conn) {
+#ifndef WIN32
+            unix_conn.close();
+#endif
+            tcp_conn.close();
+            continue;
+        }
+        if (isDebug) {
+            std::cout << "before write: \n";
+        }
+        const ssize_t res = conn->write(request);
+
+        if (isDebug) {
+            std::cout << "after write: " << res << "\n";            
+        }
+        if (res == -1) {
+            conn->close();
+#ifndef WIN32
+            unix_conn.close();
+#endif
+            tcp_conn.close();
+            continue;
+        }
+        const std::chrono::time_point<std::chrono::system_clock> before_sleep = std::chrono::system_clock::now();
+        const time_t in_time_t = std::chrono::system_clock::to_time_t(before_sleep);
+        std::string time_str(20, '\0');
+        std::strftime(&time_str[0], time_str.size(), "%Y-%m-%d %H:%M:%S", std::localtime(&in_time_t));
+        std::cout << "now: " << time_str << "\n";
+        std::cout << "request bytes: " << request.size() << "\n";
+        std::cout << "write bytes: " << res << "\n";
+
+        if (isDebug) {
+            std::cout << "after write, before read: \n";
+        }
+
+        size_t n = buffer_size;
+        stream.str("");
+        while (n == buffer_size) {
+            n = conn->read(buffer, sizeof(buffer));
+            if (isDebug) {
+                std::cout << "result size: " << n << "\n";
+                std::cout << "buffer: " << buffer << "\n";
+            }
+            if (n > 0) {
+                std::string str(buffer, buffer + n);
+                stream << str;
+            }
+        }
+        std::string str = stream.str();
+        str = str.substr(0, str.size() - 1);
+        std::cout << "read bytes: " << str.size() << "\n";
+        if(str.empty()) {
+            continue;
+        }
+        std::ofstream outFile;
+        std::stringstream file_name;
+        file_name.str("");
+        file_name << prefix_file_name << suffix_file_idx << ".txt";
+        outFile.open(file_name.str());
+        outFile << str;
+        std::cout << "write to file: " << prefix_file_name << suffix_file_idx << ".txt\n\n";
+        ++suffix_file_idx;
+        outFile.flush();
+        outFile.close();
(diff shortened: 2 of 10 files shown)


AenBleidd:
1. I have doubts this sample is useful, so it should be discussed first.
2. Code contains mixed styles including c++11/14/17 code that is not currently allowed in this repo.
3. In case of positive decision of point 1, this PR requires careful review and testing.

For now I moved this PR to draft to coma back to it later

talregev:
Please highlight all the places that there is c++11 style code

AenBleidd:
FIrst it should be decided if this is needed at all, otherwise it will be just a waste of time to do such review, including architecture review (e.g. it's not clear how long this task should run: any task including NCI mustn't be infinite, because  then there will be nothing reported back to the server).
Also, I didn't find any code related to communication with running BOINC client.

> Please highlight all the places that there is c++11 style code

You can do this by yourself enforcing particular c++ standard usage (e.g. in gcc pass `-std=c++03 -pedantic-errors` to build command)

@talregev, please do a proper testing of this PR, because then it will save time for all of us that we might spend on review.

Thanks in advance.

talregev:
> 

I already test this many time, and run it on various os including linux, windows and android.
I didn't write the code in c++11 style. so if there any, you should highlight it.
After we clear that, I can give some of the reasons why this app is useful.

talregev:
This is an example app. this code doesn't going to production. when it does, we can decide all the parameters, including what to save.

AenBleidd:
> I didn't write the code in c++11 style. so if there any, you should highlight it.

Please do this by yourself. Why you want me to spend my time to do a review if you can easily do this by yourself?
Obvious stuff that is not in c++03:
- auto -> part of the c++11
- constexpr -> part of c++17

Please compile this application with the parameters I sent you.

AenBleidd:
> This is an example app. this code doesn't going to production. when it does, we can decide all the parameters, including what to save.

It still should be complete to show your intention. In current state this PR is not ready.

talregev:
The main reason I am doing this example. 
There is only one project that collect data, and it unofficial. and also it doesn't work on android when the Android remote is activate. 
This app example should be in boinc and use as a beacon to write a new project / official app by other.

AenBleidd:
> The main reason I am doing this example. 
> 
> There is only one project that collect data, and it unofficial. and also it doesn't work on android when the Android remote is activate. 
> 
> This app example should be in boinc and use as a beacon to write a new project / official app by others. 

Then it should work OOTB.

And I'm still don't agree that it's really needed in BOINC.
You had to open the discussion before starting to implement it.

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+    const int16_t port = 31416;
+    const int buffer_size = 4096;
+    const int wake_minutes = 30;
+    const int exec_hourly = 5;
+    int suffix_file_idx = 0;
+    const std::string android_address = "edu_berkeley_boinc_client_socket";
+    const std::string unix_address = "boinc_socket";
+    const std::string tcp_address = "localhost";
+    const std::string prefix_file_name = "boinc";
+    const std::chrono::minutes wake_every_minutes = std::chrono::minutes(wake_minutes);
```
AenBleidd, line 85: This is available starting from c++11 only
https://www.cplusplus.com/reference/chrono/

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+#include <sstream>
+#include <fstream>
+#include <iostream>
+#include <string>
+#include <chrono>
+#include <thread>
+#include <sockpp/socket.h>
+#include <sockpp/tcp_connector.h>
+#if !_WIN32
```
AenBleidd, line 29: `#ifndef WIN32`

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+    #include <sockpp/unix_connector.h>
+#endif
+
+void my_sleep(std::chrono::nanoseconds nanoseconds) {
```
AenBleidd, line 33: There is no need to declare one-line function that just makes a call to the library function with the same data.

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    sockpp::socket_initializer sockInit;
+    sockpp::tcp_connector tcp_conn;
+    sockpp::connector* array_connection[] = { &tcp_conn, NULL };
+#if !_WIN32
+    sockpp::unix_connector unix_conn;
+    array_connection[1] = &unix_conn;
+#endif
+    sockpp::connector* conn = NULL;
+    const int arrSize = sizeof(array_connection) / sizeof(array_connection[0]);
```
AenBleidd, line 74: should be made `unsigned int` or `size_t` since size must never be negative

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include <sockpp/tcp_connector.h>
+#if !_WIN32
+    #include <sockpp/unix_connector.h>
+#endif
+
+void my_sleep(std::chrono::nanoseconds nanoseconds) {
+    std::this_thread::sleep_for(nanoseconds);
+}
+
+sockpp::connector* get_connected(sockpp::connector* array_connection[], int arrSize) {
```
AenBleidd, line 37: `arrSize` should be `unsigned int` or `size_t`

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+    if (NULL == array_connection || 0 == arrSize) {
```
AenBleidd, line 38: If you change `arrSize` to `unsigned int` or `size_t`, this check will be valid. Now it will lead to an exception if negative value is passed, and it will pass this check since negative value != 0

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+        return NULL;
+    }
+    for (int i = 0; i < arrSize; ++i) {
```
AenBleidd, line 41: Same here

samples/collect_data/collect_data.cpp
AenBleidd, line 33: `std::chrono` is not available in pre-C++11 standard

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+        sockpp::connector* conn = array_connection[i];
+        if (conn != NULL && conn->is_connected()) {
+            return conn;
+        }
+    }
+    return NULL;
+}
+
+bool is_connected(sockpp::connector* array_connection[], int arrSize) {
```
AenBleidd, line 50: Same issue with `arrSIze` here

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+    if (NULL == array_connection || 0 == arrSize) {
+        return NULL;
+    }
+    for (int i = 0; i < arrSize; ++i) {
```
AenBleidd, line 54: `i` should be changed to `unsigned int` or `size_t` for the reasons mentioned above

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+    return false;
+}
+
+int main(int, char**) {
+    //boinc_init();
+
+    sockpp::socket_initializer sockInit;
+    sockpp::tcp_connector tcp_conn;
+    sockpp::connector* array_connection[] = { &tcp_conn, NULL };
+#if !_WIN32
```
AenBleidd, line 69: `#ifndef WIN32`

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+    sockpp::unix_connector unix_conn;
```
AenBleidd, line 70: Why non-Windows OSs required additional connection (as far as I see, Windows will have connection, other OSs will have two parallel connections). Is this intended?

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+    array_connection[1] = &unix_conn;
+#endif
+    sockpp::connector* conn = NULL;
+    const int arrSize = sizeof(array_connection) / sizeof(array_connection[0]);
+    const bool isDebug = false;
+    const int16_t port = 31416;
```
AenBleidd, line 76: could be just `int`

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+    const std::string android_address = "edu_berkeley_boinc_client_socket";
+    const std::string unix_address = "boinc_socket";
+    const std::string tcp_address = "localhost";
+    const std::string prefix_file_name = "boinc";
+    const std::chrono::minutes wake_every_minutes = std::chrono::minutes(wake_minutes);
+    const std::chrono::minutes exec_time_hours = std::chrono::minutes(exec_hourly);
+    
+    std::stringstream stream;
+    char buffer[buffer_size];
+    const std::chrono::time_point<std::chrono::system_clock> starttime = std::chrono::system_clock::now();
```
AenBleidd, line 90: Same issue with `std::chrono`

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+    while (true) {
+        int attemped_connections = 0;
+        while (!is_connected(array_connection, arrSize)) {
+            attemped_connections++;
```
AenBleidd, line 94: Postfix operator could be changed to prefix operator

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+            std::cout << "attemped connections: " << attemped_connections << "\n";
+            if (!tcp_conn.connect(sockpp::inet_address(tcp_address, port))) {
+                std::cout << "tcp connection fail:\n";
+                std::cout << tcp_conn.last_error_str() << "\n";
+                tcp_conn.clear();
+            }
+#if! _WIN32
```
AenBleidd, line 101: `#ifnded WIN32`

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+                std::cout << "android unix connection fail:\n";
+                std::cout << unix_conn.last_error_str() << "\n";
+                unix_conn.clear();                
+            }
+#endif
+            if (is_connected(array_connection, arrSize)) {
+                conn = get_connected(array_connection, arrSize);
+            }
+            std::cout << "\n";
+            my_sleep(std::chrono::seconds(1));
```
AenBleidd, line 118: Don't you think that doing this every second is necessary?

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        std::string request = "<boinc_gui_rpc_request><get_state/></boinc_gui_rpc_request>\3";
```
AenBleidd, line 121: Could be made `const`

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+        if (isDebug) {
+            std::cout << "before NULL == conn: " << conn << "\n";
+        }
+        if(NULL == conn) {
+#if !_WIN32
```
AenBleidd, line 126: `#ifndef WIN32`

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+            std::cout << "before write: \n";
+        }
+        const ssize_t res = conn->write(request);
+
+        if (isDebug) {
+            std::cout << "after write: " << res << "\n";            
+        }
+        if (res == -1) {
+            conn->close();
+#if !_WIN32
```
AenBleidd, line 142: `#ifndef WIN32`

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+            unix_conn.close();
+#endif
+            tcp_conn.close();
+            continue;
+        }
+        const std::chrono::time_point<std::chrono::system_clock> before_sleep = std::chrono::system_clock::now();
```
AenBleidd, line 148: The same issue with `std::chrono`

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+        std::strftime(&time_str[0], time_str.size(), "%Y-%m-%d %H:%M:%S", std::localtime(&in_time_t));
+        std::cout << "now: " << time_str << "\n";
+        std::cout << "request bytes: " << request.size() << "\n";
+        std::cout << "write bytes: " << res << "\n";
+
+        if (isDebug) {
+            std::cout << "after write, before read: \n";
+        }
+
+        ssize_t n = buffer_size;
```
AenBleidd, line 160: Could be made `const`
Also, better use `size_t`

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+        str = str.substr(0, str.size() - 1);
+        std::cout << "read bytes: " << str.size() << "\n";
+        if(str.empty()) {
+            continue;
+        }
+        std::ofstream outFile;
+        std::stringstream file_name;
+        file_name.str("");
+        file_name << prefix_file_name << suffix_file_idx << ".txt";
+        outFile.open(file_name.str());
```
AenBleidd, line 183: File will be overwritten every time. Is this by intention?

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+        outFile << str;
+        std::cout << "write to file: " << prefix_file_name << suffix_file_idx << ".txt\n\n";
+        suffix_file_idx++;
```
AenBleidd, line 186: Postfix operator could be changed to prefix operator

samples/collect_data/collect_data.cpp
```diff
@@ -0,0 +1,196 @@
...
+        outFile.flush();
+        outFile.close();
+        const std::chrono::time_point<std::chrono::system_clock> now = std::chrono::system_clock::now();
```
AenBleidd, line 189: The same issue with `std::chrono` here and one line below

AenBleidd (changes requested):
Current implementation is not finished and can't be used as a sample.
Please add interaction with boinc client (you can see an example here: https://github.com/BOINC/boinc/blob/master/samples/example_app/uc2.cpp)

samples/collect_data/collect_data.cpp
talregev: Windows have only 1 option connection tcp mode.
The linux/android can have either tcp connection or linux connection (file connection). 
The program try each connection. if it succeeded, it stay on the connection, if not, it switch to the next connection.

samples/collect_data/collect_data.cpp
talregev: Yes.

samples/collect_data/collect_data.cpp
talregev: It cannot made const. I change the value few lines later.

samples/collect_data/collect_data.cpp
talregev: Files will overwritten every run. It intention. the idea, that the app will send it over the server then delete the files. 
This is an example, not a real app.

samples/collect_data/collect_data.cpp
talregev: done.

samples/collect_data/collect_data.cpp
talregev: change it to size_t

samples/collect_data/collect_data.cpp
talregev: done.

samples/collect_data/collect_data.cpp
talregev: done.

samples/collect_data/collect_data.cpp
talregev: done.

samples/collect_data/collect_data.cpp
talregev: done.

samples/collect_data/collect_data.cpp
AenBleidd: Usually files/results are sent to server when task is done. So it's better to append to the file, otherwise server will just receive the last result that is probably not what you're expecting.

win_build/collect_data_vs2019.vcxproj
```diff
@@ -0,0 +1,164 @@
...
+    <IntDir Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">$(SolutionDir)Build\$(Platform)\$(Configuration)\$(ProjectName)\obj\</IntDir>
+    <LinkIncremental Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" />
+    <CodeAnalysisRuleSet Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">AllRules.ruleset</CodeAnalysisRuleSet>
+    <CodeAnalysisRules Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" />
+    <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" />
+    <CodeAnalysisRuleSet Condition="'$(Configuration)|$(Platform)'=='Release|x64'">AllRules.ruleset</CodeAnalysisRuleSet>
+    <CodeAnalysisRules Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
+    <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
```
AenBleidd, line 56: You can remove this PropertyGroup

win_build/collect_data_vs2019.vcxproj
```diff
@@ -0,0 +1,164 @@
...
+    <TargetName>$(ProjectName)</TargetName>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
```
AenBleidd, line 59: You can remove this property group

deploy/prepare_deployment.py
```diff
@@ -151,6 +158,7 @@
     './win_build/Build/x64/Release/slide_show.exe',
     './win_build/Build/x64/Release/test*.exe',
     './win_build/Build/x64/Release/wrappture*.exe',
+    './win_build/Build/x64/Release/collect_data*.exe'
```
AenBleidd, line 161: you can change it to `'./win_build/Build/x64/Release/collect_data.exe'`

AenBleidd: Also you missed `,` at the end of the line

davidpanderson:
What does this app do?  What's the goal?

samples/collect_data/collect_data.cpp
talregev: ok. I understand it now. I will do it.

talregev:
@davidpanderson 
I want to be able get answers for this questions to get decisions in development:

Old cpu:
How many devices are running armv6?
What projects are running this cpu devices?
How long this projects are running per day/week/month?
How many crush / error get this devices?

GPU:
How many devices have gpus and not use it?
What type of the gpu devices for non using gpu?
What projects they are running?
How many devices boinc fail to recognize the gpu? - assuming we detect some brand of know laptop or phone that we know come with gpu and we fail to detect.
How many android phone have gpu?
What the type?
How much percentage from that phone we fail to detect gpu?

Https
How many devices are using non https projects?
What other projects they are running?
What version of boinc they are running?


I am sure you can ask many other interesting question that can lead to better boinc development.

deploy/prepare_deployment.py
talregev: I missed the `,`. Thank you! It solve the issue.

win_build/collect_data_vs2019.vcxproj
talregev: done.

win_build/collect_data_vs2019.vcxproj
talregev: done.

talregev:
@davidpanderson 
More interesting reading about Data driven decision making.
This is the direction I want achieve in boinc throw data collect app/project. 

https://en.wikipedia.org/wiki/Data_based_decision_making

https://www.tableau.com/learn/articles/data-driven-decision-making#:~:text=What%20is%20data%2Ddriven%20decision,goals%2C%20objectives%2C%20and%20initiatives

PR #4616, opened 2022-01-28
Disable new version check

LocutusOfBorg:
followup PR after https://github.com/BOINC/boinc/pull/4272

### DIFF
--- a/lib/cc_config.cpp
+++ b/lib/cc_config.cpp
@@ -229,6 +229,7 @@ void CC_CONFIG::defaults() {
         ignore_gpu_instance[i].clear();
     }
     max_event_log_lines = DEFAULT_MAX_EVENT_LOG_LINES;
+    disable_version_check = false;
     max_file_xfers = 8;
     max_file_xfers_per_project = 2;
     max_stderr_file_size = 0;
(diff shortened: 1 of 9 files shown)


LocutusOfBorg:
note: I don't know how to test because I didn't yet figure out how the parsing works (and how to change the xml file)

talregev:
copy this xml:
https://github.com/BOINC/boinc/blob/master/android/BOINC/app/src/main/assets/cc_config.xml

add these lines:

```xml
 <disable_version_check>1</disable_version_check>
```

LocutusOfBorg:
well so my conf file was correct...
cat /etc/boinc-client/cc_config.xml
```
<!--
This is a minimal configuration file cc_config.xml of the BOINC core client.
For a complete list of all available options and logging flags and their
meaning see: https://boinc.berkeley.edu/wiki/client_configuration
-->
<cc_config>
  <log_flags>
    <task>1</task>
    <file_xfer>1</file_xfer>
    <sched_ops>1</sched_ops>
  </log_flags>
  <options>
    <disable_version_check>1</disable_version_check>
  </options>
</cc_config>
```

but for some reasons I still see the "update" mechanism shown...

LocutusOfBorg:
I'm debugging the parser code

talregev:
I think you need to delete the  xml you copy.
 let boinc create one by default, the search for it, and edit with correct option.
Then you understand what to do.

lib/cc_config.cpp
```diff
@@ -389,6 +391,7 @@ int CC_CONFIG::parse_options(XML_PARSER& xp) {
             continue;
         }
         if (xp.parse_int("max_event_log_lines", max_event_log_lines)) continue;
+        if (xp.parse_bool("disable_version_check", disable_version_check)) continue;
```
AenBleidd, line 394: You already parsed this flag on the line 322

AenBleidd (changes requested):

talregev:
@LocutusOfBorg Do you need some help?

AenBleidd:
Since we have now a CI flow that builds the linux packages for 4 different Linux distros (for more details see https://github.com/BOINC/boinc/wiki/Linux-DEB-and-RPM-support), and we also want to encourage our users to use the latest version of our software, there is no reason to implement this functionality.
For the distros we're currently missing, it's quite easy to add them and publish the same way from the BOINC official web-site (and this is something we are gonna do).
Thus, I'm closing this PR without merge.

PR #4619, opened 2022-02-02
[WIP] Test reasons for build failure under Github Workflow

CharlieFenton:
DO NOT MERGE - TESTING ONLY

Issue #4612, opened 2022-01-23
OSX CI build failed but shown as 'passed'


AenBleidd:
CI build Is failed with the next error:

```
Linking boinc
ld: symbol(s) not found for architecture x86_64

clang: error: linker command failed with exit code 1 (use -v to see invocation)

** BUILD FAILED **

The following build commands failed:

	Ld /Users/runner/work/boinc/boinc/mac_build/build/boinc.build/Deployment/BOINC_Client.build/Objects-normal/x86_64/Binary/boinc normal x86_64

(1 failure)
```

https://github.com/BOINC/boinc/runs/4912711265?check_suite_focus=true

But build script ignores this.

@CharlieFenton, could you please take a look and fix both script and build?

Thanks in advance.

CharlieFenton:
This issue mentions two separate but related problems. Both of these appear to be problems in the implementation of Github Workflow CI, not anything in my code or that I can fix. I recommend asking @ChristianBeer for assistance, since he created the CI workflows and has been their primary maintainer.

[1] The Mac build errors were fixed each time I forced a rebuild of the dependency cache. But the link errors returned the next time I caused the Github Actions CI to run without rebuilding the cache again (so it used the cached dependent libraries.) This leads me to believe that there is a problem with the dependency cache logic, which was created and has been maintained primarily by @ChristianBeer.

I thought that perhaps another Github Action might be corrupting the macOS dependency cache, but the same problem occurred when I removed all *.yml files from _.github/workflows/_ other than _osx.yml_.

One way to force a rebuild of the dependency cache is to add the following before line 148 in _3rdParty/buildMacDependencies.sh_:
`doclean="yes"`
as I did in my commit 298d5e9e2d. I don't know whether the rebuild affects only the one branch. Note that it is important to then remove that line after the cache has been rebuilt.

[2] Regarding why CI showed the failed MacOS build as 'passed':

If you look at lines 113-115 of the shell script _mac_build/buildMacBOINC-CI.sh_, you can see that it should return 1 if the build of the Manager failed:
`    if [ $retval -ne 0 ]; then cd ..; exit 1; fi
`
and similarly if the build fails for any of the other components. This script is used only for the CI builds, and was created and maintained primarily by @ChristianBeer. I have not been directly involved with work on the CI builds since coordinating with Christian when he first developed the Travis CI support in 2017, and am not familiar with the syntax or other details.

[...]

PR #4622, opened 2022-02-03
[CI] Add *.yml files to the hash while resolving cache, remove restore-key

AenBleidd:

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -34,7 +34,7 @@ jobs:
           path: |
             3rdParty/buildCache
             !3rdParty/buildCache/android/vcpkgcache/
-          key: android-${{ matrix.type }}-${{ hashFiles('android/*.sh') }}
+          key: android-${{ matrix.type }}-${{ hashFiles('android/*.sh', '.github/workflows/android.yml') }}
 
       - name: Configure Python
         if: ${{ success() && contains(matrix.type, 'vcpkg') }}
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -25,7 +25,7 @@ jobs:
           path: |
             3rdParty/buildCache
             !3rdParty/buildCache/linux/vcpkgcache/
-          key: linux-coverity
+          key: linux-coverity-${{ hashFiles('3rdParty/*Linux*.sh', 'linux/*.sh', '.github/workflows/coverity.yml') }}
 
       - name: Automake
         if: success()
--- a/.github/workflows/flutter-macos.yml
+++ b/.github/workflows/flutter-macos.yml
@@ -18,17 +18,17 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          fetch-depth: 2          
-          
+          fetch-depth: 2
+
       - name: Setup xcode
         uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: '12.5.1'         
+          xcode-version: '12.5.1'
 
       - name: Setup flutter
         uses: subosito/flutter-action@v1
         with:
-          channel: 'beta'      
+          channel: 'beta'
 
       - name: Setup macos desktop
         if: matrix.type == 'macos'
@@ -39,8 +39,7 @@ jobs:
         if: matrix.type == 'macos'
         with:
           path: 3rdParty/buildCache
-          key: osx-flutter-${{ hashFiles('3rdParty/*Mac*.sh', 'mac_build/dependencyNames.sh', 'mac_build/[bB]uild*.sh') }}
-          restore-keys: osx-flutter-
+          key: osx-flutter-${{ hashFiles('3rdParty/*Mac*.sh', 'mac_build/setupForBOINC.sh', 'mac_build/dependencyNames.sh', 'mac_build/[bB]uild*.sh', '.github/workflows/flutter-macos.yml') }}
 
       - name: Build macos client
         if: matrix.type == 'macos'
@@ -54,7 +53,7 @@ jobs:
         run: |
           cd samples/flutter/
           flutter create boinc
-          cd boinc/ 
+          cd boinc/
           flutter build macos
 
       - name: Build ios
@@ -62,5 +61,5 @@ jobs:
         run:  |
           cd samples/flutter/
           flutter create boinc
-          cd boinc/ 
+          cd boinc/
           flutter build ios --release --no-codesign
--- a/.github/workflows/flutter-ubuntu.yml
+++ b/.github/workflows/flutter-ubuntu.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
-          
+
       - name: Setup Java
         if: matrix.type == 'android'
         uses: actions/setup-java@v1
@@ -37,7 +37,7 @@ jobs:
 
       - name: Setup linux desktop
         if: matrix.type == 'linux'
-        run: flutter config --enable-linux-desktop  
+        run: flutter config --enable-linux-desktop
 
       - name: Install linux dependencies
         if: matrix.type == 'linux'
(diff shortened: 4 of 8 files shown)


CharlieFenton (commented):
Sorry, I know absolutely nothing about YML syntax or how Github Actions work, so I can't offer any help.

AenBleidd:
@CharlieFenton, at least you can check that this fixes the initial issue with cache

CharlieFenton:
I think you would be better asking @ChristianBeer. I have added him as a reviewer.
@ChristianBeer please see Issue #4612.

> @CharlieFenton, at least you can check that this fixes the initial issue with cache

OK, I will do that, but not tonight.

CharlieFenton:
@AenBleidd But how do I clone your changes into my feature branch from yours, or make commits to this branch? I have already deleted my branch when I closed #4619, so I will have to create a new brand and a new PR.

AenBleidd:
@CharlieFenton, you can do this next way:

```
git checkout -b AenBleidd-vko_fix_cache_issue master
git pull https://github.com/AenBleidd/boinc.git vko_fix_cache_issue
```

CharlieFenton:
@AenBleidd In my test branch [here](https://github.com/BOINC/boinc/pull/4625/commits/8ee751b67aa3808c8efd84a2bacc4458a9de76f2) It is rebuilding the cache even though I removed the patch to force a rebuild. So this appears to not work correctly. But I will try again after this I run is finished, changing a different file to trigger another CI run.

CharlieFenton:
[Here ](https://github.com/BOINC/boinc/runs/5060395122?check_suite_focus=true) is the CI output from my first run, with the patch to force a cache rebuild in _3rdParty/buildMacDependencies.sh_.

CharlieFenton:
My test PR is #4625 where the second run is not finding the cache created in the first (original) run.

CharlieFenton:
I just ran another test PR #4626 and found that a newly created branch off master and PR normally do not build a new dependency cache for MacOS, but this PR #4622, and my PR #4625 which branched  of this one both did build a new cache instead of using the existing one. I believe my PR #4625 should have used the same cache your #4622 had previously created. But i am still investigating.

AenBleidd:
@CharlieFenton, this is basically the problem of the GitHub Actions cache mechanism: it saves multiple caches (based on hash) so when you force a rebuild - it will definitely do a rebuild even if the cache will be found, but when you change it back - old cache will be used because it will be found by the cache, so the only way to not use the old cache is to add smth to the files that affect the cache to make it unique again.
I can basically change it to self-implemented cache mechanism (which will take time, of course, but still possible to do) to work the way similar to TravisCI way when only last cache for every CI flow was saved.
Probably I'll start working on its implementation this weekend.

CharlieFenton:
After changing a different file (one that is not used by the build) the next CI run of PR #4625 did find and re-use the old cache and built without errors. [Here](https://github.com/BOINC/boinc/runs/5061646220?check_suite_focus=true) is the build log from that one. So it does appear that your changes fixed the cache problem. I will now do another small change to trigger another CI run to confirm it is working correctly and will report back. If i t works correctly, shall I merge it into master?

I will also run some tests on my PR #4626 to see if I can get more information on why the build was failing before your fix.

AenBleidd:
@CharlieFenton,

> If i t works correctly, shall I merge it into master?

Yes, go ahead.

Thanks for the testing.

CharlieFenton:
My [last test](https://github.com/BOINC/boinc/runs/5063163956?check_suite_focus=true) worked again, so I think you have fixed the cache issue. But I will wait to merge this so that you can continue to reproduce the MacOS build errors in branches off master. This will be needed to debug why CI thinks the build succeeded when it failed. Please let me know when I should merge.

CharlieFenton:
> when you force a rebuild - it will definitely do a rebuild even if the cache will be found, but when you change it back - old cache will be used

I finally was able to determine what missing symbols caused the linker errors for MacOS, and I think we may be misunderstanding the problem. I suspect that maybe the dependency cache for GIT Master may out of date, and that is the cache being used for all branches created from master. How we can we remove the old cache from master and force a new cache build there? 

To summarize what I found: it appears that libcurl is looking for the libnghttp2 and libidn2 libraries, which are not present on the Mac by default. I updated the libcurl build script on 16 November 2021 to not require those libraries in commit 4010e8d2db which you merged into master with commit 7d01ee565e. The change added the following to the configure command:
`--without-libidn --without-libidn2 --without-nghttp2 --without-ngtcp2 --without-nghttp3 --without-quiche
`
I rebuilt the cache on my feature branch for testing, but this did not trigger a cache rebuild in master, and I suspect the master cache has not been rebuilt since before then. We need to find a way to ensure that the master cache is properly rebuilt when such changes are made. I also wonder whether the cache will automatically be rebuilt when the Xcode project and dependent library build scripts are updated to use a new version of a dependent library. This may also be an issue for other platforms besides MacOS.

I also wonder if your changes in this PR #4622 gave the appearance of solving the problem only because by adding a new items to the cache in each *.yml file you forced a one-time cache rebuild, but that this one-time change might not solve this problem for future changes of the sort I made on 16 November.

Should I do a commit directly to master of _3rdParty/buildMacDependencies.sh_, adding `doclean="yes"` to force a cache rebuild, and then doing another commit with that removed? Will that work? If I do this, I can create a new test feature branch from master and a new test PR to see i f that solves the problem without your changes in this PR #4622. What do you think?

CharlieFenton (commented):
Please see my [previous comment](https://github.com/BOINC/boinc/pull/4622#issuecomment-1029854463)

AenBleidd:
> Should I do a commit directly to master of 3rdParty/buildMacDependencies.sh, adding doclean="yes" to force a cache rebuild, and then doing another commit with that removed? Will that work? If I do this, I can create a new test feature branch from master and a new test PR to see i f that solves the problem without your changes in this PR https://github.com/BOINC/boinc/pull/4622. What do you think?

Unfortunately, no. GitHub Acvtions saves several caches.
The situation is next:
- Let's say, current hash on master is XXXX
- You do a commit, hash changes to YYY, cache rebuilds
- You do another commit to revert your change, hash becomes XXX again, and older cache will be used.

I can include XCode project files to hash, so when you change XCode project files without changing build scripts, hash will change and cache will be invalidated.

If you want this - just tell me what files to include in hash and I'll do this (or you can do this by yourself, I think it shouldn't be too complicated for you since you see how I added YML files to the hash).

Please tell me what you think.

CharlieFenton:
In that case, we should probably add the following to the hash:
3rdParty/buildMacDependencies.sh
mac_build/setupForBOINC.sh
mac_build/buildopenssl.sh
mac_build/buildc-ares.sh
mac_build/buildcurl.sh
mac_build/buildWxMac.sh
mac_build/buildfreetype.sh
mac_build/buildFTGL.sh

The above should take care of all changes to the dependent libraries for MacOS, and also if we add or remove needed libraries. Please do **not** add the Xcode project to the hash. That is not needed because changes to the Xcode project should not affect the dependent libraries and it would trigger unnecessary cache rebuilds for every change to the project.

You may also want to add  the other _3rdParty/build*.sh_ files to the appropriate hashes. I don't know whether the .yml files which you added in this PR #4622 still need to be part of the hash; I will let you decide that.

Let me know when you have finished this and I will merge it into master. That should of course immediately cause a cache rebuild so I expect that will fix the build errors for now. But I don't know how we will be able to determine whether it fixes it for future changes of the type I made on 16 November. Do you have any thoughts about that?

AenBleidd:
@CharlieFenton, all these files should be already in hash for OSX: https://github.com/BOINC/boinc/blob/35e3d70704b8431fa91b7fe943f87ec1d4882048/.github/workflows/osx.yml#L29

So my PR should fix the issue that we have now.
For the possible future cases I can switch to our own cache implementation (that I can implement) and store only 1 cache for every flow (that should make it the same way TravisCI did and that is more predictable in my opinion).

I think we're good to go with current solution, and when I will be ready with our cache implementation, we can test this properly together and decided whether it's better or not.

AenBleidd:
Ah, no, we are missing one file, `mac_build/setupForBOINC.sh`
I'll add it to my PR now

CharlieFenton:
I will merge this after all checks have run, but probably tomorrow since I will be going to bed soon.

AenBleidd:
@CharlieFenton, thank you for your help!

CharlieFenton (approved):

CharlieFenton:
@AenBleidd Thank you for explaining the use of a hash for the dependency cache under Github Actions. Now that I understand how tis works, I am confident that these changes are the correct solution for MacOS, so I don't feel that you need to spend time on creating a custom cache implementation. But I strongly recommend doing the same thing for the other platforms as you did for MacOS. 

But I don't know how you would do this for the vcpkg runs to assure that the CI runs will use a cache with the same library versions as are being used for release builds.

AenBleidd:
@CharlieFenton, other platforms should be already ok with this. I checked build scripts and looks like all of them already affect the  hash.
Regarding vcpkg, we use completely different cache mechanism because Github Actions cache is not suitable for it, and it works fine also.
Thank you for merging this.

PR #4630, opened 2022-02-07
Configure more vram to avoid VMs get stuck during boot

computezrmle:
**Description of the Change**
Roughly 10% of Rosetta's vbox tasks get stuck when the VM starts.
This is most likely caused by not enough vram given to the virtual graphics adapter.

For years vboxwrapper uses a hard-wired value of 8 MB.
Recent VirtualBox versions (6.1.30) configure 16 MB when a VM is configured manually and a Linux 64-bit template is used.
A site moderator from the VirtualBox forum even requests 128 MB vram:
https://forums.virtualbox.org/viewtopic.php?f=3&t=105195#p513161

This seems to be OK for Windows guests but far too much for Linux guests.
A test with nearly 100 Rosetta vbox tasks over the last 24 h didn't result in a single stuck task when 16 MB vram are configured.

This PR makes the vram size configurable via vbox_job.xml in case future VirtualBox releases require higher settings.
In addition the virtual graphics adapter type is also configurable.
Valid adapter types depend on the VirtualBox Extensions installed in the VM.
See:
https://www.virtualbox.org/manual/ch03.html#settings-display

**Alternate Designs**
N/A

**Release Notes**
Makes VRAM size and virtual graphics adapter type for VirtualBox VMs configurable via vbox_job.xml

### DIFF
--- a/samples/vboxwrapper/vbox_vboxmanage.cpp
+++ b/samples/vboxwrapper/vbox_vboxmanage.cpp
@@ -250,6 +250,18 @@ namespace vboxmanage {
         retval = vbm_popen(command, output, "modifychipset");
         if (retval) return retval;
 
+        // Tweak the VM's Graphics Controller Options
+        //
+        vboxlog_msg("Setting Graphics Controller Options for VM.");
+        sprintf(buf, "%d", (int)vram_size_mb);
+
+        command  = "modifyvm \"" + vm_name + "\" ";
+        command += "--vram " + string(buf) + " ";
+        command += "--graphicscontroller " + vm_graphics_controller_type + " ";
+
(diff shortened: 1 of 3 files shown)


davidpanderson:
We'd need to update the vboxwrapper doc:
https://boinc.berkeley.edu/trac/wiki/VboxApps
with instructions for when and how to set the new options.
I can give you write access, or just send me the text I'll update the page.
-- David

computezrmle:
@davidpanderson 

> I can give you write access

Let's try it this way first.
If you need my email address - you should already have it since I recently sent you a mail regarding a technical issue with the main BOINC forum that couldn't be solved differently.

AenBleidd (approved):

PR #4646, opened 2022-02-21
Change error message

computezrmle:
**Description of the Change**
A volunteer recently complained it's bad English because it's past tense instead of present tense.

**Alternate Designs**
N/A

**Release Notes**
N/A

### DIFF
--- a/samples/vboxwrapper/vbox_common.cpp
+++ b/samples/vboxwrapper/vbox_common.cpp
@@ -398,7 +398,7 @@ string VBOX_BASE::get_error(int num){
           "   NOTE: VM failed to enter an online state within the timeout period.\n \
               This might be a temporary problem and so this job will be rescheduled for another time.\n",
 
-          "VM environment needed to be cleaned up.",
+          "VM environment needs to be cleaned up.",
 
           "Foreign VM Hypervisor locked hardware acceleration features.",
 


CharlieFenton:
Please remember that when you change even one character of a user-facing message, it invalidates the translation of that string for all languages. The translations of the old version of this message will be ignored by the software. This change means that this string will now appear only in English worldwide (in any new releases of BOINC) until the translation volunteers for each language update it. That won't happen until BOINC (@AenBleidd these days) creates a new set of files for the translators to work from and notifies them that there are changes needing updates.

This needs to be done well before the next release, since the translators for many languages often take a long time to respond. Many, perhaps most, languages have not been updated in years.

In general, I feel that user-facing strings should be changed only when absolutely necessary because of the problems and delays getting translations. In my opinion, this change does not meet that standard.

computezrmle:
Independend of the extent of any language related change this

> ... it invalidates the translation of that string for all languages.

should not happen, especially in connection with this

> Many ... languages have not been updated in years.

The usual way to handle this would be to define one single source language and a set of target languages.
In case of a change in the source language the corresponding objects (sentences, paras, pages, files, ...) in the target languages should stay valid but get flagged "review/approval required".

Otherwise waiting for any text to be translated into uncommon languages like Vogon or Penguin could lead to an infinite delay of any new release.

AenBleidd:
> Otherwise waiting for any text to be translated into uncommon languages like Vogon or Penguin could lead to an infinite delay of any new release.

I'll handle this when times comes.
And of course we're not waiting for all translations

PR #4652, opened 2022-02-27
Fix rendering/parsing of some cc_config attributes

vmc-coding:
These attributes are integers, let's render and parse them accordingly.


davidpanderson:
?? They're doubles, not ints.
We use doubles to store files sizes, which can be >32 bits

vmc-coding:
Oh, I've looked at an old branch (client_release/7/7.14) on which these are integers. This had been changed with a79d60b8891237e011bdaad3bd1a7c1572e2f3b7.

Sorry for the waste of time.

PR #4655, opened 2022-03-01
[Windows][vboxwrapper] Add support of VBox 6.1.* to vboxwrapper COM implementation

### LINKED ISSUE #4638: [vboxwrapper] COM implementation crashes with VBox 6.1.(32)

AenBleidd:
This fixes #4638.


samples/vboxwrapper/vbox_mscom61.cpp
```diff
@@ -0,0 +1,64 @@
...
+//
+// BOINC is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
+
+#define _VIRTUALBOX60_
```
computezrmle, line 18: Shouldn't this be
`#define _VIRTUALBOX61_`

AenBleidd: No. Maybe it's a good idea to rename `_VIRTUALBOX60_` to `_VIRTUALBOX6_` and change its usage

computezrmle:
@AenBleidd 
Since support for VirtualBox 4.2 the filename pattern has always been:
vbox**ab**.tlb, vbox_mscom**ab**.cpp, vbox_mscom**ab**.h

In vbox_mscom**ab**.cpp there has always been a corresponding line:
`#define _VIRTUALBOXab_`


Would like to understand why you suggest to change this for 61, especially since the filenames follow the old pattern.

AenBleidd:
> @AenBleidd 
> 
> Since support for VirtualBox 4.2 the filename pattern has always been:
> 
> vbox**ab**.tlb, vbox_mscom**ab**.cpp, vbox_mscom**ab**.h
> 
> 
> 
> In vbox_mscom**ab**.cpp there has always been a corresponding line:
> 
> `#define _VIRTUALBOXab_`
> 
> 
> 
> 
> 
> Would like to understand why you suggest to change this for 61, especially since the filenames follow the old pattern.
> 
> 

Because the changes in API that were introduced in 6.0 are still the same in 6.1, and I don't want to write now smth like this:
#id defined (_VIRTUALBOX60_) or defined (_VIRTYALBOX61_)

And then every new version add this new macro. These macro are used not to represent particular version but to apply necessary differences while building the code. Please take a look at the code and the macro usages.

davidpanderson:
This should (but doesn't) have instructions for how to update vboxwrapper for a new vbox version:
https://boinc.berkeley.edu/trac/wiki/AdminWrappers

Also: should we even be supporting the COM interface?  The Vbox Manage approach seems to be more stable.

computezrmle:
Related to the instructions here:
https://boinc.berkeley.edu/trac/wiki/AdminWrappers


At least for Linux builds this line
`./configure --disable-server --disable-client --disable-manager`
should be extended to
`./configure --disable-server --disable-client --disable-manager --enable-optimize`


As of my experience on heavily loaded systems vbox tasks regularly (daily) run into "Postponed ..." errors if "enable-optimized" is not set. The error disappears if "enable-optimized" is used.

AenBleidd:
> Also: should we even be supporting the COM interface? The Vbox Manage approach seems to be more stable.

I completely agree, VBoxManage approach is way more stable than COM and could work with newest versions of VBox without rebuild (at least in a majority of times).

I'm closing this PR then and open new one to remove COM support

PR #4659, opened 2022-03-05
Print correct logmessage

computezrmle:
**Description of the Change**
see title.

**Alternate Designs**
N/A
**Release Notes**
N/A

### DIFF
--- a/client/cs_proxy.cpp
+++ b/client/cs_proxy.cpp
@@ -76,7 +76,7 @@ void select_proxy_info() {
         working_proxy_info = config_proxy_info;
         msg_printf(0, MSG_INFO, "Using proxy info from cc_config.xml");
         if (env_var_proxy_info.present) {
-            msg_printf(0, MSG_INFO, "Proxy info env vars overridden by GUI");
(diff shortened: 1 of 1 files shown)


AenBleidd:
This is mostly the correct message because in most cases these settings were changed by GUI

client/cs_proxy.cpp
```diff
@@ -76,7 +76,7 @@ void select_proxy_info() {
         working_proxy_info = config_proxy_info;
         msg_printf(0, MSG_INFO, "Using proxy info from cc_config.xml");
         if (env_var_proxy_info.present) {
-            msg_printf(0, MSG_INFO, "Proxy info env vars overridden by GUI");
+            msg_printf(0, MSG_INFO, "Proxy info env vars overridden by cc_config.xml");
```
computezrmle, line 79: Line 79 is only executed if the proxy settings are made via cc_config.xml and a proxy is **not** set via GUI.
The GUI settings are reported by the code in lines 66-74.

AenBleidd (approved):

PR #4661, opened 2022-03-06
Computing prefs dialog: use %f instead of %g

### LINKED ISSUE #4650: Large number in "switch between applications" is misunderstood
If you enter 1000000 minutes into "switch between applications", it saves ok, but next time this is displayed as 1+e006, and saving it like that tells me it's an invalid number.

davidpanderson:
(which writes scientific notation for large numbers,
which then can't be read).

Fixes #4650

### DIFF
--- a/clientgui/DlgAdvPreferences.cpp
+++ b/clientgui/DlgAdvPreferences.cpp
@@ -252,7 +252,7 @@ void CDlgAdvPreferences::DisplayValue(double value, wxTextCtrl* textCtrl, wxChec
             return;
         }
     }
-    buffer.Printf(wxT("%g"), value);
+    buffer.Printf(wxT("%.2f"), value);
     textCtrl->ChangeValue(buffer);
     textCtrl->Enable();
 }


Vulpine05:
@davidpanderson , I think this could still cause a problem if a very large number is entered.  Could you review my comments in #4650 from the past few days, as well as #4660?

AenBleidd:
'%f' always displays number as float and never switch to use exponent, at least according to the documentation: https://en.wikipedia.org/wiki/Printf_format_string

RichardHaselgrove:
%f also unconditionally adds the full fixed-point allocation of decimal places, whereas %g suppresses them when inappropriate, as @vulpine05 demonstrated in https://github.com/BOINC/boinc/issues/4650#issuecomment-1051299345.

The config files are dual-use: sometimes they are used for automated message passing between the GUI and the client, and sometimes they are edited manually (by people who prefer the CLI, or read the user manual).

Two things are important:
1) the intended data must be human-readable, to minimise confusion and errors
2) the actual data must be machine-readable, so that it can be acted upon.

Vulpine05:
I think a minor detail of David’s PR is being missed.  The change is not just using %f, but %.2f.  This will print out the value to the hundredths, even if it is a whole number.  I think this is good because settings are only stored (and actually rounded to) the hundredths.  I didn’t include this change because I think it would require (or at least would look better with) some text window widening (something I couldn’t test on all platforms).  Also, and to Richard’s second point, this still leaves the program open to the possibility of storing an invalid number because it isn’t validated with an upper limit.

I think both PRs #4660 and #4661 are valid.  By using %.2f, I think my PR can be modified to a higher upper limit than 10,000, if we feel it won’t cause any problems.

AenBleidd (approved):

PR #4664, opened 2022-03-09
[Manager][Windows] Make Manager DPI unaware to let wxWidgets and Windows scale GUI elements properly

AenBleidd:

### DIFF
--- a/clientgui/sg_PanelBase.h
+++ b/clientgui/sg_PanelBase.h
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2008 University of California
+// Copyright (C) 2022 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -18,7 +18,7 @@
 #ifndef BOINC_SG_PANELBASE_H
 #define BOINC_SG_PANELBASE_H
 
-#include "sg_CustomControls.h" 
+#include "sg_CustomControls.h"
 #include "sg_BoincSimpleFrame.h"
 #include "MainDocument.h"
 
@@ -35,8 +35,7 @@
 #define LARGE_FONT 16
 #endif
 
-#define SIDEMARGINS ADJUSTFORXDPI(30)
-
+const int sideMargins = 30;
 
 ///////////////////////////////////////////////////////////////////////////////
 /// Class CSimplePanelBase
@@ -49,7 +48,7 @@
 #endif
 
 
-class CSimplePanelBase : public wxPanel 
+class CSimplePanelBase : public wxPanel
 {
     DECLARE_DYNAMIC_CLASS( CSimplePanelBase )
     DECLARE_EVENT_TABLE()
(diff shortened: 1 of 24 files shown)


AenBleidd:
@CharlieFenton, please test this PR. It might not affect OSX but just in case I touched smth that might break smth

AenBleidd:
linux-mingw failed build is not related to this PR

clientgui/sg_PanelBase.h
```diff
@@ -35,9 +35,6 @@
 #define LARGE_FONT 16
 #endif
 
-#define SIDEMARGINS ADJUSTFORXDPI(30)
```
CharlieFenton, line 38: I encourage you to restore this #define (eliminating the ADJUSTFORXDPI()) of course, thus:
`#define SIDEMARGINS 30
`
and please restore the use of `SIDEMARGINS ` in all the places you have changed  it to a literal `30`. It is important that this value be the same in all those places. It also makes it easier to change the margins if it is ever desired to do so in the future.
This value is uses in sg_BoincSimpleFrame.cpp, sg_PanelBase.cpp, sg_ProjectPanel.cpp and sg_TaskPanel.cpp, and must be the same in all those files.

clientgui/sg_TaskPanel.cpp
```diff
@@ -29,13 +29,8 @@
 
 
 #define SORTTASKLIST 1  /* TRUE to sort task selection control alphabetically */
-#define SLIDESHOWWIDTH ADJUSTFORXDPI(290)
```
CharlieFenton, line 32: I feel it is better to leave these #define statements, just eliminating ADJUSTFORXDPI() and ADJUSTFORYDPI(), and restore them where they were used before instead of using a number. This is a form of comment in the places they are used, to help others understand how the values were obtained. An example is the old line 514:
` m_SlideShowArea->SetMinSize(wxSize(SLIDESHOWWIDTH+(2*SLIDESHOWBORDER), SLIDESHOWHEIGHT+(2*SLIDESHOWBORDER)));`
The compiler will replace these with the computed values, so the run time efficiency is not reduced by doing this.

CharlieFenton:
My comment above refers to more than just the 2 lines shown there. It includes all these:
```
#define SLIDESHOWWIDTH ADJUSTFORXDPI(290)
#define SLIDESHOWHEIGHT ADJUSTFORYDPI(126)
#define SLIDESHOWBORDER 1
#define HIDEDEFAULTSLIDE 1
#define TESTALLDESCRIPTIONS 0
#define SCROLLBARSPACER 8

```

clientgui/sg_PanelBase.h
AenBleidd: @CharlieFenton, I'll restore this and all others but change them to `const` instead of `#define` if you don't mind and have no strong arguments to use exactly macros instead of constants.

CharlieFenton:
@AenBleidd I don't think it's necessary to change all constants back to symbols for constants if they are are used only once or twice. I just felt the ones I specifically mentioned were better as symbols. 
Using `const` is fine with me, assuming that the compiler will do any math at compile time in the same way as it would for `#define`.

AenBleidd:
> I just felt the ones I specifically mentioned were better as symbols.

@CharlieFenton, that was exactly what I meant. Sorry for telling that not accurate enough.

CharlieFenton:
I have built and tested on MacOS and I have found no problems.

CharlieFenton (approved):

PR #4666, opened 2022-03-13
Systemd service order

computezrmle:
**Description of the Change**
Linux/Systemd

If service dependencies are not explicitly configured systemd tries to start services in parallel.
This may occasionally lead to a situation where BOINC checks for VirtualBox although that service is not yet present.

A Rosetta user recently posted an issue that might be caused by that:
https://boinc.bakerlab.org/rosetta/forum_thread.php?id=14966&postid=105428

This PR sets a clear dependency and starting order between BOINC and VirtualBox but also ensures BOINC doesn't fail if VirtualBox is not used.

**Alternate Designs**
N/A

**Release Notes**
N/A

### DIFF
--- a/client/scripts/boinc-client.service.in
+++ b/client/scripts/boinc-client.service.in
@@ -1,7 +1,8 @@
 [Unit]
 Description=Berkeley Open Infrastructure Network Computing Client
 Documentation=man:boinc(1)
-After=network-online.target
+Wants=vboxdrv.service
+After=vboxdrv.service network-online.target
 
 [Service]
 Type=simple


RichardHaselgrove:
The same problem occurs if systemd launches the boinc client service before the graphics drivers have initialised - GPUs can't be detected either.

Could that be handled in the same PR, or should I dig out the previous references for your kind attention?

RichardHaselgrove:
Previous reference: #3150

computezrmle:
Regarding VirtualBox there is only 1 well known dependency and it can easily be controlled by systemd.
Regarding BOINC/GPU drivers there are lots of complex dependencies, e.g. headless vs. non-headless or NVIDIA/AMD/Intel drivers.

Hence, the starting order VirtualBox/BOINC can be made a standard while the GPU issues appear to be too complex for the moment.

AenBleidd:
I'm not an expert in linux, but this change looks legit.

AenBleidd (approved):

PR #4678, opened 2022-03-25
[Windows] Clean apps windows projects

talregev:
Clean apps windows projects.

### DIFF
--- a/win_build/htmlgfx_vs2019.vcxproj
+++ b/win_build/htmlgfx_vs2019.vcxproj
@@ -95,10 +95,8 @@
     <CodeAnalysisRules Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" />
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">htmlgfx_26203_windows_x86_64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">htmlgfx_26203_windows_arm64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Release|x64'">htmlgfx_26203_windows_x86_64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">htmlgfx_26203_windows_arm64</TargetName>
+    <TargetName>$(ProjectName)_26203_windows_$(Platform)</TargetName>
+
   </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <Midl>
--- a/win_build/vboxwrapper_vs2019.vcxproj
+++ b/win_build/vboxwrapper_vs2019.vcxproj
@@ -95,10 +95,7 @@
     <CodeAnalysisRules Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" />
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">vboxwrapper_26205_windows_x86_64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">vboxwrapper_26205_windows_arm64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Release|x64'">vboxwrapper_26205_windows_x86_64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">vboxwrapper_26205_windows_arm64</TargetName>
+    <TargetName >$(ProjectName)_26205_windows_$(Platform)</TargetName>
   </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <Midl>
--- a/win_build/worker_vs2019.vcxproj
+++ b/win_build/worker_vs2019.vcxproj
@@ -92,17 +92,8 @@
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" />
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
-    <TargetName>worker_windows_x86_64</TargetName>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">
-    <TargetName>worker_windows_x86_64</TargetName>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
-    <TargetName>worker_windows_x86_64</TargetName>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">
-    <TargetName>worker_windows_x86_64</TargetName>
+  <PropertyGroup >
+    <TargetName>$(ProjectName)_windows_$(Platform)</TargetName>
   </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <Midl>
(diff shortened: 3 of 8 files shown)


win_build/htmlgfx_vs2019.vcxproj
```diff
@@ -95,10 +95,8 @@
     <CodeAnalysisRules Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" />
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">htmlgfx_26203_windows_x86_64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">htmlgfx_26203_windows_arm64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Release|x64'">htmlgfx_26203_windows_x86_64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">htmlgfx_26203_windows_arm64</TargetName>
+    <TargetName>$(ProjectName)_26203_windows_$(Platform)</TargetName>
```
AenBleidd, line 98: This will change architecture name from `x86_64` to `x64`. For compatibility target name architecture should be `x86_64`.

win_build/vboxwrapper_vs2019.vcxproj
```diff
@@ -95,10 +95,7 @@
     <CodeAnalysisRules Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" />
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">vboxwrapper_26203_windows_x86_64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">vboxwrapper_26203_windows_arm64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Release|x64'">vboxwrapper_26203_windows_x86_64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">vboxwrapper_26203_windows_arm64</TargetName>
+    <TargetName >$(ProjectName)_26203_windows_$(Platform)</TargetName>
```
AenBleidd, line 98: This will change architecture name from `x86_64` to `x64`. For compatibility target name architecture should be `x86_64`.

win_build/worker_vs2019.vcxproj
```diff
@@ -92,17 +92,8 @@
...
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">
-    <TargetName>worker_windows_x86_64</TargetName>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
-    <TargetName>worker_windows_x86_64</TargetName>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">
-    <TargetName>worker_windows_x86_64</TargetName>
+  <PropertyGroup >
+    <TargetName>$(ProjectName)_windows_$(Platform)</TargetName>
```
AenBleidd, line 96: This will change architecture name from `x86_64` to `x64`. For compatibility target name architecture should be `x86_64`.

win_build/wrapper_vs2019.vcxproj
```diff
@@ -90,10 +90,7 @@
     <CodeAnalysisRules Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" />
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Release|x64'">wrapper_26016_windows_x86_64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">wrapper_26016_windows_arm64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">wrapper_26016_windows_x86_64</TargetName>
-    <TargetName Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">wrapper_26016_windows_arm64</TargetName>
+    <TargetName>$(ProjectName)_26016_windows_$(Platform)</TargetName>
```
AenBleidd, line 93: This will change architecture name from `x86_64` to `x64`. For compatibility target name architecture should be `x86_64`.

AenBleidd (changes requested):

win_build/htmlgfx_vs2019.vcxproj
talregev: The architecture is x64.

win_build/vboxwrapper_vs2019.vcxproj
talregev: The architecture is x64.

win_build/worker_vs2019.vcxproj
talregev: The architecture is x64.

win_build/wrapper_vs2019.vcxproj
talregev: The architecture is x64.

win_build/htmlgfx_vs2019.vcxproj
AenBleidd: And should be `x86_64`

win_build/vboxwrapper_vs2019.vcxproj
AenBleidd: And should be `x86_64`

win_build/wrapper_vs2019.vcxproj
AenBleidd: And should be `x86_64`

win_build/worker_vs2019.vcxproj
AenBleidd: And should be `x86_64`

AenBleidd:
image

win_build/vboxwrapper_vs2019.vcxproj
talregev: And should be `x64`

win_build/worker_vs2019.vcxproj
talregev: And should be `x64`

win_build/wrapper_vs2019.vcxproj
talregev: And should be `x64`

win_build/htmlgfx_vs2019.vcxproj
talregev: And should be `x64`

AenBleidd:
@talregev, if you don't want to collaborate - I see no reason to continue useless discussion

talregev:
When you give a real explanation for the naming, and not copy part, with strong reasoning, We can start collaborate.
I agree your arguments was useless.

AenBleidd:
> When you give a real explanation for the naming, and not copy past, with strong reasoning, We can start collaborate. I agree your arguments was useless.

I wrote to you:

> This will change architecture name from `x86_64` to `x64`. For compatibility target name architecture should be `x86_64`.

What is not clear with the words `for compatibility`?

talregev:
It not strong reason. and you need to explain more if you want to preserve this compatibility

AenBleidd:
Why I should spend time to explain you what does `for compatibility` mean?
We have such a naming scheme for years (for example, you could see it here: https://boinc.berkeley.edu/dl/), and this is exactly `you` who need a strong reason to break this scheme now.
If I told you that smth should be made some way - that means that there is a reason for this, not just my wish. And I don't need to explain every time why.
I don't want to waste my time arguing for nothing.

talregev:
Don't waste your time, I just create another PR.

talregev:
I create this change mainly because we have a bug in our rename process:
worker of arm64 is also the same name of worker of x64. 

https://github.com/BOINC/boinc/pull/4678/files#diff-971d8b10f183ac019a1e5d61cbe4673ee113dddd77ed3e1441cc25b2fddfb343L105

You try to explain your logic but I am still not agree to it:
> I'll repeat one more time: some of the binaries are expected to have the exact architecture 'x86_64' because historically we have exactly it. And to not break plenty of projects that uses our binaries, we must not just rename it because you think it would be more correct.

Because we change the version number of all the binaries you complain with the architecture name. It mean the maintainers need to adapt the new binary with the new number. Also you remove versioning from some of the binaries, so the maintainers should to the new names as well.

So my strong motivation is to prevent this bug I show up, by generalize the process name, and make them more accurate. 
So in my precpective, you save some of backward compatibility in name, and some not. And if we already made a change, let fix the bug and have more new and accurate and automated binaries name.

talregev:
Please remove the wafting for author respond, because I already respond.

AenBleidd:
Since the main question is not resolved, I can't merge it. I keep it open for further discussion

AenBleidd (changes requested):
Requires further discussion

Issue #4116, opened 2020-12-01
Recommended Version Logic ignores OS Version requirements


CharlieFenton:
Although the examples and specific cases I have listed here are for Macintosh, this issue applies to all platforms.

This bug has four parts:

[1] The web logic which determines which "Recommended version" to offer a user fails to take into account the minimum version of the operating system each version of BOINC may require, or the CPU architecture it supports. For example, the latest version of BOINC for Macintosh (BOINC 7.16.14) requires at least MacOS 10.9.0, but it is shown on https://boinc.berkeley.edu/download.php for all Macintosh users including those running MacOS 10.7.5. It should instead offer the most recent recommended version for MacOS 10.7.5, which is BOINC 7.16.11.

In addition, BONC has builds available for PowerPC Macs, 32-bit Intel Macs, 64-bit Intel Macs and now the new arm64 "Apple Silicon" Macs. The web logic needs to take all these factors into account.

[2] Similarly, the code in the BOINC client which checks for a new version of BOINC fails to consider the version of the operating system or the CPU architecture. When BOINC 7.16.13 was released for macOS 10.9+, the client incorrectly told users running older versions of MacOS that a newer version is available for their computers.

[3] The BOINC wiki page https://boinc.berkeley.edu/wiki/System_requirements is misleading where it says:
> Operating system
Mac OS X 10.4.0 and later
Hardware
Macintosh computer with an Intel x86 or PowerPC G3, G4, or G5 processor

As an illustration of the confusion this causes, a user wrote [here](https://boinc.berkeley.edu/forum_thread.php?id=14054&postid=101759):
> the wiki says that the minimum system requirement is mac os 10.4, and i'm running 10.7.5 on a nice little i7-powered mini server. so, i think, i meet the minimum

The wiki page should somehow make clear that BOINC has builds available for each of these configurations, but some are supported only by older versions.

[4] The [download all page](https://boinc.berkeley.edu/download_all.php) needs a separate section for "MacOS (64-bit Intel and Apple Silicon) version 10.9.0+". 

The section "Mac OS X (64-bit Intel) Version 10.7.0+" should be changed to "MacOS (64-bit Intel) Version 10.7.0 - 10.15.7".

The section "Mac OS X (32-bit or 64-bit Intel) Version 10.4.0+" should be changed to "MacOS (32-bit or 64-bit Intel) Version 10.4.0 - 10.15.7".

[...]

PR #4685, opened 2022-04-01
client: Fixes for new version check

### LINKED ISSUE #4551: BOINC check for new versions won't work on arm64 Macs
**Describe the bug**
New Mac clients are universal binaries containing both arm64 and x86_64 code. 
  On Intel Macs, the primary platform is x86_64-apple-darwin and no secondary platform
  On arm64 Macs, the primary platform is arm64-apple-darwin and the secondary platform is x86_64-apple-darwin

But https://boinc.berkeley.edu/download.php shows only x86_64-apple-darwin:
```
<version>
    <platform>Apple Mac (64-bit Intel and Apple Silicon)</platform>
    <dbplatform>x86_64-apple-darwin</dbplatform>
    <description>Recommended version (Unix command-line version)</description>
    <date>16 Sep 2021</date>
    <version_num>7.16.19</version_num>
    <url>https://boinc.berkeley.edu/dl/boinc_7.16.19_universal-apple-darwin.dmg</url>
    <filename>boinc_7.16.19_universal-apple-darwin.dmg</filename>
    <size_mb>3.77</size_mb>
    <installer>Command-line client (no GUI)</installer>
</version>

```
universal binaries either need to be in there twice, once for each platform, or there needs to be a provision for multiple primary platforms in both client/current_vrsion.cpp
and https://boinc.berkeley.edu/download.php

**NOTE:** The first solution will work with existing clients, but the second solution will only work with new clients.  
 - OS: MacOS 11
 - BOINC Version: 7.16.19

CharlieFenton:
* Allow restricting to specified OS versions (implemented for MacOS only so far)
 * Show newest acceptable BOINC version even if not in ascending order in versions.inc and get_current_version.xml files
 * Remove cached newer version number from client_state file if the build was removed from versions.inc (e.g., if a bug was found)
* Fix typos and add comments

Fixes #4551

### DIFF
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2020 University of California
+// Copyright (C) 2022 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -645,7 +645,7 @@ int CLIENT_STATE::init() {
 
     // inform the user if there's a newer version of client
     // NOTE: this must be called AFTER
-    // read_vc_config_file()
+    // read_nvc_config_file()
     //
     newer_version_startup_check();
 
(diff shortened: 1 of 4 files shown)


AenBleidd (approved):
Code looks fine for me. I can't test this, and it would be also nice to let @davidpanderson review this also

CharlieFenton:
Fixes #4116

PR #4688, opened 2022-04-06
[WIP] Fix Xcode 13 (Apple clang version 13.1.6) compiler warnings part 2

CharlieFenton:
Fix Xcode 13 (Apple clang version 13.1.6) compiler warnings in mutl-platform source files

### DIFF
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2019 University of California
+// Copyright (C) 2022 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -152,7 +152,8 @@ static void handle_get_project_status(GUI_RPC_CONN& grc) {
 
 static void handle_get_disk_usage(GUI_RPC_CONN& grc) {
     unsigned int i;
-    double size, boinc_non_project, d_allowed, boinc_total;
+    double size, boinc_non_project, d_allowed;
+//    double boinc_total;
 
     grc.mfout.printf("<disk_usage_summary>\n");
     int retval = get_filesystem_info(
@@ -184,7 +185,7 @@ static void handle_get_disk_usage(GUI_RPC_CONN& grc) {
         }
     }
 #endif
-    boinc_total = boinc_non_project;
+//    boinc_total = boinc_non_project;
     gstate.get_disk_usages();
     for (i=0; i<gstate.projects.size(); i++) {
         PROJECT* p = gstate.projects[i];
@@ -195,7 +196,7 @@ static void handle_get_disk_usage(GUI_RPC_CONN& grc) {
             "</project>\n",
             p->master_url, p->disk_usage
         );
-        boinc_total += p->disk_usage;
+//        boinc_total += p->disk_usage;
     }
     d_allowed = gstate.allowed_disk_usage(gstate.total_disk_usage);
     grc.mfout.printf(
--- a/clientgui/AsyncRPC.cpp
+++ b/clientgui/AsyncRPC.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2013 University of California
+// Copyright (C) 2022 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -139,8 +139,10 @@ void *RPCThread::Entry() {
     CRPCFinishedEvent RPC_done_event( wxEVT_RPC_FINISHED );
     ASYNC_RPC_REQUEST *current_request;
     double startTime = 0;
+#if wxDEBUG_LEVEL
     wxMutexError mutexErr = wxMUTEX_NO_ERROR;
     wxCondError condErr = wxCOND_NO_ERROR;
+#endif
 
 #ifdef HAVE__CONFIGTHREADLOCALE
     _configthreadlocale(_ENABLE_PER_THREAD_LOCALE);
@@ -159,7 +161,10 @@ void *RPCThread::Entry() {
         // This does the following:
         // (1) Unlocks the Mutex and puts the RPC thread to sleep as an atomic operation.
         // (2) On Signal from main thread: locks Mutex again and wakes the RPC thread.
-        condErr = m_pRPC_Thread_Condition->Wait();
+#if wxDEBUG_LEVEL
+        condErr =
+#endif
+        m_pRPC_Thread_Condition->Wait();
         wxASSERT(condErr == wxCOND_NO_ERROR);
         
         if (m_pDoc->m_bShutDownRPCThread) {
@@ -188,9 +193,11 @@ void *RPCThread::Entry() {
         
         current_request->retval = retval;
 
-        mutexErr = m_pRPC_Request_Mutex->Lock();
+#if wxDEBUG_LEVEL
+        mutexErr =
+#endif
+        m_pRPC_Request_Mutex->Lock();
         wxASSERT(mutexErr == wxMUTEX_NO_ERROR);
-
         current_request->isActive = false;
         wxPostEvent( wxTheApp, RPC_done_event );
 
@@ -198,7 +205,10 @@ void *RPCThread::Entry() {
         // currently blocked by m_pRPC_Request_Condition->Wait[Timeout]()
         m_pRPC_Request_Condition->Signal();
         
-        mutexErr = m_pRPC_Request_Mutex->Unlock();
+#if wxDEBUG_LEVEL
+        mutexErr =
+#endif
+        m_pRPC_Request_Mutex->Unlock();
         wxASSERT(mutexErr == wxMUTEX_NO_ERROR);
     }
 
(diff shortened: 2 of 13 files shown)


CharlieFenton:
All the changes in this PR are in files used by BOINC Manager (except _clientscr/Mac_Saver_ModuleView.m_).  Most of these compiler warnings were "variable set but not used."

Some of these variables are used only to trigger asserts when there is a problem, using the `wxAssert` macro, does nothing when `wxDEBUG_LEVEL` is zero. So these variables are used only in debug builds, not in production (deployment) builds, which set wxDEBUG_LEVEL` to zero. In those cases, I fixed them by guarding the declaration and setting of the variable with `#if wxDEBUG_LEVEL`.

In a few cases, I have used the compiler directive `__attribute__((unused))`. Since all checks have passed (except one that failed for a reason not related to this PR), it this directive appears to be recognized at least on MacOS, MS Windows and Linux.

clientgui/AsyncRPC.cpp
```diff
@@ -159,7 +161,10 @@ void *RPCThread::Entry() {
         // This does the following:
         // (1) Unlocks the Mutex and puts the RPC thread to sleep as an atomic operation.
         // (2) On Signal from main thread: locks Mutex again and wakes the RPC thread.
-        condErr = m_pRPC_Thread_Condition->Wait();
+#if wxDEBUG_LEVEL
+        condErr =
```
AenBleidd, line 165: Maybe it would be better to check these function calls results in both Debug and Release? Just ignoring return value of async functions might lead to application crash or deadlock that is not nice either. And having `wxASSERT` here is not a solution either because such an errors might happen on Release only code when it's optimized (assuming that we're using at least 3 compilers (Microsoft cl for Windows, gcc for linux and clang for OSX)) that in result might produce the code that have different behavior on every platform.
So here I'd personally suggest to rewrite this function and make proper error handling instead of just ignoring function calls results in Release mode.

clientgui/AsyncRPC.cpp
```diff
@@ -614,7 +638,10 @@ int CMainDocument::RequestRPC(ASYNC_RPC_REQUEST& request, bool hasPriority) {
             // (2) On Signal from RPC thread: locks Mutex again and wakes the main thread.
             m_pRPC_Request_Condition->WaitTimeout(timeToSleep);
 
-            mutexErr = m_pRPC_Request_Mutex->Unlock();
+#if wxDEBUG_LEVEL
+            mutexErr =
+            m_pRPC_Request_Mutex->Unlock();
```
AenBleidd, line 643: I suspect this line shouldn't be in the `#if...#endif` block. Otherwise in release we will never unlock this mutex and will have a deadlock here

clientgui/BOINCBaseView.cpp
```diff
@@ -250,10 +250,15 @@ void CBOINCBaseView::OnListRender(wxTimerEvent& event) {
...
             } else {
                 int iIndex = 0;
+#if wxDEBUG_LEVEL
                 int iReturnValue = -1;
+#endif
                 if (iDocCount > iCacheCount) {
                     for (iIndex = 0; iIndex < (iDocCount - iCacheCount); iIndex++) {
-                        iReturnValue = AddCacheElement();
+#if wxDEBUG_LEVEL
+                        iReturnValue =
```
AenBleidd, line 259: Here's the same as before: I strongly suggest to check return value in Release mode also, and do this properly, otherwise in Release we might have unexpected results.

clientgui/BOINCTaskBar.cpp
```diff
@@ -433,7 +435,10 @@ bool CTaskBarIcon::SetIcon(const wxIcon& icon, const wxString& ) {
     else if (m_iconTaskBarSnooze.IsSameAs(icon))
         macIcon = wxImage(macsnoozebadge);
     else {
-        err = SetDockBadge(NULL);
+#if wxDEBUG_LEVEL
+        err =
```
AenBleidd, line 439: The same as described above: I suggest to add proper error handling in both Debug and Release modes

clientgui/MainDocument.cpp
```diff
@@ -1719,7 +1719,7 @@ void CMainDocument::KillGraphicsApp(int pid) {
         argv[4] = thePIDbuf;
         argv[5] = 0;
 
-        iRetVal = run_program(
+        run_program(
```
AenBleidd, line 1722: I'd here add proper error handling in case there was unsuccessful try to run this application (if we can't properly deal with this situation, we should at least write a message to the log file). Ignoring errors silently is not nice.

clientgui/WizardAttach.cpp
```diff
@@ -580,6 +579,9 @@ void CWizardAttach::_ProcessCancelEvent( wxWizardExEvent& event ) {
     GetNextButton()->Enable(HasNextPage(page));
     GetBackButton()->Enable(HasPrevPage(page));
 
+#if 0   // These appear to no longer be used
```
AenBleidd, line 582: I don't like dead code. So I'd suggest to remove it until you have strong arguments why we should kept it.

clientgui/sg_BoincSimpleFrame.cpp
```diff
@@ -830,7 +832,9 @@ void CSimpleFrame::OnConnect(CFrameEvent& WXUNUSED(event)) {
         // take care of attaching to projects when it completes the RPCs
         //
     } else if (ami.acct_mgr_url.size() && !ami.have_credentials) {
+#ifndef __WXMAC__
         wasShown = IsShown();
+    #endif
```
AenBleidd, line 837: Please align this line properly.

clientgui/sg_ProjectPanel.cpp
```diff
@@ -298,7 +298,7 @@ void CSimpleProjectPanel::ReskinInterface() {
 
     CMainDocument* pDoc = wxGetApp().GetDocument();
     ProjectSelectionData* selData;
-    PROJECT* project;
+    PROJECT* project __attribute__((unused));
```
AenBleidd, line 301: I believe this line and line 311 `project = pDoc->state.lookup_project(ctrl_url);` might be safely removed since they are doing nothing

clientgui/wizardex.h
```diff
@@ -246,7 +246,7 @@ class WXDLLIMPEXP_ADV wxWizardEx : public wxWizardBaseEx
     wxStaticBitmap *m_statbmp;  // the control for the bitmap
 
     // Whether user called SetBorder()
-    bool m_calledSetBorder;
+//    bool m_calledSetBorder;
```
AenBleidd, line 249: I believe we can just remove this dead code

AenBleidd (changes requested):
There is one error that should be definitely fixed. My other comments are mostly recommendations, and could be discussed.

clientgui/wizardex.h
CharlieFenton: This code was not written by anyone at BOINC. It was taken by Rom Walton from wxWidgets; see the copyright at the top of the file. So I am reluctant to change it beyond adding the `//`.

clientgui/AsyncRPC.cpp
CharlieFenton: The problem with your suggestion is that I have no idea what we could do in production if the call results in an error. So the asserts are there for debugging purposes. You definitely don't want a production app to crash unnecessarily, which is what an assert appears to do when not run under a debugger.

This code was very difficult to get working originally, and I have no interest in doing major work on it. In any case, what you are suggesting is well beyond the scope of this PR. If you want, please create a separate issue for it.

client/gui_rpc_server_ops.cpp
```diff
@@ -152,7 +152,8 @@ static void handle_get_project_status(GUI_RPC_CONN& grc) {
 
 static void handle_get_disk_usage(GUI_RPC_CONN& grc) {
     unsigned int i;
-    double size, boinc_non_project, d_allowed, boinc_total;
+    double size, boinc_non_project, d_allowed;
+//    double boinc_total;
```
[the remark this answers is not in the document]
CharlieFenton: I don't know why this code was there, so I am reluctant to completely remove it.

clientgui/AsyncRPC.cpp
AenBleidd: Ok, I agree with you. Let's keep it then 'as-is' for now. I'll put this in my 'nice to do' list and probably handle this properly one day.

clientgui/wizardex.h
AenBleidd: Ok, one day I'll go through all the commented code and remove it (especially all the code that was commented decades ago ^^)

clientgui/BOINCBaseView.cpp
CharlieFenton: This is not my code. Rom Walton wrote it many years ago. While I agree with you, it is well beyond the scope of this PR and not something I am willing to do. If you feel strongly, I suggest you create a separate issue.

clientgui/BOINCTaskBar.cpp
CharlieFenton: Again, this is not my code. Rom Walton wrote it many years ago. While I agree with you, it is well beyond the scope of this PR and not something I am willing to do. If you feel strongly, I suggest you create a separate issue.

clientgui/MainDocument.cpp
CharlieFenton: Again, this is not my code. Rom Walton wrote it many years ago. While I agree with you, it is well beyond the scope of this PR and not something I am willing to do. If you feel strongly, I suggest you create a separate issue.

AenBleidd (approved):

AenBleidd:
Ok, looks fine for me now. WIll merge it once all CI finish (mingw-vcpkg is expected to fail, so could be ignored for now)

PR #4695, opened 2022-04-07
[CI] Add openssl custom port to fix build issues

AenBleidd:


talregev (approved):
CI compile all android and windows. This is should be only temporary solution.

PR #4699, opened 2022-04-08
[cmake] Compile boinc libs with cmake

talregev:
Compile boinc libs with cmake.


talregev:
The cmake is inside lib folder and it compile with vcpk in manifest mode.
PR is ready.

AenBleidd (approved):
LGTM

PR #4702, opened 2022-04-09
[Android][vcpkg] Add android manifests vcpkg

### LINKED ISSUE #4696: [Android][vcpkg] Remove vcpkg pinning
Currently there is a major issue with Android and OpenSSL 3.0.2:
- Android vcpkg build is failed because of clang issue (https://github.com/openssl/openssl/issues/18059). It would be possible to workaround this if https://github.com/openssl/openssl/pull/18056 is merged but still using custom port only.

When https://github.com/microsoft/vcpkg/pull/24004 and https://github.com/openssl/openssl/pull/18056 are merged, new port should be created to properly fix that in vcpkg upstream

talregev:
Add android manifests vcpkg
Fixes #4696


android/buildAndroidBOINC-CI.sh
```diff
@@ -258,49 +258,44 @@ if [ $build_with_vcpkg = "yes" ]; then
...
         export ANDROID_NDK_HOME=$NDK_ARMV6_ROOT
 
-        $VCPKG_ROOT/vcpkg install $packages $vcpkg_flags --triplet=$(getTripletName $arch)
+        $VCPKG_ROOT/vcpkg install $manifests $vcpkg_flags --triplet=$(getTripletName $arch)
     fi
     if [ $arch = "arm" ]; then
         export ANDROID_NDK_HOME=$NDK_ROOT
 
-        $VCPKG_ROOT/vcpkg install $packages $vcpkg_flags --triplet=$(getTripletName $arch)
-        $VCPKG_ROOT/vcpkg install $packages $vcpkg_flags --triplet=$(getTripletName neon)
```
AenBleidd, line 285: Why neon is removed?

android/vcpkg_json_apps/vcpkg.json
```diff
@@ -0,0 +1,19 @@
+{
```
AenBleidd, line 1: Please rename this folder to one of the next:
- apps
- vcpkg-config-apps
- vcpkg-apps

Other folders should be renamed correspondingly

android/vcpkg_json_apps/vcpkg.json
```diff
@@ -0,0 +1,19 @@
...
+    "name": "libboinc",
```
AenBleidd, line 2: Change this to 'boinc-apps'

Other components rename correspondingly

AenBleidd (changes requested):

android/buildAndroidBOINC-CI.sh
talregev: Because each time it delete the previous triplet. In manifests mode currently cannot compile multiple triplets.
I can change the order, but I see this pointless.

AenBleidd: Then we need a new target (e.g. arm -neon) because I believe we still need this functionality. Otherwise it's a regression and need to be separately discussed.

android/vcpkg_config_libs/vcpkg.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "boinc-libs",
```
AenBleidd, line 2: Hm, my bad. This should be still 'libboinc'

AenBleidd (changes requested):

android/buildAndroidBOINC-CI.sh
talregev: We never had it. it for different PR.

AenBleidd: Then why did you add it at all?

android/vcpkg_config_libs/vcpkg.json
AenBleidd: This is still need to be fixed

AenBleidd (changes requested):
One issue left unresolved

android/vcpkg_config_libs/vcpkg.json
talregev: This is not the same as libboinc cmake.
This is libs component of boinc, as the name suggest.

android/vcpkg_config_client/vcpkg.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "boinc-client",
+    "version-string": "7.19.0",
+    "builtin-baseline":"bd1ef2df46303989eeb048eb7aa9b816aa46365e",
+    "dependencies":
+    [
+        {
+          "name": "curl",
```
AenBleidd, line 8: When we are building current master with curl, we build it with `openssl` feature only: `curl[core,openssl]:armv6-android -> 7.82.0`
After this change additional configuration is passed to curl that will have negative impact: `curl[core,non-http,openssl,ssl]:armv6-android -> 7.82.0`
This should be fixed here and in two other vcpkg.json files

AenBleidd (changes requested):

android/vcpkg_config_client/vcpkg.json
talregev: Cannot do it with manifest mode.

AenBleidd: You can do it in manifest mode. E.g. https://github.com/BOINC/boinc-apps/blob/master/boinc-autodock-vina/vcpkg.json#L12-L15

AenBleidd (approved):
LGTM now

PR #4703, opened 2022-04-10
[CI] Try to fix Android build with OpenSSL 3.0.2

AenBleidd:


CharlieFenton:
@AenBleidd One way I have fixed similar problems with dependent libraries on the Mac is to add code to the library build script that applies a patch to the library source code. You can find some examples in mac_build/buildc-ares.sh, mac_build/buildcurl.sh and mac_build/buildWxMac.sh.

AenBleidd:
@CharlieFenton, yeah, I also thought about this. Not really sure this is needed. We're building x86 code on 14(!) years old compiler. Not sure anyone still run this Android app on some 14 years old hardware that is even not a cell phone.
So if might be a good idea just to update toolkit and go with it.
In case anyone will complain - we can then consider whether we want to support that particular case and create a custom patch for OpenSSL or just drop support of such an old OS/hardware.

PR #4710, opened 2022-04-15
[MinGW][cmake] Add cmake libs in mingw

talregev:
Add cmake libs in mingw.

### DIFF
--- a/.github/workflows/mingw.yml
+++ b/.github/workflows/mingw.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        type: [libs-mingw, apps-mingw, apps-mingw-vcpkg]
+        type: [libs-mingw, apps-mingw, apps-mingw-vcpkg, libs-mingw-cmake]
       fail-fast: false
     steps:
       - uses: actions/checkout@v2
@@ -62,8 +62,8 @@ jobs:
           # Install PowerShell
           sudo apt install -y powershell
 
-      - name: Automake vcpkg apps
-        if: success() && matrix.type == 'apps-mingw-vcpkg'
+      - name: Automake vcpkg and cmake
+        if: success() && (contains(matrix.type, 'vcpkg') || contains(matrix.type, 'cmake'))
         run: ./_autosetup
 
       - name: Configure mingw vcpkg apps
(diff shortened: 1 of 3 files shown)


talregev:
Ready for review.

.github/workflows/mingw.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        type: [libs-mingw, apps-mingw, apps-mingw-vcpkg]
+        type: [libs-mingw, apps-mingw, apps-mingw-vcpkg, libs-cmake]
```
AenBleidd, line 21: Better to rename to 'libs-mingw-cmake'

mingw/bootstrap_vcpkg_cmake.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+set -e
+
+OS_NAME="mingw"
```
AenBleidd, line 4: 'mingw' is not an OS, maybe better to rename 'OS_NAME' to 'TARGET_PLATFORM_NAME' or smth similar?

AenBleidd (changes requested):

mingw/bootstrap_vcpkg_cmake.sh
talregev: plt_name

talregev: `PLATFORM_NAME`

talregev: done.

.github/workflows/mingw.yml
talregev: done.

mingw/bootstrap_vcpkg_cmake.sh
AenBleidd: You changed this in ci_configure_libs_cmake.sh but not in this file. Please fix it here also

AenBleidd (changes requested):

mingw/bootstrap_vcpkg_cmake.sh
talregev: done.

AenBleidd (approved):
LGTM

PR #4713, opened 2022-04-16
[Android][cmake] Add android libs cmake

talregev:
Add android libs cmake.

### DIFF
--- a/android/ci_build_libs_cmake.sh
+++ b/android/ci_build_libs_cmake.sh
@@ -1,0 +1,42 @@
+#!/bin/sh
+set -e
+
+PLATFORM_NAME="android"
+
+if [ ! -d "$PLATFORM_NAME" ]; then
+    echo "start this script in the source root directory"
+    exit 1
+fi
+
+BUILD_DIR="$PWD/3rdParty/$PLATFORM_NAME"
+VCPKG_ROOT="$BUILD_DIR/vcpkg"
+VCPKG_PORTS="$PWD/3rdParty/vcpkg_ports"
+. $PLATFORM_NAME/ndk_common.sh
+
+if [ ! -d $NDK_ROOT ]; then
+    createNDKFolder
+fi
+
+if [ ! -d $NDK_ARMV6_ROOT ]; then
+    createNDKARMV6Folder
+fi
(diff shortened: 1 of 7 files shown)


talregev:
Ready for review.

android/ci_build_libs_cmake.sh
```diff
@@ -0,0 +1,58 @@
...
+set -e
+
+PLATFORM_NAME="android"
+
+if [ ! -d "$PLATFORM_NAME" ]; then
+    echo "start this script in the source root directory"
+    exit 1
+fi
+
+BUILD_DIR="$PWD/3rdParty/$PLATFORM_NAME"
```
AenBleidd, line 11: This file contains a lot of copy-pasted code from this file: https://github.com/BOINC/boinc/blob/master/android/buildAndroidBOINC-CI.sh
Maybe this is the right place to move common code to some other file and call it from there?

talregev: I want to think about this.
Maybe import common bash file.

talregev: I create a common ndk file.

talregev:
Ready for review.

android/ci_build_libs_cmake.sh
```diff
@@ -0,0 +1,40 @@
...
+VCPKG_ROOT="$BUILD_DIR/vcpkg"
+VCPKG_PORTS="$PWD/3rdParty/vcpkg_ports"
+. $PLATFORM_NAME/common_ndk.sh
```
AenBleidd, line 14: Please rename to 'ndk_common' because we do not have common ndk but instead we have some common stuff for ndk, so this name will be more logically correct

AenBleidd (changes requested):

android/ci_build_libs_cmake.sh
talregev: What name you want me to rename?

talregev: `ndk_common` sound more reasonable, thank you.

AenBleidd (approved):

PR #4726, opened 2022-04-23
[CI] Add Transifex flow to automatically update translations and templates

AenBleidd:

### DIFF
--- a/.github/workflows/transifex.yml
+++ b/.github/workflows/transifex.yml
@@ -1,0 +1,44 @@
+name: Transifex
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    - cron:  '25 12 * * 0'
+
+jobs:
+  build:
+    name: ${{ matrix.type }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        type: [update-templates, update-translations]
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - name: Install dependencies
+        run: |
+          sudo apt-get -qq update
+          sudo apt-get install -y gettext translate-toolkit transifex-client
+      - name: Configure Transifex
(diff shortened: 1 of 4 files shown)


.github/workflows/transifex.yml
```diff
@@ -0,0 +1,39 @@
...
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - name: Install dependencies
+        run: |
+          sudo apt-get -qq update
+          sudo apt-get install -y gettext translate-toolkit transifex-client
+      - name: Configure Transifex
+        run: |
+          echo "[https://www.transifex.com]\n" > ~/.transifexrc
```
talregev, line 28: You should remove \n from the end. It already separate the line for each echo.

.github/workflows/transifex.yml
```diff
@@ -0,0 +1,39 @@
...
+          echo "hostname = https://www.transifex.com\n" >> ~/.transifexrc
+          echo "username = api\n" >> ~/.transifexrc
+          echo "token = ${{ secrets.TRANSIFEX_API_KEY }}\n" >> ~/.transifexrc
```
talregev, line 31: Also I believe the `secrets.TRANSIFEX_API_KEY` is empty.

.github/workflows/transifex.yml
AenBleidd: Yes, you're right. Removed

.github/workflows/transifex.yml
AenBleidd: No, it's not. I added it, you will just never see it on CI

talregev:
```bash
*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"
```
You should look solution like this: 
https://github.com/marketplace/actions/add-commit

talregev:
I can help you if you want.

AenBleidd:
Thanks, I'm fine

PR #4727, opened 2022-04-23
[CI] Add cron job to update nightly new translations from Transifex

AenBleidd:

### DIFF
--- a/locale/update_translations.sh
+++ b/locale/update_translations.sh
@@ -17,7 +17,8 @@ command -v tx >/dev/null 2>&1 || { echo >&2 "tx (transifex-client) is needed but
 
 # check if working directory is clean to ensure we only commit localization changes
 if test 0 -ne `git status -s -uno |wc -l`; then
-  echo "Warning: You have pending changes! Please make sure to only commit localization changes!"
+  echo "Error: You have pending changes! Please make sure to only commit localization changes!"
+  exit 1
 fi
 
 echo "pulling translations from transifex"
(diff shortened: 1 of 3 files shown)


AenBleidd:
@talregev, PR is ready for review

locale/update_translations.sh
```diff
@@ -57,10 +58,9 @@ done
 
 cd ${srcdir}
 
-echo "Translations compiled successfully. Now some manual steps:"
-echo " 1. less BOINC-Manager-pofilter.txt # check output from pofilter and adjust translations then start this script again"
-echo " 2. git add -u # only update already tracked files (add new files when needed too)"
-echo " 3. git commit -m \"Locale: Update localization files [skip ci]\""
-echo " 4. git push"
+git add .
```
talregev, line 61: After it git add, what happen? The code is not updated. Why you delete git push?

locale/update_translations.sh
```diff
@@ -57,10 +58,9 @@ done
...
+git diff-index --cached --quiet HEAD || { git commit -m "Locale: Update localization files"; git push origin HEAD; }
```
talregev, line 62: Now I see the code is change here. How it push without a password?

AenBleidd: GitHub Actions have embedded credentials, so when you do a push from the CI, you are able to push commits to the repo from which CI is run.

talregev (approved):
I didn't test it. The changes seems ok. We should test it.

AenBleidd:
After careful consideration and offline discussion with @CharlieFenton, I decided not to continue with this functionality and decided to keep this a part of Release process. Closing this for now.

PR #4728, opened 2022-04-23
[CI] Update templates on Transifex

AenBleidd:

### DIFF
--- a/.github/workflows/transifex-update-templates.yml
+++ b/.github/workflows/transifex-update-templates.yml
@@ -1,0 +1,38 @@
+name: Transifex
+on:
+  push:
+    branches: [ master ]
+  schedule:
+    - cron:  '0 0 * * *'
+
+jobs:
+  build:
+    name: ${{ matrix.type }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        type: [update-templates]
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - name: Install dependencies
+        run: |
+          sudo apt-get -qq update
+          sudo apt-get install -y gettext translate-toolkit transifex-client
+      - name: Configure Transifex
+        run: |
+          echo $'[https://www.transifex.com]' > ~/.transifexrc
+          echo $'api_hostname = https://api.transifex.com' >> ~/.transifexrc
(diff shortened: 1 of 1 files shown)


AenBleidd:
@talregev, ready for review

talregev:
Hi, How I know this is not fail? Currently it only schedule. 
Also you want bot will commit changes? 
I think it the best if it will be per PR, and not just schedule job. Will you consider to switch?

talregev:
I think I didn't understand the purpose of this PR.

.github/workflows/transifex-update-templates.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Configure Transifex
+        run: |
+          echo $'[https://www.transifex.com]' > ~/.transifexrc
+          echo $'api_hostname = https://api.transifex.com' >> ~/.transifexrc
+          echo $'hostname = https://www.transifex.com' >> ~/.transifexrc
+          echo $'username = api' >> ~/.transifexrc
+          echo $'password = ${{ secrets.TRANSIFEX_API_KEY }}' >> ~/.transifexrc
+      - name: Setup git config
+        run: |
+          git config user.name "GitHub Actions Bot"
```
talregev, line 33: Why you update user name and email it there isn't git change? 
Should be remove.

AenBleidd: This is a git setup. It will be required if git commit should be made from this PR. See no reason to add a condition to add git credentials when making commit only.

AenBleidd:
> Hi, How I know this is not fail? Currently it only schedule. 

I tested this locally, so just to commit and wait for scheduled run.

> 
> Also you want bot will commit changes? 

Yes

> I think it the best if it will be per PR, and not just schedule job. Will you consider to switch? 

These are trivial changes: making this through PR will only add additional complexity and will require PR review between two scheduled runs, otherwise there will be a mess.

.github/workflows/transifex-update-templates.yml
talregev: Where you commit/push from this PR?

talregev (approved):
I didn't test it. The changes seems ok. More testing is require.

AenBleidd:
After careful consideration and offline discussion with @CharlieFenton, I decided not to continue with this functionality and decided to keep this a part of Release process. Closing this for now.

PR #4734, opened 2022-04-28
Remove outdated files

AenBleidd:


CharlieFenton:
@AenBleidd Please don't remove _clientscr/BOINCSaver_MacOS10_6_7.zip_ yet, although you are correct that it is obsolete. The Mac installer and the scripts that build the installer (_mac_installer/release_boinc.sh_ and  _mac_installer/release_brand.sh_) still expect it to be present. 

Versions of MacOS before OS 10.8 required an older build of the BOINC screensaver which could not be built with newer versions of Xcode. So the installer contained this prebuilt older screensaver to install for those older operating systems.

Now that the minimum MacOS we support is 10.10, _clientscr/BOINCSaver_MacOS10_6_7.zip_. is no longer needed. But I need to modify the installer and scripts before we can remove it I'll create a PR in the next few days to do that.

CharlieFenton:
@AenBleidd On second thought, you can go ahead and remove _clientscr/BOINCSaver_MacOS10_6_7.zip_ since I will soon create a PR that removes references to it in the installer and scripts.

I see no other issues with the Mac files you are removing.

CharlieFenton (approved):

AenBleidd:
@CharlieFenton, we can go then other way: I'll remove this file from my PR (so once merged, this archive will still be there), and you can later remove it in your PR that also removes all references to it. In this case we will have a stable state of OSX build and you will not feel pushed from my side to create this PR asap :)

AenBleidd:
@davidpanderson, could you please also review this file and verify that these files are no longer needed.
Thanks in advance

CharlieFenton:
> we will have a stable state of OSX build and you will not feel pushed from my side to create this PR asap

That is correct only if by _stable state of OSX build_ you don't mean anything that we can release. 
The script which builds the installer will have an error because of the missing file until I create my PR and it is pushed.

CharlieFenton:
@AenBleidd Sorry, I misunderstood your [comment](https://github.com/BOINC/boinc/pull/4734#issuecomment-1112388855). I will create the new PR and remove the outdated file as part of my PR.

PR #4739, opened 2022-04-30
[Bugfix] Fix condor app to compile with curl vcpkg

talregev:
Fix condor app to compile with curl vcpkg.

### DIFF
--- a/mingw/ci_make_apps.sh
+++ b/mingw/ci_make_apps.sh
@@ -14,7 +14,7 @@ export CXX="x86_64-w64-mingw32-g++"
 export CC="x86_64-w64-mingw32-gcc"
 # condor
 export MINGW_FLAGS="-Dflockfile=_lock_file -Dfunlockfile=_unlock_file"
-export MINGW_CURL="-lcurl -lwinmm -lws2_32 -lpthread -lssl -lcrypto -lcrypt32 -lws2_32 -lzlib -ladvapi32 -lcrypt32"
+export CURL_EXTRA_LDFLAGS="-lcurl -lwinmm -lpthread -lssl -lcrypto -lws2_32 -lzlib -ladvapi32 -lcrypt32"
 # wrapper
 export MINGW_WRAPPER_FLAGS="-DEINSTEINATHOME_CROSS_BUILD -DMINGW_WIN32 -DHAVE_STRCASECMP -D_WINDOWS -D_WIN32 -DWIN32 -DWINVER=0x0500 -D_WIN32_WINNT=0x0500 -D_MT -DBOINC -DNODB -D_CONSOLE -fexceptions"
 export MINGW_ZIP_FIX="../../zip/zip/__p___mb_cur_max.c"
--- a/samples/condor/Makefile
+++ b/samples/condor/Makefile
@@ -17,6 +17,28 @@ else
   MAKEFILE_LDFLAGS  = -lpthread
 endif
 
+CXXFLAGS += \
+    -I$(BOINC_DIR) \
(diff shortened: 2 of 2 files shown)


mingw/ci_make_apps.sh
```diff
@@ -14,7 +14,7 @@ export CXX="x86_64-w64-mingw32-g++"
 export CC="x86_64-w64-mingw32-gcc"
 # condor
 export MINGW_FLAGS="-Dflockfile=_lock_file -Dfunlockfile=_unlock_file"
-export MINGW_CURL="-lcurl -lwinmm -lws2_32 -lpthread -lssl -lcrypto -lcrypt32 -lws2_32 -lzlib -ladvapi32 -lcrypt32"
+export CURL_EXTRA_LDFLAGS="-lcurl -lwinmm -lws2_32 -lpthread -lssl -lcrypto -lcrypt32 -lws2_32 -lzlib -ladvapi32 -lcrypt32"
```
AenBleidd, line 17: '-lws2_32' and '-lcrypt32' are specified twice. Please remove duplicates.

samples/condor/Makefile
```diff
@@ -29,6 +51,6 @@ distclean: clean
 distclean-recursive: clean
 
 boinc_gahp: boinc_gahp.cpp $(BOINC_SOURCE_LIB_DIR)/remote_submit.h $(BOINC_SOURCE_LIB_DIR)/remote_submit.cpp $(BOINC_DIR)/svn_version.h $(BOINC_DIR)/version.h
-	$(CXX) $(CXXFLAGS) $(CPPFLAGS) $(LDFLAGS) $(MINGW_FLAGS) -g -O2 -I$(BOINC_DIR) -I$(BOINC_SOURCE_LIB_DIR) \
-	-I$(VCPKG_DIR)/include -o boinc_gahp boinc_gahp.cpp $(BOINC_SOURCE_LIB_DIR)/remote_submit.cpp \
-	-L$(BOINC_LIB_DIR) -L$(VCPKG_DIR)/lib -lcurl $(MINGW_CURL) -lssl -lcrypto -lboinc $(MAKEFILE_LDFLAGS) $(STDCPPTC)
+	$(CXX) $(CXXFLAGS) $(CPPFLAGS) $(LDFLAGS) $(MINGW_FLAGS) -g -O2 \
+	-o boinc_gahp boinc_gahp.cpp $(BOINC_SOURCE_LIB_DIR)/remote_submit.cpp \
+	-lcurl -lssl -lcrypto -lboinc $(CURL_EXTRA_LDFLAGS) $(MAKEFILE_LDFLAGS) $(STDCPPTC)
```
AenBleidd, line 56: In case on non-vcpkg build, '-lcurl', '-lssl' and '-lcrypto' are already specified in 'CURL_EXTRA_LDFLAGS'.
Maybe it's reasonable to remove duplicates?

AenBleidd (changes requested):

mingw/ci_make_apps.sh
talregev: I can try to remove them, but:
- This is the exact flags taken from makefile. 
- Sometimes duplicate flags is needed to specify the compiler to use this flag for each compiler, especially in old compilers. 

My recommendation is to leave it as it is.

samples/condor/Makefile
talregev: It not specify in the `CURL_EXTRA_LDFLAGS`. If I will remove them from here, I will need to create another variable for all that it will be the same.

AenBleidd (approved):

PR #4742, opened 2022-05-06
[Android] test openssl 3.0.3 android client

talregev:
test openssl 3.0.3 android client.


talregev:
openssl 3.0.3 doesn't work for android. We will wait for official fix.

AenBleidd:
It is required to fix the port by providing an additional configure parameter. I'll fix it in two weeks

talregev:
Will you fix it on vcpkg?

AenBleidd:
Yes

talregev:
> Yes

Thank you

AenBleidd:
@talregev, I believe you mentioned several times that I never do any vcpkg port changes that are never pushed to upstream (only in case this is needed for testing or while waiting for approval on upstream).

PR #4744, opened 2022-05-10
[Scripts] Unite vcpkg link in shell scripts in linux, mingw, android and osx platforms

talregev:
Unite vcpkg link in shell scripts in linux, mingw, android and osx platforms


AenBleidd:
This PR might not solve an issue as you want: usually when you test smth with custom vcpkg link, you also specify a custom branch to checkout. Don't you think this case should be handles as well, @talregev?

talregev:
This case also can handle. For example:
`export VCPKG_LINK="https://github.com/talregev/vcpkg -b custom_branch"`

AenBleidd (changes requested):
We have already vcpkg_ports folder in 3rdParty directory. I think it would be logically correct to put this new configuration file to 3rdParty directory as well than just create a new folder for this purpose. Also it will keep vcpkg configuration closer to other vcpkg stuff, so this would be more convenient to work with it.

AenBleidd (approved):

PR #4747, opened 2022-05-13
[Android] Android get latest libs: curl,zlib,rappture

talregev:
Android get latest libs: curl,zlib,rappture

### DIFF
--- a/android/buildAndroidBOINC-CI.sh
+++ b/android/buildAndroidBOINC-CI.sh
@@ -252,11 +252,11 @@ if [ $build_with_vcpkg = "yes" ]; then
         git -C $BUILD_DIR clone $VCPKG_LINK
     fi
     if [ ! -e /tmp/vcpkg_updated ]; then
-        git -C $VCPKG_ROOT reset --hard
         git -C $VCPKG_ROOT pull
         $VCPKG_ROOT/bootstrap-vcpkg.sh
         touch /tmp/vcpkg_updated
     fi
+    sed -i "s/\"builtin-baseline\":.*,/\"builtin-baseline\": \"$(git -C $VCPKG_ROOT rev-parse HEAD)\",/" vcpkg_config_$component/vcpkg.json
     if [ $arch = "armv6" ]; then
         export ANDROID_NDK_HOME=$NDK_ARMV6_ROOT
 
--- a/android/vcpkg_config_apps/vcpkg.json
+++ b/android/vcpkg_config_apps/vcpkg.json
@@ -1,7 +1,7 @@
 {
     "name": "boinc-apps",
     "version-string": "7.19.0",
-    "builtin-baseline":"bd1ef2df46303989eeb048eb7aa9b816aa46365e",
+    "builtin-baseline": "",
     "dependencies":
     [
         {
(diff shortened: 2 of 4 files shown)


AenBleidd:
Does this mean that OpenSSL3 now builds without any issue on Android x86?

talregev:
> Does this mean that OpenSSL3 now builds without any issue on Android x86?

No. it still take the old openssl 1.1.1n.
But take the latest other libraries.
It just updated automated the latest vcpkg master inside the json file.

talregev:
> Does this mean that OpenSSL3 now builds without any issue on Android x86?

We still need your fix for openssl 3.0.3 for Android.

AenBleidd:
If you want - you can implement this fix by yourself and don't wait for me.

talregev:
> If you want - you can implement this fix by yourself and don't wait for me.



> If you want - you can implement this fix by yourself and don't wait for me.

I didn't understand your fix, so it better you do it. Also, you can review my PR until you have time for your fix.

talregev:
> If you want - you can implement this fix by yourself and don't wait for me.



> If you want - you can implement this fix by yourself and don't wait for me.

I didn't understand your fix, so it better you do it. Also, you can review my PR until you have time for your fix.

AenBleidd:
There is an issue with this PR: every time you build binaries locally, vcpkg.json files will be changed, and every time you need to remember to not commit these changes.
Is there any way to eliminate this?

talregev:
> There is an issue with this PR: every time you build binaries locally, vcpkg.json files are changed, and every time you need to remember to not commit these changes. Is there any way to eliminate this?

Tell me the final thing you want to achieve.

talregev:
> There is an issue with this PR: every time you build binaries locally, vcpkg.json files are changed, and every time you need to remember to not commit these changes. Is there any way to eliminate this?

avoid commits unchanged files?
I think we can put them in ignore

AenBleidd:
We can't put them in ignore because then you have to remember to force add them when you change them (e.g. by setting a new version).

I want to avoid errors as much as possible, and currently your solution will create errors later.

As far as I see, this later will not be needed when OpenSSL 3 issue for Android x86 is fixed.

So I'd not merge it at all for now.
The only case when we might need this is if we want to fix some particular version of any dependency on release branch but in that case vcpkg.json file change is expected.

So maybe it's better to always keep the last committed value of 'builtin-baseline' and update it from the script on demand only by passing some specific parameter to build script?

talregev:
> We can't put them in ignore because then you have to remember to force add them when you change them (e.g. by setting a new version).
> 
> I want to avoid errors as much as possible, and currently your solution will create errors later.
> 
> As far as I see, this later will not be needed when OpenSSL 3 issue for Android x86 is fixed.
> 
> So I'd not merge it at all for now. The only case when we might need this is if we want to fix some particular version of any dependency on release branch but in that case vcpkg.json file change is expected.
> 
> So maybe it's better to always keep the last committed value of 'builtin-baseline' and update it from the script on demand only by passing some specific parameter to build script?

@AenBleidd When are you going to fix android openssl vcpkg? If it soon, there is no point to discuss.

AenBleidd:
Fix is provided and currently in review: https://github.com/microsoft/vcpkg/pull/24801
@talregev, should we close this PR without merging or you have some other plans on it?

talregev:
> Fix is provided and currently in review: https://github.com/microsoft/vcpkg/pull/24801
> @talregev, should we close this PR without merging or you have some other plans on it?

We should wait for your fix. After that we can discuss. Also if your fix is delay we can discuss again.

PR #4749, opened 2022-05-19
Test fix

AenBleidd:


talregev:
This should be on vcpkg, and not in our end

PR #4750, opened 2022-05-20
[Test] Android openssl fix

talregev:
Test Android openssl fix.
- [ ] Wait for merge: https://github.com/microsoft/vcpkg/pull/24801
- [ ] Change vcpkg link


AenBleidd:
This is a duplicate of #4749. Why do you need to test it again?
And btw. your test is not correct because you're still overriding vcpkg to use OpenSSL 1.1.1* instead of OpenSSL3.*

talregev:
> This is a duplicate of #4749. Why do you need to test it again

I think this is the way we should test the fix. 

> your test is not correct because you're still overriding vcpkg to use OpenSSL 1.1.1* instead of OpenSSL3.*
I agree. I will change it here.

talregev:
After your PR will merge, I can change the link again, and This PR will ready to do to boinc.

PR #4751, opened 2022-05-20
Android openssl atomic bugfix with ndk update

### LINKED ISSUE #4696: [Android][vcpkg] Remove vcpkg pinning

talregev:
Android openssl atomic bugfix with ndk update.
Fixes https://github.com/microsoft/vcpkg/issues/24021 
Fixes #4696


talregev:
@AenBleidd This is fix the openssl problem atomic with update of clang (update ndk). I think this fix is better, and we should cancel your fix to vcpkg.

talregev:
@AenBleidd This fix interduce a new problem. That vcpkg compile arm with neon, even if the neon set to False.
I want to investigate this problem. It ok if you put your vcpkg PR on draft?

android/ci_build_libs_cmake.sh
```diff
@@ -35,7 +35,7 @@ for TRIPLET in $TRIPLETS_LIST ; do
         export ANDROID_NDK_HOME=$NDK_ROOT
     fi
     BUILD_TRIPLET=build-$TRIPLET
-    cmake lib -B $BUILD_TRIPLET -DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake -DVCPKG_OVERLAY_PORTS=$VCPKG_PORTS/ports -DVCPKG_OVERLAY_TRIPLETS=$VCPKG_PORTS/triplets/ci -DVCPKG_TARGET_TRIPLET=$TRIPLET -DVCPKG_MANIFEST_DIR=lib/vcpkg_config_android -DVCPKG_INSTALL_OPTIONS=--clean-after-build
+    cmake lib -B $BUILD_TRIPLET -DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake -DVCPKG_OVERLAY_PORTS=$VCPKG_PORTS/ports -DVCPKG_OVERLAY_TRIPLETS=$VCPKG_PORTS/triplets/ci -DVCPKG_TARGET_TRIPLET=$TRIPLET -DVCPKG_INSTALL_OPTIONS=--clean-after-build
```
AenBleidd, line 38: Why did you remove this manifest file?

talregev: Because we now can use the same manifest cmake file for all os. (linux, windows and android).

AenBleidd (approved):

PR #4752, opened 2022-05-21
[Android] Update android dependencies

talregev:
Update android dependencies.
Compile, and run and crunch on One plus One (android 9) and emulator Arm7 (android 4.1)

### DIFF
--- a/android/BOINC/.idea/misc.xml
+++ b/android/BOINC/.idea/misc.xml
@@ -45,5 +45,5 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_X" project-jdk-name="11" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_16" project-jdk-name="11" project-jdk-type="JavaSDK" />
 </project>
--- a/android/BOINC/app/src/main/AndroidManifest.xml
+++ b/android/BOINC/app/src/main/AndroidManifest.xml
@@ -17,7 +17,6 @@
   along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-        package="edu.berkeley.boinc"
         android:installLocation="internalOnly">
     <!-- installation on SD card would break boot receiver -->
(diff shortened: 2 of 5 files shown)


android/BOINC/.idea/misc.xml
```diff
@@ -45,5 +45,5 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_X" project-jdk-name="11" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_16" project-jdk-name="11" project-jdk-type="JavaSDK" />
```
AenBleidd, line 48: This file is autogenerated and should be removed from the repo at all. Why did you make this change here?

android/BOINC/app/src/main/AndroidManifest.xml
```diff
@@ -17,7 +17,6 @@
   along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-        package="edu.berkeley.boinc"
```
AenBleidd, line 20: Why did you remove this?

AenBleidd (changes requested):

android/BOINC/app/src/main/AndroidManifest.xml
talregev: Android studio suggest to remove it from here and add it on build.gradle. I didn't object to that suggestion.

android/BOINC/.idea/misc.xml
talregev: We have partial android studio project in our repo. In first I was object but now I understand the advantage of that. That it faster to load your android studio and it similar dev for all dev.

android/BOINC/app/src/main/AndroidManifest.xml
talregev: It did automated autogenerated suggestion

talregev:
I am putting this PR on hold. I want to test more.

talregev:
I tested it again couple time and check against the master, and I see that we good to go.
@AenBleidd Do you want to test it local also from your side?

talregev:
Thank you!

PR #4755, opened 2022-05-30
Install headers required to build wrapper app

CyberTailor:


CyberTailor:
I package `boinc-wrapper` as a standalone application and I need those headers :)

PR #4766, opened 2022-06-09
Mac: eliminate spurious error message in release scripts

CharlieFenton:
The Mac release scripts _release_boinc.sh_ and _release_brand.sh_ invoke the hdiutil command to create the disk image file used to distribute the bare core (command-line) client. This command sometimes expects user password authentication but does not actually request it when running in a shell script, so it displays "Authentication error" in the terminal. 

This change provides the authentication by running it with `sudo` since the user has previously provided the admin password to the script's earlier invocations of `sudo`, and subsequent invocations of `sudo` within a certain time limit don't require repeated entry of the password.

Eliminating the "Authentication error" message is merely a cosmetic change. In my experience, the disk image file is created and populated correctly by these scripts even when the message appears.

### DIFF
--- a/mac_installer/release_boinc.sh
+++ b/mac_installer/release_boinc.sh
@@ -2,7 +2,7 @@
 
 # This file is part of BOINC.
 # http://boinc.berkeley.edu
-# Copyright (C) 2021 University of California
+# Copyright (C) 2022 University of California
 #
 # BOINC is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


CharlieFenton:
This PR is ready for merge when the checks have passed (though the files it changes are not involved in any of the CI checks.)

AenBleidd (approved):
lgtm. Thanks, @CharlieFenton.
I'll wait to CI to finish and merge it afterwards.

PR #4770, opened 2022-06-09
Web: fix some PHP 8 deprecation warnings

davidpanderson:
Note: if a function expects, say, a string (and not null)
and you get a warning about a null arg,
don't fix it by casting the param to string.
That masks possible bugs.
Instead, find out where null is being passed, and fix it there.

web: server status: fix case where app has no results

web: fix case where get_logged_in_user() is called twice,
first with must_be_logged_in == false and later true.

web: xml_error(): $num default is -1

web: validate_email_addr(): tell user to check spam folder if no email

Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/html/inc/boinc_db.inc
+++ b/html/inc/boinc_db.inc
@@ -144,7 +144,7 @@ class BoincDb extends DbConn {
 
     static function escape_string($string) {
         $db = self::get();
-        return $db->base_escape_string(trim((string) $string));
+        return $db->base_escape_string(trim($string));
     }
     static function error() {
         $db = self::get();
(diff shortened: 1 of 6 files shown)


html/inc/boinc_db.inc
```diff
@@ -144,7 +144,7 @@ class BoincDb extends DbConn {
 
     static function escape_string($string) {
         $db = self::get();
-        return $db->base_escape_string(trim((string) $string));
+        return $db->base_escape_string(trim($string));
```
AenBleidd, line 147: @davidpanderson, are you sure this revert is correct?

davidpanderson:
See commit comment; the idea is to fix problems upstream rather than downstream.

AenBleidd:
Looks good to me.
@davidpanderson, feel free to merge it when ready

PR #4775, opened 2022-06-11
[Wasm] Compile wasm with html

talregev:
Compile wasm with html.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -231,7 +231,10 @@ build-*/
 # WASM
 *.wasm
 a.out.js
-boinc.js
 boinc_client.js
 boinccmd.js
 switcher.js
+boinc.html
+boinc_client.html
+boinccmd.html
+switcher.html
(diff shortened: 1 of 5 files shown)


AenBleidd:
Not sure we need 'boinc' to be built as html. JS is more convenient

talregev:
> Not sure we need 'boinc' to be built as html. JS is more convenient

You didn't try. The js is the same, but now it also build the html that show the output of boinc.
Please try and understand.
Create an html server with python on the folder of the artifacts after you extract them. 
` python3 -m http.server`

AenBleidd:
It looks good for now but later we might want to embed it in our own html page, that's why we would need a bare js file without any html

talregev:
> It looks good for now but later we might want to embed it in our own html page, that's why we would need a bare js file without any html

We create also the same bare JavaScript file `boinc_client.js` that doesn't depends on the html, and we can create our own html.

I can insert back the previous html as additional to this one.

AenBleidd:
No need for now. Keep it as is

talregev:
@AenBleidd Ready for review.

AenBleidd (approved):

Issue #4784, opened 2022-06-14
Linux: Manager crashes when switching between views


RichardHaselgrove:
**Describe the bug**
Launching the Manager in simple view, and switching to advanced view, works.
Launching in Advanced view, and making a double switch (from advanced to simple view,  and again from simple to advanced view) crashes the Manager. No warning, no dialog, no messages - it just disappears.

**Steps To Reproduce**
1. Open Manager
2. Switch views once or twice, depending where you start from

**Expected behavior**
No crash.

**System Information**
 - OS: Linux Mint 20.3 Una
 - BOINC Version: v7.20.0 pre-release from @LocutusOfBorg PPA

![image](https://user-images.githubusercontent.com/14886436/173558638-973787a1-c638-4749-bcda-6e33fcfea2e0.png)

**Additional context**
Found while testing for #4783. The wxWidgets debug alert doesn't appear in Linux, but this does happen.

AenBleidd:
@RichardHaselgrove, are you able to reproduce this on master with latest build artifacts?

RichardHaselgrove:
Um. It's not the same, but...

Tested with `linux_manager-with-webview_1bbc873314e287cc924fdac5e11f0617132fe668.zip` from #4778 from yesterday. The 'About' dialog shows

![image](https://user-images.githubusercontent.com/14886436/173562797-848abad9-b19d-469b-9dc2-28f3c60bd1db.png)

It doesn't crash, but simple view displays as

![image](https://user-images.githubusercontent.com/14886436/173562936-1196b9ca-3422-4116-9607-1a6ff9373503.png)

It should be

![image](https://user-images.githubusercontent.com/14886436/173563379-75ea2fc4-5698-4381-93c7-c51b07ccfa48.png)

or thereabouts.

Note that @LocutusOfBorg's PPA is an automatically updated test tool, intended for exactly this sort of work: but it is based on certain Debian tweaks to the original BOINC code. In particular, certain menu options in the Manager are removed: there are no separate options for 'minimise window' and 'exit application': 'Close window' (ctrl-W) in advanced view becomes a synonym for 'exit application'.

AenBleidd:
This is exactly the reason why I hate wxWidgets.
@RichardHaselgrove, I'll try to check it out and send you some executables for testing because I have no linux system to test it by myself.

Vulpine05:
[...]

PR #4785, opened 2022-06-14
[Manager] Create proper sizing on SimpleView

### LINKED ISSUE #4783: [Manager] wxWidgets debug alert when switching between advanced and simple view.
**Describe the bug**
wxWidgets has an error message when switching from advanced to simple view in the manager.  Or, the Manager will close unexpectedly (see additional context below).

**Steps To Reproduce**
1. Open BOINC Manager (already in advanced view).
2. Switch to simple view, either by mouse click in menu or with ctrl + shift + A
3. See screenshot below for the debug message.  Pressing stop will close the Manager.  Hitting continue will display another debug message (similar information).  You need to click few a few of them before simple view is displayed.

**Expected behavior**
A seamless transition from the two views.

**Screenshots**
image

**System Information**
 - OS: Windows 10, simliar for Ubuntu 18.04
 - BOINC Version: 7.20.0 Pre-Release, wxWidgets 3.1.6 (Windows), 3.0.4 (Ubuntu 18.04)

**Additional context**
For my Ubuntu system, it switches to simple view with no problems, but it will close the Manager unexpectedly when switching back to advance view.  No debug or other warning message displayed.

AenBleidd:
This fixes #4783.

### DIFF
--- a/clientgui/sg_BoincSimpleFrame.cpp
+++ b/clientgui/sg_BoincSimpleFrame.cpp
@@ -118,6 +118,9 @@ CSimpleFrame::CSimpleFrame(wxString title, wxIconBundle* icons, wxPoint position
     CreateMenus();
     dlgMsgsPtr = NULL;
     m_pBackgroundPanel = new CSimpleGUIPanel(this);
+    mainSizer = new wxBoxSizer(wxVERTICAL);
+    mainSizer->Add(m_pBackgroundPanel, 1, wxLEFT | wxRIGHT | wxEXPAND, 0);
+    SetSizerAndFit(mainSizer);
     RestoreState();
 }
 
@@ -1006,7 +1009,6 @@ CSimpleGUIPanel::CSimpleGUIPanel(wxWindow* parent) :
 
     Layout();
     SetSizerAndFit(mainSizer);
-    parent->SetSizerAndFit(mainSizer);
 
     SetBackgroundBitmap();
 
@@ -1033,8 +1035,6 @@ CSimpleGUIPanel::~CSimpleGUIPanel()
     checkForNewNoticesTimer->Stop();
 	delete checkForNewNoticesTimer;
     m_bmpBg = wxNullBitmap; // Deletes old bitmap via reference counting
-
-    GetParent()->SetSizer(NULL, false); // Avoid trying to delete mainSizer twice
 
     wxLogTrace(wxT("Function Start/End"), wxT("CSimpleGUIPanel::CSimpleGUIPanel - Destructor Function End"));
 }
--- a/clientgui/sg_BoincSimpleFrame.h
+++ b/clientgui/sg_BoincSimpleFrame.h
@@ -153,6 +153,7 @@ protected:
 
 private:
     CDlgMessages* dlgMsgsPtr;
+    wxBoxSizer* mainSizer;
 
     DECLARE_EVENT_TABLE()
 };


AenBleidd:
@CharlieFenton, please review and test on OSX.
This should be a proper fix of the #4783.
@RichardHaselgrove, please test this on Windows and linux, eventually this might fix #4784

RichardHaselgrove:
OK, Linux first.

With linux_manager-with-webview_0fb2cb615a5d17a81eaff8a3f49e18f565dbd576.zip (this PR's artifact), I get

image

That's not bad, and similar to the PPA. But it (like the PPA) is hiding the icon in the combo box in the Tasks section - but I can't get it to screenshot with the combo box showing the list. I'll try to show what I'm seeing with the Windows build (it's a full green circle - just a tiny arc is showing here).

CharlieFenton (approved):
It seems ot work correctly on MacOs 12.4 Monterey. I switched back and forth several times between Simple and Advanced Views and saw no problems. Thank you for the fix.

AenBleidd:
@RichardHaselgrove, thanks for the quick fix.
if you don't mind, I'll open a separate ticket for cutted icon to not block this PR.

AenBleidd:
> Since the Simple View has worked this way for years, I think the warning was incorrect. I will be curious whether the warning still occurs with wxWidgets 3.1.7. Please let me know

Actually, the warning looks correct: we were creating an elements on one window, assign them to the sizer on it, and then assign this sizer to the parent window. So this was already some kind of a hack to get proper sizing of parent Window.
And, actually, that fact that it was working for years, doesn't mean that it was working properly ;)
So, I'd say, the way I fixed it in this PR is more correct than it was before.

P.S. I haven't checked this on 3.1.7 yet, but this particular check was not changed between 3.1.6 and 3.1.7

AenBleidd:
> > With linux_manager-with-webview_0fb2cb615a5d17a81eaff8a3f49e18f565dbd576.zip (this PR's artifact), I get
> 
> More than that, the entire background image is missing. Here is what it should look like (from my Mac): Screen Shot 2022-06-14 at 2 26 00 PM

This because it was started not from the folder where BOINC installed but from some other, and that other folder doesn't contain skins folder. So this is like an expected behavior in this case.

RichardHaselgrove:
> This because it was started not from the folder where BOINC installed but from some other, and that other folder doesn't contain skins folder. So this is like an expected behavior in this case.

Windows doesn't start it from the same folder, either - but it has more robust mechanisms for finding out out where the data folder is located.

RichardHaselgrove:
Actually, I think that might be wrong. I've looked in 

* /var/lib/boinc-client
* /usr/bin
* /etc/boinc-client

and there's no sign of a skins folder in any of them. Are the skins even deployed on Linux?

AenBleidd:
@RichardHaselgrove, looks like it's not included in `boinc-manager` package.
May I ask you to report this issue to @LocutusOfBorg probably here https://launchpad.net/~costamagnagianfranco?

RichardHaselgrove:
Actually, I think @LocutusOfBorg just puts an automation layer on top of the Debian distro. We may have to go back to the top of the distribution chain to find the origin of this. What were the communications like between development and distribution when the skins were introduced, in this case presumably with simple view itself? I'd better dig into the history before I end up with egg on my face.

AenBleidd:
@RichardHaselgrove, try next:
- find the location of boincmgr executable
- put skins folder nearby of it
- start any shell (e.g. bash)
- navigate to the location of boincmgr
- run ./boincmgr

This should let it find skins correctly.

But this might be not a solution for *nix systems because the are not expect to have non-executable files in the location where all executables are held. Instead. this folder should be in /etc/boinc folder (or similar) but as far as I see, we're not looking at that location. And here I'm wondering, if this was ever working on linux?

So this might be a separate ticket for feature implementation and it's probably worths to open one and continue conversation there.
@RichardHaselgrove, if you agree - could you please open a separate ticket to discuss this issue further?

RichardHaselgrove:
Well, that was easy. Mint gives me a GUI file manager: I can navigate to anywhere, either as myself or as SuperUser. And I can open a terminal in either guise, and run shell commands as either user.

Binaries are in /usr/bin: that's where I've been dropping test versions all along. And it's also where I dropped the skins this morning, as SU. Running ./boincmgr as SU gets me the manager as usual, except that it doesn't pick up my personal layout settings - good test of who I am today. No skin backgrounds in simple view.

But I learned something new. The Linux distro wallahs have **removed** commands like 'shut down connected client' in Advanced view, but they have **not** disabled the exit dialog in simple view. Ergo, corporates don't use simple view.

There would be a simple answer to this: ask an IBM exec if their copy displays the WCG skin in simple view. Except, IBM aren't running WCG any more, and WCG are out for the count.

You're right - we should take this elsewhere. I'll ask around on the message boards, and come back to open an issue if I can't get a definitive answer.

CharlieFenton:
Looking at clientgui/SkinManager.cpp, I see the following code:
```
wxString CSkinManager::GetSkinsLocation() {
    // Construct path to skins directory
    wxString strSkinLocation = wxEmptyString;

#ifdef __WXMSW__
    strSkinLocation  = wxGetApp().GetRootDirectory();
    strSkinLocation += wxFileName::GetPathSeparator();
    strSkinLocation += wxT("skins");
#else
    strSkinLocation = wxString(wxGetCwd() + wxString(wxFileName::GetPathSeparator()) + wxT("skins"));
#endif

    return strSkinLocation;
}

```

AenBleidd:
@CharlieFenton, yep, and not sure it could ever work for linux

RichardHaselgrove:
That code dates back to 2007 (https://github.com/BOINC/boinc/commit/6b2c7f38a28f1656e2bf10ea608c8b43e3e16bfa)

If we haven't noticed until now that it doesn't work, this perhaps reverts to being a design decision: do we want or need a skinned simple view Manager in Linux? That's probably a question for the providers of skinned bespoke derivatives of the raw BOINC platform, but I suspect that the answer might be "won't fix".

In which case, we can forget it, and go back to sorting out the functional bugs.

AenBleidd:
@RichardHaselgrove, I'd keep it on backlog because it's always good to have at least default skin than just a grey emptiness

PR #4795, opened 2022-06-17
[wasm] Disable get mac address in wasm.

### LINKED ISSUE #4798: [wasm] Generate device or browser unique id
When BOINC client runs on Desktop, it generates a unique device id even if Data folder was removed between runs. This helps to avoid device duplicates on user profile page.
This behavior is achieved by reading device MAC address.

For WASM hardware identifiers could not be read.

There should be implemented some functionality to generate unique id for device (or at least browser) from JS code.

Ref: https://github.com/fingerprintjs/fingerprintjs

talregev:
Disable get mac address in wasm.

After the change we have some progress in client on the web:

```
17-Jun-2022 06:33:17 [---] cc_config.xml not found - using defaults
17-Jun-2022 06:33:17 [---] Starting BOINC client version 7.21.0 for i686-pc-linux-gnu
17-Jun-2022 06:33:17 [---] This a development version of BOINC and may not function properly
17-Jun-2022 06:33:17 [---] log flags: file_xfer, sched_ops, task
17-Jun-2022 06:33:17 [---] Libraries: libcurl/7.83.1-DEV OpenSSL/3.0.3 zlib/1.2.12
17-Jun-2022 06:33:17 [---] Data directory: /
17-Jun-2022 06:33:17 [---] No usable GPUs found
17-Jun-2022 06:33:17 [---] Creating new client state file
17-Jun-2022 06:33:17 [---] Host name: emscripten
17-Jun-2022 06:33:17 [---] Processor: 1 WASM WASM
17-Jun-2022 06:33:17 [---] Processor features: 
17-Jun-2022 06:33:17 [---] OS: Emscripten: 3.1.13
17-Jun-2022 06:33:17 [---] Memory: 16.00 MB physical, 0 bytes virtual
17-Jun-2022 06:33:17 [---] Disk: 3.81 GB total, 1.91 GB free
17-Jun-2022 06:33:17 [---] Local time is UTC +3 hours
17-Jun-2022 06:33:17 [---] No general preferences found - using defaults
17-Jun-2022 06:33:17 [---] Preferences:
17-Jun-2022 06:33:17 [---]    max memory usage when active: 8.00 MB
17-Jun-2022 06:33:17 [---]    max memory usage when idle: 14.40 MB
17-Jun-2022 06:33:17 [---]    max disk usage: 1.81 GB
17-Jun-2022 06:33:17 [---]    don't use GPU while active
17-Jun-2022 06:33:17 [---]    suspend work if non-BOINC CPU load exceeds 25%
17-Jun-2022 06:33:17 [---]    (to change preferences, visit a project web site or select Preferences in the Manager)
17-Jun-2022 06:33:17 [---] Setting up project and slot directories
17-Jun-2022 06:33:17 [---] Checking active tasks
17-Jun-2022 06:33:17 [---] Setting up GUI RPC socket
17-Jun-2022 06:33:46 [---] GUI RPC listen failed: -1
```

image

Implement unique user device id in js
fixed #4798

### DIFF
--- a/client/hostinfo_network.cpp
+++ b/client/hostinfo_network.cpp
@@ -57,6 +57,27 @@
 
 #include "hostinfo.h"
 
+#if WASM
+    #include <emscripten.h>
+#endif
+
+#if WASM
+    // unique user device in js.
+    EM_JS(int, get_uuid, (char* buf), {
+        const fpPromise = import('https://openfpcdn.io/fingerprintjs/v3').then(FingerprintJS => FingerprintJS.load());
+
+        // Get the visitor identifier when you need it.
+        fpPromise.then(fp => fp.get()).then(result => {
+                // This is the visitor identifier:
+                const visitorId = result.visitorId;
+                for (let i = 0; i < visitorId.length; i++) {
+                    buf[i] = visitorId[i];
+                }
+        });
+        return 0;
+    });
+#endif
+
 // get domain name and IP address of this host
 // Android: if domain_name is empty, set it to android_xxxxxxxx
 //
@@ -145,11 +166,16 @@ void HOST_INFO::generate_host_cpid() {
     char buf[256+MAXPATHLEN];
     char dir[MAXPATHLEN];
 
+#if WASM
+    // unique user device in js.
+    retval = get_uuid(buf);
+#else
     // if a MAC address is available, compute an ID based on it;
     // this has the advantage of stability
     // (a given host will get the same ID each time BOINC is reinstalled)
     //
     retval = get_mac_address(buf);
+#endif
     if (retval) {
         make_random_string("", host_cpid);
         return;


talregev:
@AenBleidd Ready for review.

AenBleidd (approved):

talregev:
> Ok, so the results of wasm testing are next:
- Chrome and Edge on the same system give different IDs - PASSED
- ID is the same after browser restart - PASSED
- ID is the same after browser restart with cookies and cache clean-up - PASSED
- ID is the same on two VMs running in parallel (same configuration, different instances) - FAILED

> Not sure we can do anything with the last test case, so for now it's fine.

@AenBleidd

CharlieFenton:
Did you confirm that you get different IDs on two different but identical computers running the same browser?

AenBleidd:
> Did you confirm that you get different IDs on two different but identical computers running the same browser?

@CharlieFenton, thank you for your question.

Unfortunately, if I create a VM and run two instances in parallel (the the only difference is IP/MAC address), these two instances have the same ID.

For now this is fine. If this will be an issue in future - it could be revisited.
Currently we're just exploring is it really possible/reasonable to run BOINC in browser.

CharlieFenton:
`if I create a VM and run two instances in parallel (the the only difference is IP/MAC address), these two instances have the same ID.

For now this is fine.`

@AenBleidd I think that is a different test than comparing the ID generated by two different but identical physical computers. After all, the purpose of the ID is to uniquely and reproducibly identify one particular computer so that all results form that computer can be grouped together, without confusing them with results from a different computer.

AenBleidd:
> I think that is a different test than comparing the ID generated by two different but identical physical computers.

I completely agree with you.
Unfortunately I have no identical computers so I'm unable to perform such a test.

PR #4800, opened 2022-06-18
client: if project has zero resource share AND work buf min is zero, don't fetch from it unless device instance is actually idle.

### LINKED ISSUE #4396: client: fix #2837 removed a special use case where idle was desired before work fetch
**Describe the bug**
PrimeGrid and other projects where being "first" with a result is important used zero resource share and zero cache to avoid ANY prefetch of tasks before the current one was completed. fix #2837 changed the test from "idle instance" to "three minutes work remaining". This is a good change for MOST people but broke the special-use case for some.

**Steps To Reproduce**
1. Set resource share on your only project to zero
2. Set cache to 0 days and 'extra' to 0 days

**Expected behavior**
A single task should download and execute. When it completes - and not before! - another task should be downloaded.

**Additional context**
Suggested fix: in work_fetch.cpp line 115 replace

`if (p->resource_share == 0 && rwf.saturated_time > WF_EST_FETCH_TIME) {
     return RSC_REASON_ZERO_SHARE;`

with

`if (p->resource_share == 0 && rwf.saturated_time > WF_EST_FETCH_TIME && gstate.work_buf_min() > WF_EST_FETCH_TIME) {
             return RSC_REASON_ZERO_SHARE;`

This would in addition to zero resource share, check for the zero cache setting. I have NOT been able to test this fix (although it compiles clean), and am not familiar with the internals (are both values in seconds?), so someone much more knowledgeable than I should verify that this will work and not break a hundred other things.

davidpanderson:
Apparently this is desired by users who compete to return results first.

Fixes #4396

### DIFF
--- a/client/work_fetch.cpp
+++ b/client/work_fetch.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2014 University of California
+// Copyright (C) 2022 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -36,6 +36,7 @@
 #include "work_fetch.h"
 
 using std::vector;
+using std::min;
 
 RSC_WORK_FETCH rsc_work_fetch[MAX_RSC];
 WORK_FETCH work_fetch;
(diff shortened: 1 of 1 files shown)


AenBleidd:
This is gonna be tested by @pg-tern

AenBleidd (approved):

PR #4808, opened 2022-06-23
[Windows][Manager] Do proper verifications when saving state on application exit

### LINKED ISSUE #4793: [Manager] Memory access error on closedown
**Describe the bug**
I've seen this popup a couple of times since installing the test version of v7.20.0

image

**Steps To Reproduce**
I think it occurs when the Manager (running mimised) is force-closed by the operating system in preparation for a reboot. Further testing required.

**System Information**
 - OS: Windows 7/64 professional
 - BOINC Version: 7.20.0

AenBleidd:
Sometimes Advanced Frame is destroyed much more later, when e.g. MainDocument is already destroyed.

This fixes #4793.

### DIFF
--- a/clientgui/ViewWork.cpp
+++ b/clientgui/ViewWork.cpp
@@ -616,7 +616,7 @@ void CViewWork::OnShowItemProperties( wxCommandEvent& WXUNUSED(event) ) {
 
 bool CViewWork::OnSaveState(wxConfigBase* pConfig) {
     bool bReturnValue = true;
-    CMainDocument* pDoc     = wxGetApp().GetDocument();
+    CMainDocument* pDoc = wxGetApp().GetDocument();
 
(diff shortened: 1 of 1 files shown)


AenBleidd:
@CharlieFenton, could you please review this?
In understand, that this is not reproducible on any OS except Windows because only on Windows we have special procedure to shutdown on system log out, but the piece of code I fixed is a common one, so it would be nice to have it reviewed.
Briefly, additionally to Debug asserts I'm verifying now all necessary variables in Release mode also, and in this case to not have some values saved on forced shutdown of BOINC is much more better for users than a message about application crash.

CharlieFenton (approved):
Code looks good to me, and seems to work correctly on Mac. The state was saved correctly both on manual exit from BOINC and on forced shutdown.

PR #4828, opened 2022-07-06
client: if bandwidth limit is removed (up or down), remove limit from active transfers.

### LINKED ISSUE #1071: Removing an upload/download rate preference does not return speeds to maximum for in-progress uploads/downloads
**Reported by JacobKlein on 29 Jul 41291932 02:13 UTC**
If I change a download rate preference to something like 5KB/sec, I can see the speed drop down to match it.

However, if I then clear the preference (or type 0.00 in it), BOINC appears to keep the speed at 5KB/sec unless I restart BOINC.

If I change the preference to something like 9999 before restarting BOINC, I can see the speeds resume.

Essentially, the mechanism that sets BOINC to max speed is not being restored when a speed limit preference is removed.

Migrated-From: http://boinc.berkeley.edu/trac/ticket/1095

davidpanderson:
Logic cleanup in this area.

Fixes #1071

### DIFF
--- a/client/file_xfer.cpp
+++ b/client/file_xfer.cpp
@@ -215,7 +215,7 @@ FILE_XFER_SET::FILE_XFER_SET(HTTP_OP_SET* p) {
 int FILE_XFER_SET::insert(FILE_XFER* fxp) {
     http_ops->insert(fxp);
     file_xfers.push_back(fxp);
-    set_bandwidth_limits(fxp->is_upload);
+    enforce_bandwidth_limits(fxp->is_upload);
     return 0;
 }
 
@@ -230,7 +230,7 @@ int FILE_XFER_SET::remove(FILE_XFER* fxp) {
     while (iter != file_xfers.end()) {
         if (*iter == fxp) {
             iter = file_xfers.erase(iter);
-            set_bandwidth_limits(fxp->is_upload);
+            enforce_bandwidth_limits(fxp->is_upload);
             return 0;
         }
         ++iter;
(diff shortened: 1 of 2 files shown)


AenBleidd:
@Vulpine05, could you please test it?

Vulpine05:
@AenBleidd , yes, I can test.  I glanced through the code and it looks good, but I'll test it out to confirm later this week.

Vulpine05:
Good news and bad news.  The good news is that the PR appears to fix the bug.  If I have a very low transfer rate files move very slowly.  When I clear the transfer rate, they move really quickly.  I also tried this same procedure with 7.20.1, and when I cleared the transfer limits nothing different happened.  I tried capturing this via debug logs, but I didn't get good information out of my control (7.20.1).  I'll have to run this again tomorrow when I have more files to transfer.

However, as I was testing this PR, I had a debug alert pop up:

Screencap-debug-4828

I don't know if this is related to this PR or not, but several files had transferred already when it occured.  I clicked on retry and the Manager closed out.

I'm not sure where to go from here, but let me know if you need me to test something else.

davidpanderson:
The crash was in the Manager.  I don't think it's related to this PR.

BTW, I'm confused by the crash: the Manager doesn't use any deques, as far as I can tell.

davidpanderson:
... that is, the BOINC Manager code doesn't use deques.  WxWidgets probably does.

AenBleidd:
I'll check this debug asserts message

AenBleidd (approved):

PR #4829, opened 2022-07-07
client: fix bug in work fetch logic for case of zero resource share and zero work buf.

### LINKED ISSUE #4396: client: fix #2837 removed a special use case where idle was desired before work fetch

davidpanderson:
Fixes #4396

### DIFF
--- a/client/file_xfer.cpp
+++ b/client/file_xfer.cpp
@@ -215,7 +215,7 @@ FILE_XFER_SET::FILE_XFER_SET(HTTP_OP_SET* p) {
 int FILE_XFER_SET::insert(FILE_XFER* fxp) {
     http_ops->insert(fxp);
     file_xfers.push_back(fxp);
-    set_bandwidth_limits(fxp->is_upload);
+    enforce_bandwidth_limits(fxp->is_upload);
     return 0;
 }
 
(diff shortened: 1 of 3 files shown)


AenBleidd:
This PR contains two different fixes that do not relate to each other.
@davidpanderson, I'll close this PR and create a new one with relevant commit only if you don't mind

AenBleidd:
Closed in favor of #4830 that doesn't contain unrelated commits

PR #4836, opened 2022-07-11
[CI] Enable OSX screensaver build

AenBleidd:

### DIFF
--- a/deploy/prepare_deployment.py
+++ b/deploy/prepare_deployment.py
@@ -214,6 +214,7 @@ logs_list = [
     '3rdParty/mingw/vcpkg/buildtrees/*.log',
     '3rdParty/Windows/vcpkg/buildtrees/*.log',
     'android/BOINC/app/build/reports/',
+    'mac_build/*.log',
 ]
 
 def prepare_7z_archive(archive_name, target_directory, files_list):
--- a/deploy/prepare_deployment.py
+++ b/deploy/prepare_deployment.py
@@ -214,7 +214,6 @@ logs_list = [
     '3rdParty/mingw/vcpkg/buildtrees/*.log',
     '3rdParty/Windows/vcpkg/buildtrees/*.log',
     'android/BOINC/app/build/reports/',
-    'mac_build/*.log',
 ]
 
 def prepare_7z_archive(archive_name, target_directory, files_list):
--- a/mac_build/BuildMacBOINC.sh
+++ b/mac_build/BuildMacBOINC.sh
@@ -172,7 +172,8 @@ done
 
 ## For unknown reasons, this xcodebuild call generates syntax errors under zsh
 ## unless we enclose the command in quotes and invoke it with eval.
-eval "xcodebuild -project boinc.xcodeproj ${targets} -configuration ${style} -sdk \"${SDKPATH}\" ${doclean} build ${uselibcplusplus} ${theSettings};result=$?"
+eval "xcodebuild -project boinc.xcodeproj ${targets} -configuration ${style} -sdk \"${SDKPATH}\" ${doclean} build ${uselibcplusplus} ${theSettings}"
+result=$?
 
 if [ $result -eq 0 ]; then
     # build ibboinc_zip.a for -all or -lib or default, where
@@ -184,7 +185,5 @@ if [ $result -eq 0 ]; then
         fi
     fi
 fi
-
-echo "Result is: ${result}"
 
 return $result
--- a/mac_build/BuildMacBOINC.sh
+++ b/mac_build/BuildMacBOINC.sh
@@ -171,7 +171,7 @@ for f in "${settings[@]}"; do
 done
 
 ## For unknown reasons, this xcodebuild call generates syntax errors under zsh
-## unless we enclose the command in quotes and invoke it with eval. 
+## unless we enclose the command in quotes and invoke it with eval.
 eval "xcodebuild -project boinc.xcodeproj ${targets} -configuration ${style} -sdk \"${SDKPATH}\" ${doclean} build ${uselibcplusplus} ${theSettings};result=$?"
 
 if [ $result -eq 0 ]; then
@@ -184,5 +184,7 @@ if [ $result -eq 0 ]; then
         fi
     fi
 fi
+
+echo "Result is: ${result}"
 
 return $result
--- a/mac_build/BuildMacBOINC.sh
+++ b/mac_build/BuildMacBOINC.sh
@@ -173,7 +173,7 @@ done
 
 ## For unknown reasons, this xcodebuild call generates syntax errors under zsh
 ## unless we enclose the command in quotes and invoke it with eval.
-eval "xcodebuild -project boinc.xcodeproj ${targets} -configuration ${style} -sdk \"${SDKPATH}\" ${doclean} build ${uselibcplusplus} ${theSettings} OTHER_CFLAGS=\"-Xclang -v\""
+eval "xcodebuild -verbose -project boinc.xcodeproj ${targets} -configuration ${style} -sdk \"${SDKPATH}\" ${doclean} build ${uselibcplusplus} ${theSettings}"
 result=$?
 
 if [ $result -eq 0 ]; then
--- a/mac_build/BuildMacBOINC.sh
+++ b/mac_build/BuildMacBOINC.sh
@@ -173,7 +173,7 @@ done
 
 ## For unknown reasons, this xcodebuild call generates syntax errors under zsh
 ## unless we enclose the command in quotes and invoke it with eval.
-eval "xcodebuild -project boinc.xcodeproj ${targets} -configuration ${style} -sdk \"${SDKPATH}\" ${doclean} build ${uselibcplusplus} ${theSettings}"
+eval "xcodebuild -project boinc.xcodeproj ${targets} -configuration ${style} -sdk \"${SDKPATH}\" ${doclean} build ${uselibcplusplus} ${theSettings} OTHER_CFLAGS=\"-Xclang -v\""
 result=$?
 
 if [ $result -eq 0 ]; then
--- a/mac_build/buildFTGL.sh
+++ b/mac_build/buildFTGL.sh
@@ -61,7 +61,7 @@ while [[ $# -gt 0 ]]; do
         -prefix|--prefix)
         lprefix="$2"
         libPath="${lprefix}/lib"
-#        libftpath="${lprefix}"
+        libftpath="${lprefix}"
         shift
         ;;
         -q|--quiet)
--- a/mac_build/buildFTGL.sh
+++ b/mac_build/buildFTGL.sh
@@ -61,7 +61,7 @@ while [[ $# -gt 0 ]]; do
         -prefix|--prefix)
         lprefix="$2"
         libPath="${lprefix}/lib"
-        libftpath="${lprefix}"
+#        libftpath="${lprefix}"
         shift
         ;;
         -q|--quiet)
@@ -104,12 +104,12 @@ if [ "${doclean}" != "yes" ]; then
             lipo "${libPath}/libftgl.a" -verify_arch x86_64
             if [ $? -ne 0 ]; then alreadyBuilt=0; doclean="yes"; fi
         fi
-        
+
         if [ $alreadyBuilt -eq 1 ] && [ $GCC_can_build_arm64 = "yes" ]; then
             lipo "${libPath}/libftgl.a" -verify_arch arm64
             if [ $? -ne 0 ]; then alreadyBuilt=0; doclean="yes"; fi
         fi
-        
+
         if [ $alreadyBuilt -eq 1 ]; then
             cwd=$(pwd)
             dirname=${cwd##*/}
--- a/mac_build/buildMacBOINC-CI.sh
+++ b/mac_build/buildMacBOINC-CI.sh
@@ -181,7 +181,7 @@ fi
 echo "Building ${target}...done"
 
 # screensaver disabled because Travis can't build some library correctly, see https://github.com/BOINC/boinc/issues/2662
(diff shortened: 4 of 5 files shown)


talregev:
Please add a screenshot.

AenBleidd:
> Please add a screenshot.

Pardon, what screenshot? xD

talregev:
> > Please add a screenshot.
> 
> Pardon, what screenshot? xD

If you had a mac, you can take a photo with your cell phone of the screen server.

AenBleidd:
> > > Please add a screenshot.
> > 
> > 
> > Pardon, what screenshot? xD
> 
> If you had a mac, you can take a photo with your cell phone of the screen server. :P

I have no Mac. And in any case screenshot is not required, because this is just a try to fix our CI

AenBleidd:
@talregev, https://github.com/BOINC/boinc/issues/2662

talregev:
You can install libpng with brew?

CharlieFenton:
If you look at mac_build/buildfreetype.sh line 173 you will see that it feels freestyle not to use libpng:
`./configure --enable-shared=NO --prefix=${lprefix} --enable-freetype-config --without-png --host=x86_64`

CharlieFenton:
Also line 195:
`    ./configure --enable-shared=NO --prefix=${lprefix} --enable-freetype-config --without-png --host=arm
`

CharlieFenton:
I may have found the problem with the CI build. Building FreeType with mac_build/buildfreetype.sh creates a subdirectory freetype_install, in the same directory as freetype-2.11.0. When I run mac_build/buildFTGL.sh  on my Mac as part of building BOINC for distribution, I do not pass the` --prefix` argument to mac_build/buildFTGL.sh, so `libftpath="`pwd`/../freetype_install/"`. But building for CI on GitHub calls buildFTGL.sh from 3rdParty/buildMacDependencies.sh, which passes a value for ` --prefix`. This means that the path specified in freetype_install/lib/pkgconfig/freetype2.pc is ignored.

This makes me wonder if perhaps FTGL is using a different build of freetype then the one we created under CI, one that requires libpng. Notice this comment on line 74 of mac_build/buildFTGL.sh, and the following line 75:
```
# needed for ftgl to find our freetype build not the system one
export PKG_CONFIG_PATH=${libftpath}/lib/pkgconfig:${PKG_CONFIG_PATH}
```

All that may be needed is to remove line 64 of mac_build/buildFTGL.sh
`        libftpath="${lprefix}"
`
while leaving line 63:
`libPath="${lprefix}/lib"
`
In other words, `libftpath` should not be set to the the same value as `libpath`.

AenBleidd:
@CharlieFenton, thanks for the hint, I'll try it. Could you please meanwhile  review #4837?

CharlieFenton:
@AenBleidd You may need to force a rebuild of the dependent libraries in order to have it build libftgl correctly. I think it may rebuild the dependent libraries automatically if you create a new PR or a new branch, though I'm not sure. 

The way I have forced a library rebuild in the past is to temporarily add the line
`doclean="yes"`
before line 148 of 3rdParty/buildMacDependencies.sh.
After having rebuilt the cache, remember to revert that change.

AenBleidd:
I believe it does a rebuild already because now every build  takes more than 1 hour, but I'll check the logs, and if some libraries were not rebuilt, I'll definitely do a temporary set of `doclean="yes"` to the build scripts.
Thanks, @CharlieFenton

CharlieFenton:
@AenBleidd Yes, I see it is rebuilding the dependency cache now. If it is rebuilding the cache every time, that is wrong and should be fixed.

AenBleidd:
@CharlieFenton, no it rebuilds it as requested.
As you can see here
https://github.com/BOINC/boinc/blob/4e02993c7ad94fc1c35ed04e1ab4b4acf62e4dc3/.github/workflows/osx.yml#L38
we invalidate the cache every time any of the files mentioned in the key is changed.
So since I'm actively changing these files, cache is always invalidated and rebuilt.
Once merged to master, there will be no cache misses anymore

CharlieFenton:
@AenBleidd I see this in the CI build output:
> Run edgarrc/action-7z@v1.0.4
Error: Container action is only supported on Linux

I think this is what @talregev fixed in PR #4838. You may also need to implement the change from PR #4838.

AenBleidd:
@CharlieFenton, yeah, I forgot to merge from master. Thank you for the reminder. In any case, this is definitely not the final version of this PR, so I'll do that in the next commit.

talregev:
No need to reimplemented. Just rebase.

CharlieFenton:
@AenBleidd I think we may be almost there. In CI run #2276, I see the following:

line 6762:
`Checking for freetype-config... /Users/runner/work/boinc/boinc/3rdParty/buildCache/mac/bin/freetype-config`

In CI run #2288 it is perhaps closer:
lines 6758 - 6759:
```
checking for freetype-config... /Users/runner/work/boinc/boinc/3rdParty/mac/ftgl-2.1.3~rc5/../freetype_install//bin/freetype-config
./configure: line 20879: /Users/runner/work/boinc/boinc/3rdParty/mac/ftgl-2.1.3~rc5/../freetype_install//bin/freetype-config: No such file or directory
```
Note the double forward slash between `freetype_install//bin`

AenBleidd:
@CharlieFenton,
But search in `/Users/runner/work/boinc/boinc/3rdParty/buildCache/mac/bin/freetype-config` is successful, indicating that it takes 
the build from the correct place...

AenBleidd:
@CharlieFenton,

> I'll try fixing that locally and report back.

Sure, thanks

CharlieFenton:
@AenBleidd I removed the trailing slash in line 54 of mac_build/buildFTGL.sh thus:
`libftpath="`pwd`/../freetype_install"
`
and everything builds correctly. (I suspect that my mac does not have a preinstalled libfreetype so it used the one I built even without the correct path to freetype-config.)

Could you try that change and see if it helps on CI?

AenBleidd:
@CharlieFenton, I mean, before my change it was looking for a config in that directory and didn't complain, so I suppose it found the library correctly.

Speaking about CI, you can access directories only via build scripts, no she'll access in interactive mode AFAIK

AenBleidd:
@CharlieFenton 

> Could you try that change and see if it helps on CI?

Sure

CharlieFenton:
@AenBleidd 
> I mean, before my change it was looking for a config in that directory and didn't complain, so I suppose it found the library correctly.

Yes, I wonder about that too. But when I ran the script locally on my Mac, it did not complain about this path with the double slash, which I would expect to be invalid:
`checking for freetype-config... /Volumes/Dev/BOINC_GIT/ftgl-2.1.3~rc5/../freetype_install//bin/freetype-config
`

Now this is surprising to me: the `ls` command accepts the path with both the single slash and the double slash:
```
charliefenton@M1-iMac ~ % ls -al /Volumes/Dev/BOINC_GIT/ftgl-2.1.3~rc5/../freetype_install//bin/freetype-config
-rwxr-xr-x@ 1 charliefenton  staff  5415 Feb  1 23:23 /Volumes/Dev/BOINC_GIT/ftgl-2.1.3~rc5/../freetype_install//bin/freetype-config
charliefenton@M1-iMac ~ % ls -al /Volumes/Dev/BOINC_GIT/ftgl-2.1.3~rc5/../freetype_install/bin/freetype-config 
-rwxr-xr-x@ 1 charliefenton  staff  5415 Feb  1 23:23 /Volumes/Dev/BOINC_GIT/ftgl-2.1.3~rc5/../freetype_install/bin/freetype-config

```

AenBleidd:
@CharlieFenton, ok, I believe original path is correct because when building FreeType on CI, we also specify this path:
https://github.com/BOINC/boinc/blob/4e02993c7ad94fc1c35ed04e1ab4b4acf62e4dc3/mac_build/buildfreetype.sh#L82-L83
So, removing `libftpath` will definitely point to the wrong location.
Maybe, there is another reason of this fail?
Is it possible to generate more verbose output when building with XCode?
Because currently I even don't see what went wrong with linkage, and what is exactly missing...

CharlieFenton:
Well, I learned something today:
https://unix.stackexchange.com/questions/11964/what-do-double-slashes-mean-in-unix-path-is-cd-dir-subdir-valid

CharlieFenton:
@AenBleidd libftpath is different from lib path. But I admit that I don't know what the code in freetype make files does. So I suspect you may be correct.

However, looking back at the errors shown [here](https://github.com/BOINC/boinc/issues/2662#issue-354099453), all the PNG related calls were due to calls from libfreetype.a, not libFTGL.a. And lines 6212 and 6399 of the lates CI build say:
` ./configure  '--enable-shared=NO' '--prefix=/Users/runner/work/boinc/boinc/3rdParty/buildCache/mac' '--enable-freetype-config' '--without-png' '--host=x86_64'` and lines 6359 and 6546 say:
`libpng:        no`

This makes me still suspect that the CI build may be linking with a different copy of the freetype library. And I just noticed that lines 184 to 190 of BuildMacBOINC-CI.sh tell it to search for libraries in ./build/Deployment or ./build/Development before looking in the dependencies cache directory. None of the other build targets do that.

I suggest you try changing line 184 from this:
`libSearchPath=""`
to this:
libSearchPath="./build/Deployment"

CharlieFenton:
@AenBleidd Correction: I wrote that backwards. I meant that I suggest you try changing line 184 from this:
`libSearchPath="./build/Deployment"`
to this:
`libSearchPath=""`

CharlieFenton:
> Is it possible to generate more verbose output when building with XCode?

The error message on line 1074 of the Make section of CI run 2292 says:
> clang: error: linker command failed with exit code 1 (use -v to see invocation)

but since the linker is invoked internally by xcodebuild, I don't know of any way to do that. When building using the Xcode GUI rather than the command line xcodebuild, it automatically shows you the details. But again, we can't do that under CI.

CharlieFenton:
@AenBleidd I am leaving for tonight. Let me know whether changing line 184 of BuildMacBOINC-CI.sh helped.

CharlieFenton:
@AenBleidd
> Is it possible to generate more verbose output when building with XCode?

You could try temporarily adding `-verbose` on line 176 of BuildMacBOINC.sh:
`eval "xcodebuild -verbose -project boinc.xcodeproj ${targets} -configuration ${style} -sdk \"${SDKPATH}\" ${doclean} build ${uselibcplusplus} ${theSettings}"`

 I don't know if this will give us a list of undefined linker symbols, but it might.

talregev:
For this PR you should delete all the other flow, and when you successful,just bring them back. it will save you time.

AenBleidd:
@talregev, not really: OSX build I'm working with in any case takes more than 1 hour to build. Removing all other flows will not save me any time)

talregev:
Switch off no. Delete them next push force, yes.
As you wish. It will save you about 10 min each update.

CharlieFenton:
@AenBleidd the Xcode project for ss_app has the following setting for OTHER_LDFLAGS: 
`-lresolv -ljpeg -lfreetype -lftgl -lz -lbz2`

So instead of `-setting LDFLAGS "-v"` you might want to try `-setting OTHER_LDFLAGS "-v -lresolv -ljpeg -lfreetype -lftgl -lz -lbz2"`. I don't know if that will help, but worth a try.

AenBleidd:
@CharlieFenton, I think I found the issue: could you please check when building freetype library locally, do you have `brotli` support enabled? In case of CI it's found by PKG_CONFIG but libraries are not passed. If in your case freetype is build **with** brotli support. then this means that for some reason brotli libraries are not passed correctly

AenBleidd:
@CharlieFenton, just for reference, from the CI when building freetype:
```
Library configuration:
  external zlib: yes (pkg-config)
  bzip2:         yes (autoconf test)
configure: WARNING:
  `make refdoc' will fail since pip package `docwriter' is not installed.
  To install, run `python3 -m pip install docwriter', or to use a Python
  virtual environment, run `make refdoc-venv' (requires pip package
  `virtualenv').  These operations require Python >= 3.5.
  libpng:        no
  harfbuzz:      no
  brotli:        yes (pkg-config)
  pthread:       yes
```

CharlieFenton:
@AenBleidd When I buildFreeType locally, I see:
`checking for BROTLI... no`
and later:
```
Library configuration:
  external zlib: yes (autoconf test)
  bzip2:         yes (autoconf test)
  libpng:        no
  harfbuzz:      no
  brotli:        no
  pthread:       yes
```

Apparently it does not build with brotli on my Mac because brotli is not installed. I tested adding `--without-brotli` to the two calls of `./configure` in the buildfreetype.sh script and it seems to be fine on my Mac. So go ahead and make that change. Remember that you must do it in two places, as there is a `./configure` call for x86_64 and another for arm64.

I notice you only did this for the arm64 invocation of `./configure`, so this would not fix the issue for the x86_64 build, which is done first.

As before, please remember to update the copyright date and to add add a description of your change to the edit history after line 36.

CharlieFenton:
@AenBleidd I also suggest creating a new branch for the change to buildfreetype.sh, to avoid unnecessary changes to mac_build/BuildMacBOINC.sh and mac_build/BuildFTGL.sh. You probably also want to undo some of your experimental changes to mac_build/BuildMacBOINC-CI.sh (such as restoring the beautifier call.)

AenBleidd:
@CharlieFenton,

> I notice you only did this for the arm64 invocation of ./configure, so this would not fix the issue for the x86_64 build, which is done first.

3am not always a good time to make a proper commits xD
Thanks for the hint.

> As before, please remember to update the copyright date and to add add a description of your change to the edit history after line 36.

For sure, I'll definitely do this when PR is ready.

> I also suggest creating a new branch

Definitely, I'll do this like a did with the previous PR when moving retval from eval string.

This is mostly a test PR, because there is no other way for me to test OSX builds, so this PR won't be merged in it's current state.

PR #4837, opened 2022-07-11
[OSX] Move get command exit code from eval string.

### LINKED ISSUE #4612: OSX CI build failed but shown as 'passed'

AenBleidd:
This fixes #4612.

### DIFF
--- a/mac_build/BuildMacBOINC.sh
+++ b/mac_build/BuildMacBOINC.sh
@@ -2,7 +2,7 @@
 
 # This file is part of BOINC.
 # http://boinc.berkeley.edu
-# Copyright (C) 2017 University of California
+# Copyright (C) 2022 University of California
 #
 # BOINC is free software; you can redistribute it and/or modify it
 # under the terms of the GNU Lesser General Public License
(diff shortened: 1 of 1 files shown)


CharlieFenton:
@AenBleidd Why are you changing this? This script is not  used by CI; CI uses buildMacBOINC-CI.sh, so this file is not the reason the CI builds are not showing build failures.

CharlieFenton:
My mistake; I now see that buildMacBOINC-CI.sh calls buildMacBOINC.sh. I will test it.

AenBleidd:
@CharlieFenton, this script is called from the `buildMacBOINC-CI.sh` script.
And I checked, that this doesn't work on CI because we do 
```
source BuildMacBOINC.sh ${config} -noclean -target ${target} | tee xcodebuild_${target}.log | $beautifier; retval=${PIPESTATUS[0]}
```
And when in this script there is `eval` with `result=$?` **inside** this `eval`, `result` value is never reported as non-zero.

AenBleidd:
> My mistake; I now see that buildMacBOINC-CI.sh calls buildMacBOINC.sh. I will test it.

No problem :)

CharlieFenton (approved):
Works correctly on my Mac under zsh. Please remember to update the copyright year in line 5. Also, I would appreciate it if you would add a description of your change to the edit history after line 37.

PR #4841, opened 2022-07-12
[CI][osx] Add mac osx logs files

talregev:
Add mac osx logs files.

### DIFF
--- a/deploy/prepare_deployment.py
+++ b/deploy/prepare_deployment.py
@@ -214,6 +214,7 @@ logs_list = [
     '3rdParty/mingw/vcpkg/buildtrees/*.log',
     '3rdParty/Windows/vcpkg/buildtrees/*.log',
     'android/BOINC/app/build/reports/',
+    'mac_build/xcodebuild_*.log',
 ]
 
 def prepare_7z_archive(archive_name, target_directory, files_list):


AenBleidd:
This is not needed. All these logs are already appear as a build log and could be read directly from the CI flow, no need to duplicate them and upload

talregev:
> This is not needed. All these logs are already appear as a build log and could be read directly from the CI flow, no need to duplicate them and upload

I am not duplicated. this logs already created. Also so I upload them as debug process as we can downloads files and search for it, or compare other run or other debug ideas.

AenBleidd:
Again: these logs are the same that you can see in the output log, no additional info there

talregev:
> Again: these logs are the same that you can see in the output log, no additional info there

I am not talking on additional info, as also it can be hint to add other files. I am talking other tool for debug. Not all the devs debug like you ...

AenBleidd:
@talregev, then why we don't upload everything that is printed to the output to the logs when flow fails? Maybe because it's not needed? Here's the same. Why should you download any file if you can directly see the same logs in your browser?

talregev:
> @talregev, then why we don't upload everything that is printed to the output to the logs when flow fails? Maybe because it's not needed? Here's the same. Why should you download any file if you can directly see the same logs in your browser?

I give you some example above. In the browser for example, you cannot compare 2 logs files in different runs. 
And we can think other stuff. I don't tell you to upload everything. Just key point logs files for mac.

AenBleidd:
> I give you some example above. In the browser for example, you cannot compare 2 logs files in different runs.

Nobody stops you from opening two tabs in browser with different runs

AenBleidd:
Again: we never upload the same that is already printed on screen, only some additional info that is not shown during the build.
Why OSX flow should be different?

talregev:
> Nobody stops you from opening two tabs in browser with different

Let's see you compare 2 tabs in a long texts, instead to compare 2 files with diff. Are you really want to go of that direction?

AenBleidd:
> Are you really want to go to that direction?

We already do this for **every other flow**. Why should OSX be different?

talregev:
> > Are you really want to go to that direction?
> 
> We already do this for **every other flow**. Why should OSX be different?

I never compare 2 tabs in the browser.

AenBleidd:
We also never zip output of any other flow.

talregev:
> We also never zip output of any other flow.

We actually do this for Linux. That how I solve the bug with the share compilation. I compare 2 files before and after my change in opencl.

AenBleidd:
What exact files are you talking about?
'config.log' is not printed to output by default

talregev:
Printed some of it. and save config.log. Even if save only the print output I will never compare 2 tabs with long text. It not logical. I compare 2 logs files.

deploy/prepare_deployment.py
```diff
@@ -214,79 +214,115 @@
     '3rdParty/mingw/vcpkg/buildtrees/*.log',
     '3rdParty/Windows/vcpkg/buildtrees/*.log',
     'android/BOINC/app/build/reports/',
+    'mac_build/xcodebuild_*.log',
```
AenBleidd, line 217: @talregev, please remove all changes but this line and I'll merge this PR

AenBleidd: Already done here: https://github.com/BOINC/boinc/pull/4845

PR #4843, opened 2022-07-13
Vboxwrapper improvements

computezrmle:
**Description of the Change**
Vboxwrapper 26204 calls "vboxmanage -q list hdds" which can lead to a race condition when multiple instances of vboxwrapper (even older versions) are running concurrently.
This PR replaces "list hdds" with "showhdinfo ..." which implicitly asks for a lock.

Since "showhdinfo" requires a bit more error handling this is used to automatically set a new UUID to the vdi if necessary.

**Alternate Designs**
N/A

**Release Notes**
Fix possible race condition in Vboxwrapper 26204

### DIFF
--- a/samples/vboxwrapper/vbox_common.cpp
+++ b/samples/vboxwrapper/vbox_common.cpp
@@ -661,7 +661,7 @@ int VBOX_BASE::get_scratch_directory(string& dir) {
 int VBOX_BASE::get_slot_directory(string& dir) {
     char slot_dir[256];
 
-    if (getcwd(slot_dir, sizeof(slot_dir))) {;}
+    getcwd(slot_dir, sizeof(slot_dir));
     dir = slot_dir;
 
     if (!dir.empty()) {
--- a/samples/vboxwrapper/vbox_common.cpp
+++ b/samples/vboxwrapper/vbox_common.cpp
@@ -661,7 +661,7 @@ int VBOX_BASE::get_scratch_directory(string& dir) {
 int VBOX_BASE::get_slot_directory(string& dir) {
     char slot_dir[256];
 
-    getcwd(slot_dir, sizeof(slot_dir));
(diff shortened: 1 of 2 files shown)


davidpanderson:
Looks fine to me.

AenBleidd:
I'll test this in weekend.

computezrmle:
Extended test at LHC@home -dev have shown that there are still severe errors when multiple instances of vboxwrapper run concurrently.
This needs more investigation and additional error handling.

lfield:
Please merge and release. We will test it on LHC@home Dev.

AenBleidd:
I'll make a review and merge it ASAP

computezrmle:
Tasks from LHC@home -dev are using vboxwrapper 26205 (Windows/Linux) from an artifact here for a couple of days.
They do not show any vboxwrapper related issues.
A couple of (Linux) tasks from Rosetta also succeeded.

Hence, this PR may be reviewed/merged and vboxwrapper 26205 may be released.

AenBleidd:
@computezrmle, thank you for the update. I'll make review tonight.

samples/vboxwrapper/vbox_common.cpp
```diff
@@ -661,7 +661,7 @@ int VBOX_BASE::get_scratch_directory(string& dir) {
 int VBOX_BASE::get_slot_directory(string& dir) {
     char slot_dir[256];
 
-    getcwd(slot_dir, sizeof(slot_dir));
+    if (getcwd(slot_dir, sizeof(slot_dir))) {;}
```
AenBleidd, line 664: Maybe instead of this weird looking code is better to check for an error? In case of function failure the content of `slot_dir` might contain thrash and further crash the whole application.

computezrmle: Cancelled this change and reverted back to the original code.

PR #4849, opened 2022-07-16
[osx] Add osx check for universal lib

talregev:
Add osx check for universal lib.


talregev:
@AenBleidd ready for review.

CharlieFenton:
I am unclear which libraries you want to test, the dependent libraries (libssl.a, libcurl.a etc.) or the BOINC libraries (libboinc.a, libboinc_api.a, etc.) If you want to test the dependent libraries, that is already being done by 3rdParty/buildMacDependencies.sh. If either architecture is missing from a dependent library, that library is rebuilt. Of course, that does not check the library after it is rebuilt.

In any case, the easiest way to check for the presence of both x86_64 and arm64 in a file is:
`lipo ${path} -verify_arch x86_64 arm64`
This will return a result of 0 if both architectures are present, or 1 if either or both is missing.

talregev:
@CharlieFenton Thank you for review my code.
My intention is to check the dependent libraries after they rebuild, and BOINC libraries and binaries of boinc if they are universal. That if some bug happen that we know about that. That if for example one of the boinc library is no longer universal we can know quickly if it come from  dependent libraries.
Thank you for your advice how to check universal, I will change my code accordingly. 
The binaries list is not complete. Can you tell me the complete list? If not I can do it in next PR.

talregev:
It not so good for me to do as you suggested:
`lipo ${path} -verify_arch x86_64 arm64`
from here:
https://ss64.com/osx/lipo.html
> Take one input file and verify the specified arch_types are present in the file. 
If so then exit with a status of 0 else exit with a status of 1.

I don't it exit immediately. I want to type the some information before exit.
Can you verify that my way is also work?

CharlieFenton:
@talregev The man page on my Mac for lip has additional information, which told me _more than one arch_type can be
              verified at once_:
> -verify_arch arch_type ...
              Take one input file and verify the specified arch_types are
              present in the file.  If so then exit with a status of 0 else exit
              with a status of 1. Because more than one arch_type can be
              verified at once, all of the input files must appear before the
              -verify_arch flag on the command-line.

I tested this and confirmed that it works on my Mac using a path to _libboinc.a_:
`lipo ${path} -verify_arch x86_64 arm64` succeeds
`lipo ${path} -verify_arch x86_64 ppc` fails

I will check your code later tonight; I don't have time right now.

CharlieFenton:
@talregev Your code seems to work correctly, but I believe this would accomplish the same thing:
```
lipo ${lib} -verify_arch x86_64 arm64
     if [ $? -ne 0 ]; then
        echo "lib: $(basename $lib) is not universal"
         lipo -info $lib
         exit 1
     fi
```

talregev:
> @talregev Your code seems to work correctly, but I believe this would accomplish the same thing:
> ```
> lipo ${lib} -verify_arch x86_64 arm64
>      if [ $? -ne 0 ]; then
>         echo "lib: $(basename $lib) is not universal"
>          lipo -info $lib
>          exit 1
>      fi
> ```

I want to do your code but:
- When process exit with 1, the ci immidaitly fail.
- I try before -verify_arch with 2 arch and it gives me an error. Maybe I need to write twice -verify_arch on the same command?

I will try both to show you.

talregev:
@CharlieFenton Thank you for your comment and code. 
I use your code. Inside the if, the process is not exit when it finish with 1 as I wanted. 
Can you review my code again? 
Thank you.

CharlieFenton:
@talregev I think the problem with your new code is that you are testing for output from the lipo `-verify_arch` command. But lipo never generates output to stdout for the `-verify_arch` command, so you must test the result with `if [ $? -ne 0 ]` as I did in my suggested code.

talregev:
@CharlieFenton The if condition is checking status code of the process. 
You can check on here:
https://stackoverflow.com/a/26675771

Also note I check on vcpkg libs too and it fail there because they are not universal and it let me know.

CharlieFenton:
> The if condition is checking status code of the process.
You can check on here:
https://stackoverflow.com/a/26675771

@talregev I am not an expert on shell scripts, but my reading of that article is different from yours. It says:
> That being said, if you are running the command and are wanting to test its output, using the following is often more straightforward.
if some_command; then
    echo command returned true
else
    echo command returned some error
fi

And it also says:
> Note though that neither of those cares what the error code is. If you know you only care about a specific error code then you need to check $? manually.

As I wrote earlier, the lipo -verify_arch command has no _output_, so the form you used will always give the same result. As I understand, the _status code_ is just another way of saying _error code_ or _exit status_. This is different from the _output_.

talregev:
> And it also says:
> 
> > Note though that neither of those cares what the error code is. If you know you only care about a specific error code then you need to check $? manually.
> 
> As I wrote earlier, the lipo -verify_arch command has no _output_, so the form you used will always give the same result. As I understand, the _status code_ is just another way of saying _error code_ or _exit status_. This is different from the _output_.

@CharlieFenton Thank you for your time to look again and verify it. 
When I write status code, I ment error code/ exit code. all these names the same.
If condition is checking error code. As is written as a side node, that you cannot distinguish on error code. For example if you have error code 1 do foo1 and if you have error code 2, do foo2. It just success or fails as we needed. 
For summary, the if condition don't check the output of the command. (there is not -ne 0 or similar check).
The if condition checking the status code / error code / exit code of the process. (for example, if you type false on bash then check the error code, you will see it 1. for true is 0).

talregev:
I cannot run your command as it is because on the ci it stop immediately when process exit with error code 1 as it should. (like lipo -verify_arch).
You can read more on if condition from official manual:

https://www.gnu.org/software/bash/manual/html_node/Conditional-Constructs.html

.github/workflows/osx.yml
```diff
@@ -55,10 +55,18 @@ jobs:
         if: success() && ! contains(matrix.type, 'cmake')
         run: ./mac_build/buildMacBOINC-CI.sh --no_shared_headers
 
+      - name: test universal
```
AenBleidd, line 58: I don't know why this test is outside the build procedure. For me, if build script has failed to build any binary in an expected mode, it should immediately fail, so I'd suggest to put this verification to `./mac_build/buildMacBOINC-CI.sh` script and run on every built target (and not just by enumerating files in the directory). Then it will be more consistent and clear.

mac_build/test_universal.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+set -e
+
+folders_list="3rdParty/buildCache/mac/lib mac_build/build/Deployment"
+libs_list=$(find $folders_list -type f -name "*.a" -o -name "boinc")
+
+for lib in $libs_list; do
+    echo "Test lib: $lib"
+    if ! lipo $lib -verify_arch x86_64 arm64; then
```
AenBleidd, line 9: This `if` statement looks weird but according to my internal testing, it should work

CharlieFenton:
@talregev I just ran some tests again and now your code is working properly for me. I don't know why it did not seem to work when i tried it yesterday.
Testing this code also works for me:
```
% lipo $lib -verify_arch x86_64 arm64;if [ $? -ne 0 ]; then echo "lib: $(basename $lib) is not universal"; else echo "lib: $(basename $lib) is universal"; fi
lib: libboinc.a is universal
% lipo $lib -verify_arch x86_64 ppc;if [ $? -ne 0 ]; then echo "lib: $(basename $lib) is not universal"; else echo "lib: $(basename $lib) is universal"; fi
lib: libboinc.a is not universal
%
```
> I cannot run your command as it is because on the ci it stop immediately when process exit with error code 1 as it should. (like lipo -verify_arch).

I am not sure I understand what you are saying. My sample code contains `exit 1` because I was adapting your code which also exited if the library is not universal. You have two copies of your code. The version in _mac_build/  test_universal_ has the `exit 1` but the version in osx/test_universal does not exit. 

But since your code works, use it if you are happy with it. However, I do agree with @AenBleidd that it should be in the script to build dependent libraries. Perhaps `3rdParty/buildMacDependencies.sh` would be an even better place than `./mac_build/buildMacBOINC-CI.sh`. Note that `3rdParty/buildMacDependencies.sh` already checks the architecture of libraries before deciding whether to rebuild them; you could add another check after each build.

_Important_: I just looked at the code in `3rdParty/buildMacDependencies.sh` and it is checking only for x86_64. Apparently this was never updated to include both x86_64 and arm64. Please fix this. I believe you can just change "x86_64" to "x86_64 arm64" for each of the `download_and_build` calls at the end of the script.

talregev:
> ```
% lipo $lib -verify_arch x86_64 arm64;if [ $? -ne 0 ]; then echo "lib: $(basename $lib) is not universal"; else echo "lib: $(basename $lib) is universal"; fi
lib: libboinc.a is universal
% lipo $lib -verify_arch x86_64 ppc;if [ $? -ne 0 ]; then echo "lib: $(basename $lib) is not universal"; else echo "lib: $(basename $lib) is universal"; fi
lib: libboinc.a is not universal
%
```
You code is working and correct. But don't work on the ci.

> I am not sure I understand what you are saying. My sample code contains exit 1 because I was adapting your code which also exited if the library is not universal. You have two copies of your code. The version in mac_build/ test_universal has the exit 1 but the version in osx/test_universal does not exit.

When I talk about error code 1, I talk the error code 1 for this command: `lipo $lib -verify_arch x86_64 arm64`
When this command don't find the exact arches as we specify it, it give error code 1 as it should.
Also the ci stop immediately after each command that give error code 1. (like  `lipo $lib -verify_arch ...`).
Then it will not continue to the rest of the if and not print the information we need:
```
     if [ $? -ne 0 ]; then
        echo "lib: $(basename $lib) is not universal"
         lipo -info $lib
         exit 1
     fi
```
I preferer to do external check and not inside the scrip because if we change the check, it will not validate the cache. Also more safe without bugs. (ratter to change your script that already working, and we need to verify it again, and maybe input more bugs). 

I will try to add all targets/binaries and I wanted and @AenBleidd suggested. If you have a list of binaries it will be helpful

> Important: I just looked at the code in 3rdParty/buildMacDependencies.sh and it is checking only for x86_64. Apparently this was never updated to include both x86_64 and arm64. Please fix this. I believe you can just change "x86_64" to "x86_64 arm64" for each of the download_and_build calls at the end of the script.

done.

Thank you again for your time and review and for verify my way.

File: mac_build/test_universal.sh (hunk 2 above)
Comment (talregev) [reply mac_build/test_universal.sh:9]: Charlie and you both verify it working.

**Comment by CharlieFenton** [comment]
> If you have a list of all binaries it will be helpful.

OpenSSL creates two libraries, libssl.a and libcrypto.a, so there are 7 third-party dependent libraries:
libcares.a
libcurl.a
libfreetype.a
libftgl.a
libssl.a
libcrypto.a
libwx_osx_cocoa_static.a

If you want a list of BOINC binaries built by `buildMacBOINC-CI.sh`, the BOINC libraries are:
libboinc.a
libboinc_graphics2.a
libboinc_opencl.a

`buildMacBOINC-CI.sh` builds the following bare executables (command-line executables):
switcher
gfx_switcher
boincscr
setprojectgrp
boinccmd
SetUpSecurity
AddRemoveUser

`buildMacBOINC-CI.sh` also builds these applications and plug-ins. But note that these are "bundles" which are directories containing the actual executable file several levels deep. To test their architectures, you must run `lipo` on the executable code. For example, to test the architectures of _BOINCManager.app_, the executable is at _BOINCManager.app/Contents/MacOS/BOINCManager_.
BOINCManager.app
PostInstall.app
BOINC Installer.app
Uninstall BOINC.app
BOINCSaver.saver

The following command-line executables were added after @ChristianBeer created `buildMacBOINC-CI.sh` and so are not currently being built under CI, though they probably should be:
boinc_finish_install
detect_rosetta_cpu
gfx_cleanup

`buildMacBOINC-CI.sh` does not build these libraries, which are used only by projects:
libboinc_api.a
libboinc_zip.a

**Comment by CharlieFenton** [comment]
> The following command-line executables were added after @ChristianBeer created buildMacBOINC-CI.sh and so are not currently being built under CI, though they probably should be:

the corresponding target for each of these has the same name as the binary it builds:
boinc_finish_install
detect_rosetta_cpu
gfx_cleanup

**Comment by CharlieFenton** [comment]
> Also the ci stop immediately after each command that give error code 1. (like lipo $lib -verify_arch ...).
Then it will not continue to the rest of the if and not print the information we need:

Why does this happen with my code but not yours? Both have the same call of `lipo $lib -verify_arch ....` Is it because you did not remove the `exit 1` from my suggested code (which I only included because you had it in your earlier code)? Or is it because yours is inside an `if` statement?

**Comment by talregev** [comment]
> > Also the ci stop immediately after each command that give error code 1. (like lipo $lib -verify_arch ...).
> Then it will not continue to the rest of the if and not print the information we need:
> 
> Why does this happen with my code but not yours? Both have the same call of `lipo $lib -verify_arch ....` Is it because you did not remove the `exit 1` from my suggested code (which I only included because you had it in your earlier code)? Or is it because yours is inside an `if` statement? 
> 

I think it connect to `set -e` in the beginning of the script.
The if condition don't report the output of the testing commads. that why inside the if it working.

**Comment by talregev** [comment]
> > If you have a list of all binaries it will be helpful.
> 
> OpenSSL creates two libraries, libssl.a and libcrypto.a, so there are 7 third-party dependent libraries: libcares.a libcurl.a libfreetype.a libftgl.a libssl.a libcrypto.a libwx_osx_cocoa_static.a
> 
> If you want a list of BOINC binaries built by `buildMacBOINC-CI.sh`, the BOINC libraries are: libboinc.a libboinc_graphics2.a libboinc_opencl.a
> 
> `buildMacBOINC-CI.sh` builds the following bare executables (command-line executables): switcher gfx_switcher boincscr setprojectgrp boinccmd SetUpSecurity AddRemoveUser
> 
> `buildMacBOINC-CI.sh` also builds these applications and plug-ins. But note that these are "bundles" which are directories containing the actual executable file several levels deep. To test their architectures, you must run `lipo` on the executable code. For example, to test the architectures of _BOINCManager.app_, the executable is at _BOINCManager.app/Contents/MacOS/BOINCManager_. BOINCManager.app PostInstall.app BOINC Installer.app Uninstall BOINC.app BOINCSaver.saver
> 
> The following command-line executables were added after @ChristianBeer created `buildMacBOINC-CI.sh` and so are not currently being built under CI, though they probably should be: boinc_finish_install detect_rosetta_cpu gfx_cleanup
> 
> `buildMacBOINC-CI.sh` does not build these libraries, which are used only by projects: libboinc_api.a libboinc_zip.a

Thank you for the complete list of binaries!!!! It will very helpful.

**Comment by talregev** [comment]
> are not currently being built under CI, though they probably should be

> does not build these libraries, which are used only by projects

We can think how to build them on the ci. I am in favor of that. It out of the scope of this pr and we can discuss it on next issue/pr how to do it.

**Comment by talregev** [comment]
@CharlieFenton @AenBleidd 
Ready for review.

File: .github/workflows/osx.yml (hunk 1 above)
Comment (talregev) [reply .github/workflows/osx.yml:58]: I preferer to do external check and not inside the scrip because if we change the check, it invalidate the cache and take more time to build. Also it I am doing this check after all the build is done, to change the script for every target I will insert some bugs, ratter to change your script that already working, and we need to verify it again. It will take more time.
Also we have many lines of output in the test, we need some other part to see it very quickly what happen. It  more readability.

**Comment by talregev** [comment]
> I assume that _mac_build/test_universal.sh_ is intended to be used only in the Ci builds, based on line 5: `folders_list="3rdParty/buildCache/mac/lib mac_build/build/Deployment"` If so, please change the name of this file to _test_universal-CI.sh_ to make it clear this does not apply to production builds on an individual Mac computer but only to CI builds. This is the convention with the naming of _mac_build/buildMacBOINC-CI.sh_ to distinguish it from _buildMacBOINC.sh_.
> 
> Beyond that, the code looks OK to me, though of course I don't really have detailed knowledge of programming for the CI system.

Thank you for your review. Yes this scrip is design to run on ci, but I think it can also run after production run. (I think it the same folders. Folders are relative.). I rename the scrip as you requested.

File: .github/workflows/osx.yml (hunk 1 above)
Comment (AenBleidd) [reply .github/workflows/osx.yml:58]: I still would like to have this inside the ./mac_build/buildMacBOINC-CI.sh script because in this case it will be more logical: we build binary and check it right after the build. Now your solution have decoupled binary verification and it's hard to remember to fix verification list in case of adding/removing binaries from the build.
So I strongly recommend to follow my suggestion and move this check to the ./mac_build/buildMacBOINC-CI.sh file.

**Comment by CharlieFenton** [comment]
@talregev 
> Yes this scrip is design to run on ci, but I think it can also run after production run.

I always build releases using Xcode app GUI directly rather than the `BuildMacBOINC.sh` script, because it provides information about build issues in a much more easily viewed format. The output then goes to a completely different directory. But my `mac_installer/release_boinc.sh` script knows where to find them whichever way it is built. Whether the build is done using the Xcode app GUI or the `BuildMacBOINC.sh ` script, on an individual Mac , there is never a _3rdParty/buildCache_ directory.

**Review by CharlieFenton — APPROVED** [review]
Thank you for  changing the file name. I tentatively approve this request, but I still agree with @AenBleidd that it would be better to have the test inside `mac_buid/BuildMacBOINC-CI.sh`. If you do that, you should make the test a subroutine inside that script instead of a separate script.

**Review by CharlieFenton — APPROVED** [review]

File: mac_build/buildMacBOINC-CI.sh [hunk 3]
```diff
@@ -257,3 +257,30 @@ fi
...
+
+set -e
+ 
+echo "Testing libs and binaries for univaral build."
+echo ""
+
+binaries_libs_list="*.a boinc switcher gfx_switcher boincscr setprojectgrp boinccmd SetUpSecurity AddRemoveUser BOINCManager PostInstall BOINCSaver *Installer Uninstall*"
+folders_list="3rdParty/buildCache/mac/lib mac_build/build/Deployment"
+
+for object in $binaries_libs_list; do
```
AenBleidd, line 269: Please move this check to the corresponding block above and test only those binaries that are build in those block. We don't need to wait to build everything if any of the targets was build with the wrong architecture.
Please do this as I suggested you before.

mac_build/buildMacBOINC-CI.sh
```diff
@@ -257,3 +257,30 @@ fi
 echo "Building ${target}...done"
 
 cd ..
+
+set -e
```
AenBleidd, line 261: Will this affect build script in any negative way if put this at the very beginning of the file?

AenBleidd (changes requested):

mac_build/buildMacBOINC-CI.sh
talregev: I don't have a mac to understand which target belong to which binary. Also on android we test after all the binary are build. Mac when cache, is build around 5 min. So waiting for results for that, that ok. 
There never wasn't a test before, and now you sudden rush as immediate if the binary are wrong to fail the CI?
I will not invest more time for that. I am doing more meaningful tasks.

AenBleidd: I didn't ask you to add this check. You decided to do this by yourself. And telling me now that you don't have Mac and you can't check smth is not an excuse to leave the code in a non-appropriate state. From the very beginning I told you have I want to see this functionality, so don't blame me that you decided to ignore me and do it your own way.
If you don't want to make it properly - leave it, maybe one day I will find some time to make it properly.
I can't approve this PR in its current state.

talregev: When I do something here because you told me?
I am doing stuff here because I care, because I want to change, and it for common good, and not between you and me.
The test is working, and Charlie already approve it. You are very good to preserve your own time, but want to waste other time. You can accept this code, and "fix" it when you "have time". 
Don't hold it up, and don't waste my time like you don't want to waste yours.

talregev: Please remove wait for author respond. I already told my respond.

AenBleidd: At first, calm down. 
Second, I don't waste time of anyone. You don't want to work on smth - nobody forces you to do that. But if you started smth - please be patient and finish your work.
Third, that fact that you care about smth doesn't mean that you should produce unfinished or incorrect code. In this case improvement is required to make this code looks nice and professional. If you're professional - you should listen to the other people and hear their arguments: "I don't have a mac" and "I am doing more meaningful tasks." are not valid arguments from your side to prove that your opinion is correct.
I gave my arguments: checking built binaries right after they are built is way more efficient than waiting for all of them to build and then try to understand which target was not build correctly.
If you don't agree and don't want to proceed - fine, leave it and I'll finish it when I have time for this.

AenBleidd: > Please remove wait for author respond. I already told my respond.

I made a new label specially for you.

mac_build/buildMacBOINC-CI.sh
```diff
@@ -30,6 +30,8 @@
 # --no_shared_headers will build targets individually instead of in one call of BuildMacBOINC.sh (NOT recommended)
 
 # check working directory because the script needs to be called like: ./mac_build/buildMacBOINC-CI.sh
+set -e
```
AenBleidd, line 33: Since there is no check anymore, this string is redundant. Please remove

AenBleidd (changes requested):

talregev:
I don't allow to copy my work.

PR #4811, opened 2022-06-24
[Manager][linux] Search for skins in /usr/(local)/share/boinc-manager/skins or ./skins folders.

AenBleidd:
This fixes #4809.

Signed-off-by: Vitalii Koshura <lestat.de.lionkur@gmail.com>

PR #4851, opened 2022-07-22
[snap] Add snap build

talregev:
Add snap build.


talregev:
Is working. Currently limitation: Skins is not correct. Cannot forward x11.
image

talregev:
locale and other project working too:
image

AenBleidd:
Why skins are not included in the package? Is it possible to fix this?

talregev:
> Why skins are not included in the package? Is it possible to fix this?

The skins are included. I need to understand why it not recognized it. It still on develop mode.

AenBleidd:
> > Why skins are not included in the package? Is it possible to fix this?
> 
> 
> 
> The skins are included. I need to understand why it not recognized it. It still on develop mode. 

Check the path where skins are unpacked. It might be the wrong location, so either you need to unpack it to the correct place or fix the code that looks for skins.

talregev:
from here:
https://github.com/BOINC/boinc/pull/4811/files#diff-ed2d87bd7bb9a24765f0dfe83cdf2de42bde16f60be8fec9b865678447caf842R982

```c++
    strSkinLocation = wxGetApp().GetRootDirectory();
    if (strSkinLocation.StartsWith("/usr/")) {
        strSkinLocation += wxT("/../share/boinc-manager/skins");
    }
```

Linux guys install it on /usr/local? 
Why you going down in the folder? (`../share`)?

AenBleidd:
When installed via package manager, it goes to /usr. When via 'make install' - it goes to /usr/local.
That's the Linux rule. share folder is for non-executable files (resources).

talregev:
> When installed via package manager, it goes to /usr. When via 'make install' - it goes to /usr/local. That's the Linux rule. share folder is for non-executable files (resources).

You miss my point. I was referring the `..` in the line. I think I understand, because the binary is found on the bin folder, then you exit it, then you go to the correct way. And I think I know why it not working. I will try to do some fix in the code. 
By the way, the locale is found. How we check on the locale? Can you find me the corresponding line for the locale?

AenBleidd:
> How we check on the locale? Can you find me the corresponding line for the locale?

https://github.com/BOINC/boinc/blob/3df4f63f2f418ad1f1d90d68fac3b8e3ab944a2b/clientgui/BOINCGUIApp.cpp#L276-L288

talregev:
> > How we check on the locale? Can you find me the corresponding line for the locale?
> 
> https://github.com/BOINC/boinc/blob/3df4f63f2f418ad1f1d90d68fac3b8e3ab944a2b/clientgui/BOINCGUIApp.cpp#L276-L288

Thank you! I am not sure I understand this logic, but it different from the skin. I will try with my own fix to see it working.

talregev:
I fix the skins lookup. 
image

talregev:
Fix x11 forwarding: 
image

talregev:
The skins still working after the change.

talregev:
I compile with wxwidgets that come with ubuntu (3.0.4) Then I was able to switch to advanced mode:
image

talregev:
```
> boinc --help
The command-line options for /snap/boinc/x1/usr/bin/boinc are intended for debugging.
The recommended command-line interface is a separate program,'boinccmd'.
Run boinccmd in the same directory as /snap/boinc/x1/usr/bin/boinc.

Usage: /snap/boinc/x1/usr/bin/boinc [options]
    --abort_jobs_on_exit           when client exits, abort and report jobs
    --allow_remote_gui_rpc         allow remote GUI RPC connections
    --allow_multiple_clients       allow >1 instances per host
    --attach_project <URL> <key>   attach to a project
    --check_all_logins             for idle detection, check remote logins too
    --daemon                       run as daemon (Unix)
    --detach_console               detach from console (Windows)
    --detach_project <URL>         detach from a project
    --dir <path>                   use given dir as BOINC home
    --exit_after_app_start N       exit N seconds after an app starts
    --exit_after_finish            exit right after finishing a job
    --exit_before_start            exit right before starting a job
    --exit_before_upload           exit right before starting an upload
    --exit_when_idle               exit when there are no results
    --fetch_minimal_work           fetch only 1 job per device
    --file_xfer_giveup_period N    give up on file xfers after N sec
    --gui_rpc_port <port>          port for GUI RPCs
    --gui_rpc_unix_domain          use Unix domain for GUI RPCs
    --help                         show options
    --insecure                     disable app sandboxing (Unix)
    --master_fetch_interval N      limiting period of master retry
    --master_fetch_period N        reload master URL after N RPC failures
    --master_fetch_retry_cap N     exponential backoff limit
    --no_gpus                      don't check for GPUs
    --no_gui_rpc                   don't allow GUI RPC, don't make socket
    --no_info_fetch                don't fetch project list or client version info
    --no_priority_change           run apps at same priority as client
    --pers_giveup N                giveup time for persistent file xfer
    --pers_retry_delay_max N       max for file xfer exponential backoff
    --pers_retry_delay_min N       min for file xfer exponential backoff
    --redirectio                   redirect stdout and stderr to log files
    --reset_project <URL>          reset (clear) a project
    --retry_cap N                  exponential backoff limit
    --run_cpu_benchmarks           run the CPU benchmarks
    --run_by_updater               set by updater
    --sched_retry_delay_max N      max for RPC exponential backoff
    --sched_retry_delay_min N      min for RPC exponential backoff
    --show_projects                show attached projects
    --skip_cpu_benchmarks          don't run CPU benchmarks
    --start_delay X                delay starting apps for X secs
    --suppress_net_info            don't send network addrs to server
    --unsigned_apps_ok             allow unsigned apps (for testing)
    --update_prefs <URL>           contact a project to update preferences
    --version                      show version info
```

```
> boinc.manager --help
Usage: boincmgr [-h] [--verbose] [-a] [-e <str>] [-d <str>] [-n <str>] [-g <str>] [-p <str>] [-b <str>] [-i] [-c] [-m] [-nd]
  -h, --help                    show this help message
  --verbose                     generate verbose log messages
  -a, --autostart               BOINC Manager was started by the operating system automatically
  -e, --clientdir=<str>         Directory containing the BOINC Client executable
  -d, --datadir=<str>           BOINC data directory
  -n, --namehost=<str>          Host name or IP address
```
@AenBleidd 
To run boinc with dir option it mean `boinc --dir <some dir>`
I need to run the boinc manager with datadir option? 
`boincmgr --datadir <the same some dir>`
?

AenBleidd:
@talregev, yes

talregev:
> @talregev, yes

Thank you.

talregev:
@AenBleidd @LocutusOfBorg I am not sure if always let boinc run on their home snap dir, or let the user decide? 
That it will not contaminant the folder the user choose to run boinc without any flags. 
How is in linux from apt? I know it going to the var folder?

AenBleidd:
@talregev, boinc data dir is configured via 'init.d' configuration file. See example in the attachmen
[boinc-client.txt](https://github.com/BOINC/boinc/files/9221993/boinc-client.txt)
t

talregev:
> @talregev, boinc data dir is configured via 'init.d' configuration file. See example in the attachment [boinc-client.txt](https://github.com/BOINC/boinc/files/9221993/boinc-client.txt)

Thank you for share with me. I see:
`BOINC_DIR=/var/lib/boinc-client`
That mean when linux user run boinc, it direct it to this folder? without any flag?

AenBleidd:
@talregev, init.d config file is for init.d subsystem: https://www.novell.com/documentation/suse91/suselinux-adminguide/html/ch13s04.html
So basically linux user don't run boinc directly, but when it's run via init.d, it uses data from this configuration file without any additional flags

talregev:
> @talregev, init.d config file is for init.d subsystem: https://www.novell.com/documentation/suse91/suselinux-adminguide/html/ch13s04.html So basically linux user don't run boinc directly, but when it's run via init.d, it uses data from this configuration file without any additional flags

So for snap, if user run boinc snap, it will redirect all data to home snap application. I think it good behavior. 
What do you think?

AenBleidd:
@talregev, it depends where snap keeps 'home': boinc data dir might be very big, so if everything comes to the main hard disk - that might be not good, and users might want to change default location. For new they are doing this by mounting (or making symlink) to this folder other bigger drive. Not sure this is an option for snap, but I know nothing about snap, so I can't answer this question.

talregev:
> @talregev, it depends on the location where snap keeps 'home': boinc data dir might be very big, so if everything comes to the main hard disk - that might be not good, and users might want to change default location. For now they are doing this by mounting (or making symlink) to this folder other bigger drive. Not sure this is an option for snap, but I know nothing about snap, so I can't answer this question.

They mound /var/lib/boinc-client to different location?
The same can done for snap home application.
`/home/$USER/snap/boinc/x1`

If I will set it to var it will be
`$SNAP/var/lib/boinc-client`

Any of this folder are prefer of one another?

AenBleidd:
`/home/$USER/snap/boinc/x1` this looks better to me

talregev:
> `/home/$USER/snap/boinc/x1` this looks better to me

ok. So it already set to there.

talregev:
So this PR is ready for testing. I will make it ready for review once we merge osx PR.

AenBleidd:
This PR and #4849 are not related to each other so they could be tested/fixed/merged separately.

talregev:
> `/home/$USER/snap/boinc/x1` this looks better to me

I was thinking about this.
The home path can be change if the boinc snap been upgraded. 
Also when remove boinc snap, it also remove.
So I want to put the boinc data dir here:
/home/$USER/.boinc-client in hidden folder inside home.

AenBleidd:
@talregev, it's better then to put it to the same location that is used on linux now. It's easier for users to search for it

talregev:
> 



> better then to put it to the same location that is used on linux now. It's easier

It will be similar: `$SNAP/var/lib/boinc-client`  mean `snap/boinc/x1/var/lib/boinc-client`
And I am not sure how it handle the permission.
The  `snap/boinc/x1/var/lib/boinc-client` is acceptable?

AenBleidd:
@talregev, wait, above you wrote next path `/home/$USER/.boinc-client` and it contains no 'snap' prefix. So if you can put it here, why you can't put it into `/var/lib/boinc-client`?

talregev:
> @talregev, wait, above you wrote next path `/home/$USER/.boinc-client` and it contains no 'snap' prefix. So if you can put it here, why you can't put it into `/var/lib/boinc-client`?

Because snap give the real home directory. with special var, and I can give the app the premotion for home directory. 
I cannot do the same for system.

talregev:
> @talregev, wait, above you wrote next path `/home/$USER/.boinc-client` and it contains no 'snap' prefix. So if you can put it here, why you can't put it into `/var/lib/boinc-client`?

Also I don't think I can put in hidden folder in home. also another limitation.

AenBleidd:
@talregev, then put it somewhere in home directory, and if it's not hidden - that's completely fine

talregev:
> @talregev, then put it somewhere in home directory, and if it's not hidden - that's completely fine

Currently if the folder is not exist, boinc fail to run. So For now I am switching to snap home:
`/home/$USER/snap/boinc/x1` until we change the code to handle that.

talregev:
compile boinc manager snap with wxwidgets 3.2.0 and can switch advanced mode, also reduce file size.
image

talregev:
Boinc now have desktop file, can show from menu:
image

talregev:
I don't allow to copy my work.

PR #4853, opened 2022-07-27
Avoid gcc warnings

computezrmle:
If '-Wunused-result' is set compilers usually print a warning when they expect a function to return 'something' but that 'something' is not used in the subsequent code.
The usual method to omit the warning is to precede the function call with '(void)'.

In case of gcc this doesn't work for years:
```
switcher.cpp: In function ‘int main(int, char**)’:
switcher.cpp:118:26: warning: ignoring return value of ‘int setgid(__gid_t)’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
  118 |             (void) setgid(grp->gr_gid);
      |                    ~~~~~~^~~~~~~~~~~~~
switcher.cpp:125:26: warning: ignoring return value of ‘int setuid(__uid_t)’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
  125 |             (void) setuid(pw->pw_uid);
      |                    ~~~~~~^~~~~~~~~~~~

```
The changes here compile without a warning.

### DIFF
--- a/client/switcher.cpp
+++ b/client/switcher.cpp
@@ -115,14 +115,14 @@ int main(int /*argc*/, char** argv) {
         // effective group ID and saved set_group-ID for this process
         grp = getgrnam(boinc_project_group_name);
         if (grp) {
-            (void) setgid(grp->gr_gid);
+            if (setgid(grp->gr_gid)) {;}
         }
 
         // We are running setuid root, so setuid() sets real user ID,
         // effective user ID and saved set_user-ID for this process
         pw = getpwnam(boinc_project_user_name);
         if (pw) {
-            (void) setuid(pw->pw_uid);
+            if (setuid(pw->pw_uid)) {;}
         }
     }
 


CharlieFenton:
@computezrmle Please note that _switcher_ is used only on the Macintosh. It should **never** be built for Linux or Windows systems. This has come up before, and no one has ever fixed the make files. If you are encountering this when building for a non_Mac system using `make`, the correct fix is to modify _client/makefile.am_ so it does not build _switcher_ for any system other than Macintosh.

I am no expert in make files, but I believe that this incorrect code in _client/makefile.am_:
```
if OS_DARWIN
   LIBS += -framework IOKit -framework Foundation -framework ScreenSaver -framework Cocoa -framework CoreServices
endif

bin_PROGRAMS = boinc_client boinccmd boinc
if !OS_WIN32
bin_PROGRAMS += switcher
endif
```
should be changed to this:
```
bin_PROGRAMS = boinc_client boinccmd boinc

if OS_DARWIN
   bin_PROGRAMS += switcher
   LIBS += -framework IOKit -framework Foundation -framework ScreenSaver -framework Cocoa -framework CoreServices
endif

```
Xcode uses the LLVM Apple clang compiler, not gcc. I don't see these warnings when building switcher either directly with the Xcode GUI or with _mac_build/BuildMacBOINC.sh_

computezrmle:
@CharlieFenton
Thanks for pointing this out.
I didn't knew that switcher is Mac specific.
Just saw the warnings during compilation.

I agree that this is gcc specific.
I corrected the same issue (1 line) in vboxwrapper's vbox_common.cpp recently and it passed all automatic tests.
Hence, it looks like the fix doesn't hurt if other compilers are used.

If nobody disagrees I'd like to keep this PR but I can also close it.

computezrmle:
This PR is obsolete if #4854 is accepted instead.

CharlieFenton:
@computezrmle Thank you for your new PR #4854.
> If nobody disagrees I'd like to keep this PR but I can also close it.

I prefer we not make unnecessary changes to switcher.

PR #4336, opened 2021-04-19
client: add a cooldown period before resuming operations

Flowdalic:
Add a cooldown period, currently 30 seconds, before resuming
operations to prevent frequent running/suspended switches.

BOINCs non-BOINC CPU usage determination is often incorrect (see issue
core machine which compiles LLVM while BOINC is running. BOINC is
configured to suspend computation once non-BOINC CPU usage exceeds
10%. As you can see BOINC frequently reports 0.00% CPU usage. This is
due issue #4327. A workaround is suggested in pull request #4328. This
would change the incorrectly reported value from 0% to 50% non-BOINC
CPU usage. However there are still times where a non-BOINC CPU usage
below 10% is reported. At those times, BOINC would resume its
operation. However in the next cycle, the non-BOINC CPU usage exceeds
10% again. To avoid frequent running/suspended switches of BOINC, this
adds a cooldown period of 30 seconds before BOINC resumes computation.

A logical next step would be to make this cooldown period
configurable, including an option to completely disable the cooldown
period (to restore the original behavior).

15:25:08 [---] [mem_usage] non-BOINC CPU usage: 0.66%
15:25:18 [---] [mem_usage] non-BOINC CPU usage: 0.61%
15:25:28 [---] [mem_usage] non-BOINC CPU usage: 0.57%
15:25:38 [---] [mem_usage] non-BOINC CPU usage: 0.60%
15:25:48 [---] [mem_usage] non-BOINC CPU usage: 0.62%
15:25:58 [---] [mem_usage] non-BOINC CPU usage: 0.62%
15:26:08 [---] [mem_usage] non-BOINC CPU usage: 0.62%
15:26:18 [---] [mem_usage] non-BOINC CPU usage: 0.71%
15:26:28 [---] [mem_usage] non-BOINC CPU usage: 0.65%
15:26:38 [---] [mem_usage] non-BOINC CPU usage: 9.05%
15:26:48 [---] [mem_usage] non-BOINC CPU usage: 10.87%
15:26:58 [---] [mem_usage] non-BOINC CPU usage: 0.00%
15:27:08 [---] [mem_usage] non-BOINC CPU usage: 0.76%
15:27:18 [---] [mem_usage] non-BOINC CPU usage: 0.62%
15:27:29 [---] [mem_usage] non-BOINC CPU usage: 2.00%
15:27:39 [---] [mem_usage] non-BOINC CPU usage: 1.80%
15:27:49 [---] [mem_usage] non-BOINC CPU usage: 2.04%
[...]

PR #4859, opened 2022-07-30
fix non-BOINC CPU usage limit on Win and Linux

### LINKED ISSUE #4327: Non-BOINC CPU usage time is incorrect

davidpanderson:
Fixes #4327, #4328

The pref "suspend_cpu_usage" lets you suspend BOINC computing
when the non-BOINC CPU usage (NBCU) exceeds a threshold T.

NBCU excludes BOINC-related processes
- the BOINC client
- the BOINC Manager
- BOINC apps and their process trees
- graphics apps
- VBox daemon
    This can use significant time.
    We count it as BOINC-related only if BOINC is running a Vbox app

NBCU also excludes low-priority processors:
"niced" on Linux, idle class on Win.
We don't want such processes to keep BOINC from computing.

This preference is enforced every 10 seconds.
The client estimates the NBCU over the last 10 sec,
and suspends BOINC processing if the load exceeds T

Currently, NBCU is estimated by scanning the list of all processes,
skipping the types of processes listed above,
and adding up the user and kernel time of others.
This gives a value Tcur.
Let Tprev be the previous value (from 10 seconds ago).
Let X = (Tcur - Tprev)/(ncpus*10).
We suspend processing if X/(ncpus*10) > T

This approach is fundamentally flawed.
It would be fine if processes never exited.
But if a process exits during a given 10-sec period,
its time is included in Tprev but not Tcur,
in which case X is wrong; it may be negative.
This can cause false negatives (not suspending when we should have suspended).

To fix this, we need a different approach.
We shouldn't estimate total CPU by adding up current processes.
Fortunately, most OS's have an API for getting total CPU time.
On Linux this is broken down into a lot of categories,
including "niced" processes.
On Win it's only broken down by user/kernel.

The new approach is basically:
every 10 sec, get the increase Y in total CPU (as reported by the OS).
Then estimate the total CPU time B
used by BOINC-related and low-priority processes in the last 10 sec,
and let X = Y - B.
Suspend processing if X/(ncpus*10) > T

We'll estimate B by scanning the process list and adding up
the totals for BOINC-related processes,
and computing the delta since 10 sec ago.
This is subject to the same error as above:
when these processes exit, B will be underestimated (typically negative),
so X will be overestimated, causing a false positive.
However, this happens infrequently;
BOINC jobs low-priority processes are generally long-running.
And we can avoid false positives by setting X to zero if B is negative.

Because of the possible error in X,
we'll use a conservative approach to resuming from a suspension:
if we're currently suspended because of CPU usage,
unsuspend only if X < T for two consecutive 10-sec periods

### DIFF
--- a/lib/procinfo.cpp
+++ b/lib/procinfo.cpp
@@ -135,8 +135,32 @@ void procinfo_non_boinc(PROCINFO& procinfo, PROC_MAP& pm) {
         procinfo.working_set_size += p.working_set_size;
     }
 #if 0
-    fprintf(stderr, "total non-boinc: %f %f\n", procinfo.user_time, procinfo.kernel_time);
+    fprintf(stderr,
+        "total non-boinc: %f %f\n", procinfo.user_time, procinfo.kernel_time
+    );
 #endif
+}
+
+// get CPU time of BOINC-related processes, low-priority processes,
+// and (if we're using Vbox) the Vbox daemon.
+//
+double boinc_related_cpu_time(PROC_MAP& pm, bool using_vbox) {
+    double sum = 0;
+    PROC_MAP::iterator i;
+    for (i=pm.begin(); i!=pm.end(); ++i) {
+        PROCINFO& p = i->second;
+#ifdef _WIN32
+        if (p.id == 0) continue;    // idle process
+#endif
+        if (
+            p.is_boinc_app
+            || p.is_low_priority
+            || (using_vbox && strstr(p.command, "VBoxSVC"))
+        ) {
+            sum += p.user_time;
+        }
+    }
+    return sum;
 }
 
 double process_tree_cpu_time(int pid) {
--- a/lib/procinfo.h
+++ b/lib/procinfo.h
@@ -83,7 +83,7 @@ extern double process_tree_cpu_time(int pid);
     // get the CPU time of the given process and its descendants
 
 extern double total_cpu_time();
-    // total CPU time, as reported by OS
+    // total user-mode CPU time, as reported by OS
 
 extern double boinc_related_cpu_time(PROC_MAP&, bool using_vbox);
     // total CPU of current BOINC processes, low-priority processes,
--- a/lib/procinfo.h
+++ b/lib/procinfo.h
@@ -82,4 +82,10 @@ extern void procinfo_non_boinc(PROCINFO&, PROC_MAP&);
 extern double process_tree_cpu_time(int pid);
     // get the CPU time of the given process and its descendants
 
+extern double total_cpu_time();
+    // total CPU time, as reported by OS
+
+extern double boinc_related_cpu_time(PROC_MAP&, bool using_vbox);
+    // total CPU of current BOINC processes, low-priority processes,
+    // and (if using vbox) the Vbox daemon
 #endif
--- a/lib/procinfo_unix.cpp
+++ b/lib/procinfo_unix.cpp
@@ -263,3 +263,24 @@ int procinfo_setup(PROC_MAP& pm) {
     find_children(pm);
     return 0;
 }
+
+// get total user-mode CPU time
+// see https://www.baeldung.com/linux/get-cpu-usage
+//
+double total_cpu_time() {
+    char buf[1024];
+    static FILE *f=NULL;
+    static double scale;
+    if (!f) {
+        f = fopen("/proc/stat", "r");
+        long hz = sysconf(_SC_CLK_TCK);
+        scale = 1./hz;
+    } else {
+        fflush(f);
+        rewind(f);
+    }
+    if (!fgets(buf, 256, f)) return 0;
+    double user, nice;
+    sscanf(buf, "cpu %lf %lf", &user, &nice);
+    return (user+nice)*scale;
+}
--- a/lib/procinfo_unix.cpp
+++ b/lib/procinfo_unix.cpp
@@ -273,14 +273,25 @@ double total_cpu_time() {
     static double scale;
     if (!f) {
         f = fopen("/proc/stat", "r");
+        if (!f) {
+            fprintf(stderr, "can't open /proc/stat\n");
+            return 0;
+        }
         long hz = sysconf(_SC_CLK_TCK);
         scale = 1./hz;
     } else {
         fflush(f);
         rewind(f);
     }
-    if (!fgets(buf, 256, f)) return 0;
+    if (!fgets(buf, 256, f)) {
+        fprintf(stderr, "can't read /proc/stat\n");
+        return 0;
+    }
     double user, nice;
-    sscanf(buf, "cpu %lf %lf", &user, &nice);
+    int n = sscanf(buf, "cpu %lf %lf", &user, &nice);
+    if (n != 2) {
+        fprintf(stderr, "can't parse /proc/stat: %s\n", buf);
+        return 0;
+    }
     return (user+nice)*scale;
 }
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -359,8 +359,8 @@ void ACTIVE_TASK_SET::get_memory_usage() {
     unsigned int i;
     int retval;
     static bool first = true;
-    static double last_cpu_time;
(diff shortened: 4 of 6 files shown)


AenBleidd:
@Flowdalic, could you please test this and confirm that it works in your case?

AenBleidd (approved):
Code itself looks ok to me. Still need to be verified IMHO

Vulpine05:
> Code itself looks ok to me. Still need to be verified IMHO

I am building it right now and was going to test it.  I cannot say I know exactly how to test, but I will at least see if I observe anything different.  I will see what I can do in the next week, but don't wait for me if you are ready to merge.

AenBleidd:
@Vulpine05, we do not rush here. And mostly this need to be tested by @Flowdalic since they have a very special conditions

davidpanderson:
I tested it (not exhaustively) on Win and Linux.  I'm working with Charlie to see if there's a "get total CPU" API on Mac OS as well.

lib/procinfo_unix.cpp
```diff
@@ -263,3 +263,24 @@ int procinfo_setup(PROC_MAP& pm) {
...
+double total_cpu_time() {
+    char buf[1024];
+    static FILE *f=NULL;
+    static double scale;
+    if (!f) {
+        f = fopen("/proc/stat", "r");
+        long hz = sysconf(_SC_CLK_TCK);
+        scale = 1./hz;
+    } else {
+        fflush(f);
```
Flowdalic, line 279: I believe `fflush()` is unnecessary if followed by `rewind()`  (which is basically `fseek()`).

lib/procinfo_unix.cpp
```diff
@@ -263,3 +263,24 @@ int procinfo_setup(PROC_MAP& pm) {
...
+        rewind(f);
+    }
+    if (!fgets(buf, 256, f)) return 0;
```
Flowdalic, line 282: Why not `return -1` to signal the error condition?

lib/procinfo_unix.cpp
```diff
@@ -263,3 +263,24 @@ int procinfo_setup(PROC_MAP& pm) {
...
+    double user, nice;
+    sscanf(buf, "cpu %lf %lf", &user, &nice);
```
Flowdalic, line 284: Missing error handling for `sscanf()` and `user` and `nice` could be declared as `long`.

lib/procinfo.h
```diff
@@ -82,4 +82,10 @@ extern void procinfo_non_boinc(PROCINFO&, PROC_MAP&);
 extern double process_tree_cpu_time(int pid);
     // get the CPU time of the given process and its descendants
 
+extern double total_cpu_time();
+    // total CPU time, as reported by OS
```
Flowdalic, line 86: The comment in `procinfo_unix.cpp` more accurately states that this returns *user-mode* CPU time. Both comments are missing the returned unit, which is, I believe, seconds.

lib/procinfo.cpp
```diff
@@ -135,10 +135,34 @@ void procinfo_non_boinc(PROCINFO& procinfo, PROC_MAP& pm) {
...
+    double sum = 0;
+    PROC_MAP::iterator i;
+    for (i=pm.begin(); i!=pm.end(); ++i) {
+        PROCINFO& p = i->second;
+#ifdef _WIN32
+        if (p.id == 0) continue;    // idle process
+#endif
+        if (
+            p.is_boinc_app
+            || p.is_low_priority
```
Flowdalic, line 157: The following is on indirectly related to this PR.

 `is_low_priority()` currently only does consider the "lowest-priority" nice value (unless I am mistaken). Once could argue that all "low-priority" nice values should be considered.

In any case, `is_low_priority()` should also return true for processes running under `SCHED_IDLE`.

Flowdalic:
Glad to see that there is finally movement towards fixing this long standing issue. Thanks everyone! :)

> And mostly this need to be tested by @Flowdalic since they have a very special conditions

I tend to believe that we do *not* have "very special conditions". We have a bunch of machines for continuous integration and scientific computation. We donated CPU time by running BOINC on those as low-priority batch job, so that it wont affect the primary users. For that we just set it to run as SCHED_IDLE job. This worked perfectly, especially with [kernels >= 5.4 where SCHED_IDLE got significantly improved](https://lwn.net/Articles/805317/). We also set BOINC to stop if the non-BOINC usage was above a threshold. Just to make extra sure that nobody could complain about BOINC draining resources from their jobs.

I believe it is very common for people and/or organization that they are willing to donate CPU time by running BOINC as low-priority batch job on their machines. With the only requirement being that it does not affect normal operations.

What we saw was that BOINC would often not release the resources, because of the non-BOINC usage calculation being flawed. Then I patched BOINC with https://github.com/BOINC/boinc/pull/4336 and https://github.com/BOINC/boinc/pull/4328, which improved the user experience.

Ultimately, we had to to turn off BOINC, because building and distributing a custom BOINC .deb was not longer sustainable. I can not tell when we will re-enabled BOINC again, or if we do so ever. Hence I am unfortunately unable to test this PR.

However, I left some comments in-line. I guess my main point is not directly related to this PR, but `is_low_priority()` should also return `true` for processes that run with the `SCHED_IDLE` scheduling policy.

lib/procinfo_unix.cpp
davidpanderson: without fflush(), fgets() will return the same buffered data

lib/procinfo_unix.cpp
davidpanderson: I added several error checks.  nice is double to avoid 32-bit overflow

lib/procinfo.h
davidpanderson: fixed.  Time in BOINC are always seconds unless otherwise specified

lib/procinfo_unix.cpp
davidpanderson: for now, we'll handle errors by writing an error message and returning 0 (so the limit will have no effect).

lib/procinfo.cpp
davidpanderson: Maybe do this later.  It would require calling sched_getscheduler() for every process, every 10 sec.

CharlieFenton:
It appears to me that your code is looking at total CPU time since the system was booted, rather than the recent CPU times, for both `boinc_related_cpu_time()` and `total_cpu_time()`. I think you need to calculate the equivalent of `(new_cpu_time - last_cpu_time)/diff`for both  in `ACTIVE_TASK_SET::get_memory_usage()`both `boinc_related_cpu_time()` and `total_cpu_time()`, as you do for the older method.

lib/procinfo_unix.cpp
Flowdalic: FWIW I am not totally convinced that this is the case, since fseek() will also flush the data (and rewind() is based on fseek()). At least if my memory serves me right.

davidpanderson:
The code computes the difference between successive totals.
See app.cpp line 495

CharlieFenton:
> The code computes the difference between successive totals.
See app.cpp line 495

@davidpanderson Thanks, I see that now.

PR #4864, opened 2022-08-03
[client] Replace sprintf with snprintf

computezrmle:
Replaces nearly all occurrences of 'sprintf' with 'snprintf' within the sources forming the BOINC client.
Few occurrences are left unchanged as the still return compiler warnings.


AenBleidd (approved):
Looks good to me. Thank you for the contribution!

PR #4865, opened 2022-08-03
Fix missing '-lz' parameter when building 'stage_file_native' tool

AenBleidd:


AenBleidd:
Small change that doesn't require actual review. Just fixing the build

PR #4866, opened 2022-08-04
[vboxwrapper] Replace sprintf with snprintf

computezrmle:
Replaces all occurrences of 'sprintf' with 'snprintf' within the sources forming vboxwrapper.


AenBleidd (approved):
Looks good to me. Thanks for the improvement

PR #4867, opened 2022-08-04
Fix gcc compiler warnings

computezrmle:
... like these:

```
MainDocument.cpp: In member function ‘void CMainDocument::KillGraphicsApp(int)’:
MainDocument.cpp:1715:19: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
 1715 |         argv[0] = "switcher";
      |                   ^~~~~~~~~~
MainDocument.cpp: In function ‘void color_cycle(int, int, wxColour&)’:
MainDocument.cpp:2655:42: warning: ‘b’ may be used uninitialized [-Wmaybe-uninitialized]
 2655 |     unsigned char cb = (unsigned char) (b*256);
```

### DIFF
--- a/clientgui/MainDocument.cpp
+++ b/clientgui/MainDocument.cpp
@@ -1712,10 +1712,10 @@ void CMainDocument::KillGraphicsApp(int pid) {
 
     if (g_use_sandbox) {
         snprintf(thePIDbuf, sizeof(thePIDbuf), "%d", pid);
-        argv[0] = (char *)"switcher";
-        argv[1] = (char *)"/bin/kill";
-        argv[2] = (char *)"kill";
-        argv[3] = (char *)"-KILL";
+        argv[0] = "switcher";
+        argv[1] = "/bin/kill";
+        argv[2] =  "kill";
+        argv[3] = "-KILL";
         argv[4] = thePIDbuf;
         argv[5] = 0;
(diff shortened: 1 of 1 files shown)


clientgui/MainDocument.cpp
```diff
@@ -1712,10 +1712,10 @@ void CMainDocument::KillGraphicsApp(int pid) {
 
     if (g_use_sandbox) {
         snprintf(thePIDbuf, sizeof(thePIDbuf), "%d", pid);
-        argv[0] = "switcher";
-        argv[1] = "/bin/kill";
-        argv[2] =  "kill";
-        argv[3] = "-KILL";
+        argv[0] = (char *)"switcher";
```
AenBleidd, line 1715: This is not a correct change. By default these strings are located in the application static resources and when you convert this to char* that is not constant, and then someone will device to change the value this pointer pointing at, there will be a crash. So I suggest not to make this change at all. Maybe one day we will consider rewriting this using std::string, and this issue will be solved then in a more sustainable way.

AenBleidd (changes requested):

AenBleidd (approved):

AenBleidd:
Looks good to me. Thanks for the fix!

PR #4868, opened 2022-08-04
Fix pragma warning

computezrmle:
Not sure, but shouldn't the file reference point to the header file instead of the cpp file?

```
DlgDiagnosticLogFlags.cpp:19:24: warning: ‘#pragma implementation’ for ‘DlgDiagnosticLogFlags.cpp’ appears after file is included
   19 | #pragma implementation "DlgDiagnosticLogFlags.cpp"
      |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/clientgui/DlgDiagnosticLogFlags.cpp
+++ b/clientgui/DlgDiagnosticLogFlags.cpp
@@ -16,7 +16,7 @@
 // along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 //
 #if defined(__GNUG__) && !defined(__APPLE__)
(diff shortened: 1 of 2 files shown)


AenBleidd:
There is one more file with the same issue: https://github.com/BOINC/boinc/blob/76dffd0dda2ee7a7f881e3dd08c35edec497a504/clientgui/DlgHiddenColumns.cpp#L19
I believe it could be fixed too.

computezrmle:
Done.

AenBleidd (approved):

AenBleidd:
Looks good to me. Thanks for the fix!

Issue #1872, opened 2017-04-18
Computing Preferences window is too big under 1366*768 resolution


truboxl:
![screenshot from 2017-04-18 21-00-33](https://cloud.githubusercontent.com/assets/12658589/25131992/84be3a50-247a-11e7-9ff6-4e38b80f59e1.png)

Fedora 25 Workstation
GNOME 3
BOINC version 7.6.22
wxWidgets version 3.0.2

I don't understand why recent BOINC Managers keep growing the size of that window

AenBleidd:
Looks like some problems with Gnome 3 themes. 
I see no problems on Windows with default text size.
Do you use normal font size?

TruboXL:
![screenshot from 2017-04-18 21-22-38](https://cloud.githubusercontent.com/assets/12658589/25132761/2d65851c-247d-11e7-851f-8b93f2885559.png)

These are the normal defaults on Fedora I guess...

PR #4871, opened 2022-08-07
Add not-in-use computing prefs

### LINKED ISSUE #41: differentiate between number of CPUs to use in idle and busy mode
**Reported by KSMarksPsych on 12 Oct 37296831 05:46 UTC**
Posted by: 
Date: 3:05 PM 04-07-2007
BOINC detects if the machine is busy and can limit the number of CPUs to use. I'd like to see a feature, that I can define 2 numbers of CPUs to use, one if the machine is idle and one if the machine is busy. Right now I've got a dual processor machine and allow BOINC to use one of them so that my work is not slowed down. But when the machine is idle, BOINC could use both CPUs. Right now, this is not possible. With the advent of multi-processor/core machines spreading out to end users, this would allow more CPU power to be used by BOINC and the necessary preconditions are already implented. 

Migrated-From: http://boinc.berkeley.edu/trac/ticket/41

davidpanderson:
Fixes #41 (!)
Previously there were single prefs for
- % of CPUs
- % of CPU time
- suspend if non-BOINC CPU load above X
These applied whether or not the computer was in use.
This PR adds separate prefs for when the computer is not in use.
It includes both web-based and Manager interfaces for editing them.
Backwards compatible; if the new prefs are not specified, it uses the old ones for both cases.

### DIFF
--- a/client/client_state.h
+++ b/client/client_state.h
@@ -18,7 +18,6 @@
 #ifndef BOINC_CLIENT_STATE_H
 #define BOINC_CLIENT_STATE_H
 
-#define NEW_CPU_THROTTLE
 // do CPU throttling using a separate thread.
 // This makes it possible to throttle faster than the client's 1-sec poll period
 // NOTE: we can't actually do this because the runtime system's
@@ -35,9 +34,7 @@ using std::vector;
 
 #include "coproc.h"
 #include "util.h"
-#ifdef NEW_CPU_THROTTLE
 #include "thread.h"
-#endif
 
 #include "acct_mgr.h"
 #include "acct_setup.h"
@@ -345,7 +342,7 @@ struct CLIENT_STATE {
         // - an app fails to start (CS::schedule_cpus())
         // - any project op is done via RPC (suspend/resume)
         // - any result op is done via RPC (suspend/resume)
-    void set_ncpus();
+    void set_n_usable_cpus();
 
 // --------------- cs_account.cpp:
     int add_project(
@@ -363,12 +360,13 @@ struct CLIENT_STATE {
     double get_fraction_done(RESULT* result);
     int input_files_available(RESULT*, bool, FILE_INFO** f=0);
     ACTIVE_TASK* lookup_active_task_by_result(RESULT*);
-    int ncpus;
-        // Act like there are this many CPUs.
+    int n_usable_cpus;
+        // number of usable CPUs
         // By default this is the # of physical CPUs,
         // but it can be changed in two ways:
-        // - type <ncpus>N</ncpus> in the config file
-        // - type the max_ncpus_pct pref
+        // - <ncpus>N</ncpus> in cc_config.xml
+        //      (for debugging; can be > # physical CPUs)
+        // - the max_ncpus_pct and niu_max_ncpus_pct prefs
 
     int latest_version(APP*, char*);
     int app_finished(ACTIVE_TASK&);
@@ -521,6 +519,24 @@ struct CLIENT_STATE {
 #endif
 
     KEYWORDS keywords;
+
+    double current_cpu_usage_limit() {
+        double x = global_prefs.cpu_usage_limit;
+        if (!user_active && global_prefs.niu_cpu_usage_limit>=0) {
+            x = global_prefs.niu_cpu_usage_limit;
+        }
+        if (x < 0.005 || x > 99) {
+            x = 100;
+        }
+        return x;
+    }
+    double current_suspend_cpu_usage() {
+        double x = global_prefs.suspend_cpu_usage;
+        if (!user_active && global_prefs.niu_suspend_cpu_usage>=0) {
+            x = global_prefs.niu_suspend_cpu_usage;
+        }
+        return x;
+    }
 };
 
 extern CLIENT_STATE gstate;
@@ -535,10 +551,8 @@ extern double calculate_exponential_backoff(
     int n, double MIN, double MAX
 );
 
-#ifdef NEW_CPU_THROTTLE
 extern THREAD_LOCK client_mutex;
 extern THREAD throttle_thread;
-#endif
 
 //////// TIME-RELATED CONSTANTS ////////////
 
--- a/client/client_state.h
+++ b/client/client_state.h
@@ -424,6 +424,7 @@ struct CLIENT_STATE {
         const char* fname = GLOBAL_PREFS_FILE_NAME,
         const char* override_fname = GLOBAL_PREFS_OVERRIDE_FILE
     );
+    void print_global_prefs();
     int save_global_prefs(const char* prefs, char* url, char* sched);
     double available_ram();
     double max_available_ram();
@@ -525,7 +526,7 @@ struct CLIENT_STATE {
         if (!user_active && global_prefs.niu_cpu_usage_limit>=0) {
             x = global_prefs.niu_cpu_usage_limit;
         }
-        if (x < 0.005 || x > 99) {
+        if (x < 0.005 || x > 99.99) {
             x = 100;
         }
         return x;
--- a/clientgui/DlgAdvPreferences.cpp
+++ b/clientgui/DlgAdvPreferences.cpp
@@ -672,6 +672,7 @@ void CDlgAdvPreferences::UpdateControlStates() {
     if (m_chkProcInUse->IsChecked()) m_chkGPUProcInUse->SetValue(true);
 
     m_txtMaxLoad->Enable(m_chkMaxLoad->IsChecked());
+    m_txtMaxLoadNotInUse->Enable(m_chkMaxLoadNotInUse->IsChecked());
 
     // ######### disk and memory usage page
     m_txtDiskMaxSpace->Enable(m_chkDiskMaxSpace->IsChecked());
@@ -1066,7 +1067,9 @@ void CDlgAdvPreferences::OnHandleCommandEvent(wxCommandEvent& ev) {
     case ID_CHKMAXLOAD:
         DisplayValue(defaultPrefs.suspend_cpu_usage, m_txtMaxLoad, m_chkMaxLoad);
         break;
-
+    case ID_CHKMAXLOADNOTINUSE:
+        DisplayValue(defaultPrefs.niu_suspend_cpu_usage, m_txtMaxLoadNotInUse, m_chkMaxLoadNotInUse);
+        break;
     // network usage page
     case ID_CHKNETDOWNLOADRATE:
         DisplayValue((defaultPrefs.max_bytes_sec_down / 1024), m_txtNetDownloadRate, m_chkNetDownloadRate);
--- a/clientgui/DlgAdvPreferences.cpp
+++ b/clientgui/DlgAdvPreferences.cpp
@@ -337,7 +337,9 @@ void CDlgAdvPreferences::ReadPreferenceSettings() {
         m_txtProcIdleFor->Disable();
     }
 
+    m_chkNoRecentInput->SetValue(prefs.suspend_if_no_recent_input > 0.0);
     DisplayValue(prefs.suspend_if_no_recent_input, m_txtNoRecentInput);
+
     m_chkMaxLoad->SetValue(prefs.suspend_cpu_usage > 0.0);
     DisplayValue(prefs.suspend_cpu_usage, m_txtMaxLoad, m_chkMaxLoad);
     m_chkMaxLoadNotInUse->SetValue(prefs.niu_suspend_cpu_usage > 0.0);
@@ -479,9 +481,13 @@ bool CDlgAdvPreferences::SavePreferencesSettings() {
         mask.idle_time_to_run=true;
     }
 
-    m_txtNoRecentInput->GetValue().ToDouble(&td);
-    prefs.suspend_if_no_recent_input = RoundToHundredths(td);
-       mask.suspend_if_no_recent_input = true;
+    if (m_chkNoRecentInput->IsChecked()) {
+        m_txtNoRecentInput->GetValue().ToDouble(&td);
+        prefs.suspend_if_no_recent_input = RoundToHundredths(td);
+    } else {
+        prefs.suspend_if_no_recent_input = 0;
+    }
+    mask.suspend_if_no_recent_input = true;
 
     if (m_chkMaxLoad->IsChecked()) {
         m_txtMaxLoad->GetValue().ToDouble(&td);
@@ -673,6 +679,7 @@ void CDlgAdvPreferences::UpdateControlStates() {
 
     m_txtMaxLoad->Enable(m_chkMaxLoad->IsChecked());
     m_txtMaxLoadNotInUse->Enable(m_chkMaxLoadNotInUse->IsChecked());
+    m_txtNoRecentInput->Enable(m_chkNoRecentInput->IsChecked());
 
     // ######### disk and memory usage page
     m_txtDiskMaxSpace->Enable(m_chkDiskMaxSpace->IsChecked());
@@ -737,10 +744,12 @@ bool CDlgAdvPreferences::ValidateInput() {
         }
     }
 
-    buffer = m_txtNoRecentInput->GetValue();
-    if (!IsValidFloatValueBetween(buffer, 0, 9999999999999.99)) {
-        ShowErrorMessage(invMsgFloat, m_txtNoRecentInput);
-        return false;
+    if (m_chkNoRecentInput->IsChecked()) {
+        buffer = m_txtNoRecentInput->GetValue();
+        if (!IsValidFloatValueBetween(buffer, 0, 9999999999999.99)) {
+            ShowErrorMessage(invMsgFloat, m_txtNoRecentInput);
+            return false;
+        }
     }
 
     if (m_chkMaxLoad->IsChecked()) {
@@ -750,6 +759,7 @@ bool CDlgAdvPreferences::ValidateInput() {
             return false;
         }
     }
+
     if (m_chkMaxLoadNotInUse->IsChecked()) {
         buffer = m_txtMaxLoadNotInUse->GetValue();
         if(!IsValidFloatValueBetween(buffer, 1.0, 100.0)) {
@@ -1069,6 +1079,9 @@ void CDlgAdvPreferences::OnHandleCommandEvent(wxCommandEvent& ev) {
         break;
     case ID_CHKMAXLOADNOTINUSE:
         DisplayValue(defaultPrefs.niu_suspend_cpu_usage, m_txtMaxLoadNotInUse, m_chkMaxLoadNotInUse);
+        break;
+    case ID_CHKNORECENTINPUT:
+        DisplayValue(defaultPrefs.suspend_if_no_recent_input, m_txtNoRecentInput, m_chkNoRecentInput);
         break;
     // network usage page
     case ID_CHKNETDOWNLOADRATE:
--- a/clientgui/DlgAdvPreferences.cpp
+++ b/clientgui/DlgAdvPreferences.cpp
@@ -131,7 +131,9 @@ void CDlgAdvPreferences::SetValidators() {
 
     // ######### proc usage page
     m_txtProcUseProcessors->SetValidator(wxTextValidator(wxFILTER_NUMERIC));
+    m_txtProcUseProcessorsNotInUse->SetValidator(wxTextValidator(wxFILTER_NUMERIC));
     m_txtProcUseCPUTime->SetValidator(wxTextValidator(wxFILTER_NUMERIC));
+    m_txtProcUseCPUTimeNotInUse->SetValidator(wxTextValidator(wxFILTER_NUMERIC));
 
     m_txtProcIdleFor->SetValidator(wxTextValidator(wxFILTER_NUMERIC));
     m_txtProcIdleFor->SetMaxLength(16);
@@ -276,7 +278,8 @@ void CDlgAdvPreferences::DisplayValue(double value, wxTextCtrl* textCtrl, wxChec
 }
 
 
-/* read preferences from core client and initialize control values */
+// read preferences from core client and initialize control values
+//
 void CDlgAdvPreferences::ReadPreferenceSettings() {
     CMainDocument* pDoc = wxGetApp().GetDocument();
     int retval;
@@ -303,11 +306,13 @@ void CDlgAdvPreferences::ReadPreferenceSettings() {
     // 0 means "no restriction" but we don't use a checkbox here
     if (prefs.max_ncpus_pct == 0.0) prefs.max_ncpus_pct = 100.0;
     DisplayValue(prefs.max_ncpus_pct, m_txtProcUseProcessors);
+    DisplayValue(prefs.niu_max_ncpus_pct, m_txtProcUseProcessorsNotInUse);
 
-            //cpu limit
+    // cpu limit
     // 0 means "no restriction" but we don't use a checkbox here
     if (prefs.cpu_usage_limit == 0.0) prefs.cpu_usage_limit = 100.0;
     DisplayValue(prefs.cpu_usage_limit, m_txtProcUseCPUTime);
+    DisplayValue(prefs.niu_cpu_usage_limit, m_txtProcUseCPUTimeNotInUse);
 
     // on batteries
     m_chkProcOnBatteries->SetValue(! prefs.run_on_batteries);
@@ -335,6 +340,8 @@ void CDlgAdvPreferences::ReadPreferenceSettings() {
     DisplayValue(prefs.suspend_if_no_recent_input, m_txtNoRecentInput);
     m_chkMaxLoad->SetValue(prefs.suspend_cpu_usage > 0.0);
     DisplayValue(prefs.suspend_cpu_usage, m_txtMaxLoad, m_chkMaxLoad);
+    m_chkMaxLoadNotInUse->SetValue(prefs.niu_suspend_cpu_usage > 0.0);
+    DisplayValue(prefs.niu_suspend_cpu_usage, m_txtMaxLoadNotInUse, m_chkMaxLoadNotInUse);
 
     // connection interval
     DisplayValue(prefs.work_buf_min_days, m_txtNetConnectInterval);
@@ -434,8 +441,9 @@ void CDlgAdvPreferences::ReadPreferenceSettings() {
     this->UpdateControlStates();
 }
 
-/* write overridden preferences to disk (global_prefs_override.xml) */
-/* IMPORTANT: Any items added here must be checked in ValidateInput()! */
+// write overridden preferences to disk (global_prefs_override.xml)
+// IMPORTANT: Any items added here must be checked in ValidateInput()!
+//
 bool CDlgAdvPreferences::SavePreferencesSettings() {
     double td;
 
@@ -445,11 +453,16 @@ bool CDlgAdvPreferences::SavePreferencesSettings() {
     m_txtProcUseProcessors->GetValue().ToDouble(&td);
     prefs.max_ncpus_pct = RoundToHundredths(td);
     mask.max_ncpus_pct=true;
+    m_txtProcUseProcessorsNotInUse->GetValue().ToDouble(&td);
+    prefs.niu_max_ncpus_pct = RoundToHundredths(td);
+    mask.niu_max_ncpus_pct=true;
 
-    //
     m_txtProcUseCPUTime->GetValue().ToDouble(&td);
     prefs.cpu_usage_limit=RoundToHundredths(td);
     mask.cpu_usage_limit=true;
+    m_txtProcUseCPUTimeNotInUse->GetValue().ToDouble(&td);
+    prefs.niu_cpu_usage_limit = RoundToHundredths(td);
+    mask.niu_cpu_usage_limit = true;
 
     prefs.run_on_batteries = ! (m_chkProcOnBatteries->GetValue());
     mask.run_on_batteries=true;
@@ -466,12 +479,10 @@ bool CDlgAdvPreferences::SavePreferencesSettings() {
         mask.idle_time_to_run=true;
     }
 
-    //
     m_txtNoRecentInput->GetValue().ToDouble(&td);
     prefs.suspend_if_no_recent_input = RoundToHundredths(td);
        mask.suspend_if_no_recent_input = true;
 
-    //
     if (m_chkMaxLoad->IsChecked()) {
         m_txtMaxLoad->GetValue().ToDouble(&td);
         prefs.suspend_cpu_usage=RoundToHundredths(td);
@@ -479,6 +490,14 @@ bool CDlgAdvPreferences::SavePreferencesSettings() {
         prefs.suspend_cpu_usage = 0.0;
     }
     mask.suspend_cpu_usage=true;
+
+    if (m_chkMaxLoadNotInUse->IsChecked()) {
+        m_txtMaxLoadNotInUse->GetValue().ToDouble(&td);
+        prefs.niu_suspend_cpu_usage=RoundToHundredths(td);
+    } else {
+        prefs.niu_suspend_cpu_usage = 0.0;
+    }
+    mask.niu_suspend_cpu_usage=true;
 
     m_txtNetConnectInterval->GetValue().ToDouble(&td);
     prefs.work_buf_min_days=RoundToHundredths(td);
@@ -697,10 +716,15 @@ bool CDlgAdvPreferences::ValidateInput() {
         ShowErrorMessage(invMsgLimit100, m_txtProcUseProcessors);
         return false;
     }
-
-    buffer = m_txtProcUseCPUTime->GetValue();
+    buffer = m_txtProcUseProcessorsNotInUse->GetValue();
     if(!IsValidFloatValueBetween(buffer, 0.0, 100.0)) {
-        ShowErrorMessage(invMsgLimit100, m_txtProcUseCPUTime);
+        ShowErrorMessage(invMsgLimit100, m_txtProcUseProcessorsNotInUse);
+        return false;
+    }
+
+    buffer = m_txtProcUseCPUTimeNotInUse->GetValue();
+    if(!IsValidFloatValueBetween(buffer, 0.0, 100.0)) {
+        ShowErrorMessage(invMsgLimit100, m_txtProcUseCPUTimeNotInUse);
         return false;
     }
 
@@ -725,8 +749,15 @@ bool CDlgAdvPreferences::ValidateInput() {
             return false;
         }
     }
+    if (m_chkMaxLoadNotInUse->IsChecked()) {
+        buffer = m_txtMaxLoadNotInUse->GetValue();
+        if(!IsValidFloatValueBetween(buffer, 1.0, 100.0)) {
+            ShowErrorMessage(invMsgLimit1_100, m_txtMaxLoadNotInUse);
+            return false;
+        }
+    }
 
-    //limit additional days from 0 to 10
+    // limit additional days from 0 to 10
     buffer = m_txtNetConnectInterval->GetValue();
     if(!IsValidFloatValueBetween(buffer, 0.0, 10.0)) {
         ShowErrorMessage(invMsgLimit10,m_txtNetConnectInterval);
--- a/clientgui/DlgAdvPreferences.cpp
+++ b/clientgui/DlgAdvPreferences.cpp
@@ -338,7 +338,7 @@ void CDlgAdvPreferences::ReadPreferenceSettings() {
     }
 
     m_chkNoRecentInput->SetValue(prefs.suspend_if_no_recent_input > 0.0);
-    DisplayValue(prefs.suspend_if_no_recent_input, m_txtNoRecentInput);
+    DisplayValue(prefs.suspend_if_no_recent_input, m_txtNoRecentInput, m_chkNoRecentInput);
 
     m_chkMaxLoad->SetValue(prefs.suspend_cpu_usage > 0.0);
     DisplayValue(prefs.suspend_cpu_usage, m_txtMaxLoad, m_chkMaxLoad);
--- a/clientgui/DlgAdvPreferences.cpp
+++ b/clientgui/DlgAdvPreferences.cpp
@@ -277,6 +277,27 @@ void CDlgAdvPreferences::DisplayValue(double value, wxTextCtrl* textCtrl, wxChec
     textCtrl->Enable();
 }
 
+void CDlgAdvPreferences::EnableDisableInUseItem(wxTextCtrl* textCtrl, bool doEnable, double value) {
+    if (doEnable) {
+        if (! textCtrl->IsEnabled()) {
+            textCtrl->Enable();
+            DisplayValue(value, textCtrl);
+        }
+    } else {
+        textCtrl->Clear();
+        textCtrl->Disable();
+    }
+}
+
+void CDlgAdvPreferences::EnableDisableInUseItems() {
+    bool doEnable = !(m_chkProcInUse->IsChecked());
+    EnableDisableInUseItem(m_txtProcUseProcessors, doEnable, 
+                            defaultPrefs.max_ncpus_pct > 0.0 ? defaultPrefs.max_ncpus_pct : 100.0);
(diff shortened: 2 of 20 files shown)


clientgui/DlgAdvPreferencesBase.cpp
```diff
@@ -250,233 +249,296 @@ wxPanel* CDlgAdvPreferencesBase::createProcessorTab(wxNotebook* notebook)
...
+    m_txtProcUseProcessorsNotInUse = new wxTextCtrl(box, ID_TXTPROCUSEPROCESSORSNOTINUSE, wxEmptyString, wxDefaultPosition, textCtrlSize, wxTE_RIGHT);
+    /*xgettext:no-c-format*/
+    wxStaticText* staticText21niu = new wxStaticText(box, ID_DEFAULT, _("% of the CPUs"), wxDefaultPosition, wxDefaultSize, 0);
+    addNewRowToSizer(box_sizer, MaxCPUPctTTniu, staticText20niu, m_txtProcUseProcessorsNotInUse, staticText21niu);
 
-    wxStaticText* staticText27 = new wxStaticText(
-        suspendComputingStaticBox, ID_DEFAULT,
-        _("Suspend when no mouse/keyboard input in last"),
+    // CPU throttling
+    //
```
CharlieFenton, line 374: Precede line 375 with ` /*xgettext:no-c-format*/`

AenBleidd:
On Windows it's failing on this:
```
         sim.cpp
    27>D:\a\boinc\boinc\client\sim.cpp(302,24): error C2065: 'ncpus': undeclared identifier [D:\a\boinc\boinc\win_build\sim_vs2019.vcxproj]
    27>D:\a\boinc\boinc\client\sim.cpp(350,24): warning C4459: declaration of 'delta' hides global declaration [D:\a\boinc\boinc\win_build\sim_vs2019.vcxproj]
       D:\a\boinc\boinc\client\sim.cpp(91,26): message : see declaration of 'delta' [D:\a\boinc\boinc\win_build\sim_vs2019.vcxproj]
    27>D:\a\boinc\boinc\client\sim.cpp(405,26): warning C4458: declaration of 'apps' hides class member [D:\a\boinc\boinc\win_build\sim_vs2019.vcxproj]
       D:\a\boinc\boinc\client\client_state.h(75,18): message : see declaration of 'CLIENT_STATE::apps' [D:\a\boinc\boinc\win_build\sim_vs2019.vcxproj]
    27>D:\a\boinc\boinc\client\sim.cpp(415,36): error C2065: 'ncpus': undeclared identifier [D:\a\boinc\boinc\win_build\sim_vs2019.vcxproj]
    27>D:\a\boinc\boinc\client\sim.cpp(456,32): warning C4477: '_snprintf' : format string '%lu' requires an argument of type 'unsigned long', but variadic argument 1 has type 'unsigned __int64' [D:\a\boinc\boinc\win_build\sim_vs2019.vcxproj]
       D:\a\boinc\boinc\client\sim.cpp(456,32): message : consider using '%zu' in the format string [D:\a\boinc\boinc\win_build\sim_vs2019.vcxproj]
    27>D:\a\boinc\boinc\client\sim.cpp(1498,12): error C2039: 'set_ncpus': is not a member of 'CLIENT_STATE' [D:\a\boinc\boinc\win_build\sim_vs2019.vcxproj]
       D:\a\boinc\boinc\client\client_state.h(71): message : see declaration of 'CLIENT_STATE' [D:\a\boinc\boinc\win_build\sim_vs2019.vcxproj]
 ```
 
 On linux it's failing on:
 ```
 DlgAdvPreferencesBase.cpp: In member function ‘wxPanel* CDlgAdvPreferencesBase::createProcessorTab(wxNotebook*)’:
DlgAdvPreferencesBase.cpp:335:9: error: cannot bind non-const lvalue reference of type ‘wxString&’ to an rvalue of type ‘wxString’
  335 |         wxString (_("Suspend computing when your computer is busy running other programs.")),
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
DlgAdvPreferencesBase.cpp:205:45: note:   initializing argument 2 of ‘void CDlgAdvPreferencesBase::addNewRowToSizer(wxSizer*, wxString&, wxWindow*, wxWindow*, wxWindow*, wxWindow*, wxWindow*)’
  205 |                 wxSizer* toSizer, wxString& toolTipText,
      |                                   ~~~~~~~~~~^~~~~~~~~~~
make[2]: *** [Makefile:1158: boincmgr-DlgAdvPreferencesBase.o] Error 1
```

clientgui/DlgAdvPreferencesBase.cpp
```diff
@@ -250,233 +249,296 @@ wxPanel* CDlgAdvPreferencesBase::createProcessorTab(wxNotebook* notebook)
...
+    m_chkMaxLoad = new wxCheckBox(
+        box, ID_CHKMAXLOAD,
+        wxString(_("Suspend when non-BOINC CPU usage is above")),
         wxDefaultPosition, wxDefaultSize, 0
     );
+    m_txtMaxLoad = new wxTextCtrl(
+        box, ID_TXTMAXLOAD, wxEmptyString, wxDefaultPosition, getTextCtrlSize(wxT("100.00")), wxTE_RIGHT
+    );
+    addNewRowToSizer(box_sizer,
```
CharlieFenton, line 334: To eliminate the compiler syntax error, you must create a separate variable with this string thus:
`    wxString SuspendOnBusyTT(_("Suspend computing when your computer is busy running other programs."));`
and pass `SuspendOnBusyTT ` in the call.

CharlieFenton (commented):
@AenBleidd This is a substantial amount of added and changed code. I will try to review this within the next week.

AenBleidd:
@CharlieFenton, I completely agree with you. I don't push you to make it just immediately. But this change is quite significant so I believe your review is essential.

RichardHaselgrove:
It would reassure me while testing if the scheduler at SETI@home (or another project with the provisional web selection tools) could be re-activated so that I can test the automatic propagation of global preferences from project to project (propagation considers the time and project of the most recent update).

davidpanderson:
I fixed the compile error.  Oddly, it wasn't an error in VS

davidpanderson:
I fixed the sim build error

davidpanderson:
I put the new web code on Science United.

Vulpine05 (changes requested):

client/client_state.h
```diff
@@ -521,6 +519,24 @@ struct CLIENT_STATE {
 #endif
 
     KEYWORDS keywords;
+
+    double current_cpu_usage_limit() {
```
Vulpine05, line 523: Should this function be moved to the .cpp file?

client/client_state.h
```diff
@@ -521,6 +519,24 @@ struct CLIENT_STATE {
...
+        double x = global_prefs.cpu_usage_limit;
+        if (!user_active && global_prefs.niu_cpu_usage_limit>=0) {
+            x = global_prefs.niu_cpu_usage_limit;
+        }
+        if (x < 0.005 || x > 99) {
+            x = 100;
+        }
+        return x;
+    }
+    double current_suspend_cpu_usage() {
```
Vulpine05, line 533: Should this also be moved to the cpp file?

davidpanderson: same

client/client_state.h
davidpanderson: I like to put short functions in the .h file.

lib/prefs.cpp
```diff
@@ -246,7 +255,7 @@ void GLOBAL_PREFS::defaults() {
 #else
     suspend_cpu_usage = 25;
 #endif
-    suspend_if_no_recent_input = 0;
+    suspend_if_no_recent_input = 60;
```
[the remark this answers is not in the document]
davidpanderson: I think this works as intended.

client/client_state.h
```diff
@@ -521,6 +519,24 @@ struct CLIENT_STATE {
 #endif
 
     KEYWORDS keywords;
+
+    double current_cpu_usage_limit() {
+        double x = global_prefs.cpu_usage_limit;
+        if (!user_active && global_prefs.niu_cpu_usage_limit>=0) {
+            x = global_prefs.niu_cpu_usage_limit;
+        }
+        if (x < 0.005 || x > 99) {
```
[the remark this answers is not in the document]
davidpanderson: I'll change this

CharlieFenton:
@davidpanderson Since you are still making changes, please let me know when you are done so I can review the code in its final form.

clientgui/DlgAdvPreferencesBase.cpp
CharlieFenton: @davidpanderson You have not fixed this.  For an explanation why this is needed, see [here](https://stackoverflow.com/questions/70490036/xgettext-incorrectly-infers-c-format-in-cpp-files) and [here](https://www.mit.edu/afs.new/athena/astaff/source/src-9.4/third/gettext2/gettext-tools/doc/gettext_3.html#SEC18).

(Of course the `/*xgettext:no-c-format*/` directive must now precede line 376 in the latest version of this file.)

CharlieFenton (changes requested):
Before even testing whether the new preferences work, I see 2 show stopper bugs:

[1] The Computing Preferences dialog is too tall to fit  on many computer screens. On a Mac, it has a size of 572 W X 864 H. Apple's Retina display uses 4X4 physical pixels per logical pixel, so although a 13-inch MacBook Pro has a screen resolution of 2560 x 1600 physical pixels, a window of 1280 X 800 pixels will completely fill the screen. In addition, the menubar takes up the top 24 logical pixels of the screen, leaving a usable height of only 776 logical pixels.

Assuming the dialog height is similar on Windows and Linux, those platforms will have  the same problem. According to [this](https://www.hobo-web.co.uk/best-screen-size/) article from July 2022:
> Most Common Desktop Screen Resolution Sizes Worldwide:
1920×1080 – 22.97%
1366×768 – 17.88%
1536×864 – 11.34%
1280×720 – 6.18%
1440×900 – 5.85%
1600×900 – 3.58%

> Most Common Desktop Screen Resolution Size in the United States Of America
1920×1080 – 23.8%
1366×768 – 10.63%
1536×864 – 8.61%
1440×900 – 7.97%
1280×720 – 5.04%
1600×900 – 3.71%

(I assume they include laptops under desktop computers in these statistics.)

[2] All items in "When computer is not in use" section should be disabled if "Suspend when computer is in use" is checked, as they do not apply. Their purpose is confusing if they appear active with that checked.

I will resume looking at this PR after these problems are corrected.

AenBleidd:
This window is too big even now on Gnome3: #1872

Maybe we should brainstorm this to redesign Preferences window completely? Since we're not planning a new release next few months, I could find some time to work on this topic.

RichardHaselgrove:
May I add the resolutions of two of my screens:

1920 x 1200
1600 x 1200

They're desktop IPS screens, marketed under the 'DELL Ultrasharp' brand, so not entirely niche. Obviously, the increased height means they are less affected by this issue - but we must remember to design for different aspect ratios, as well as different pixel counts. I've kept an old CRT multisync monitor in the museum to test the edge cases...

davidpanderson:
I can trim 5 lines from the dialog and tighten up the spacing a little; that should restore it to the previous size

davidpanderson:
I shortened the dialog quite a bit;
it's just slightly longer than the old one.

I'm not going to selectively disable the in-use items for now.
We could eventually do this,
but the issue existed before this PR.

clientgui/DlgAdvPreferencesBase.cpp
```diff
@@ -250,234 +279,335 @@ wxPanel* CDlgAdvPreferencesBase::createProcessorTab(wxNotebook* notebook)
...
-
-    m_txtProcIdleFor = new wxTextCtrl(
-        suspendComputingStaticBox, ID_TXTPROCIDLEFOR, wxEmptyString, wxDefaultPosition, getTextCtrlSize(wxT("999.99")), wxTE_RIGHT
+    // max # CPUs and throttling
+    //
+    wxStaticText* staticText20 = new wxStaticText(
+        box, ID_DEFAULT, _("Use at most"), wxDefaultPosition, wxDefaultSize, 0
+    );
+    m_txtProcUseProcessors = new wxTextCtrl(box, ID_TXTPROCUSEPROCESSORS, wxEmptyString, wxDefaultPosition, textCtrlSize, wxTE_RIGHT );
+    wxStaticText* staticText22 = new wxStaticText(
```
CharlieFenton, line 336: Precede line 336 with `/*xgettext:no-c-format*/`. For an explanation why this is needed, see [here](https://stackoverflow.com/questions/70490036/xgettext-incorrectly-infers-c-format-in-cpp-files) and [here](https://www.mit.edu/afs.new/athena/astaff/source/src-9.4/third/gettext2/gettext-tools/doc/gettext_3.html#SEC18).

CharlieFenton:
> I'm not going to selectively disable the in-use items for now.
We could eventually do this,
but the issue existed before this PR.

Before this PR, there were all items in this dialog still applied whether or not the "Suspend when computer is in use" checkbox was set. While I recognize that doing this requires significant additional effort, I cannot approve this PR without this because good UI practice dictates it and the dialog is confusing without it. 

I will resume reviewing this PR when the 2 changes I requested have been completed.

davidpanderson:
Actually there were separate memory limits for in use and not in use.
The in-use setting was not grayed out if suspend-while-in-use was set.

I think it's fine the way it is.  Not confusing.

CharlieFenton:
> Actually there were separate memory limits for in use and not in use.
The in-use setting was not grayed out if suspend-while-in-use was set.

OK, I see that now. But they were not on the same tab / page as the checkbox in question. That was definitely an oversight which we should have fixed earlier. It is good that we can fix it now.

CharlieFenton:
Now that the items in the Advanced Prefs dialog ave been rearranged and in some cases two items have been combined into one, they no longer match the arrangement on the new web prefs forms. Is that a problem?

Backward compatibility questions:
 * How will the new web prefs interact with older versions of the BOINC Manager, and how will the new BOINC Manager prefs dialog interact with older web prefs on projects that are slow to upgrade their web interface? 
 * How will the changes affect the migration of updated web preferences between projects that have and have not updated their web interface?
 * How will a new version of BOINC Manager work with a remote older version of the client, and vice versa?
 * If a user installs a new version of BOINC Manager and uses the new Advanced Prefs dialog, then reverts to an older version of BOINC, how will the client handle the unrecognized XML tags in global_prefs.xml?

RichardHaselgrove:
> Backward compatibility questions:

To which add:

- compatibility with third-party implementations, like Einstein's Drupal

There's still no scheduler at SETI, so I can't even start testing:

13/08/2022 11:53:23 | SETI@home | Scheduler request failed: HTTP file not found

Vulpine05:
> This window is too big even now on Gnome3: #1872
> 
> Maybe we should brainstorm this to redesign Preferences window completely? Since we're not planning a new release next few months, I could find some time to work on this topic.

I know in AutoCAD, one of their setting windows has left and right arrows to "flip" through settings on a tab.  That could probably be done under the compute tab to cycle between in use, not in use, and general?

CharlieFenton:
The [help page](https://boinc.berkeley.edu/wiki/Preferences) for this dialog will also need to be updated. A recent message from the boinc_projects email list also reminded me that this has not been kept up to date; there is no mention of the _Suspend when no mouse/keyboard input in last xx minutes_ item on the current Advanced Computing Preferences dialog.

CharlieFenton:
> I know in AutoCAD, one of their setting windows has left and right arrows to "flip" through settings on a tab. That could probably be done under the compute tab to cycle between in use, not in use, and general?

A simpler and probably clearer option would be to make the dialog wider and put the "when in use" and "when not  in use" sections side by side, with a divider line between them. This just requires adding one more `boxsizer` to the code.

Vulpine05:
> A simpler and probably clearer option would be to make the dialog wider and put the "when in use" and "when not  in use" sections side by side, with a divider line between them. This just requires adding one more `boxsizer` to the code.

That sounds better to me!

CharlieFenton:
@davidpanderson I got tired of waiting for you so I went ahead and updated the code to disable those items which don't apply when _Suspend when computer is in use_ is set.

To be consistent with the existing code, their values are restored to defaults when they are again enabled. I think it would be better if all items which can selectively be disabled (based on the state of a checkbox) were restored to their last saved values when re-enabled rather than the defaults. Unfortunately, that is beyond the scope of this PR.

I also added the missing `/*xgettext:no-c-format*/`.

CharlieFenton (commented):
I will continue testing and reviewing this PR after the issue is fixed of the dialog still not fitting on the GNOME3 screen. Please see my suggested fix [here](https://github.com/BOINC/boinc/pull/4871#issuecomment-1214249201). Also the The [help page](https://boinc.berkeley.edu/wiki/Preferences) for this dialog will also need to be updated and [my](https://github.com/BOINC/boinc/pull/4871#issuecomment-1214126019) and [Richard's](https://github.com/BOINC/boinc/pull/4871#issuecomment-1214137714) compatibility questions need to be addressed.

RichardHaselgrove:
> [Richard's](https://github.com/BOINC/boinc/pull/4871#issuecomment-1214137714) compatibility questions need to be addressed.

I'll write to Eric Korpela and Jeff Cobb tomorrow and see if they can help with the missing scheduler.

davidpanderson:
We're not going to revive SETI@home to test a web feature.
The feature is deployed on Science United and can be tested there.

davidpanderson:
What is the pixel height of the dialog on the problem system?  What is the screen size?

davidpanderson:
AFAIK there are no compatibility issues.
If the new client gets a set of prefs from a project that doesn't include the not-in-use prefs,
it will use the corresponding in-use prefs,
i.e. it will behave like the old client.

If an old client gets a set of web prefs that includes the new prefs,
it will ignore them.
(the web GUI has a warning to this effect)

RichardHaselgrove:
> We're not going to revive SETI@home

I'm not asking for the project to be revived. I'm asking for a scheduler to to be in place to respond to random requests. That would have the useful side effect of enforcing the 24 hour delay that was put in place after the hibernation - reducing Berkeley's bandwidth costs.

CharlieFenton:
@davidpanderson: On the subject of compatibility: My understanding is that with the current Advanced Computing Preferences Dialog (as of BOINC 7.20.2) these items always apply both when the computer is not in use, and if the _suspend when not in use_ checkbox is **not** set they also apply when the computer is in use:
* Use at most xx% of the CPUs - `max_ncpus_pct`
* Use at most xx% ofCPU time - `cpu_usage_limit`
* Suspend when non-BOINC CPU usage is above xx% - `suspend_cpu_usage`

Therefore, when a user installs a future version of BOINC to replace 7.20.2, BOINC should propagate these items to the new "when not in use" prefs, to preserve the behavior the user has previously requested without requiring her to run the Advanced Computing Preferences Dialog after upgrading BOINC: 
* Use at most xx% of the CPUs - `niu_max_ncpus_pct`
* Use at most xx% ofCPU time - `niu_cpu_usage_limit`
* Suspend when non-BOINC CPU usage is above xx% - `niu_suspend_cpu_usage`

Please make sure that this automatic propagation is in the code for this PR.

clientgui/DlgAdvPreferences.cpp
```diff
@@ -303,11 +327,13 @@ void CDlgAdvPreferences::ReadPreferenceSettings() {
     // 0 means "no restriction" but we don't use a checkbox here
     if (prefs.max_ncpus_pct == 0.0) prefs.max_ncpus_pct = 100.0;
     DisplayValue(prefs.max_ncpus_pct, m_txtProcUseProcessors);
+    DisplayValue(prefs.niu_max_ncpus_pct, m_txtProcUseProcessorsNotInUse);
```
CharlieFenton, line 331: missing: `if (prefs.niu_max_ncpus_pct == 0.0) prefs.niu_max_ncpus_pct = 100.0;`

clientgui/DlgAdvPreferences.cpp
```diff
@@ -303,11 +327,13 @@ void CDlgAdvPreferences::ReadPreferenceSettings() {
...
 
-            //cpu limit
+    // cpu limit
     // 0 means "no restriction" but we don't use a checkbox here
     if (prefs.cpu_usage_limit == 0.0) prefs.cpu_usage_limit = 100.0;
     DisplayValue(prefs.cpu_usage_limit, m_txtProcUseCPUTime);
+    DisplayValue(prefs.niu_cpu_usage_limit, m_txtProcUseCPUTimeNotInUse);
```
CharlieFenton, line 337: Missing: `if (prefs.niu_cpu_usage_limit == 0.0) prefs.niu_cpu_usage_limit = 100.0;`

CharlieFenton:
> Therefore, when a user installs a future version of BOINC to replace 7.20.2, BOINC should propagate these items to the new "when not in use" prefs, to preserve the behavior the user has previously requested without requiring her to run the Advanced Computing Preferences Dialog after upgrading BOINC:
 o Use at most xx% of the CPUs - niu_max_ncpus_pct
 o Use at most xx% ofCPU time - niu_cpu_usage_limit
 o Suspend when non-BOINC CPU usage is above xx% - niu_suspend_cpu_usage

This also applies to how the server code handles web prefs:
> How will the changes affect the migration of updated web preferences between projects that have and have not updated their web interface?

When web prefs are updated on a project without the new not-in-use preference items and then propagated to a project with the new items, the new not_in_use items must be set to the corresponding when-in-use items. To test this, I changed the values of _Use at most xx% of the CPUs_ and _Use at most xx% ofCPU time_  on Einstein@home, updated that project on BOINC 7.20.2, then updated SETI@home (also on BOINC 7.20.2.) The _sched_request_setiathome.berkeley.edu.xml_ file sent the updated values for _cpu_usage_limit_ and _max_ncpus_pct_ to the S@h server, but the _sched_reply_setiathome.berkeley.edu.xml_ shows a 404 error _The requested URL was not found on this server._.

Please confirm that the server code  propagates when-in-use values to the new not-in-use values both in this situation and also when a project's server code is updated. Note also that the propagation of these values when the server code is  updated must not interfere with a later update from a client (i.e., it must not update the modification time.)

CharlieFenton:
Consider this sequence of events. A user has an old _global_prefs_override.xml_ file because he has changed a value unrelated to those covered by this PR. He updates the new web prefs on a project website to set values for the new not-in-use settings different from the when-in-use settings (which he leaves unchanged.) But he does not run the Advanced Computing Prefs dialog and so does not update the Prefs Override file because it does not seem relevant. Line 403 of prefs.cpp will then continue to use the when-in-use settings when not in use, which is not what the user expected.

RichardHaselgrove:
> the _sched_reply_setiathome.berkeley.edu.xml_ shows a 404 error

I wrote to Eric and Jeff yesterday, but as yet haven't received a reply.

CharlieFenton:
I will be occupied with other projects until Thursday (18 August 2022.) I will return to this PR after that.

CharlieFenton:
> The feature is deployed on Science United and can be tested there.

@davidpanderson: I created an account on Science United. But because it is an account manager it does not let me do [this](https://github.com/BOINC/boinc/pull/4871#issuecomment-1216352004) test. I agree with @RichardHaselgrove that you need to enable the S@h scheduler to allow proper testing.

Also, I think it is a bad idea to implement this untested development code on an active live project or (in this case) account manager such as Science United.

RichardHaselgrove:
Many other projects down the years have implemented parallel test projects to facilitate this sort of testing. If you can find one that's still live, and will let you deploy test code remotely, that might help.

There's even SETI Beta, but that's even deader ...

16/08/2022 11:16:12 | SETI@home Beta Test | [sched_op] Fetching master file
16/08/2022 11:16:13 | SETI@home Beta Test | [sched_op] Reason: Master file request failed

Vulpine05:
Instead of asking a project to test the changes with their server, is there a way to create a test server and a test project?  Or is that not possible?

davidpanderson:
It's up to project admins to decide what web code to run.  I've tested on Science United and everything works.  I encourage others to do same.

CharlieFenton:
> It's up to project admins to decide what web code to run.

I'm unclear which comments or concerns this is intended to address. Obviously, we can only test the "standard" web code we will provide. 

> I've tested on Science United and everything works.

My experience attaching to Science United was not so great. It ignored my existing global prefs and replaced them with default values. Since I've never attached to Science United before, I don't know if this is a new problem associated with this PR.

As I wrote [here](https://github.com/BOINC/boinc/pull/4871#issuecomment-1216432839), testing with an account manager does not allow us to try many of the situations we feel needs to be tested. It's great that you have tested it (as one would expect of anyone writing new code); but this PR involves extensive changes and so it needs testing by as many people as possible and in as many scenarios as possible, especially the edge cases which could easily be overlooked.

As just one example, I don't see you having addressed either [this](https://github.com/BOINC/boinc/pull/4871#pullrequestreview-1072436875) or [this](https://github.com/BOINC/boinc/pull/4871#pullrequestreview-1072436906), either by adding the missing code or explaining why it should not be added.

davidpanderson:
From the point of view of prefs, an account manager is the same as a project.

clientgui/DlgAdvPreferences.cpp
davidpanderson: not needed - they're initialized in GLOBAL_PREFS::parse_override()

CharlieFenton:
> What is the pixel height of the dialog on the problem system? What is the screen size?

The [report](https://github.com/BOINC/boinc/pull/4871#issuecomment-1214157173) referenced issue #1872 which says the screen size is 1366 pixels wide by 768 pixels high, but the screen shot shows that the menu bar takes up part of that heigh, so the dialog needs to be be shorter than that.

CharlieFenton:
@davidpanderson Thank you for setting up the test project to allow us to test the code for this PR. So far, I have found the following problems:

[1] The test project had my old web prefs before you updated the server code. But when you updated the server code, it set the new "When computer is not in use" values to the new default. It should have propagated the old "When computer is in use" values to the new fields, for consistency with the way the client code works and with what the user would expect.

[2] Changes made to web computing prefs on another project (I used Einstein@home) never propagate to the test project, even though I updated the projects from the client in the correct sequence. This was true whether using BOINC 7.20.2 or the new BOINC client and Manager built from this PR.

[3] I commented [here](https://github.com/BOINC/boinc/pull/4871#pullrequestreview-1072436875) that a line of code seemed to be missing: `if (prefs.niu_max_ncpus_pct == 0.0) prefs.niu_max_ncpus_pct = 100.0;` and [here](https://github.com/BOINC/boinc/pull/4871#pullrequestreview-1072436906) that this seemed to be missing: `if (prefs.niu_cpu_usage_limit == 0.0) prefs.niu_cpu_usage_limit = 100.0;`. You responded:
> not needed - they're initialized in GLOBAL_PREFS::parse_override()

However, the purpose of the corresponding code (from BOINC V7.20.2 and earlier) is to ensure that the dialog displays a 0 value as 100%, not to fix how it is handled internally by the code. Because these lines are missing for the new prefs, the dialog displays the new values as 0 rather than 100 when the dialog is saved and again opened. This is inconsistent with the way the dialog displays 0 values for the corresponding `prefs.max_ncpus_pct` and `prefs.cpu_usage_limit`.

[4] If the user enters 0 for `prefs.niu_max_ncpus_pct` it is set to the value of `prefs.max_ncpus_pct` rather than to 0 or 100.

Because of problem [2], I have not been able to complete my testing. Please let us know when you have fixed these problems (as well as [fitting the dialog to the GNOME3 screen](https://github.com/BOINC/boinc/pull/4871#issuecomment-1220315293)) so that Richard and I can continue testing.

RichardHaselgrove:
I confirm point [2] above: changes made at Einstein are not imported by the test project, although other projects do import them successfully. I attach a global_prefs.xml file (renamed) generated by a v7.20.2 client following an Einstein update: the ordering of the values is completely different, because they use a full XML handler. @ChristianBeer (when he was working at Einstein) told me that in Drupal they had little control over the sequencing of the elements, because the file was written as a single complete entity.

[global_prefs.txt](https://github.com/BOINC/boinc/files/9381192/global_prefs.txt)

Please disregard a private email I sent yesterday: the changes did eventually reach Einstein correctly via propagation. I've performed a more careful and through test today, controlling propagation by blocking networking on all my machines, and only allowing one transfer at a time. That way, I've successfully updated CPDN, GPUGrid, NumberFields, and WCG.

I'll try a reverse transfer from NumberFields to the test project, to see if the ordering makes the difference, or if something else is in play.

RichardHaselgrove:
I've carried out the NumberFields test, and the updated file **was** transferred to the test site. The generated file (below) was in the traditional format, with the venues **below** the uncategorised entries.

[global_prefs_nf.txt](https://github.com/BOINC/boinc/files/9381774/global_prefs_nf.txt)

CharlieFenton:
@RichardHaselgrove Are you saying that when you log in to the test project and look at https://boinc.berkeley.edu/test/prefs.php?subset=global the values match the global_prefs.nf.txt file you linked? 

If I understand you correctly, they propagate correctly to the test project if the source is from NumberFields but not if it is from Einstein? But it did propagate correctly to several other projects you tried even if the source was Einstein. That certainly implies a problem with the new web code on the test project.

RichardHaselgrove:
@CharlieFenton: That would appear to be the situation. I made a trivial (non-visible) change at Numberfields, but that was enough to trigger a full re-write, and update the timestamp. It was the changed timestamp on the test server `prefs.php?subset=global` page that suggested to me that the new file had been accepted: we probably do need to perform a full audit on that, but we would a second project with the revised code, and swap settings between them. Later.

@davidpanderson: Could you please reinstate the normal `prefs.php?subset=global` link from the home page of the test site? The page is fully functional, just unnecessarily awkward to reach.

CharlieFenton:
I now modified the web preferences on the Rosetta web site, but I still can't get the prefs to propagate to the test project.

davidpanderson:
The semantics of prefs are:
- if you set them with the manager, it uses those
- else it uses the last-modified prefs from the projects you're attached to

At one point there was the idea of prefs propagating between projects.
This doesn't make sense if projects have different prefs web code,
so it's no longer supported.

davidpanderson:
It's under Computing / Preferences on the test project

RichardHaselgrove:
> It's under Computing / Preferences on the test project

Thanks - so it is. But that's just one subset.

CharlieFenton:
> if you set them with the manager, it uses those
You can't modify the _global_prefs.xml_ file from the Manager. Any changes in the Manager go into the _global_prefs_override.xml_ file, which by design does not propagate to the project web servers.

> else it uses the last-modified prefs from the projects you're attached to

Agreed. And that is what I am testing, thus: 
I first open the computing preferences on the project web site (I first tried this with Einstein, later with Rosetta) and modify them. 

Next, I update that project (Einstein or Rosetta) with the Manager. I wait until the Event Log shows _Scheduler request completed_ for that project.

Next, I examine my _global_prefs.xml_ file in my local BOINC Data directory and / or using the _Options->Computing preferences_ menu item in the Manager, to confirm that they match the changes I made on the project web site. 

Next I update the test project from the Manager, confirming that the Event Log shows _Scheduler request completed_ for the test project. 

Finally, I examine my computing preferences on the test site at https://boinc.berkeley.edu/test/prefs.php?subset=global and see that they have not changed and the preferences there have not been modified since 6 May 2013.

RichardHaselgrove:
@CharlieFenton - another check you can perform on this is to visually inspect, and if necessary do a plain-text edit on, the local global_prefs.xml file. During my testing for Christian on 2016, it transpired that there are certain unwritten assumptions about the format of that file, which Einstein broke (and still beaks).
 
In particular, it appeared that the `<mod_time>` tag had to appear in the file before before the first `<venue name=` tag. If it's lower down, move it before the venue sections, bump the timestamp by one second, and try again.

CharlieFenton:
@davidpanderson I have done more extensive testing and have found that I am after all able to propagate to and from the Test Project prefs, though there I have found another problem with this PR: If the prefs are propagating from BOINC 7.20.2, the new when-not-in-use values are set to defaults rather than to the when-in-use values from the client. This is a similar issue to item [1] [here](https://github.com/BOINC/boinc/pull/4871#issuecomment-1220434617) button the same, though possibly they are related. 

Apparently my test was failing because I was testing with Einstein and Rosetta. So you can ignore item [2] in [this list](https://github.com/BOINC/boinc/pull/4871#issuecomment-1220434617), but not the other items there. 

I have now tested with every possible combination of the projects to which I am currently attached (excluding SETI@home): Rosetta@home, Einstein@home, NFS@home, PrimeGrid, LHC@home and BOINC Test Project.

I found tonight that if my global_prefs.xml file has prefs from either Rosetta@home or Einstein@home, I cannot get them to propagate at all to any of the other projects (or to each other.) If the prefs are from NFS@home, PrimeGrid, LHC@home or BOINC Test Project, they propagate correctly to all of the others from that list, but do not propagate at all to Rosetta@home or Einstein@home. (I will create a new issue for this.)

So propagation to and from the Test Project does appear to work correctly with those 4 other projects, except for the one new problem I explained above.

CharlieFenton:
> In particular, it appeared that the <mod_time> tag had to appear in the file before before the first <venue name= tag. If it's lower down, move it before the venue sections, bump the timestamp by one second, and try again.

Interesting. I have not set up any venues, but looking at the _global_prefs.xml_ file I got from Einstein@home, the <max_ncpus> line does come before the <mod_time> line, so that may be part of the problem. However, the order of the entries on the  global_prefs.xml file from Rosetta@home is the same as from the projects which did correctly propagate.

CharlieFenton:
@davidpanderson For your convenience, here is a summary of the 9 unresolved problems to date with this PR, plus one that I have not been able to test. Please fix these so we can finish this PR.

The first 4 items are from [here](https://github.com/BOINC/boinc/pull/4871#issuecomment-1220434617):
[1] The test project had my old web prefs before you updated the server code. But when you updated the server code, it set the new "When computer is not in use" values to the new default. It should have propagated the old "When computer is in use" values to the new fields, for consistency with the way the client code works and with what the user would expect.

[2]  [RESOLVED](https://github.com/BOINC/boinc/pull/4871#issuecomment-1221293645)

[3] Because [this](https://github.com/BOINC/boinc/pull/4871#pullrequestreview-1072436875) and [this](https://github.com/BOINC/boinc/pull/4871#pullrequestreview-1072436906) line of code are missing for the new prefs, the dialog displays the new values as 0 rather than 100 when the dialog is saved and again opened, as explained [here](https://github.com/BOINC/boinc/pull/4871#issuecomment-1220434617).

[4] If the user enters 0 for prefs.niu_max_ncpus_pct it is set to the value of prefs.max_ncpus_pct rather than to 0 or 100.

[5] From [here](https://github.com/BOINC/boinc/pull/4871#issuecomment-1221293645): If the current web prefs are propagating from BOINC 7.20.2 to a project whose web prefs page has _when-not-in-use_ prefs, the new _when-not-in-use_ values on the web site are set to the new defaults rather than to the _when-in-use_ values from the client. This is a similar issue to item [1] but not the same, though possibly they are related.

[6] Let's say there are two projects, A and B, both of which have the new web prefs code, and the user is still running BOINC 7.20.2. The user sets web prefs on A, including custom values for _when-not-in-use_ prefs, and then does an update of the project from the client to update his _global_prefs.xml_ file. Those custom _when-not-in-use_ prefs are ignored by the BOINC 7.20.2 client. The user then does an update for project B from the client. Since the client has ignored the custom _when-not-in-use_ prefs, they are not propagated to project B's web prefs. Do you have a solution for this?

[7] From [here](https://github.com/BOINC/boinc/pull/4871#issuecomment-1214157173) and [here](https://github.com/BOINC/boinc/pull/4871#issuecomment-1220315293): This window is too big even now on Gnome3: #1872, which says the screen size is 1366 pixels wide by 768 pixels high, but the screen shot shows that the menu bar takes up part of that heigh, so the dialog needs to be be shorter than that.

[8] From [here](https://github.com/BOINC/boinc/pull/4871#issuecomment-1214240194): The help page for this dialog will also need to be updated.

[9] @davidpanderson wrote [here](https://github.com/BOINC/boinc/pull/4871#issuecomment-1214447201):
> If an old client gets a set of web prefs that includes the new prefs, it will ignore them. (the web GUI has a warning to this effect)

But there is no such warning at https://boinc.berkeley.edu/test/prefs.php?subset=global

[10] In addition, I have not yet been able to test this question from [here](https://github.com/BOINC/boinc/pull/4871#issuecomment-1214126019) because I currently have only one computer: 
> How will a new version of BOINC Manager work with a remote older version of the client, and vice versa?

CharlieFenton:
@davidpanderson I was able to test [10] by launching the old client first and then the new Manager, and vice versa. I found no problem with either of those combinations. So this leaves only problems [1] and [3] through [9] needing to be fixed.

CharlieFenton:
@davidpanderson I've edited the [list of problems](https://github.com/BOINC/boinc/pull/4871#issuecomment-1223941090) to correct some typos and omissions. Also, it would be nice if you can find a solution for item [6], but I feel it is not critical if there is no fix for that one.

davidpanderson:
I changed the web code to fix 1 and (I think) 5.
I changed the way NIU prefs are parsed, fixing 3 and 4.

6 doesn't happen.  When the client gets prefs from a project,
it puts them in global_prefs.xml and doesn't modify it.
It only modified global_prefs_override.xml

7: the dialog is now 644 pixels high on Win.  That's short enough.
We can't support tiny displays, and they can edit prefs other ways.

9: the web form says "Requires BOINC 7.20.3+"

10: works AFAIK

CharlieFenton:
@davidpanderson I have confirmed you have fixed all the items on my list (except [6] which probably has no solution.

I found a couple of additional problems, but I have fixed them in the code myself. (For some reason, a couple of the changes you made in commit 27aff32ff1 were lost; I replaced them.)

Regarding item 6, you wrote:
> 6 doesn't happen. When the client gets prefs from a project,
it puts them in global_prefs.xml and doesn't modify it.
It only modified global_prefs_override.xml

I'm sorry if I was not clear. I was referring only to changes on a project web site, so the propagation happens via _global_prefs.xml_. I agree that updating the client transfers the updated web prefs to _global_prefs.xml_, which is in effect updating it. 

When you then update another project, that project's web prefs receive the updated data from _global_prefs.xml_. The scenario I asked about was when a user is running BOINC 7.20.2 or earlier, which does not save the not-in-use prefs from the first project in _global_prefs.xml_. As a result, the second project does not receive the not-in-use prefs and sets them to the in-use prefs (the default.) This scenario does not involve _global_prefs_override.xml_ at all; the customizing I mentioned is done entirely on project websites.

But, as I wrote earlier, I suspect there is no fix for this and it is not critical.

CharlieFenton (approved):

CharlieFenton:
@davidpanderson Please remember to update the [documentation](https://boinc.berkeley.edu/wiki/Preferences), and add the missing description of _Suspend when no mouse/keyboard input in last xx minutes_ .

davidpanderson:
I added the "no input in X min" to the mediawiki page
(which had a number of other errors, some of which I fixed).
I'll update this to the new version when we release it.

AenBleidd:
For Ubuntu with the screen resolution 1366x768 the window is too big (the same as for the current version).
Not sure if we need to do smth with it in this particular PR.
Old version:
Screenshot from 2022-09-13 11-19-47
New version:
Screenshot from 2022-09-13 11-20-14

AenBleidd:
I suggest to go ahead and merge this, and fix the size issue later in the corresponding ticket instead https://github.com/BOINC/boinc/issues/1872
@davidpanderson, @CharlieFenton, what do you think?

AenBleidd (approved):

CharlieFenton:
> I suggest to go ahead and merge this, and fix the size issue later in the corresponding ticket instead https://github.com/BOINC/boinc/issues/1872

I agree.

RichardHaselgrove:
Likewise. I agree.

davidpanderson:
On Linux can you drag the dialog to make the bottom visible?

AenBleidd:
No. The window is already on its top most position.

AenBleidd:
As discussed, I'm merging it now.
I have an idea how to solve this issue #1872, so I'll take a couple of weeks to try it out before pushing new strings for localization to Transifex.

PR #4875, opened 2022-08-09
Mac: compute non-BOINC CPU usage in a more accurate way

### LINKED ISSUE #4327: Non-BOINC CPU usage time is incorrect

CharlieFenton:
Fixes #4327, #4328

**Description of the Change**
Implement #4859 for Macintosh computers

### DIFF
--- a/lib/procinfo_mac.cpp
+++ b/lib/procinfo_mac.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2020 University of California
+// Copyright (C) 2022 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -185,7 +185,7 @@ double total_cpu_time() {
     static double scale;
     uint64_t totalUserTime = 0;
 
-    if (!first) {
+    if (first) {
         first = false;
         long hz = sysconf(_SC_CLK_TCK);
         scale = 1./hz;
--- a/lib/procinfo_mac.cpp
+++ b/lib/procinfo_mac.cpp
@@ -26,6 +26,8 @@
 #include <unistd.h>
 #include <string>
 #include <locale.h>
+#include <sys/sysctl.h>
+#include <mach/mach.h>
 
 #if SHOW_TIMING
 #include <Carbon/Carbon.h>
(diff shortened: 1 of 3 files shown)


lib/procinfo_mac.cpp
```diff
@@ -167,3 +169,42 @@ int procinfo_setup(PROC_MAP& pm) {
...
+// From usr/include/mach/processor_info.h:
+// struct processor_cpu_load_info {             /* number of ticks while running... */
+//	 unsigned int    cpu_ticks[CPU_STATE_MAX]; /* ... in the given mode */
+// };
+//
+double total_cpu_time() {
+    static natural_t processorCount = 0;
+    processor_cpu_load_info_t cpuLoad;
+    mach_msg_type_number_t processorMsgCount;
+    static double scale;
```
AenBleidd, line 184: There is no default value for `scale`.
In case `processorCount` is not equal zero, this value will be set to `0` be default, so later this function will always return `0` because the last statement is multiplication with `scale`.
If this is intended - then fine, but I believe this is a mistake

AenBleidd: P.S. If you want just to initialize this static variable once, then it's better to have either some flag or just verify if this value is zero (and then initialize it with a proper value) otherwise just continue execution

CharlieFenton: @AenBleidd I intended this. I have no better solution in case of such errors because the code by @davidpanderson ( line 495 in _client/app.cpp_) which calls this function has no ability to handle errors returned from `total_cpu_time()`. The UNIX / Linux version of `total_cpu_time()` created for #4859 in _lib/procinfo_unix.cpp_ also returns 0 in case of any error opening or parsing `/proc/stat`. The MSW version of `total_cpu_time()` in _lib/procinfo_win.cpp_ from #4859 does no error checking at all.

I agree that the code which calls this should take some sort of appropriate action if `total_cpu_time()` indicates an error by returning 0.

AenBleidd (approved):

lib/procinfo_mac.cpp
AenBleidd: Ok, code looks OK to me. Please let me know when it's tested and ready to merge

CharlieFenton: One possibility for handling errors from `total_cpu_time()` might be to revert to the old method of calculating non_boinc CPU usage (lines 502 - 523 in _client/app.cpp_).

CharlieFenton: > P.S. If you want just to initialize this static variable once, then it's better to have either some flag or just verify if this value is zero (and then initialize it with a proper value) otherwise just continue execution

I'm not sure I understand what you are suggesting. `processorCount ` is always updated by the `host_processor_info()` call; there is no way to avoid that. But I am having it do double duty, both as a flag allowing me to call `sysconf(_SC_CLK_TCK)` only once, and also to indicate an error by returning 0 if for some reason `processorCount ` is 0 (as you pointed out in the first comment above.)

AenBleidd: This double duty is exactly that I don't like: it's an additional complication and fragile point, and in case you need to change the code, you still should remember that this variable serves as a flag too.
I believe in this case it's acceptable but in general case I'd recommend to use a separate variable as a flag.

davidpanderson:
I suggest using 
static bool first = true
if (first) {
   first = false
   ... get CPU count etc.

Also: total_cpu_time() should return 0 if any error occurred.

CharlieFenton:
This code is now ready to be merged.

I tested this by setting local preferences to suspend if non-BOINC CPU usage > 50%, setting the _mem_usage_debug_ flag  and letting BOINC run while telling Xcode to do a full build of BOINC. This computer has 4 CPUs, so the Xcode build runs hundreds of instances of the compiler (one per source file) 4 at a time. Each compiler instance runs for a fraction of a second.

With BOINC 7.20.2 (using the old code), BOINC failed to detect the high non-BOINC CPU usage. It reported mostly  between 4% and 15%, with 25% a few times. The old code never suspended BOINC.

With the new code, it reported mostly 80% to 90% non-BOINC CPU usage and correctly suspended until the Xcode build finished. So the new code is accomplishing what was intended: better recognizing the CPU load when there are multiple short duration processes.

PR #4876, opened 2022-08-09
[Bugfix][Manager] Check linux location for skins

talregev:
Check linux location for skins.


clientgui/SkinManager.cpp
```diff
@@ -19,6 +19,8 @@
 #pragma implementation "SkinManager.h"
 #endif
 
+#include <sys/types.h>
```
AenBleidd, line 22: In this particular case these headers are used for linux-only builds.
Please add the corresponding macro defined to exclude them from any other builds

AenBleidd (changes requested):

talregev:
I don't allow to copy my work

PR #4894, opened 2022-08-19
[wrapper] Update version

AenBleidd:


AenBleidd:
Trivial change, no review required

CharlieFenton:
The current wrapper version should be 26017, not 26018. The last wrapper released until my new Mac release earlier today had been 26016, released on 2015-11-19, as you can see [here](https://boinc.berkeley.edu/dl/?C=N;O=D).

CharlieFenton:
> The current wrapper version should be 26017, not 26018. The last wrapper released until my new Mac release earlier today had been 26016, released on 2015-11-19, as you can see [here](https://boinc.berkeley.edu/dl/?C=N;O=D). 

That is why I used 26017 for my Mac release. Please don't create a new version number without further changing something else; it will create confusion because now the Mac 26017 is the same code as Linux ARM64 26018.

AenBleidd:
I did this intentionally: wrapper was not released for a while, and being built for arm64 it was linked against latest BOINC library that is definitely more fresh that the on etwas used to build wrapper for other platforms.
And I increased version number to show this. Sorry for the confusion, I had to explain this before making this change and probably only after discussion with you.

PR #4897, opened 2022-08-22
[Manager] Add check if new columns are to be visible by default

Vulpine05:
**Description of the Change**
If a developer adds a new column to a tab in the Manager, that column will be added regardless of the default visibility.  There is no way of having the column hide by default unless the user resets to default columns.  This PR revises the code so that if a column is hidden by default, it will not be visible the first time the Manager is started.

### DIFF
--- a/clientgui/BOINCListCtrl.cpp
+++ b/clientgui/BOINCListCtrl.cpp
@@ -252,7 +252,6 @@ bool CBOINCListCtrl::OnRestoreState(wxConfigBase* pConfig) {
             int shownCount = orderArray.size();
             int hiddenCount = hiddenArray.size();
             int totalCount = pView->m_aStdColNameOrder->size();
-            int defaultCount = pView->m_iNumDefaultShownColumns;
             for (int i = 0; i < totalCount; ++i) {  // cycles through updated array of columns.
                 wxString columnNameToFind = pView->m_aStdColNameOrder->Item(i);
                 bool found = false;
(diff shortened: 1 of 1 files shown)


clientgui/BOINCListCtrl.cpp
```diff
@@ -240,42 +239,67 @@ bool CBOINCListCtrl::OnRestoreState(wxConfigBase* pConfig) {
...
+                foundNewColumns = true;
+                // If we got this far, then we know this column is new.
+                // Now it needs to be determined if the new column should be shown by default or not.
+                // Create array of default columns.
+                //
+                defaultArray.Clear();
+                for (int k = 0; k < pView->m_iNumDefaultShownColumns; ++k) {
+                    defaultArray.Add(pView->m_aStdColNameOrder->Item(pView->m_iDefaultShownColumns[k]));
+                }
+                for (int k = 0; k < defaultCount; ++k) {
```
AenBleidd, line 284: On the line 255 you saved `defaultCount` as a value from `pView->m_iNumDefaultShownColumns`. Then you put into `defaultArray` `pView->m_iNumDefaultShownColumns` number of values on the line 281. Here you iterate `defaultArray` using `defaultValue` again. For me it looks not safe and too compicated. Why just don't write
```
for (int k = 0; k < defaultArray.GetCount(); ++k) {
```
instead of 
```
for (int k = 0; k < defaultCount; ++k) {
```
?
Then you don't need `defaultCount` variable at all.

AenBleidd (changes requested):

clientgui/BOINCListCtrl.cpp
Vulpine05: I see what you're saying, I agree.  Fixed now.

AenBleidd (approved):

CharlieFenton:
@Vulpine05 Please create a new PR and fix these compiler warnings I believe you introduced with this and perhaps other PRs:
clientgui/BOINCListCtrl.cpp:246:14: warning: variable 'foundNewHiddenColumns' set but not used [-Wunused-but-set-variable]
        `bool foundNewHiddenColumns = false;`
             ^
clientgui/BOINCListCtrl.cpp:342:9: warning: unused variable 'colCount' [-Wunused-variable]
    `int colCount = GetColumnCount();`
        ^
clientgui/BOINCListCtrl.cpp:341:9: warning: unused variable 'i' [-Wunused-variable]
    `int i, stdCount, columnPosition;`

Thank you.

Vulpine05:
@CharlieFenton, I will definitely take a look at this.  At a glance, I don't think I introduced the last two (L341, L342).  The only changes I made in that function were deleting whitespace.  Regardless, I will take a look at all three and correct them.

PR #4907, opened 2022-08-28
Client: don't crash when <exclude_gpu> has bad device #

### LINKED ISSUE #4906: Boinc client won't start with cc_config, exclude_gpu, device_num=99
**Describe the bug**
Boinc client fails to start when cc_config.xml have exclude_gpu, device_num=99. The only error message is from BOINC manager, "The BOINC client has exited unexpectedly 3 times within the last 3 minutes".

**Steps To Reproduce**
1. Open cc_config.xml
2. `<cc_config><exclude_gpu><url>http://milkyway.cs.rpi.edu/milkyway/</url><device_num>99</device_num><type>ATI</type></exclude_gpu></cc_config>`

**Expected behavior**
Starts up and work, maybe with a warning about invalid device_num

**System Information**
 - OS: Windows 10
 - BOINC Version: 7.20.2

**Additional context**
Old Boinc version 7.16.20 did start ok, but may have some random stability problems with invalid device_num.
[cc_config.txt](https://github.com/BOINC/boinc/files/9438900/cc_config.txt)

davidpanderson:
Fixes #4906

### DIFF
--- a/lib/cc_config.cpp
+++ b/lib/cc_config.cpp
@@ -274,6 +274,9 @@ int EXCLUDE_GPU::parse(XML_PARSER& xp) {
         if (!xp.is_tag) continue;
         if (xp.match_tag("/exclude_gpu")) {
             if (!found_url) return ERR_XML_PARSE;
+            if (device_num >= MAX_COPROC_INSTANCES) {
+                return ERR_XML_PARSE;
+            }
             return 0;
         }
         if (xp.parse_string("url", url)) {


Vulpine05:
I see now that there is a limit of 64 coprocessors.  Perhaps there should be a new error message saying the coprocessor ID needs to be within range?  That could help users troubleshoot their error message without asking the same question on the message boards (not that this should be a common problem, hopefully).

AenBleidd (approved):

AenBleidd:
Since this happens in XML parsing (and ideally it should not be changed manually), current fix is ok.

PR #4912, opened 2022-09-05
[Manager] Freeze view when modifying columns

### LINKED ISSUE #4888: [Manager] When hiding a column after moving scroll bar, Manager crashes
**Describe the bug**
When hiding a column, the Manager will crash if the scroll bar is not in the far-left position.  This only happens when you hide a column on the tab that is currently in view, AND there are items in the tab.  I can reproduce this easily in the Projects and Tasks tabs, but I couldn't reproduce this in the Transfers tab, probably because there were no items in that tab.

When the Manger crashes, the active tasks continue to run.  When I open the Manager again it returns to the tab that was opened when the Manager started previously, and the column that was requested to be hidden is still visible.

I tried reproducing this on another computer running Ubuntu 18.04 with Gianfranco Costamanga's PPA (7.20.2), but I could not get it to reproduce.  I would be interested to see if this happens on a Mac, or maybe an older version of Windows or BOINC.

**Steps To Reproduce**
1. In BOINC Manager, move the scroll bar in any tab to the right.  It can be all the way, or just a little bit.  The key is for the scroll bar to not be at the furthest left position.
2. Options->Select Columns...
3. Hide a column, click save.
4.  "Select Columns" window closes, the scroll bar moves to the left, and then Manager exits without warning.

**Expected behavior**
The column should just hide without crashing.

**System Information**
 - OS: Windows 10 Home, 10.0.19043 Build 19043
 - BOINC Version: 7.20.2, wxWidgets 3.1.7

Vulpine05:
Fixes #4888

**Description of the Change**
Implements a freeze and thaw of the active view window when adding or removing columns.

Also unrelated, but wxWidgets has a remove all columns function.  I've added this to replace the iterative remove column function.

**Alternate Designs**
N/A

**Release Notes**
Fixed crash when removing columns when horizontal scroll bar is not at left-most position

### DIFF
--- a/clientgui/BOINCListCtrl.cpp
+++ b/clientgui/BOINCListCtrl.cpp
@@ -349,11 +349,12 @@ void CBOINCListCtrl::SetListColumnOrder(wxArrayString& orderArray) {
     CBOINCBaseView* pView = (CBOINCBaseView*)GetParent();
     wxASSERT(wxDynamicCast(pView, CBOINCBaseView));
 
+    // Manager will crash if the scroll bar is not at the left-most position on the
+    // current view if columns are modified.
+    //
+    pView->Freeze();
     pView->m_iColumnIndexToColumnID.Clear();
-    for (i=colCount-1; i>=0; --i) {
-        DeleteColumn(i);
-    }
-
+    DeleteAllColumns();
     stdCount = pView->m_aStdColNameOrder->GetCount();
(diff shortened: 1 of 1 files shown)


AenBleidd (approved):

AenBleidd:
LGTM, thank you for the fix!

PR #4938, opened 2022-09-29
Client:  Update cpu/gpu resources when app_config is re-read

### LINKED ISSUE #1903: Displayed CPU/GPU count for app is not updated when app_config.xml changes them
When you change number of required CPUs and/or GPUs for app using app_config.xml, GUI is not updated - it still shows old values. 3rd party software like BoincTasks also shows it in this way, so BOINC Client sends them incorrect values. This is display-only issue, number of running tasks is properly adjusted. New tasks downloaded later shows correct value. If I remember correctly, BOINC restart also fixes this.

Vulpine05:
Fixes #1903

**Description of the Change**
When the app_config is re-read, the CPU/GPU resources will be updated in the Manager and boinccmd without requiring BOINC to restart.

Edit:  Changes are made in the client.  The CPU/GPU string stored in "resources" is printed only once.  If there is a string present, the client was not reprinting the resources if they changed.  This PR now passes a bool for the case when app_config is re-read so the resources can be reprinted.  No changes were made to the Manager.

Also note that if app_config is removed, the CPU/GPU resources do not change.  The client doesn't update these currently, it needs information from the server to provide default CPU/GPU usage.

**Alternate Designs**
None.

**Release Notes**
CPU/GPU usage in Manager displays correctly immediately when app_config is re-read.

### DIFF
--- a/client/client_state.h
+++ b/client/client_state.h
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2020 University of California
+// Copyright (C) 2022 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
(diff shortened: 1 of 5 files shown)


AenBleidd:

AenBleidd (approved):

RichardHaselgrove:
Tomorrow morning.

RichardHaselgrove:
OK, testing under Windows 7. This is my normal configuration, showing three of the normal cases:

image

Loading a revised app_config.xml changed it to this:

image

I think that's what we wanted to see. Reverting the change restored the status quo ante. No side effects observed, so I think this can be merged.

AenBleidd (approved):

PR #4940, opened 2022-09-29
[Manager] Convert ProcessorTab in Preferences dialog from wxPanel to wxScrolledWindow to show all content on small screens (e.g. 1024x768).

### LINKED ISSUE #1872: Computing Preferences window is too big under 1366*768 resolution

AenBleidd:
This fixes #1872.

### DIFF
--- a/clientgui/DlgAdvPreferencesBase.cpp
+++ b/clientgui/DlgAdvPreferencesBase.cpp
@@ -200,6 +200,7 @@ CDlgAdvPreferencesBase::CDlgAdvPreferencesBase( wxWindow* parent, int id, wxStri
 
     dialogSizer->Fit( this );
     this->SetSizer( dialogSizer );
+    m_panelProcessor->SetVirtualSize(notebookSizer->GetMinSize());
 }
 
 #define PAD0    1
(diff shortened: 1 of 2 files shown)


AenBleidd:
@CharlieFenton, could you please review and test this? I know this is mainly for linux. but I just want to be sure I broke nothing for OSX as well. Thank you in advance

AenBleidd:
The only difference that it's not looks like this even on a big screens:
image
But for me this is ok.
@davidpanderson, is it ok for you as well or it's better to spend more time and make it looks the old way (with no scrolling) on big screens?

Vulpine05:
> The only difference that it's not looks like this even on a big screens

I'm not sure I understand what you are saying.  Are you saying on larger resolutions (screens), the window is the size as you show, and the scrollbar is by default?  I assume resizing the window would remove the scrollbar.

My preference would be to not have the scrollbar show if the screen can fit a larger window.  How many users are using the smaller resolution compared to 1080?

AenBleidd:
Ok, I'm closing this for now to find the better solution

PR #4955, opened 2022-10-08
[Manager] [Advanced View] Add warning message in preferences for suspension

### LINKED ISSUE #4939: "In Use" and "Not In Use" suspension preferences could prevent computing indefinitely
**Describe the bug**

**Steps To Reproduce**

**Expected behavior**

**System Information**
 - OS: Windows 10

**Additional context**

Vulpine05:
Partially fixes #4939

**Description of the Change**
This commit checks for the condition stated in #4939 when advanced preferences are changed locally.  An error message is created if the condition exists.  The user will have to modify one of the values in order to save their preferences.  The revised error message is:

image

**Alternate Designs**
#4634 was my previous draft.  The conditional logic was slightly different.  I decided this time to validate both values first separately, then check to see if both doubles are "active".  This minimizes additional variables for a condition that is probably rare.

**Release Notes**
Fixed indefinite suspension when suspension times were not set correctly in preferences.

### DIFF
--- a/clientgui/DlgAdvPreferences.cpp
+++ b/clientgui/DlgAdvPreferences.cpp
@@ -744,6 +744,7 @@ bool CDlgAdvPreferences::ValidateInput() {
     wxString invMsgLimit10 = _("Number must be between 0 and 10");
     wxString invMsgLimit100 = _("Number must be between 0 and 100");
     wxString invMsgLimit1_100 = _("Number must be between 1 and 100");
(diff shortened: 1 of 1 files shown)


Vulpine05:
Hang on, I think I just came up with a more straightforward solution.  Marking WIP for now while I test it.

Vulpine05:
@AenBleidd sorry for the back and forth, but I think this is ready for review once it passes checks.

AenBleidd (approved):

Issue #4414, opened 2021-06-22
[Android] Convert left *.java files to Kotlin


AenBleidd:
- [x] #4445
- [x] #4518
- [x] #4944 
- [ ] #4932 
- [x] #4945 
- [ ] #4934 
- [ ] #4935 
- [ ] #4936 
- [ ] client/ClientStatus
- [ ] client/DeviceStatus
- [ ] #4977 
- [x] #4949
- [ ] ui.eventlog/EventLogGuiFragment

AenBleidd:
Assigned to @pAkalpa

PR #4977, opened 2022-10-15
Remove NoticeNotification.java

pAkalpa:
Fixes #4414 

**Description of the Change**

Convert NoticeNotification.java to NoticeNotification.kt

**Alternate Designs**

nil

**Release Notes**


AenBleidd (changes requested):
Could be nice to fix this two hints:
- Missing `PendingIntent` mutability flag (Line 67)
- Missing `PendingIntent` mutability flag (Line 132)

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/NoticeNotification.kt
```diff
@@ -0,0 +1,215 @@
...
+import edu.berkeley.boinc.R.plurals
+import edu.berkeley.boinc.R.string
+import edu.berkeley.boinc.rpc.Notice
+import edu.berkeley.boinc.utils.Logging.Category.CLIENT
+import edu.berkeley.boinc.utils.Logging.logException
+import edu.berkeley.boinc.utils.getBitmapFromVectorDrawable
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@RequiresApi(VERSION_CODES.M)
```
AenBleidd, line 49: We support Android versions starting from Android 4.1 (API level 16).
If this set, this will not work for any that is older than Android 6.0 (API level 23).
So you should either find a way to fix that warning without increasing making ANdroid 6.0 as a minimal supported version or just rollback and leave it as-is (the first in preferred).

AenBleidd (changes requested):

android/BOINC/app/src/main/java/edu/berkeley/boinc/client/NoticeNotification.kt
pAkalpa: I can add the below code, but a deprecated warning won't go away
`        
contentIntent = if (VERSION.SDK_INT >= VERSION_CODES.M)	{	
PendingIntent.getActivity(context, 0, intent,PendingIntent.FLAG_IMMUTABLE or PendingIntent.FLAG_UPDATE_CURRENT) } else { PendingIntent.getActivity(context, 0, intent, PendingIntent.FLAG_UPDATE_CURRENT) }
`

AenBleidd: Then just rollback your last commit and it should be ok

AenBleidd: And add this annotation at the top of this class:
```
@SuppressLint("UnspecifiedImmutableFlag")
```

AenBleidd (approved):

PR #4995, opened 2022-11-05
web: show disk size in TB if needed

davidpanderson:


AenBleidd:
@lfield, please review.

PR #5002, opened 2022-11-07
Mac: Fixes for MacOS 13 and Xcode 14

CharlieFenton:
Fixes BOINC screensaver and Manager "Show Graphics" button functionality to work with MacOS 13 Ventura. This also required a different implementation of security to prevent access to user's files in the unlikely case of  malevolent or buggy project graphics apps.

As of MacOS 13.0 Ventura IOSurface cannot be used to share graphics between apps unless they are running as the same user, so  the screensaver can no longer run the graphics apps as user boinc_master or user boinc_project. Similarly, project graphics no longer work if run as user boinc_master or user boinc_project when run from the Manager's _Show Graphics_ button, though I don't fully understand why.

As a result, they now must be run as the logged in user, creating a potential security risk. I have implemented a technique to block access to the user's files and other sensitive directories based on an API that Apple uses extensively but has marked deprecated because it is an Apple private API. Apple warns it has deprecated the API because the syntax used to specify the access restrictions one passes to this API is subject to change without notice. But the syntax has remained unchanged for many years, this API is used widely in Apple's software, and the security profile elements we use are quite basic and so are very unlikely to change. I can find no other way to provide this security.

### DIFF
--- a/clientgui/MainDocument.cpp
+++ b/clientgui/MainDocument.cpp
@@ -1646,6 +1646,10 @@ RUNNING_GFX_APP* CMainDocument::GetRunningGraphicsApp(RESULT* rp) {
        }
 
         // Either the graphics app had already exited or we just killed it
+#ifdef __APPLE__
+        // Graphics app wrote files in slot directory as logged in user, not boinc_master
+        fix_slot_file_owners((*gfx_app_iter).slot);
+#endif
         (*gfx_app_iter).name.clear();
         (*gfx_app_iter).project_url.clear();
         m_running_gfx_apps.erase(gfx_app_iter);
@@ -1693,6 +1697,9 @@ void CMainDocument::KillInactiveGraphicsApps() {
             if (GetGFXPIDFromForkedPID(&(*gfx_app_iter))) { 
                 KillGraphicsApp((*gfx_app_iter).gfx_pid);
                 KillGraphicsApp((*gfx_app_iter).pid);
+            
+                // Graphics app wrote files in slot directory as logged in user, not boinc_master
+                fix_slot_file_owners((*gfx_app_iter).slot);
            }
 #else
             KillGraphicsApp((*gfx_app_iter).pid);
@@ -1724,9 +1731,13 @@ void CMainDocument::KillAllRunningGraphicsApps()
             continue;
         }
         KillGraphicsApp((*gfx_app_iter).gfx_pid);
-#endif
-
         KillGraphicsApp((*gfx_app_iter).pid);
+
+        // Graphics app wrote files in slot directory as logged in user, not boinc_master
+        fix_slot_file_owners((*gfx_app_iter).slot);
+#else
+        KillGraphicsApp((*gfx_app_iter).pid);
+#endif
         (*gfx_app_iter).name.clear();
         (*gfx_app_iter).project_url.clear();
         m_running_gfx_apps.erase(gfx_app_iter);
@@ -1775,6 +1786,15 @@ int CMainDocument::GetGFXPIDFromForkedPID(RUNNING_GFX_APP* gfx_app) {
     pclose(fp);
     return gfx_app->gfx_pid;
 }
+
+
+int CMainDocument::fix_slot_file_owners(int slot) {
+        // Graphics apps run by Manager write files in slot directory
+        // as logged in user, not boinc_master. This ugly hack tells 
+        // BOINC client to fix all ownerships in this slot directory
+        rpcClient.run_graphics_app("stop", slot, "");
+    return 0;
+}
 #endif
 
 
@@ -1820,6 +1840,8 @@ int CMainDocument::WorkShowGraphics(RESULT* rp) {
             } else {
                 KillGraphicsApp(previous_gfx_app->pid); // User clicked on "Stop graphics" button
             }
+            
+            GetRunningGraphicsApp(rp);  // Let GetRunningGraphicsApp() do necessary clean up            
 #else
         // On other platforms, don't launch a second instance if graphics app is already running
 #endif
@@ -1832,7 +1854,7 @@ int CMainDocument::WorkShowGraphics(RESULT* rp) {
             NULL
         };
 
-#ifndef __WXMSW__
+#ifdef __APPLE__
         if (g_use_sandbox) {    // Used only by Mac
             int pid = fork();
             char path[MAXPATHLEN];
@@ -1867,7 +1889,10 @@ int CMainDocument::WorkShowGraphics(RESULT* rp) {
             gfx_app.pid = id;
             gfx_app.gfx_pid = 0;    // Initialize GetGFXPIDFromForkedPID()
             gfx_app.gfx_pid = GetGFXPIDFromForkedPID(&gfx_app);
-        } else {    // !g_use_sandbox
+        } else  // !g_use_sandbox
+#endif  // __APPLE__
+#ifndef __WXMSW__
+        {
             iRetVal = run_program(
                 rp->slot_path,
                 rp->graphics_exec_path,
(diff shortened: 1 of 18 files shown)


CharlieFenton:
I will be releasing this hot fix for the Mac only. While I have made a major effort to not affect the graphics on other platforms, especially for the "Show graphics" button in BOINC Manager, I have no way to test this on other platforms. That functionality needs to be confirmed on Linux, MS Windows, etc. before the next release for those platforms.

I am less concerned about the screensaver code for MS Windows because the bulk of my changes are in Mac-specific files or guarded by `#ifdef __APPLE__`. But that functionality should also be thoroughly tested on MS Windows before a release for that platform.

clientgui/MainDocument.cpp
```diff
@@ -1750,61 +1795,89 @@ int CMainDocument::WorkShowGraphics(RESULT* rp) {
...
+                // graphics" button.
+                if (GetGFXPIDFromForkedPID(previous_gfx_app)) {
+                    KillGraphicsApp(previous_gfx_app->pid); // User clicked on "Stop graphics" button
+                    KillGraphicsApp(previous_gfx_app->gfx_pid);
+                }
+            } else {
+                KillGraphicsApp(previous_gfx_app->pid); // User clicked on "Stop graphics" button
+            }
+#else
+        // On other platforms, don't launch a second instance if graphics app is already running
```
AenBleidd, line 1824: Shouldn't we kill already running graphics app on other platforms as well?

AenBleidd:
@CharlieFenton, on linux compilation is failed:
```
MainDocument.cpp: In member function ‘int CMainDocument::WorkShowGraphics(RESULT*)’:
MainDocument.cpp:1855:17: error: ‘getPathToThisApp’ was not declared in this scope
 1855 |                 getPathToThisApp(path, sizeof(path));
      |                 ^~~~~~~~~~~~~~~~
MainDocument.cpp:1861:27: error: ‘callPosixSpawn’ was not declared in this scope
 1861 |                 iRetVal = callPosixSpawn(cmd);
      |                           ^~~~~~~~~~~~~~
MainDocument.cpp:1868:21: error: ‘struct RUNNING_GFX_APP’ has no member named ‘gfx_pid’
 1868 |             gfx_app.gfx_pid = 0;    // Initialize GetGFXPIDFromForkedPID()
      |                     ^~~~~~~
MainDocument.cpp:1869:21: error: ‘struct RUNNING_GFX_APP’ has no member named ‘gfx_pid’
 1869 |             gfx_app.gfx_pid = GetGFXPIDFromForkedPID(&gfx_app);
      |                     ^~~~~~~
MainDocument.cpp:1869:31: error: ‘GetGFXPIDFromForkedPID’ was not declared in this scope
 1869 |             gfx_app.gfx_pid = GetGFXPIDFromForkedPID(&gfx_app);
      |                               ^~~~~~~~~~~~~~~~~~~~~~
```

clientgui/MainDocument.cpp
CharlieFenton: The code is correct as written.

This part of the code is refactored from lines 1799-1801 of _Maindocument.cpp_ in GIT master. On other platforms, it is executed only when the user clicks on "Show graphics" in BOINC Manager for a selected task; if the graphics for the selected task is already running, we just want to ignore this click. 

On Mac only, this button is changed to "Stop graphics" when the task is running, so we want to kill it on Mac only. I made the button change to "Stop graphics" is that some project graphics apps (e.g., Einstein) don't respond to the close box on newer versions of MacOS.

CharlieFenton:
I'm not sure, but I think the Linux failure appears to be due to an internal problem with GitHub Actions rather than related to the code:
> Error: Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255

AenBleidd:
> I'm not sure, but I think the Linux failure appears to be due to an internal problem with GitHub Actions rather than related to the code:
> 
> > Error: Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255

Yes, you can ignore it, I'll rerun it if needed

AenBleidd:
Android failure is unrelated to this PR and will be fixed separately

AenBleidd (approved):
Looks good to me

AenBleidd:
@RichardHaselgrove, I don't want to block this from being merged and released for OSX only, so could you please make a note to yourself somewhere to test this functionality later when we will be testing a next major release? Thank you in advance

RichardHaselgrove:
I'll put it on the ToDo list.

Item zero: find a project that still issues graphics apps...

CharlieFenton:
> Item zero: find a project that still issues graphics apps...

I tested with World Community Grid and Einstein, both of which have graphics on the Mac. I believe Rosetta does, also, but I don't currently have any Rosetta tasks to confirm that.

RichardHaselgrove:
Thanks. I run both those projects, but only on GPUs: the GPU versions of their apps don't include the graphics element (subject to more detailed checking later). I do have CPU-only devices, but they're currently resting because of the price of electricity. I'll fire one or two up again for testing.

Ken-Dinwiddie:
You guys must have analyzed the complete error message already. But, because I have seen it so many tooth-grinding times, here is the complete (although discontinuous) sucker. I'm too lazy to dig the continuous version out of the UNIX weeds. Hope the corrected code gets folded in to a new release soon. And how about fixing VirtualBox for late model macOS (Ventura and later) running on M[1,2,3...] processors?
[Crash.pdf](https://github.com/BOINC/boinc/files/9995884/Crash.pdf)

CharlieFenton:
@Ken-Dinwiddie The only changes in [this BOINC 7.20.4 build](https://boinc.berkeley.edu/dl/boinc_7.20.4_macOSX_universal.zip) from 7.20.2 should be fixes to the "Show Graphics" functionality in BOINC Manager and to the screensaver. Please download and try this build and let us know if you find any problems.

Note: I have noticed that some World Community Grid graphics apps appear to be buggy. They sometime work and sometimes don't, both with this new version 7.20.4 and the previous version 7.20.2, so that problem is not due to these changes in BOINC.

CharlieFenton:
Note: there is a known problem under MacOS 11 and later: the close button no longer works on some graphics apps, though it continues to work on others. It is unclear why. Since the graphics apps are provided by the projects, we have no way to fix this at our end. Our solution (really a workaround) was to modify the BOINC Manager so the "Show Graphics" button changes to "Stop Graphics" when the graphics app is displayed in PR #3970 and #3975.

Ken-Dinwiddie:
Thanks very much. I shall try v 7.20.4 and let you know…

From: CharlieFenton ***@***.***>
Date: Sunday, November 13, 2022 at 03:35
To: BOINC/boinc ***@***.***>
Cc: Kendall Dinwiddie ***@***.***>, Mention ***@***.***>
Subject: Re: [BOINC/boinc] Mac: Fixes for MacOS 13 and Xcode 14 (PR #5002)

@Ken-Dinwiddie<https://github.com/Ken-Dinwiddie> The only changes in this BOINC 7.20.4 build<https://boinc.berkeley.edu/dl/boinc_7.20.4_macOSX_universal.zip> from 7.20.2 should be fixes to the "Show Graphics" functionality in BOINC Manager and to the screensaver. Please download and try this build and let us know if you find any problems.

Note: I have noticed that some World Community Grid graphics apps appear to be buggy. They sometime work and sometimes don't, both with this new version 7.20.4 and the previous version 7.20.2, so that problem is not due to these changes in BOINC.

RichardHaselgrove:
OK, rebooted, and the screensaver is firing up - we're in business. Is that a known thing?

RichardHaselgrove:
Sorry, that was a follow-up to https://github.com/BOINC/boinc/pull/5010#issuecomment-1312571325 in the other PR.

PR #5007, opened 2022-11-09
Set the corrrect permissions

lfield:
Sets the permissions to be -rwxrwx---

Fixes an issues found during testing of the server release 1.4

### DIFF
--- a/py/Boinc/setup_project.py
+++ b/py/Boinc/setup_project.py
@@ -297,7 +297,7 @@ def create_project_dirs(dest_dir):
         'html/user_profile/images',
     ]
     for d in directories:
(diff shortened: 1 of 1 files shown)


davidpanderson:
what is the issue and how does this fix it?

lfield:
Sorry, here is the bug report. #5008

AenBleidd:
@davidpanderson, could you please take a look at this PR one more time?

davidpanderson:
I don't understand what the problem was, or how this change fixes it.
The 2770 sets the "setgid" bit in the permissions, which is necessary.
This has been here forever and seems to work.

lfield:
I am not sure this is needed.

PR #5017, opened 2022-11-17
scheduler: add a debug printf for multi-thread issue

davidpanderson:


lfield:
Already tested on LHC@home Dev.

PR #5021, opened 2022-11-19
Fix broken confirm in process_version

davidbolvansky:
PHP Notice:  Undefined variable: confirm in /home/boincadm/projects/fitcrack/bin/update_versions on line 480


davidpanderson:
Thanks!

PR #5025, opened 2022-11-22
Fix for issue 5024

davidbolvansky:
Works fine now.

image

### DIFF
--- a/clientgui/AccountManagerProcessingPage.cpp
+++ b/clientgui/AccountManagerProcessingPage.cpp
@@ -263,7 +263,7 @@ void CAccountManagerProcessingPage::OnStateChange( CAccountManagerProcessingPage
             }
 
             if (pWA->project_config.uses_username) {
-                username = (const char*)pWA->GetAccountUsername().mb_str();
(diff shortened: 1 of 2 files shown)


davidpanderson:
We should make the same change for account manager user name:
AccountManagerProcessingPage.cpp:266:                username = (const char*)pWA->GetAccountUsername().mb_str();

I guess we can assume that passwords, proxy names, etc. won't have problem characters.

RichardHaselgrove:
@lfield - can we squeeze the consequentials and testing into server release 1.4.0? Else, when would it be deployed?

davidbolvansky:
> We should make the same change for account manager user name: AccountManagerProcessingPage.cpp:266: username = (const char*)pWA->GetAccountUsername().mb_str();

Done.

> I guess we can assume that passwords, proxy names, etc. won't have problem characters.

Right.

davidpanderson:
Thanks for the fix.  This should appear in a client release fairly soon.

davidpanderson:
BTW, this is a client (manager) change, not a server change.

davidbolvansky:
Thanks!

PR #5027, opened 2022-11-27
Canonicalize URLs to lower case

### LINKED ISSUE #4937: Make BOINC urls case-insensitive
**Describe the problem**
When a user is given a URL to attach to a BOINC project, it will not work if the case is not correct. For example: boincproject.com/boinc would work but boincproject.com/BOINC would not. Domain names are case-insensitive, and I see no reason for case-sensitive URL handling. There is a chance that a user without the correct case-sensitive URL may try to attach to a project and therefore have issues doing so. Making URL handling case-insensitive would eliminate this source of error and confusion.

This is a further extension of https://github.com/BOINC/boinc/issues/4801

I realize this is a small issue and there are more important ones, just wanted to raise it as an idea.

Make BOINC server URL handling case-insensitive

hristog:
Fixes #4937

**Description of the Change**
This introduces new functionality to `canonicalize_master_url`, namely converting all input ASCII alphabet characters to their respective lower-case representations.
`std::tolower(static_cast<unsigned char>(ch))` is utilised for this purpose from `ctypes.h`.
Unit tests have been added correspondingly.

**Alternate Designs**

N/A

**Release Notes**

Updates canonicalize_master_url to normalize all URLs to lower-case, wherever applicable.

### DIFF
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -908,43 +908,14 @@ static void handle_project_attach(GUI_RPC_CONN& grc) {
         }
     }
 
-	// remove http(s):// at the beginning of project address
-	// there is no reason to connect to secure address project
-	// if we're already connected to the non-secure address
-	// or vice versa
-	// also clear last '/' character if present
-
-    const string http = "http://";
-    const string https = "https://";
-
-    string new_project_url = url;
-    size_t pos = new_project_url.find(http);
-    if (pos != string::npos) {
-        new_project_url.erase(pos, http.length());
-    }
-    else if ((pos = new_project_url.find(https)) != string::npos) {
-        new_project_url.erase(pos, https.length());
-    }
-    if (new_project_url.length() >= 1 && new_project_url[new_project_url.length() - 1] == '/') {
-        new_project_url.erase(new_project_url.length() - 1, 1);
-	}
+    canonicalize_master_url(url);
 
     for (i=0; i<gstate.projects.size(); i++) {
         PROJECT* p = gstate.projects[i];
         string project_url = p->master_url;
+        canonicalize_master_url(project_url);
 
-        pos = project_url.find(http);
-        if (pos != string::npos) {
-            project_url.erase(pos, http.length());
-        }
-        else if ((pos = project_url.find(https)) != string::npos) {
-            project_url.erase(pos, https.length());
-        }
-        if (project_url.length() >= 1 && project_url[project_url.length() - 1] == '/') {
-            project_url.erase(project_url.length() - 1, 1);
-        }
-
-        if (new_project_url == project_url) {
+        if (url == project_url) {
             already_attached = true;
             break;
         }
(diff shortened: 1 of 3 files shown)


client/gui_rpc_server_ops.cpp
```diff
@@ -908,43 +908,13 @@ static void handle_project_attach(GUI_RPC_CONN& grc) {
...
-        if (pos != string::npos) {
-            project_url.erase(pos, http.length());
-        }
-        else if ((pos = project_url.find(https)) != string::npos) {
-            project_url.erase(pos, https.length());
-        }
-        if (project_url.length() >= 1 && project_url[project_url.length() - 1] == '/') {
-            project_url.erase(project_url.length() - 1, 1);
-        }
+        string project_url = canonicalize_master_url(p->master_url);
```
hristog, line 915: Is it safe to modify `p->master_url` here? Would it have undesired implications outside of the scope of this function?

lib/url.cpp
```diff
@@ -203,6 +203,7 @@ void escape_url_readable(char *in, char* out) {
 // or prepend it
 //   - Remove double slashes in the rest
 //   - Add a trailing slash if necessary
+//   - Convert all alphabet characters to lower case
```
hristog, line 206: Should it explicitly be indicated that only ASCII characters are relevant in this case.
Do we need to do anything specific, regarding Unicode URLs (I'm not familiar with how those are handled by BOINC)?

hristog:
If this change is accepted, should the note from [this](https://github.com/BOINC/boinc/pull/4803/files) PR be removed, accordingly?
To me it seems, it definitely should.

davidpanderson:
Domain names are case-insensitive, but the filename part of a URL (after the /) is case sensitive.
If you change its case, the URL won't work anymore.
So I'm not sure this change is a good idea.

Vulpine05:
> Domain names are case-insensitive, but the filename part of a URL (after the /) is case sensitive. If you change its case, the URL won't work anymore. So I'm not sure this change is a good idea.

I see that now.  Amicable Numbers, for example, works if any letter is capitalization is switched before the first single '/', but won't after the /.  Trying the same test in a web browser has the same results (web site access if case sensitivity matches, 404 if it doesn't).

So I would think #4937 should be closed because there is no bug.  This PR could maybe be changed to just canonicalize the URL properly?

hristog:
> Domain names are case-insensitive, but the filename part of a URL (after the /) is case sensitive.
That's a great point - thanks!

> This PR could maybe be changed to just canonicalize the URL properly?
I like the suggestion, and will update the PR accordingly, if that's okay, @davidpanderson?

hristog:
@davidpanderson @AenBleidd I've updated the PR accordingly and would appreciate a review, when you get a chance!
Please, feel free to reject the PR, if you believe that the added value would be insufficient.

davidpanderson:
This looks OK to me.  @AenBleidd, please review and merge if OK.

AenBleidd (approved):

AenBleidd:
Looks good to me as well.
@hristog, thanks for contribution!

PR #5030, opened 2022-11-29
web: fix recaptcha validation

davidpanderson:
The old code stopped working; I don't know why.
Maybe some change at Google's end.
I switched to a simpler method, as described here: https://developers.google.com/recaptcha/docs/verify


lfield:
Works for Seti and CPDN. Will also re-test on the LHC@home Dev server before release.

PR #5047, opened 2022-12-27
scheduler: set type of Apple GPUs to "Apple"

### LINKED ISSUE #5036: Create "Apple" GPU scheduling group
Currently there are three hardcoded scheduling groups for GPUs, Nvidia, AMD, and Intel.

When a Mac requests a task for GPU, it does so as an unknown vendor and includes the full description; ie "Apple M1"

The server must know "Apple M1" as a coprocessor in order to send tasks. However, there's now an Apple M1 Max, M1 Pro, M1 Ultra, M2... with more to come. On the server side, each of these must be individually hard coded; currently 8(?) with the probability of continuous changes in the future.

If "Apple" was also a scheduling group, all Mxxxxx GPUs would be lumped together by the client as "Apple GPU" and the server would only have one "thing" to deal with. I realize this requires a change to both client and server, and would only work with those running up-to-date code on both sides, but it would be an improvement. At the least, a project could say "if you're running a Mac M3 ProUltraMaxExtreme or newer, you must use BOINC 7.22.7 or later" instead of having a list a mile long.

davidpanderson:
When the client reports an OpenCL GPU (other than NVIDIA/AMD/Intel) in a scheduler request, the "type" field is the model name. In the case of Apple GPUs this is "Apple M1" or "Apple M2" (and other models as time goes on).
The type field is used in plan class descriptions.

In the case of Apple the models are (I think) equivalent software-wise, and it would be a nuisance to have to make a new plan class (and app versions) for each model.

So - in the case of Apple GPUs - use "Apple" as the type. The model name is still available (as opencl_prop.name) if needed.

Fixes #5036

### DIFF
--- a/lib/coproc.cpp
+++ b/lib/coproc.cpp
@@ -140,8 +140,8 @@ int COPROC::parse(XML_PARSER& xp) {
     while (!xp.get_tag()) {
         if (!xp.is_tag) continue;
         if (xp.match_tag("/coproc")) {
-            // the client reports the type of Apple GPUs as the model
-            // (e.g. "Apple M1")
+            // The client reports the type of Apple GPUs as the model
+            // (e.g. "Apple M1").
             // Change it to just "Apple".
             //
             if (strstr(type, "Apple")) {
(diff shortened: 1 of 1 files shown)


CharlieFenton (approved):
I have no way to test this but it looks OK to me.

PR #5060, opened 2023-01-12
configure.ac: Do not probe for <avxintrin.h>

fweimer-rh:
With GCC 12, this header file cannot be included on its own. Doing so fails with #error, but also produces many warnings related to implicit ints and implicit function declarations. This is inconvenient for distributions who are trying to switch to newer C versions (without implicit ints/function declarations), and monitor builds for suspicious warnings related to obsolete C features.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC


AenBleidd (approved):
LGTM. Thank for the contribution!

PR #5073, opened 2023-01-25
client: enable CPU throttling

### LINKED ISSUE #5072: 7.22.0 CPU usage FULL THROTTLE on macOS Ventura.2
**Describe the bug**
The CPU usage rate set by the web server cannot be reflected, and the CPU usage rate becomes full throttle.

**Steps To Reproduce**
1. upgrade to 7.22.0 from 7.20.4
2. launch

**Expected behavior**
It works according to the CPU usage rate set on the web server.

**System Information**
- OS: macOS Ventura.2
- BOINC Version: 7.22.0

davidpanderson:
Somehow I had removed
#define NEW_CPU_THROTTLE in an earlier commit.
Fix a compile warning.

Fixes #5072

### DIFF
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -1197,7 +1197,6 @@ void ACTIVE_TASK::set_task_state(int val, const char* where) {
 }
 
 #ifndef SIM
-#ifdef NEW_CPU_THROTTLE
 #ifdef _WIN32
 DWORD WINAPI throttler(LPVOID) {
 #else
@@ -1249,5 +1248,4 @@ void* throttler(void*) {
     }
     return 0;
 }
-#endif
 #endif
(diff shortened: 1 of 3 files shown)


AenBleidd:
This particular fix looks ok but the original issue also includes some additional comments from @CharlieFenton that looks like separate issues. Not sure they all should be fixed via this PR, but they should be definitely reviewed and tested as well.
@CharlieFenton, could you please test this PR and if your additional comments are not fixed by it, please open a separate issue so we could track and fix those issues as well.

AenBleidd (approved):

CharlieFenton:
Sorry, I don't have any more time to work on BOINC this week. O'll try to look at it next week.

CharlieFenton:
> @CharlieFenton, could you please test this PR and if your additional comments are not fixed by it, please open a separate issue so we could track and fix those issues as well.

Please see issue #5076

PR #5074, opened 2023-01-25
client: don't print a (confusing) warning if we're using an https:// URL

davidpanderson:
for a project and a scheduler reply contains an http:// URL. Keep using https://


AenBleidd (approved):
Looks good to me.

PR #5076, opened 2023-01-26
Manager: don't clear in-use prefs when disable in-use computing

davidpanderson:
In-use prefs:
- cpu_usage_limit
- max_ncpus_pct
- ram_max_used_busy_frac
- suspend_cpu_usage
Don't clear their values when in-use computing is disabled, and write them to the override file.
That way if the user enables in-use computing again, they'll see the values they had set before.

Fixes some of Charlie's comments in #5072


AenBleidd (approved):

AenBleidd:
Code looks good to me but testing is required.

PR #5084, opened 2023-02-05
libboinc: fix include order

CyberTailor:
Compilation fails if "filesys.h" is included after "boinc_fcgi.h".

I don't know what's the reason but here's error message:

```
In file included from /usr/include/features.h:490,
                 from /usr/include/bits/libc-header-start.h:33,
                 from /usr/include/stdio.h:27,
                 from /usr/include/fcgi_stdio.h:18,
                 from boinc_fcgi.h:19,
                 from coproc.cpp:22:
/usr/include/wchar.h:582:24: error: 'malloc' attribute argument 1 is ambiguous
  582 |   __attribute_malloc__ __attr_dealloc_fclose;
      |                        ^~~~~~~~~~~~~~~~~~~~~
/usr/include/wchar.h:582:24: note: use a cast to the expected type to disambiguate
```

### DIFF
--- a/lib/coproc.cpp
+++ b/lib/coproc.cpp
@@ -18,11 +18,6 @@
 #if defined(_WIN32)
 #include "boinc_win.h"
 #else
-#ifdef _USING_FCGI_
-#include "boinc_fcgi.h"
-#else
-#include <cstdio>
-#endif
 #include <cstring>
 #include <cstdlib>
 #include <cmath>
@@ -47,6 +42,14 @@
 #include "parse.h"
 #include "str_replace.h"
 #include "util.h"
+
+#if !defined(_WIN32)
+#ifdef _USING_FCGI_
+#include "boinc_fcgi.h"
+#else
+#include <cstdio>
+#endif
+#endif
 
 #include "coproc.h"
 
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -19,7 +19,7 @@ env:
 jobs:
   build:
     name: ${{ matrix.type }}-build
-    runs-on: ubuntu-20.04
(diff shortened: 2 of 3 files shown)


lib/coproc.cpp
```diff
@@ -48,6 +43,12 @@
 #include "str_replace.h"
 #include "util.h"
 
+#ifdef _USING_FCGI_
```
AenBleidd, line 46: Previously this piece of code was for non '_WIN32' only.
Could you please keep it?

AenBleidd (changes requested):
Please specify os and compiler version where you see this error.

CyberTailor:
> Please specify os and compiler version where you see this error.

Linux
GCC 12.2.1
Glibc 2.36

AenBleidd:
There is at least one more place that is failing with the same issue: `lib/filesys.cpp`
Could you please fix that as well?
Ref:
- https://github.com/BOINC/boinc/actions/runs/4096879926/jobs/7064842121
- https://github.com/BOINC/boinc/actions/runs/4096879926/jobs/7064842000

And to test this fix please change this line https://github.com/BOINC/boinc/blob/aa31e3d2f8d7d8edfef1142f45d136dcd7fb870b/.github/workflows/linux.yml#L22
to `runs-on: ubuntu-latest`

CyberTailor:
Missed that! Fixed now (hopefully)

AenBleidd:
Still failing...

AenBleidd:
I believe I found the root cause of this issue.
I'll try to fix it separately.
For now I'll keep this PR open in case I'm wrong

AenBleidd:
Closing this in favor of #5096 that has a complete fix of the issue

PR #5085, opened 2023-02-05
Account for systems where library directory is named `lib64`

CyberTailor:
* Use `$(libdir)` in automake
* Use **GNUInstallDirs** in CMake

### DIFF
--- a/apps/Makefile.am
+++ b/apps/Makefile.am
@@ -3,7 +3,7 @@
 
 include $(top_srcdir)/Makefile.incl
 
-exampledir = $(prefix)/lib/boinc-apps-examples
+exampledir = $(libdir)/boinc-apps-examples
 example_PROGRAMS = upper_case concat 1sec
 
 upper_case_SOURCES = upper_case.cpp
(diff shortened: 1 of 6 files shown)


apps/Makefile.am
```diff
@@ -3,7 +3,7 @@
 
 include $(top_srcdir)/Makefile.incl
 
-exampledir = $(prefix)/lib/boinc-apps-examples
+exampledir = $(libdir)/boinc-apps-examples
```
AenBleidd, line 6: Doesn't this mean that after your change there is no way to use custom location for the installation?

CyberTailor:
This variable respects --prefix

AenBleidd (approved):

AenBleidd:
I believe this should be fixed, bit reverted.
@CyberTailor, I believe when installing executables, ${libexecdir} should be used instead.
Could you please take a look and fix?

CyberTailor:
> I believe this should be fixed, bit reverted. @CyberTailor, I believe when installing executables, ${libexecdir} should be used instead. Could you please take a look and fix?

I didn't note that. Yes, ${libexecdir} or ${bindir} should be used for daemon programs.

CyberTailor:
This is not urgent because some software installs executables to libdir too.

Created #5239

PR #5096, opened 2023-02-11
[linux] Move all fcgi dependent stdio functions to the separate 'boinc_stdio.h' header only file

AenBleidd:


AenBleidd:
@davidpanderson, could you please review this?
This fix is needed to fix build of libboinc and server tools when using FastFCGI with the new gcc 11 that has a breaking change of malloc prototype making it incompatible with FastCGI methods overloads.

@lfield, since this is quite a big fix that could probably touched smth that was not intended to be touched, could you please test this on the test server with FastCGI enabled?

Thank you both in advance.

AenBleidd:
@davidpanderson, could you please review this PR? Thank you in advance

AenBleidd:
> BTW, for large changes like this it's better to not include unrelated edits like trailing spaces;
that makes the diff harder to read.

That's done automatically by my code editor. Pardon, will clean it up next time.

Thank you for the review

CharlieFenton:
Unnecessary changes such as deleting trailing spaces also interferes with using GIT's _blame_ function when trying to determine when and why a line of code had functional changes, and who made the change.

PR #5099, opened 2023-02-13
[Linux][Manager][vcpkg] Compile Linux vcpkg manager in manifest mode.

talregev:
Compile Linux vcpkg manager in manifest mode.
tested on wsl2 ubuntu 22.04.

### DIFF
--- a/3rdParty/vcpkg_ports/configs/manager/linux/vcpkg.json
+++ b/3rdParty/vcpkg_ports/configs/manager/linux/vcpkg.json
@@ -1,0 +1,17 @@
+{
+    "name": "boinc-manager",
+    "version-string": "7.23.0",
+    "dependencies":
+    [
+        "ftgl",
+        {
+          "name": "curl",
+          "features": ["openssl"],
+          "default-features": false
+        },
+        {
+          "name": "wxwidgets",
+          "default-features": false
+        }
+    ]
+}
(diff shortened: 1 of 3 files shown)


talregev:
image

3rdParty/vcpkg_ports/configs/manager/vcpkg.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "boinc-msbuild-x64",
```
AenBleidd, line 2: Is this is for linux, then please change it to `boinc-manager` and put into 'linux' subdirectory

AenBleidd (changes requested):
Please update https://github.com/BOINC/boinc/blob/master/set-client-version.py script to include your new file

talregev:
We should remove all versions from vcpkg.json. there is not meaning / not effecting boinc versioning. 
I will test it in other PR.

3rdParty/vcpkg_ports/configs/manager/vcpkg.json
talregev: done.

AenBleidd (approved):

AenBleidd:
Looks good to me.
@talregev, thank you for the contribution.

PR #5109, opened 2023-02-21
[Linux][amr64][snap] Add arm64 snap build

talregev:
Add arm64 snap build.

Time to compile vcpkg dependencies is around ~6hours. When it cache this job take around 20min. 
That why now this job will fail. it exceed the 6 hours limit. 

**I am looking way to cache it before. I will happy for suggestions.**

I also reduce the x64 snap size from ~60m to ~20m.
Tested the x64 vcpkg linux version on wsl2.
Tested the x64 snap on vm and with x11.


linux/update_vcpkg_manager.sh
```diff
@@ -24,5 +24,5 @@ if [ ! -d $VCPKG_ROOT ]; then
 fi
 
 git -C $VCPKG_ROOT pull
-$VCPKG_ROOT/bootstrap-vcpkg.sh
-$VCPKG_ROOT/vcpkg install  --x-manifest-root=3rdParty/vcpkg_ports/configs/manager/linux --x-install-root=$VCPKG_ROOT/installed/ --overlay-ports=$VCPKG_PORTS/ports --overlay-triplets=$VCPKG_PORTS/triplets/ci --triplet=$TRIPLET --clean-after-build
+$VCPKG_ROOT/bootstrap-vcpkg.sh -musl
```
AenBleidd, line 27: Why this was added?
We definitely not gonna do all builds for linux with `musl` library.
This library is required for the very specific builds of linux (usually for IoT), and this is not smth we should use for regular builds

AenBleidd (changes requested):

linux/update_vcpkg_manager.sh
talregev: We don't use musl dependency to our linux. This is vcpkg binary that build with musl library that can run on arm64 and arm without compilation. It save build time.

AenBleidd: Pardon for some lack of knowledge from my side, but does this PR means that you run arm64 build on **arm64** image of linux on qemu?
If that's the case - is there a way to build arm64 on amd64 OS image?
I'm asking this because on other project that runs CI on GitHub Actions I found that arm64 running on qemu with very poor performance, and sometimes even slower than running arm (32-bit) on the same qemu.
This could be quite a bad if we need to rerun build several times to make it green while caching vcpkg binaries with several retries.
vcpkg cache is very fragile, and could be invalidated with some very minor update of the top-most dependency, that will lead to full rebuild.
Please explore other ways to build arm64 if this is possible (I think it should be possible, because I don't believe it's a strong requirement to build arm64 snap packages on arm64 OS).

talregev: > Please explore other ways to build arm64 if this is possible (I think it should be possible, because I don't believe it's a strong requirement to build arm64 snap packages on arm64 OS).

Easy to say than done. I already made the effort to compile it to this state. For now I am not going to explore cross compiling on snap builds.

AenBleidd: In this case I can't accept this PR in its state. Build performance of arm64 is too poor: after almost 30 minutes it just reached vcpkg and haven't start building yet:
image

I'm sorry, but this is unacceptable to let anyone wait for several hours to find out whether their build successful or not

AenBleidd:
Converted to 'Draft' because of the poor build performance of arm64 build

linux/update_vcpkg_manager.sh
talregev: As I say, when it cache it will take 20min. No one will wait 6h. only first time.

AenBleidd: You haven't read what I wrote, so I have to quote myself:

> This could be quite a bad if we need to rerun build several times to make it green while caching vcpkg binaries with several retries.
vcpkg cache is very fragile, and could be invalidated with some very minor update of the top-most dependency, that will lead to full rebuild.

talregev: I already read you. I don't push this PR. Also looking for contractive solution. You just looking how to be destructive. 
Also even on worlds build that is a rate thing in vcpkg, PRs can advanced and ignore when the build. 
Then it will cache again. 
Appen many time with others externals problem. The recent one is with osx builds.

AenBleidd: > I don't push this PR. I am Also looking for contractive solution regarding the caching.

> For now I am not going to explore cross compiling on snap builds.

These two statements don't look like a PR that is ready to be reviewed.

If you're still working on it - please, mark it as a draft.
This will significantly save time of reviewers.
Thank you in advance.

talregev: When I say I don't push it, I mean I don't push you to take it. Don't change my meaning. 
When I say I am not going to explore cross compiling as you ask, because you asking do it again from scratch.

AenBleidd: > When I say I am not going to explore cross compiling as you ask, because you asking do it again from scratch.

That's why before starting implementing anything you should ask me first.
As a maintainer of this project, I have experience in different fields, and either could suggest smth or at least review your idea before you start spending time on it.

So, you decided to do this task in your way. This implementation has serious issues that prevent me from accepting it in its current way, and you blame me that I'm asking you to explore other ways to do this task.
Yes, I'm asking you this, because there is no way I accept this PR in it's current state.

Don't put your errors and bad decisions on others.
We all do some mistakes, and as a professionals, we should accept that fact that sometime we just fuck up something.
And this is ok. Just no need to blame others.

talregev: I am not blaming you for asking, I just tell you I am not currently take this direction. I asked other constrictive in idea for caching. You couldn't say because you asking to do something from scratch, this PR is not ready.  
I do accept your view point that it taking long way, and you are not willing to accept. As I say before, I am not pushing you.
I suggested some workaround. You can also put some ideas if you like other then implement it from scratch.

AenBleidd: I have an idea that's flying around since a while, but I'll not like it either.
I'll just put it here in case you want to think about it once.

So, for now we have all needed dependencies built within one flow. And in case we could reuse the same dependencies in different flow (e.g. osx-manager-build and osx-samples-makefiles-build) that could save us building time significantly.
The same way we could make for Flutter that is currently disabled: we could reuse already built dependencies from another flow.
Here the same: if you want to build snap package on arm64, you don't need to actually build vcpkg on the same os, because you can cross-build it on amd64 and reuse on arm64.
The solution is to reorganize our CI flow step-by-step to use this approach: https://docs.github.com/en/actions/using-workflows/reusing-workflows
I don't ask you to do this for all out CIs (also, not all of them requires this), but this PR could be a good starting point of doing this (for snap-arm64 only, because all other flows should be done in a separate PR(s)).
What do you think about this?

talregev: I didn't understand the idea. I just briefly read the reusing workflow. It will take me time to understand.

AenBleidd: We can discuss it privately, if you want.
Briefly - split CI flow (e.g. snap-arm64) into two workflows:
- build dependencies
- build actual snap package (and reuse already build and cached dependencies from the previous separate build.
In any case, just read the document, and if you want to try it out - please let me know.
We could discuss it in details later.

talregev: Split the snap arm64 to build vcpkg dependencies, then another workflow that do the rest still will not be sufficient time, because it take more then 6hours to just for vcpkg to work. It going the most fast that it can. 
It mean that it needed to split the vcpkg dependencies work. Are you agree to this concept?

AenBleidd: I suggested to cross-build vcpkg arm64 dependencies on **amd64** machine that will be much faster, and then reuse them on arm64 os image.
Please re-read my message above one more time, so you could be on the same page with me.

talregev: I already try to test cross compiling direction. In vcpkg there are some broken packages that is not compile in cross compiling. It not going to be in the near future.

AenBleidd:
Hey, @talregev, any plans to continue working on this PR?

talregev:
@davidpanderson 
Do it as  a bounty over 1000$.

AenBleidd:
@talregev, here is the list of the current open bounties: https://github.com/TheSCInitiative/bounties/issues
Feel free to choose any, fix it and you'll get a reward.
Please take into the consideration, that these bounties are not financed by anyone from the BOINC, that is why they are located in a separate repo of the separate organization.

PR #5111, opened 2023-02-22
client: show file size in event log message when file transfer finishes

### LINKED ISSUE #5105: Feature Req: Show network data usage per project as running total over last 1,3,7,14 etc. days for current host in BOINC Manager
**Describe the problem**
Currently BOINC does provide good CPU and disk statistics per project, but it does NOT provide any network usage statistics per project. Since my internet provider implements a monthly data cap, I had set a network usage limit in BOINC Manager of 30GB per 30 days. However in less than 14 days my projects blasted through this limit.

Unfortunately there appears to be no way within BOINC Manager to determine, which of my projects were the main user of network data usage, so I had to rely on posting on the forums of the main "suspect" to find out what their network data usage typically is.

BOINC Manager should provide information on network data usage. In addition to the "Statistics" and "Disk" tab, there should be another tab "Network". On that tab it should show on a per-project basis the network data consumption for the **current host**.

This could e.g. be done as bar graphs showing running totals per day over certain intervals (e.g. 1 day, 3 days, 7 days, 14 days). E.g. if project X consumed a total of 30GB in the last 30 days and 14 GB over the last 7 days, then the UI would show a running total of 1GB/day over the last 30 days and 2GB/day over the last 14 days. (Normalizing this to the "per day" usage avoid the problem of having to show different magnitudes of numbers, at least when data usage is more-less consistent.)

While it would be nice to also see the data for **all hosts** ("host total" as it's called on the "Statistics" tab), I am assuming that this data might not be available via general BOINC statistics and that each host can only determine their own network data usage.

**Additional context**
I am currently running BOINC Manager 7.20.2 x64 under Windows 10 Pro, 21H2 with Virtual Box available for projects.

davidpanderson:
Fixes #5105 (partly)

### DIFF
--- a/client/pers_file_xfer.cpp
+++ b/client/pers_file_xfer.cpp
@@ -234,8 +234,8 @@ bool PERS_FILE_XFER::poll() {
             fip->project->file_xfer_backoff(is_upload).file_xfer_succeeded();
             if (log_flags.file_xfer) {
                 msg_printf(
-                    fip->project, MSG_INFO, "Finished %s of %s",
-                    is_upload?"upload":"download", fip->name
+                    fip->project, MSG_INFO, "Finished %s of %s (%.0f bytes)",
+                    is_upload?"upload":"download", fip->name, fip->nbytes
                 );
             }
             if (log_flags.file_xfer_debug) {


AenBleidd (approved):
Looks good to me.
 Thanks!

PR #5112, opened 2023-02-22
[osx][vcpkg] Bugfix osx vcpkg dependencies

talregev:
Bugfix osx vcpkg dependencies.

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -48,8 +48,12 @@ jobs:
           echo "VCPKG_BINARY_SOURCES=clear;x-aws-config,no-sign-request;x-aws,s3://vcpkg.cache.boinc/,read" >> $GITHUB_ENV
 
       - name: install dependencies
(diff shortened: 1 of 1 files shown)


AenBleidd:
This issue will be fixed in https://github.com/actions/runner-images/releases/tag/macOS-12%2F20230219.1
so we can just wait for the deployment and do not add a temporary solution since we have no urgency

talregev:
> This issue will be fixed in https://github.com/actions/runner-images/releases/tag/macOS-12%2F20230219.1 so we can just wait for the deployment and do not add a temporary solution since we have no urgency

I think if the image will be with this dependency, we should put it in our ci for knowledge. images can change many times.

AenBleidd:
This was a defect in their image release. No need to put this into our ci scripts even for our knowledge because this adds redundant complexity

talregev:
> This was a defect in their image release. No need to put this into our ci scripts even for our knowledge because this adds redundant complexity

I don't agree.

PR #5116, opened 2023-02-24
[CI] Change test report

talregev:
Change test report.
I made a mistake and added the test report badge to readme. This badge don't indicate if the tests are past. it just indicate if it succeeded to create the job for it. It misleading. 
So I implement it differently, more naturally. The test report now wait for the job with all matrix option, then download the artifact and create the report similar to before. But now it on the same workflow. 
The report still appear on random workflow. It an know bug that developer of the reporter need better api from github action. 
After this PR there is no need for extra workflow.


.github/workflows/android.yml
```diff
@@ -143,3 +143,21 @@ jobs:
         with:
           fail_ci_if_error: true
           verbose: false
+  report:
+    needs: build
+    runs-on: ubuntu-latest
```
AenBleidd, line 148: Do we really need to run this report on a separate build flow (build machine to be exact) and can't just run it in the same flow?

talregev: I think there is something connected to permission. That why it needed to split to 2 workflow, as suggested in the readme of the dev github test reporter.

AenBleidd (approved):
Looks good to me

talregev:
> Looks good to me

I test it more before merge.

AenBleidd:
> I test it more before merge.

Yes please, because currently looks like it's failing trying to create report

.github/workflows/android.yml
talregev: That why it cannot be on the same workflow, even as I did here:
`Resource not accessible by integration`

talregev:
> > I test it more before merge.
> 
> Yes please, because currently looks like it's failing trying to create report

I will create another PR just delete the misleading badge.

Issue #2485, opened 2018-04-27
Flatpak/Snap support


ghost:
Dear developers,

could you provide Flatpak and/or Snap support for Linux distributions? Both are new packaging formats that work distribution independently and offer sandbox isolation and other other neat features.

Regards

Jakob

TheAspens:
In the duplicate issue #2715 - @davidpanderson noted:

> email on boinc_alpha:
> 
> I am not succeeding in running BOINC with openSUSE. The program opens but does not communicate with the server. It's running on Slackware but it was not something simple. I had to follow specific instructions. Is not it possible to package BOINC using flatpak?

jakobjakobson18:
An initial flatpak manifest could look similar to the following one:
```
app-id: org.boinc.boinc
runtime: org.freedesktop.Platform
runtime-version: "18.08"
sdk: org.freedesktop.Sdk
command: ???
modules:
  - name: wxWidgets
    buildsystem: autotools
    config-opts:
      - --with-gtk
      - --disable-shared
    sources:
      - type: archive
        url: https://github.com/wxWidgets/wxWidgets/releases/download/v3.0.4/wxWidgets-3.0.4.tar.bz2
        sha256: 96157f988d261b7368e5340afa1a0cad943768f35929c22841f62c25b17bf7f0

#  - name: libnotify
#    buildsystem: autotools
#    sources:
#      - type: git
#        url: git://salsa.debian.org/gnome-team/libnotify

  - name: boinc
    buildsystem: simple
    build-commands:
      - ./_autosetup
      - ./configure --disable-server --enable-client CXXFLAGS="-O3 "
      - make
      - cd packages/generic/sea/
      - make
    sources:
      - type: git
        url: git://github.com/BOINC/boinc

```

Without the `libnotify` section the manifest succeeds in building the `wxWidgets` and going into the `boinc` build process. However in that case the libnotify library is missing.

With the `libnotify` section the manifest can't pull the libnotify git repository (even though I don't understand why).

jakobjakobson18:
As boinc provides a script that extracts the needed files, the following manifest could also be a starting point:

```
app-id: org.boinc.boinc
runtime: org.gnome.Platform
runtime-version: '3.28'
sdk: org.gnome.Sdk
command: boinc
modules:
- name: boinc
  buildsystem: simple
  build-commands:
  - ???
  - ???
  sources:
  - type: file
[...]

PR #5119, opened 2023-02-27
Add flatpack cross build

talregev:
Add flatpack cross build.


AenBleidd:
BOINC is already listed on FlatHub, thus there is no need to add the same functionality here.
Please check these links:
- https://github.com/flathub/edu.berkeley.BOINC
- https://flathub.org/apps/details/edu.berkeley.BOINC

talregev:
They use system lib not with vcpkg.

AenBleidd:
> They use system libs not with vcpkg.

We're not supporting vcpkg exclusively, and integrators are willing to build it the way they want.
Also, there is no need to have two separate builds of the same product on FlatHub.

> Also, when we change stuff here, we don't know how it effect the flatpak build.

The build on Flatpak is very simple and clear. Please check the links I provided.

AenBleidd:
@talregev,
Please before starting doing smth, check that this was not already resolved.
For example, ticket about flatpak was closed September 25 2022: https://github.com/BOINC/boinc/issues/2485
That means that there is no need to work on this topic further (at least without discussion with project maintainers first), because this will be just a waste of time.
If you think that Flatpak should be built with vcpkg instead of system libraries, you should contact Flatpak package maintainer directly and discuss this with them.
As I already wrote above, integrators could use any of the supported ways of building our product. Also, since vcpkg doesn't provide any valuable profit for end users of BOINC in this case, I see no reason to force its usage anywhere.
We are exclusively using vcpkg for Windows only, because otherwise it's almost impossible to build all dependencies manually.

So I have to ask you one more time: before working on any topic that has no open ticket - ask maintainers first. This way you'll save not only your time, but the time of maintainers as well.
And that fact that you already made this implementation doesn't give you any additional credits trying to convince me to accept smth that is not needed.

Thank you in advance.

PR #5120, opened 2023-02-28
[CI] Add emoji to ci names

talregev:
Add a nice touch to ci, by adding emoji to ci names.


AenBleidd:
This is useless and distracting change, that gives no benefit.
We already have a clear name of every CI flow, there is no need to make a duplication with any emoji.

PR #5127, opened 2023-03-06
[ci] Fix: vcpkg will clone from $VCPKG_LINK

talregev:
Fix: vcpkg will clone from $VCPKG_LINK.


AenBleidd (approved):
LGTM. Thanks for the fix

PR #5135, opened 2023-03-11
boinc_cmd: add --get_task_summary command.

### LINKED ISSUE #5130: boinccmd --get_tasks simpler formatting
**Describe the problem**
I would like to easily be able to see a list of running tasks with 1 line per task

Maybe an additional command like `--get_tasks_list` that returns something like this:

```
$ boinccmd --get_tasks_list
WU Name ------- Deadline ----------- State ----------- % Done
wu_1234567 ---- 1/1/1/ 11:11:11 ---- Uploading ------- 100%
wu_7890123 ---- 1/1/1/ 11:11:11 ---- Executing ------- 90%
wu_4567890 ---- 1/1/1/ 11:11:11 ---- Executing ------- 50%
wu_1234567 ---- 1/1/1/ 11:11:11 ---- Pending --------- 0%
```

davidpanderson:
Shows task summaries, one per line

Fixes #5130

### DIFF
--- a/client/boinc_cmd.cpp
+++ b/client/boinc_cmd.cpp
@@ -206,7 +206,7 @@ typedef vector<const char*> STR_LIST;
 //
 void show_str_lists(vector<STR_LIST> &lines, size_t ncols) {
     vector<int> lengths;
-    uint i;
+    size_t i;
     for (i=0; i<ncols; i++) {
         size_t max = 0;
         for (const STR_LIST& s: lines) {
--- a/client/boinc_cmd.cpp
+++ b/client/boinc_cmd.cpp
@@ -188,6 +188,111 @@ void show_alerts(RPC_CLIENT &rpc) {
     }
 }
 
+////////////// --get_task_summary start ////////////////
+
+typedef vector<const char*> STR_LIST;
+
+// given: a list of lines, each consisting of n columns
+// for each column: find the longest entry.
+// Then display the lines so the columns line up.
+//
+void show_str_lists(vector<STR_LIST> lines, int ncols) {
+    vector<int> lengths;
+    int i;
+    for (i=0; i<ncols; i++) {
+        size_t max = 0;
+        for (STR_LIST& s: lines) {
+            max = std::max(max, strlen(s[i]));
+        }
+        lengths.push_back(max);
+    }
+    for (STR_LIST &line : lines) {
+        for (i=0; i<ncols; i++) {
+            printf("%-*s  ", lengths[i], line[i]);
+        }
+        printf("\n");
+    }
+}
+
+// given 
+// show list of tasks, 1 line per task,
+// similar to the Manager's Tasks pane
+// fields per task:
+// - project name
+// - progress %
+// - elapsed time
+// - deadline
+// - status
+// - resource usage
+// - WU name
+//
+
+struct RESULT_INFO {
+    char project_name[256];
+    char pct_done[256];
+    char elapsed_time[256];
+    char deadline[256];
+    char status[256];
+    char resource_usage[256];
+    char wu_name[256];
+};
+#define NCOLS 7
+
+int show_task_summary(RPC_CLIENT &rpc) {
+    int retval;
+    PROJECTS ps;    // need for project names
+    retval = rpc.get_project_status(ps);
+    if (retval) return retval;
+    RESULTS results;
+    retval = rpc.get_results(results);
+    if (retval) return retval;
+    vector<STR_LIST> lines;
+    STR_LIST title;
+    title.push_back("Project");
+    title.push_back("% Done");
+    title.push_back("Elapsed");
+    title.push_back("Deadline");
+    title.push_back("Status");
+    title.push_back("Procs");
+    title.push_back("WU name");
+    lines.push_back(title);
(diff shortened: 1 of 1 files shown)


bdwyer2 (commented):
Thanks for the fast PR!

client/boinc_cmd.cpp
```diff
@@ -188,6 +196,167 @@ void show_alerts(RPC_CLIENT &rpc) {
...
 
+////////////// --get_task_summary start ////////////////
+
+typedef vector<const char*> STR_LIST;
+
+// given: a list of lines, each consisting of n columns
+// for each column: find the longest entry.
+// Then display the lines so the columns line up.
+//
+void show_str_lists(vector<STR_LIST> lines, int ncols) {
```
AenBleidd, line 207: 'vector' is not a trivial object, so in this case it can't be automatically moved by the compiler, and a copy will be created.
I suggest to pass lines as a constant reference:
`const vector<STR_LIST>& lines)`

AenBleidd, line 207: I believe `ncols` will always be greater than 0. I suggest changing it's type to `size_t`

client/boinc_cmd.cpp
```diff
@@ -188,6 +196,167 @@ void show_alerts(RPC_CLIENT &rpc) {
...
+    vector<int> lengths;
+    int i;
+    for (i=0; i<ncols; i++) {
+        size_t max = 0;
+        for (STR_LIST& s: lines) {
```
AenBleidd, line 212: since there is no plans to change the value of `s` I suggest to make it `const`

client/boinc_cmd.cpp
```diff
@@ -188,6 +196,167 @@ void show_alerts(RPC_CLIENT &rpc) {
...
+            max = std::max(max, strlen(s[i]));
+        }
+        lengths.push_back(max);
+    }
+    for (STR_LIST &line : lines) {
```
AenBleidd, line 217: since there is no plans to change the value of `len` I suggest to make it `const`

client/boinc_cmd.cpp
```diff
@@ -188,6 +196,167 @@ void show_alerts(RPC_CLIENT &rpc) {
...
+        case 'w':
+            break;
+        default:
+            printf("Invalid options %s\n", options.c_str());
+            printf("Options:\n"
+                "   p: project name\n"
+                "   c: completion %%\n"
+                "   e: elapsed time\n"
+                "   d: deadline\n"
+                "   s: statue\n"
```
AenBleidd, line 267: `statue` -> `status`

client/boinc_cmd.cpp
```diff
@@ -188,6 +196,167 @@ void show_alerts(RPC_CLIENT &rpc) {
...
+                "   r: processor usage\n"
+                "   w: workunit name\n"
+            );
+            exit(1);
+        }
+    }
+}
+
+int show_task_summary(RPC_CLIENT &rpc, string &options) {
```
AenBleidd, line 276: It's better to explicitly declare both parameters as `const`

client/boinc_cmd.cpp
```diff
@@ -188,6 +196,167 @@ void show_alerts(RPC_CLIENT &rpc) {
...
+    int retval;
```
AenBleidd, line 277: This variable declaration could be safely moved to the line 280

AenBleidd (changes requested):

davidpanderson:
I added most of these.  One of them produced an error:
boinc_cmd.cpp: In function 'int show_task_summary(const RPC_CLIENT&, const string&)':
boinc_cmd.cpp:279:41: error: passing 'const RPC_CLIENT' as 'this' argument discards qualifiers [-fpermissive]
     int retval = rpc.get_results(results);
                                         ^
In file included from boinc_cmd.cpp:38:0:
../lib/gui_rpc_client.h:693:9: note:   in call to 'int RPC_CLIENT::get_results(RESULTS&, bool)'
     int get_results(RESULTS&, bool active_only = false);
         ^~~~~~~~~~~
boinc_cmd.cpp:298:43: error: passing 'const RPC_CLIENT' as 'this' argument discards qualifiers [-fpermissive]
         retval = rpc.get_project_status(ps);
                                           ^

davidpanderson:
I generally avoid qualifiers like const, unsigned etc.  I've never seen a bug that they would have prevented, and they make the code longer.

AenBleidd:
@davidpanderson, I generally agree with your arguments, also they were mostly recommendations.
BOINC codebase in overall have mixed code style, so it's fine to have some code that could be done better but with more refactoring, that is definitely out of scope of this task.
Please fix build error, and I'll merge it since I'm ok with this PR.

davidpanderson:
It builds now (I didn't add the const to RPC_CLIENT, which is what caused the problem)

AenBleidd:
It fails on Windows: https://github.com/BOINC/boinc/actions/runs/4396431268/jobs/7698857928

AenBleidd:
@davidpanderson, it fails because 'uint' is not a variable type on Windows, you should use 'size_t' instead.

AenBleidd (approved):

PR #5144, opened 2023-03-13
client, Win: add a cc_config.xml option <no_rdp_check>.

### LINKED ISSUE #5143: Remote desktop disables GPU, but it doesn't need to
**Describe the bug**
Boinc disables the GPU when Microsoft Remote Desktop Connection is in use.  But according to this webpage, it doesn't have to:
https://knowledge.civilgeo.com/knowledge-base/enabling-gpu-rendering-for-microsoft-remote-desktop/

It says you can enable GPU computing for a remote desktop, and I've checked, it works ok with Folding@Home, and if a Boinc GPU task is already running it keeps running ok for a minute until Boinc disables it.

This is nothing to do with the other report I made [#5132](https://github.com/BOINC/boinc/issues/5132), this one is not to do with multiple Boinc instances.

**Steps To Reproduce**
1. Follow instructions on https://knowledge.civilgeo.com/knowledge-base/enabling-gpu-rendering-for-microsoft-remote-desktop/ for a host running GPU tasks in Boinc.
2. Connect a remote desktop session to the host.

**Expected behavior**
Boinc should continue processing GPU tasks, but instead Boinc gives the error "Remote desktop in use; disabling GPU tasks".

**System Information**
 - OS: Windows 11
 - BOINC Version: 7.20.2

davidpanderson:
If set, it skips the check to see if Remote Desktop Protocol (RDP) is in use, allowing GPU apps to run while using RDP.
This should be done only if you have configured RDP as described here: https://knowledge.civilgeo.com/knowledge-base/enabling-gpu-rendering-for-microsoft-remote-desktop/

Fixes #5143

### DIFF
--- a/client/cpu_sched.cpp
+++ b/client/cpu_sched.cpp
@@ -269,6 +269,7 @@ bool gpus_usable = true;
 //
 bool check_coprocs_usable() {
 #ifdef _WIN32
+    if (cc_config.no_rdp_check) return false;
     unsigned int i;
     bool new_usable = !is_remote_desktop();
     if (gpus_usable) {
--- a/client/log_flags.cpp
+++ b/client/log_flags.cpp
@@ -265,6 +265,9 @@ void CC_CONFIG::show() {
     if (no_priority_change) {
         msg_printf(NULL, MSG_INFO, "Config: run apps at regular priority");
     }
+    if (no_rdp_check) {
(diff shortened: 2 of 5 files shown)


AenBleidd (approved):

hucker75:
Actually, it appears you don't have to follow all those steps. I just Remote Desktop connected to another machine where I hadn't followed the steps, and Folding@Home continued to function. They also use OpenCL, so I guess Boinc would keep working too.

PR #5147, opened 2023-03-14
errorcode returns add

### LINKED ISSUE #4178: Handle leak in launcher code
Successful calls to `CreateProcess` and `CreateProcessAsUser` return a thread and process handle to the caller via the `ProcInfo` parameter. These need to be explicitly closed otherwise they will be leaked. 

https://github.com/BOINC/boinc/blob/03b2e00990b54caab9a2d725104f1f69b469b505/clientsetup/win/launcher.cpp#L245

https://github.com/BOINC/boinc/blob/03b2e00990b54caab9a2d725104f1f69b469b505/clientsetup/win/launcher.cpp#L278

See remarks here:
https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/ns-processthreadsapi-process_information

mcagriaksoy:
Fixes #4178

**Description of the Change**
immediate return add after fault

**Alternate Designs**
does not have idea

**Release Notes**
immediate return add after fault

### DIFF
--- a/clientsetup/win/launcher.cpp
+++ b/clientsetup/win/launcher.cpp
@@ -256,7 +256,7 @@ HRESULT CreateProcessWithExplorerIL(LPWSTR szProcessName, LPWSTR szCmdLine)
                                 &ProcInfo
                             );
                             if(!bRet) {
-								hr = HRESULT_FROM_WIN32(GetLastError());
+								return(HRESULT_FROM_WIN32(GetLastError()));
                             }
 						}
                         CloseHandle(hNewToken);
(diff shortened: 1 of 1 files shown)


clientsetup/win/launcher.cpp
```diff
@@ -256,7 +256,7 @@ HRESULT CreateProcessWithExplorerIL(LPWSTR szProcessName, LPWSTR szCmdLine)
                                 &ProcInfo
                             );
                             if(!bRet) {
-								hr = HRESULT_FROM_WIN32(GetLastError());
+								return(HRESULT_FROM_WIN32(GetLastError()));
```
AenBleidd, line 259: At this moment hNewToken, hToken, hProcess won't be properly released

AenBleidd (changes requested):

PR #5148, opened 2023-03-14
[Android][CI] Split build to run faster and do less rebuilds

AenBleidd:


.github/workflows/android.yml
```diff
@@ -35,11 +35,319 @@ jobs:
...
+      - name: Setup Android SDK
+        uses: android-actions/setup-android@v2
+
+      - name: Cache dependencies
+        uses: actions/cache@v3
+        with:
+          path: |
+            3rdParty/buildCache
+            !3rdParty/buildCache/android/vcpkgcache/
+          key: android-vcpkg-${{ matrix.arch }}-${{ hashFiles('android/*.sh', '.github/workflows/android.yml') }}
```
talregev, line 97: We don't use cache in vcpkg.

.github/workflows/android.yml
```diff
@@ -35,11 +35,308 @@ jobs:
...
+        run: python ./deploy/prepare_deployment.py logs
+
+      - name: Upload logs on failure
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: android_logs_dependencies_${{ matrix.arch }}_${{ github.event.pull_request.head.sha }}
+          path: deploy/logs.7z
+
+  build-dependencies-vcpkg:
```
talregev, line 70: In case in vcpkg, don't use dependency. If the ports not install, it will make them compile twice.

talregev: Compile twice
image

talregev:
Compile twice:
image

talregev:
I still not recommend on this split. On the practical level it swarm boinc repo with a  lot of vm, and it will starve PR that will wait longer.

Also You should remove the native ones. You don't use them anyway. For prove you still use openssl 1.1.1 on native.

AenBleidd:
> I still not recommend on this split. On the practical level it swarm boinc repo with a lot of vm, and it will starve PRs that will wait longer.

You might be correct. I was thinking to split this to make possible in future to build Flutter, but probably it would be easier to do it with vcpkg, and this split become unneeded at all.

> Also You should remove the native ones. You don't use them anyway. For proof: you still use openssl 1.1.1 on native.

We still support two ways of building Android BOINC: with vcpkg (official release) and without vcpkg (e.g. F-Droid).

talregev:
You should convert F-droid compile with vcpkg.

AenBleidd:
> You should convert F-droid compile with vcpkg.

I won't do that. If you want to do that - go ahead.

talregev:
> > You should convert F-droid compile with vcpkg.
> 
> I won't do that. If you want to do that - go ahead.

I can do that. You should approach me and be nicer.

PR #5149, opened 2023-03-15
Web: remove XSS vulnerability.

davidpanderson:
team_members.php and top_*.php have a textual parameter "sort"by". Validate it.


lfield:
Tested on LHC@home

PR #5153, opened 2023-03-17
fix potential bug in building process tree on Win

### LINKED ISSUE #5152: Task descendants can be calculated incorrectly due to PID recycling on Windows
Windows [readily](https://devblogs.microsoft.com/oldnewthing/20110106-00/?p=11813 "Old New Thing") recycles process IDs, and does not update a running process’s ‘parent PID’ when its parent exits. This can lead to the situation where BOINC thinks that unrelated processes are descendants of tasks it is managing, which causes trouble in two ways:

1. When a task is killed, `kill_descendants()` kills things it shouldn’t.
_Example:_ At logon, `userinit.exe` (PID 1000) launches `explorer.exe` (PID 1234), then exits. Some time later, a BOINC task happens to receive PID 1000. Killing that task also kills Explorer.

2. CPU time is calculated incorrectly, which can prevent checkpointing from working properly.
_Example:_ At startup, `boinc.exe` (PID 2000) spawns itself (creating PID 2345) and exits to return control to the invoking console. Some time later, a BOINC task happens to receive PID 2000. Computing that task’s CPU time counts the CPU time of all descendants of the BOINC client, which could be a huge value. As tasks finish, subsequent calculations happen to compute a smaller CPU time for PID 2000, but these are never applied (because of the check in `TASK::cpu_time()` in the wrapper), meaning `checkpoint_cpu_time` appears not to change, which prevents the task state file getting created/​updated in `ACTIVE_TASK_SET::get_msgs()`.

One possible fix might be to consider process creation time when building the process tree, and only allowing parent-child relationships to be established where the child is younger than the parent. (Though this could break if time goes backwards due to clock correction.)

davidpanderson:
client, win: when building the global process tree, don't make a parent->child link if the parent is newer than the child (this can happen in Win because a process's parentID isn't changed when its parent exits).
This prevents two potential problems:
- when killing a runaway app, killing unrelated processes too
- incorrect CPU time accounting

Fixes #5152

### DIFF
--- a/lib/procinfo.cpp
+++ b/lib/procinfo.cpp
@@ -100,6 +100,25 @@ void find_children(PROC_MAP& pm) {
         int parentid = i->second.parentid;
         PROC_MAP::iterator j = pm.find(parentid);
         if (j == pm.end()) continue;    // should never happen
+#ifdef _WIN32
+        // In Windows:
+        // 1) PIDs are reused, possibly quickly
+        // 2) if a process creates children and then exits,
+        //    the parentID of the children are not cleared,
+        //    so they may soon refer to an unrelated process.
+        //    (this is horrible design, BTW)
+        // This can cause problems:
+        // - when we abort a BOINC app we kill the process and its descendants
+        //   (based on parent ID).  These could be unrelated processes.
+        // - If a BOINC app gets a process ID that is the parentID of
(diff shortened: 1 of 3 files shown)


AenBleidd (approved):

PR #5156, opened 2023-03-18
[CI][Android] Deprecated android native in ci

talregev:
After you try to split android in this PR #5148
It become harder to maintain both native and vcpkg build for android.
As you say in this PR:
vcpkg it the official release, and the native you don't maintain anymore. For evidence, the native android still compile with openssl 1.1.1.
Now when Fdroid is switch to vcpkg build too, we should try to minimize our ci smaller and easier to maintain android builds.
https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/edu.berkeley.boinc.yml

@davidpanderson for your attention.


.github/workflows/android.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        type: [manager, libs, apps, manager-vcpkg, libs-vcpkg, apps-vcpkg, libs-cmake]
+        type: [manager-vcpkg, libs-vcpkg, apps-vcpkg, libs-cmake]
```
AenBleidd, line 25: Since there are no more native builds, there is no reason to keep '-vcpkg' suffix.
Please remove

.github/workflows/android.yml
```diff
@@ -67,19 +67,6 @@ jobs:
         run: |
           echo ANDROID_NDK_ROOT="" >> $GITHUB_ENV
 
-      - name: Build manager
-        if: ${{ success() && matrix.type == 'manager' }}
-        run:  |
-          ./android/ci_build_manager.sh
```
AenBleidd, line 73: This and other files are not used anymore but still present it the repo. Please do a cascade removal of all outdated scripts

AenBleidd (changes requested):

.github/workflows/android.yml
talregev: Clean the scripts can done in other PR.

AenBleidd: Why not to clean them in one PR since they are not used anymore?

talregev: There more scripts for android that can remove, and they need to remove and discuss in different PR.

AenBleidd: I kindly ask you to remove outdated scripts in this PR

AenBleidd (changes requested):
Please remove these files as well:

build_openssl_armv6.sh
build_curl_armv6.sh
build_openssl_arm.sh
build_curl_arm.sh
build_openssl_arm64.sh
build_curl_arm64.sh
build_openssl_x86.sh
build_curl_x86.sh
build_openssl_x86_64.sh
build_curl_x86_64.sh

Also, please go though all scripts and remove `--with-vcpkg` parameter and clean-up all code that is related to native build.

AenBleidd:
Marking this PR as a draft until the code related to native builds is not removed from the scripts

android/buildAndroidBOINC-CI.sh
```diff
@@ -239,54 +205,52 @@ getTripletName()
...
+if [ ! -d "$VCPKG_ROOT" ]; then
+    mkdir -p $BUILD_DIR
+    git -C $BUILD_DIR clone $VCPKG_LINK
+fi
+if [ ! -e /tmp/vcpkg_updated ]; then
+    git -C $VCPKG_ROOT pull
+    $VCPKG_ROOT/bootstrap-vcpkg.sh
+    touch /tmp/vcpkg_updated
+fi
+if [ $arch = "armv6" ]; then
```
AenBleidd, line 229: In this block of code, only `armv6` (because of the NDK folder) and `arm` (because of the additional `neon` configuration) are different. Others could be safely combined in one statement

AenBleidd (changes requested):

android/buildAndroidBOINC-CI.sh
```diff
@@ -22,8 +22,6 @@ set -e
 # When you want to invalidate openssl and curl without change their versions.
 export REV=1
```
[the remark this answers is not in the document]
talregev: It for ndk changes.

android/buildAndroidBOINC-CI.sh
talregev: It should remain for clearance of the script. Script also should be readable and not supper efficient.

android/buildAndroidBOINC-CI.sh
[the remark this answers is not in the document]
AenBleidd: Yes, but why do we still need revisions for the same version of the NDK?

android/buildAndroidBOINC-CI.sh
AenBleidd: Less code duplication is still better, and makes the script clearer

AenBleidd (approved):
Looks good to me now. Thanks for the PR.

PR #5157, opened 2023-03-18
[Android] Remove hard coded Android manifest string

talregev:
Remove hard coded Android manifest string.


AenBleidd (approved):

PR #5160, opened 2023-03-20
web: fix problems with download page on project sites

### LINKED ISSUE #5129: BOINC Linux Download Page is confusing
**Describe the bug**
The BOINC download page (https://boinc.berkeley.edu/download.php) for Linux is giving users information that IMO is incorrect. The big green "download BOINC" button links to the Berkeley installer but my understanding is that this installer is only suggested if user's don't have a package manager as the next line says, in smaller text, "We recommend that instead you[ install BOINC using a package manager, run it using Docker, or use a bootable ISO image". Most Linux users are using an OS that comes with a package manager, that should be the default option represented in the big green button, especially since package manager will automatically update BOINC if a security release is issued.

I am not aware if the packages in the Debian or Ubuntu repos currently work, I know the PPA doesn't work with Mint and I suspect Ubuntu as well, so until that is fixed that option should be removed as it doesn't work out of the box.  (see https://github.com/BOINC/boinc/issues/5124)

**Suggested Resolution**

- Big green button should link to https://boinc.berkeley.edu/wiki/Installing_BOINC#Linux
- Otherwise make it clear which is the suggested and best supported option for Linux users, depending on their distro
- Install on Ubuntu page should be updated to use apt instead of aptitude, I believe aptitude is deprecated https://boinc.berkeley.edu/wiki/Installing_BOINC_on_Ubuntu
- Suggest language change "To install BOINC on Ubuntu, type: " -> "To install BOINC on Ubuntu, type the following commands in the terminal: " with each command being on a separate line so users unfamiliar with apt do not think it is a single command
- Synaptic package manager is no longer installed by default on Ubuntu, so should be removed
- Potentially include boinc-client-opencl boinc-client-nvidia-cuda as suggested packages to install. My understanding is that they won't harm anything if installed and un-needed, but may be otherwise critical for BOINC to use the GPU. Users may not know which GPU they have.

davidpanderson:
The Linux installers on the BOINC web site (.sh files) have out-of-date binaries, and are not real installers. We shouldn't direct users to them.
But projects and AMs that use auto-attach were directing Linux users to these installers.

This change fixes that.
Linux users are not shown a download button, but are instead directed to https://boinc.berkeley.edu/wiki/Installing_on_Linux which lists various options for installing BOINC on Linux (starting with package managers).

Also: the download page (download_software.php) is also used by account managers (Science United).
In this case, it should show instructions for
adding an account manager, not attaching to a project.

Note: there's an analogous update to download.php on the BOINC web site: if the machine is Linux,
show a link to the Linux instructions rather than a download button.

Fixes #5129

### DIFF
--- a/html/inc/util.inc
+++ b/html/inc/util.inc
@@ -900,19 +900,23 @@ function current_url() {
 // @param text The text to display on the button
 // @param desc The title of the destination - typically used as a popup
 // @param class The optional CSS class of the button. Defaults to a standard button
+// @params extra Additional text in href tag
 //
 
-function button_text($url, $text, $desc=null, $class="btn-success btn-sm") {
+function button_text($url, $text, $desc=null, $class=null, $extra='') {
     if (!$desc) {
         $desc = $text;
     }
-    return sprintf(' <a href="%s" title="%s" class="btn %s">%s</a>',
-        $url, $desc, $class, $text
+    if (!$class) {
+        $class = "btn-success btn-sm";
+    }
+    return sprintf(' <a href="%s" title="%s" class="btn %s" %s>%s</a>',
+        $url, $desc, $class, $extra, $text
     );
 }
 
-function show_button($url, $text, $desc=null, $class="btn-success btn-sm") {
-    echo button_text($url, $text, $desc, $class);
+function show_button($url, $text, $desc=null, $class=null, $extra=null) {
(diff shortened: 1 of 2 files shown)


davidpanderson:
The installation procedure apparently doesn't work in some cases: #5129 

What's going on?

PR #5161, opened 2023-03-22
API: print Y/M/D as well as time in log messages

### LINKED ISSUE #5155: Add date to task log messages
[`boinc_msg_prefix()`](https://github.com/BOINC/boinc/blob/master/api/boinc_api.cpp#L277 "boinc_api.cpp") embeds the time of day into log messages. For diagnosing issues with long tasks that run for multiple days, it could be useful to have the date in there as well. (Preferably in [ISO format](https://xkcd.com/1179/ "xkcd" ) to avoid ambiguity: `%Y-%m-%d`, or `%F` if you can be sure `strftime()` everywhere supports it.)

davidpanderson:
Fixes #5155

### DIFF
--- a/api/boinc_api.cpp
+++ b/api/boinc_api.cpp
@@ -274,7 +274,7 @@ char* boinc_msg_prefix(char* sbuf, int len) {
         strlcpy(sbuf, "localtime() failed", len);
         return sbuf;
     }
-    if (strftime(buf, sizeof(buf)-1, "%H:%M:%S", tmp) == 0) {
+    if (strftime(buf, sizeof(buf)-1, "%F %H:%M:%S", tmp) == 0) {
         strlcpy(sbuf, "strftime() failed", len);
         return sbuf;
     }


AenBleidd (approved):
Looks good to me.
Thank you for the fix.

BrianNixon:
Does `WRAPPER_RELEASE` need to be bumped for a change like this?

AenBleidd:
> Does `WRAPPER_RELEASE` need to be bumped for a change like this?

Only when making new release. Not sure this worth to make it now since there are no other significant changes, and this is not a critical bug fix.

@davidpanderson, What do you think?
If you want, I can make a new release of BOINC Wrapper.

BrianNixon:
> Only when making new release.

Fair enough.

> If you want, I can make a new release of BOINC Wrapper.

I’m not pushing for a release; just wanted to check that an update hadn’t gone overlooked…

(Some SCM policies require _every_ change to a component to change its version number, otherwise you can end up with two copies of a program claiming to be the same version but producing different output.)

BrianNixon:
#6363

PR #5167, opened 2023-03-26
client: improve file checking on task start

### LINKED ISSUE #5165: Get rid of verify_files_on_app_start
This is a project flag (in config.xml, send in scheduler replies) that tells the client to verify
the MD5 of all files (executable and input) prior to starting each job.

If files are big (e.g. VM images) this produces a big burst of disk I/O.
Disk I/O isn't prioritized so this can cause the system to become unresponsive.
Also, it's synchronous (the client doesn't overlap it with RPCs)
so the BOINC Manager shows a "communicating with client" while it's happening.

The file check doesn't accomplish anything.
Apparently people thought it would prevent users from maliciously replacing project executables.
But it doesn't do this.
Users can substitute whatever file they want, and modify client_state.xml to change
the nbytes and md5_cksum.

So let's just get rid of it.

davidpanderson:
When a task starts:
- don't re-checksum its files.
    This was already done after the file was downloaded.
    Doing it again doesn't accomplish anything,
    and causes performance problems if the file is big.
- do check that the files exist and have the right size
- if the program launch (e.g. exec()) fails,
    re-checksum the app version files on the (unlikely) chance
    that one of them was accidentally corrupted.

Fixes #5165

### DIFF
--- a/client/app_start.cpp
+++ b/client/app_start.cpp
@@ -540,7 +540,6 @@ int ACTIVE_TASK::start(bool test) {
     char cmdline[80000];    // 64KB plus some extra
     unsigned int i;
     FILE_REF fref;
-    FILE_INFO* fip;
     int retval;
     APP_INIT_DATA aid;
 #ifdef _WIN32
@@ -569,18 +568,14 @@ int ACTIVE_TASK::start(bool test) {
 
     // make sure the task files exist
     //
-    fip=0;
+    FILE_INFO* fip = 0;
     retval = gstate.task_files_present(result, true, &fip);
     if (retval) {
-        if (fip) {
-            snprintf(
-                buf, sizeof(buf),
-                "Input file %s missing or invalid: %s",
-                fip->name, boincerror(retval)
-            );
-        } else {
-            safe_strcpy(buf, "Input file missing or invalid");
-        }
+        snprintf(
+            buf, sizeof(buf),
+            "Task file %s: %s",
+            fip->name, boincerror(retval)
+        );
         goto error;
     }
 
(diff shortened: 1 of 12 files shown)


client/app_start.cpp
```diff
@@ -567,21 +567,21 @@ int ACTIVE_TASK::start(bool test) {
...
-            goto error;
+    // make sure the task files exist
+    //
+    fip=0;
+    retval = gstate.task_files_present(result, true, &fip);
+    if (retval) {
+        if (fip) {
+            snprintf(
+                buf, sizeof(buf),
+                "Input file %s missing or invalid: %s",
```
AenBleidd, line 578: Since we're not verify the content of the file, this message is misleading. Should it be changed to:
`Input file %s is missing: %s` ?

client/app_start.cpp
```diff
@@ -567,21 +567,21 @@ int ACTIVE_TASK::start(bool test) {
...
+                fip->name, boincerror(retval)
+            );
+        } else {
+            safe_strcpy(buf, "Input file missing or invalid");
```
AenBleidd, line 582: Here's the same

AenBleidd (changes requested):

BrianNixon (commented):
`CLIENT_STATE::task_files_present()`: Suggest mentioning the side-effect of `check_size`, otherwise it’s not obvious why the early returns get removed

client/cs_files.cpp
```diff
@@ -466,6 +466,34 @@ bool CLIENT_STATE::create_and_delete_pers_file_xfers() {
 }
 #endif
 
+// check whether file exists and has the right size
```
BrianNixon, line 469:
```suggestion
// Check whether file exists and has the right size.
// If the size on disk does not match the expected size, delete the file.
```

AenBleidd (approved):

PR #5170, opened 2023-03-27
[Windows Installer] Prevent handle leaks

### LINKED ISSUE #4178: Handle leak in launcher code

BrianNixon:
Fixes #4178

### DIFF
--- a/clientsetup/win/launcher.cpp
+++ b/clientsetup/win/launcher.cpp
@@ -255,7 +255,10 @@ HRESULT CreateProcessWithExplorerIL(LPWSTR szProcessName, LPWSTR szCmdLine)
                                 &StartupInfo,
                                 &ProcInfo
                             );
-                            if(!bRet) {
+                            if(bRet) {
+                                CloseHandle(ProcInfo.hThread);
+                                CloseHandle(ProcInfo.hProcess);
+                            } else {
 								hr = HRESULT_FROM_WIN32(GetLastError());
                             }
 						}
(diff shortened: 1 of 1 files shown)


talregev:
Is it interesting you to do this research?

BrianNixon:
> Is it interesting you to do this research?

At the moment I don’t know enough about installers, or BOINC’s setup needs, to be able to take on that task.

I just picked an easy ‘Newbie’ bug to fix…

talregev:
> > Is it interesting you to do this research?
> 
> At the moment I don’t know enough about installers, or BOINC’s setup needs, to be able to take on that task.
> 
> I just picked an easy ‘Newbie’ bug to fix…

Let me know when you feel ready. Thank you for your fix!

AenBleidd (approved):
Looks good to me

talregev:
> > As you pointed out in the issue you don't have time for that. So stop blocking others for helping the boinc community.
> 
> I don't stop anybody. I'm just saying that this is a complex task, and it's probably better if I do this. If somebody else has experience - they are always welcome to contribute.

If you say it probably should done by me, it blocking people from learning and give the information for windows installer.

AenBleidd:
> If you say it probably should done by me, it blocking people from learning and give the information for windows installer.

At first, this is not a correct place for such a discussion.
Second, I expressed my opinion, because this is really something that could be more effectively done by me rather than by someone who just starts to learn BOINC codebase. But I still don't block anyone from doing that.

Issue #5168, opened 2023-03-26
Exclusive app is running... Show which app.


4e576rt8uh9ij9okp:
**Describe the problem**
So BOINC wasn't running over night because apparently an exclusive app was running so no credits for me sadly.

**Describe the solution you'd like**
I would like to see the exclusive app's name being logged into the log feed so I can see what is causing the stop.

**Additional context**
I use my PC also as a gaming PC so having the exclusive app' being logged into the feed can help me debug the issue.

davidpanderson:
This is shown with the "mem_usage_debug" flag, but that shows a bunch of other stuff too, and it shows it every 10 seconds.

We should show a message only for the start and end of the execlusive app period.
What option should this be under?  Maybe cpu_sched?

AenBleidd:
> What option should this be under?  Maybe cpu_sched?

I believe this is a good flag to show such information

PR #5172, opened 2023-03-28
client: print log message (conditioned on log_flags.task) when an exclusive app or exclusive GPU app starts or stops.

davidpanderson:
Fixes: #5168


AenBleidd (approved):
Looks good to me

PR #5173, opened 2023-03-29
Client, Win: include kernel time in non-BOINC CPU usage

### LINKED ISSUE #3935: BOINC isn't suspending computation when non-BOINC cpu usage is high
**Describe the bug**
BOINC isn't suspending computation when non-BOINC cpu usage is high, despite having settings to suspend computation to that effect.

**Steps To Reproduce**
1. Set BOINC up:
    - Use local preferences
    - Use 50% of processors on a high core count machine
    - 100% of cpu time used.
2. Set "suspend when non-boinc cpu usage is above": 30% (also tried 50%)
3. Launch a heavy-load, multi-core task

**Expected behavior**
BOINC should suspend computation when the heavy load task runs

**Screenshots**
image

**System Information**
 - OS: Windows Server 2008 R2
 - BOINC Version: 7.16.7

**Additional context**
Hardware: 2x 6c/12t Xeon L5640 (24 logical processors total)
60GB RAM

This machine is used for multithreaded compilation which bumps CPU usage outside of BOINC to 95%+ when active, and BOINC should suspend, but it doesn't. Even if fully loaded, BOINC continues to compute.
Even suspending and resuming manually while the task is running resumes the BOINC computation tasks when CPU usage is already over 95%.
It happens irrespective of which projects attached to (seems to be a clear BOINC scheduling problem)

davidpanderson:
Fixes #3935

### DIFF
--- a/lib/procinfo.cpp
+++ b/lib/procinfo.cpp
@@ -124,6 +124,8 @@ void find_children(PROC_MAP& pm) {
 }
 
 // get resource usage of non-BOINC apps
+// NOTE: this is flawed because it doesn't account for short-lived processes.
+// It's not used on Win, Mac, or Linux, which have better ways of getting total CPU usage.
 //
 void procinfo_non_boinc(PROCINFO& procinfo, PROC_MAP& pm) {
     procinfo.clear();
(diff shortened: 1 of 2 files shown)


AenBleidd (approved):
Looks good to me

PR #5176, opened 2023-03-31
[Manager] Show native names for language options, and only those for which translations are available

### LINKED ISSUE #2061: Show language options in actual language, not all in English (only)
This is a request to make the Options->Other options->Language drop down menu show the optional languages in those languages instead of all in English. Or show them in dual language. For example Deutsch (German), Nederlands (Dutch) , Français (French) etc. 

For consistency this is also needed at the web sites.

BrianNixon:
Fixes #2061

**Release Notes**
Manager ‘Other options’ dialog now shows native language names as well as English, and only offers languages for which a translation is available.


BrianNixon:
This is no good. I shouldn’t have fiddled with my wxWidgets configuration…

AenBleidd:
> Are we allowed to use C++11 yet? #3635

We don't have yet any guidelines about c++11 and later. So if that's possible - please try to avoid using it now.
Thank you for understanding.

BrianNixon:
This isn’t going to work on Mac without a wxWidgets update.

Assuming @CharlieFenton has neither the time nor the inclination to do that, do we allow platforms to diverge, or just drop this cosmetic enhancement?

AenBleidd:
@BrianNixon, What is the minimal required version of wxWidgets that supports this feature?

BrianNixon:
3.1.6
https://github.com/wxWidgets/wxWidgets/commit/deef116a09748796711d1e3509965ee208dcdf0b

talregev:
> This isn’t going to work on Mac without a wxWidgets update.
> 
> Assuming @CharlieFenton has neither the time nor the inclination to do that, do we allow platforms to diverge, or just drop this cosmetic enhancement?

There is another tool that can update quickly wxwidgets to 3.1.6 on mac.
But it not just the fact of the update the wxwidgets.
On mac wxwidgets is a modify version for boinc, also insist to support older mac. (That may have a solution for that even with heigh wxwidgets version).

BrianNixon:
Yeah – I saw that discussion, which is why I’m conscious that any update needs a robust justification…

This branch builds now, at least – but it leaves the Mac version different from the other platforms. I’ll hold off on the C++11 removal until there’s further guidance.

clientgui/BOINCGUIApp.cpp
```diff
@@ -891,23 +898,48 @@ void CBOINCGUIApp::DetectDataDirectory() {
...
-    for (iIndex = 0; iIndex <= wxLANGUAGE_USER_DEFINED; iIndex++) {
-        liLanguage = wxLocale::GetLanguageInfo(iIndex);
-        if (liLanguage) {
-            m_astrLanguages[iIndex] = liLanguage->Description;
+    // Synthesize labels to be used in the options dialog
+    m_astrLanguages.reserve(langs.size() + 1);  // +1 for the entry for "English"
+    bool has_translation_en = false;
+    for (const auto& pLI : langs) {
+        wxString Label =
+#if wxCHECK_VERSION(3,1,6)
```
talregev, line 919: This is mean 3.1.6 or higher? On all other os we already at 3.2.1.1

BrianNixon: Yes: the native language description was added in wxWidgets 3.1.6

AenBleidd: We have now wxWidgets 3.1.5 for OSX. It's highly possible that in the future release we will update it to 3.1.6 or later.
So please keep this PR in draft, so we can come back to it when time comes.

BrianNixon:
Reducing the language list to only those for which translations are available has a nasty side-effect because there is no separation in the Manager between _display language_ (really meaning only text translation) and _locale_ (which also influences things like date formats). This means regional variations are lost now that only the base language is selectable.

Example: my system locale is `en-GB`, but if I now set the BOINC Manager language to ‘English’ (because ‘English (United Kingdom)’ is no longer available), dates get displayed in U.S. format. This is unacceptable.

Needs more work…

CharlieFenton:
@BrianNixon I've taken only a cursory look at your code (or the original code). Please excuse my ignorance as I ask a few questions. 

Why do you need to eliminate the many regional variants we currently support? How do the current regional designations get converted from _de_CH_ to _German (Swiss)_? In any case, BOINC doesn't actually have a separate translation for _de_CH_ but only a generic _de_. What happens under the current code if one selects _German (Swiss)_ or _English (U.K.)_ in BOINC's menu? BOINC does support a few separate regional translations, such as _zh_CN_ and _zh_TW_.

Would it be possible to keep the current list (including the regional variants) and use the translation options in wxWidgets to translate the language name only (using some sort of lookup) in our original list so we could keep the regional variant names? I realize this would not translate the regional designations such as _Austrian_.

Perhaps it makes sense to not use the wxWidgets feature at all, and just create a list of translated language names ourselves (such as you already did in _html/inc/language_names.inc_) but including the regional variation names. This would of course require that we manually update that list if we add or remove languages we support. For robustness, it could be an N X 2 array, with the English and localized language names in each pair and using a lookup function to translate the displayed list.

This latter approach would also eliminate the need to  update wxidgets on the Mac. That said, upgrading from 3.1.5 to 3.1.6 is probably not very difficult. An upgrade to the current 3.2.2 would require considerably more testing to ensure no bugs were introduced.

BrianNixon:
Hi @CharlieFenton,

> Why do you need to eliminate the many regional variants we currently support?

We don’t _need to_; it was a side-effect of making the list of languages in the drop-down match what is actually available for the UI (as suggested in [the issue](https://github.com/BOINC/boinc/issues/2061 "#2061")).

Does it make sense to offer _German (Switzerland)_ as a language, when selecting it does not in fact translate the UI into Swiss German? Or _Irish_, when selecting it will leave the UI in English? Is it a good user experience to be presented with a list of _877_ choices?


> How do the current regional designations get converted from _de_CH_ to _German (Swiss)_?

The display strings are all compiled in to wxWidgets.


> What happens under the current code if one selects _German (Swiss)_ or _English (U.K.)_ in BOINC's menu?

The Manager _locale_ (influencing number and date formats) is set exactly as specified, using wxWidgets’ implementation (or that supplied by the operating system). The UI _display language_ (the words that appear on screen) falls back to the `de` translation supplied by BOINC when it doesn’t find one for `de_CH`, and to the compiled-in strings when it finds no translation for either `en_GB` or `en`.


> Would it be possible to keep the current list (including the regional variants) and use the translation options in wxWidgets to translate the language name only (using some sort of lookup) in our original list so we could keep the regional variant names? I realize this would not translate the regional designations such as _Austrian_.

Certainly it’s possible, but what would that add? If we keep the regional variants, why not use the native name for the full thing (which wxWidgets will do for us)? _German (Austria) – Deutsch (Österreich)_


The ‘right’ answer is probably to separate the setting of display language and locale (and wxWidgets seems to be encouraging a move from `wxLocale` to `wxUILocale` to help address problems like this), but that would seem to add unnecessary complexity for negligible benefit. Approximately nobody is ever going to change the setting from the automatic choice anyway.


Maybe the halfway house is to keep the regional variants, but only those with a base language for which a translation is available? (Though that still leaves the list pretty unwieldy: there are 108 variants of _English_ alone…)

CharlieFenton:
> Why do you need to eliminate the many regional variants we currently support?
We don’t need to; it was a side-effect of making the list of languages in the drop-down match what is actually available for the UI (as suggested in https://github.com/BOINC/boinc/issues/2061).

> Approximately nobody is ever going to change the setting from the automatic choice anyway.

Just my opinion, but I think you have found a number of good reasons to keep our current list with all the region codes, even when we don't actually support some of them directly. So I would disregard [this portion](https://github.com/BOINC/boinc/issues/2061#issuecomment-326766164) of the original issue as it applies to region codes. (If the drop down includes languages we don't port at all, then those certainly should not be listed.)

talregev:
> Does it make sense to offer _German (Switzerland)_ as a language, when selecting it does not in fact translate the UI into Swiss German? Or _Irish_, when selecting it will leave the UI in English? Is it a good user experience to be presented with a list of _877_ choices?
> 
> >

@CharlieFenton You ignore valid and the most important point. The most important is user experience as you seems to ignore. 
The very long list that cause me difficulty to change to my language and it very annoying. And to learn the most of the language there is not even translated is disappointing. I imagine for other user that switch to his language and it even not translated, after he struggle with the long list like I do. My feeling with this long list is common in the users of boinc.
This long list should be eliminated.

AenBleidd:
> The very long list that cause me difficulty to change to my language and it very annoying.

You want to say your language was not correctly detected automatically?

> This long list should be eliminated. 

I don't agree. As it was correctly pointed above, this is not a question of translation but the local settings that could be different for the same language but in different country.

And we can't make duplicate translations for every language that has different local settings since it will be too annoying for our translators as well, because this will duplicate their work, and we have no valid mechanism neither to validate nor to merge in a correct way the same language translations for different locations.

talregev:
> You want to say your language was not correctly detected automatically?

No. My windows is set to English. And I wanted to switch my local Hebrew language to tests it and see it.
I had a bad user experience.

You should find a solution without duplicate the translation and keep this list small for that only translated or partial translated.

CharlieFenton:
@talregev you wrote:

> You should find a solution without duplicate the translation and keep this list small for that only translated or partial translated.

I agree with you. As I wrote:
> If the drop down includes languages we don't port at all, then those certainly should not be listed

Apparently autocorrect changed this. What I intended to write was:
If the drop down includes languages we don't support at all, then those certainly should not be listed.
This should significantly shorten the list.

AenBleidd:
I doubt we should get rid of the language options that we have no translations for. Mainly because in this case users will lose their regional (local) settings that is not good.
Windows itself has also very long list of languages (yes, probably they have translations for all of them but still):
image
We definitely can't add translations to all possible localizations.
Best (probably) solution would be to have two separate lists:
1 - Available localizations (that will apply UI translations)
2 - All possible Regions (that will apply regional settings like 'coma separator', etc).

Potentially, we could just have current list of languages as-is just with translated labels (I don't know whether it's supported OOTB by wxWidgets) that will solve directly the reported issue., and if user choses the language we have no translation for, application will just fallback to the closes available option (that is currently does).

Everything else requires more conversation and is currently not clear to me whether it's really needed, and I doubt it should be solved in this particular PR.

@CharlieFenton, @davidpanderson, what do you think?

talregev:
What is the different between עברית‏‏‎‎ and עברית (ישראל)‏‏‎ ? (Hebrew and Hebrew (Israel))‎?
We should have only one Hebrew.

AenBleidd:
> What is the different between עברית‏‏‎‎ and עברית (ישראל)‏‏‎ ? (Hebrew and Hebrew (Israel))‎? We should have only one Hebrew.

https://stackoverflow.com/a/25171976/4210508

talregev:
> > What is the different between עברית‏‏‎‎ and עברית (ישראל)‏‏‎ ? (Hebrew and Hebrew (Israel))‎? We should have only one Hebrew.
> 
> https://stackoverflow.com/a/25171976/4210508

From stack overflow:
> There numerous languages (Hebrew, Japanese, Greek, etc) that only define one region **so in practise there isn't much difference (outside potential future proofing).**

AenBleidd:
@talregev, and this is the answer. If you want to know why wxWidgets includes both 'he' and 'he-IL' - it's better to ask them.
In our particular case, we have only 'he' translation available.
So when users selects' he-IL' they will receive the same UI language but may receive different 'regional' settings (that we do not actually control in BOINC Manager and definitely won't do)

BrianNixon:
There are multiple entries because locales exist in both a ‘base’ form (like _Portuguese_) and with regional variations (like _Portuguese (Angola)_ and _Portuguese (Brazil)_). The base form always resolves to one of the more-specific regional variations. For Hebrew, there is only one regional variation, so in this instance there is no effective difference between the two. We could probably detect cases like this and prune the list.

AenBleidd:
> We should have a map from very similar / exact translation codes to the same translation that we don't duplicate the translations.

We already have a separate ticket for that: https://github.com/BOINC/boinc/issues/5180

talregev:
> So when users selects' he-IL' they will receive the same UI language but may receive different 'regional' settings (that we do not actually control in BOINC Manager and definitely won't do)

That fine.

AenBleidd:
> We could probably detect cases like this and prune the list.

I'd not do this because it's hard to manage this and keep up-to-date

BrianNixon:
> > We could probably detect cases like this and prune the list.
> 
> I'd not do this because it's hard to manage this and keep up-to-date

Well, it’s certainly not top priority…
But it doesn’t feel like it should be too hard. For each base locale, count the number of regional variants. If there is only one, remove it. Fully automatic. _(He says, confidently…)_

davidpanderson:
Presumably other WxWidgets apps have dealt with this.  I wonder if we can get help on their forums:
https://forums.wxwidgets.org/

CharlieFenton:
> Except for Mac unless we update wxWidgets.

I suspect (hope) that upgrading from wxWidgets 3.1.5 to 3.1.6 will be trivial and won't introduce any significant bugs. If so, then I feel it is worth doing to support this PR. I will try to look into this within the next week and perhaps create a new PR for that.

BrianNixon:
Now implemented as proposed:
Screenshot of the Options dialog in English in its new state
Screenshot of the Options dialog in Hebrew in its new state


With this, the list is down to 267 entries. Pruning the cases where there is only a single regional variant of a base language would reduce it by 13. I know I claimed that isn’t difficult to implement, but I’m not convinced it’s worth it.

@talregev: Does the right-to-left layout look correct to you? (Word order, brackets, …)

Obviously adding the native description doubles the width of the list box. Is that an issue?

Observe that the list only has has the two explicit variants of Portuguese that BOINC currently ships with. This is as expected. Renaming `pt_PT` to `pt` (#5180) will allow all the other Portuguese-speaking regions to appear as well.

CharlieFenton:
I just confirmed this works on the Mac. However, the list of languages with regional codes is extremely long, at least on my Mac. It took 5 screenshots to capture them all; attached below.

Screenshot languages 1png

Screenshot langages 2

Screenshot languages 3

Screenshot langages 4

Screenshot languages 5

BrianNixon:
Interestingly the labels for Hebrew are the opposite way round in Windows and Mac. Maybe they need more direction markers.

CharlieFenton:
> I suspect (hope) that upgrading from wxWidgets 3.1.5 to 3.1.6 will be trivial

That was wishful thinking. It was considerably more effort than I had hoped, but I think I've fixed all the issues now.

BrianNixon:
> considerably more effort than I had hoped

Nothing in software is straightforward any more…

Thank you for taking the time to do it.

talregev:
The Hebrew sentences looks good. There is another problem that don't concern this PR.
the alignment of the sentences.   It happen also in English and Hebrew (left and right languages).
The name of the field is align toward the value, instead to the opposite  direction to start of the window. That all the start of sentences will align to one straight line.

talregev:
So final result is 267 option instead 877? 
Most of the 267 are translate? 
It improvement, but we can do better.

BrianNixon:
> Most of the 267 are translated?

There aren’t 267 **translations**; just 267 **variants** of the 33 languages for which there are translations of the Manager UI text.

talregev:
> > Most of the 267 are translated?
> 
> There aren’t 267 **translations**; just 267 **variants** of the 33 languages for which there are translations of the Manager UI text.

If I choose one of the 267 variants it map to one of the 33 correct languages?

BrianNixon:
> Each of the 267 variants map to one of the 33 languages correctly?

If you’re asking me whether I have tested all 267 options individually to verify that: the answer is no.

But the list is populated by comparing the language tag of each locale known to wxWidgets to the translations available to BOINC, so it shouldn’t be possible for there to be any mismatch.

talregev:
> > Each of the 267 variants map to one of the 33 languages correctly?
> 
> If you’re asking me whether I have tested all 267 options individually to verify that: the answer is no.
> 
> But the list is populated by comparing the language tag of each locale known to wxWidgets to the translations available to BOINC, so it shouldn’t be possible for there to be any mismatch.

I am not asking for you to testing all the 267 options. This task is for @AenBleidd when he release new version with the alpha testers.
If you test some of the variations (for example 2 variations of Hebrew are going to 1 translated Hebrew), and all other should work similar that ok.

I approve this PR, and @AenBleidd Please review it again.

talregev (approved):
I approve.

AenBleidd (approved):
Looks good to me

AenBleidd (approved):
Looks good to me

AenBleidd:
@BrianNixon, could you please update this PR to resolve conflicts?
Thank you in advance

PR #5179, opened 2023-04-04
client: don't change user CPID if detach from oldest project

### LINKED ISSUE #5177: Feature Request: Option to manually override external CPID
**Describe the problem**
Background: BOINC uses several forms of CPID, this is in regards to the external CPID, which is used to track statistics across projects. A user, when signing up to multiple projects, initially has several different external CPIDs which eventually "converge" onto one "dominant" CPID. This dominant CPID is the oldest CPID.

Problem: There are several things which can cause this external CPID to change:

- Adding removing projects from BOINC manager: If user removes oldest project, CPID changes. 
- Installing new copy of OS or BOINC but being unable to attach to oldest project because it no longer exists or is down

When this CPID changes, it can cause problems with other services which rely on this CPID, such as websites which track user stats across projects. A user's cross-project stats should not depend on them keeping their oldest project attached.

Additionally, sometimes it may take a while for the CPID's to converge on a new setup (such as one project being down or some other edge cases which I don't know about) leading to "split CPID" which is a difficult problem to solve.

Make it possible to force BOINC to use a specific external CPID by issuing a command via the api (boinccmd and RPC) or by setting a field in BOINC manager preferences.

davidpanderson:
- Maintain a map from email hash to the oldest user CPID and its time.
        This doesn't change if that project is detached.

 - Write/parse write this in client_state.xml

- Update it when handling a scheduler reply

- Use it when sending a scheduler request

 - If, on startup, the map is empty, populate it from PROJECT data.
        This should happen only once, when update to this version

Fixes #5177

### DIFF
--- a/client/acct_mgr.cpp
+++ b/client/acct_mgr.cpp
@@ -767,7 +767,7 @@ void ACCT_MGR_OP::handle_reply(int http_op_retval) {
                     "Attaching to %s", acct.url.c_str()
                 );
                 gstate.add_project(
-                    acct.url.c_str(), acct.authenticator.c_str(), "", true
+                    acct.url.c_str(), acct.authenticator.c_str(), "", "", true
                 );
                 pp = gstate.lookup_project(acct.url.c_str());
                 if (pp) {
--- a/client/acct_setup.cpp
+++ b/client/acct_setup.cpp
@@ -458,7 +458,7 @@ void LOOKUP_LOGIN_TOKEN_OP::handle_reply(int http_op_retval) {
         msg_printf(NULL, MSG_INFO, "Attaching to project %s", pli->name.c_str());
         gstate.add_project(
             pli->master_url.c_str(), authenticator.c_str(),
-            pli->name.c_str(), false
+            pli->name.c_str(), "", false
         );
         PROJECT *p = gstate.lookup_project(pli->master_url.c_str());
         if (p) {
--- a/client/boinc_cmd.cpp
+++ b/client/boinc_cmd.cpp
@@ -592,7 +592,7 @@ int main(int argc, char** argv) {
         safe_strcpy(url, next_arg(argc, argv, i));
         canonicalize_master_url(url, sizeof(url));
         char* auth = next_arg(argc, argv, i);
-        retval = rpc.project_attach(url, auth, "");
+        retval = rpc.project_attach(url, auth, "", "");
         show_alerts(rpc);
     } else if (!strcmp(cmd, "--file_transfer")) {
         FILE_TRANSFER ft;
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -757,10 +757,14 @@ int CLIENT_STATE::init() {
     process_autologin(true);
     acct_mgr_info.init();
     project_init.init();
+
     // if project_init.xml specifies an account, attach
     //
     if (strlen(project_init.url) && strlen(project_init.account_key)) {
-        add_project(project_init.url, project_init.account_key, project_init.name, false);
+        add_project(
+            project_init.url, project_init.account_key, project_init.name, "",
+            false
+        );
         project_init.remove();
     }
 
--- a/client/client_state.h
+++ b/client/client_state.h
@@ -347,7 +347,8 @@ struct CLIENT_STATE {
 // --------------- cs_account.cpp:
     int add_project(
         const char* master_url, const char* authenticator,
-        const char* project_name, bool attached_via_acct_mgr
+        const char* project_name, const char* email_addr,
+        bool attached_via_acct_mgr
     );
 
     int parse_account_files();
--- a/client/client_types.cpp
+++ b/client/client_types.cpp
@@ -1349,3 +1349,89 @@ double RUN_MODE::delay() {
         return 0;
     }
 }
+
+int USER_CPID::parse(XML_PARSER& xp) {
+    clear();
+    while (!xp.get_tag()) {
+        if (xp.match_tag("/user_cpid")) {
+            if (!strlen(email_hash) || !strlen(cpid) || !time) {
+                msg_printf(0, MSG_INTERNAL_ERROR, "USER_CPID::parse() failed");
+                return ERR_XML_PARSE;
+            }
+            break;
+        }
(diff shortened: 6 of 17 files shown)


AenBleidd (approved):

AenBleidd:
@KeithMyers, @makeasnek, any chance you could test this?

makeasnek:
I have a Windows 10 machine I can test it on if you get me an exe or installer :).

AenBleidd:
@makeasnek, we don't have automatically built installer, but we have binaries. Here's a built client for this PR: https://github.com/BOINC/boinc/suites/12005652776/artifacts/630473582
You should stop both BOINC Manager and BOINC client, go to the BOINC installation location (be default it's `C:\Program Files\ BOINC`), make a backup copy of `boinc.exe`, and then overwrite it from the archive (please remember you should first extract the content of the archive in some folder, because otherwise Windows might prevent you from copying file to the `Program Files` directory directly from the downloaded archive.
Please don't forget also to backup your BOINC Data directory (by default located in the hidden folder `C:\ProgramData\BOINC`) to prevent any possible data loss after you decide to switch back to the original installation version.
After that you could start BOINC Manager again (that should automatically start the client) and test the fix.

makeasnek:
Excellent thank you I will test this tomorrow

KeithMyers:
> @KeithMyers, @makeasnek, any chance you could test this?

I can build it.  But how would you set up conditions to test if the fix works?  I was thinking that you would have to already have split CPIDs in your client state file.  Then run the new client and then see if the split CPIDs were converged to the oldest CPID

Does that sound like the correct test scenario?

makeasnek:
In each of these tests, I am copying external CPID directly from client_state.xml file and reloading the file into notepad between checks

Test #1:
Add new BOINC project to a host with existing BOINC projects all converged on a single CPID
Result: New project initially has a new CPID. After clicking "update" CPID changed to existing converged CPID

Test #2
1. Remove all existing projects from BOINC
2. Create new BOINC account with new e-mail, attach to project A, write down that project's CPID
3. Create another new BOINC account with same e-mail, attach to project B, write down that project's CPID (this project (GPUGrid) never got an external CPID at all despite having fetched work for it and attempting several force updates, idk why). 
4. Create another new BOINC account with same e-mail, attach to project C, write down that project's CPID (this project got a new CPID, but converged on oldest CPID upon force update), 
5. Verify CPIDs have converged
6. Remove project A (oldest project) to verify CPID does not change on project C
7. Verification passed even after doing force update on Project C

So it looks like the intended functionality is working as removing the oldest project does not change CPID. A few things to note here:

- Even when doing a force update, the "new" CPID given to a boinc client does not automatically converge. I noticed this when a force update ("not requesting tasks don't need") didn't work. It will eventually, but forcing a request doesn't immediately solve issue. Deleting all present tasks and forcing update does seem to work, as it then needs tasks.
- This "force update" step shouldn't need to be required. I am not well versed enough in the client<->server protocol to know, but is there a way to automatically do a "force update" when a project is added, just to prevent CPID problems? Maybe without actually requesting work?
- I have attached relevant portions of event log if you want to see what it was saying while all this was happening. I tried a number of projects before I found a project that would let me create account from BOINC manager so the log is a little cluttered as a result.

[testnotes.txt](https://github.com/BOINC/boinc/files/11154451/testnotes.txt)

AenBleidd:
@davidpanderson, could you please take a look at the test results above?
Thank you in advance

davidpanderson:
I couldn't quite follow the part after "a few things to note".
But I think the issue is:
if you attach to a new project, its web site will show the correct user CPID only after the 2nd scheduler RPC.
That's because:
- when the client makes the first scheduler RPC, it doesn't know the email addr of the account,
   just the authenticator.  So it doesn't send a CPID in the request message.
  The reply message includes the email addr hash.

- The 2nd scheduler RPC request will include the correct user CPID
  (possibly from an earlier account with the same email addr).

The bottom line: CPIDs converge after 2 RPCs, which should happen pretty quickly.
It's always been this way.
This is as good as we can do without a larger code change.

makeasnek:
Can we not just issue an "update project" command/second RPC request automatically right after a project is added? That would solve the problem. Here's the scenario I'm envisioning:

- A person tells their friend about BOINC Games and how they're fun and how they should sign up and join their team to support the cause
- Their friend downloads BOINC, but they can't join BOINC games because their CPID hasn't synced, so now they have to look up documentation on how to fix their CPID and/or get told to just "wait a while" which could be hours to days depending on how often BOINC would organically make the next RPC request. <-- This experience is totally avoidable with a force update immediately after adding a project. It would make CPID syncing totally invisible to the user.

KeithMyers:
For one thing, the code change you are requesting is in a completely different area of the client code.  Do you want to wait some more for a "nice to have" change, or just get the main benefit of the originally asked for improvement merged?

I don't think that a simple manual project update is too much to ask from the new user which achieves the same thing you ask.

Or just wait for the second rpc that is going to happen regardless in a matter of minutes or at most a couple of hours.

davidpanderson:
Actually it doesn't require a 2nd scheduler RPC
It's not that big a change so I'll go ahead and do it.

makeasnek:
Awesome thank you David!

KeithMyers:
Thanks for the quick fix David.

davidpanderson:
Seems to work.  The test case is:
- attach to an old project; note the user CPID on the web site
- attach to a brand-new project using the same email addr
- after the first RPC to the new project, look at the user CPID
  on its web site; it should be the same

AenBleidd (approved):

AenBleidd:
@makeasnek, did you test the latest build? If not, please use these artifacts:
- client: https://github.com/BOINC/boinc/suites/12115307954/artifacts/638510688
- manager: https://github.com/BOINC/boinc/suites/12115307954/artifacts/638510690

makeasnek:
Tested this situation with the latest build:

1. Inside BOINC manager, sign up for Amicable numbers, get CPID A
2. Inside BOINC manager, sign up for Primegrid, get CPID B, which does not equal CPID A. I waited 15 seconds or so in case a new RPC request had to go through first.
3. After pressing update on Primegrid, CPID B now changes to CPID A

PR #5186, opened 2023-04-08
[Client] Correct limit check in throttler thread

### LINKED ISSUE #5072: 7.22.0 CPU usage FULL THROTTLE on macOS Ventura.2

BrianNixon:
Fixes #5072 more

Without this, if the CPU limit is 100%, the throttler thread suspends all tasks, resumes them all 1 s later, then goes to sleep forever (meaning subsequent changes to the limit are never applied)

(Commit 3b1ebfd1b0c3aefd45446d1ec6dacb6d7be4d4c5 changed the ‘nothing to do’ check to `limit == 0` at the same time as implementing `CLIENT_STATE::current_cpu_usage_limit()` which makes that condition impossible)

### DIFF
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -1254,7 +1254,8 @@ void* throttler(void*) {
     while (1) {
         client_mutex.lock();
         double limit = gstate.current_cpu_usage_limit();
-        if (gstate.tasks_suspended || limit > 99.99) {
+        const double CPU_USAGE_UNLIMITED = 99.99;
+        if (gstate.tasks_suspended || limit >= CPU_USAGE_UNLIMITED) {
             client_mutex.unlock();
 //            ::Sleep((int)(1000*10));  // for Win debugging
             boinc_sleep(10);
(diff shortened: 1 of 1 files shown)


AenBleidd:
@davidpanderson, could you please take a look at this PR?

davidpanderson:
Good catch!  This PR is fine and can be merged.
You can change the 99.99 to 100 since current_cpu_usage_limit() returns that if the pref is > 99.99

BrianNixon:
In general it’s good practice to avoid strict equality checks with `double`s. While a literal `100` is not a general case, a threshold comparison keeps the code slightly more resistant to future inadvertent breakage. (Imagine somebody decides to replace `100` with `limit_from_some_clever_dynamic_calculation` which comes out close to, but not exactly, 100.)

AenBleidd (approved):

PR #5189, opened 2023-04-09
[Manager] Avoid crash if account manager list is empty

BrianNixon:


davidpanderson:
Thanks.  I guess this would happen if all_projects_list.xml was missing.

talregev:
I think is my bug.

BrianNixon:
> I think this is my bug fix.

You can judge better than I can – but given the description of that bug, I doubt this patch fixes it. It only prevents a crash when selecting “Next>” on the first page of the ‘Use account manager’ wizard.

PR #5190, opened 2023-04-10
client: don't hang if detection processes hang

davidpanderson:
The client runs two "detection processes":
    - GPU detection
    - (Mac) check if in CPU emulation mode
The client was waiting indefinitely for these to exit.
If for some reason they don't exit, the client hangs.

Fix: wait no more than 10 sec for them to exit.
If still running, kill them and move on.


lib/proc_control.cpp
```diff
@@ -50,6 +52,232 @@ using std::vector;
...
+        fclose(file);
+        if (n != 2) return 0;
+    }
+    return (double)(utime + stime)/100;
+}
+#endif
+#endif
+
+// chdir into the given directory, and run a program there.
+// If nsecs is nonzero, make sure it's still running after that many seconds.
```
BrianNixon, line 140: `nsecs` got removed; this line of the comment is no longer relevant

lib/proc_control.h
```diff
@@ -23,6 +25,54 @@
...
+
+extern int run_program(
+    const char* dir,        // directory to run program in; NULL if current dir
+    const char* file,       // path of executable
+    int argc,
+    char *const argv[],     // cmdline args, UNIX-style
+    PROCESS_ID&             // ID of child process
+);
+extern int kill_process(PROCESS_ID);
+extern int get_exit_status(PROCESS_ID, int& status, double dt);
```
BrianNixon, line 61: Could make this `dt=-1.0`, to avoid the caller needing explicitly to specify the timeout when they want an infinite wait?

davidpanderson: Turns out that's never used.

client/auto_update.cpp
```diff
@@ -15,6 +15,8 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 
+DEPRECATED - we're not going to do auto-update
```
AenBleidd, line 18: Maybe it's still better to make this line a proper comment?

clientsetup/win/CACreateProjectInitFile.cpp
```diff
@@ -20,8 +20,8 @@
 
 #include "stdafx.h"
 #include "boinccas.h"
-#include "win_util.h"
-#include "base64.h"
+//#include "win_util.h"
```
AenBleidd, line 23: Why keep these commented? Maybe it's better to remove them at all to make code clearer?

AenBleidd (changes requested):

clientsetup/win/CACreateProjectInitFile.cpp
davidpanderson: done

client/auto_update.cpp
davidpanderson: This is what I do with deprecated files.
They're not supposed to compile, and this way people will see that immediately.

PR #5191, opened 2023-04-10
[Lib] Improve Windows diagnostics

BrianNixon:
Fixes all the bugs the compiler has been warning about, and several others I found while I was in there

### DIFF
--- a/lib/diagnostics.cpp
+++ b/lib/diagnostics.cpp
@@ -500,7 +500,7 @@ int diagnostics_init(
         //   that for future use.
         lReturnValue = RegOpenKeyEx(
             HKEY_LOCAL_MACHINE,
-            TEXT("SOFTWARE\\Space Sciences Laboratory, U.C. Berkeley\\BOINC Setup"),
+            _T("SOFTWARE\\Space Sciences Laboratory, U.C. Berkeley\\BOINC Setup"),
 	        0,
             KEY_READ,
             &hkSetupHive
@@ -509,9 +509,9 @@ int diagnostics_init(
             // How large does our buffer need to be?
             dwSize = sizeof(boinc_install_dir);
 
-            lReturnValue = RegQueryValueExA(
+            lReturnValue = RegQueryValueEx(
                 hkSetupHive,
-                "INSTALLDIR",
+                _T("INSTALLDIR"),
                 NULL,
                 NULL,
                 (LPBYTE)&boinc_install_dir,
--- a/lib/diagnostics.cpp
+++ b/lib/diagnostics.cpp
@@ -628,7 +628,7 @@ char* diagnostics_get_symstore() {
 // store the location of the symbol store.
 //
 int diagnostics_set_symstore(const char* project_symstore) {
-    strlcpy(symstore, project_symstore, sizeof(symstore));
+    safe_strcpy(symstore, project_symstore);
     return 0;
 }
 
--- a/lib/diagnostics.cpp
+++ b/lib/diagnostics.cpp
@@ -933,7 +933,7 @@ void boinc_trace(const char *pszFormat, ...) {
 #else
         if (flags & BOINC_DIAG_TRACETOSTDERR) {
 #ifdef _WIN32
-            n = fprintf(stderr, "[%s %s] TRACE [%d]: %s\n", szDate, szTime, GetCurrentThreadId(), szBuffer);
+            n = fprintf(stderr, "[%s %s] TRACE [%lu]: %s\n", szDate, szTime, GetCurrentThreadId(), szBuffer);
 #else
             n = fprintf(stderr, "[%s] TRACE: %s\n", szTime, szBuffer);
 #endif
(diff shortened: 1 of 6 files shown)


lib/diagnostics.cpp
```diff
@@ -496,7 +500,7 @@ int diagnostics_init(
         //   that for future use.
         lReturnValue = RegOpenKeyEx(
             HKEY_LOCAL_MACHINE,
-            _T("SOFTWARE\\Space Sciences Laboratory, U.C. Berkeley\\BOINC Setup"),
+            TEXT("SOFTWARE\\Space Sciences Laboratory, U.C. Berkeley\\BOINC Setup"),
```
AenBleidd, line 503: `_T` correctly resolves to Unicode if needed.
`TEXT` by default is just a string literal as-is.
What is the reason of this change?

lib/diagnostics.cpp
```diff
@@ -505,9 +509,9 @@ int diagnostics_init(
             // How large does our buffer need to be?
             dwSize = sizeof(boinc_install_dir);
 
-            lReturnValue = RegQueryValueEx(
+            lReturnValue = RegQueryValueExA(
```
AenBleidd, line 512: Why changing from Unicode function to ANSI?

lib/diagnostics.cpp
```diff
@@ -623,13 +627,8 @@ char* diagnostics_get_symstore() {
...
 //
-int diagnostics_set_symstore(char* project_symstore) {
-    if (!strlen(symstore)) {
-        int buffer_used = snprintf(symstore, sizeof(symstore), "%s", project_symstore);
-        if ((sizeof(symstore) == buffer_used) || (-1 == buffer_used)) {
-            symstore[sizeof(symstore)-1] = '\0';
-        }
-    }
+int diagnostics_set_symstore(const char* project_symstore) {
+    strlcpy(symstore, project_symstore, sizeof(symstore));
```
AenBleidd, line 631: Please use `safe_strcpy` instead

lib/diagnostics_win.cpp
```diff
@@ -303,9 +303,9 @@ int diagnostics_update_thread_list() {
                 pThreadEntry = diagnostics_find_thread_entry((DWORD)(uintptr_t)pThread->ClientId.UniqueThread);
 
                 if (pThreadEntry) {
-                    pThreadEntry->crash_kernel_time = (FLOAT)pThread->KernelTime.QuadPart;
-                    pThreadEntry->crash_user_time = (FLOAT)pThread->UserTime.QuadPart;
-                    pThreadEntry->crash_wait_time = (FLOAT)pThread->WaitTime;
+                    pThreadEntry->crash_kernel_time = pThread->KernelTime.QuadPart / 10000000.0;
```
AenBleidd, line 306: `1e7` is shorter, I believe

AenBleidd (changes requested):

lib/diagnostics.cpp
BrianNixon: API correctness
`TEXT` is the generic-text macro for the Windows API
`_T` is for the CRT

lib/diagnostics.cpp
BrianNixon: API correctness
`boinc_install_dir` is a `char` string
`RegQueryValueEx` is a generic-text function
The two should never be mixed

AenBleidd: `RegQueryValueEx` is a macro (since `RegQueryValueExW` was added), and where it depends on the project settings what it will be resolved to, `RegQueryValueExA` or `RegQueryValueExW`.
Currently it resolves to `RegQueryValueExA`.
I suggest to keep it as is because in case we decide to make it wide char, it would be much easier to change to `boinc_install_dir` value on build error, rather than looking for all functions with `A` suffix at the end.

BrianNixon: Changing to widechar is not going to be easy whichever way it’s done. In BOINC it could only reasonably be implemented in the context of ‘UTF-⁠8 everywhere’, in which case simply changing ANSI to Unicode and seeing what breaks will not even come close to getting it right. And the risk of that approach is that people will ‘fix’ the breakage by throwing in casts to make the compiler warnings go away while accusing Microsoft of not knowing what they’re doing… (Like [here](https://github.com/BOINC/boinc/blob/master/apps/concat.cpp#L95-L98 "apps/concat.cpp"), for instance.)

In theory, everything should _compile_ cleanly in both ANSI and Unicode. If it doesn’t, the code is using the API wrong. (Whether the application _works_ in both modes is an entirely different question…)

There are dozens of other examples of this inconsistency throughout BOINC; this instance only showed up as overlap between two concurrent patchsets which I haven’t kept in separate branches. The bigger question would be: are you interested in that kind of ‘code quality’ patch, or is it beyond the kind of thing that is worth spending time on?

AenBleidd: In this particular case I prefer to keep it as-is, and do not change macro to the function with `A` suffix.
Thank you for understanding.

Answering you bigger question, it depends: I can't answer for all cases, each should be reviewed separately.

davidpanderson:
General comments:
- we don't support 32-bit CPUs on Win anymore
- probably better to limit this PR to format codes

AenBleidd (approved):

AenBleidd:
@BrianNixon, in general I'm ok with this PR. But next time please try to make PRs as atomic as possible.
Thank you for your contribution

PR #5197, opened 2023-04-13
clean up screensaver code

davidpanderson:
Fix formatting in screensaver_x11.cpp

Remove unused RESULT* arg in terminate functions

Don't use reference params if no reason to


AenBleidd (approved):
Looks ok to me. It would be nice if @CharlieFenton reviews this as well

CharlieFenton (approved):

PR #5198, opened 2023-04-15
[Linux][vcpkg] Detect vcpkg freeglut for linux

talregev:
Detect vcpkg freeglut for linux

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1433,6 +1433,10 @@ if test "$enable_libraries" = "yes" -a "$have_glut" = yes -a "$have_jpeg" = 1; t
   configured_to_build="${configured_to_build} libraries_graphics"
 fi
 
+if test "${enable_manager}" = yes -a "$have_Xcb" = yes -a "$have_Xcbatom" = yes; then
+  configured_to_build="${configured_to_build} manager_clientscr"
+fi
+
 if test "$enable_apps" = yes -a "$enable_apps_gui" = yes; then
   configured_to_build="${configured_to_build} apps_gui"
 fi
--- a/3rdParty/vcpkg_ports/configs/manager/linux/vcpkg.json
+++ b/3rdParty/vcpkg_ports/configs/manager/linux/vcpkg.json
@@ -2,7 +2,6 @@
     "name": "boinc-manager",
     "dependencies":
     [
-        "ftgl",
         {
           "name": "curl",
           "features": ["openssl"],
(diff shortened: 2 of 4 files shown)


configure.ac
```diff
@@ -1433,6 +1433,10 @@ if test "$enable_libraries" = "yes" -a "$have_glut" = yes -a "$have_jpeg" = 1; t
   configured_to_build="${configured_to_build} libraries_graphics"
 fi
 
+if test "${enable_manager}" = yes -a "$have_Xcb" = yes -a "$have_Xcbatom" = yes; then
+  configured_to_build="${configured_to_build} manager_screen_saver"
```
AenBleidd, line 1437: This is already present in this file: https://github.com/BOINC/boinc/blob/a9866080c5f9854530df5fae7f8827e6e5fffe67/configure.ac#L1098-L1102

AenBleidd (changes requested):
If screensaver was not built before but now is building with the changes in this PR, please update https://github.com/BOINC/boinc/blob/master/deploy/prepare_deployment.py file to include new artifact.
Also, it would be nice (but not mandatory) if you check that screensaver could be built without vcpkg and if not, update our CI to include needed dependencies.

Thank you in advance.

talregev:
The screen server was build before. Nothing change about that. only msg in config process was added.

configure.ac
talregev: Yes, I check this conditions again and just add a nice msg to the user that it know that it going to build a screen server. like I did to graphic libs. the same way.

AenBleidd: We have no 'Manager screensaver'. Either name it 'Client Screensaver' (the same way the folder is named) or 'X11 Screensaver' as it named in the scripts

talregev: I name it on the condition. It can build only when you build the manager.

AenBleidd: Yeah, my bad: 'apps_gui' should be renamed as well to avoid misleading

talregev: I can change it to `manager_clientscr`

AenBleidd (changes requested):
Could you please add 'clientscr' file to the https://github.com/BOINC/boinc/blob/master/deploy/prepare_deployment.py file? Because it's obviously missed and should be there.

Thank you in advance.

talregev:
> Could you please add 'clientscr' file to the https://github.com/BOINC/boinc/blob/master/deploy/prepare_deployment.py file? Because it's obviously missed and should be there.
> 
> Thank you in advance.

No. It for another PR. This PR is for Detect vcpkg freeglut.

AenBleidd (approved):

PR #5199, opened 2023-04-16
remove Bossa and Bolt files.

davidpanderson:
These aren't part of BOINC.  I moved them to new repos.


AenBleidd (approved):
Looks good to me

AenBleidd:
@davidpanderson, just found that there are bolt/bossa files in 'html/ops' folder. Should they be removed as well?

davidpanderson:
I'll do it.

PR #5201, opened 2023-04-16
[Lib][Windows] Correct flags for `FormatMessage()`

BrianNixon:
Use `FORMAT_MESSAGE_IGNORE_INSERTS` to prevent trying to replace insert sequences in system message definitions with nonexistent arguments


AenBleidd (approved):

PR #5202, opened 2023-04-17
[Manager] Improve consistency of control labels and accelerators

BrianNixon:
Make accelerators consistent throughout:
* Exit &rarr; Ex&#x0332;it
* Close &rarr; C&#x0332;lose
* C&#x0332;ancel &rarr; Cancel (doesn’t need an accelerator; that’s what the `Esc` key is for)
* Save &rarr; S&#x0332;ave
* Help &rarr; H&#x0332;elp (except in simple view, where none of the buttons has an accelerator)

Also remove some extraneous full stops from the end of tooltips. (There are many more which could be changed, but they would need all the translations updating, which is a much bigger job.)


talregev:
Please share a screenshots.

BrianNixon:
Actually most people won’t see any change, given that UIs these days by default hide keyboard accelerators…

Before:
Screenshot of the advanced options dialog buttons, before the proposed change


After:
Screenshot of the advanced options dialog buttons, after the proposed change

talregev:
> Actually most people won’t see any change, given that UIs these days by default hide keyboard accelerators…
> 
> Before: Screenshot of the advanced options dialog buttons, before the proposed change
> 
> After: Screenshot of the advanced options dialog buttons, after the proposed change

Thank you!

AenBleidd:
While in general PR looks ok, I have to put it on hold because we have plans to do a release soon, and this particular PR invalidates a lot of translation strings. I don't want to push translators now, and don't want to delay a release another month.
Thank you for understanding.

BrianNixon:
> this particular PR invalidates a lot of translation strings

Which ones?

AenBleidd:
> Which ones? AFAICT, every ‘new’ string in this PR is already in use somewhere.

Every translatable string. Even if you change just a dot, it will invalidate the translated line completely.

AenBleidd:
Ok, I'll double check that later

AenBleidd:
@BrianNixon, could you please update your PR to resolve conflicts?
Thank you in advance.
Everything else looks fine to me.

AenBleidd (approved):

PR #5203, opened 2023-04-17
[Windows] Favor `_beginthreadex()` over `CreateThread()`

BrianNixon:
Also fixes the handle leaks from discarding the return value


AenBleidd:
And currently I see that this PR contains a lot of changes that are not connected to each other, and it could be effectively splitter into at least 3 independent PRs.
Please do PRs more atomic for easier review and later handling of them.
Thank you in advance

PR #5206, opened 2023-04-22
web: when update venue, call it a "computer" rather than a "host".

davidpanderson:
(should do this uniformly in user-facing web pages)


AenBleidd:
@davidpanderson, do we call in this case Android devices and computers as well? Technically this is not correct while still quite understandable

davidpanderson:
The choices are "device" and "computer".
A while back we chose "computer" because it's more accurate.
Smartphones are computers too.

AenBleidd (approved):

PR #5207, opened 2023-04-27
Work around uselocale() apparently broken in MacOS 13.3.1

### LINKED ISSUE #5205: Boinc manager not reporting tasks details macOS 13.3.1
I just upgraded my Intel iMac (one of the latest late 2020) from macOS 12 to latest 13.3.1.

System is working fine and I didn't experience any flaw but Boinc Manager 7.20.4 is only listing the tasks and their status, but not anymore the run time / % / remaining / deadline... I see only 0 or --- instead of numbers and %. 

All the text looks OK : project name, application name, task name, status. 

BoincTasks Js is doing the job, though.

I reinstalled boinc (7.20.4) but same behavior. 

Capture d’écran 2023-04-21 à 23 38 26

CharlieFenton:
MacOS 13.3.1 apparently broke per-thread locale uselocale() so we must use our SET_LOCALE struct on MacOS even though HAVE_USELOCALE is defined.

When BOINC Manager is run with a locale which uses comma instead of period as the decimal delimiter, this bug in MacOS causes strtod() to stop on the period decimal point. XML_PARSER::parse_double() then returns false because strtod() reports that it has processed fewer characters than expected. This results in all values obtained by parse_double() being set to zero.

Fixes #5205

### DIFF
--- a/lib/gui_rpc_client.h
+++ b/lib/gui_rpc_client.h
@@ -788,7 +788,9 @@ struct RPC {
 };
 
 
-#if defined(HAVE__CONFIGTHREADLOCALE) || defined(HAVE_USELOCALE)
+// MacOS 13.3.1 apparently broke per-thread locale uselocale() so we must
+// use our SET_LOCALE struct on MacOS even though HAVE_USELOCALE is defined.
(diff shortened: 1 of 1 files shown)


AenBleidd:
Android is expected to fail (will be fixed later) and not related to this particular PR.
All other fails look random, and will be handled by me later.

AenBleidd (approved):

PR #5229, opened 2023-05-09
Apparently ftok() is missing in the wasm C library.  Fix this the right way.

davidpanderson:
BTW, why is there no mention of a wasm build in
https://github.com/BOINC/boinc/wiki/BuildSystem


AenBleidd:
Because it's not ready yet.
Yes, we have it compiling, but in reality the client doesn't work.
It's far from being ready even for testing.

AenBleidd (approved):

PR #5238, opened 2023-05-17
Mac: add script to notarize BOINC executables

CharlieFenton:
Mac: add script to notarize BOINC Installer, Uninstaller, stand-alone client and AddRemoveUser utility.

As of MacOS 10.15 Catalina, the OS does not allow the user to run downloaded software unless it has been "notarized" by Apple. This PR adds the mac_installer/notarize_boinc.sh script, which can be run to notarize the downloadable BOINC executables after building the installer, uninstaller, etc. with the mac_installer/release_boinc.sh script.

Instructions for running both mac_installer/release_boinc.sh and mac_installer/notarize_boinc.sh can be found in the comments near the top of each script.

I have tested these changes and this PR will be ready to merge as soon as CI checks finish.

### DIFF
--- a/mac_installer/AddRemoveUser_ReadMe.rtf
+++ b/mac_installer/AddRemoveUser_ReadMe.rtf
@@ -1,0 +1,42 @@
+{\rtf1\ansi\ansicpg1252\cocoartf2709
+\cocoatextscaling0\cocoaplatform0{\fonttbl\f0\fswiss\fcharset0 Helvetica;\f1\fswiss\fcharset0 Helvetica-Oblique;\f2\fmodern\fcharset0 Courier;
+\f3\fnil\fcharset0 TrebuchetMS;\f4\fnil\fcharset0 Menlo-Regular;}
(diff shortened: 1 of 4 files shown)


AenBleidd:
@CharlieFenton, I can fix trailing spaces for you if you want

CharlieFenton:
@AenBleidd Thank you. I already did it.

Though I recently set my Xcode to delete trailing spaces, it only does that for lines I edit. These lines with trailing spaces were present in my file (locally) before I changed that Xcode setting, and they were not lines I modified today. Since this is the first time I've added this file to the repository, your earlier elimination of trailing spaces was not applied to it. 

The biggest annoyance about this source code check is that all CI checks are started again after the developer fixes the trailing spaces. Hopefully, i won't run into this again.

AenBleidd:
> The biggest annoyance about this source code check is that all CI checks are started again after the developer fixes the trailing spaces. Hopefully, i won't run into this again.

I was thinking about optimizations of our CI to skip unrelated checks (e.g. if there are only Android changes, there is no need to rerun all other CIs) but I haven't looked into this more deeply.

Thank you for the fix. PR looks good to me.

PR #5251, opened 2023-05-23
client: the default value for the "suspend_if_no_recent_input" pref should to 0, not 60

davidpanderson:
Otherwise BOINC will stop computing after 60 minutes of idleness.

Is this a show-stopper?  Quite possibly.

BTW, this should have been called "suspend_if_no_recent_input_min" (times are in seconds unless otherwise specified)

### DIFF
--- a/lib/prefs.cpp
+++ b/lib/prefs.cpp
@@ -252,7 +252,7 @@ void GLOBAL_PREFS::defaults() {
 #else
     suspend_cpu_usage = 25;
 #endif
-    suspend_if_no_recent_input = 60;
+    suspend_if_no_recent_input = 0;
     vm_max_used_frac = 0.75;
     work_buf_additional_days = 0.5;
     work_buf_min_days = 0.1;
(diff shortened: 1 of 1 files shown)


davidpanderson:
I know where it was introduced.
I'm not sure why I made that change; there's no comment or commit note,
and it wasn't caught in the code review.

RichardHaselgrove:
Exactly. That's why we need to check for 'errors like that' during user testing.

CharlieFenton:
Isn't the default also for the checkbox to be off for that option? If so, then the value for the number of seconds doesn't matter, because it won't be used.

CharlieFenton:
@davidpanderson When a user attaches to their first project, doesn't it get their web preferences and then create the global_prefs.xml? If so, then the situation you describe can't happen once they get their first tasks from the project.

davidpanderson:
When you create an account on a project, it initially doesn't have any global prefs.

davidpanderson:
I see my mistake; I initialized it to 60 in defaults() rather than enabled_defaults().

CharlieFenton:
> When you create an account on a project, it initially doesn't have any global prefs.

Doesn't it have the web prefs defaults? Also, I seem to remember that at least some of the projects take you to their web page to set up your prefs as soon as yo attach, but i could be wrong.

In any case, if we need to build 7.22.3, I won't be able to do that until after I return from my trip on June 3.

AenBleidd (approved):

PR #5260, opened 2023-06-04
web: job-submission tweaks

davidpanderson:
Add function to map sandbox logical name to physical.
Uncomment permissions functions in app manage page.
Add basic function for staging an input file.


AenBleidd (approved):
Can't test it but generally looks fine to me.

PR #5269, opened 2023-06-13
Manager, computing prefs dialog: Fix bug where the "In use means..." value mysteriously showed up in the "suspend when non-BOINC..." field.

davidpanderson:
The problem was that wxWidgets' ToDouble(&x) returns an error (which we weren't checking for)
and doesn't modify x if the string is empty.
So we were getting the value from a previous field.

Solution: set x to zero before each such call.

Fixes the bug reported in #5266

### DIFF
--- a/clientgui/DlgAdvPreferences.cpp
+++ b/clientgui/DlgAdvPreferences.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2022 University of California
+// Copyright (C) 2023 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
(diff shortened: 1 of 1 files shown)


AenBleidd:
@davidpanderson, please fix trailing whitespaces issue: 
```
Found files with errors:
./clientgui/DlgAdvPreferences.cpp:[5](https://github.com/BOINC/boinc/actions/runs/5249917126/jobs/9483276673?pr=5269#step:4:6)73
./clientgui/DlgAdvPreferences.cpp:[6](https://github.com/BOINC/boinc/actions/runs/5249917126/jobs/9483276673?pr=5269#step:4:7)45
To fix errors run locally:
```

Vulpine05:
I tested this and it looks like it fixed the problem.  Thanks!

AenBleidd (approved):

PR #5270, opened 2023-06-13
[Client] Add alert if idle time to resume computing is greater than idle time to suspend computing

### LINKED ISSUE #4939: "In Use" and "Not In Use" suspension preferences could prevent computing indefinitely

Vulpine05:
Partially fixes #4939 

**Description of the Change**
When global preferences are read in the client, a check is in place to see if the idle time to resume computing is greater than the idle time to suspend computing, but only if CPU or GPU computing is to be suspended when the computer is in use.  If this condition is true, CPU or GPU computing could be suspended indefinitely regardless of the activity of the computer.  This check will send an alert to the user to make them aware.

**Alternate Designs**
Previous ideas were discussed in #4960, but this was favored since it did not modify preferences.  Thank you to @AenBleidd and @davidpanderson for the feedback!

**Release Notes**
[Client] Alert added if idle time to resume computing exceeds idle time to suspend computing.

### DIFF
--- a/client/cs_prefs.cpp
+++ b/client/cs_prefs.cpp
@@ -675,7 +675,7 @@ void CLIENT_STATE::print_global_prefs() {
     //
     msg_printf(NULL, MSG_INFO, "-  When computer is in use");
     msg_printf(NULL, MSG_INFO,
-        "-     'In use' means mouse/keyboard input in last %.1f minutes",
+        "-     'In use' means mouse/keyboard input in last %.2f minutes",
         global_prefs.idle_time_to_run
     );
     if (!global_prefs.run_if_user_active) {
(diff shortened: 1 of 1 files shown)


Vulpine05:
Apologizes, I meant to include screen captures of what the alert looks like.  Screen captures are in 1920x1080.  I'm open to suggestions to making the alert shorter and still informative.

Under Notices:
image

In the Event Log:
image

davidpanderson:
I'd shorten it to:

Preference settings don't allow computing (9.00 > 8.00).  Please review.

The numbers tell the user which prefs we're talking about.

Vulpine05:
Great suggestion, thank you!  See revised:
image

image

I also noticed I missed in my logic that suspend if no recent input had to be enabled, too.  That is fixed in the latest commit.

Hmm... @davidpanderson, I see in the event log that the two times have different significant digits, %.1f and %f, respectively.  I used %.2f for the alert.  Do you think we should have these match?

davidpanderson:
It might be good to use %.2f for minutes, everywhere.

Vulpine05:
> It might be good to use %.2f for minutes, everywhere.

Fixed.

PR #5278, opened 2023-06-21
Rpm repo workflow

parvit:
Fixes https://github.com/TheSCInitiative/bounties/issues/1, please review and send feedback.

## Description of the Change
The change implements a new workflow that allows to generate and manage an RPM repo.

## Versions
Support is implemented currently for:
* Fedora 38, 37
* OpenSuse Leap 15.5, 15.4

More versions or distros rpm-based can be included as necessary.

## Channels
Two channels will be available: 
* alpha
* stable

Both are uploaded to the https://boinc.berkeley.edu/dl/linux url, which can however be changed in the
workflow environment BASEREPO.

To add stable and alpha channel download the repository definition file, eg.:
> https://boinc.berkeley.edu/dl/linux/stable/fc38/boinc-stable-fc38.repo

And install it in the dnf/yum/zypper folder for repositories.

## Supported operations
Two operations are implemented:
* Repository update with new package
  * Done at the same time for *all* the distro versions
  * Will fail if the same package is present with same version in specified channel

* Repository update with package remove
  * Done at the same time for *all* the distro versions
  * Will fail if the same package is not present with indicated version in specified channel
  * Will fail if the repo cannot be mirrored

The managing of the repo is done by:
1.  mirroring the published repo
2.  updating it with new package / remove indicated package
  * Both operations keep the old versions of the packages previously published (if not explicitly removed)
4.  upload of the new repo state

If for some reason any step fails, the published repo remains untouched, unless the _allow_repo_create_ workflow parameter is true, in that case the repo is recreated losing
the previous state.

## Signing key
In addition the public key under https://boinc.berkeley.edu/dl/linux/<channel>/<version>/boinc.gpg (name can be changed in workflow environment PUBKEY) must be imported for the repo commands to work correctly.

The sign keys are provided via github secret and are _always_ removed on workflow finish.

## Secrets
Three Github secrets are necessary for correct operation:
* **secrets.REPO_KEY**: GPG Public key to be used to sign the packages and repo, must be exported in armor format
* **secrets.REPO_PRIV_KEY**: GPG Private key to be used to sign the packages and repo, must be exported in armor format
* **secrets.BOINC_AUTH**: Strong authenticator key for the service account that uploads to the boinc server

## Integration
Currently the workflow only has manual triggering implemented, but can be integrated into the linux.yml flow if deemed necessary.

## Testing
For convenience, under the `rpmrepo` dir, two `Dockerfile` have been prepared to execute the steps necessary to install the repo and a package, example execution:

`docker build -t fc38-boinc --build-arg PACKAGE=boinc-linux-client --build-arg VERSION=1.0.0-1 -f ./Dockerfile.fedora .`

Please refer to the Dockerfile for additional parameters available.

## Release Notes
Workflow for RPM repository management


parvit:
To be resubmitted as some changes entered that do not belong.

PR #5279, opened 2023-06-21
Mac: Use strtod_l() only in Manager. Use strtod() in client, etc.

### LINKED ISSUE #5249: Newest version of BOINC 7.20.7 slows down processing of tasks
After upgrading BOINC to the v 7.20.7 I noticed that tasks are processed slower. In my case it is about 5% but it's not only an E@H problem. Currently from time to time I also do tasks for Denis@Home and situations there is similar. I didn't change my local settings nor account settings for these projects, however increasing settings isn't help this situation. Problem was also reported by another user on [this thread](https://einsteinathome.org/content/boinc-7207-slows-down-eh-wus-mac) of Einstein@Home project.

**My computer:**

MacBook M1 8GB RAM, SSD 256GB, Mac OS Ventura 13.4 (updated few days ago)
Laptop bought over a year ago and turned on most of the time. However, I'm a volunteer from 2009 and there's no problem on Win machines in past years (currently I don't have one to verify if the difference is present there).

**I use local settings:**

12.5% of cores (1 core) / 90%

I also do tasks for World Community Grid but unfortunately, now I cannot check it because I don't have long tasks, for example ARP, which normally lasts ~6 hours on my computer.

CharlieFenton:
Fixes #5249, #5211

PR #5211 implemented a workaround for MacOS 13.3.1 having apparently broken the per-thread locale API `useloclae()`, by using the `strtod_l()` function instead of `strtod()`. This change was needed only in the Manager (since the client doesn't use a separate RPC thread), but the changes were in _lib/parse.cpp_ and  _lib/parse.h_ which are used by the BOINC client as well as the BOINC Manager (and also other components of BOINC.)

While my tests indicate that `strtod_l()` actually takes less CPU time than `strtod()` on the M1 Mac, this PR lists the use of `strtod_l()` to BIONC Manager just in case its use in the client might contribute to the long task CPU times reported in issue #5249.

### DIFF
--- a/lib/parse.cpp
+++ b/lib/parse.cpp
@@ -747,7 +747,7 @@ bool XML_PARSER::parse_double(const char* start_tag, double& x) {
         }
     }
     errno = 0;
-#ifdef __APPLE__
+#if (defined(__APPLE__) && defined(BUILDING_MANAGER))
 // MacOS 13.3.1 apparently broke per-thread locale uselocale()
     double val = strtod_l(buf, &end, LC_C_LOCALE);
 #else
--- a/lib/parse.h
+++ b/lib/parse.h
@@ -312,7 +312,7 @@ inline bool parse_double(const char* buf, const char* tag, double& x) {
     const char* p = strstr(buf, tag);
     if (!p) return false;
     errno = 0;
-#ifdef __APPLE__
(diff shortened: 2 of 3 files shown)


CharlieFenton:
This PR can be merged as soon as CI checks pass.

AenBleidd (approved):

PR #5294, opened 2023-06-29
Make delimiters argument const so it can (legally) be a string literal.

Russ741:
Fixes #

**Description of the Change**
This no-op change to TokenizedStringToArray's signature resolves the following compiler warning:
BOINCListCtrl.cpp:251:54: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]

**Alternate Designs**
N/A

**Release Notes**
N/A


AenBleidd (approved):

AenBleidd:
Thank you for the fix!

PR #5313, opened 2023-07-21
[Manager] Add button in event log to display only alerts (errors)

### LINKED ISSUE #4258: Error filter.
**Describe the problem**

Add a filter in the event log to show only errors and to have BOINC warn you instantly that an error has occurred.

**Additional context**

Vulpine05:
Fixes #4258 

**Description of the Change**
Button added to the event log window to filter out alerts (aka errors).  Button toggles to display only alerts and to all message types.  Functionality was also added to the "show all messages" button as well.  For example, if messages are filtered by a project and then the alert button is clicked, then only alerts for that specific project are displayed.

Screen shots:
 
New button added:
New button

When alerts are filtered:
Alerts filtered

When alerts and a project are filtered:
Alert and project filtered

**Alternate Designs**
None considered.

**Release Notes**
Add button in event log to display only alerts.

### DIFF
--- a/clientgui/DlgEventLog.h
+++ b/clientgui/DlgEventLog.h
@@ -202,8 +202,8 @@ private:
 
     void                    ResetMessageFiltering();
 
-    void                    finderrormessages(bool isfiltered);
-    void                    findprojectmessages(bool isfiltered);
+    void                    FindErrorMessages(bool isFiltered);
+    void                    FindProjectMessages(bool isFiltered);
 
     bool                    EnsureLastItemVisible();
     wxInt32                 FormatProjectName( wxInt32 item, wxString& strBuffer ) const;
--- a/clientgui/DlgEventLog.h
+++ b/clientgui/DlgEventLog.h
@@ -199,6 +199,9 @@ private:
 
     void                    ResetMessageFiltering();
 
+    void                    finderrormessages(bool isfiltered);
+    void                    findprojectmessages(bool isfiltered);
+
     bool                    EnsureLastItemVisible();
     wxInt32                 FormatProjectName( wxInt32 item, wxString& strBuffer ) const;
     wxInt32                 FormatTime( wxInt32 item, wxString& strBuffer ) const;
--- a/clientgui/DlgEventLog.h
+++ b/clientgui/DlgEventLog.h
@@ -124,6 +124,9 @@ public:
     /// wxEVT_COMMAND_BUTTON_CLICKED event handler for ID_COPYSELECTED
     void OnMessagesCopySelected( wxCommandEvent& event );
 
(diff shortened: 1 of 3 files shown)


clientgui/DlgEventLog.h
```diff
@@ -198,6 +202,9 @@ class CDlgEventLog : public DlgEventLogBase
 
     void                    ResetMessageFiltering();
 
+    void                    finderrormessages(bool isfiltered);
```
AenBleidd, line 205: Please rename these functions and their parameters to the 'camel-case' name convention:
`finderrormessages` -> `FindErrorMessages`
`findprojectmessages` -> `FindProjectMessages`
`isfiltered` -> `isFiltered`
This increases code readability.

AenBleidd (changes requested):
Thank you for the PR. It looks great.

Please fix the names and also remove trailing workspaces (see the failed CI check).

davidpanderson:
Nice!
Please change text to "Show only alerts" and "Show all"
(the tooltips can be longer)

clientgui/DlgEventLog.h
Vulpine05: Noted on the whitespace (I'm not sure how I missed that, but I'll check).

As for the function names, I completely agree on the readability.  However, this statement is in the [Coding Style](https://github.com/BOINC/boinc/wiki/CodingStyle#namingl):

> Variable and function names are all lower case, with underscores to separate words.

I was just trying to follow this rule.  Let me know if I should proceed with camel-case or not.

Vulpine05:
> Nice! Please change text to "Show only alerts" and "Show all" (the tooltips can be longer)

I'm fine with "Show only alerts", but I'm concerned "show all" may imply that all filtering is being reset.  All it does is reset the filter for message priority.  This allows the user to filter by a project, and then toggle a filter for alerts if they want.

clientgui/DlgEventLog.h
AenBleidd: Please proceed with camel-case since all other functions are written this way. Thank you.

davidpanderson:
I think it would be clear - the user clicks "show only alerts"
and the button changes to "show all".
The tooltip would say something like "show both alerts and non-alert messages".

clientgui/DlgEventLog.h
Vulpine05: Fixed camel case.

Vulpine05:
@davidpanderson, fixed the button text as requested.  Additionally, I had adjust the minimum width for the window to account for the old button text; I have not reverted it since it is no longer needed.

image

image

AenBleidd (approved):

PR #5322, opened 2023-08-09
Fix Mac crash bug introduced by commit e0ffbf106d

### LINKED ISSUE #5321: CRASH ! BOINC Manager 7.24.0 on macOS
**Describe the bug**
launch BOINC manager 7.24.0 on macOS Ventura.5 on MacBook Air Late 2020.

**Steps To Reproduce**
1. overwrite install from 7.22.2 to 7.24.0
2. launch 7.24.0

**Expected behavior**
normal launch.

**Screenshots**
attached error log

**System Information**
 - OS: macOS Ventura.5
 - BOINC Version: 7.24.0

**Additional context**
nothing
[BOINC 7.24.0 error.txt.zip](https://github.com/BOINC/boinc/files/12287106/BOINC.7.24.0.error.txt.zip)

CharlieFenton:
Fixes #5321 

Commit e0ffbf106d failed t check for a NULL return from `wxLocale::GetLanguageInfo(wxLANGUAGE_DEFAULT)` which causes BOINC Manager to crash if the default system locale is unknown to wxWidgets.

This also means that the code added by #5176 which adds automatic selection of the language to the list of languages does not work on the Mac. In fact, that code broke automatic language selection on the Mac, which worked correctly in BOINC 7.22.2.

I have tried to guard the new automatic language selection code with `#ifndef __WXMAC__`, but I am getting an assert from wxWidgets "Invalid combobox index" if I scroll all the way to the top of the language selection combo box. 
@BrianNixon can you please look at this and tel me what more I need to do. Also, please note that line 752 of DlgOptions.cpp also calls `wxLocale::GetLanguageInfo(wxLANGUAGE_DEFAULT)` without checking for a NULL return.

I have also fixed 4 compiler warnings introduced by commit 59eb2c20ea.

### DIFF
--- a/clientgui/BOINCGUIApp.cpp
+++ b/clientgui/BOINCGUIApp.cpp
@@ -205,10 +205,17 @@ bool CBOINCGUIApp::OnInit() {
 #endif
     m_pConfig->Read(wxT("DisableAutoStart"), &m_iBOINCMGRDisableAutoStart, 0L);
     m_pConfig->Read(wxT("LanguageISO"), &m_strISOLanguageCode, wxT(""));
-    bool bUseDefaultLocaleDefault =
+#ifdef __WXMAC__   // wxWidgets System language detection does not work on Mac
+    m_bUseDefaultLocale = false;
+#else
+    bool bUseDefaultLocaleDefault = false;
+    const wxLanguageInfo *defaultLanguageInfo = wxLocale::GetLanguageInfo(wxLANGUAGE_DEFAULT);
+    if (defaultLanguageInfo != NULL) {
         // Migration: assume a selected language code that matches the system default means "auto select"
-        m_strISOLanguageCode == wxLocale::GetLanguageInfo(wxLANGUAGE_DEFAULT)->CanonicalName;
+        bUseDefaultLocaleDefault = m_strISOLanguageCode == defaultLanguageInfo->CanonicalName;;
+    }
     m_pConfig->Read(wxT("UseDefaultLocale"), &m_bUseDefaultLocale, bUseDefaultLocaleDefault);
+#endif
     m_pConfig->Read(wxT("GUISelection"), &m_iGUISelected, BOINC_SIMPLEGUI);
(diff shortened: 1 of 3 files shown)


CharlieFenton:
Additional explanation of the problem:
wxWidgets does not properly detect the default system locale (language and country / region codes) for any language other than English. It's not clear to me whether this is due to a bug in macOS or if wxWidgets is not doing it correctly, but I suspect the problem may be in MacOS. The language portion of the code returned by the Mac is always "en" but the region / country code is correct. So for Japanese (Japan), we get _en_JP_. For France we get _en_FR_, and so forth.  Since these are nonsense codes, wxWidgets treats it as` wxLANGUAGE_UNKNOWN` and returns NULL for the language info.

This means that all the logic he put in for Automatic Detection of the language is not useable on the Mac, and so the "Automatic Detection" item Brian Nixon added to the language menu should not appear on the Mac since it does not work. Note that automatic detection worked correctly on the Mac before #5176, though the Other Options dialog may not have always shown the language which BOINC was actually using.

Rather than totally eliminating all the changes for automatic language detection from #5176, it would be better to just remove that portion which broke automatic detection on the Mac.

CharlieFenton:
I think I've figured out how to make this work for the Mac and still keep the _(Automatic Detection)_ menu item. I'll work on this more later this week.

CharlieFenton:
@AenBleidd I believe this is fixed properly now. Please merge when checks pass.

AenBleidd (approved):

AenBleidd:
Looks good to me. Thank you for the fix!

BrianNixon:
@CharlieFenton: Thank you for fixing this. Evidently I had assumed that asking wxWidgets about the default language couldn’t fail – though I don’t recall what led me to that line of thinking, as even a cursory look at [the wxWidgets source](https://github.com/wxWidgets/wxWidgets/blob/v3.1.6/src/common/uilocale.cpp#L688) shows that it can.

PR #5323, opened 2023-08-10
Scheduler: add <min_libc_version> option to plan_class_spec.xml

davidpanderson:
(Linux) If you build an app on a machine with a given version of libc, it won't run on hosts with older versions.
Until now you had to build on a super-old system.
With this change you can build on a newer system,
and give the app version a plan class that limits it to hosts with that min libc version.


AenBleidd (approved):

AenBleidd:
Looks good to me. Thank you for the feature implementation!

PR #5332, opened 2023-08-16
Mgr: Offer only languages for which we have translations

CharlieFenton:
Before BOINC 7.24.0, the _Other Options_ dialog offered users all 897 languages and language variants possibly supported by wxWidgets. PR #5176 reduced this number by eliminating those for which we don't have a reasonable substitute translation for the requested language variant. For example, if the user selected the French dialect spoken in Chad (ISO language code `fr_TD`), BOINC substitutes generic French (ISO code `fr`); so the dialog offered `fr_TD` even though it would actually provide a generic French translation.

This is still an unwieldy number of choices and gives the false impression that BOINC actually provides all the listed dialects. This PR modifies the code to offer only the language dialects for which we provide actual translations. For backward compatibility, if the user had selected a dialect not in the new list while running an earlier version of BOINC, that dialect will continue to be listed.

In addition, this change makes it possible for me to make the improvement I unsuccessfully tried to implement in PR #5329: _Compare best translations for selected languages_.

@AenBleidd and @BrianNixon please tell me what you think.

Here is how the new trimmed language selection dialog looks now on the Mac:

Screenshot 2023-08-16 at 3 34 39 AM

### DIFF
--- a/clientgui/BOINCGUIApp.cpp
+++ b/clientgui/BOINCGUIApp.cpp
@@ -188,12 +188,19 @@ bool CBOINCGUIApp::OnInit() {
     SetAppDisplayName(wxString(displayName)); // {ass the display name to wxWidgets
 #endif
 
+    // Detect where BOINC Manager executable name.
+    DetectExecutableName();
+
+    // Detect where BOINC Manager was installed too.
+    DetectRootDirectory();
+
+    // Detect where the BOINC Data files are.
+    DetectDataDirectory();
 
     // Initialize the configuration storage module
     m_pConfig = new wxConfig(GetAppName());
     wxConfigBase::Set(m_pConfig);
     wxASSERT(m_pConfig);
-
 
     // Restore Application State
     m_pConfig->SetPath(wxT("/"));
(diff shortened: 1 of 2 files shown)


CharlieFenton:
@BrianNixon Thank you for the clear explanation. I am closing this PR. For this reason, my idea of comparing the best translations rather than the full ISO code in  PR #5329 was wrong, since it only considered the translations and not the regional locale attributes.

But this led me to do some testing and discovered that the regional attributes (date, time and numeric [decimal point & thousands separator] formats) are not getting set properly on the Mac when Automatic Detection is used, though they are correct when the Country and region is explicitly selected from the list in the _Other Options_ dialog. This was true in 7.22.2 (automatic detection was used if the _LanguageISO_ preference tag was an empty string, thus: `LanguageISO=`.) I believe this has been true for a very long time, and may never have worked.

This problem occurred because  `wxLocale::GetLanguageInfo(wxLANGUAGE_DEFAULT)`  doesn't get the correct default language code on the Mac for  the reason explained in https://stackoverflow.com/questions/48136456. On the Mac, we must use `wxUILocale::GetSystemLanguage()`. The wxWidgets documentation for wxLocale warns of this.

I will be submitting a new PR to fix this.

PR #5337, opened 2023-08-18
Mac: Fix Automatic language detection

CharlieFenton:
This PR fixes a problem on MacOS automatically detecting the language set for system to use in menus, dialogs, etc.

The regional attributes (date, time and numeric [decimal point & thousands separator] formats) were not getting set properly on the Mac when Automatic Detection was used, though they were correct when the country and region were explicitly selected from the list in the Other Options dialog. This was true in 7.22.2. (BOINC has long used automatic detection if the _LanguageISO_ preference tag was an empty string, thus: `LanguageISO=`.) I believe this issue has been present for a very long time, and this may never have worked on the Mac.

This occurred because _wxLocale_ for _wxLANGUAGE_DEFAULT_ does not work as expected on the Mac. Mac applications normally provide localization by including a directory for each supported language in the application's bundle. These directories are named xx.lproj, where xx is the 2-letter ISO code for the language; each of these directories contains the translated strings for that language. (MacOS also accepts English.lproj as an alias for en.lproj.)

On Macs, the _wxLocale_ class gets the default ISO language code from MacOS using the Cocoa API `[NSLocale currentLocale]`. As explained [here](https://stackoverflow.com/questions/48136456), this API doesn't not necessarily return the actual system language. Rather, it returns the **best match** to the system language **for which there is an xx.lproj directory in the calling application**.

But BOINC doesn't use Apple's localization scheme; we use wxWidgets' localization, bead on the UNIX _gettext_ translation scheme. Since the only localization in BOINC Manager recognized by MacOS is English, `[NSLocale currentLocale]` always returns English as the default language for BOINC Manager. This results in such odd ISO language codes as _en_FR_ (which would mean English as spoken in France.)

In fact, the [documentation](https://docs.wxwidgets.org/3.2.2/classwx_locale.html) for wxLocale warns:
> This class is known to have several problems under macOS

and rsays:
> it is recommended to use [wxUILocale](https://docs.wxwidgets.org/3.2.2/classwx_u_i_locale.html) instead of this class for the applications targeting macOS.

The solution is to detect the system language using `wxUILocale::GetSystemLanguage()` and then passing that language instead of _wxLANGUAGE_DEFAULT_ to the wxLocale APIs.

### DIFF
--- a/clientgui/BOINCGUIApp.cpp
+++ b/clientgui/BOINCGUIApp.cpp
@@ -21,6 +21,7 @@
 
 #ifdef __WXMAC__
 #include <Carbon/Carbon.h>
+#include <wx/uilocale.h>
 #include "filesys.h"
 #include "util.h"
 #include "mac_util.h"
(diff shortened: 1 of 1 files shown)


CharlieFenton:
This PR is ready to merge when CLI checks are complete.

AenBleidd (approved):

Issue #5226, opened 2023-05-05
redesign Linux idle detection


davidpanderson:
Related issues: #3405, #1187

The notion of "in use" (for computing preferences) means any interactive use of the host.
This includes:
1) a physical terminal connected via RS-232
2) remote login with ssh
3) an xterm window on the local X11 server
4) other interactive X11 applications
... am I missing anything?

Currently the client checks for input in two ways:
a) looking at the access time of tty and pty entries in /dev.
  In theory this should handle cases 1-3);
  remote logins and xterms create a pseudo-tty.
b) attempting to connect to the local X11 server,
  and doing a XScreenSaverQueryInfo() call.
  This is intended to handle case 4.

b) apparently doesn't work when the client runs as use "boinc".
And it has been argued that the client shouldn't be connecting
to ports like 6000; who knows what's there, and the client can hang.

Proposal:

- verify that method a) actually works like we think it does,
  on all major distros, no matter what user/group the client runs as.

- Don't do X11 stuff in the client.  Instead, do what we do on Windows:
  A small "monitor" program (to be run as the desktop user) creates a shared-mem segment,
  periodically queries the X server and writes the time of last user input to shared mem.
  The client checks for the existence of this segment and reads it.
  
  Note: the Manager (which runs as the desktop user) could and should perform the same function
  as the monitor program.
  We should provide the monitor program for people who don't want to use the Manager.

jamescowens:
@davidpanderson I have been discussing this with @AenBleidd and am looking at taking a stab at it.

Some initial comments:

1. I am not sure that we should necessarily include tty activity by default. Most of the reasons people want to quiesce BOINC on the machine is to improve the foreground interactivity on GUI sessions, and most of this at this point involves the GPU lagging when under GPU task loads. We may want to include an option to select in or out including tty activity for the idle detection.
2. Given the rise of Wayland and the deprecation of X, along with the replacement of the old init with systemd, it may be better to go down the route of interfacing with logind via the systemd callable API rather than creating a custom memory segment sharing solution. I don't know of any Linux distributions that have not moved to systemd at this point. This should work for both X and Wayland sessions. I believe this also works cross-user.

More to come as I review the relevant tickets more thoroughly and do a code review of the relevant code. Looks like most of it is in hostinfo_unix.cpp.

davidpanderson:
Sounds good.
BTW, hostinfo_unix.cpp is bloated.
I suggest we split it into two parts:
- static info: CPU type etc.
[...]

PR #5354, opened 2023-09-09
adding a method to patch linux idle check with required libraries.

0x0OZ:
Fixes # Linux checking being idle, A fix for #1187 

**Description of the Change**

added a method that does the check without really replacing it with the existing poor check to be reviewed and added by a C++ developer.
**Alternate Designs**

**Release Notes**

Fixing linux idle check by listening for input events in `/dev/input/*`.

If this change is not user-facing or notable enough to be included in release notes you may use the string "N/A" here. -->
To this run correctly, the user running the client has to have read and execute permissions on /dev/input/* files, this can be done using `setfacl`

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -181,10 +181,11 @@ extern "C" {
 #define LINUX_LIKE_SYSTEM 1
 #include <libevdev-1.0/libevdev/libevdev.h>
 #include <fcntl.h>
-// #include <unistd.h> || alraedy included
+// variables for getting linux IDLE time on linux.
+size_t deviceCount = 0;
+libevdev* devices[256]; // Assuming a maximum of 256 input devices
+time_t linuxUserLastSeen = time(nullptr);
 #include <glob.h>
-// #include <ctime> || already included
-
 #endif
 
 #if WASM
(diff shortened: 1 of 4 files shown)


client/hostinfo_unix.cpp
```diff
@@ -2003,6 +2010,75 @@ const vector<string> X_display_values_initialize() {
...
+        }
+
+        if (libevdev_new_from_fd(fd, &devices[i]) < 0) {
+            std::cerr << "Failed to initialize libevdev for device: " << devicePath << std::endl;
+            return;
+        }
+    }
+
+    // Main loop to monitor input events
+    while (true) {
```
AenBleidd, line 2041: This will lead to the dead lock of the BOINC Client. Instead, this functions should be called outside and return result as fast as possible.
Please be sure to test your changes before making a PR.

client/hostinfo_unix.cpp
```diff
@@ -2003,6 +2010,75 @@ const vector<string> X_display_values_initialize() {
     return display_values;
 }
 
+void getIdleTime() {
+    long idleTime = 0;
+    // Use glob to enumerate input devices in /dev/input/
+    glob_t globbuf;
+    if (glob("/dev/input/event*", GLOB_NOSORT, nullptr, &globbuf) != 0) {
+        std::cerr << "Failed to enumerate input devices. " << std::endl;
```
AenBleidd, line 2018: `msg_printf`should be used instead of the direct `std::` calls.

AenBleidd (changes requested):

client/hostinfo_unix.cpp
0x0OZ: I want to have an opinion before making any further changes, This code could be added separately, runs in the background while the client is running, and writes the time of being idle to a tmp file, which can be read inside the client at any time.
But I am not sure if you agree with this idea to be added to the code, 

Also, as I mentioned it will need permissions to read and execute /dev/input which regular users don't have by default.
So separating it could be a better approach to give the file its own permissions because having such permissions to a whole user is a security risk.

AenBleidd: Don't get me wrong, but this is a not a good approach: you did some implementation, but you haven't tested it. What is more important - we can't test it as well, because we can't just use your code and run any test, because obviously everything will just hang in an infinite loop.
If you have an idea but don't know how to implement it properly - it's better to describe your idea in detail and ask for advice.

0x0OZ: You are right.
I've removed the infinite loop and fixed the stupid logic.
But I also haven't removed any existing code.

I'll push the updates to the pull (when I figure out how).

AenBleidd: While on this branch:
git add . && git commit && git push

client/hostinfo_unix.cpp
AenBleidd: Please fix this comment as well

0x0OZ: It should be all good now.

AenBleidd:
@0x0OZ, Client is failed to build as well, you need to include 'libevdev' here to install it on CI: https://github.com/BOINC/boinc/blob/eac3e3a2d9fc74b95290273ebe2e9b0b9b133997/.github/workflows/linux.yml#L53

0x0OZ:
@AenBleidd client still fails,
I need to add `-levdev`  flag to the MakeFile, but I am not sure how it's being built

AenBleidd:
> @AenBleidd client still fails,
> 
> I need to add `-levdev`  flag to the MakeFile, but I am not sure how it's being built

You need to add this linker flags somewhere to this file: https://github.com/BOINC/boinc/blob/master/configure.ac

0x0OZ:
> > @AenBleidd client still fails,
> > I need to add `-levdev`  flag to the MakeFile, but I am not sure how it's being built
> 
> You need to add this linker flags somewhere to this file: https://github.com/BOINC/boinc/blob/master/configure.ac

Like I understand a single word 😅, I'll try to.

0x0OZ:
I made it worse with the last change to `configure.ac`...
I will try to find out how this thing work.

0x0OZ:
@AenBleidd Linux client now builds successfully, 
But other clients still fail because they try to import libedev when it _should_ be included only in Linux env.

0x0OZ:
Hey @AenBleidd,
I want to remind you of this pull,
is there anything I need to do from my side?

AenBleidd:
@0x0OZ, it would be nice to fix this build for all other OSs...

0x0OZ:
I didn't really think it would work on the first try!
@AenBleidd I think things all good now?

0x0OZ:
> @0x0OZ, it would be nice to fix this build for all other OSs...

That's done. just need a quick review.

0x0OZ:
Snap failed, I thought it did not as at first it was skipped and didn't give an error.

0x0OZ:
@AenBleidd All success but snap, and I know what's wrong but I don't know what could be done because I can't reproduce the error and it's done in a container...
It seems_ that snap doesn't go over the libs path `/usr/lib/x86_64-linux-gnu/`, because apt installs the missing lib there.
but I don't know what causes this.
I installed an Ubuntu VM and tried Ubnutu container inside that VM, Both recognized `libevdev.so.2`.

0x0OZ:
I now have more time to work on this,  Apart from the dead end for snap that I couldn't reproduce the bug and find a solution but it seems a docker thing? I have seen some issues that the libs installed during init weren't being recognized...
Also it doesn't detect ssh connection and it requires some privileges to access `/dev/input/event*` files.
because of this I think it would be better to make a stand-alone tool for an *idle timer* for every linux system.

This would be a better approach because of the permissions issue.

AenBleidd:
> because of this I think it would be better to make a stand-alone tool for an idle timer for every linux system.

And that is why we have this idea: #5226.

Since you're already have some experience with that, maybe you could work on this tool?
IT would be also nice, if before starting a coding you could do an investigation first and present your solution in the ticket I just mentioned and afterwards start doing coding?

0x0OZ:
> > because of this I think it would be better to make a stand-alone tool for an idle timer for every linux system.
> 
> And that is why we have this idea: #5226.
> 
> Since you're already have some experience with that, maybe you could work on this tool? IT would be also nice, if before starting a coding you could do an investigation first and present your solution in the ticket I just mentioned and afterwards start doing coding?

Yes, I was thinking of doing that myself, I saw this ticket which gave me some insight.
After doing enough investigation I will share what I will build there which *should* solve the problem in all mentioned cases and be easily adaptable.

AenBleidd:
@0x0OZ, nice! Thank you very much! In case you have any questions - feel free to reach out to me here or in BOINC Discord: https://discord.gg/wPRafUq

PR #5365, opened 2023-09-19
client: add app testing feature

davidpanderson:
Add a cmdline option '--app_test filename'.
Runs a job with the given executable, in a slot directory slots/app_test.
No XML files are required.

This is a new and simpler way of testing an application,
interacting with the client, without involving a project.
It's documented here: https://github.com/BOINC/boinc/wiki/AppDebug

I added this to facilitate development of the 'sporadic app' feature,
which involves complex client/app interaction.

### DIFF
--- a/client/Makefile.linux
+++ b/client/Makefile.linux
@@ -6,7 +6,8 @@
 #    make -f Makefile.linux clean all
 # 3) do the same in this dir
 
-CC = g++ -O4 -Wall -I ../ -I ../lib/
+//CC = g++ -O4 -Wall -I ../ -I ../lib/
+CC = g++ -g -Wall -I ../ -I ../lib/
 
 PROGS = boinc boinccmd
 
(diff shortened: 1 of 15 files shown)


client/Makefile.linux
```diff
@@ -6,7 +6,8 @@
 #    make -f Makefile.linux clean all
 # 3) do the same in this dir
 
-CC = g++ -O4 -Wall -I ../ -I ../lib/
+//CC = g++ -O4 -Wall -I ../ -I ../lib/
+CC = g++ -g -Wall -I ../ -I ../lib/
```
AenBleidd, line 10: I assume, this was added for debugging, and now can be removed safely.

davidpanderson:
AFAIK, Makfile.linux is used only for debugging (by me) so this may as well stay.

I used this because configure --enable-debug doesn't work (I made an issue for this).

AenBleidd (approved):

PR #5370, opened 2023-09-21
Mac: Fix build break introduced by commit 8ea2aba1d8

CharlieFenton:
Remove app_test.cpp from build of switcher

Also, add switcher and setprojectgrp targets as dependencies to client so CI builds will detect problems building them.


CharlieFenton:
@davidpanderson I don't understand why you added _app_test.cpp_ to the switcher target. But remember that you must never edit the Xcode project file directly using a text editor. It must be modified only using Xcode.

CharlieFenton:
This PR is ready to be merged when the CI checks finish.

AenBleidd (approved):

davidpanderson:
I added it using xcode, but I guess I got it wrong.
The xcode UI is confusing (to me).

PR #5373, opened 2023-09-22
Correctly detect CPU cores when cpuset cgroup is used

CallMeFoxie:
**Description of the Change**
CPU cores are not correctly detected when eg running within a container that uses cgroup cpusets. This change should improve the detection. I am not 100% sure about other platforms and may need some ifdefing, but I do not have many platforms to test it on.

**Alternate Designs**
this code was checked against `nproc` which uses the same method

**Release Notes**
Detect CPU cores when using cpuset to limit core count

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1334,6 +1334,12 @@ int HOST_INFO::get_cpu_count() {
     if(cpus_sys_path > p_ncpus){
         p_ncpus = cpus_sys_path;
     }
+#elif __GNU_LIBRARY__ /* glibc */
+    cpu_set_t set;
+
+    if (sched_getaffinity (0, sizeof (set), &set) == 0) {
+        p_ncpus = CPU_COUNT (&set);
+    }
 #elif defined(_SC_NPROCESSORS_ONLN) && !defined(__EMX__) && !defined(__APPLE__)
     // sysconf not working on OS2
     p_ncpus = sysconf(_SC_NPROCESSORS_ONLN);


AenBleidd:
@davidpanderson, could you please review this improvement?
@CharlieFenton, could you please verify that this works on MacOS without any issues?
Thank you both in advance.

CallMeFoxie:
For the record the origin is from `nproc` code here https://github.com/coreutils/gnulib/blob/724096f6e860a107d66f7a20cd3a9538a8b7a418/lib/nproc.c#L127

CharlieFenton:
@AenBleidd:     _\_\_GNU_LIBRARY\_\__ is not defined on the Mac, so the added code has no effect. The Mac has no _sched_getaffinity()_ function, but a quick Google search came up with [this](https://stackoverflow.com/questions/45227009/alternative-to-shed-getaffinity-cpu-set-t-etc) as a possible workaround.

I am not familiar with the concept of cgroup cpusets or CPU affinity, but I am puzzled by the implementation in this PR because the file _client/hostinfor_unix.cpp_ is used only in the client, and not in any BOINC libraries. So I suspect the code guarded by `#elif __GNU_LIBRARY__` might never be compiled. Also, the [Linux man page for sched_getaffinity()](https://linux.die.net/man/2/sched_getaffinity) says:
> If pid is zero, then the mask of the calling process is returned.

So this code will only return information for the client process itself. It's unclear to me whether this is what @CallMeFoxie intended.

AenBleidd:
@CharlieFenton,

I was mostly worried that this code might break some functionality on MacOS. If it will have no effect - that is perfectly fine.

> __GNU_LIBRARY__ is not defined on the Mac, so the added code has no effect. The Mac has no sched_getaffinity() function

This is completely fine. I have doubts anyone uses `cgroup` on MacOS.

> So this code will only return information for the client process itself. It's unclear to me whether this is what @CallMeFoxie intended.

I believe, the intention is to limit CPU usage of BOINC using native linux tools (for whatever reasons).
@CallMeFoxie, please correct me if I'm wrong.

CallMeFoxie:
> @AenBleidd: ___GNU_LIBRARY___ is not defined on the Mac, so the added code has no effect. The Mac has no _sched_getaffinity()_ function, but a quick Google search came up with [this](https://stackoverflow.com/questions/45227009/alternative-to-shed-getaffinity-cpu-set-t-etc) as a possible workaround.
> 
> I am not familiar with the concept of cgroup cpusets or CPU affinity, but I am puzzled by the implementation in this PR because the file _client/hostinfor_unix.cpp_ is used only in the client, and not in any BOINC libraries. So I suspect the code guarded by `#elif __GNU_LIBRARY__` might never be compiled. Also, the [Linux man page for sched_getaffinity()](https://linux.die.net/man/2/sched_getaffinity) says:
> 
> > If pid is zero, then the mask of the calling process is returned.
> 
> So this code will only return information for the client process itself. It's unclear to me whether this is what @CallMeFoxie intended.

I think it is working correctly? Because after the patch I can do
`docker update --cpuset-cpus="0-9" boinc_test`
and within the container I get 
```
root@db4289a2920b:/src/boinc# nproc
10
root@db4289a2920b:/src/boinc# boinc
22-Sep-2023 08:39:14 [---] Starting BOINC client version 7.25.0 for x86_64-pc-linux-gnu
22-Sep-2023 08:39:14 [---] This a development version of BOINC and may not function properly
...
22-Sep-2023 08:39:15 [---] Processor: 10 GenuineIntel Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz [Family 6 Model 79 Stepping 1]
```
yet the original method was returning all 56 cores as lscpu does:
```
root@db4289a2920b:/src/boinc# lscpu
...
CPU(s):                  56
```

CallMeFoxie:
Basically the whole idea is that BOINC sees the actual assigned cpu core count, not the whole machine, because if it runs in some container it may have some limits imposed (be it docker or just general cgroups limits outside of any container)

davidpanderson:
Sorry to pick nits, but please use 4-space indentation and space in 'if ('.
Otherwise fine with me.

AenBleidd (approved):

AenBleidd:
@CallMeFoxie, thank you for this fix!

Issue #5378, opened 2023-09-26
Remove two unnecessary null pointer checks for the variable “m_RPCWaitDlg”


elfring:
:eyes: The member function “`CMainDocument::RequestRPC`” contains the statement “`m_RPCWaitDlg = new AsyncRPCDlg();`”.
[Redundant checks are still applied](https://github.com/BOINC/boinc/blob/ed69f05b396fcd09007181bb72ea61927f182bb0/clientgui/AsyncRPC.cpp#L659 "Update candidate: CMainDocument::RequestRPC()") (despite of this implementation detail).
:thought_balloon: Thus I suggest to remove them.

AenBleidd:
You can go ahead and create a PR with the fix 😉

elfring:
:crystal_ball: [The software evolution can be continued accordingly](https://github.com/BOINC/boinc/pull/5379 "Delete two unnecessary null pointer checks for the variable “m_RPCWaitDlg”"), can't it?

elfring:
>…, and create the dialog only if it's actually used.

Will further source code adjustments belong to other issues? :thinking:

AenBleidd:
@elfring, this particular request belongs to the same PR: #5379.

elfring:
@davidpanderson did not add his comment as a direct contribution to my pull request for this issue.

AenBleidd:
@elfring, ok, if you want, I can say next way: request of @davidpanderson belongs to this particular issue (#5378) and the corresponding PR (#5379)

elfring:
:thought_balloon: I would prefer a clearer separation between remaining open issues for the affected software components.

AenBleidd:
@elfring, you're the author of this issue. It was a request from the maintainer to improve your PR. This improvement belongs to the same component and to the same function that is touched by your PR.
Please, take into the consideration the comment from the maintainer and improve your PR.
Thank you in advance

AenBleidd:
Closing this as 'not needed'. Please read more detailed explanation here: https://github.com/BOINC/boinc/pull/5379#issuecomment-1738925231

Issue #4048, opened 2020-10-05
BOINC Manager interface should not fully block interface when connecting with client


AustinConlon:
At the very least the close/minimize/maximize buttons at the top of the window should be clickable.

![Screen Shot 2020-10-04 at 2 36 31 PM](https://user-images.githubusercontent.com/1305957/95033188-21521d00-0672-11eb-811c-b18534b07071.png)

CharlieFenton:
Just to clarify, the "Communicating with client" alert is different from "Connecting to client." If you see that alert, the Manager is already connected to the client, but has been waiting over 1.5 seconds for the client to respond to a _Demand RPC_ request. A _Demand RPC_ is one in which the Manager cannot proceed until it receives the response from the client, and so the Manager's event processing must be suspended. 

I have edited the title of this PR accordingly.

It might be possible to implement this RPC by adding those menu items to `CBOINCGUIApp::FilterEvent(wxEvent &event)`, but I don't think so since that is a modal dialog. See [this documentation](https://docs.wxwidgets.org/3.1.0/classwx_event_filter.html).

Since the Manager's event processing must be suspended, I think it might require a major rewrite of the Manager logic to allow handling of any menu items, which is why I implemented this as a modal dialog. Ideally, this dialog should never appear, but we know from experience that is not the reality. 

I retrofitted the Manager years ago to allow for asynchronous RPCs to minimize the situations in which the manager becomes unresponsive to the user, but due to the Manager's basic design I could not find a way to allow the Manager to remain responsive while waiting for certainRPCs to complete. 

The best I could come up with for that situation is to display the dreaded "Communicating with client" dialog to let the user know why the Manager was not responding, and to allow bailing out by providing the "Exit BOINC Manager" button ("Quit BOINC Manager" on Macs) and  "Cancel" button on the dialog. The Cancel option cancels that one RPC, which sometimes resolves the problem but can produce unpredictable results. Unfortunately, the same situation quickly occurs again with another RPC after the cancel.

I called these _Demand RPCs_ or synchronous RPCs; the others I called asynchronous. You can see  these definitions in `enum ASYNC_RPC_TYPE` in _clientgui/AsyncRPC.h_. 

The logic to retrofit the Manager for asynchronous RPCs was quite complex and tricky to prevent problems from race conditions, etc. Changing the "Communicating with client" dialog to a modeless dialog to enable menu items might be possible, but would require extra code to disable all other events and would need to be done very carefully to avoid accidentally breaking things.

[...]

PR #5379, opened 2023-09-26
Delete two unnecessary null pointer checks for the variable “m_RPCWaitDlg”

elfring:
**Description of the Change**
:eyes: The member function “`CMainDocument::RequestRPC`” contains the statement “`m_RPCWaitDlg = new AsyncRPCDlg();`”.
Redundant checks were still applied (despite of this implementation detail).
:thought_balloon: Thus remove them.

**Release Notes**
Two null pointer checks for the variable “m_RPCWaitDlg” could be omitted from the implementation of the function “CMainDocument::RequestRPC”.

### DIFF
--- a/clientgui/AsyncRPC.cpp
+++ b/clientgui/AsyncRPC.cpp
@@ -656,48 +656,45 @@ int CMainDocument::RequestRPC(ASYNC_RPC_REQUEST& request, bool hasPriority) {
         // posting of more RPC requests while in this dialog, to prevent
         // undesirable recursion.
         //
-        if (m_RPCWaitDlg) {
-            response = m_RPCWaitDlg->ShowModal();
-            // Remember time the dialog was closed for use by RunPeriodicRPCs()
-            m_dtLasAsyncRPCDlgTime = wxDateTime::Now();
-            if (response != wxID_OK) {
-                // TODO: If user presses Cancel in Please Wait dialog but request
-                // has not yet been started, should we just remove it from queue?
-                // If we make that change, should we also add a separate menu item
-                // to reset the RPC connection (or does one already exist)?
+        response = m_RPCWaitDlg->ShowModal();
+        // Remember time the dialog was closed for use by RunPeriodicRPCs()
+        m_dtLasAsyncRPCDlgTime = wxDateTime::Now();
+        if (response != wxID_OK) {
+            // TODO: If user presses Cancel in Please Wait dialog but request
+            // has not yet been started, should we just remove it from queue?
+            // If we make that change, should we also add a separate menu item
+            // to reset the RPC connection (or does one already exist)?
 
-                retval = -1;
-                // If the RPC continues to get data after we return to
-                // our caller, it may try to write into a buffer or struct
-                // which the caller has already deleted.  To prevent this,
-                // we close the socket (disconnect) and kill the RPC thread.
-                // This is ugly but necessary.  We must then reconnect and
-                // start a new RPC thread.
-                if (current_rpc_request.isActive) {
-                    current_rpc_request.isActive = false;
-                    rpcClient.close();
-                    RPC_requests.clear();
-                    current_rpc_request.clear();
-                    m_bNeedRefresh = false;
-                    m_bNeedTaskBarRefresh = false;
+            retval = -1;
+            // If the RPC continues to get data after we return to
+            // our caller, it may try to write into a buffer or struct
+            // which the caller has already deleted.  To prevent this,
+            // we close the socket (disconnect) and kill the RPC thread.
+            // This is ugly but necessary.  We must then reconnect and
+            // start a new RPC thread.
+            if (current_rpc_request.isActive) {
+                current_rpc_request.isActive = false;
+                rpcClient.close();
+                RPC_requests.clear();
+                current_rpc_request.clear();
+                m_bNeedRefresh = false;
+                m_bNeedTaskBarRefresh = false;
 
-                    // We will be reconnected to the same client (if possible) by
-                    // CBOINCDialUpManager::OnPoll() and CNetworkConnection::Poll().
-                    m_pNetworkConnection->SetStateDisconnected();
-                }
-                if (response == wxID_EXIT) {
-                    pFrame = wxGetApp().GetFrame();
-                    wxCommandEvent evt(wxEVT_COMMAND_MENU_SELECTED, wxID_EXIT);
-                    s_bSkipExitConfirmation = true;
-                    pFrame->GetEventHandler()->AddPendingEvent(evt);
-                }
+                // We will be reconnected to the same client (if possible) by
+                // CBOINCDialUpManager::OnPoll() and CNetworkConnection::Poll().
+                m_pNetworkConnection->SetStateDisconnected();
             }
-            if (m_RPCWaitDlg) {
-                m_RPCWaitDlg->Destroy();
+            if (response == wxID_EXIT) {
+                pFrame = wxGetApp().GetFrame();
+                wxCommandEvent evt(wxEVT_COMMAND_MENU_SELECTED, wxID_EXIT);
+                s_bSkipExitConfirmation = true;
+                pFrame->GetEventHandler()->AddPendingEvent(evt);
             }
-            m_RPCWaitDlg = NULL;
-            m_bWaitingForRPC = false;
         }
+
+        m_RPCWaitDlg->Destroy();
+        m_RPCWaitDlg = NULL;
+        m_bWaitingForRPC = false;
     }
     return retval;
 }


elfring:
Can you accept that the avoidance of redundant pointer checks and [a questionable dialogue creation](https://github.com/BOINC/boinc/issues/5378#issuecomment-1736055641 "Comment by David Anderson from 2023.09-26") should be treated as separate issues?

AenBleidd:
@elfring, why not to fix it right here? I took a look at the code, and it won't be a big code change, but that change will affect the code that is currently changed here.
So why do we need to split this into two PRs?

elfring:
>So why do we need to split this into two PRs?

:thought_balloon: Clear separation of involved development concerns

AenBleidd:
> > So why do we need to split this into two PRs?
> 
> 💭 Clear separation of involved development concerns

This concern was raised by you. Project maintainer asked you to improve your PR to make a fix more complete and logically correct.

elfring:
@davidpanderson picked the opportunity up to indicate an additional change possibility which might be not so clear so far. :thinking:

AenBleidd:
@elfring, now you have a chance to make it clear and improve your PR

elfring:
What does hinder you to keep a strict development concern separation here?

AenBleidd:
@elfring, because the change you're requested to do will touch the same code again, thus there is no reason to have two separate PRs and instead make the real improvement, and not just the change in the code that will have no effect for the users, because there is a high chance that these particular checks are optimized by the compiler, and removed completely in the resulting code.
My personal opinion: making a requested change will take less time than the time that is already spent  just for this discussion.

CharlieFenton:
I am the author of this code, and I strongly object to this PR. Due to the asynchronous logic, race conditions can occur where `CMainDocument::OnRPCComplete(CRPCFinishedEvent&)` is called at any time causing `CMainDocument::HandleCompletedRPC()` tp destroy the dialog. In that case, if `CMainDocument::RequestRPC(ASYNC_RPC_REQUEST& request, bool hasPriority)` again calls `m_RPCWaitDlg->Destroy();` it will cause the Manager to crash.

This code is quite complex and I put in extra NULL checks as a safety measure to guard against developers inadvertently causing such situations if they modify the code in the future, as well as to guard against unintended consequences in my own code. The cost of doing NULL checks is trivial. Good programming practice dictates **always** making these type of checks throughout, even when they appear to be redundant.

CharlieFenton:
> f we're going to clean up the logic here let's finish the job, and create the dialog only if it's actually used.

@davidpanderson I can't remember now why I created the dialog before it is actually needed, but I seem to remember there was a reason I felt there would be a problem trying to  create it while waiting for a non-responsive asynchronous RPC, or that it would be better to only need to call ShowModal().

CharlieFenton:
> @davidpanderson I can't remember now why I created the dialog before it is actually needed

As I think about this, I suspect it was because of concerns about potential race conditions.

CharlieFenton:
A bit of background: _AsyncRPC.cpp,.h_ and associated code in _MainDocument.cpp,.h_ and elsewhere were a huge hack to implement asynchronous RPCs in the Manager without requiring a near total rewrite, in order to improve UI responsiveness. Therefore this is perhaps the most fragile code in all of BOINC, and should not be modified unless needed to fix a serious bug. Even the slightest changes to this code will require extremely extensive testing.

And it certainly should not be the first code touched by someone who is not yet throughly familiar with both the BOINC code and with asynchronous programming in general.

CharlieFenton:
I have been trying to follow the email threads and conversations leading up to this PR. If I understand correctly, @elfring was seeing frequent and persistent appearances of the dreaded "communicating with client" dialog on @elfring's computer and wanted to understand the reason and find a solution. It appears this was something specific to @elfring's computer, as we have not had many such reports.

If I understand correctly, @elfring determined the hung RPC was _RPC_AUTHORIZE_, (actually _rpc.authorize_). @davidpanderson is the expert on RPCs, but I think this checks the _gui_rpc_auth.cfg_ file in the BOINC Data folder to ensure the user has access to it. I believe this is done before many RPCs, though it may only be done once when first needed and the result retained; I'm not sure.

If that is where the RPC communication is hanging, it probably implies the client is pretty completely failing to communicate with the Manager, and that is probably where you should be looking.

@davidpanderson can you correct me if I am mistaken and perhaps elaborate further?

elfring:
Your feedback indicates that you are looking more for changes in other directions than the integration of a tiny source code cleanup.

:thought_balloon: Mentioned concerns affect also [the contribution “Mgr: Don't call get_notices RPC before previous one is processed to prevent multiple display of notices”](https://github.com/BOINC/boinc/commit/c5e31b7d952bdd9d60fbabb6eecf76a1bbc36b13 "Software evolution") from 2013-03-04.

elfring:
>… If I understand correctly, @elfring determined the hung RPC was RPC_AUTHORIZE, …

:thought_balloon: The desired clarification for such a technical detail should be continued with [the corresponding issue “BOINC Manager interface shouldn't fully block user interaction when in communicating with client dialog”](https://github.com/BOINC/boinc/issues/4048#issuecomment-1727647700 "Bug report from 2020-10-05").

CharlieFenton:
I looked at the code and I was mistaken about when _RPC_AUTHORIZE_ (actually _rpc.authorize_) is used. The Manager  calls it only once when first connecting to the client, to confirm that the Manager is authorized to communicate with the client. This is the first RPC the Manager sends to the client, so If the client is not responding to this it means the client has _never_ responded to the Manager. As requested, i will also post this comment to issue #4048.

AenBleidd:
I made a detailed review of the whole `RequestRPC` function that is touched by this code.
From one hand, there is no reason to check if `m_RPCWaitDlg` is NULL or not because at the moment of this check it will never be NULL. The only case when it might be NULL is when at the moment of the `new` call there was not enough memory, and thus object was't be created. However, in this case exception will be thrown, and the all the code below won't be executed at all.
From the other hand, this verification for NULL doesn't add any significant additional overhead to the function. Assuming that this `m_RPCWaitDlg` is created as an answer on the user initiated action, we're not in a rush, and thus could take some additional CPU cycles to perform this check (however, I don't believe this check will still be present in the code after the optimization by the compiler, but anyway).
Regarding the comment of @davidpanderson to create this dialog when it's really needed, I don't think this is very important in this particular case, because we (as I mentioned above) are waiting for an answer on a user initiated action, and thus while waiting for completion of the RPC, we can create a dialog beforehands, and delete it afterwards. Of course, generally it's a good practice to create objects as close to the point of their usage as possible, but in this particular case there is no need to do that: we don't have any performance related issues in this particular piece of code.

To summarise: current code is good enough, and this particular PR (and the corresponding issue) are not needed, that is why I'm closing both of them.

elfring:
>The only case when it might be NULL is when at the moment of the new call there was not enough memory, and thus object was't be created.

Such a description should be reconsidered also according to subsequent information.


>However, in this case exception will be thrown, and the all the code below won't be executed at all.

This is usual C++ software behaviour.


>To summarise: current code is good enough, …

Further implementation details are waiting on better solutions as it is indicated by some open issues.

AenBleidd:
> Further implementation details are waiting on better solutions as it is indicated by some open issues.

There is no issue in this particular function, and this PR doesn't fix any issue and provides no improvements to the current software behaviour.

elfring:
We represent different software development opinions here.

AenBleidd:
@elfring, you're completely right.
If you have in mind an improvement that will fix existing issues or provide additional useful functionality - please go ahead and create a ticket.
As you might notice, your PR (even if applied) won't fix any issue, and will not provide any improvement.
That is why if you think that this particular function has any issues - please describe this in the ticket. You might be right, and there is something we don't see, or you're wrong because you don't have so much experience with BOINC as maintainers have.
We will never find who is right and who is wrong if you will not provide any details for your statements.

In my message before, I described why I see no issue in this particular function, and described possible cases when something might go wrong, and why I don't see that this particular PR will improve anything there. Also, I see no room for improvements in this particular function.

I don't say I'm completely right, but currently you haven't provided any information and/or details that might prove that me or other BOINC maintainers are wrong.

elfring:
>…, and will not provide any improvement.

I find also a tiny source code reduction helpful.


>We will never find who is right and who is wrong …

* You indicated another agreement (in principle) because of the information “From one hand, there is no reason to check if `m_RPCWaitDlg` is NULL or not because at the moment of this check it will never be NULL.”.
* The information “Of course, generally it's a good practice to create objects as close to the point of their usage as possible, …” indicated potential for further collateral evolution.

AenBleidd:
> I find also a tiny source code reduction helpful.

As you were already told, we are quite limited in development resources, thus we are trying to fix existing issues and implement those improvements that directly impact our users in a positive way.
If something works for years without any issues - there is no need to change that.

That is why we will not rewrite software with the new language, with the newest standard, etc.

Issue #1721, opened 2016-11-25
Remove unnecessary null pointer checks


elfring:
[An extra null pointer check is not needed in functions](http://pubs.opengroup.org/onlinepubs/9699919799/functions/free.html "Do I need to check for null before free p?") like the following.
- [MFILE](https://github.com/BOINC/boinc/blob/a44f7ad49d47bdc66647e34bb678ba672fe1289f/lib/mfile.cpp#L45 "Destructor for MFILE")
- [SCHEDULER_REPLY](https://github.com/BOINC/boinc/blob/5c36ce25daae8532f44d81e9f4c5527bde5e0e51/client/scheduler_op.cpp#L566 "Destructor for SCHEDULER_REPLY")
- [STARFIELD::build_stars](https://github.com/BOINC/boinc/blob/365ca7960f751a501fc195670abd0acfc160f8d3/api/gutil.cpp#L466 "STARFIELD::build_stars function")

ChristianBeer:
Can you please create a Pull request with your proposed solution and the reasoning why you've done so? I frankly don't see a reason to remove the checks because at least for `MFILE` we allocate memory in the constructor why shouldn't we free in the destructor? The other cases are similar.

elfring:
How do you think about to [avoid redundant checks](https://stackoverflow.com/questions/18775608/free-a-null-pointer-anyway-or-check-first "Free a null pointer anyway or check first?")?

ChristianBeer:
Ok, now I understand what you mean. Although the SO users say it is not strictly required if you use a standard compliant library it also creates no performance problems. If you can guarantee that no one will ever try to build BOINC with a non standards compliant C library I'm fine with removing the checks. Since you can't do that and they don't harm us now, I don't see a reason to remove them and potentially create a problem in the future.

davidpanderson:
Let's not worry about this.  A bigger concern is making the client behave deterministically (e.g. log an error message, wait 2 seconds for the GUI to get it, then exit) on a memory alloc failure.  This could be an issue on Android.

elfring:
* How do you think about cooperation with configuration checks like [the macro call "`AC_FUNC_MALLOC`"](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Particular-Functions.html#index-AC_005fFUNC_005fMALLOC-453 "Description for AC_FUNC_MALLOC") in your build specification?
[...]

PR #5382, opened 2023-09-28
Delete unnecessary checks before calls of the function “free”

elfring:
**Description of the Change**
[The function “free”](https://stackoverflow.com/questions/18775608/free-a-null-pointer-anyway-or-check-first "Free a null pointer anyway or check first?") is documented in the way that no action shall occur for a passed null pointer.
:thought_balloon: Redundant safety checks can be avoided.

**Release Notes**
An analysis by the means of the Coccinelle software pointed the potential out to simplify some source code places.
Thus extra null pointer checks could be omitted in some function implementations.

### DIFF
--- a/api/boinc_api.cpp
+++ b/api/boinc_api.cpp
@@ -922,10 +922,8 @@ int boinc_parse_init_data_file() {
     int retval;
     char buf[256];
 
-    if (aid.project_preferences) {
-        free(aid.project_preferences);
-        aid.project_preferences = NULL;
-    }
+    free(aid.project_preferences);
+    aid.project_preferences = NULL;
     aid.clear();
     aid.checkpoint_period = DEFAULT_CHECKPOINT_PERIOD;
 
--- a/client/cs_trickle.cpp
+++ b/client/cs_trickle.cpp
@@ -317,8 +317,6 @@ void TRICKLE_UP_OP::handle_reply(int http_op_retval) {
             "[trickle] Replicated trickle done; retval %d", http_op_retval
         );
     }
-    if (req_buf) {
-        free(req_buf);
-        req_buf = 0;
-    }
+    free(req_buf);
+    req_buf = 0;
 }
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1137,7 +1137,7 @@ int get_network_usage_totals(
 
     if (sysctl(mib, 6, NULL, &currentSize, NULL, 0) != 0) return errno;
     if (!sysctlBuffer || (currentSize > sysctlBufferSize)) {
-        if (sysctlBuffer) free(sysctlBuffer);
+        free(sysctlBuffer);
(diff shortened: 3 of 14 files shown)


AenBleidd:
Unfortunately, I can't accept this PR. The reason is the same as 7 years ago: https://github.com/BOINC/boinc/issues/1721#issuecomment-265589820

elfring:
:crystal_ball: Would you still like to support programming interfaces for the function “free” which are not standard compliant?

AenBleidd:
> 🔮 Would you still like to support programming interfaces for the function “free” which are not standard compliant?

If everything works currently without issues - then why not?

How this PR improves anything?
Does it make code faster - I doubt.
Does it improves user experience - definitely no.
Does it improves code readability - also no, because the purpose of these checks is quite obvious.

So why to spend time doing PRs that will fix nothing and improve nothing instead of fixing real existing issues?
We have 506 open tickets. Are you sure there is no any single ticket that you could fix and really improve the BOINC?

elfring:
>If everything works currently without issues - then why not?

I assume that the desire for benefits from standard compliant programming interfaces can grow.


>How this PR improves anything?

I tend to interpret the presented change possibilities in this way.


>Does it make code faster? - I doubt.

A bit less source code can influence run time characteristics also in desirable directions.


>Does it improves user experience? - definitely no.

This was not directly the intention for the shown transformation.


>Does it improves code readability? - also no, because the purpose of these checks is quite obvious.

The C++ programming language supports improved coding styles, doesn't it?

AenBleidd:
> I assume that the desire for benefits from standard compliant programming interfaces can grow.

We support building BOINC for different platforms on different OSs. We can't verify that every library on all these OSs are complaint.

> A bit less source code can influence run time characteristics also in desirable directions.

I highly doubt there will be any significant performance improvements.

> The C++ programming language supports improved coding styles, doesn't it?

True, but it also doesn't force you to not do any additional checks if you want to be sure your code is safe and stable.

The solution you propose is fine, but it doesn't fix anything and improve nothing. At the same time your change might break some builds we're not aware about. And we have no possibility to test upfront on every platform we currently support.

If you want to improve the code and fix code issues that might have impact - I can grant you access to our Coverity project.

Otherwise before opening any PR, please open the corresponding ticket to start discussion, and then you will save your time by not implementing something that will be rejecting, and our time by eliminating such a discussions.

Thank you in advance.

elfring:
>We can't verify that every library on all these OSs are complaint.

:thought_balloon: I imagine that software build systems can be designed in the way that such development concerns would be handled better.


>I highly doubt there will be any significant performance improvements.

:crystal_ball: Will interests grow so that corresponding measurements will become more attractive?

AenBleidd:
> Will interests grow so that corresponding measurements will become more attractive?

If you want to do such measurements - sure, go ahead.
I specifically checked every change in this PR: there is no a single change here that affects a function that is called very often to make any performance impact.

PR #5385, opened 2023-10-03
Client: add sporadic app feature

### LINKED ISSUE #5358: support sporadic apps
There are proposals for apps that need to compute sporadically; I'll call them 'sporadic apps'.
Typically, the computer running BOINC is acting as a worker node in a distributed computing system
outside of BOINC.
Such systems often involve some form of parallel computing:
lots of worker nodes must run at the same time,
typically exchanging data with other worker nodes using P2P communication or via a relay.
The app communicates with a server of this system to get work requests.
When a work request is active, BOINC should let the app use all the computing resources it needs,
preempting other jobs if necessary.

Implementing this requires some additions to BOINC's app API,
changes to the client, changes to vboxwrapper,
and changes to the logic of the app.

Proposed design:

client->app messages:
ok_to_compute (true or false)
    false if:
    - CPU (or GPU if relevant) computing is suspended
    - other projects have priority now (e.g. deadline crunch)
    This is sent every N sec for reliability

app->client messages:
computing_state
    - not computing and don't want to
    - not computing but want to
    - currently computing
    sent every N sec for reliability

client: new app type 'sporadic'
(along with regular and non-CPU-intensive).
Sporadic apps always run.
Various changes to scheduling logic (details TBD)

vboxwrapper: forward messages to/from the VM

app:
loop
   if get not OK to compute from BOINC client
      if computing, stop
      send "can't compute" message to server
  if get OK to compute from BOINC client
      send "can compute" to server
  if get request from server
     state = currently computing
     start computing

davidpanderson:
See https://github.com/BOINC/boinc/wiki/Sporadic-Applications
Fixes #5358

This should have no effect if no sporadic apps are present.

Passes basic tests using a simple sporadic app.
Additional testing/debugging will be needed when/if there are real sporadic apps.

### DIFF
--- a/android/buildAndroidBOINC-CI.sh
+++ b/android/buildAndroidBOINC-CI.sh
@@ -260,7 +260,7 @@ vcpkgDir()
     echo $vcpkg_dir
 }
 
-list_apps_name="boinc_gahp uc2 ucn multi_thread sleeper worker wrapper wrappture_example fermi"
+list_apps_name="boinc_gahp uc2 ucn multi_thread sleeper sporadic worker wrapper wrappture_example fermi"
 
 NeonTest()
 {
@@ -327,6 +327,7 @@ RenameAllApps()
                 ../samples/example_app/ ucn
                 ../samples/multi_thread/ multi_thread
                 ../samples/sleeper/ sleeper
+                ../samples/sporadic/ sporadic
                 ../samples/worker/ worker
                 ../samples/wrapper/ wrapper
                 ../samples/wrappture/ wrappture_example
--- a/api/boinc_api.cpp
+++ b/api/boinc_api.cpp
@@ -473,7 +473,7 @@ static void handle_heartbeat_msg() {
     if (parse_bool(buf, "suspend_network", btemp)) {
         boinc_status.network_suspended = btemp;
     }
-    if (parse_int(buf, "sporadic_ca", i)) {
+    if (parse_int(buf, "<sporadic_ca>", i)) {
         boinc_status.ca_state = (SPORADIC_CA_STATE)i;
     }
 }
--- a/api/boinc_api.cpp
+++ b/api/boinc_api.cpp
@@ -194,6 +194,7 @@ char remote_desktop_addr[256];
 bool send_remote_desktop_addr = false;
 int app_min_checkpoint_period = 0;
     // min checkpoint period requested by app
+SPORADIC_AC_STATE ac_state;
 
 #define TIMER_PERIOD 0.1
     // Sleep interval for timer thread;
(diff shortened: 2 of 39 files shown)


AenBleidd:
@davidpanderson, currently the build is broken for Windows because of the new file added.
Let's handle this offline (will write an email to you with the explanations what should be fixed).
I'll put this PR to draft right now to prevent it from being merged and to not block the current master branch.

CharlieFenton (approved):
This is far outside my area of expertise so I don't feel I can review it other than to say that it builds without warnings or errors on my Mac.

davidpanderson:
The new code for sporadic apps never gets called
unless there are sporadic apps; this should be easy to check.
So it shouldn't break anything.

We can't fully debug it until we have a real sporadic app;
the distributed AI people will hopefully have one soon.

-- D

CharlieFenton:
> Add sporadic sample app to the linux and android CI pipelines

My understanding is that this is not actually a sample / example but a test application. It is unclear how it is to be used. Perhaps @davidpanderson can clarify this.

AenBleidd:
@CharlieFenton, I got a different impression. But I might be wrong here, of course

AenBleidd:
@CharlieFenton, we have most of our samples built on the CI for MacOS.
Do we need to build sporadic sample all as well?

CharlieFenton:
The first line of samples/sporadic/sporadic.cpp is
`// test sporadic app`

AenBleidd (approved):

davidpanderson:
The apps in samples/ have various purposes.
Some are for direct use by projects (wrapper, vboxwrapper).
Others are templates for projects (nvcuda).
Others are for testing the API (sleeper, worker, sporadic)

CharlieFenton:
Wouldn't it be more appropriate for the test apps to be in the tests directory rather than samples?

AenBleidd:
'Tests' directory usually contains unit/integration/module tests.
And not really a test applications.
I'd personally keep them as-is in the samples folder.

CharlieFenton:
> @CharlieFenton, we have most of our samples built on the CI for MacOS.
Do we need to build sporadic sample all as well?

On the Mac we build the example_app (uppercase, uppercase graphics & slideshow), VBoxWrapper, wrapper, OpenCLApp, and the zip apps, all  of which are samples or for direct use by projects. SInce it is a test app, I don't think this in the same category. 

I don't understand how samples/sporadic is to be used, if it is needed on more than one platform or even if it is needed on a regular basis after the sporadic apps functionality has been debugged.

I really don't want to spend time getting it to build on the Mac unless there is a compelling reason to do so.

davidpanderson:
We won't need it on the Mac.  Linux and Win is enough to debug API stuff.

AenBleidd:
>  if it is needed on a regular basis after the sporadic apps functionality has been debugged.

I believe, it's always good to have the code available somewhere in case you need later require to test some additional functionality or to debug existing one because of some issue or corner case found.

Then I'll proceed with this PR and merge it since it's done.

AenBleidd:
@davidpanderson, @CharlieFenton, thank you for working on this important topic!

PR #5123, opened 2023-03-02
client and server: increase p_features size from 1024 to 8192.

davidpanderson:
1024 wasn't enough for some CPUs.
This caused some hosts to not get sse3 app versions, for example.

Note: ideally we'd change this to std::string.
But that would have disrupted the code that populates it (hostinfo_unix.cpp, hostinfo_win.cpp).
Easier/safer to do it this way.

Fixes #5122

PR #5397, opened 2023-10-16
complete #5123: client and server: increase p_features size from 1024 to 8192

bema-aei:
Fixes #5123

In the current version of the code (master branch), p_features size of the host record is still fixed 1024. That doesn't look right (and probably won't help):

char p_features[1024];

https://github.com/BOINC/boinc/blob/master/db/boinc_db_types.h#L369

### DIFF
--- a/db/boinc_db_types.h
+++ b/db/boinc_db_types.h
@@ -366,7 +366,7 @@ struct HOST {
     // but not stored in the DB
     // TODO: move this stuff to a derived class HOST_SCHED
     //
-    char p_features[1024];
+    char p_features[P_FEATURES_SIZE];
     char virtualbox_version[256];
     bool p_vm_extensions_disabled;
     int num_opencl_cpu_platforms;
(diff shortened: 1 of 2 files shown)


AenBleidd (approved):

AenBleidd:
@bema-aei, could you please fix the size of the `buf` to use the same define instead of the number?https://github.com/BOINC/boinc/blob/74d2d456a4a3ea54b7d457e4c5b50d48c8d5a72a/sched/plan_class_spec.cpp#L272

AenBleidd (approved):

bema-aei:
Regarding the failes checks: "Please upload with the Codecov repository upload token to resolve issue." Nothing I can do about that.

AenBleidd:
> Regarding the failes checks: "Please upload with the Codecov repository upload token to resolve issue." Nothing I can do about that.

You can ignore them, this happens from time to time, and this is nothing related to the changes in this PR.

AenBleidd:
Thank you for the fix!

PR #5399, opened 2023-10-17
Update weak_auth.php

Caboose700:
**Description of the Change**

Add text to explain that changing the email address connected to the account will also change the weak account key.

**Alternate Designs**

N/A

**Release Notes**

N/A


AenBleidd:
@davidpanderson, could you please verify that this is correct?

davidpanderson:
Yup, it's correct.  Thanks for pointing this out.

I made another PR, #5400, which includes this, and also
adds text to the forms for changing email addr and password
saying that doing so will change the weak account key.

PR #5411, opened 2023-10-28
Mac screensaver: update for MacOS 14.0 Sonoma

CharlieFenton:
MacOS 14.0 Sonoma broke many third-party screensavers, including BOINC's. This PR addresses several issues:

[1] The most serious was that once the BOINC screensaver runs, the desktop remained black, leaving the user unable to access the desktop. The user had to log out using the keyboard shortcut to regain access. This is fixed.

[2] The BOINC screensaver can no longer run the default graphics _boincscr_ or any project graphics, though project graphics still work when launched from the BOINC Manager using the _Show graphics_ button.

To get around Apple's earlier sandbox restrictions on screensavers, we use IOSurfaceBuffers to allow the screensaver to display the project application's or _boincscr's_ graphics. This involves the use of Mach communication. But Apple has added a new sandbox restriction on screensavers preventing the use of Mach communication.

I have been working with Apple Developer Technical Support trying to find a solution. At their recommendation, I have filed a bug report with Apple asking them to remove this new restriction, and we are waiting for their decision. 

If Apple won't remove the new restriction, I have a possible new approach, but it involves significant changes in the code and would require all projects to at least relink their graphics apps with our updated libraries.

For now, the updated screensaver in this PR just shows a message "BOINC can't show screensaver graphics on this version of MacOS"

[3] Our installer has asked the user if they would like to set BOINC as their screensaver. If they answered yes, it would do that automatically. This ability is broken under MacOS 14.0. This is recognized as a bug, and I have filed a separate bug report with Apple for this issue, also at the recommendation of Apple Developer Technical Support. For now, the Installer does not offer that option when run on MacOS 14.

### DIFF
--- a/clientscr/Mac_Saver_Module.h
+++ b/clientscr/Mac_Saver_Module.h
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2020 University of California
+// Copyright (C) 2023 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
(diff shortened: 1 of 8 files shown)


CharlieFenton:
This PR is ready to be merged when CI checks have passed

AenBleidd (approved):

PR #5433, opened 2023-11-22
5428: Replace distutils with setuptools

Delizald:
Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

### DIFF
--- a/.github/workflows/linux-package.yml
+++ b/.github/workflows/linux-package.yml
@@ -67,7 +67,7 @@ jobs:
       - name: Install dependencies
         run: |
           apt-get -qq update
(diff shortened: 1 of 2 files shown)


Delizald:
Looks like there's more distutil references 
image

AenBleidd:
Could you please replace those as well?

davidpanderson:
The important pieces of Python code are tools/make_project and sched/start.

We need to decide whether to use exclusively Python3.
I'd be fine with this, although it might inconvenience projects running on old servers.

BTW, (although I ike Python) the use of Python in BOINC is old and messy;
for example, the path_config.py kludge (see py/Boinc).
At one point I converted 'update_versions' from Python to PHP
and it was much shorter and simpler.
We could do the same for make_project and start.

PR #5451, opened 2023-12-09
Add check for stuck jobs in poll()

### LINKED ISSUE #5352: stuck jobs
Reportedly, some VM jobs (and possibly others) get in a "stuck" state where they
don't make progress: no fraction done change, and little CPU usage.
These jobs will eventually be aborted when their elapsed time reaches the rsc_fpops_bound limit,
but this could take weeks or months depending on the limit.

Proposal: have the client try to figure out when a job is stuck:
e.g. in the last hour of running, the fraction done hasn't changed,
and the CPU time is < 10s.
At that point, the client could
1) notify the user, suggesting that they abort the job
2) abort the job

We could do 1) for starters, to make sure that the logic is right,
then at some point do 2).

FTang21:
Fixes #5352

**Description of the Change**
Add a way to check if jobs were stuck by adding an addition poll that occurs every hour, it checks if an active_tasks fraction_done doesn't change and current_cpu_time < 10s. 

**Alternate Designs**

**Release Notes**

### DIFF
--- a/client/app_control.cpp
+++ b/client/app_control.cpp
@@ -149,6 +149,25 @@ bool ACTIVE_TASK_SET::poll() {
             }
         }
     }
+
+    // check if a job is "stuck" (did not make progress in the last hour)
+    // notify the user about the issue
+    // abort after some time
+    static double last_fraction_done_check_time = 0;
+    if (gstate.now - last_fraction_done_check_time > FRACTION_DONE_POLL_PERIOD) {
+        for (i=0; i<active_tasks.size(); i++){
+            ACTIVE_TASK* atp = active_tasks[i];
+            // do not consider the first pass
+            if (last_fraction_done_check_time == 0  && atp->stuck_fraction_done == atp->fraction_done 
+                    && atp->current_cpu_time < 10) {
+                msg_printf(atp->result->project, MSG_INFO, "Task has not made progress in last hour, consider aborting");
+                // atp->abort_task(EXIT_TASK_STUCK, "Task has not made progress in last hour, begin to abort");
+            }
+            atp->stuck_fraction_done = atp->fraction_done;
+        }
+        last_fraction_done_check_time = gstate.now;
+    }
+
     if (action) {
         gstate.set_client_state_dirty("ACTIVE_TASK_SET::poll");
     }
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -114,6 +114,7 @@ ACTIVE_TASK::ACTIVE_TASK() {
     fraction_done_elapsed_time = 0;
     first_fraction_done = 0;
     first_fraction_done_elapsed_time = 0;
+    stuck_fraction_done = 0;
     scheduler_state = CPU_SCHED_UNINITIALIZED;
     next_scheduler_state = CPU_SCHED_UNINITIALIZED;
     signal = 0;
--- a/client/app.h
+++ b/client/app.h
@@ -112,6 +112,8 @@ struct ACTIVE_TASK {
         // first frac done reported during this run of task
     double first_fraction_done_elapsed_time;
         // elapsed time when the above was reported
+    double stuck_fraction_done;
+        // keeps track if fraction has change to ensure it is not stuck
     SCHEDULER_STATE scheduler_state;
     SCHEDULER_STATE next_scheduler_state; // temp
     int signal;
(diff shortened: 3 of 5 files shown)


AenBleidd:
This might be an issue for really big tasks (e.g. ClimatePrediction.net)

FTang21:
Would this be more of an issue of an hour is too short of time frame, or more with the implementation?

AenBleidd:
@FTang21, ah, no, my bad: in the original proposal there was an additional verification method for long running jobs: CPU time, and I completely missed that this was the part of the implementation in this PR.

@davidpanderson, could you please review this and verify that this is a desired implementation of the original proposal?

davidpanderson:
There are various problems with this.
I updated the issue to clarify what needs to be done:
https://github.com/BOINC/boinc/issues/5352

FTang21:
@davidpanderson I updated the implementation based on the updated issue. Lmk if this is ok. Should I add the abort on its own after some time or this is fine for now? Would this also be preferable as it function?

davidpanderson:
For now, let's just show a message using
msg_printf(atp->project, MSG_USER_ALERT...)

... telling the user which job is stuck,
and that they should consider aborting it.

This will be useful for testing because we can see the stuck job
and decide if it's really stuck.

FTang21:
Gotcha, I updated it to MSG_USER_ALERT

client/app_control.cpp
```diff
@@ -149,6 +149,36 @@ bool ACTIVE_TASK_SET::poll() {
...
     }
+
+    // check if a job is "stuck" (did not make progress in the last hour)
+    // notify the user about the issue
+    // abort after some time
+    static double last_stuck_check_time = 0;
+    if (gstate.now - last_stuck_check_time > STUCK_CHECK_POLL_PERIOD) {
+        for (i=0; i<active_tasks.size(); i++){
+            ACTIVE_TASK* atp = active_tasks[i];
+            if (!atp->non_cpu_intensive()) continue;
```
AenBleidd, line 160: Correct me if I'm wrong but this check will skip all the normal tasks, and will iterate through Non-CPU intensive tasks only.
Is this a typo and `!` is not needed here?

FTang21: Ah i believe that is correct, that `!` shouldn't be there

davidpanderson:
Almost but not quite.  Please review my pseudo-code.

FTang21:
Ah I see what I missed, it should match the order provided.

client/app_control.cpp
```diff
@@ -149,6 +149,37 @@ bool ACTIVE_TASK_SET::poll() {
...
+                atp->stuck_check_elapsed_time = atp->elapsed_time;
+                atp->stuck_check_fraction_done = atp->fraction_done;
+                atp->stuck_check_cpu_time = atp->current_cpu_time;
+                continue;
+            }
+            if (atp->elapsed_time < atp->stuck_check_elapsed_time + STUCK_CHECK_POLL_PERIOD) continue;
+            if (atp->stuck_check_fraction_done == atp->fraction_done &&
+                    (atp->current_cpu_time - atp->stuck_check_cpu_time) < 10) {
+                // if fraction done does not change and cpu time is <10, message the user
+                msg_printf(atp->result->project, MSG_USER_ALERT,
```
AenBleidd, line 174: Let's imagine, user has several tasks of the same project, and only one of them stuck.
How the user should know which one to abort?
I believe in the message the WU name should be shown

davidpanderson: Better to show the result name;
that's what we show in the Manager,
and in theory there could be two jobs with the same workunit.

AenBleidd: Agree, result name would be even better 👍

FTang21: Just added the result name into the msg_printf

client/app_control.cpp
```diff
@@ -149,6 +149,39 @@ bool ACTIVE_TASK_SET::poll() {
...
+                    "[task] has not made progress in last hour, consider aborting task %s",
```
AenBleidd, line 175: Maybe the better working should be
```
Task has not made any progress within the last hour, consider aborting it: %s
```
@davidpanderson, could you please verify the spelling?

RichardHaselgrove: I'm worried about the direction this might take the BOINC client (eventually), or might lead the user.

Not every wrapper-using project successfully manages to implement a realistic measure of progress made. This direction of thought changes the progress report from 'nice to have' towards 'essential'. It needs to be explained carefully.

davidpanderson: A job is regarded as stuck if, in the last hour of running,
a) it accumulates less than 10 sec of CPU time
AND
b) its fraction done is unchanged.

b) by itself doesn't make it stuck.

PR #5482, opened 2024-01-08
client: handle changes to project master URLs

davidpanderson:
The Einstein@home web site says its master URL is
https://einsteinathome.org.
But its scheduler replies say the master URL is
https://einstein.phys.uwm.edu
If you attach to URL1 the client creates a file
account_einsteinathome.org.xml
Then it does a scheduler RPC and gets a reply saying the master URL is URL2. It then creates a file
account_einstein.phys.uwm.edu.xml

If you then detach from the project, it deletes the 2nd account file but the first one is still there,
so the next time you run the client you're still attached to Einstein@home!

To fix this situation:

1) on startup, check for misnamed account files
(file name doesn't match master URL in the file).
Delete (if there's already a file of the right name) or rename.

2) on scheduler RPC reply with a URL that differs by more than http(s)://, rename the account file and the project directory. Reset the project (remove jobs and other info)
since soft links point to the old project dir.

### DIFF
--- a/client/cs_account.cpp
+++ b/client/cs_account.cpp
@@ -323,6 +323,7 @@ int CLIENT_STATE::parse_account_files() {
     PROJECT* project;
     FILE* f;
     int retval;
+    char path[MAXPATHLEN];
 
     DirScanner dir(".");
     while (dir.scan(name)) {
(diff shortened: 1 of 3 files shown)


AenBleidd:
In general I agree with the code, but in case project changes url for some reason, and there are some running/pending tasks, they will be cancelled. Is this the desired behavior? Maybe it make sense to not accept new tasks, and reser the project the next time when there are no tasks running and finished but not reported yet?

davidpanderson:
The client needs to rename the project directory to the current URL.
That means that the soft links in the slot directories (to ../../projects/proj_url)
aren't valid anymore, so we can't resume the jobs.

But it doesn't matter - this happens only once.

AenBleidd:
> But it doesn't matter - this happens only once.

Sometimes project could be misconfigurated, and I saw several times when you connected to the project with the URL XXX, BOINC client says you need to join the project using the URL YYY, but when you connect to the project using URL YYY, BOINC client starts to say that you need to join the project using the XXX URL.
I understand that is an issue on the project side mostly, but it might affect users in a non-good way.
Again, since this happens very rare, I assume this fix is fine.

AenBleidd (approved):

PR #5490, opened 2024-01-17
changes to support job-processing cookbook

davidpanderson:
Add Python tools for submitting batches of jobs (sample_submit_batch)
and for assimilating jobs from these batches (sample_assimilate.py).
Extend script_assimilator to supply the data needed by the latter.

### DIFF
--- a/tools/sample_assimilate.py
+++ b/tools/sample_assimilate.py
@@ -16,15 +16,22 @@ import sys, os
 if sys.argv[1] == '--error':
     error_code = sys.argv[2]
     wu_name = sys.argv[3]
-    batch_id = sys.argv[4]
-    outdir = 'sample_results/%s'%(batch_id)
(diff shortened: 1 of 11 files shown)


tools/sample_assimilate.py
```diff
@@ -0,0 +1,37 @@
...
+# in the 2nd case, append a line of the form
+#   wu_name error_code
+# to samples_results/batch_id/errors
+
+import sys, os
+
+if sys.argv[1] == '--error':
+    error_code = sys.argv[2]
+    wu_name = sys.argv[3]
+    wu_id = sys.argv[3]
```
AenBleidd, line 19: @davidpanderson, here's a typo, I suppose it should be `sys.argv[4]`

AenBleidd (changes requested):
Please fix one critical typo

AenBleidd (approved):

PR #5495, opened 2024-01-19
Web: sanitize user URLs to prevent XSS attacks

### LINKED ISSUE #5491: URL field not sanitized on edit_user_info_form.php
**Description**

When editing user data upon registration on the project, or via "home.php", the user is directed to edit_user_info_form.php (with a token) and can update fields like URL, country and so on.

The URL field is not sanitized, hence there is a possibility to insert malicious code or cross-site scripting in the field, that will be displayed when other users consult the user information with show_user.php, e.g. via the message boards.

The URL field and other fields on this page should be checked and sanitized to prevent XSS or other undesirable content.

**Steps To Reproduce**
1. Edit user information (from home.php or upon account registration)
2. It is possible to insert the following exemplary value into the “URL” field:
`_" style="display: block; position: fixed; top: 0; left: 0; z-index: 99999; width: 9999px; height: 9999px;" onmouseover="alert('xss')" _ `

The field will be accepted and when the user profile page is opened up again, the XSS is triggered in the browser, as a popup.
Please note that a similar XSS problem exists when changing the e-mail address of a user. It can, for example, be triggered using the page: ` edit_email_action.php?email_addr=<script>alert('XSS')<%2Fscript>&passwd=`

**Expected behavior**
It is recommended to both validate whether the entered data is indeed a correct URL and to apply output encoding mechanisms, so that user data cannot influence the application’s HTML structure. 

**Screenshots**
Please see POC from our project: 
image

**System Information**
 - OS: Multiple, server platforms e,g,, CentOS 7 and RHEL8
 - BOINC Version: Server release 1.4.2.

**Additional context**

davidpanderson:
Add a function sanitize_user_url() to do this.
It accepts things like

google.com
http://google.com
https://google.com?blah=foo&x=y

but nothing else.
There doesn't seem to be a PHP function that works, and stack overflow didn't yield anything plausible.

When showing a user page,
show 'Invalid URL' if it's nonempty and doesn't pass this.

When a user edits their info, show error page if they enter an invalid URL.

Fixes #5491

### DIFF
--- a/html/inc/user.inc
+++ b/html/inc/user.inc
@@ -226,6 +226,8 @@ function weak_auth($user) {
 // it to a canonical form, with the protocol prefix.
 //
 function normalize_user_url($url) {
+    $url = sanitize_user_url($url);
+    if (!$url) return '';
     $x = strtolower($url);
     if (substr($x, 0, 7) == 'http://') {
         return 'http://'.substr($url, 7);
@@ -251,7 +253,10 @@ function show_user_info_private($user) {
     }
     if (USER_URL && $user->url) {
         $u = normalize_user_url($user->url);
-        row2(tra("URL"), sprintf('<a href="%s">%s</a>', $u, $u));
+        row2(
+            tra("URL"),
+            $u?sprintf('<a href="%s">%s</a>', $u, $u):tra('Invalid URL')
+        );
     }
     if (USER_COUNTRY) {
         row2(tra("Country"), $user->country);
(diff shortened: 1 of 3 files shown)


AenBleidd:
@nilsjh, could you please check that this PR fixes the issues?
Thank you in advance.

AenBleidd (approved):

nilsjh:
Thanks! When applied to our dev project, I get "Invalid URL" when trying both the XSS POF and other invalid URLs.

PR #5502, opened 2024-01-26
Mac: fix Apple Silicon GPU errors after MacOS update

### LINKED ISSUE #5456: [macOS] [OpenCL] [ARM] Boinc AppleM (ARM) client cannot run OpenCL programs after update to Sonoma
**Describe the bug**

Important Note: this bug may be specific only to AppleM (i.e. ARM) version of Mac OS and/or client running on M1/M2 CPU.

After recent update of MacOS to Sonoma, all OpenCL programs from all projects started to fail with following error message from Mac OpenCL stack:
```
UNSUPPORTED (log once): buildComputeProgram: cl2Metal failed
<built-in>:1:10: fatal error: cannot open file './opencl-c.h': Permission denied
#include "opencl-c.h"
         ^
```
And application exits because OpenCL kernel cannot be compiled.

I've asked one of users to run this application from command line under normal user account. It worked without errors.

```
geneferg version 22.12.2 (macOS arm64, clang-15.0.3, boinc-7.20.2)
Copyright (c) 2022, Yves Gallot
genefer is free source code, under the MIT license.

Command line: '-h'

Running on device 'Apple M1', vendor 'Apple', version 'OpenCL 1.2 ', driver '1.2 1.0'.
330000000^{2^15} + 1: 00:02:16, 0.147 ms/bit, data size: 1.12 MB.
200000000^{2^16} + 1: 00:08:17, 0.275 ms/bit, data size: 2.25 MB.
... (and so on) ...
```

This makes me think that Mac ARM Boinc client or its account **has not enough privileges**, causing OpenCL stack to fail in its operations when application is started from Boinc client.

**Steps To Reproduce**
1. Get a Mac with M1 or M2 processor.
2. Attach to project with native support for M1/M2 GPU (for example, any of "Genefer" or "AP27" applications on PrimeGrid).
3. All tasks will fail with error above.

To check that application is working under normal user account:
1. Download Genefer AppleM GPU application from https://primegrid.com/download/genefer22g_macarm64_22.12.02
2. If gmp is not installed in the system (using homebrew or somehow else), download https://www.primegrid.com/download/macarm64_libgmp.10.dylib , rename it to `libgmp.10.dylib` and put in same folder as main executable.
3. Make main file executable: `chmod +x genefer22g_macarm64_22.12.02`
4. Run program in "benchmark" mode: `./genefer22g_macarm64_22.12.02 -h`
5. The program should run and print some information about throughput of your system like in the log above.

**Expected behavior**
It worked before update to Sonoma

**System Information**
 - OS: Mac OS ARM, M1/M2 CPU/GPU.
 - BOINC Version: 7.22.2, 7.24.2

CharlieFenton:
User boinc_master should have primary group boinc_master. User boincproject should have primary group boincproject. 
But updating MacOS to a new version sometimes changes the PrimaryGroupID of users boinc_master and boincproject to staff (20). This seems to be responsible for OpenCL projects failing after a MacOS update with the following error:
`UNSUPPORTED (log once): buildComputeProgram: cl2Metal failed
<built-in>:1:10: fatal error: cannot open file './opencl-c.h': Permission denied
#include "opencl-c.h"`
 
Fixes #5456

This PR makes the following changes:
* Added a check of the Primary Group IDs of users boinc_master and  boinc_project along with the other permissions checks when the Mac Manager or client is launched. If incorrect, refuses to run and displays an alert "BOINC ownership or permissions are not set properly; please reinstall BOINC".
* The Mac Installer: always sets Primary Group of users boinc_master boinc_project. (It previously did that  only if those users did not already exist.)

### DIFF
--- a/client/check_security.cpp
+++ b/client/check_security.cpp
@@ -80,12 +80,17 @@ int use_sandbox, int isManager, char* path_to_error, int len
     struct stat         sbuf;
     int                 retval;
     int                 useFakeProjectUserAndGroup = 0;
+    int                 isMacInstaller = 0;
 
     useFakeProjectUserAndGroup = ! use_sandbox;
 #ifdef _DEBUG
 #ifdef DEBUG_WITH_FAKE_PROJECT_USER_AND_GROUP
     useFakeProjectUserAndGroup = 1;
 #endif
+#ifdef _MAC_INSTALLER
+    isMacInstaller = 1;
+#endif
+
 #if (defined(__APPLE__) &&  ! defined(_MAC_INSTALLER))
     // Debugging Mac client or Mac BOINC Manager
     // Normally, the Mac Development (Debug) builds do not define SANDBOX, so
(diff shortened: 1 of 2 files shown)


AenBleidd (approved):

CharlieFenton:
I have built and  uploaded BOINC 7.24.3. This is a hot fix for issue #5456. The changes affect only the permissions checks when BOINC is started and the way the BOINC installer sets permissions. It makes no changes to the way BOINC computes or interacts with project servers. Therefore, I have immediately promoted it to recommended version.

PR #5504, opened 2024-01-29
[wrapper] improve handling of descendant processes

piedar:
**Description of the Change**

There are two commits here - a small (hopefully) uncontroversial fix and a proposed wrapper change.

I found a need for these while running boinc tasks as a subprocess of a [custom profiler](https://github.com/piedar/perfdata/blob/main/perfdata).  My worker processes were getting killed too soon, losing hours worth of profiling data.  First I discovered that systemd was killing my processes too early, but this was easily solved by setting [`KillMode=mixed`](https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html#KillMode=).

Then I noticed that the boinc wrapper was killing my processes immediately on shutdown, even though it's supposed to wait for 10 seconds.  It seems that `sleep()` can return early if a signal is handled during the sleep.  Luckily there is already a `boinc_sleep()` function which takes care of this situation.

- [fix: make sure sleep completes before killing descendants](https://github.com/BOINC/boinc/commit/5fa8994e71d1f6b66948329296716454d6bea314)

Then I wondered if 10 seconds might not be enough for all cases, so I also propose only killing wrapper descendants in the same process group as the wrapper process.

- [feat: filter kill_descendants by process group](https://github.com/BOINC/boinc/commit/75ea5fdbdde736cc9c3b47ced26b89455805027d)

It seems like `kill_descendants()` is meant to prevent defunct processes hanging around when the job process fails to clean them up properly.  But I think this is a reasonable compromise - if a process specifically puts its children in another process group, that's a good sign that they shouldn't be killed by the supervisor.

**Alternate Designs**

- considered passing a predicate function to filter descendant processes
  - adds complexity with lambdas and `auto` variables
  - depends on newer c++ version
- considered using `::kill(-child_process, SIGTERM)` to kill the entire process group
  - could simplify code, but would have been a larger change
  - might not work reliably, as there's no guarantee that the `child_process` is also the group leader
- considered filtering based on [session vs process group](https://stackoverflow.com/questions/6548823/use-and-meaning-of-session-and-process-group-in-unix)
  - session seems to be related more with interactive terminals

**Release Notes**

The boinc wrapper cleans up processes more reliably, and does not kill descendants which have been moved to other process groups.


davidpanderson:
I can't figure out what problem is being addressed here.
If you want to profile a program why involve the wrapper?

AenBleidd:
I don't agree with this PR. We need to be sure that we're keeping device as clean as possible, to avoiding to kill processes in a different group might lead to the situation when badly designed application could easily left a ton of processes that consumes the resources, and a regular user will never know what to do.

piedar:
> I can't figure out what problem is being addressed here. If you want to profile a program why involve the wrapper?

The boinc wrapper?  It handles RPC calls from boinc-client and supervises the application process.  (In my case, this is a script which starts the real application along with an attached profiler.)  I guess I could replace it with a custom wrapper which works more how I want.

> I don't agree with this PR. We need to be sure that we're keeping device as clean as possible, to avoiding to kill processes in a different group might lead to the situation when badly designed application could easily left a ton of processes that consumes the resources, and a regular user will never know what to do.

Applications can already catch `SIGTERM` and keep going.  The idea was to allow them to mark certain subprocesses as vital so they can finish up in their own time rather than being terminated after 10 seconds.  But I see your point, and I'll gladly withdraw the second commit and find another way to accomplish this.

davidpanderson:
That change looks good - thanks.

PR #5505, opened 2024-01-29
client/server/web: add 'Apple GPU' as an official GPU type.

davidpanderson:
For now, Apple GPUs are detected only via OpenCL (like Intel GPUs).
Eventually we'll detect via Metal as well.

Plan classes for Apple GPU must include 'apple_gpu' in their name.

The changes are mostly copy-and-paste of Intel GPU code.
I hate duplicate code but I don't know how to avoid it here.


lib/gui_rpc_client_print.cpp
```diff
@@ -299,6 +299,20 @@ void HOST_INFO::print() {
...
+    if (cap.count) {
+        printf("  Apple GPU\n");
+        if (cap.count > 1) {
+            printf("    Count: %d\n", cap.count);
+        }
+        if (cap.have_opencl) {
+            cap.opencl_prop.peak_flops = cap.peak_flops;
+            cap.opencl_prop.opencl_available_ram = cap.opencl_prop.global_mem_size;
+            cap.opencl_prop.is_used = COPROC_USED;
+            cap.opencl_prop.description(buf, sizeof(buf), "Intel GPU");
```
AenBleidd, line 312: Is `Intel GPU` valid description here? Shouldn't it be `Apple GPU` instead?

AenBleidd:
For now it's ok to have code duplicates. The whole GPU detection class should be refactored once, but I'm ok with the current state of it.
Also, as a first solution, I think we can use the OpenCL detection, since it's still used by Apple and right now this makes out code more uniform and easy to use.
Later we can think about the implementation of the same functionality using the Metal API (I can take a look at it if needed).

lib/gui_rpc_client_print.cpp
davidpanderson: Oops!  Fixed.

CharlieFenton:
I think you want `#if (defined (__APPLE__) && defined (__arm64__))` rather than just `ifdef __APPLE__` guarding some or all of these changes. Otherwise, I think it will apply the code to ATI, NVIDIA and Intel GPUs on Intel Macs while you intended it to be used only for Apple Silicon (Apple branded) GPUs.

davidpanderson:
I tested this on an M2 Mac, and it works;
it reports the GPU and able to do OpenCL and Metal,
and reports the 'Metal support' number (3).
When I turn off OpenCL, it correctly reports the GPU via Metal.

I added plan class support for Apple GPU, and documented it.

I think this is ready to start testing with a live project like E@h.

CharlieFenton:
You need to first test it on an Intel Mac to ensure you don't get false positives. See my [previous comment](https://github.com/BOINC/boinc/pull/5505#issuecomment-1926431335).

davidpanderson:
The code ignores non-Apple Silicon GPUs.

AenBleidd:
@davidpanderson, there is one minor issue with the code:
```
Found files with errors:
./client/gpu_detect.cpp:[6](https://github.com/BOINC/boinc/actions/runs/7781387969/job/21215728195?pr=5505#step:4:7)2
To fix errors run locally:
```
Could you please fix that? That's not critical, and I can fix it before merging your PR.

@CharlieFenton, please review this PR.

For me it looks fine. I do not like compiling and running this code on non-ARM laptops, but I can live with that. @CharlieFenton, @davidpanderson, I'll leave this decision to you.

davidpanderson:
I tested this on an Intel Mac and it works.

CharlieFenton:
> @CharlieFenton, please review this PR.

I think it looks reasonable.

PR #5550, opened 2024-03-24
Add an optional sleep function to lib/util.cpp

computezrmle:
Typical usecase:
Sleep commands are often called in loops when a process waits for the availability of a resource.
BOINC provides a universal sleep function "boinc_sleep" in lib/util.cpp.

This PR adds "boinc_wobble_sleep" to lib/util.cpp.
It can be seen as a wrapper to boinc_sleep adding some configurable randomness.

### DIFF
--- a/lib/util.cpp
+++ b/lib/util.cpp
@@ -31,6 +31,10 @@
 #define M_LN2      0.693147180559945309417
 #endif
 
+#ifndef BOINC_WOBBLE_SLEEP_MIN_SECONDS
+#define BOINC_WOBBLE_SLEEP_MIN_SECONDS 0.01
+#endif
+
 #include "boinc_stdio.h"
 
 #ifndef _WIN32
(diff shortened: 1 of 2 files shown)


davidpanderson:
I don't see where this would be useful, and it seems overly complex

computezrmle:
I'm working on a real usecase related to vboxwrapper where this is very helpful.
Just want to share it with others in case somebody has similar requirements.
lib/utils.cpp seems to be the right place for it.

AenBleidd:
I don't see a use case either.
Also, in its current state PR is failed to build on some platforms (we can't use C++11 when building libboinc)

computezrmle:
> we can't use C++11 when building libboinc
Any alternative?
It compiles without issues on my local machine using the most recent source code copy.

AenBleidd:
@computezrmle, maybe smth like this: https://github.com/BOINC/boinc/blob/b2af30757e1a8e8009edb6877869d515103b8d7e/lib/crypt_prog.cpp#L88-L121
?

computezrmle:
@AenBleidd 
How can I reference random_int() from util.cpp?
I would expect a header file like crypt_prog.h where it is declared but can't find any.

What works (at least locally for Linux) is to copy the code to util.cpp and the declare the random_int() in util.h.
Not an elegant solution, right?

Same for the small helper function "die".

AenBleidd:
@computezrmle, yes, the best would be to move this function to some other file (e.g. util.cpp as you suggested), and reference it both from your function and `crypt_prog` application too.
Regarding the `die()` function, it's better to change the `crypt_prog.cpp` to get the return value from the function and call `die()` at the place of `random_int()` usage (the best for me in case of error to throw an exception in the function `random_int()` and then catch it where this function called, but this need to be mentioned in the function description that it might throw an exception when fails).

AenBleidd:
@computezrmle, but before doing that, it still would be nice to discuss the use case, and also the signature of the function. Currently it looks overcomplicate.

computezrmle:
The usecase (in short)

Long term experience at LHC@home shows that vboxwrapper is prone to a race condition when several events happen together:
- an app_version uses multiattach vdis
- a previous VM has finished -> vbox moves the parent disk entry to the wrong location
- **many** subsequent VMs start concurrently

If only 1 VM starts it can sanitize this situation in most cases.
Many concurrently starting VMs often fail since deregistering is part of the workaround that must be done.

I plan to cover the critical workaround code with a lock.
Since other vboxwrapper instances must wait until the lock is released they should call sleep (better boinc_sleep) in a loop.

First tests on different computers have shown that a wobbling sleep results in much less loop cycles per process and a shorter total time until all instances have left the critical section.


More details when the development and tests have gone more forward.



The wobbling sleep function came out as an independent side effect and as already mentioned it may be useful in other cases.
Hence the idea to put it in the global utils beside boinc_sleep.


I wonder where you see too much complexity.
The source code comments should explain each step and finally it calls the already existing boinc_sleep to avoid exactly this.

davidpanderson:
You can do, e.g.
boinc_sleep(1+0.1*drand())

computezrmle:
> You can do, e.g. boinc_sleep(1+0.1*drand())

boinc_sleep(x+y*drand()) works fine, especially if x and y are well chosen.
Will close this PR.
Please mark it as "not necessary".

PR #5555, opened 2024-03-26
Move random_int() to util.cpp

computezrmle:
`random_int()` can easily be used as seed for `srand()` from various functions if it is in `util.cpp`.
This is not possible if it is in `crypt_prog.cpp` since `crypt_prog.cpp` is not included in the standard build process.
Instead, `tools/makefile_sign_executable` must be used to build it.

See:
https://boinc.berkeley.edu/trac/wiki/KeySetup#Filesigningutilities

### DIFF
--- a/lib/crypt_prog.cpp
+++ b/lib/crypt_prog.cpp
@@ -90,6 +90,7 @@ int main(int argc, char** argv) {
     R_RSA_PUBLIC_KEY public_key;
     R_RSA_PRIVATE_KEY private_key;
     int i, n, retval;
+    unsigned int srand_seed;
     bool is_valid;
     DATA_BLOCK signature, in, out;
     unsigned char signature_buf[256], buf[256], buf2[256];
@@ -120,7 +121,18 @@ int main(int argc, char** argv) {
         printf("creating keys in %s and %s\n", argv[3], argv[4]);
         n = atoi(argv[2]);
 
-        srand(random_int());
+        retval = random_int(srand_seed);
+        if (retval == 1) {
+            die("can't load ADVAPI32.DLL");
+        } else if (retval == 2) {
+            die("can't open /dev/random");
+        } else if (retval == 3) {
+            die("can't read from /dev/random");
+        } else if (retval) {
+            die("random_int");
+        }
+        srand(srand_seed);
+        
         e = BN_new();
         retval = BN_set_word(e, (unsigned long)65537);
         if (retval != 1) {
--- a/lib/crypt_prog.cpp
+++ b/lib/crypt_prog.cpp
@@ -132,7 +132,7 @@ int main(int argc, char** argv) {
             die("random_int");
         }
         srand(srand_seed);
-        
+
         e = BN_new();
         retval = BN_set_word(e, (unsigned long)65537);
         if (retval != 1) {
--- a/lib/crypt_prog.cpp
+++ b/lib/crypt_prog.cpp
@@ -56,6 +56,7 @@
 
 #include "crypt.h"
 #include "md5_file.h"
+#include "util.h"
 
 void die(const char* p) {
     fprintf(stderr, "Error: %s\n", p);
@@ -83,41 +84,6 @@ void usage() {
         "-cert_verify file signature certificate_dir\n"
         "    verify a signature using a directory of certificates\n"
    );
-}
-
-unsigned int random_int() {
-    unsigned int n;
-#if defined(_WIN32)
-#if defined(__CYGWIN32__)
-    HMODULE hLib=LoadLibrary((const char *)"ADVAPI32.DLL");
-#else
-    HMODULE hLib=LoadLibrary("ADVAPI32.DLL");
-#endif
-    if (!hLib) {
-        die("Can't load ADVAPI32.DLL");
-    }
-    BOOLEAN (APIENTRY *pfn)(void*, ULONG) =
-        (BOOLEAN (APIENTRY *)(void*,ULONG))GetProcAddress(hLib,"SystemFunction036");
-    if (pfn) {
-        char buff[32];
-        ULONG ulCbBuff = sizeof(buff);
-        if(pfn(buff,ulCbBuff)) {
-            // use buff full of random goop
-            memcpy(&n,buff,sizeof(n));
-        }
-    }
-    FreeLibrary(hLib);
-#else
-    FILE* f = fopen("/dev/random", "r");
-    if (!f) {
-        die("can't open /dev/random\n");
-    }
-    if (1 != fread(&n, sizeof(n), 1, f)) {
-        die("couldn't read from /dev/random\n");
-    }
-    fclose(f);
-#endif
-    return n;
 }
 
 int main(int argc, char** argv) {
--- a/lib/util.cpp
+++ b/lib/util.cpp
@@ -565,13 +565,14 @@ int random_int(unsigned int &n) {
     }
     FreeLibrary(hLib);
 #else
-    FILE* f = fopen("/dev/random", "r");
+    FILE* f = boinc::fopen("/dev/random", "r");
     if (!f) {
         return 2;
     }
-    if (1 != fread(&n, sizeof(n), 1, f)) {
+    if (1 != boinc::fread(&n, sizeof(n), 1, f)) {
         return 3;
     }
-    fclose(f);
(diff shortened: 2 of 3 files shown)


AenBleidd:
@computezrmle, please fix the failed build.
Thank you

computezrmle:
As for Release-x64-msbuild
`2024-03-26T21:24:45.8536712Z          D:\a\boinc\boinc\win_build\vcpkg_3rdparty_dependencies.vcxproj(182,5): error MSB3073: The command "vcpkg.exe install --x-manifest-root=D:\a\boinc\boinc\win_build\..\3rdParty\vcpkg_ports\configs\msbuild/x64 --x-install-root=D:\a\boinc\boinc\win_build\..\3rdParty\Windows\vcpkg/installed/ --overlay-ports=../../vcpkg_ports/ports --overlay-triplets=../../vcpkg_ports/triplets/ci --triplet x64-windows-static --clean-after-build" exited with code -1.`
Is it caused by a missing entry in `win_build\vcpkg_3rdparty_dependencies.vcxproj`?
If so, I would need a hint since I'm not familiar with the syntax.

As for Release-ARM64-msbuild
`2024-03-26T21:25:11.3790575Z          D:\a\boinc\boinc\lib\util.cpp(552,18): error C2664: 'HINSTANCE LoadLibrary(LPCWSTR)': cannot convert argument 1 from 'const char [13]' to 'LPCWSTR' [D:\a\boinc\boinc\win_build\boincmgr.vcxproj]`
Could this be a multi-byte/unicode issue?
If so, could it be solved using `LoadLibrary(L"ADVAPI32.DLL")`?

lib/util.cpp
```diff
@@ -542,3 +542,41 @@ bool process_exists(int pid) {
...
 
 #endif
+
+unsigned int random_int() {
+    unsigned int n;
+#if defined(_WIN32)
+#if defined(__CYGWIN32__)
+    HMODULE hLib=LoadLibrary((const char *)"ADVAPI32.DLL");
+#else
+    HMODULE hLib=LoadLibrary("ADVAPI32.DLL");
```
AenBleidd, line 552:
```suggestion
    HMODULE hLib=LoadLibraryA("ADVAPI32.DLL");
```

lib/util.cpp
```diff
@@ -542,3 +542,41 @@ bool process_exists(int pid) {
...
+#endif
+    if (!hLib) {
+        fprintf(stderr, "Error: can't load ADVAPI32.DLL\n");
+        exit(2);
```
AenBleidd, line 556: Function call should not exit application in case of failure.

AenBleidd (changes requested):

lib/util.cpp
```diff
@@ -542,3 +542,38 @@ bool process_exists(int pid) {
...
+#if defined(_WIN32)
+#if defined(__CYGWIN32__)
+    HMODULE hLib=LoadLibrary((const char *)"ADVAPI32.DLL");
+#else
+    HMODULE hLib=LoadLibraryA("ADVAPI32.DLL");
+#endif
+    if (!hLib) {
+        fprintf(stderr, "Error: can't load ADVAPI32.DLL\n");
+    }
+    BOOLEAN (APIENTRY *pfn)(void*, ULONG) =
```
AenBleidd, line 557: If the library was not loaded correctly two lines above - this will fail with an exception

lib/util.cpp
```diff
@@ -542,3 +542,38 @@ bool process_exists(int pid) {
...
+    FILE* f = fopen("/dev/random", "r");
+    if (!f) {
+        fprintf(stderr, "Error: can't open /dev/random\n");
+    }
+    if (1 != fread(&n, sizeof(n), 1, f)) {
+        fprintf(stderr, "Error: can't read from /dev/random\n");
+    }
+    fclose(f);
+#endif
+    return n;
```
AenBleidd, line 578: What will be returned in case of an error?
We should indicate somehow that the function returns a proper result.

lib/util.cpp
```diff
@@ -542,3 +542,38 @@ bool process_exists(int pid) {
 }
 
 #endif
+
+unsigned int random_int() {
```
AenBleidd, line 546: I suggest to change signature function to something like this:
```bool random_int(unsigned int& value)```
And return the random value using the output parameter, and return `true` on success and `false` in case of an error

Comment (AenBleidd) [reply lib/util.cpp:546]: Also, then you need to adjust `crypt_prog.cpp` with the new function signature

**Review by AenBleidd — CHANGES_REQUESTED** [review]
There are still two failing builds.
Please check them and fix.
Thank you in advance.

**Comment by davidpanderson** [comment]
Is there a need for this outside of crypt_prog?
You can call srand() with, e.g. the time of day in microseconds.

**Comment by computezrmle** [comment]
Can the recent build issues be solved without too much effort?

1st idea to get a random number was to use the standard library functions from C++11.
Looks like certain C++11 functions (including those) don't work for BOINC.

Next idea was to use `rand()` which needs `srand(seed)`.
Since `seed` is expected to be an unsigned int `random_int()` from `crypt_prog` seemed to be a good idea as
- it is available in the source tree
- it delivers an unsigned int
- it uses a "natural" random source like `/dev/random`.

As `random_int()` has build issues for the client when used from `crypt_prog` the next idea was to move it to `lib/util` which would make it globally available.
The build issues for the server it has now came unexpected.
If they can be solved, why not leave `random_int()` in `lib/util`?

Other sources for seed appear to be less "natural".
A seed base on daytime may deliver a good seed if it is not too coarse.

Can you confirm that `dtime()` from `lib/util` returns a double like 86345.123456?
I'm sure it does on Linux.
What about Windows or GCL_SIMULATOR?

If yes, this may be fine enough to cover concurrently starting processes:
`seed = (unsigned int)(100000 * dtime())`

or, to avoid `seed` becomes greater than 32-bit max unsigned int:
`seed = (unsigned int)(1000 * dtime())`
This would give a granularity of 0.1 ms.

File: lib/util.cpp [hunk 6]
```diff
@@ -542,3 +542,36 @@ bool process_exists(int pid) {
...
+        char buff[32];
+        ULONG ulCbBuff = sizeof(buff);
+        if(pfn(buff,ulCbBuff)) {
+            // use buff full of random goop
+            memcpy(&n,buff,sizeof(n));
+        }
+    }
+    FreeLibrary(hLib);
+#else
+    FILE* f = fopen("/dev/random", "r");
```
AenBleidd, line 568:
```suggestion
    FILE* f = boinc::fopen("/dev/random", "r");
```

lib/util.cpp
```diff
@@ -542,3 +542,36 @@ bool process_exists(int pid) {
...
+    if (!f) {
+        return 2;
+    }
+    if (1 != fread(&n, sizeof(n), 1, f)) {
```
AenBleidd, line 572:
```suggestion
    if (1 != boinc::fread(&n, sizeof(n), 1, f)) {
```

lib/util.cpp
```diff
@@ -542,3 +542,36 @@ bool process_exists(int pid) {
...
+        return 3;
+    }
+    fclose(f);
```
AenBleidd, line 575:
```suggestion
    boinc::fclose(f);
```

lib/crypt_prog.cpp
```diff
@@ -154,7 +121,18 @@ int main(int argc, char** argv) {
...
+            die("can't load ADVAPI32.DLL");
+        } else if (retval == 2) {
+            die("can't open /dev/random");
+        } else if (retval == 3) {
+            die("can't read from /dev/random");
+        } else if (retval) {
+            die("random_int");
+        }
+        srand(srand_seed);
+
```
AenBleidd, line 135: run locally this command to remove trailing spaces:
`python3 ci_tools/trailing_whitespaces_check.py . --fix`

AenBleidd (changes requested):

AenBleidd:
> Is there a need for this outside of crypt_prog? You can call srand() with, e.g. the time of day in microseconds.

I think, having this as a generally available function won't hurt and could be helpful.

AenBleidd (approved):
Looks good to me.
@davidpanderson, if you are ok with this - feel free to merge it

davidpanderson:
Sorry, but I don't see a use case for this.

PR #5577, opened 2024-04-11
client: if GPU is missing, discard app versions and results that refer to it.

davidpanderson:
This addresses an issue introduced when we changed GPU names from (e.g.) 'Apple M3 Pro' to 'apple_gpu'.

Note: this means that if a host has in-progress jobs using a discrete GPU, and you remove that GPU, those jobs will be discarded. This is a change, which I think is good because otherwise you'd get error messages forever.

### DIFF
--- a/client/cs_statefile.cpp
+++ b/client/cs_statefile.cpp
@@ -293,13 +293,18 @@ int CLIENT_STATE::parse_state_file_aux(const char* fname) {
                 }
             }
             if (avp->missing_coproc) {
-                msg_printf(project, MSG_INFO,
-                    "App version uses missing GPU '%s' - discarding",
-                    avp->missing_coproc_name
-                );
                 if (strstr(avp->missing_coproc_name, "Apple ")) {
+                    msg_printf(project, MSG_INFO,
+                        "App version uses deprecated GPU type '%s' - discarding",
+                        avp->missing_coproc_name
+                    );
                     delete avp;
                     continue;
+                } else {
+                    msg_printf(project, MSG_INFO,
+                        "App version uses missing GPU '%s'",
+                        avp->missing_coproc_name
+                    );
                 }
             }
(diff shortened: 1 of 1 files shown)


AenBleidd:
@davidpanderson, I think this is quite a dangerous change. Please see my email for more details.

RichardHaselgrove:
I agree with Vitalii - this change needs, at the minimum, more thought. One significant problem: with modern versions of Windows, an operating system update and reboot can take place "outside working hours" without the explicit approval of the user. The sequence of component restarts can be ill-defined: it is possible for BOINC to restart before the video drivers have loaded and are ready for use. In that case, the GPUs are logged as being missing - but it's only a transient situation. Shutting down the client and restarting it resolves the problem.

Discarding the task is wasteful of project resources and the user's internet bandwidth.

AenBleidd:
I'm making this PR a draft to discuss this change.

CharlieFenton:
Instead of this why not just allow the old _Apple GPU_ (or whatever it was) as an alias for the new name? That would provide backward compatibility.

CharlieFenton:
… or accept any GPU name containing _Apple_

davidpanderson:
It doesn't matter if a few jobs get discarded.

client/cs_statefile.cpp
```diff
@@ -294,9 +294,13 @@ int CLIENT_STATE::parse_state_file_aux(const char* fname) {
             }
             if (avp->missing_coproc) {
                 msg_printf(project, MSG_INFO,
-                    "Application uses missing %s GPU",
+                    "App version uses missing GPU '%s' - discarding",
```
AenBleidd, line 297: Is it really needed to show `discarding` for all cases, not only when Apple GPU is missing?
It might be a little bit misleading for the users.

davidpanderson:
Oops!  I fixed the messages in each case

PR #5608, opened 2024-05-01
Fix 'File name too long' on Darwin

computezrmle:
On LHC@home a user running Darwin 19.6.0 and Darwin 23.4.0 got the following error:

```
Could not set race mitigation lock.
Lockname: '/boinc_vboxwrapper_lock_c94b628801c684e7'
Error: 63, File name too long
```

A couple of internet sources state that on Darwin the filename size is limited to 31 characters plus trailing '0'.

### DIFF
--- a/samples/vboxwrapper/vbox_vboxmanage.cpp
+++ b/samples/vboxwrapper/vbox_vboxmanage.cpp
@@ -2342,7 +2342,7 @@ int VBOX_VM::set_race_mitigation_lock(int& fd_race_mitigator, string& lock_name,
     // lock_name should be derived from the full path of the disk's filename.
     //
     // Darwin limits the file name size to 32 characters (including the trailing '0').
-    // On POSIX we set a '/' as prefix. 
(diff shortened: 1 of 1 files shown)


CharlieFenton:
On my M1 iMac I have a file name with 78 characters plus period and 7 characters suffix. [This page](https://discussions.apple.com/thread/8370573?sortBy=best) suggests a reason:
> It isn't the zip file name that it is complaining about, it is the combination of path and filename length internal to the zip that may have exceeded the maximum 1024 byte path, and 255 character filename limit.

computezrmle:
Since LHC@home uses a vboxwrapper with this fix the Apple/Darwin computers previously returned `ENAMETOOLONG` now returning valid results.

I's currently used for CMS and Theory.

davidpanderson:
The name in this case is of a shared memory segment, not a file.
On Linux both have a limit of 255 chars, but perhaps they differ on Darwin.

PR #5620, opened 2024-05-08
web: add option for making user name unique

davidpanderson:
We currently allow different accounts to have the same user name.
Hence in places where we need to identify users
(sending PMs, blocking users) we have to use numeric IDs,
which means that we have to display them in various places.
This is ugly; I'm not sure why I thought it was a good idea.

Anyway.  Add a project.inc option UNIQUE_USER_NAME.  If set:
- it won't let you create an account with a dup name,
    of change your name to a dup
- it doesn't show user IDs anywhere
- in places where you specify users, you use names

This is now the default for new projects.
Existing projects - which have lots of duplicate names -
can't use it unless they de-dup their names somehow
(and this might anger users).

Also some minor code cleanup.  sprintf() is our friend.

### DIFF
--- a/db/constraints.sql
+++ b/db/constraints.sql
@@ -12,7 +12,7 @@ alter table user
     add unique(email_addr),
     add unique(authenticator),
     add index ind_tid (teamid),
-    add index user_name(name),
+    add unique(name),
     add index user_tot (total_credit desc),
         -- db_dump.C
     add index user_avg (expavg_credit desc),
--- a/html/inc/bootstrap.inc
+++ b/html/inc/bootstrap.inc
@@ -81,7 +81,7 @@ function navbar_right($user) {
         if ($user) {
             echo sprintf('
                 <li><a href=%s%s>%s</a></li>
-                ', url_base(), USER_HOME, $user->name
+                ', url_base(), HOME_PAGE, $user->name
             );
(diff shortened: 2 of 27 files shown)


cwallbaum:
So when there's new projects on the horizon and others are faster than me, it can easily happen that at project x I'm walli, at project y woofy and at z I'll have to become Chewbacca?! Sorry, but that's exactly what I find ugly. With this new default setting, you now have even fewer options to recognize me across different projects than with the old standard behaviour. And if I wanted to have different usernames, that was already possible before, now I'm more or less forced to that situation (as far as new project admins are unaware of this setting)... I don't like it and I'm in favour of not merging this pull request or at least please don't set this behaviour as the new default.

davidpanderson:
Currently, when you specify a user (e.g. to send a PM to several users) you enter their user name.
But if it's not unique you have to enter their database ID instead.
(How do you know if the name is unique?  Trial and error.)

So everywhere we show a user name we need to show their database ID too, like 'John (33)'.
But you can also have 'John (33)' as a user name.

This is a mess.  There should be one and only one way to identify a user.

Having the same name across projects is fine,
but it doesn't accomplish much if there are other users with the same name on some of the projects.

AenBleidd (approved):

PR #5632, opened 2024-05-23
[linux][client] Fix check whether the client is configured to run as a daemon.

### LINKED ISSUE #5601: Recent commit to the BOINC Manager, to not enable stopping the BOINC client when configured as a systemd service, results in anomaly.
The commit is:
https://github.com/BOINC/boinc/commit/82a4fc80613264b18f358963b76aeb4697deb8c5

This greys out (disables) the File/Shut down connected client.. option if BOINC is running as a systemd service on the current system.

This works, but produces an anomaly.
I do not think it is working as intended.
It is certainly not working in any logical manner.

The check is done on the local system. However, the _same_ check is used (_i.e_ against the local system) even when the Manager is connected to a _remote_ system.

The result is that when I run the BOINC Manager on my Desktop system (which has BOINC running as a system service) it does not let me stop the client on **any** system.
However, on my laptop (which isn't running any BOINC client) the option to stop the client _is_ enabled when I connect to _any_ remote system - _including that Desktop system_ (the one where I cannot stop it when logged in locally).

I suspect that the option should always be disabled when connected to a remote system.

And I think there should be an option in the **Manager** code to enable the "Shut down" option for any connected, logged in client.

AenBleidd:
Remove 'Stop running tasks...' checkbox from the Exit Confirmation dialog on linux when the client is configured to run as a daemon. Small refactoring.

This fixes #5601.

### DIFF
--- a/clientgui/BOINCClientManager.cpp
+++ b/clientgui/BOINCClientManager.cpp
@@ -59,6 +59,7 @@ extern int diagnostics_get_process_information(PVOID* ppBuffer, PULONG pcbBuffer
 
 #else
 #include <sys/wait.h>
+#include <cstdio>
 #endif
 
 CBOINCClientManager::CBOINCClientManager() {
@@ -135,15 +136,18 @@ int CBOINCClientManager::IsBOINCConfiguredAsDaemon() {
 #if   defined(__WXMSW__)
     if (is_daemon_installed()) bReturnValue = 1;
 #elif defined(__WXMAC__)
-    if ( boinc_file_exists("/Library/LaunchDaemons/edu.berkeley.boinc.plist")) {
+    if (boinc_file_exists("/Library/LaunchDaemons/edu.berkeley.boinc.plist")) {
         bReturnValue = NewStyleDaemon;                      // New-style daemon uses launchd
     }
     if (boinc_file_exists("/Library/StartupItems/boinc/boinc") ) {
         bReturnValue = OldStyleDaemon;                      // Old-style daemon uses StartupItem
     }
 #else
-    if (boinc_file_exists("/usr/lib/systemd/system/boinc-client.service")) {
-        bReturnValue = 1;
+    FILE *f = popen("systemctl -q is-active boinc-client.service", "r");
+    if (f) {
+        if (pclose(f) == 0) {
+            bReturnValue = 1;
+        }
     }
 #endif
     return bReturnValue;
--- a/clientgui/DlgExitMessage.cpp
+++ b/clientgui/DlgExitMessage.cpp
@@ -29,6 +29,8 @@
 #include "miofile.h"
 #include "parse.h"
 #include "BOINCGUIApp.h"
+#include "MainDocument.h"
+#include "BOINCClientManager.h"
 #include "SkinManager.h"
 #include "DlgExitMessage.h"
 #ifdef __WXMAC__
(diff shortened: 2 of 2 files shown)


original-birdman:
> This fixes https://github.com/BOINC/boinc/issues/5601.

Difficult to tell for sure from just the diffs, but I'm not convinced.

AFAICT you are still just checking for boinc running as a service on the _local_ system even when the Manager is connected to a remote on.

AenBleidd:
@original-birdman, yes, that's true. And that will remain this way for Linux. To improve it and make more flexible, we will need to implement additional logic in both client and manager, and this is not planned for now.

original-birdman:
Well, I've just built it and now the option is greyed-out even for remote systems when I run it on my laptop (where BOINC is not running).
Which is odd, as I reckon that `IsBOINCConfiguredAsDaemon()` will be 0 (False) and so `ShowShutdownCoreClientCheckbox()` should return the opposite, i.e. True.

But I now get the same result whether or not BOINC is set to run from systemd.

Indeed, if I rename the `boinc-client.service` on a system (so that the systemctl is-active command returns False), the option is _still_ not enabled.

So I'm not convinced this is working as intended.

AenBleidd:
@original-birdman, what do you mean 'greyed-out'?
If the BOINC client on linux is configured to run as a daemon, then there will be no checkbox at all.

original-birdman:
No. It's there in the menu, but greyed out.

AenBleidd:
@original-birdman, then your BOINC Manager is not connected to the BOINC Client

original-birdman:
True when on my laptop (where there is no client) but I saw the same on a system where BOINC was running.

AenBleidd:
I'll double check that

original-birdman:
Screenshot_20240523_183047

AenBleidd:
And on the screenshot above you have the client running as a daemon?

original-birdman:
It's runnign out of systemd.

And I get the same greyed-out option on a system that doesn't run BOINC at all.

Maeda1:
Hi,

I think there is a mistake in merge pull request #5632 : I'm running version 8.0.2 and installed Boinc as a service. The option is not anymore in the GUI but closing Boincmgr always stopping the active tasks. That command which seems to be used as per last commit does NOT correctly show the is-active state:

in commit line 146 in clientgui/BOINCClientManager.cpp:
`    FILE *f = popen("systemctl -q is-active boinc-client.service", "r");`

On my system that returns no value, but using:
`systemctl is-active boinc-client.service`

returns **active**

original-birdman:
> 
> in commit line 146 in clientgui/BOINCClientManager.cpp: ` FILE *f = popen("systemctl -q is-active boinc-client.service", "r");`
> 
> On my system that returns no value, ...

All that does is set the exit status (which `popen()` can check directly).
It specifically doesn't print any output as it isn't needed to get the result.

Try echoing $? immediately after running the command...

Maeda1:
Did it: $? equals so 0. Expected value?

original-birdman:
```
root@ganuc:~# systemctl -q is-active boinc-client.service
root@ganuc:~# echo $?
0
root@ganuc:~# systemctl stop boinc-client.service
root@ganuc:~# systemctl -q is-active boinc-client.service
root@ganuc:~# echo $?
3
```

It was running at the start.

Maeda1:
OK clear. Thanks.

PR #5642, opened 2024-06-06
Client: don't start jobs that will exceed RAM

### LINKED ISSUE #5641: Client starts set of jobs too large for system memory
(This problem was reported by Glenn from CPDN)

For each runnable job we have several estimates of its future WSS:

a) the project-supplied value rsc_memory_bound
b) APP_VERSION::max_working_set_size:
    the max measured WSS of jobs using this app version
    since the client started (not saved to state file)
c) if the job has already run, ACTIVE_TASK::working_set_size_smoothed:
    recent average (on the order of 1 min) of its WSS

Current policy:
In job scheduling, for job WSS we use
if job has run: c)
else if b) is nonzero: b)
else a)

Problem:
CPDN jobs run for a few minutes with small WSS (say, 1MB).
Then they grow to full WSS (say, 6 GB).
There are various scenarios in which this leads to problems.
E.g. suppose the host has 16GB RAM and 8 cores.
It gets 8 CPDN jobs.
It starts 2 of them.
Their WSS is measured as 1MB.
On the next reschedule the client starts 2 more CPDN jobs.
Eventually all 4 jobs expand to 6GB WSS.
This is bigger than RAM and maybe bigger than swap space.
Some of the jobs fail with memory allocation failure.

Solution:
Change the WSS policy to:
If job has run: max(a, c)
else max(b, c)

Note:
What happens if project's rsc_memory_bound is wildly wrong?
if too large:
    client may run fewer jobs for that project
if too small:
    same as current: client may run too many jobs,
    and they may fail (and possibly cause jobs of other projects to fail)
    with malloc failure

davidpanderson:
Fixes #5641

### DIFF
--- a/client/cpu_sched.cpp
+++ b/client/cpu_sched.cpp
@@ -1309,26 +1309,33 @@ bool CLIENT_STATE::enforce_run_list(vector<RESULT*>& run_list) {
         //      until it reaches deadline pressure.
         // So we'll go with 1).
 
-        // skip jobs whose working set is too large to fit in available RAM
+        // skip jobs whose 'expected working set size' (EWSS)
+        // is too large to fit in available RAM.
+        // The EWSS is the max of
+        // 1) the recent average WSS of the job
+        // 2) if it hasn't run yet,
+        //      the max WSS of other jobs of this app version
+        // 3) the WU's rsc_memory_bound
+        //      This handles the case of CPDN, where jobs run with
+        //      small WSS for a while and then get big.
         //
-        double wss = 0;
+        double ewss = 0;
         if (atp) {
             atp->too_large = false;
-            wss = atp->procinfo.working_set_size_smoothed;
+            ewss = atp->procinfo.working_set_size_smoothed;
         } else {
-            wss = rp->avp->max_working_set_size;
+            ewss = rp->avp->max_working_set_size;
         }
-        if (wss == 0) {
-            wss = rp->wup->rsc_memory_bound;
-        }
-        if (wss > ram_left) {
+        ewss = std::max(ewss, rp->wup->rsc_memory_bound);
+
+        if (ewss > ram_left) {
             if (atp) {
                 atp->too_large = true;
             }
             if (log_flags.cpu_sched_debug || log_flags.mem_usage_debug) {
                 msg_printf(rp->project, MSG_INFO,
-                    "[cpu_sched_debug] can't run %s: WS too big %.2fMB > %.2fMB",
-                    rp->name,  wss/MEGA, ram_left/MEGA
+                    "[cpu_sched_debug] skipping %s: estimated WSS (%.2fMB) exceeds RAM left (%.2fMB)",
+                    rp->name,  ewss/MEGA, ram_left/MEGA
                 );
             }
             continue;
@@ -1357,7 +1364,7 @@ bool CLIENT_STATE::enforce_run_list(vector<RESULT*>& run_list) {
 
         ncpus_used += rp->avp->avg_ncpus;
         atp->next_scheduler_state = CPU_SCHED_SCHEDULED;
-        ram_left -= wss;
+        ram_left -= ewss;
         if (have_max_concurrent) {
             max_concurrent_inc(rp);
         }


AenBleidd (approved):

Chamiu:
Oh ?

Until 8.0.2, it was able to process 4WU of [DENIS@home](https://denis.usj.es/denisathome/) at 16GBx4%, but when I installed 8.0.3dev, it was waiting for memory 1WU of 2WU.
[DENIS@home](https://denis.usj.es/denisathome/)'s 1WU need 1.7MB.
8.0.3dev. need 16GBx6%.

8.0.3dev want a 50% more RAM of 8.0.2 at start.

Should I verify it?
0. setting anytime 2WU.
1. down grade to 8.0.2
2. check memory usage (BOINC Manager and 2WU) at Activity Monitor.app
3. upgrade to 8.0.3dev.
4. check memory usage (BOINC Manager and 2WU) at Activity Monitor.app

Environment : macOS Sonoma.5 on MacBook Air Late 2020 (Apple M1)

Will only macOS behave in the opposite way ?
Can't it be verified without the situation in each OS ?

davidpanderson:
I don't understand the above.

What is rsc_memory_bound for those jobs?

Chamiu:
2.
BOINC 8 0 2
4.
BOINC 8 0 3dev

RichardHaselgrove:
Assistance please. I have a Linux Mint 20.3 machine running BOINC 8.0.2 under systemd from @LocutusOfBorg PPA. I'm working with Glenn to help him adjust to the BOINC ecosystem (he has an HPC background), and am trying to test this PR for CPDN.

I have downloaded linux_client_5e0d0db47cad2f1f27df4ae69175d5b813c21e90.zip from the artifacts for this PR - but it won't run on this machine: systemd won't start it, but no other error message or reason given.

Have I found the right artifact? Is there another one I could try, or should I resort to a private build?

AenBleidd:
@RichardHaselgrove, please use our own Linux installer: https://boinc.berkeley.edu/linux_install.php
Please keep in mind that this particular PR is in 8.0.3 alpha version only

RichardHaselgrove:
OK, I'll try that when I've reported all current work, and re-confirmed the mapping between Ubuntu and Mint version numbers.

Of course I know that 8.0.3 is alpha and bespoke for CPDN - that's why we want to test it.

We might need a Plan C ...

RichardHaselgrove:
Well, that was a complete disaster.

1) re the instructions to install OpenCL: the response was
```
Package opencl-icd is a virtual package provided by:
  libnvidia-compute-418-server 418.226.00-0ubuntu0.20.04.2
  mesa-opencl-icd 21.2.6-0ubuntu0.1~20.04.2
  nvidia-opencl-icd-340 340.108-0ubuntu5.20.04.2
  libnvidia-compute-535-server 535.161.08-0ubuntu2.20.04.1
  libnvidia-compute-535 535.171.04-0ubuntu0.20.04.1
  libnvidia-compute-470-server 470.239.06-0ubuntu0.20.04.1
  libnvidia-compute-470 470.239.06-0ubuntu0.20.04.1
  libnvidia-compute-450-server 450.248.02-0ubuntu0.20.04.1
  libnvidia-compute-390 390.157-0ubuntu0.20.04.1
  pocl-opencl-icd 1.4-6
  intel-opencl-icd 20.13.16352-1
  beignet-opencl-icd 1.3.2-7build1
You should explicitly select one to install.

E: Package 'opencl-icd' has no installation candidate
```
2) installed boinc-client boinc-manager

I now have the shared libraries boinc and boincmgr in /usr/bin (as before), but they are both datestamped Sun 19 Apr 2020 16:07:33 BST - I think that equates to about v7.16.6. And I can't start either of them.

Reverting to the PPA overnight.

RichardHaselgrove:
And we're running again. the same files are now datestamped Tue 28 May 2024 04:18:35 BST, and 2 GPU and 4 CPU tasks are running. Manager and client agree version is 8.0.2

Now about that Plan C ...

AenBleidd:
@RichardHaselgrove, because we are installing binaries to `/usr/local/bin` since our packages are not officially distributed with the corresponding distro

RichardHaselgrove:
Well, where did it get v7.16.6 from, then? And why wouldn't that even run?If there are hidden gotchas like that in the 'new' instructions, you need to be explicit about them.

AenBleidd:
@RichardHaselgrove, for me instructions are quite clear, but I'm a developer, and I see the things from a little bit different perspective. So let's make instructions better together ;)
1. For the user there should not be any difference where the binaries are located. We're putting them in that particular location because this is a recommended location for the packages that are not provided with a distro.
2. Everything should work OOTB ('should' doesn't mean it does), at least in my case all the files are located in the folder I specified.
3. If in your case it's different - than that means that you haven't installed **our** package. I do not now if the PPA provides 8.0.2 version, but you can try to update to 8.0.3 from the alpha and see, if you get it (please keep in mind that if you have boinc running as a daemon before upgrade, you need either to restart the boinc-client service manually or do the machine restart).

RichardHaselgrove:
OK, but not tonight - I've only just got back from a long road trip, and I'm knackered - barely thinking straight. Which is **not** the right time to do complicated software installs.

The PPA currently offers v8.0.2, so that's working as it's supposed to. I'm very used to running stop / disable / enable / start, to keep systemd happy around system restarts.

Perhaps the biggest take-home from today is that we need to consider both brand-new installs, and upgrades from old and messy working boxes - that'll be the difficult one to write.

AenBleidd:
> Perhaps the biggest take-home from today is that we need to consider both brand-new installs, and upgrades from old and messy working boxes - that'll be the difficult one to write.

This is something I have tried to address already in 8.0.2.
I have checked this manually some time ago, and it was working fine for me.
But if you find anything additional - then we can improve that together.
Thank you for your help!

PR #5658, opened 2024-06-16
vboxwrapper: don't create projects/virtualbox

davidpanderson:
If Vboxwrapper couldn't find the Vbox 'home dir' (normally ~/.VirtualBox) based on env vars, it would create one: projects/virtualbox (owned by boinc_projects, not boinc_master)

This is wrong because
1) on startup, the client deletes anything in projects/
    that's not a project directory
2) on Mac it causes installer error messages
    because the dir is not owned by boinc_master.

Fix: don't create projects/virtualbox.

### DIFF
--- a/samples/vboxwrapper/vbox_vboxmanage.cpp
+++ b/samples/vboxwrapper/vbox_vboxmanage.cpp
@@ -38,6 +38,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <sys/mman.h>
+#include <pwd.h>
 #endif
 
 using std::string;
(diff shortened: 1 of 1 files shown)


AenBleidd (approved):

AenBleidd:
@computezrmle, could you please review this change?

computezrmle:
The docs mention different directories to put `VirtualBox.xml` in depending on the host OS:
```
Linux and Oracle Solaris:$HOME/.config/VirtualBox.
Windows:$HOME/.VirtualBox.
macOS:$HOME/Library/VirtualBox.
```

It looks like the patch covers the situation on Windows only.

On Linux and Oracle Solaris `XDG_CONFIG_HOME` can be used instead of `VBOX_USER_HOME`.
This case may also be implemented.

davidpanderson:
Other than the comment, is the change OK?

PR #5694, opened 2024-07-18
sched/db_purge: if a result fails to be deleted, don't purge the referencing workunit

bema-aei:
In db_purge, the return value of purge_and_archive_results() is never checked and not dealt with. If a result fails to be deleted from the DB (for whatever reason), the referencing workunit is still purged, leaving this result behind "orphaned", i.e. without reference. It will never get purged from the DB.

This change will skip this workunit in case of an error from purge_and_archive_results(), so the purger can have another try when it comes across that workunit again.

### DIFF
--- a/sched/db_purge.cpp
+++ b/sched/db_purge.cpp
@@ -588,7 +588,12 @@ int purge_and_archive_results(DB_WORKUNIT& wu, int& number_results) {
             );
         } else {
             retval = result.delete_from_db();
-            if (retval) return retval;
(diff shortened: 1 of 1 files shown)


brevilo (approved):

bema-aei:
Note: this may lead to some results (the results up to the one where the error occurred) being archived in a different file (i.e. with a different timestamp) than the workunit referencing these. One should be aware of this when searching the archives for specific results.

Issue #5691, opened 2024-07-17
Missing DB connection closing


brevilo:
Various BOINC components don't properly close the DB connection they opened before. This is a problem because:
* Open connections are a limited resource on the database server and new clients might be rejected when all are used up.
* Left-open connections will eventually be aborted by the database server and logged as a `WARNING`, at least by [MariaDB](https://mariadb.com/kb/en/error-log/#verbosity-level-2). [MySQL](https://dev.mysql.com/doc/refman/8.4/en/communication-errors.html) still logs it as `INFORMATION` but MariaDB has a good reason to warn about those: the root cause can in fact be caused by misconfiguration (e.g. [`max_allowed_packet`](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_max_allowed_packet)) or networking issues. Both should not get masked by otherwise harmless missing connection closing.

Sampling only the current C API (e.g. used by daemons), you'll find:
* files issuing `boinc_db.open()`: [30](https://github.com/search?q=repo%3ABOINC%2Fboinc%20boinc_db.open(&type=code))
* files issuing `boinc_db.close()`: [9](https://github.com/search?q=repo%3ABOINC%2Fboinc%20boinc_db.close()&type=code)

This indicates a general issue, not just in the C API.

Thanks

bema-aei:
Also the Db_conn of the web code (in html/inc/db_conn.inc) is lacking a method to close a DB connection.

bema-aei:
There should at least be the possibility given to do it.

PR #5695, opened 2024-07-18
Allow to address a second replica DB "replica2" in the web code

bema-aei:
The current setup of Einstein@Home uses two replicas. This patch is to allow accessing the second replica in the web code by calling get_aux(2); get_aux(1) or get_aux(true) still addresses the standard replica.

For this to work, config.xml needs to have a second set of tags "<replica2_db_(host|name|user|password)>" analog to the current "<replica_db_(host|name|user|password)>".

This code will probably never be used by anyone else, it's just to bring upstream BOINC in sync with the web code that we're using on E@H for easier updates.


brevilo (commented):
There seem to be a few logic issues in `get_aux()`, at least when comparing code vs. comments. Eg. *original* lines:
- 39 vs. 60: should not error out for `fallback_mode = 1`
- 69 vs. 75: the else block doesn't imply `fallback_mode == 0`, if `(!$u || !$p || !$n) == false`
- 69 vs. 80: likewise, `fallback_mode == 0` isn't guaranteed here. If so, does the logic really represent the logic described for the three fallback modes?

I know these issues are independent of this PR but they need fixing anyway (if I'm not mistaken).

AenBleidd:
@davidpanderson, could you please take a look at this PR?

brevilo:
Also, thanks to @RichardHaselgrove, we should also make sure the [docs](https://github.com/BOINC/boinc/wiki/ProjectOptions#replica-database-info) get updated.

@davidpanderson while at it, please note that that [doc](https://github.com/BOINC/boinc/wiki/ProjectOptions#job-limits) contains a lot of syntax errors, presumably by a flawed trac -> markdown conversion. Just search for "\<" in it to see what I mean.

Thanks

davidpanderson:
The logic in get_aux() is too complex.
I'd prefer to clean it up (and handle replica2 in the process)

Is $fallback_mode needed?  I can't find any place that uses it.
It would be simpler if we assume that
- each replica has its own user/host/passwd in config.xml
- if connection to replica 1 or 2 fails,  use main DB

bema-aei:
The fallback_mode was introduced by Christian in 2015, see 3481806b6d418c0a33be08166f007d3de539b89c No idea (anymore) what it was used for, or in which projects, or which projects would break if it was removed.

davidpanderson:
I added support for multiple readonly replicas in #5695,
using the same <replica2_db_host> config syntax.
This PR also greatly simplifies the get_aux() logic.
Please see if it works for you.

brevilo:
You mean #5697

bema-aei:
I just pushed a new version of the branch that takes David's commits from #5697 and applies a few changes:

- including a change that affects the whole web code ("use mysqli everywhere") in another commit with very limited scope and not even mentioning it in the commit message is just asking for trouble.. So I split that commit in two (leaving the author intact).
- I added commit d1f99043877552ea48ac2912dc4ec9ac9a1efdddI mentioned above that is required to make this work for us. It allows to address multiple DB hosts in sequence.
- I also changed the debug output to suit my needs, then reverted it. You can re-add it by reverting the 'Revert' commit if needed.

html/inc/boinc_db.inc
```diff
@@ -95,7 +96,7 @@ class BoincDb extends DbConn {
     //
     static function get($dbnum = 0) {
         global $generating_xml;
-        if (isset(self::$instance)) {
+        if (isset(self::$instance) && intval(self::$dbnum) == intval($dbnum)) {
```
brevilo, line 97: The usual comment: changing `$dbnum` creates a new instance, leaving the old connection open until garbage collection kicks in. It would be better to actively close the connection as soon as it isn't needed anymore.

bema-aei: Yes, but it's the duty of the calling web code to decide how long a specific instance is needed. That's why I asked for a close function in #5691. We in our server status page need the second connection only for a single query, then go back to using the first instance. Should we only use one at. atime?

brevilo: Note, #5691 deals with the C-API. Here you should be able to use `self::$instance->db_conn->close()`, no?

> it's the duty of the calling web code to decide how long a specific instance is needed

I disagree in this case since you explicitly "clobber" the existing instance. Callers don't have a say in what's happening here. Unless...

> We do save what get() returns in a local variable and don't care about the global instance.

... this changes things. Ok.

davidpanderson:
What's the use case?  i.e. why would you need to access one replica, then another?

brevilo:
Different parts of the system use different replicas to spread the load and we're also showing the replication lag (of both replicas) on the SSP such that people can judge whether all data they is up to date.

bema-aei:
Actually the server status page where we show the status of all the replicas is the only single page where we need to access all of these on the same page. With a bit of internal caching it should be possible to make sure there's only one connection open at a time, but I didn't think about that initially.

PR #5697, opened 2024-07-19
PHP DB code: clean up the logic, and allow for > 1 readonly replica

davidpanderson:
Also always use mysqli interface

### DIFF
--- a/html/inc/boinc_db.inc
+++ b/html/inc/boinc_db.inc
@@ -105,7 +105,10 @@ class BoincDb {
     static function get($dbnum = 0) {
         global $generating_xml;
         if (isset(self::$instance)) {
-            return self::$instance;
+            if (self::$dbnum == $dbnum) {
+                return self::$instance;
+            }
+            close();
         }
         if (web_stopped()) {
             if ($generating_xml) {
--- a/html/inc/boinc_db.inc
+++ b/html/inc/boinc_db.inc
@@ -1,7 +1,7 @@
 <?php
 // This file is part of BOINC.
-// http://boinc.berkeley.edu
-// Copyright (C) 2008 University of California
+// https://boinc.berkeley.edu
+// Copyright (C) 2024 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -16,6 +16,19 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 
+// A project can have one or more BOINC databases:
+// DB 0:
+//      the main DB; read/write
+//      identified in config file by db_host, db_name, db_user, db_passwd
+//      db_host defaults to localhost
+// DB 1:
+//      read-only replica; identified by
+//      replica_db_host/name/user/passwd (must include all)
+// DB 2:
+//      read-only replica; identified by
+//      replica2_db_host/name/user/passwd (must include all)
+// ... and potentially more
+
 function incs() {
     $d = dirname(__FILE__);
     require_once("$d/db_conn.inc");
@@ -24,21 +37,15 @@ function incs() {
 
 incs();
 
-class BoincDb extends DbConn {
-    static $instance;
+// class BoincDb represents a connection to a BOINC database.
+// All its members are static, so there's only 1 connection at a time.
+// get(n) establishes a connection to DB n,
+// or DB 0 if that fails or doesn't exit.
+// close() closes the connection.
 
-    // A project can have one or more databases:
-    // DB 0:
-    //      the main DB; read/write
-    //      identified in config file by db_host, db_name, db_user, db_passwd
-    //      db_host defaults to localhost
-    // DB 1:
-    //      read-only replica; identified by
-    //      replica_db_host/name/user/passwd (must include all)
-    // DB 2:
-    //      read-only replica; identified by
-    //      replica2_db_host/name/user/passwd (must include all)
-    // ... and potentially more
+class BoincDb {
+    static $instance;       // a DbConn object, or null
+    static $dbnum;          // which replica we're connected to
 
     // connect to DB $dbnum (0, 1, ...)
     // If the requested DB doesn't exist or connection fails, connect to DB 0.
@@ -59,6 +66,7 @@ class BoincDb extends DbConn {
                 if ($retval) {
                     //error_log("BoincDb::get_aux(): connected to replica DB $dbnum");
                     self::$instance = $instance;
+                    self::$dbnum = $dbnum;
                     return;
                 }
             }
@@ -77,6 +85,7 @@ class BoincDb extends DbConn {
         if ($retval) {
             //error_log("BoincDb::get_aux(): connected to DB $dbnum");
             self::$instance = $instance;
+            self::$dbnum = 0;
             return;
         }
         error_log("BoincDb::get_aux(): Couldn't connect to DB $dbnum");
@@ -113,6 +122,13 @@ class BoincDb extends DbConn {
             xml_error(-138, "Can't connect to database");
         } else {
             error_page("Can't connect to database");
+        }
+    }
+
+    static function close() {
+        if (isset(self::$instance)) {
+            self::$instance->close();
+            self::$instance = null;
         }
     }
 
--- a/html/inc/boinc_db.inc
+++ b/html/inc/boinc_db.inc
@@ -57,31 +57,27 @@ class BoincDb extends DbConn {
             if ($host && $name && $user && $passwd) {
                 $retval = $instance->init_conn($user, $passwd, $host, $name);
                 if ($retval) {
-                    if ($instance->do_query("use $name")) {
-                        //error_log("BoincDb::get_aux(): connected to replica DB $dbnum");
-                        self::$instance = $instance;
-                        return;
-                    }
+                    //error_log("BoincDb::get_aux(): connected to replica DB $dbnum");
+                    self::$instance = $instance;
+                    return;
                 }
             }
             // if can't connect to replica, fall through and try DB 0
         }
         $host = parse_config($config, '<db_host>');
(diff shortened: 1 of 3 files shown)


brevilo:
> Also always use mysqli interface

Please, use atomic commits so that one can review separate topics separately.

html/inc/boinc_db.inc
```diff
@@ -27,119 +27,97 @@ incs();
...
+        $passwd = parse_config($config, '<db_passwd>');
+        // OK if host is null; it will use localhost
+        if (!$name || !$user || !$passwd) {
+            error_log("BoincDb::get_aux(): must specify DB name, user, passwd");
+            return;
+        }
+        $retval = $instance->init_conn($user, $passwd, $host, $name);
+        if ($retval) {
+            if ($instance->do_query("use $name")) {
+                //error_log("BoincDb::get_aux(): connected to DB $dbnum");
```
brevilo, line 81: Can be removed.

html/inc/boinc_db.inc
```diff
@@ -27,119 +27,97 @@ incs();
...
+            $passwd = parse_config($config, sprintf('<replica%s_db_passwd>', $r));
+            if ($host && $name && $user && $passwd) {
+                $retval = $instance->init_conn($user, $passwd, $host, $name);
                 if ($retval) {
-                    // needed for places where we do direct queries
-                    if (!$instance->do_query("use $name")) {
-                        error_log("BoincDb::get_aux(): Couldn't select database $name on $replica_host (giving up)");
-                        $instance = null;
+                    if ($instance->do_query("use $name")) {
+                        //error_log("BoincDb::get_aux(): connected to replica DB $dbnum");
```
brevilo, line 61: Can be removed.

html/inc/boinc_db.inc
```diff
@@ -27,119 +27,97 @@ incs();
...
-        } else {
-            // needed for places where we do direct queries
-            if (!$instance->do_query("use $name")) {
-                error_log("BoincDb::get_aux(): Couldn't select database $name on $host (giving up)");
-                $instance = null;
+        $host = parse_config($config, '<db_host>');
+        $user = parse_config($config, '<db_user>');
+        $name = parse_config($config, '<db_name>');
+        $passwd = parse_config($config, '<db_passwd>');
+        // OK if host is null; it will use localhost
```
brevilo, line 73: How? I don't see it being set explicitly. Do you depend on `mysqli` to do that? If so, it [falls back](https://www.php.net/manual/en/mysqli.construct.php) to use [`mysqli.default_host`](https://www.php.net/manual/en/mysqli.configuration.php#ini.mysqli.default-host) which doesn't have a [default value](https://www.php.net/manual/en/mysqli.configuration.php) configured, and I can't find one explicitly defined here either.

brevilo:
Apart from the above I like the approach. Flexible while not overly complex.

davidpanderson:
I removed the 'use db_name' query (not needed with mysqli)
and made the default 'localhost' explicit.

html/inc/boinc_db.inc
brevilo: Thanks

html/inc/boinc_db.inc
```diff
@@ -57,31 +57,27 @@ class BoincDb extends DbConn {
             if ($host && $name && $user && $passwd) {
                 $retval = $instance->init_conn($user, $passwd, $host, $name);
                 if ($retval) {
-                    if ($instance->do_query("use $name")) {
-                        //error_log("BoincDb::get_aux(): connected to replica DB $dbnum");
-                        self::$instance = $instance;
-                        return;
-                    }
+                    //error_log("BoincDb::get_aux(): connected to replica DB $dbnum");
```
brevilo, line 60: Please remove dead code like this. You can add it back for debugging if needed. Alternatively, use pre-processor macros.

html/inc/boinc_db.inc
```diff
@@ -57,31 +57,27 @@ class BoincDb extends DbConn {
...
             return;
         }
         $retval = $instance->init_conn($user, $passwd, $host, $name);
         if ($retval) {
-            if ($instance->do_query("use $name")) {
-                //error_log("BoincDb::get_aux(): connected to DB $dbnum");
-                self::$instance = $instance;
-                return;
-            }
+            //error_log("BoincDb::get_aux(): connected to DB $dbnum");
```
brevilo, line 78: Please remove dead code like this. You can add it back for debugging if needed. Alternatively, use pre-processor macros.

brevilo:
@bema-aei should ultimately review this as well.

html/inc/boinc_db.inc
```diff
@@ -27,119 +27,93 @@ incs();
...
-                }
-            }
-            self::get_aux($readonly, $fallback_mode);
-            if (!self::$instance) {
-                if ($generating_xml) {
-                    xml_error(-138, "Can't connect to database");
-                } else {
-                    error_page("Can't connect to database");
-                }
+        if (isset(self::$instance)) {
```
brevilo, line 98: Hm, this implies that there can only be a single instance at a time. What we need here is a hash of instances (indexed by `$dbnum`) such that multiple instances are retained and/or can be switched between. E.g. the server status page needs to query different databases (here: replicas).

bema-aei: I think a hash of instances might be a bit of an overkill, I think E@H is the only project that uses as much as two. However a requirement for us would be to access different replicas at least in sequence. I did this in d1f99043877552ea48ac2912dc4ec9ac9a1efddd

bema-aei:
This works fro us in principle, with the change I mentioned above. I updated my original branch with your changes, with two important changes:
- including a change that affects the whole web code ("use mysqli everywhere") in another commit with very limited scope and not even mentioning it in the commit message is just asking for trouble. It's difficult to review, difficult to test. Worst of all,  if someone of some project later finds it to cause trouble for him, he is unlikely to ever find the commit that breaks his project. So I split your commit in two (leaving the author intact).
- I added the commit I mentioned above that is required to make this work for us. It allows to address multiple DB hosts in sequence without much effort.
- I also changed the debug output to suit my needs, then reverted it. You can re-add it by reverting the 'Revert' commit if needed.
Please have a look at that branch.

html/inc/boinc_db.inc
brevilo: Ok, but please add a comment that states that there's only ever be one current DB instance at a time. IOW, the current instance gets created, reused or replaced by a new one.

html/inc/boinc_db.inc
brevilo: Fixed by @bema-aei

html/inc/boinc_db.inc
brevilo: Fixed by @bema-aei

html/inc/boinc_db.inc
bema-aei: Is this a requirement?

brevilo: Well no, but that's how you implemented it, isn't it?

brevilo: Resolved [here](https://github.com/BOINC/boinc/pull/5695#discussion_r1698695506)

AenBleidd:
@bema-aei, @brevilo, are you ok with this PR? Does it need anything else (like testing on your side) or we can move forward and merge it?

AenBleidd:
@bema-aei, @brevilo, could you please take a look at this PR and test it?
Thank you in advance.

bema-aei:
First of all I like the idea of allowing an arbitrary number of replicas. And thank you for implementing a close() function.

[This code](https://github.com/BOINC/boinc/blob/4b4ba9dbf83384f272f083875cb7f187bc6f6f67/html/inc/boinc_db.inc#L107-L109) in get() is still unchanged, i.e. if $instance is set, get() will return the same instance even if $dbnum is different.

At least this doesn't correspond to what I proposed in my branch. Is the "user" of this code required to close one connection / instance before opening another? If so, it should be noted somewhere, I didn't expect that.

bema-aei:
Also too I there is an uncaught exception if the DB connection is refused for some reason. I get:
```
PHP Fatal error:  Uncaught mysqli_sql_exception: Connection refused in /srv/BOINC/live-webcode-test/html/inc/db_conn.inc:37
Stack trace:
#0 /srv/BOINC/live-webcode-test/html/inc/db_conn.inc(37): mysqli->__construct()
#1 /srv/BOINC/live-webcode-test/html/inc/boinc_db.inc(65): DbConn->init_conn()
#2 /srv/BOINC/live-webcode-test/html/inc/boinc_db.inc(117): BoincDb::get_aux()
#3 /srv/BOINC/live-webcode-test/html/ops/eah_server_status.php(681): BoincDb::get()
#4 /srv/BOINC/live-webcode-test/html/ops/eah_server_status.php(1108): seconds_behind_master()
#5 {main}
  thrown in /srv/BOINC/live-webcode-test/html/inc/db_conn.inc on line 37
```

davidpanderson:
- I added the check for different dbnum in get()
- Yes, you can only use on DB at a time
- An exception would be thrown if had previously called
  mysqli_report(MYSQLI_REPORT_STRICT);
  presumably because you want to handle exceptions.
  But in any case I added a try/catch.

bema-aei:
Thanks, this works for us technically. We do use a recent MariaDB, which has some defaults different from MySQL, hence the exception.

However, for reasons I pointed out already I can't really approve the first commit which combines different features w/o even mentioning in the commit message, so I can't recommend to merge this branch as it is.

brevilo:
> We do use a recent MariaDB, which has some defaults different from MySQL, hence the exception.

How does a PHP module's config relate to the MySQL flavor used as the DB backend?

bema-aei:
> > We do use a recent MariaDB, which has some defaults different from MySQL, hence the exception.
> 
> How does a PHP module's (`mysqli`) config relate to the MySQL flavor used as the DB backend?

Oh. Sorry, was mislead. Indeed, that's the mysqli config.

PR #5704, opened 2024-07-23
client: fix minor prefs parsing bug

### LINKED ISSUE #5682: Relax XML parser to handle global preferences gracefully
The global preferences XML gets processed by and transferred between all clients, projects and account managers used by a volunteer. Any part of this diverse software ecosystem could have introduced certain features into the XML structure over time that the current XML parser doesn't expect, resulting in broken parsing. This has been a known and recurring issue for years and previous attempts to fix the situation have apparently been incomplete. Thus we should relax the BOINC XML parser(s) as much as possible such that it handles incoming XML as gracefully as possible.

For instance, [this line](https://github.com/BOINC/boinc/blob/6baf185406e698cf8146fe3d9dcc448cefb3d758/lib/prefs.cpp#L414) implies that any elements following the (current) closing `venue` element aren't parsed. But the reality is that some XML fragments have a different order, for whatever reason. These fragments are technically valid XML and they also contain valid preferences sets, default and one or more venues - just that the venues might sometimes be listed first, or the `mod_time` follows the venue elements. Please relax the parser to make it more robust for cases like these. For this case I suggest replacing the `return 0` by a `continue` to just keep processing.

I haven't put this into a PR since I'm not familiar enough with this part of the codebase and maybe there are other places that need to be taken into account as well. I hope the general idea came across, though.

Thanks

davidpanderson:
If we parse for a target venue, and it's present,
and mod_time/source_project/source_scheduler appear after the venue, we were failing to parse them.

Fixes #5682

### DIFF
--- a/lib/prefs.cpp
+++ b/lib/prefs.cpp
@@ -408,15 +408,28 @@ int GLOBAL_PREFS::parse_override(
             }
             return 0;
         }
+        // parse these first; they're independent of venue
+        //
+        if (xp.parse_str("source_project", source_project, sizeof(source_project))) {
+            continue;
+        }
+        if (xp.parse_str("source_scheduler", source_scheduler, sizeof(source_scheduler))) {
+            continue;
+        }
+        if (xp.parse_double("mod_time", mod_time)) {
+            double now = dtime();
+            if (mod_time > now) {
+                mod_time = now;
+            }
+            continue;
+        }
         if (in_venue) {
             if (xp.match_tag("/venue")) {
-                if (in_correct_venue) {
-                    return 0;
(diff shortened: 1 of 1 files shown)


bema-aei:
@davidpanderson Do I understand your fix correctly that whenever there is some global preference property (and tag) added that doesn't appear in the venues, there has to be code added here to treat that specifically (as it's currently done with source_project, source_scheduler, mod_time now)? I really don't know how likely such an addition is, though.

davidpanderson:
All items need code to parse them.  It's a question of where the code goes.  See my patch.

PR #5707, opened 2024-07-24
Server: sched/validator: DON'T set assimilate_state by the validator.

### LINKED ISSUE #5706: validator - assimilator race condition - should validator set assimilate_state?
Occasionally Einstein@Home encounters the following situation:
- quorum of a workunit is completed, the workunit is validated, a canonical result is found and the validator sets assimilate_state =1
- while the assimilator is working on the workunit, another result is reported by the scheduler, need_validate is set  and the validator reads the workunit record
- the assimilator finishes assimilation and sets assimilate_state=2. file deletion is triggered
- the validator finishes validation and sets assimilate_state=1, overriding the value the assimilator just set
- the assimilator tries to assimilate the same workunit again.
Depending on whether the file deleter has already finished, the assimilator either tries to assimilate the same canonical result files a second time or it doesn't find these at all. Both conditions cause our assimilator to halt (for good reason), and require manual intervention to analyze and resolve the situation.

Also, when setting assimilate_state, the validator doesn't care about results of the workunit that are still "in progress". Triggering assimilation (and subsequently file deletion) of a workunit immediately after finding a canonical result means that these "late" results can not be (successfully) validated, assimilated and credited even if these are valid and arrive on time, i.e. before their deadline.

I was about to modify the validator to not set assimilate_state directly, but just let the transitioner know to take a look at that workunit (immediately). It does handle such late results correctly and, and the transitioner doesn't have such a delay between reading the workuits and writing modifications as the validator has (at least occasionally).

But then I came across that comment in the validator:

https://github.com/BOINC/boinc/blob/master/sched/validator.cpp#L677-L679
```
                // if we found a canonical result,
                // trigger the assimilator, but do NOT trigger
                // the transitioner - doing so creates a race condition
```

What race condition is this comment referring to, and how would we resolve the problems above then?

bema-aei:
- The assimilator may have already started to assimilate this WU. The validator may take quite some time to chew through a chunk of workunits. Resetting the assimilate_state after that may be harmful and override the update made by the assimilator. Instead, let the transitioner decide what to do with this workunit.

- Also too, avoid overriding a non-zero error mask that might have been set in the meantime by some other daemon

Fixes #5706

### DIFF
--- a/db/boinc_db.cpp
+++ b/db/boinc_db.cpp
@@ -2032,8 +2032,9 @@ int DB_VALIDATOR_ITEM_SET::update_workunit(WORKUNIT& wu) {
     char query[MAX_QUERY_LEN];
 
     sprintf(query,
-        "update workunit set need_validate=0, error_mask=%d, "
-        "assimilate_state=%d, transition_time=%d, "
+        "update workunit set need_validate=0, "
+        "error_mask=error_mask|%d, "
+        "transition_time=%d, "
         "target_nresults=%d, "
         "canonical_resultid=%lu, canonical_credit=%.15e "
         "where id=%lu",
--- a/sched/validator.cpp
+++ b/sched/validator.cpp
@@ -685,7 +685,11 @@ int handle_wu(
                 );
                 wu.canonical_resultid = canonicalid;
                 wu.canonical_credit = credit;
-                wu.assimilate_state = ASSIMILATE_READY;
+
+                // DON'T modify assimilate_state here, under bad conditions the
+                // assimilator may have already started to work on this WU.
+                // Instead, let the transitioner decide what to do with it.
+                transition_time = IMMEDIATE;
 
                 // don't need to send any more results
                 //


brevilo:
Crappy GitHub doesn't let me comment outside a hunk's context. I presume [this comment](https://github.com/BOINC/boinc/pull/5707/files#diff-7ea9559350700313814aead405e39b324dd3cad697733e8d5abbd1bfb83e9c6cR677) need to be removed as they contradict the new logic (and comment), I think.

bema-aei:
Right. And I think [this loop](https://github.com/BOINC/boinc/blob/bed25a896e7975550376bca41ba54f3841a3737f/sched/validator.cpp#L696-L716) should be deleted as well, as this should be done by the transitioner then anyway.

bema-aei:
I'm afraid the current transitioner relies on the validator setting assimilate_state as a sign of finished validation, at least under certain conditions. I need to think about this a little more.

PR #5745, opened 2024-08-09
clarify meaning of DEBUG in C++ code

davidpanderson:
I was testing boinc_cmd and was surprised to see that it was
printing the GUI RPC XML messages.
Turned out this was because of a puts(req) conditioned on DEBUG,
which is set (globally) when you configure with --enable-debug.
I looked around and there were several other cases of
very specific debugging printf's enabled by DEBUG.

This isn't the right idea.
A global flag should do global things (like compile with -g).
Use specific #defines (like SHOW_MSGS for GUI RPCs).

### DIFF
--- a/api/ttfont.cpp
+++ b/api/ttfont.cpp
@@ -31,6 +31,8 @@
 
 // originally adapted by Carl Christensen
 
+//#define TTFONT_DEBUG
+
 #ifdef _WIN32
 #include "boinc_win.h"
 #endif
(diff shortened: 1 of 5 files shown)


AenBleidd:
This is quite a controversial topic. From one hand I agree with you, from the other hand, the purpose of `DEBUG` macro is to either enable debug-specific behavior (like additional output) or to add some other compiler-specific parameters.
I tend to reject this PR and keep the current behavior.
@davidpanderson, let's discuss this separately if you don't agree with me, and try to find the best suitable solution.

PR #5747, opened 2024-08-11
[Windows] Detect WSL version of enabled distros.

AenBleidd:
In the most cases, WSL2 (currently the latest) is the recommended version, but there are some cases when WSL1 is more suitable. For example, WSL1 is faster in some cases (e.g. file operations), and it is the only option for Windows 10 Home users. More details here: https://docs.microsoft.com/en-us/windows/wsl/compare-versions

### DIFF
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
-// http://boinc.berkeley.edu
-// Copyright (C) 2022 University of California
+// https://boinc.berkeley.edu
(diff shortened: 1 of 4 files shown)


AenBleidd:

davidpanderson:
struct WSL has fields 'name' and 'version'.
Please change them to say what they refer to
(i.e. name and version of what)

AenBleidd:
@davidpanderson, done.
Please re-review.

PR #5754, opened 2024-08-13
[sched] add wsl plan class

AenBleidd:


AenBleidd:
@davidpanderson, please review. This is something that was documented (https://github.com/BOINC/boinc/wiki/AppPlan) but never properly implemented.
I haven't tested it, so please pay additional attention to something I might missed.
Thank you in advance.

davidpanderson:
I think this is premature.  An app that uses WSL is going to need to specify
an distro (name/version) and possibly whether it uses WSL 1 or 2.
Let's add this when we have a complete example.

AenBleidd:
Will just leave it for now then.

AenBleidd:
@davidpanderson, I was thinking a little bit about this.
I do believe we can go iterative approach there, and add required functionality when really needed.
Speaking about version of the WSL (1 or 2) and the distro to run - I don't think this will be a real request from the projects, and we can think here about WSL as just a Linux: for linux we don't actually have a lot of requirements to be me to run project application (bitness, cpu features, etc), but currently the amount of WSL images available in the Microsoft Store (we are not talking about custom WSL images since this is a very rare case) is not so big.
Let's talk about this on the next call and decide whether we want to proceed with that or keep it on hold for a while.

PR #5755, opened 2024-08-13
address issue with max concurrent and work fetch

### LINKED ISSUE #5743: GPU task switching causes computation errors for asteroids@home when using 2 or more different models of gpu of the same type
**Describe the bug**
If GPU computation is suspended during use or when an exclusive application is running, when computation resumes, BOINC sometimes swaps which task is on which GPU. This causes a computation error for asteroids@home tasks when using multiple AMD GPUs, for example, an RX 7600 XT and an RX 6600. This might be an application specific issue, but it might be a good idea to have an option to not switch tasks between GPUs if possible, unless, for example, one GPU is removed, in which case all the tasks would have to run on the remaining GPU.

**Steps To Reproduce**
1. Start the asteroids@home period search application on a system with 2 different AMD gpus
2. Suspend computation partway through the computation, observing which task is on which gpu, the resume computation
3. Repeat if necessary until BOINC switches tasks between gpus resulting in a computation error.

**Expected behavior**
I would expect the task to stay on the GPU it started on if that is necessary for the task to finish. An option to disable gpu task switching is a potential solution, or tasks could specify weather or not they can be switched.

**Screenshots**

**System Information**
 - OS: Windows 10 (Latest)
 - BOINC Version: 8.0.4

**Additional context**

### LINKED ISSUE #5749: Work fetch stops before per-app concurrency limit is reached even when there are CPUs are idling
**Describe the bug**
The host has 8C/16T, currently with 4 threads busy running 2 ATLAS jobs but remaining 12 idling. Preference is set to use 100% CPUs. ATLAS app from LHC@Home is set to 4 concurrent jobs with 2 CPUs each in app_config. All other projects are set to request no more work. LHC will only provide work for ATLAS app based on online preference. 

While ATLAS alone won't be able to fill all cores, I expect the work fetch logic to get at least 4 tasks to max out the concurrency limit, occupying 8 threads. However, fetch consistently stops at 2 ATLAS WUs with no other work running. 

Only when I increase the `work_buf_min_days` to be beyond the estimated finish time of each ATLAS WU, fetching resumes and the new WUs can start right away, maxing out the concurrency limit.

**Steps To Reproduce**
1. Apply app_config.xml for LHC as shown in [app_config.txt](https://github.com/user-attachments/files/16571576/app_config.txt)
2. Set all other projects except LHC@Home to request no new work.
3. Set `work_buf_min_days` to be below estimated time of an ATLAS WU, like `0.1`. Set `work_buf_additional_days` to `0`. Set `max_ncpus_pct` to `100`. 
4. The system would start fetching the work until half of the concurrency limit is hit, instead of the full concurrent limit. Manual update won't trigger any work fetch either.

Log with work fetch flag enabled is attached: 
[lhc_fetch.txt](https://github.com/user-attachments/files/16571594/lhc_fetch.txt)
Relevant parts (based on my guess) below. 
```
[---] [work_fetch] shortfall 103680.00 nidle 12.00 saturated 0.00 busy 0.00
[LHC@home] [work_fetch] REC 1097.107 prio -1.009 can request work
[LHC@home] [work_fetch] share 1.000
[LHC@home] [work_fetch] using MC shortfall 0.000000 instead of shortfall 103680.000000
[LHC@home] Not requesting tasks: don't need (CPU: ; AMD/ATI GPU: )
```

I've also uploaded my state, prefs, cc_config files to the simulator [scenario 210
](https://boinc.berkeley.edu/dev/sim_web.php?action=show_scenario&name=210). Simulation 0 also failed to fetch work, though the log looks a bit different. 

**Expected behavior**
Given the concurrent limit is 4 while there are only 2 WUs running, I expect the client to fetch at least another 2 WUs regardless what `work_buf_min_days` is. If I remove the concurrency limit with no other changes to preference, the client would proceed to fetch enough work to fill all cores, which is expected behavior. 

**Screenshots**

**System Information**
 - OS: Ubuntu 24.04
 - BOINC Version: 7.24.1 (as packaged by Ubuntu)

**Additional context**
I was able to reproduce this on a different 16C/32T host by setting ATLAS currency limit to 8 with 2 CPUs per WU as well. Fetch stops at 4 WUs on that machine, instead of 8.

davidpanderson:
Fixes #5749 (hopefully)

Max concurrent is a limit on jobs, not processor instances.
The work fetch logic made the erroneous implicit assumption
that all jobs use 1 CPU.
So e.g. if project has max concurrent 4,
and the client has two 2-CPU jobs,
it will think (if work buf is zero) that there's
no point in fetching more work.
But in fact the project could use 8 CPUs, so 4 are idle.

Fix: if a project has MC constraints,
then for each resource compute 'mc_max_could_use':
the max # of instances the project could use, given its MC constraints.
Use this to compute the project's shortfall,
and hence to decide whether to fetch work from it.

Note: the way mc_max_could_use is computed is crude;
it takes the max over all apps,
when it's possible that only one of them has a MC constraint.
This could result in limited over-fetching,
but that's preferable to under-fetching and starvation.

Sim: show app name in timeline

### DIFF
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -2118,7 +2118,7 @@ int CLIENT_STATE::reset_project(PROJECT* project, bool detaching) {
     project->min_rpc_time = 0;
     project->pwf.reset(project);
     for (int j=0; j<coprocs.n_rsc; j++) {
-        project->rsc_pwf[j].reset();
+        project->rsc_pwf[j].reset(j);
     }
     write_state_file();
     return 0;
--- a/client/makefile_sim
+++ b/client/makefile_sim
@@ -1,5 +1,8 @@
 # makefile for client simulator
 # Do "make_clean" in client/, lib/, and sched/ first
+#
+# this doesn't have .h dependencies; if you change something,
+# do make clean and make
 
 CXXFLAGS = -g -DSIM -Wall \
     -I ../lib \
--- a/client/project.h
+++ b/client/project.h
@@ -276,9 +276,9 @@ struct PROJECT : PROJ_AM {
     //
     RSC_PROJECT_WORK_FETCH rsc_pwf[MAX_RSC];
     PROJECT_WORK_FETCH pwf;
-    inline void reset() {
+    inline void work_fetch_reset() {
         for (int i=0; i<coprocs.n_rsc; i++) {
-            rsc_pwf[i].reset();
+            rsc_pwf[i].reset(i);
         }
     }
     inline int deadlines_missed(int rsc_type) {
--- a/client/rr_sim.cpp
+++ b/client/rr_sim.cpp
@@ -111,12 +111,6 @@ struct RR_SIM {
         }
         if (have_max_concurrent) {
             max_concurrent_inc(rp);
-            if (p->rsc_pwf[0].sim_nused > p->rsc_pwf[0].max_nused) {
-                p->rsc_pwf[0].max_nused = p->rsc_pwf[0].sim_nused;
-            }
-            if (rt && p->rsc_pwf[rt].sim_nused > p->rsc_pwf[rt].max_nused) {
-                p->rsc_pwf[rt].max_nused = p->rsc_pwf[rt].sim_nused;
-            }
         }
     }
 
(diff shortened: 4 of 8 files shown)


AenBleidd (approved):

AenBleidd:
@davidpanderson, are you sure this is the fix for #5743, not for #5749?

davidpanderson:
Oops!  Fixed.

Issue #5786, opened 2024-08-31
Linux Manager: Font used in 'Computing Preferences' is too large


RichardHaselgrove:
**Describe the bug**
The text in the computing preferences dialog (bold, italic) is too large, and overspills the available space. This is the only page where this font is used.

**Screenshots**
![image](https://github.com/user-attachments/assets/7ccb8a7f-21f8-4ba2-aaaa-da373e69c739)

**System Information**
 - OS: Linux Mint 22
 - BOINC Version: 8.0.4

**Additional context**
Again, this does not apply to Windows. Please check for consequential changes in other OSs when fixing.

PR #5790, opened 2024-09-03
[linux][manager] Fix fonts on Advanced Computing Preferences dialog.

AenBleidd:
Fixes: #5786.


RichardHaselgrove:
Thanks - that looks much better (downloaded artifact)

PR #5797, opened 2024-09-06
client: get 'in use' state only if computing preferences need it.

davidpanderson:


AenBleidd:
@davidpanderson, please fix MacOS build:
```
Compiling client_state.cpp

❌  /Users/runner/work/boinc/boinc/client/client_state.cpp:1059:36: use of undeclared identifier 'idle_time'

    if (started_by_screensaver && (idle_time < 30) && (getppid() == 1)) {
                                   ^                                
```

AenBleidd (approved):

PR #5820, opened 2024-09-29
Update is_extpack_installed

computezrmle:
For many years and across a couple of major VirtualBox versions vboxwrapper was looking for "Oracle VM VirtualBox Extension Pack" to test whether the Extension Pack is present.
With VirtualBox version 7.1 Oracle changed the string to "Oracle VirtualBox Extension Pack".

Reported by an experienced volunteer at LHC@home.

To make vboxwrapper work with old as well as with recent VirtualBox versions the test string needs to be modified.

### DIFF
--- a/samples/vboxwrapper/vbox_vboxmanage.cpp
+++ b/samples/vboxwrapper/vbox_vboxmanage.cpp
@@ -1770,7 +1770,7 @@ bool VBOX_VM::is_extpack_installed() {
     command = "list extpacks";
 
     if (vbm_popen(command, output, "extpack detection", false, false) == 0) {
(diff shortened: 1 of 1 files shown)


AenBleidd (approved):

Toby-Broom:
Would it not be better to just look for VBoxVRDP as this didn't change? and allows back compatabilty from 7.0.x to 7.1.x

computezrmle:
@Toby-Broom
Thanks for the comment.
Tests have already been done with 7.0.x and 7.1.x.
They didn't show any issue.

Can you live with it if the code remains as is?

Toby-Broom:
sure no problem, thanks for making the update.

PR #5840, opened 2024-10-07
lib/Makefile.mingw: add boinc_stdio.h to install headers

bema-aei:
boinc_stdio.h should be installed as well when doing a 'make -f lib/Makefile.mingw install'


AenBleidd (approved):

AenBleidd:
@bema-aei, thank you for the fix!

PR #5845, opened 2024-10-11
Improved input handling

lfield:
Fixes #

**Description of the Change**

**Alternate Designs**

**Release Notes**

PR #5846, opened 2024-10-11
web: don't allow special chars in user names.

davidpanderson:
Check this (and show appropriate message) in both account creation and user name update


AenBleidd:
If I understand this correctly, this is a replacement for #5845.
@lfield, please take a look

lfield:
LGTM

PR #5850, opened 2024-10-18
Remove unused functions

Vulpine05:
**Description of the Change**
Browsing through the code, I have found several functions defined in header files that Visual Studio found did not have an associated function nor inline code.  Since these functions do not exist, this PR removes them.

I have broken up the PR into three commits; each commit is for a subfolder (client, manager, and lib).

This PR does not clean up ALL unused functions.  There were some functions that were empty constructors that I didn't feel comfortable removing.  Another thing I discovered were some functions were noted by Visual Studio as not defined, but the function was written in one or more preprocessor definitions.  With that said, I tried to carefully remove only functions that were obvious to me that weren't defined.  Thank you in advance for reviewing!

### DIFF
--- a/client/app.h
+++ b/client/app.h
@@ -227,7 +227,6 @@ struct ACTIVE_TASK {
         // This is compared with project-specified limits
         // to decide whether to abort job; no other use.
     int get_free_slot(RESULT*);
-    int start(bool test=false);         // start a process
 
     // Termination stuff.
     // Terminology:
@@ -339,7 +338,6 @@ public:
     void get_msgs();
     bool check_app_exited();
     bool check_rsc_limits_exceeded();
-    bool check_quit_timeout_exceeded();
     bool is_slot_in_use(int);
     bool is_slot_dir_in_use(char*);
     void send_heartbeats();
--- a/client/app.h
+++ b/client/app.h
@@ -339,7 +339,6 @@ public:
     void get_msgs();
     bool check_app_exited();
     bool check_rsc_limits_exceeded();
-    bool check_quit_timeout_exceeded();
     bool is_slot_in_use(int);
     bool is_slot_dir_in_use(char*);
     void send_heartbeats();
--- a/client/app.h
+++ b/client/app.h
@@ -227,6 +227,7 @@ struct ACTIVE_TASK {
         // This is compared with project-specified limits
         // to decide whether to abort job; no other use.
     int get_free_slot(RESULT*);
+    int start(bool test=false);         // start a process
 
     // Termination stuff.
     // Terminology:
@@ -338,6 +339,7 @@ public:
     void get_msgs();
     bool check_app_exited();
     bool check_rsc_limits_exceeded();
+    bool check_quit_timeout_exceeded();
     bool is_slot_in_use(int);
     bool is_slot_dir_in_use(char*);
     void send_heartbeats();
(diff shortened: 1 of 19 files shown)


Vulpine05:
Oops, I don't see how I missed that Start function, but it makes sense now.  I'll fix that when I'm able to.

CharlieFenton:
Please do not remove the following line from _MainDocument.h_:
`//    void                      TestAsyncRPC();      // For testing Async RPCs`
The comment clearly states that this line has been left in the code but disabled with `//` to allow easily re-enabling it for additional testing when needed. The corresponding function is in _AsyncRPC.cpp_ and disabled by `#if 0` and `#endif`, allowing for easy re-enabling when needed.

> Visual Studio found did not have an associated function nor inline code.

Since this line is disabled with `//`, I don't see how the compiler would mark it as a declaration with no corresponding function.

Also, please use caution when removing declarations, because some functions are used only by builds for certain operations systems (Windows, MacOS, Linux, Android, etc.) Ideally both the header declaration and function definition will be guarded by something like `#ifdef _WIN32` or `#ifdef __WXMAC__`, but there might be instances where that guard is missing from the header declaration. That situation would also make the compiler think there is a header declaration with no associated function. Hopefully, the CI build will fail if you remove a declaration needed when building for a different operating system.

Vulpine05:
> Please do not remove the following line from _MainDocument.h_: `// void TestAsyncRPC(); // For testing Async RPCs` The comment clearly states that this line has been left in the code but disabled with `//` to allow easily re-enabling it for additional testing when needed. The corresponding function is in _AsyncRPC.cpp_ and disabled by `#if 0` and `#endif`, allowing for easy re-enabling when needed.

Understood, apologies that I missed that the first time around.

> Since this line is disabled with `//`, I don't see how the compiler would mark it as a declaration with no corresponding function.

It didn't find it that way, I just thought it was commented out for not being used, but I was too quick with the delete key for that one.  This was an exception to how I was targeting unused functions.  Again, apologies.

> Also, please use caution when removing declarations, because some functions are used only by builds for certain operations systems (Windows, MacOS, Linux, Android, etc.) Ideally both the header declaration and function definition will be guarded by something like `#ifdef _WIN32` or `#ifdef __WXMAC__`, but there might be instances where that guard is missing from the header declaration. That situation would also make the compiler think there is a header declaration with no associated function. Hopefully, the CI build will fail if you remove a declaration needed when building for a different operating system.

Yes, I did see some of those, and for the most part I left them alone because I didn't want to break something I didn't fully understand.  For all other functions I did painstakingly search through the entire code to see if the function was used.  If I didn't find it anywhere else, I removed it.  There are some instances where a function was marked as not used, but something else was happening (different input variables, or preprocessor declarations for example), and I left those alone since it wasn't just removing a declaration.

The only declaration I see right now I have removed is part of a preprocessor is here:

https://github.com/BOINC/boinc/blob/be884908dc15de5c3d75739f9174548a18c7fc5f/client/project.h#L390

Vulpine05:
Okay, take two.  I think this is ready for review now.

AenBleidd (approved):

AenBleidd:
Ok, it was not easy.
Looks good to me.

Vulpine05:
Thank you for reviewing, @AenBleidd.  I'm sure it was harder than it seemed.

PR #5855, opened 2024-10-24
Add docker_wrapper and associated client/server changes

### LINKED ISSUE #5825: Support apps that run in Docker (or other) containers
Design notes on proposed Docker features.
Related issues: #5617, #5512

The basic idea is to make a cross-platform "Docker wrapper" that interfaces the BOINC client to the Docker engine.
The wrapper uses the docker CLI program (e.g. docker run foo) to create and control containers.

An app version would consist of:
* the wrapper
* a Docker file, which defines the image (which may use images from public or private registries)
* Additional files, e.g. executable

Issues:

## Input/output files

Goal: no copying of large files.
To do this: the wrapper mounts the slot and project dirs in the container.
The main script uses resolve.sh to map logical to physical names.

## Suspend/resume

Use 'docker run --stop name'

## CPU throttling

Does Docker have a feature for this?  Don't want to use BOINC's stop/start approach.

## Checkpointing

Can we use 'docker commit'?

## Fraction done

Do this using a file in slot dir;  app writes it.

## Cleaning up old images

When and how?

## Getting stderr output

Use a file in slot dir?  or does Docker have a feature for this?

## Finish file and exit code

How does wrapper know if app really finished?

## Getting CPU time and working set size

Can the wrapper run ps in container, like WSL wrapper?

## Windows issues

What WSL image should the wrapper use?

## Other container systems

Docker is free only for organizations with < 250 employees and  $10M revenue.
These parameters can change at any time.
Maybe we should design things so that we could also use
[Singulariy](https://sylabs.github.io/singularity101/introduction.html)
or [Podman](https://podman.io/).

davidpanderson:
Support app versions that run jobs in Docker containers.
This also works with podman, an open-source and daemonless Docker equivalent.

Also: some changes to the client in terms of how Docker and podman are detected,
and the corresponding data structures.

Also: minor changes to the scheduler for parsing this info
(to be used by plan class specs).

Fixes #5825

### DIFF
--- a/client/Makefile.linux
+++ b/client/Makefile.linux
@@ -140,7 +140,7 @@ clean:
 LIBS = ../lib/boinc.a \
     -L /usr/local/lib/ \
     -lpthread \
-    -lX11 -lXss \
+    -lX11 \
     -lcurl -lssl -lcrypto \
     -lz -ldl
 
--- a/client/app_test.cpp
+++ b/client/app_test.cpp
@@ -43,9 +43,11 @@
 
 #include "client_state.h"
 
-// set to 0 to enable app test
+#define APP_NONE            0
+#define APP_WSL_WRAPPER     0
+#define APP_DOCKER_WRAPPER  1
 
-#if 1
+#if APP_NONE
 void CLIENT_STATE::app_test_init() {}
 #else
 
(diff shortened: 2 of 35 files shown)


AenBleidd:
@davidpanderson, please fix build error: https://github.com/BOINC/boinc/actions/runs/11507373201/job/32033412490?pr=5855
And also afterwards please fix the formatting error by running next command while being in the root folder of the BOINC source code:
```
```

davidpanderson:
Done

PR #5862, opened 2024-10-28
[Manager] Gray out projects in project wizard that are already attached

### LINKED ISSUE #1552: Hide or gray out projects already added from the BOINC Manager Add Project wizard
Is it possible to add to the Add Project wizard that it hides, or grayes out those projects that are already added.

Would that be easy to add?

Vulpine05:
Fixes #1552 

**Description of the Change**
This PR adds a feature to the Add Project wizard that will change the text of a project to gray if the project is already added.  This is accomplished by storing a string array of all projects that are available in the wizard, another string array of all projects currently attached.  Each array is then canonicalized, then trimmed of "http", etc.  Arrays are then compared and if any match, the text is set to gray.  The ability to make the text gray was enabled by adding the property "wxLB_OWNERDRAW" to the wxListBox.

For a "simple" change, this is actually a bit more complicated.  A few design notes:
1. Adding the Ownerdraw property did more than provide the ability to change the text color.  The background color of the wxListBox, wherenever there is text, also changed to grey.  I discovered this was a bug and not a feature in wxWidgets, and have reported it [here.](https://github.com/wxWidgets/wxWidgets/issues/24917)  It is being fixed with [this pull request](https://github.com/wxWidgets/wxWidgets/pull/24918).  I haven't tested the fix yet (I'm not sure I have the skill set to change BOINC's wxWidgets dependency).  Therefore, if you were to built with this PR, it looks like this:
image

     I had hard coded a white background before I noticed it was a bug with wxWidgets, and with that it looks like this:
     image
     Of course, that white background would be redundant once the wxWidgets PR is merged, so that hard coding is not in this PR.

2. I have not tested this in Linux nor MacOS.  I also have not reviewed the code for anything dark mode settings.  I will need someone to check on those platforms to see if I broke anything.  FWIW, I run Windows in dark mode.
3. I am not an expert in accessibility, but a 4.6:1 contrast ratio looked good to me.  I tried to get a 7:1 ratio, but the gray looked too close to black to me and was hard to notice the distinction.

From what I have worked on, I think this PR is ready, but I kindly request others to test this on other platforms, especially with dark mode, before merging.

**Alternate Designs**
Hiding the projects that are attached didn't seem like a good idea to me.  If someone wanted to find out the information of the project in this wizard, they would not be able to unless they removed the project.  By graying out the projects that are attached, the users is able to see which projects that are still active and they are attached to.

**Release Notes**
[Manager] Projects listed in Add a Project wizard that are already added are changed to grey.

### DIFF
--- a/clientgui/ProjectInfoPage.cpp
+++ b/clientgui/ProjectInfoPage.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // https://boinc.berkeley.edu
-// Copyright (C) 2022 University of California
+// Copyright (C) 2024 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -818,36 +818,14 @@ void CProjectInfoPage::OnPageChanging( wxWizardExEvent& event ) {
 
 	std::string new_project_url = (const char*)m_strProjectURL.mb_str();
 	canonicalize_master_url(new_project_url);
-	// remove http(s):// at the beginning of project address
-	// there is no reason to connect to secure address project
-	// if we're already connected to the non-secure address
-	// or vice versa
-	// also clear last '/' character if present
-	size_t pos = new_project_url.find(http);
-	if (pos != std::string::npos) {
-		new_project_url.erase(pos, http.length());
-	}
-	else if ((pos = new_project_url.find(https)) != std::string::npos) {
-		new_project_url.erase(pos, https.length());
-	}
-	if (new_project_url.length() >= 1 && new_project_url[new_project_url.length() - 1] == '/') {
-		new_project_url.erase(new_project_url.length() - 1, 1);
-	}
+    TrimURL(new_project_url);
  	for (int i = 0; i < pDoc->GetProjectCount(); ++i) {
  	    PROJECT* project = pDoc->project(i);
         if (project) {
             std::string project_url = project->master_url;
 
             canonicalize_master_url(project_url);
-
-            if ((pos = project_url.find(http)) != std::string::npos) {
-                project_url.erase(pos, http.length());
-            } else if ((pos = project_url.find(https)) != std::string::npos) {
-                project_url.erase(pos, https.length());
-            }
-			if (project_url.length() >= 1 && project_url[project_url.length() - 1] == '/') {
-				project_url.erase(project_url.length() - 1, 1);
-			}
+            TrimURL(project_url);
 
             if (project_url == new_project_url) {
                 bAlreadyAttached = true;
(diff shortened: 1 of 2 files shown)


computezrmle:
There should be a hint explaining the meaning of the different color, like:
"To choose a project, click its name or type its URL below.
Greyed projects are already attached."

As an option or in addition something like this could be at the top of the "Project details" page:
"This project is already attached to your computer.
---"

There may also be a new entry in the "Categories" drop-down list like:
"attached"

AenBleidd:
@Vulpine05, please fix linux build

Vulpine05:
Hmm...I may have to go back to the drawing board on this one.  Apparently Owner Drawn is a feature in wxWidgets for Windows build, but not for mac nor Linux.  I suppose I could add precompiler code to use this feature only for Windows, but that kindof defeats the purpose of the PR if it can only work on one platform.

Vulpine05:
> There should be a hint explaining the meaning of the different color, like: "To choose a project, click its name or type its URL below. Greyed projects are already attached."
> 
> As an option or in addition something like this could be at the top of the "Project details" page: "This project is already attached to your computer. ---"
> 
When you click on the Next button a message does pop up stating that the project is added.  I am assuming the user has some knowledge of what project(s) they are attached to, so I am hoping that is obvious enough.  Your idea isn't bad, but in my opinion I would rather keep the window cleaner.

> There may also be a new entry in the "Categories" drop-down list like: "attached"

Neat idea, but may be not as easy to add to the code.  Again, just my opinion, so if there is strong support for this, I'd be willing to work on it.

Vulpine05:
Okay, this should be ready for review.  I only tested this on Windows, so if someone can test on Mac and Linux it would be appreciated.

@CharlieFenton, I have one concern with running this on Mac.  Per the wxWidgets manual [here](https://docs.wxwidgets.org/3.2/classwx_list_ctrl.html), it says that report mode for a wxListCtrl is generic for all modes.  I'm not sure exactly what that means.  Could you see how it looks?

Here is how it looks to me in Windows:
Test of wxListCtrl - 2024-12-22

AenBleidd (approved):
Code itself looks good to me, but it still need to be tested.
Will try to test it on Windows and Linux during the next few days

CharlieFenton:
> Per the wxWidgets manual [here](https://docs.wxwidgets.org/3.2/classwx_list_ctrl.html), it says that report mode for a wxListCtrl is generic for all modes. I'm not sure exactly what that means

The implementation on Windows uses a built-in API in the Windows OS. There is no such API on the Mac (and perhaps other platforms), so wxWidgets provides a generic implementation in code to be used on those platforms, simulating what the Widow API does.

II tested it, and it works on the Mac, more or less. while the projects to which you are already attached are shown in a lighter gray color, they are not disabled. You can still select them and click on the Next button, which then tells you you are already attached to the project. Here aer 3 screen shots.

Screenshot 2024-12-22 at 6 52 22 PM

Screenshot 2024-12-22 at 6 53 02 PM

Screenshot 2024-12-22 at 6 53 14 PM

CharlieFenton:
I assume the title of this PR "Gray out projects in project wizard that are already attached" means that the intent is to disable those projects in the list so that attempted clicks on the are ignored, which is what a gray item in a list normally means. This does not accomplish that goal, but only changes the appearance ofthose projects in the list.

However, it probably works well enough since clicking on the item brings up the "already attached" alert.

Vulpine05:
> I assume the title of this PR "Gray out projects in project wizard that are already attached" means that the intent is to disable those projects in the list so that attempted clicks on the are ignored, which is what a gray item in a list normally means. This does not accomplish that goal, but only changes the appearance ofthose projects in the list.
> 
> However, it probably works well enough since clicking on the item brings up the "already attached" alert.

Thanks for testing this and sharing screen shots, @CharlieFenton.  I'm glad to see that it functions the same as in Windows.  I understand how you and others assume that with the grayed out text then the next button should be disabled.  However, the way the project wizard currently operates is that it checks the project URL field, which is editable.  Although you can select a project and have the information filled out in the project details, the user can still modify the URL so they can add a custom project not on the list, such as SETI's former beta test project.

Does this work well enough?  I think so.  Is there a better way?  Probably, but there are enough guardrails to prevent the user from doing something that causes an error.

CharlieFenton:
> Does this work well enough? I think so

I think so too.

PR #5867, opened 2024-10-30
client: measure CPU time of VBox apps correctly

davidpanderson:
We were checking for VBoxSVC.exe and, if a VBox app is running, counting its CPU time as BOINC-related.
Should be 'VBoxHeadless.exe' on Win.
May be different on Mac/Linux, so use '*vbox*', case-insensitive

### DIFF
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -363,7 +363,7 @@ void ACTIVE_TASK_SET::get_memory_usage() {
     int retval;
     static bool first = true;
     double diff=0;
-    bool using_vbox = false;
+    bool vbox_app_running = false;
 
     if (!first) {
         diff = gstate.now - last_mem_time;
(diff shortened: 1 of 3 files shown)


computezrmle:
Not all vbox processes are necessarily BOINC related.
Imagine VMs running in the background to provide special services like
- E-mail
- DB
- a compile service

Thus, only those processes should be included that are in the BOINC process tree.
In addition there may be more than 1 BOINC client running on a computer and each of them has its own process tree.

AenBleidd:
@davidpanderson, please fix Windows build

AenBleidd:
@computezrmle, could this be later better handled with the new functionality introduced in #5870?

AenBleidd (approved):

PR #5915, opened 2024-11-23
[vboxwrapper] bump development version to 26209

AenBleidd:


NenTech:
@AenBleidd The milestone is called 26210. The version number in the wrapper itself says 26209.
What is the right version number? It should be 26209 I think.

AenBleidd:
@NenTech, Milestone is set to the next (yet unreleased) version.
Before the release the version of the vboxwrapper will be changed to be the same as the milestone

PR #5918, opened 2024-11-25
Web and server: Add web-based job submission for BUDA

davidpanderson:
Design goals:
- Be simple and user-friendly.  For example:
    - when showing jobs or files, use names that the user supplied,
        rather than DB IDs or names generated by BOINC.
    - Don't expose the distinction between logical and physical names.
    - Don't expose the idea of file immutability.
        When users submit jobs, they use whatever filenames they want.
        These files get copied to physical names (which the user doesn't see).
        Users can reuse the sames names for other files.
- Have a plan for file deletion that will eventually
    delete all unused files (app, input, output).

To do this, I redesigned a couple of server-side features.
(If any projects are currently using these features,
they'll need to do a reset; I don't think any are.)

- The user file sandbox.
    The web interface is the same, but I changed the implementation.
- The scheme for storing output files,
    used by various job-submission systems.
    These are now unified into a single scheme.


html/user/buda_submit.php
```diff
@@ -0,0 +1,351 @@
...
+        $fname
+    ));
+}
+function file_info_out($i) {
+    return sprintf(
+'    <file_info>
+        <name><OUTFILE_%d/></name>
+        <generated_locally/>
+        <upload_when_present/>
+        <max_nbytes>5000000</max_nbytes>
```
AenBleidd, line 241: Of course it depends on the application type, but I assume 5MB is too few.

samples/docker_wrapper/test_buda/worker_3_x86_64-pc-linux-gnu
AenBleidd, line 1: I don't think we should put this binary file into this directory, since this is the file we are building, and it's better to have it ignored via `.gitignore` file to keep it for local testing

AenBleidd (changes requested):

html/user/buda.php
```diff
@@ -0,0 +1,301 @@
...
+    header("Location: buda.php");
+}
+
+function view_file() {
+    global $buda_root;
+    $app = get_str('app');
+    $variant = get_str('variant');
+    $fname = get_str('fname');
+    echo "<pre>\n";
+    readfile("$buda_root/$app/$variant/$fname");
```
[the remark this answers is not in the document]
davidpanderson: I added code to validate GET args in both buda.php and buda_submit.php

html/user/buda_submit.php
davidpanderson: At some point we'll need to let the job submitter specify this; 5MB is OK for now

AenBleidd (approved):

PR #5926, opened 2024-11-29
Compile osx manager with vcpkg

talregev:
Compile osx manager with vcpkg.


AenBleidd:
Thank you for your pull request.
Currently we do not plan to support vcpkg in the MacOS build pipeline.
Will keep it in draft for future discussion and review.

CharlieFenton:
The current build system guarantees that each version of BOINC is built for MacOS with a specific versions of all the dependencies. Since vcpkg could result in the MacOS Manager being built with different versions of dependencies, I would no longer be willing to support BOINC because of the added difficulty in debugging bug reports from the field.

Issue #4697, opened 2022-04-07
[Windows] Provide an installer for arm64


AenBleidd:

martinsuchan:
Hi, I just bought my new laptop - it's a Microsoft Copilot+ PC Lenovo Yoga Slim 7x with the new **ARM64 Snapdragon X Elite CPU and Windows 11 ARM**.
I'm wondering what are the plans for releasing stable BOINC client natively compiled and running on Windows 11 ARM64? Right now I run the x64 client 8.0.2 in emulated mode. Thanks

AenBleidd:
Hello, the current plan is to have the new installer ready for testing before the end of this year.
So stay tuned.

martinsuchan:
Is there a ticket for adding the ARM64 installer to the downloads page? It's still nowhere to be found, thanks.
https://boinc.berkeley.edu/download_all.php

AenBleidd:
@martinsuchan, it's not released yet, so obviously it's not present on the download page.
Once it will be ready - we will make an alpha release, make an announcement, and it will appear on the download page.
Stay tuned ;)
If in the meanwhile you are interested in testing the dev build - please let me know and I'll send you one.

PR #5928, opened 2024-12-01
Draft: windows on arm lib support

ahorek:
Related to #4697

**Description of the Change**
this allows compiling BOINC libraries with https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/ for cross-compilation. Windows on ARM will officially be supported in the upcoming GCC 15.

**Alternate Designs**
Visual Studio is an alternative, it allows ARM targets, but it doesn't support cross-compilation, so using a GCC is more convenient because there's no need to access the real hardware.

while this change fixes the build
_./configure --host="aarch64-windows" --with-boinc-platform="windows_aarch64" --disable-server --disable-manager --disable-client --disable-shared CXXFLAGS="-O3 "_

I'm not sure what the platform string should be... ? https://boinc.berkeley.edu/trac/wiki/BoincPlatforms

I also have to run more tests to ensure it works with apps, so it's a draft for now, but any feedback is welcome. Thanks!

### DIFF
--- a/lib/boinc_win.h
+++ b/lib/boinc_win.h
@@ -148,6 +148,8 @@
 #define SECURITY_WIN32
 #endif
 
+#include <cstddef>
+
 #if !defined(__CYGWIN32__) || defined(USE_WINSOCK)
 
 /* If we're not running under CYGWIN use windows networking */
--- a/lib/hostinfo.cpp
+++ b/lib/hostinfo.cpp
@@ -75,7 +75,9 @@ void HOST_INFO::clear_host_info() {
     safe_strcpy(os_version, "");
 
 #ifdef _WIN64
+#if !defined(__aarch64__)
     wsl_distros.clear();
+#endif
 #else
     safe_strcpy(docker_version, "");
     safe_strcpy(docker_compose_version, "");
@@ -138,10 +140,12 @@ int HOST_INFO::parse(XML_PARSER& xp, bool static_items_only) {
         if (xp.parse_str("os_name", os_name, sizeof(os_name))) continue;
         if (xp.parse_str("os_version", os_version, sizeof(os_version))) continue;
 #ifdef _WIN64
+#if !defined(__aarch64__)
         if (xp.match_tag("wsl")) {
             this->wsl_distros.parse(xp);
             continue;
         }
+#endif
 #else
         if (xp.parse_str("docker_version", docker_version, sizeof(docker_version))) continue;
         if (xp.parse_str("docker_compose_version", docker_compose_version, sizeof(docker_compose_version))) continue;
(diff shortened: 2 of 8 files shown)


AenBleidd:
Thank you for your PR.
I converted it to 'Draft' since the same is mentioned in your PR title, and will review it later.
Just for the reference, Visual Studio supports cross-compilation, you just need to install required components to it, and it already builds for ARM64 on our CI:
- https://github.com/BOINC/boinc/actions/runs/12100618832/job/33739454467
- https://github.com/BOINC/boinc/actions/runs/12100618832/job/33739454280

lib/hostinfo.cpp
```diff
@@ -75,7 +75,9 @@ void HOST_INFO::clear_host_info() {
     safe_strcpy(os_version, "");
 
 #ifdef _WIN64
+#if !defined(__aarch64__)
```
AenBleidd, line 78: All these are not required since WSL is supported on ARM64 devices as well: https://learn.microsoft.com/en-us/windows/wsl/install-manual

lib/boinc_win.h
```diff
@@ -148,6 +148,8 @@
 #define SECURITY_WIN32
 #endif
 
+#include <cstddef>
```
AenBleidd, line 151: is this really required?

AenBleidd (changes requested):
Please also add one more CI flow to run ARM64 builds with mingw: https://github.com/BOINC/boinc/blob/master/.github/workflows/mingw.yml

lib/boinc_win.h
ahorek: it's for size_t https://github.com/BOINC/boinc/blob/master/lib/boinc_win.h#L164C9-L164C15

lib/hostinfo.cpp
ahorek: ok, I'm encountering a relocation error during linking... let's see if I can find a better solution

ahorek:
> Thank you for your PR. I converted it to 'Draft' since the same is mentioned in your PR title, and will review it later. Just for the reference, Visual Studio supports cross-compilation, you just need to install required components to it, and it already builds for ARM64 on our CI:
> 
>     * https://github.com/BOINC/boinc/actions/runs/12100618832/job/33739454467
> 
>     * https://github.com/BOINC/boinc/actions/runs/12100618832/job/33739454280

ok. Even with an empty hello world project, I'm getting LNK1112: module machine type 'x86' conflicts with target machine type 'ARM' , and the binary isn't generated. I'll see if there's a way to fix it somehow. Thanks

ahorek:
unfortunately, using https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/ the app is crashing at boinc_init(). Without a backtrace or tools like gdb (which aren't available on Windows ARM), I probably won't be able to fix it.

It would be great to support GCC 15 for building Win+ARM once it becomes officially available. https://www.phoronix.com/news/GCC-aarch64-w64-mingw32 but this change doesn't seem to be sufficient.

ahorek:
btw I successfully built and tested the app in Visual Studio after installing additional ARM SDK libraries.

PR #5951, opened 2024-12-13
remove useless redirect

homersimpsons:
**Description of the Change**
Remove useless call to `redirect_to_secure_url`.

The call is already made in a file that is included:
https://github.com/BOINC/boinc/blob/c67393e6d11e4dafe11b54fd9470fa254f3b1c36/html/inc/util.inc#L1167

Also, the function `redirect_to_secure_url` requires 0 parameter, here we were providing one:
https://github.com/BOINC/boinc/blob/c67393e6d11e4dafe11b54fd9470fa254f3b1c36/html/inc/util.inc#L1156

**Release Notes**
N/A


AenBleidd:

PR #5952, opened 2024-12-14
Consistent DB `enum` function

### LINKED ISSUE #1813: DB layer enum functions are inconsistent
The basic `enum*` functions in `html/inc/db_conn.inc` support an argument for _where_ and _order_ clauses and the _where_ clause must not start with "where". The table specific `enum` functions in `html/inc/boinc_db.inc` sometimes have one sometimes two arguments. See `BoincProfile::enum` and `BoincCreditUser::enum`.

This should be made consistent, as in: All table specific `enum` implementations should support two arguments. In addition a third argument for _limit_ clauses should be added to the basic functions and the class implementations. This will make the DB layer more generic and will allow switching to something other than mysql in the future.

homersimpsons:
Fixes #1813

**Description of the Change**
Make all DB `enum` functions consistent

About the following requirement from the task:
> In addition a third argument for limit clauses should be added to the basic functions and the class implementations

Should I just update the `DbConn::enum` and `DbConn::enum_fields`?

**Release Notes**
N/A

### DIFF
--- a/html/inc/boinc_db.inc
+++ b/html/inc/boinc_db.inc
@@ -204,6 +204,11 @@ class BoincUser {
         $db = BoincDb::get();
         return $db->update($this, 'user', $clause);
     }
+    /**
+     * @param string|null $where_clause
+     * @param string|null $order_clause
+     * @return list<BoincUser>
+     */
     static function enum($where_clause, $order_clause=null) {
         $db = BoincDb::get();
         return $db->enum('user', 'BoincUser', $where_clause, $order_clause);
@@ -258,6 +263,11 @@ class BoincTeam {
         $db = BoincDb::get();
         return $db->update($this, 'team', $clause);
     }
+    /**
+     * @param string|null $where_clause
+     * @param string|null $order_clause
+     * @return list<BoincTeam>
+     */
     static function enum($where_clause, $order_clause=null) {
         $db = BoincDb::get();
         return $db->enum('team', 'BoincTeam', $where_clause, $order_clause);
@@ -297,9 +307,14 @@ class BoincTeamDelta {
         $db = BoincDb::get();
         return $db->insert('team_delta', $clause);
     }
-    static function enum($where_clause) {
+    /**
+     * @param string|null $where_clause
+     * @param string|null $order_clause
+     * @return list<BoincTeamDelta>
+     */
+    static function enum($where_clause, $order_clause=null) {
         $db = BoincDb::get();
-        return $db->enum('team_delta', 'BoincTeamDelta', $where_clause);
+        return $db->enum('team_delta', 'BoincTeamDelta', $where_clause, $order_clause);
     }
     static function delete_for_user($user_id) {
         $db = BoincDb::get();
@@ -321,6 +336,11 @@ class BoincHost {
         $db = BoincDb::get();
         return $db->delete($this, 'host');
     }
+    /**
+     * @param string|null $where_clause
+     * @param string|null $order_clause
+     * @return list<BoincHost>
+     */
     static function enum($where_clause, $order_clause=null) {
         $db = BoincDb::get();
         return $db->enum('host', 'BoincHost', $where_clause, $order_clause);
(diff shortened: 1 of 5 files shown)


html/inc/boinc_db.inc
```diff
@@ -204,6 +204,11 @@ class BoincUser {
         $db = BoincDb::get();
         return $db->update($this, 'user', $clause);
     }
+    /**
+     * @param string|null $where_clause
+     * @param string|null $order_clause
+     * @return list<BoincUser>
+     */
```
homersimpsons, line 211: I added the typing. If we target recent php versions I can just add it directly in the function signature (except for the generic return).

html/ops/create_forums.php
```diff
@@ -29,7 +29,7 @@
 
 function create_category($orderID, $name, $is_helpdesk) {
     $q = "(orderID, lang, name, is_helpdesk) values ($orderID, 1, '$name', $is_helpdesk)";
-    $db = BoincDB::get();
+    $db = BoincDb::get();
```
homersimpsons, line 32: The case was incorrect. PHP is case-insensitive but it is better to us the correct casing.

html/inc/boinc_db.inc
homersimpsons, line 1: I would like to split this file to have 1 file per class for OOP. I'm not sure that this is possible with the current setup though.

html/inc/boinc_db.inc
```diff
@@ -505,6 +545,11 @@ class BoincProfile {
         $db = BoincDb::get();
         return $db->insert('profile', $clause);
     }
+    /**
+     * @param string|null $where_clause
+     * @param string|null $order_clause
+     * @return list<BoincProfile>
+     */
     static function enum($where_clause=null, $order_clause=null) {
```
homersimpsons, line 553: This is the only occurence of `$where_clause=null`, should I keep it? Or should I check for usages?

html/inc/boinc_db.inc
```diff
@@ -297,9 +307,14 @@ class BoincTeamDelta {
...
+    /**
+     * @param string|null $where_clause
+     * @param string|null $order_clause
+     * @return list<BoincTeamDelta>
+     */
+    static function enum($where_clause, $order_clause=null) {
         $db = BoincDb::get();
-        return $db->enum('team_delta', 'BoincTeamDelta', $where_clause);
+        return $db->enum('team_delta', 'BoincTeamDelta', $where_clause, $order_clause);
     }
```
homersimpsons, line 318: Would it make sense to have this function in a trait?

AenBleidd:

davidpanderson:
There's a better way to do the DB API, but a this point there's no
reason to change what we have.
Let's not change the coding style.

homersimpsons:
> There's a better way to do the DB API, but a this point there's no reason to change what we have. Let's not change the coding style.

Can you update your vision on #1813? I may be able to tackle this.

davidpanderson:
It's possible to use inheritance so that the table-specific classes (BoincUser etc.)
share the same enum(), lookup() etc. functions.
This would greatly shrink the code.
However, our current policy with the PHP code is: if it ain't broken don't fix it.

Something that would be worth doing: make sure that SQL injection attacks are impossible.
The DB APIs are often called with arguments that are user-supplied: GET/POST args and cookies.
String args are generally 'sanitized' with escape_string(), but:
- numeric args should be sanitized too, e.g. with intval()
- right now it's vague whether sanitization is the job of the caller or the API;
  this should be nailed down.

I can create an issue if you'd like.

homersimpsons:
> right now it's vague whether sanitization is the job of the caller or the API;

If we take example from PDO, DBAL and Eloquent they all favor parameterized queries. So it is the responsibility of the caller to use parameters and the API guarantee that those cannot be used as SQL injection.
If I recall correctly this also has the advantages that some SQL engine will be able to parse / analyse the SQL query only once instead of doing it every time there is a parameter that changes.

PR #5953, opened 2024-12-14
Delete html/ops/sample_table_stats.php

homersimpsons:
**Description of the Change**

This file used a function that does not exists `_mysql_select_db` so it was most likely failing.

Reading its content I believe it is vetter to remove it.

**Release Notes**
N/A


davidpanderson:
May as well just fix it; see #5959

PR #5960, opened 2024-12-15
client, server, web: enable BUDA GPU apps

davidpanderson:
With BUDA, the (BOINC) app version has just the docker wrapper.
Everything else (Dockerfile, executables) is part of the workunit,
along with the input files.

We want to support GPU applications in BUDA.
That implies that:
- The plan class (saying what GPU type and driver are needed)
    is part of the workunit rather than the app version
- The resource usage (GPU, fraction of CPU) is part of
    the workunit that's sent to and stored on the client.

This required changes to both scheduler and client
(and to a small extent web).
Fortunately I was able to keep the changes fairly simple.

DB:
    When we create a BUDA workunit, we stores its plan class
    as an element of its xml_doc, where the scheduler can see it.

Scheduler:
    If a workunit has a plan class,
    call the plan class function to see if we can send it to the host
    and if so to get the usage info.
    Include the usage info in the <workunit> element in the scheduler reply.

Feeder:
    It makes a list of GPU types the project can use;
    this is used in scheduler replies.
    This list now must reflect not only APP_VERSION plan classes,
    but also BUDA app variants.
    We do this using a file 'buda_plan_classes'
    that's maintained by the web code.

Client:
    A new struct RESOURCE_USAGE has GPU/CPU usage info.
    APP_VERSION and WORKUNIT (for BUDA jobs) both have one.
    The appropriate one is copied to RESULT when it's created.
    Scheduling and work fetch code references this copy.

Scheduler protocol:
    <workunit> now can include plan class and resource usage info

### DIFF
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -285,7 +285,7 @@ int ACTIVE_TASK::init(RESULT* rp) {
     result = rp;
     wup = rp->wup;
     app_version = rp->avp;
-    max_elapsed_time = rp->wup->rsc_fpops_bound/rp->avp->flops;
+    max_elapsed_time = rp->wup->rsc_fpops_bound/rp->resource_usage.flops;
     if (max_elapsed_time < MIN_TIME_BOUND) {
         msg_printf(wup->project, MSG_INFO,
             "Elapsed time limit %f < %f; setting to %f",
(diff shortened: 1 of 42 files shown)


AenBleidd:
@davidpanderson, please fix build errors.

AenBleidd:
@davidpanderson, unfortunately, still failing:
```
sched_shmem.cpp: In function ‘void get_buda_plan_classes(std::vector<std::__cxx11::basic_string<char> >&)’:
sched_shmem.cpp:116:28: error: cannot convert ‘FCGI_FILE*’ to ‘FILE*’
  116 |     while (fgets(buf, 256, f)) {
      |                            ^
      |                            |
      |                            FCGI_FILE*
```
also, please run `python3 ci_tools/trailing_whitespaces_check.py . --fix` to remove extra whitespaces at the end of the lines.
Thank you.

davidpanderson:
possibly fixed

AenBleidd (approved):

PR #5967, opened 2024-12-20
web: fix various vulnerabilities

davidpanderson:
- Send all cookies as HttpOnly (don't let Javascript see them)
- If page sending cookie is HTTPS, make the cookie secure
- Often we have something like: $name = get_str('name'); if (!lookup($name)) { error_page("can't find $name"); } Can't do this; it can be exploited for XSS attacks. Just say 'Can't find file' or whatever
- Don't show database error messages

### DIFF
--- a/html/inc/consent.inc
+++ b/html/inc/consent.inc
@@ -48,7 +48,9 @@ function consent_to_a_policy(
 function check_user_consent($user, $consent_name) {
     list($checkct, $ctid) = check_consent_type($consent_name);
     if ($checkct) {
-        $consent_result = BoincLatestConsent::lookup("userid={$user->id} AND consent_type_id=$ctid AND consent_flag=1");
+        $consent_result = BoincLatestConsent::lookup(
+            "userid=$user->id AND consent_type_id=$ctid AND consent_flag=1"
+        );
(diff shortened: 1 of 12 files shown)


AenBleidd (approved):

lfield:
Tested. LGTM

Rytiss:
@lfield @davidpanderson The commit is broken. https://github.com/BOINC/boinc/compare/server_release/1.4/1.4.5...server_release/1.4/1.4.6#diff-d5fbb528304cce7cc7425fdb9fc52f69865c650a7509060720c0f797d225ade3R165 contains git diff lines in the main PHP code (<<< HEAD, ===== etc) in more than one place.

Moreover, regarding is_valid_filename() - htmlspecialchars($x) is used to check for invalid characters. However, this is not appropriate for checking valid filenames. htmlspecialchars is meant for escaping HTML characters (like quotes, <, >, and &) and may incorrectly reject valid filenames with special characters. These characters are perfectly valid characters in filenames.

Could we stop panicking and making new server releases one per day, and instead wait and review stuff properly?

lfield:
Thanks for this. Something went wrong during the cherry pick.

PR #4953, opened 2022-10-06
changes systemd flag ProtectSystem from strict to full

Germano0:
Fixes issue [3355](https://github.com/BOINC/boinc/issues/3355).
Below I am copying the description originally available at https://github.com/BOINC/boinc/issues/3355#issuecomment-1264751840

**Description of the Change**
Before this change, BOINC working units that make use of Virtualbox could not properly run on Linux, like the ones from LHC@home.
To fix this problem I used the default boinc-client systemd unit file, and just replaced
```
ProtectSystem=strict
```
with
```
ProtectSystem=full
```

Before such change, I made many different attempts, while having the old value `ProtectSystem=strict`

- adding file `/etc/tmpfiles.d/boinc-vbox.conf` containing `d /tmp/.vbox-boinc-ipc 700 boinc boinc - -`
- adding `-/tmp/.vbox-boinc-ipc` in `ReadWritePaths`

but they all failed.

I also had to set SELinux to `permissive` mode in order to be able to successfully run VBox based working units. Of course this is outside what concerns BOINC code base, but it worth to be mentioned.

**INFORMATIONS ABOUT THE TEST ENVIRONMENT**

**System environment**
```
Fedora 36
systemd 250.8
Linux kernel 5.18.18
SELinux on permissive mode
BOINC 7.20.2
VirtualBox 6.1.38 and its following kernel modules: vboxnetadp, vboxnetflt, vboxdrv, vboxnetadp, vboxnetflt
```

**systemd unit file content**
```
[Unit]
Description=Berkeley Open Infrastructure Network Computing Client
Documentation=man:boinc(1)
Wants=vboxdrv.service
After=vboxdrv.service network-online.target

[Service]
Type=simple
ProtectHome=true
ProtectSystem=full
ProtectControlGroups=true
ReadWritePaths=-/var/lib/boinc -/etc/boinc-client
Nice=10
User=boinc
WorkingDirectory=/var/lib/boinc
ExecStart=/usr/bin/boinc
ExecStop=/usr/bin/boinccmd --quit
ExecReload=/usr/bin/boinccmd --read_cc_config
ExecStopPost=/bin/rm -f lockfile
IOSchedulingClass=idle

[Install]
WantedBy=multi-user.target
```

To ensure reliability of the tests, after each systemd unit file change, and/or creation or removal of `/etc/tmpfiles.d/boinc-vbox.conf`, I always have done the following:

1. erased all current/pending working units
2. blocked LHC@home from requesting new working units
[...]

Issue #3355, opened 2019-10-28
[Explained] 7.16.3 systemd startup file prevents LHC VirtualBox jobs running


original-birdman:
**Describe the bug**
I tried an update to 7.16.3, but it produced an oddity, so I switched back to 7.14.2
But I left the 7.16.3 systemd start-up file in place.

The result was that an LHC CMS task started, but then stalled.
It ran for ~6mins (during which the VirtualBox process wasn't consuming CPU) then this showed up in the log:
```
Oct 28 20:03:53 benuc boinc[16837]: 28-Oct-2019 20:03:53 [LHC@home] Task CMS_1316885_1572237828.669283_0 postponed for 86400 seconds: Communication with VM Hypervisor failed.
```

Re-installing the 7.14.2 start-up file and re-starting BOINC (no other change) has allowed the job to run.

**Steps To Reproduce**
1. Try running an LHC CMS job using the 7.16.3 start-up file.

**Expected behavior**
I'd expect jobs to be able to run successfully.

**System Information**
 - OS: Kubuntu 19.10
 - BOINC Version: 7.16.3 (start-up file only)

**Additional context**
I'll post the working and non-working systemd start-up files in the follow-ups.

original-birdman:
This is the (minorly-modified to reflect my path usage) 7.16.3 systemd config file.
This FAILS.

```
[Unit]
Description=Berkeley Open Infrastructure Network Computing Client
Documentation=man:boinc(1)
After=network-online.target

[Service]
Type=simple
ProtectHome=true
PrivateTmp=true
ProtectSystem=strict
ProtectControlGroups=true
#GML ReadWritePaths=-/var/lib/boinc -/etc/boinc-client
ReadWritePaths=-/extra/BOINC/wlib
Nice=10
User=boinc
#GML WorkingDirectory=/var/lib/boinc
#GML ExecStart=/extra/BOINC/bin/boinc
WorkingDirectory=/extra/BOINC/wlib
ExecStart=/extra/BOINC/bin/boinc --dir /extra/BOINC/wlib --start_delay 20
ExecStop=/extra/BOINC/bin/boinccmd --quit
ExecReload=/extra/BOINC/bin/boinccmd --read_cc_config
ExecStopPost=/bin/rm -f lockfile
IOSchedulingClass=idle
# The following options prevent setuid root as they imply NoNewPrivileges=true
# Since Atlas requires setuid root, they break Atlas
# In order to improve security, if you're not using Atlas,
[...]

Issue #495, opened 2015-02-04
Unable to activate 'Show Graphics' tab


romw:
**Reported by john stevens on 7 Feb 37955190 01:20 UTC**
I've installed BOINC (version 5.10.28) and am running it successfully on my computer, but when I installed it on my wife's computer (both Dells), the 'Show Graphics' button on the 'Tasks' page is grayed out (disabled) and I can't change it.

Any idea why?

John Stevens

romw:
**Commented by Ageless on 24 Jan 37958941 15:06 UTC**
BOINC by itself doesn't crunch a thing, it's a managing program. You need to attach to a project to do something with it. Not all projects have a graphics option build in. Only a handful of them do. 

I'm calling this not a bug. If you have further questions, go to [the BOINC forums](http://boinc.berkeley.edu/dev/index.php), register there and post  questions. That's where we can help you further. When you do post there, make sure to add more detail, such as which OS, what kind of videocard, DirectX version, which projects you're attach to etc. 

(Do not post your email address in a legible form either. The spam bots reading those forums and this here Trac love that).

romw:
**Commented by Nicolas on 21 Nov 37959397 11:33 UTC**
I deleted his first comment and edited the description to get rid of his email address and save him some spam.

PR #5977, opened 2024-12-28
Revert to 'ProtectSystem=strict' in boinc-client.service and make '/tmp' writable

computezrmle:
BOINC volunteers running VirtualBox tasks on Linux occasionally report errors like this if BOINC runs as a systemd service:

```
2024-12-28 01:20:15 (16867): Detected: VirtualBox VboxManage Interface (Version: 7.0.12)
2024-12-28 01:20:20 (16867): Error in host info for VM: -182
Command:
VBoxManage -q list hostinfo 
Output:
VBoxManage: error: Failed to create the VirtualBox object!
VBoxManage: error: Code NS_ERROR_SOCKET_FAIL (0xC1F30200) - IPC daemon socket error (extended info not available)
VBoxManage: error: Most likely, the VirtualBox COM server is not running or failed to start.


2024-12-28 01:20:20 (16867): WARNING: Communication with VM Hypervisor failed.
2024-12-28 01:20:20 (16867): ERROR: VBoxManage list hostinfo failed
2024-12-28 01:20:20 (16867): called boinc_finish(1)
```

Usually this can be tracked down to a too restrictive setting in BOINC's systemd file:
`ProtectSystem=strict`

In most cases experienced volunteers suggest to replace `strict` with `full`, run `systemctl daemon-reload` and restart BOINC.
This works since - unlike `strict` - `full` leaves `/tmp` writable for BOINC and it's child processes.
The disadvantage is that it also leaves many other directories writable.
A better solution might be to leave `strict` and make `/tmp` writable which can be configured using `ReadWritePaths=`.

**Background details**

_From the systemd manual_

```
ProtectSystem=
Takes a boolean argument or the special values "full" or "strict". If true, mounts the /usr/ and the boot loader directories (/boot and /efi) read-only for processes invoked by this unit. If set to "full", the /etc/ directory is mounted read-only, too. If set to "strict" the entire file system hierarchy is mounted read-only, except for the API file system subtrees /dev/, /proc/ and /sys/ (...) If this option is used, ReadWritePaths= may be used to exclude specific directories from being made read-only.
```

_From the VirtualBox manual_

```
VBoxSVC IPC Issues
On Linux, Oracle VirtualBox makes use of a custom version of Mozilla XPCOM (cross platform component object model) for interprocess and intraprocess communication (IPC). (...) Communication between the various Oracle VirtualBox components and VBoxSVC is performed through a local domain socket residing in /tmp/.vbox-username-ipc. In case there are communication problems, such as an Oracle VirtualBox application cannot communicate with VBoxSVC, terminate the daemons and remove the local domain socket directory.
```

### DIFF
--- a/client/scripts/boinc-client.service.in
+++ b/client/scripts/boinc-client.service.in
@@ -7,9 +7,9 @@ After=vboxdrv.service network-online.target
 [Service]
 Type=simple
 ProtectHome=true
-ProtectSystem=full
+ProtectSystem=strict
 ProtectControlGroups=true
-ReadWritePaths=-/var/lib/boinc -/etc/boinc-client
+ReadWritePaths=-/var/lib/boinc -/etc/boinc-client -/tmp
 Nice=10
 User=boinc
 WorkingDirectory=/var/lib/boinc
--- a/tests/linux_package_integration_tests.py
+++ b/tests/linux_package_integration_tests.py
@@ -148,6 +148,7 @@ class IntegrationTests:
     def test_selected_values_from_boinc_client_service_file(self):
         ts = testset.TestSet("Test selected values from the '/usr/lib/systemd/system/boinc-client.service' file")
         data = self._get_key_value_pairs_from_file("/usr/lib/systemd/system/boinc-client.service")
+        ts.expect_equal(data["ProtectSystem"], "strict", "Test 'ProtectSystem' is correctly set")
         ts.expect_equal(data["ReadWritePaths"], "-/var/lib/boinc -/etc/boinc-client -/tmp", "Test 'ReadWritePaths' is correctly set")
         ts.expect_equal(data["User"], "boinc", "Test 'User' is correctly set")
         ts.expect_equal(data["WorkingDirectory"], "/var/lib/boinc", "Test 'WorkingDirectory' is correctly set")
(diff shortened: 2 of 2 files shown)


AenBleidd:
@Germano0, @lfielf, could you please review this PR?
@computezrmle, please fix the tests (IF you don't want or don't know how - I can do that for you)

computezrmle:
> @computezrmle, please fix the tests ...

Done.

AenBleidd:
@Germano0, @lfield, do you have some time to review (in the better case also test) it?

lfield:
LGTM. Tested.

Germano0:
I am against messing up for the n-th time with boinc-client systemd unit file unless URL of of failed boinc working unit is being provided.
- year 2018 systemd unit change https://github.com/BOINC/boinc/pull/2873
- year 2022 systemd unit change https://github.com/BOINC/boinc/pull/4953 which fixed the endless discussion https://github.com/BOINC/boinc/issues/3355
Also, in https://github.com/BOINC/boinc/pull/495 you were not using boinc-client systemd unit file, but a custom one

computezrmle:
I don't agree with arguments like "I am against messing up for the n-th time ...".
If a change solves an issue, it should be integrated.

I remember the discussion #4953 you mentioned.
What you tested there was a combination of `ProtectSystem=strict` and `-/tmp/.vbox-boinc-ipc` in `ReadWritePaths` which didn't work:
https://github.com/BOINC/boinc/pull/4953#issue-1399392492
The reason for that was that you left `/tmp` write protected and VirtualBox could not create `.vbox-boinc-ipc` below.

The solution here is to make `/tmp` writeable which allows VirtualBox to create the required directory.
This follows the manuals from VirtualBox and systemd mentioned above and it is tested.

Whether the systemd service file is taken from the BOINC package or not is not relevant here.

PR #6006, opened 2025-01-14
Typo - FR & ES resource language fix

Borik-ctrl:
Fixes #

**Description of the Change**
I don't know it this is encouraged but I found some typo on FR and ES resource language file.

Also I have noticed a specific text part about Xiaomi, in the res about  and not sure it is up-to-date or should be kept...

**Release Notes**
N/A


AenBleidd:
We are managing translations using Transifex service: https://www.transifex.com/projects/p/boinc/
Please fix it there

PR #6013, opened 2025-01-17
client: fix bug where avg_ncpus for anonymous platform apps defaults to zero

### LINKED ISSUE #6012: app_info.xml configured projects break when 8.1.0-1586 client is restarted
### Describe the bug

When i restart boinc-client while running sidock using app_info all tasks display 0 cpus used and attempt to run all tasks at the same time disregarding use 50% of the cpus.

Image

### Steps to reproduce

1. set sidock to use app_info.xml using this guide: https://www.sidock.si/sidock/forum_thread.php?id=125&postid=2167
2. receive tasks.
3. run sudo systemctl restart boinc-client

### Expected behavior

Image

### Screenshots

_No response_

### System information

OS: Ubuntu 24.04
boinc-client 8.1.0-1586

### Additional context

It doesn't happen on 8.0.4-875

davidpanderson:
Fixes #6012

### DIFF
--- a/client/cs_statefile.cpp
+++ b/client/cs_statefile.cpp
@@ -856,7 +856,8 @@ int CLIENT_STATE::write_state_file_if_needed() {
 
 // look for app_versions.xml file in project dir.
 // If find, get app versions from there,
-// and use "anonymous platform" mechanism for this project
+// and mark project as "anonymous platform".
+// This is called before parsing client_state.xml
 //
 void CLIENT_STATE::check_anonymous() {
     unsigned int i;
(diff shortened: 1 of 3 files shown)


client/cs_statefile.cpp
```diff
@@ -856,7 +856,8 @@ int CLIENT_STATE::write_state_file_if_needed() {
 
 // look for app_versions.xml file in project dir.
 // If find, get app versions from there,
-// and use "anonymous platform" mechanism for this project
```
Vulpine05, line 859: Apologies for being the grammar police, but shouldn't the line above say "If found" instead of "If find"?

AenBleidd (approved):

client/cs_statefile.cpp
davidpanderson: I think that 'If find' is comment-shorthand for 'If you find it'.

PR #6018, opened 2025-01-22
vboxwrapper: create the 'virtualbox home directory' in the project dir

davidpanderson:
The 'VBox home directory' is where VBox writes log files
(which are read by vboxwrapper).
If this is not specified by the env var VBOX_USER_HOME,
we need to create it somewhere and set the env var to point there.

Previously we put it in <datadir>/projects.
That's no good because it's not a project, and the client erased it.

We also tried putting it in the (real) user's home dir.
That's no good because
1) we shouldn't mess with the home dir
2) in sandboxed configs we're running as user 'boinc_projects',
and don't have access to the home dir.

According to https://boinc.berkeley.edu/sandbox_design.php,
the only places 'boinc_projects' can write are
projects/, slots/, and their subdirectories.
So the logical places to put .VirtualBox are
this job's slot directory, or its project directory.
I chose the latter.

PR #6016, opened 2025-01-21
vboxwrapper (Mac): fix startup failure on Mac

davidpanderson:
The VBox home dir (where it keeps its files) is
"/Library/Application Support/BOINC Data/.VirtualBox/",
not ~/.Virtualbox (the latter may be inaccessible)

Fixes #6015

PR #6019, opened 2025-01-23
vboxwrapper: virtualbox home directory fix MacOS

NenTech:
Directory was changed but not assigned to the environmental variable. Which caused a failure on MacOS.

Only change which need to be made for this pull request to work is create directory .VirtualBox by default or give boinc_project access to BOINC_data to be able to write the folder.

Related to #6016  and #6018


NenTech:
@davidpanderson
@computezrmle 
I hope I did this the right way..

davidpanderson:
Yes, the setenv() was definitely needed.
I'm going to fold this into another PR that puts .VirtualBox
in the BOINC data directory, not the user's home directory.
I don't think BOINC should touch the home directory
(and it may not have access to it anyway, like on the Mac)

davidpanderson:
This change is incorporated in #6018

PR #6043, opened 2025-01-28
strip docker_wrapper Linux executable

davidpanderson:


AenBleidd:
We already doing that when building both docker_wrapper for release: https://github.com/BOINC/boinc/blob/97b7fde5b18a743b98362bd4785ace1e4c80f532/.github/workflows/linux.yml#L254
I suggest not doing that in the Makefile directly to avoid changing it every time when required to build debug version for local testing.

davidpanderson:
OK

Issue #1741, opened 2016-12-28
BOINC Manager mistreats its windows on Windows 8/10


victory-ja:
using 7.6.33;
how to reproduce:
1) open Manager
2) open Event log window
3) drop focus; click othe application's window
4) click BOINC window or select it from task-bar or whatever it is called
5) the other BOINC window than the one clicked/selected gets focus

right window may get focus when main window is selected,
but it always fail when selected Event log window.

Ageless93:
Using BOINC 7.8.3 on Windows 7 x64, doing the steps above, I cannot reproduce the problem. Anyone else?

AenBleidd:
I can. At least on 7.8.2. I'll check this again in an hour and attach a
video.

Best regards,
Vitalii

Sent via Android

24 окт. 2017 г. 20:35 пользователь "Jord van der Elst" <
notifications@github.com> написал:

AenBleidd:
https://drive.google.com/open?id=0B8jzqKlhXhyMT2I3cXZSSV9kazQ

kotenok2000:
I will attach video here

https://user-images.githubusercontent.com/8822292/152210089-930ce479-4cc2-4983-b088-e69540738eef.mp4

Vulpine05:
I have done a little bit of investigating into this issue, but I think I may have reached a point beyond what I understand.

This bug occurs when select either window from the taskbar icon, or if you ALT + TAB to select the window.  When you click on either window if it is up and visible on your screen, it behaves properly.  Additionally, the application has some code to keep the manager and event log together.  That is, if one is selected, the other is raised prior to raising the window that is selected (see BOINCGUIApp.cpp, L930).  By removing L945 through L955, it will fix the bug, but then the windows won't stay together when you switch between applications.  I'm guessing those were put there for the purpose of keeping all BOINC windows together, so I don't think removing those lines is the solution we want.

So, something is happening with the event for selecting either window with the taskbar icon or keyboard shortcuts that gets tripped up (or is tripping) the events when the application is activated.  What that is, I'm not sure.  If anyone has a hint how those events are processed, please let me know.

PR #6080, opened 2025-02-14
Manager, Win: activate the right window from task bar

davidpanderson:
There was some logic which didn't work.
I commented it out (on Win) and now it works.
Need to check this on Mac and Linux.

### DIFF
--- a/clientgui/BOINCGUIApp.cpp
+++ b/clientgui/BOINCGUIApp.cpp
@@ -1105,6 +1105,8 @@ int CBOINCGUIApp::IdleTrackerDetach() {
 void CBOINCGUIApp::OnActivateApp(wxActivateEvent& event) {
     m_bProcessingActivateAppEvent = true;
 
+#ifndef _WIN32  // On Win, the following raises the wrong window
+
     if (event.GetActive()) {
 #ifdef __WXMAC__
         ShowInterface();
@@ -1131,6 +1133,7 @@ void CBOINCGUIApp::OnActivateApp(wxActivateEvent& event) {
         }
 #endif
     }
+#endif  // _WIN32
 
     event.Skip();
 


AenBleidd:
@davidpanderson, I suggest changing your implementation to this:
```c++
void CBOINCGUIApp::OnActivateApp(wxActivateEvent& event) {
    m_bProcessingActivateAppEvent = true;

    if (event.GetActive()) {
#ifdef __WXMAC__
        ShowInterface();
#else
#ifdef __WXGTK__
        // Linux allows the Event Log to be brought forward and made active
        // even if we have a modal dialog displayed (associated with our
        // main frame.) This test is needed to allow bringing the modal
        // dialog forward again by clicking on its title bar.
        if (!IsModalDialogDisplayed())
        {
            bool keepEventLogInFront = m_bEventLogWasActive;

            if (m_pEventLog && !m_pEventLog->IsIconized() && !keepEventLogInFront) {
                m_pEventLog->Raise();
            }
            if (m_pFrame) {
                m_pFrame->Raise();
            }
            if (m_pEventLog && !m_pEventLog->IsIconized() && keepEventLogInFront) {
                m_pEventLog->Raise();
            }
        }
#endif
#endif
    }

    event.Skip();

    m_bProcessingActivateAppEvent = false;
}
```
It inherits the same code style that is used across the GUI implementation codebase

davidpanderson:
That removes the logic for Mac as well a Win; I didn't want to change Mac

AenBleidd:
@davidpanderson, no.
For Mac literally only `ShowInterface();` function is called both in old and new version of code

davidpanderson:
Oops yeah you're right; I'll change to your version

AenBleidd:
> Oops yeah you're right; I'll change to your version

No worries. It took me several minutes to get the logic between all these `#ifdef`s as well xD

AenBleidd (approved):

Vulpine05:
@davidpanderson, this does fix the bug of the wrong window being selected when clicking on the taskbar icon.  However, this PR seems to have removed the "feature" that was originally caused this issue - trying to keep the Manager window and the Event Log window together when selecting one or the other.  I made some notes about my observations in [this comment](https://github.com/BOINC/boinc/issues/1741#issuecomment-1320762011).

This isn't a feature I use, so there isn't any complaints from me about removing it.  I just want to make sure you were aware of this in case it got overlooked.

davidpanderson:
Yup, I'm aware of it.  Not important or even desirable in my view.

PR #6107, opened 2025-02-23
Refactor code related to wxWidgets deprecations

### LINKED ISSUE #6098: [Manager] Refactor several wxWidgets constructors and/or inputs
### Describe the problem

While building the Manager, several deprecation warnings occur.  These are all part of wxWidgets.  The warnings are:

1. Constructor for wxFont
```
AccountManagerProcessingPage.cpp: In member function ‘void CAccountManagerProcessingPage::CreateControls()’:
AccountManagerProcessingPage.cpp:139:91: warning: ‘wxFont::wxFont(int, int, int, int, bool, const wxString&, wxFontEncoding)’ is deprecated: use wxFONT{FAMILY,STYLE,WEIGHT}_XXX constants [-Wdeprecated-declarations]
  139 |     m_pTitleStaticCtrl->SetFont(wxFont(10, wxSWISS, wxNORMAL, wxBOLD, FALSE, _T("Verdana")));
```

2.  Constructor for wxTimerEvent
```
AdvancedFrame.cpp:1570:25: warning: ‘wxTimerEvent::wxTimerEvent()’ is deprecated: wxTimerEvent not supposed to be created by user code [-Wdeprecated-declarations]
 1570 |         wxTimerEvent    timerEvent;
```

3. Constructor for wxPen
```
sg_PanelBase.cpp:70:43: warning: ‘wxPen::wxPen(const wxColour&, int, int)’ is deprecated: use wxPENSTYLE_XXX constants [-Wdeprecated-declarations]
   70 |     wxPen bgPen(*wxWHITE, 1, wxTRANSPARENT);

```

4. Constructor for wxBrush
```
sg_PanelBase.cpp:174:49: warning: ‘wxBrush::wxBrush(const wxColour&, int)’ is deprecated: use wxBRUSHSTYLE_XXX constants [-Wdeprecated-declarations]
  174 |     wxBrush bgBrush(*wxLIGHT_GREY, wxTRANSPARENT);
```

5. SetWeight for wxFontBase
```
ViewStatistics.cpp:1151:34: warning: ‘void wxFontBase::SetWeight(wxDeprecatedGUIConstants)’ is deprecated: use wxFONTWEIGHT_XXX constants instead of wxLIGHT/wxNORMAL/wxBOLD [-Wdeprecated-declarations]
 1151 |         m_font_standart.SetWeight(wxNORMAL);
```

### Describe the solution you'd like

These should be updated to the current code.  However, I am not sure if it is as simple as refactoring some code, see below.

### Additional context

Updating these deprecations for the current version of wxWidgets shouldn't be a problem, except for the wxTimerEvent.  I haven't looked at that one in detail yet but it may be a separate issue.  However, do we need to keep BOINC compatible with older versions of wxWidgets?  If so, then this may get trickier.  the changes for the wxFont constructor were added in wxWidgets 3.1.1.  If someone is building the Manager for an older OS that needs to use an older version of wxWidgets, (possibly due to a wxGTK version), then this could be trickier.

I'm not sure if the appropriate fix would be to just fix it, add preprocessor headers for each constructor to make it compatible with older versions (which seems like a lot of work and difficult to cover all possibilities), or something else.

Any input is appreciated.  I would be happy to fix these deprecations as well.

Vulpine05:
Fixes #6098 (mostly)

**Description of the Change**
wxFont, wxPen, and wxBrush has deprecations for formatting, such as the font's family name, weight, style.  This PR updates these.  For example, wxSWISS is updated to wxFONTFAMILY_SWISS.

Please note that #6098 also mentions a deprecation with wxTimerEvent.  I believe fixing this deprecation is different than the styling, and has the potential to affect performance if not addressed correctly.  Therefore, I feel that should be a separate PR.

**Alternate Designs**
N/A

**Release Notes**
N/A


Vulpine05:
@CharlieFenton, there was one change to the Mac Bitmap Combo Box file.  Would you be able to take a look?

AenBleidd (approved):

CharlieFenton:
> there was one change to the Mac Bitmap Combo Box file. Would you be able to take a look?

I think that should not be a problem. Since that is only on the Mac, I'll have to test the Simple View on the build artifact to be sure. I'll do that later tonight and let uyou know how it looks.

CharlieFenton:
@Vulpine05 Your changes all seem fine. However, I noticed a problem with drawing the buttons which is unrelated to your changes. On the Mac, the buttons _Task Commands_, _Add Project_ (or _Synchronize_ if using an account manager), _Project Web Pages_ and _Project Commands_ all have gray rectangles around them as shown in this screenshot:
bad_buttons

They should look like this:
good_buttons

As best as I can tell, this seems to be caused by a change in the Xcode compiler since I built BOINC 7.24.1. If I  install the release version of BOINC 7.24.1, it does not have the issue, but if I recompile it with the current compiler the issue does appear.

Do you see this issue on Linux or Windows?

To fix this, please change the following code in _clientgui/sg_CustomControls.cpp_ lines 112 - 116 from this:
```
    bool bRetVal = wxButton::Create(parent, id, label, pos, size, style|wxTRANSPARENT_WINDOW, validator, name);

    SetBackgroundStyle(wxBG_STYLE_CUSTOM);
    SetBackgroundColour(parent->GetBackgroundColour());
    SetForegroundColour(parent->GetForegroundColour());
```
to this:
```
    SetBackgroundStyle(wxBG_STYLE_TRANSPARENT);
    bool bRetVal = wxButton::Create(parent, id, label, pos, size, style|wxTRANSPARENT_WINDOW, validator, name);

    SetBackgroundColour(parent->GetBackgroundColour());
    SetForegroundColour(parent->GetForegroundColour());
```

Note that `SetBackgroundStyle(wxBG_STYLE_TRANSPARENT);` must come before the `Create()` cal, as explained [here](https://docs.wxwidgets.org/3.2.6/classwx_window.html#af14f8fd2ed2d30a9bbb5d4f9fd6594ec).

If this change causes problems on Linux or Windows, you can guard the changes with 
```
#ifdef__WXMAC__

#endif
```
or 
```
#ifndef__WXMAC__

#endif

```

Thank  you!

CharlieFenton:
@Vulpine05 Please see also my PR #6109

Vulpine05:
@CharlieFenton, glad the refactor in the Mac file works.

I looked at the grey box in Windows, and I do not see it.  I will have to check Linux but I have not had an opportunity today.  I will have to explore that during the week sometime.

As for the code change, I don't mind adding it to this PR, but I think it is better suited as a separate PR.  I keep finding a bunch of little graphic imperfections (see some of the issues I have posted).  Let me know how you feel about that.

AenBleidd:
> As for the code change, I don't mind adding it to this PR, but I think it is better suited as a separate PR. 

Totally agree: that change is not a part of this deprecation refactoring but rather a new issue fix, so it should definitely go to the separate PR.

PR #6110, opened 2025-02-23
Merge the MRU computers list when saving

ChihweiLHBird:
This is to fix an issue causing duplicated or lost of MRU computers when multiple windows (instances) of BOINC manager are saving the MRU list into the state.

TODO before making this PR ready:
1. More details of description
2. Explore required items (e.g., tests) for the PR

### DIFF
--- a/clientgui/BOINCBaseFrame.cpp
+++ b/clientgui/BOINCBaseFrame.cpp
@@ -730,8 +730,9 @@ bool CBOINCBaseFrame::SaveState() {
     wxString        strConfigLocation;
     wxString        strPreviousLocation;
     wxString        strBuffer;
-    int             iIndex;
-    int             iItemCount;
+    size_t          iIndex;
+    size_t          iItemCount;
+    wxArrayString   existingComputers;
 
 
     // An odd case happens every once and awhile where wxWidgets looses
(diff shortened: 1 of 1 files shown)


ChihweiLHBird:
Hi @AenBleidd, any idea about adding tests for this change? I took a look but didn't find a good point for adding test for this functionality of BOINC Manager.

AenBleidd:
We don't have tests for that, so I'll test that manually.

AenBleidd:
Ok, looks like this doesn't work as intended.
Here I have tested in on the local installation.
1. Open BOINC Manager
2. Open second instance of BOINC Manager (e.g. via boincmgr.exe -m) and connect to the same machine (e.g. setting `localhost`)
3. Close second instance and the first instance
4. Open regedit.exe
5. Navigate to `Computer\HKEY_CURRENT_USER\Software\Space Sciences Laboratory, U.C. Berkeley\BOINC Manager\ComputerMRU`
6. Add duplicate entry (In my case I had `0: localhost` and I added `1: localhost`)
7. Open BOINC Manager
8. Close BOINC Manager.
9. Check that two duplicate values are still there (should be only one)

ChihweiLHBird:
@AenBleidd thanks for testing it!

I just realized the current purposed change only prevents new duplicated items from being set, but it won't clean up existing ones.

I will make a change to support the cleaning up, maybe later this week.

ChihweiLHBird:
Hi @AenBleidd, I think de-duplicating the computer list when loading the state would solve this, and it works in my local experiment.

ChihweiLHBird:
Couldn't understand why the check failed... is it a bug in the check itself?

AenBleidd (approved):

Issue #6015, opened 2025-01-19
All BOINC tasks using vboxwrapper 26208 on MacOS fail


computezrmle:
### Describe the bug

LHC@home published new vbox apps using vboxwrapper 26208 a few days ago.
On Windows and Linux they work fine but on MacOS all tasks fail:

```
<core_client_version>8.0.2</core_client_version>
<![CDATA[
<message>
process exited with code 1 (0x1, -255)</message>
<stderr_txt>
2025-01-19 13:24:28 (1952): vboxwrapper version 26208
2025-01-19 13:24:28 (1952): BOINC client version: 8.0.2
2025-01-19 13:24:28 (1952): Detected: VirtualBox VboxManage Interface (Version: 7.1.4)
2025-01-19 13:24:28 (1952): Detected: Sandbox Configuration Enabled
2025-01-19 13:24:28 (1952): WARNING: Communication with VM Hypervisor failed.
2025-01-19 13:24:28 (1952): ERROR: VBoxManage list hostinfo failed
2025-01-19 13:24:28 (1952): called boinc_finish(1)

</stderr_txt>
]]>
```

A volunteer from LHC@home running an Apple computer commented this:
https://lhcathome.cern.ch/lhcathome/forum_thread.php?id=6277&postid=51425
```
The biggest change between the vboxwrapper v26207 and v26208 is the location for the virtual machines. This moved from /Library/Application Data/Boinc/Projects/VirtualBox to ~/Library/VirtualBox

Apparently, user boinc_manager is not allowed to access the ~ folder... I think the data should be moved to the /Users/Shared directory instead of the ~ directory.


```
The related changes have been introduced in #5658.

Please verify if these changes are the root cause for the problems and if so provide a solution for computers running MacOS.

AenBleidd:
@CharlieFenton, @davidpanderson, could you please take a look at this issue?

CharlieFenton:
> Apparently, user boinc_manager is not allowed to access the ~ folder

That is correct. For security reasons, BOINC cannot access non-BOINC user data as described [here](https://github.com/BOINC/boinc/wiki/SandboxUser). I suggest putting it in "_/Library/Application Support/BOINC Data/_" (note: **not** _~/Libary/,,,_. I'm not certain this will work without additional changes to the BOINC installer, but I believe it will.

CharlieFenton:
> For security reasons, BOINC cannot access non-BOINC user data

[...]

PR #6116, opened 2025-02-25
setup vbox profile location

computezrmle:
Recently there were a lot of comments and suggestions regarding where to locate the VirtualBox configuration files.

See:
#6015
#6016
#6018

This PR is intended to combine them into a working solution on Linux/MacOS/Windows.

Major objectives:
1. use VirtualBox's default file locations where possible
2. use a well defined file location if (1.) is not possible

(1.) mitigates a possible race between VirtualBox Manager (launched by the user) and BOINC VMs (launched via VBoxHeadless).
Since the first VirtualBox process that starts up for a distinct user determines which vbox profile is used, all of them should use the same vbox profile.
(1.) also ensures that an interactive user can monitor the running VMs, e.g. on Windows.

(2.) must be used in environments where BOINC has not enough permissions to access the default vbox profile location, e.g. from inside MacOS's sandbox.


Crystal-Pellet:
Tested the **Windows** version of this PR and concluded that VirtualBox is using again the default Oracle VirtualBox location for its xml- and log-files.
Unnecessary .VirtualBox and/or VirtualBox directories are not longer created in BOINC's datadirectory.

So OK to this version of vboxwrapper.

davidpanderson:
Can you please also add comments (in vbox_common.h) for these fields:
    string virtualbox_home_directory;
    string virtualbox_scratch_directory;
    string virtualbox_install_directory;
    string virtualbox_guest_additions;

... saying what they're used for and how they're determined.

Also: I don't understand why these are elements of VBOX_BASE and not global vars.
(more generally I don't understand why we need VBOX_JOB, VBOX_VM, and VBOX_BASE at all.
But that's beyond the scope of this).

computezrmle:
> Can you please also add comments (in vbox_common.h) for these fields:
> string virtualbox_home_directory;
> string virtualbox_scratch_directory;
> string virtualbox_install_directory;
> string virtualbox_guest_additions;

As for `virtualbox_scratch_directory`
This is unused, hence I removed it.

Added comments for the others.
These variables have been introduced more than a decade ago together with comments explaining what they are used for.
Over the years the original comments obviously got lost.

VBOX_JOB, VBOX_VM, and VBOX_BASE are also used for more than a decade.
Changing this should indeed not be the scope of this PR.

lfield:
LGTM I endorse the test from Crystal-Pellet

AenBleidd:
@davidpanderson, could you please review this again? Thank you in advance.

davidpanderson:
The added comments are good, but they should say
- why does vboxwrapper need to know where the vbox profile directory is?
- Does this directory always exist already?  In what situations does vboxwrapper have to create it?  Why?
- If vboxwrapper needs to create it, where does it put it, and what's it called?

vboxwrapper may not be able to create the profile directory in the user's home directory.
All platforms have a 'sandboxed' mode in which apps (including vboxwrapper)
run under a user account that can't access anything outside the BOINC data dir.
On Mac it can't even write in that directory - only in subdirectories.
That's why we were putting the profile dir in <datadir>/projects/.

Whatever the change is, it needs to be tested on
- Windows: secure install
- Linux: secure install (I believe our .rpm does this by default)
- Mac: standard install (which is secure)

computezrmle:
> The added comments are good, but they should say
> 
> - why does vboxwrapper need to know where the vbox profile directory is?
> - Does this directory always exist already? In what situations does vboxwrapper have to create it? Why?
> - If vboxwrapper needs to create it, where does it put it, and what's it called?

I think the comments are detailed enough.
Especially since they already include a link to the VirtualBox technical documentation.
https://github.com/BOINC/boinc/blob/4216226b750c8ae9b9e6bcfffb2987f3ba99faa4/samples/vboxwrapper/vbox_vboxmanage.cpp#L109

> vboxwrapper may not be able to create the profile directory in the user's home directory.

Vboxwrapper runs under that username, hence can write to it's own home directory.

In a sandbox environment the profile now goes to `/projects/VirtualBox`.
This was the agreement for MacOS as far as I understand the PRs mentioned above.
I don't see a reason why it should be different on other platforms.

> Whatever the change is, it needs to be tested on
>
> - Windows: secure install
> - Linux: secure install (I believe our .rpm does this by default)
> - Mac: standard install (which is secure)

It has been tested on all platforms with real tasks from LHC@home.

davidpanderson:
Has it been tested with secure installs on Win and Linux?

computezrmle:
Depends on what you exactly mean by 'secure installs'.
Is it (on Linux) the installation as a service without a home directory?
If so, this works.
I didn't test it with this artifact but with the vboxwrapper before this PR which writes it's profile to `/projects`.

On Windows I don't know whether this has explicitly been tested.
I don't see a reason why it shouldn't work since it uses the same non-default location that is used on Linux and for the MacOS sandbox.

Crystal-Pellet:
> In the Windows installer there's a secure install option. That creates an unprivileged user (without a home directory, I believe) and BOINC jobs run as that user

If you mean 'Service install' with the secure option:
There is a boinc_master user created under the C:\Users and BOINC is running with the account of boinc_master.
Although a .VirtualBox directory is created under C:\Users\boinc_master's directory, VirtualBox cannot be used because of Access Denied.
```
2025-03-04 11:52:41 (10996): 
Command: VBoxManage -q list systemproperties
Exit Code: -2147024891
Output:
VBoxManage.exe: error: Failed to create the VirtualBox object!
VBoxManage.exe: error: The object is not ready
VBoxManage.exe: error: Details: code E_ACCESSDENIED (0x80070005), component VirtualBoxClientWrap, interface IVirtualBoxClient
```
It's just like with GPU-tasks, don't use BOINC's service install when you plan to run VBox-tasks.

computezrmle:
I wonder if `boinc_master` has enough access rights (via group membership) to meet the requirements mentioned in the VirtualBox manual.
If not it has to be solved in the installer rather than in vboxwrapper.

https://www.virtualbox.org/manual/topics/installation.html#install-win-installdir-req
```
Windows Installation Directory Security Requirements

The installation directory on Windows hosts must meet certain security requirements, in order to be accepted by the Windows installer.

This also applies for upgrades of Oracle VirtualBox.
For example, when installing Oracle VirtualBox into a custom location at X:\Data\MyPrograms\Oracle VirtualBox, all parent directories of this path (namely X:\Data and X:\Data\MyPrograms) must meet the following Discretionary Access Control List (DACL).

        Users               S-1-5-32-545:(OI)(CI)(RX)
        Users               S-1-5-32-545:(DE,WD,AD,WEA,WA)
        Authenticated Users S-1-5-11:(OI)(CI)(RX)
        Authenticated Users S-1-5-11:(DE,WD,AD,WEA,WA)
      

Directory inheritance must also be disabled for all parent directories.
You can use the icacls Windows command line tool to modify a directory to meet the security requirements. For example:

      icacls <Directory> /reset /t /c
      icacls <Directory> /inheritance:d /t /c
      icacls <Directory> /grant *S-1-5-32-545:(OI)(CI)(RX)
      icacls <Directory> /deny  *S-1-5-32-545:(DE,WD,AD,WEA,WA)
      icacls <Directory> /grant *S-1-5-11:(OI)(CI)(RX)
      icacls <Directory> /deny  *S-1-5-11:(DE,WD,AD,WEA,WA)
      

Note that these commands must be repeated for all parent directories (X:\Data and X:\Data\MyPrograms in this example).
```

AenBleidd:
I agree this should be solved separately in the installer.
Please open a separate ticket for that.

computezrmle:
Anything else that needs to be done?
Crystal-Pellet's last test shows that it works as intended in connection with with a secure install.
The fact that VirtualBox finally fails is not caused by vboxwrapper but most likely by a permissions issue.
```
There is a boinc_master user created under the C:\Users
(...) BOINC is running with the account of boinc_master.
(...) a .VirtualBox directory is created under C:\Users\boinc_master's directory
```

computezrmle:
@davidpanderson Please review again.
@AenBleidd After approval by reviewers please publish an official revision.

Thanks in advance.

davidpanderson:
The comments are not complete, but as long as it works, fine w/ me.

PR #6131, opened 2025-03-02
[WIP] Mac: Fix bugs in Mac installer

CharlieFenton:
This PR fixes several bugs in the Mac installe. While I'm not sure, I suspect they might have been caused by changes in behavior of the MacOS. Among the issues fixed:
 - The BOINC Manager was being launched prematurely,before our install had completed setting things up.
 -  The installer opened BOINC Manager in a strange state,  especially if it was set to open in Simple View (which is the default for a virgin install), causing missing menus and the windows to not be shown.
 -  After a log out, the next time the user logged in, the BOINC Manager was launched in the foreground instead of hidden.

### DIFF
--- a/clientgui/BOINCGUIApp.cpp
+++ b/clientgui/BOINCGUIApp.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
 // http://boinc.berkeley.edu
-// Copyright (C) 2024 University of California
+// Copyright (C) 2025 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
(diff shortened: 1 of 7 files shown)


clientgui/BOINCGUIApp.cpp
```diff
@@ -69,33 +69,7 @@ BEGIN_EVENT_TABLE (CBOINCGUIApp, wxApp)
 #endif
 END_EVENT_TABLE ()
 
-#if defined(__WXGTK__) && defined(BUILD_WITH_VCPKG)
```
AenBleidd, line 72: Please keep the next two blocks. They are very important for the Linux build

CharlieFenton:
@AenBleidd hank you for noticing this. I don't know how it got removed. I found another change I had not intended. I have now retired them.

AenBleidd (approved):

PR #6147, opened 2025-03-11
docker_wrapper: support HTML graphics

### LINKED ISSUE #6146: Dynamically create host port and enable show graphics
### Describe the problem

For the docker wrapper, to enable the show graphics button to work, the following code would be needed in the docker_wrapper to define a host port if not specified. 

https://github.com/BOINC/boinc/blob/master/samples/vboxwrapper/vbox_vboxmanage.cpp#L850

In the .toml file, only the guest port should need to be specified. The docker wrapper should set this correctly for the container and enable the graphics button.

davidpanderson:
Add 'web_graphics_guest_port' option to job.toml config file. If set:
- allocate a host port
- tell Docker to map the guest port to host port
- tell the BOINC client to get HTML graphics at http:localhost:host_port

Fixes #6146

### DIFF
--- a/samples/docker_wrapper/Makefile
+++ b/samples/docker_wrapper/Makefile
@@ -31,6 +31,11 @@ ifdef BUILD_APPS_WITH_VCPKG
     -L$(VCPKG_DIR)/lib
 endif
 
+ifdef BUILD_WITH_MINGW
+  MAKEFILE_LDFLAGS += \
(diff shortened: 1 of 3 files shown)


AenBleidd (approved):

lfield:
I know that this has already been merged but I just wanted to add that I have tested this and it works.

PR #6149, opened 2025-03-12
web (forums): fix XSS vulnerability

davidpanderson:
Check the contents of [email]...[/email] with a regex (like we do for [url])


lfield:
LGTM

PR #6154, opened 2025-03-13
client, Linux: don't check for podman

davidpanderson:
podman build with a remote FS doesn't seem to work. For now, use docker instead of podman.


AenBleidd:
While this is ok to me, it would be nice to check that podman works with the remote FS on MacOS and Windows/WSL.
Or remote FS is never the case for any other OS except Linux?

AenBleidd (approved):

PR #6164, opened 2025-03-17
web: fix XSS vulnerability in submit RPC handler

davidpanderson:


AenBleidd:
I will restart this build after the maintenance is finished and merge it afterwards if this will be reviewed and tested before.
Otherwise I'll wait for the review to be done.

lfield:
LGTM

AenBleidd (approved):

PR #6165, opened 2025-03-17
fix DB queries

### LINKED ISSUE #6160: Ops page DB errors
### Describe the bug

We encountered a few issues on the ops pages due to a Mysql DB upgrade from 8.0.35 to 8.4.2.

1.  The use of DESC in [pass_percentage_by_platform.php](https://github.com/BOINC/boinc/blob/master/html/ops/pass_percentage_by_platform.php#L109) results in it being used twice in the query causing it to fail. 
2. `ANY_VALUE(app_version_num)`  needs to be in [failure_result_summary_by_host.php#L61](https://github.com/BOINC/boinc/blob/master/html/ops/failure_result_summary_by_host.php#L61) due to MySQL's ONLY_FULL_GROUP_BY mode.
3. `host.os_name` should be added to the GROUP  BY in  [failure_result_summary_by_platform.php#L53](https://github.com/BOINC/boinc/blob/master/html/ops/failure_result_summary_by_platform.php#L53) for the same reason.

davidpanderson:
- don't fetch unused stuff
- use get_int() instead of $_GET[]
- don't put quotes around ints in queries
- don't put DESC in GROUP_BY

Fixes #6160

### DIFF
--- a/html/ops/failure_result_summary_by_host.php
+++ b/html/ops/failure_result_summary_by_host.php
@@ -21,23 +21,13 @@ require_once("../inc/util_ops.inc");
 db_init();
 admin_page_head("Failures grouped by app version and host");
 
-$query_appid = $_GET['appid'];
-$query_received_time = time() - $_GET['nsecs'];
+$query_appid = get_int('appid');
+$query_received_time = time() - get_int('nsecs');
 
 $main_query = "
 SELECT
        app_version_id,
-       app_version_num,
-       hostid AS Host_ID,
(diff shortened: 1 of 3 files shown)


AenBleidd:
I did force-push to remove unrelated commit.

I will restart this build after the maintenance is finished and merge it afterwards if this will be reviewed and tested before.
Otherwise I'll wait for the review to be done.

lfield:
LGTM

PR #6166, opened 2025-03-17
client (Win): show Docker detection details

davidpanderson:


AenBleidd:
I will restart this build after the maintenance is finished and merge it afterwards.

PR #6183, opened 2025-03-22
client (Linux): parse podman version if using podman-docker

### LINKED ISSUE #6141: docker command detected as unknown
### Describe the bug

I am using the latest code from master. 

The client detects that I have podman. Here is the entry from the event log.

Fri 07 Mar 2025 13:23:15 CET |  | Docker: version 5.0.3 (podman)

However, when the docker_wrapper runs it uses the command 'unknown'

https://lhcathomedev.cern.ch/lhcathome-dev/result.php?resultid=3387274

I can work around this by setting a symlink. 

ln -s /usr/bin/podman /usr/bin/unknown

davidpanderson:
... which makes docker an alias of podman.
Note: there should be no reason to use podman-docker now that we allow podman if data dir is not remote.

Fixes #6141

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1242,12 +1242,17 @@ bool HOST_INFO::get_docker_version_aux(DOCKER_TYPE type){
     FILE* f = popen(cmd.c_str(), "r");
     if (f) {
         char buf[256];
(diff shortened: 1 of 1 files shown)


lfield:
LGTM

AenBleidd (approved):

PR #6199, opened 2025-03-26
Ignore MacOs specific folders and files during trailing whitespace check

AenBleidd:


AenBleidd:
@CharlieFenton, please test this.
This will not remove all the findings since some of them are not related to the BOINC repo at all (looks like some custom files that you have in the repo folder) but will make this list significantly shorter.

AenBleidd:
@CharlieFenton, also I'm not completely sure about the '.DS_Store': I don't know if this is treated as an extension or the filename.
If the script will still report this file/folder - please let me know and I'll fix the script.
Thank you in advance.

CharlieFenton:
MacOS creates a .DS_Store in most directories. It should be ignored by the trailing whitespace check. .DS_Store is a hidden file; the leading period in a file or directory name makes it hidden. The trailing whitespace check should ignore all such hidden files (any file whose name begins with a period.)

CharlieFenton:
It is still reporting _.DS_Store_, as well as the compiler generated files _MultiGPUMigUser.c_, _MultiGPUMigServer.h_ and _MultiGPUMig.h_. While I suspect removing trailing whitespace from these files won't cause problems, I am not sure. So It would be better for the trailing whitespace check to ignore them. I have attached the output from this latest version.

[whitespaces_check_output2.txt](https://github.com/user-attachments/files/19479061/whitespaces_check_output2.txt)

AenBleidd:
@CharlieFenton, please check one more time.
I don't want to add this file to exclusions `/Volumes/Dev/BOINC_GIT/boinc_trunk copy/lib/gui_rpc_client-new.h:504` as this is really an unrelated file.
Everything else should now be skipped.

CharlieFenton:
It works as desired now. It did not report any inappropriate files. Thank you.

> I don't want to add this file to exclusions /Volumes/Dev/BOINC_GIT/boinc_trunk copy/lib/gui_rpc_client-new.h:504 as this is really an unrelated file.

You are correct. That is not part of the source tree and should not have been present.

PR #6201, opened 2025-03-27
client, Mac: detect podman

davidpanderson:
Also add project files for docker_wrapper on Mac


client/hostinfo_unix.cpp
```diff
@@ -1238,6 +1238,12 @@ int HOST_INFO::get_virtualbox_version() {
 //
 bool HOST_INFO::get_docker_version_aux(DOCKER_TYPE type){
     bool ret = false;
+#ifdef __apple__
```
AenBleidd, line 1241: Should be `__APPLE__` for consistency

client/hostinfo_unix.cpp
```diff
@@ -1238,6 +1238,12 @@ int HOST_INFO::get_virtualbox_version() {
...
+    if (type == PODMAN) {
```
AenBleidd, line 1242: What about Docker? Does it require similar command or works out-of-the-box?

AenBleidd (changes requested):

CharlieFenton (approved):

CharlieFenton:
I agree with the change from` __apple__` to `__APPLE__` requested by @AenBleidd. I don't know whether ` __apple__` is recognized by the Apple compiler.

AenBleidd (approved):

PR #6206, opened 2025-03-29
Mac: build docker_wrapper without providing a code signing identity

CharlieFenton:
Allow building docker_wrapper on the Mac without providing a code signing identity so no signing certificate is required.

Fixes #6205


CharlieFenton:
As explained [here](https://stackoverflow.com/questions/8136531/what-is-development-and-deployment-mode-in-xcode-active-configuration) Xcode changed the names it uses for the build configurations. Apparently Xcode continues to use the legacy names in the older projects, but cannot be told to use them with newer versions of Xcode.

I didn't think it was possible to change the names of the build configurations, but I found the way todo it [here](https://cocoacasts.com/tips-and-tricks-managing-build-configurations-in-xocde#:~:text=Select%20the%20scheme%20at%20the,Rename%20it%20to%20Staging.) .

This PR is now ready to merge.

AenBleidd (approved):

PR #6208, opened 2025-03-29
Enable docker wrapper to forward exit codes from a container to BOINC client

computezrmle:
**Current status**

Docker wrapper reports either '0' or '1' to BOINC client.
Status codes from inside a container are not forwarded.

**Expected**

In case of a container 'JOB_SUCCESS' get the status code from the main process running inside a container and forward that status code to BOINC client.

Tested with a task from lhcathomedev.cern.ch which was forced to report status 206.
https://lhcathomedev.cern.ch/lhcathome-dev/result.php?resultid=3392131


AenBleidd:

PR #6216, opened 2025-03-31
Add --passthrough_child arg to wrapper app

### LINKED ISSUE #6214: Wrapper app does not properly forward arguments to the child process
### Describe the bug

Wrapper app in samples/wrapper was advertised to passthrough arguments in argv to the child process that come after the arguments meant for use by the wrapper.

A commit added an "else" in the arguments logic, so anything that does not match what the wrapper consumes for itself will be rejected and the program will cease.

The workaround I had been utilizing up until now was to use something like `"--device \" \""` to trick the wrapper into accepting the arguments and proceeding, and everything gets passed to the child process.

This feels a bit janky, and it also doesn't remove the arguments before the arguments destined for use in the child process, so the child process must account for arguments destined for the wrapper, such as `--device` in that instance.

### Steps to reproduce

1. Call wrapper app with any arguments that should be sent to the child process which aren't valid arguments for the wrapper's consumption
2. Program will end

### Expected behavior

1. Call wrapper app with any arguments that should be sent to the child process which aren't valid arguments for the wrapper's consumption
2. Program will disregard all arguments at the end of argv that aren't meant for wrapper to use, and send them to the child process instead.

### Screenshots

_No response_

### System information

- OS: macOS, Linux, and Windows
- Arch: arm64 and amd64

### Additional context

I will have a PR to fix this today, I've already completed a code change that would resolve this but obviously it will need some review before merging into master.

It utilizes a new argument `--passthrough_child` which will send all arguments following it to the child process and stop processing further args after that.
If we update the wiki to reflect this, it should fix the problems I had with the wrapper.

BoySanic:
Fixes #6214

**Description of the Change**
Adds an argument flag `--passthrough_child` to the wrapper app.
When supplied, any arguments in argv that come after that flag will be disregarded by the wrapper app itself and instead will be passed through to the child process as the sole arguments given to it.

**Alternate Designs**
Alternative solutions may be to either:
1. Disable the logic that exits the program when it finds an invalid argument from the wrapper's perspective, since it should passthrough those arguments to the child.
2. Rely solely on job.xml for cmdline args

The reason I selected this implementation is because I utilize the cmdline args in jobsubmission to iterate through a range of variables and create the tasks for our project, and we can't easily replace job.xml for each individual task, so supplying these via jobsubmission made sense for us

**Release Notes**
N/A
For boinc admins, though, who may want to use this:
To specify args for the child process, supply `--passthrough_child`

Example:
`./wrapper --passthrough_child 1234567 input.npz 1500000 150`

Wrapper will ignore everything after `--passthrough_child` for its own uses and instead directly supply those arguments until the end of `argv` to the child process.

### DIFF
--- a/samples/wrapper/wrapper.cpp
+++ b/samples/wrapper/wrapper.cpp
@@ -1183,6 +1183,7 @@ void usage() {
         "   --trickle X\n"
         "   --version\n"
         "   --use_tstp\n"
+        "   --passthrough_child\n"
     );
     boinc_finish(EXIT_INIT_FAILURE);
 }
@@ -1196,7 +1197,9 @@ int main(int argc, char** argv) {
         // total CPU time at last checkpoint
     char buf[256];
     bool is_sporadic = false;
-
+    bool passthrough_child = false;
+    int child_arg_count = 0;
+    char** child_args;
     // Log banner
     //
     fprintf(stderr, "%s wrapper (%d.%d.%d): starting\n",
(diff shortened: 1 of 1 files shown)


davidpanderson:
Thanks.  I'll update the wiki.
After merging this I'll modify it a little to use vectors instead of malloc.

PR #6217, opened 2025-03-31
Wrapper: Replace malloc implementation with vectors, fix off-by-one e…

### LINKED ISSUE #6214: Wrapper app does not properly forward arguments to the child process

BoySanic:
Fixes #6214

**Description of the Change**
Mostly just replaced malloc'd c-style string arrays with vectors.
Also fixed off-by-one error I introduced which left `--passthrough_child` in `child_args`

**Alternate Designs**
Decided it was best to replace all instances of `char**` in reference to `argv` elsewhere in the program rather than build a `vector<string>` from a `char**` every time a `vector<string>` was needed.

I also avoided iterating twice to fill `wrapper_args` after already having iterated over argv.

A different, and maybe slightly smarter implementation might be to avoid looping a 2nd time for `child_args` and just avoid writing to `child_args` until we've seen `--passthrough-args` one time.

**Release Notes**
N/A

This is how #6216 should've been implemented to begin with.

### DIFF
--- a/samples/wrapper/wrapper.cpp
+++ b/samples/wrapper/wrapper.cpp
@@ -167,7 +167,7 @@ struct TASK {
     int parse(XML_PARSER&);
     void substitute_macros();
     bool poll(int& status);
-    int run(int argc, vector<string> argv);
+    int run(int argc, vector<string>& argv);
     void kill();
     void stop();
     void resume();
--- a/samples/wrapper/wrapper.cpp
+++ b/samples/wrapper/wrapper.cpp
@@ -596,7 +596,7 @@ int parse_job_file() {
     return ERR_XML_PARSE;
 }
 
-int start_daemons(int argc, vector<string> argv) {
+int start_daemons(int argc, vector<string>& argv) {
     for (TASK& task: daemons) {
         int retval = task.run(argc, argv);
         if (retval) return retval;
@@ -675,7 +675,7 @@ void backslash_to_slash(char* p) {
     }
 }
 
-int TASK::run(int argct, vector<string> argvt) {
+int TASK::run(int argct, vector<string>& argvt) {
     string stdout_path, stdin_path, stderr_path;
     char app_path[1024], buf[256];
 
--- a/samples/wrapper/wrapper.cpp
+++ b/samples/wrapper/wrapper.cpp
@@ -167,7 +167,7 @@ struct TASK {
     int parse(XML_PARSER&);
     void substitute_macros();
     bool poll(int& status);
-    int run(int argc, char** argv);
+    int run(int argc, vector<string> argv);
     void kill();
     void stop();
     void resume();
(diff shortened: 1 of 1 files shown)


BoySanic:
@AenBleidd @davidpanderson I appreciate your time in the #6216!

Unfortunately, I discovered I had an off-by-one which included `--passthrough_child` in `child_args`, so while I was fixing that, I figured I'd convert the relevant pieces to `vector<string>`.

Let me know if you have any questions or concerns about this.

AenBleidd:
I need to review this

samples/wrapper/wrapper.cpp
```diff
@@ -596,7 +596,7 @@ int parse_job_file() {
     return ERR_XML_PARSE;
 }
 
-int start_daemons(int argc, char** argv) {
+int start_daemons(int argc, vector<string> argv) {
```
AenBleidd, line 599: Don't pass vector by value, pass it by reference instead

samples/wrapper/wrapper.cpp
```diff
@@ -675,7 +675,7 @@ void backslash_to_slash(char* p) {
     }
 }
 
-int TASK::run(int argct, char** argvt) {
+int TASK::run(int argct, vector<string> argvt) {
```
AenBleidd, line 678: Don't pass vector by value, pass it by reference instead

AenBleidd (changes requested):

BoySanic:
I think David's version of this PR is more succinct and efficient. ( #6218)

We can close this as long as his is merged, I think.

Issue #2199, opened 2017-10-30
boinc_client gets blocked on XOpenDisplay () when 2 Xorg sessions are open


paulez:
When two graphic sessions are opened on my system with only one active, boinc_client gets eventually stuck forever waiting on XOpenDisplay(). Killing the non-active Xorg session or switching back to it will unstuck boinc_client.

```
% ps aux|grep boinc
boinc      616  0.1  0.1  76512 16276 ?        SNsl oct.25  11:27 /usr/bin/boinc_client --dir /var/lib/boinc --redirectio
boinc     6469 27.0  2.7 375236 227700 ?       SNl  oct.28 471:12 ../../projects/boinc.bakerlab.org_rosetta/minirosetta_3.78_x86_64-pc-linux-gnu @G187193_1-50_MAP_hyb_TMP_v02_i01_t000__krypton.flags -in:file:boinc_wu_zip G187193_1-50_MAP_hyb_TMP_v02_i01_t000__krypton.zip -run:protocol jd2_scripting -silent_gz -mute all -out:file:silent default.out -nstruct 10000 -cpu_run_time 28800 -boinc:max_nstruct 600 -checkpoint_interval 120 -database minirosetta_database -in::file::zip minirosetta_database.zip -boinc::watchdog -run::rng mt19937 -constant_seed -jran 1240652
boinc     8431 15.0  0.0      0     0 ?        ZN   oct.28 238:57 [minirosetta_3.7] <defunct>
boinc     9481 18.7  0.0      0     0 ?        ZN   oct.28 267:20 [wcgrid_mcm1_7.3] <defunct>
boinc     9770  7.4  0.0      0     0 ?        ZN   oct.28 102:17 [minirosetta_3.7] <defunct>
boinc     9822 18.9  0.0      0     0 ?        ZN   oct.28 257:39 [wcgrid_mcm1_7.3] <defunct>
boinc    14127  0.0  0.0      0     0 ?        ZN   oct.28   0:06 [primegrid_genef] <defunct>
boinc    14128 17.8  0.0      0     0 ?        ZN   oct.28 226:39 [wcgrid_mcm1_7.3] <defunct>
paul     21201  0.2  1.4 2969360 120644 ?      Sl   19:10   0:00 /usr/bin/boincmgr
paul     21295  0.0  0.0  11204   896 pts/1    S+   19:12   0:00 grep --color=auto boinc
```
[...]

PR #6236, opened 2025-04-07
client: Hook in reading of last_active_time from event_detect in hostinfo_unix

### LINKED ISSUE #1187: boinc_client does not notice (or does not care) that "computer is in use"

jamescowens:
hostinfo_unix.cpp, str_util.h, str_util.cpp

Fixes #1187, #2199, #2256, #3405, #5226

**Description of the Change**
This is a PR to introduce transitional code to make use of the new idle_detect service that I have been developing as a self contained idle detector. Please see https://github.com/jamescowens/idle_detect and in particular https://github.com/jamescowens/idle_detect/discussions/4.

What is meant by transitional is that user_idle_time() in hostinfo_unix.cpp first checks ReadLastActiveTimeDat() and if it is successfully read (which means that event_detect is present and running), uses the output to calculate idle time. If the file is not there, it falls back to the original code. Eventually this existing fallback code will be removed and/or simplified.

Note that the idle_detect/event_detect solution solves for 99% of the idle detection issues in Linux, including both X and Wayland KDE and Gnome sessions (including idle inhibit when watching a movie), and non KDE non Gnome sessions (X and Wayland) but without idle inhibit functionality. Additional monitor functions can be easily added to idle_detect to address Window manager specific APIs for idle inhibit.

**Alternate Designs**
The thread in #5226 in addition to the above discussion explains why this is the best approach to solve this problem.

**Release Notes**

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -22,6 +22,7 @@
 #include "version.h"         // version numbers from autoconf
 #include "cpp.h"
 #include "config.h"
+#include "str_util.h"
 
 #if !defined(_WIN32) || defined(__CYGWIN32__)
 
@@ -35,6 +36,7 @@
 #undef _LARGEFILE_SOURCE
 #undef _LARGEFILE64_SOURCE
 #include <iostream>
+#include <fstream>
 #include <vector>
 #include <string>
 #include <cstring>
@@ -914,7 +916,7 @@ static void get_cpu_info_mac(HOST_INFO& host) {
         FILE* fp = boinc_fopen(fpath, "r");
         if (fp) {
             fgets(features, sizeof(features), fp);
-	    feature_string += features;
+            feature_string += features;
             fclose(fp);
         } else if (log_flags.coproc_debug) {
             msg_printf(0, MSG_INFO, "[x86_64-M1] couldn't open file %s", fpath);
@@ -2139,10 +2141,10 @@ long xss_idle() {
         //
         if (disp == NULL) {
             if (log_flags.idle_detection_debug) {
-	            msg_printf(NULL, MSG_INFO,
-	                "[idle_detection] DISPLAY '%s' not found or insufficient access.",
-	                it->c_str()
-                );
+                msg_printf(NULL, MSG_INFO,
+                           "[idle_detection] DISPLAY '%s' not found or insufficient access.",
+                           it->c_str()
+                           );
             }
             continue;
         }
@@ -2155,10 +2157,10 @@ long xss_idle() {
             disp, &event_base_return, &error_base_return
         )){
             if (log_flags.idle_detection_debug) {
-	            msg_printf(NULL, MSG_INFO,
-	                "[idle_detection] XScreenSaver extension not available for DISPLAY '%s'.",
-	                it->c_str()
-                );
+                msg_printf(NULL, MSG_INFO,
+                           "[idle_detection] XScreenSaver extension not available for DISPLAY '%s'.",
+                           it->c_str()
+                           );
             }
             XCloseDisplay(disp);
             continue;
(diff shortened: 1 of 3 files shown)


client/hostinfo_unix.cpp
```diff
@@ -2203,35 +2205,97 @@ long xss_idle() {
...
+ * @return int64_t Timestamp read from file, or -1 if the file cannot be opened/read/parsed.
+ */
+static int64_t ReadLastActiveTimeFromFile() { // Takes no arguments
+    const char* file_path = "/run/event_detect/last_active_time.dat";
+
+    std::ifstream infile(file_path); // Open the specific file
+
+    if (!infile.is_open()) {
+        msg_printf(NULL, MSG_INTERNAL_ERROR, "%s: Could not open file %s", __func__, file_path);
+        return 0;
```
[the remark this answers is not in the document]
jamescowens: Resolved in latest force-push.

davidpanderson:
This PR is premature.
I agree that Linux idle detection should be done in a separate process.
How it communicates with the client hasn't been decided.
This can be discussed in #5226; there I proposed using a memory-mapped file.

AenBleidd:
@davidpanderson, I think memory-mapped file is more complicated than just reading a plain file on Linux.
What will be the benefit of using memory-mapped file?
From disadvantages I see more complicated process of synchronisation.

davidpanderson:
See comment in #5226

AenBleidd:
Ok. I'll take a look at the PR tonight when I come back home. If we say we need something like 'boinctray' to communicate with the new 'idle_detect' process via the file, and them communicate to the BOINC client via memory mapped file - then this is ok to me (I still want to review the PR first). However, I don't want to make the communication via the memory mapped file with the external application ('idle_detect' in this case).

lib/str_util.cpp
```diff
@@ -465,6 +465,17 @@ string timediff_format(double diff) {
 
 }
 
+[[nodiscard]] int64_t ParseStringtoInt64(const std::string& str)
```
AenBleidd, line 468: We try to keep libboinc compatible with the c++03 compilers, so please remove `nodiscard` here.

client/hostinfo_unix.cpp
```diff
@@ -2203,35 +2205,97 @@ long xss_idle() {
 }
 #endif // HAVE_XSS
 
+/**
```
AenBleidd, line 2208: We do not use Doxygen on the project, please convert this comments block to the regular comments

client/hostinfo_unix.cpp
```diff
@@ -2203,35 +2205,97 @@ long xss_idle() {
...
+ * using std::ifstream.
+ * @return int64_t Timestamp read from file, or -1 if the file cannot be opened/read/parsed.
+ */
+static int64_t ReadLastActiveTimeFromFile() { // Takes no arguments
+    const char* file_path = "/run/event_detect/last_active_time.dat";
+
+    std::ifstream infile(file_path); // Open the specific file
+
+    if (!infile.is_open()) {
+        msg_printf(NULL, MSG_INTERNAL_ERROR, "%s: Could not open file %s", __func__, file_path);
```
AenBleidd, line 2219: We don't use `__func__`, `__line__` and `__file__` macroses.
Please remove them.

client/hostinfo_unix.cpp
```diff
@@ -2203,35 +2205,97 @@ long xss_idle() {
...
+
+        strip_quotes(trimmed_line);
+        strip_whitespace(trimmed_line);
+
+        if (!trimmed_line.empty()) {
+            try {
+                timestamp = ParseStringtoInt64(trimmed_line);
+            } catch (const std::exception& e) {
+                msg_printf(NULL, MSG_INTERNAL_ERROR, "%s: Failed to parse timestamp line '%s' from %s: %s",
+                          __func__, trimmed_line.c_str(), file_path, e.what());
```
AenBleidd, line 2238: Here's the same.

client/hostinfo_unix.cpp
```diff
@@ -2203,35 +2205,97 @@ long xss_idle() {
...
+                timestamp = -1; // Reset on error
+            }
+        } else {
+            msg_printf(NULL, MSG_INTERNAL_ERROR, "%s: Read empty line after trimming from %s", __func__, file_path);
```
AenBleidd, line 2242: Here's the same.

client/hostinfo_unix.cpp
```diff
@@ -2203,35 +2205,97 @@ long xss_idle() {
...
+            timestamp = -1;
+        }
+    } else {
+        if (infile.eof()) {
+            msg_printf(NULL, MSG_INTERNAL_ERROR, "%s: Failed to read line (EOF - file empty?) from %s", __func__, file_path);
```
AenBleidd, line 2247: Here's the same.

client/hostinfo_unix.cpp
```diff
@@ -2203,35 +2205,97 @@ long xss_idle() {
...
+        } else {
+            msg_printf(NULL, MSG_INTERNAL_ERROR, "%s: Failed to read line from file %s (stream error?)", __func__, file_path);
```
AenBleidd, line 2249: Here's the same.

AenBleidd (changes requested):

lib/str_util.cpp
jamescowens: Roger.

client/hostinfo_unix.cpp
jamescowens: Done.

client/hostinfo_unix.cpp
jamescowens: Done.

client/hostinfo_unix.cpp
jamescowens: Done.

client/hostinfo_unix.cpp
jamescowens: Done.

jamescowens:
@AenBleidd I think I addressed all of your comments. Thanks.

client/hostinfo_unix.cpp
jamescowens: Done.

jamescowens:
@davidpanderson I made some detailed comments over on #5226.

AenBleidd (approved):

AenBleidd (approved):

AenBleidd (approved):

jamescowens:
David, Vitalii, and I agreed that we would use Linux shmem to share the last_active_time.dat from event_detect to the boinc-client instead of a file in /run/event_detect. I have made the appropriate modifications to the idle_detect/event_detect side, and David is going to put up a new PR to do the BOINC side, so I am closing this one.

PR #6245, opened 2025-04-12
Fix snprintf() errors introduced in PR #4864

davidpanderson:
PR #4864 introduced 4 erroneous changes of the form
   sprintf(buf, sizeof(buf), ....)
where buf is a char*.
(Note: in this case sizeof(buf) is 8.
It is NOT the size of the memory area that buf points to).

One of these (in file_names.cpp) was fixed in a later PR. This PR fixes the other 3.
Hopefully we didn't miss any.

### DIFF
--- a/client/mac_address.cpp
+++ b/client/mac_address.cpp
@@ -135,7 +135,7 @@ GetMACAddress(io_iterator_t intfIterator, char* buffer)
                 UInt8 MACAddress[ kIOEthernetAddressSize ];
 
                 CFDataGetBytes(refData, CFRangeMake(0,CFDataGetLength(refData)), MACAddress);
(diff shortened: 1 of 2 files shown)


AenBleidd:
@davidpanderson, I agree with Copilot here: it's safer to use snprintf instead of sprintf.

davidpanderson:
Yeah.  But we'd need the size of the buffer, passed in as an additional arg.

AenBleidd:
@CharlieFenton, please review

davidpanderson:
It turns out that
* other sprintf()s were added after the original PR
* get_mac_address() shouldn't exist; we don't actually need MAC addresses.

So (if OK with Charlie) let's merge this PR and I'll deal with the above later.

CharlieFenton (approved):

PR #6260, opened 2025-04-18
fix compile warnings in lib/

davidpanderson:
In lib/, fix warnings with Linux g++ version 11.4.0, using the compiler options listed here:
https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html

(except for the RSA deprecation warnings, which appear even with -Wno-deprecated)

Note: it's pretty much all type casting.
In many cases this could also be eliminated by using the right types in the first place,
e.g. size_t instead of int in various places.

### DIFF
--- a/client/Makefile.linux
+++ b/client/Makefile.linux
@@ -6,8 +6,19 @@
 #    make -f Makefile.linux clean all
 # 3) do the same in this dir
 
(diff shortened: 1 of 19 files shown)


AenBleidd:
@davidpanderson, ssize_t is not from C standard but from the POSIX. Visual Studio is not 100% POSIX-compatible

PR #6274, opened 2025-04-22
library interfaces: remove dependency on config.h

### LINKED ISSUE #6271: Installed header str_replace.h includes config.h, which should never be installed
In #4063 config.h was removed from installable headers for good reason given in related issues mentioned there. In short: (our) Apps are built from multiple libraries, that mostly use autoconf (or some other way to produce a config.h) to match the library code with a given OS and build system. BOINC (boinc_api for building apps or boinc_server for building project daemons) is only one of these. Being installed in a a common place a config.h of one library confuses the build process of other libraries and codes built afterwards. Therefore, config.h should never be installed.

However, current API code required str_replace.h, which currently includes config.h, therefore config.h would need to be installed as well - but shouldn't. Pleas fix str_replace.h to not include config.h, at least not for API builds.

davidpanderson:
We need to be able to compile with libboinc and libboinc_api without needing config.h.

To do this: avoid having .h files include other .h files
(in particular, avoid chains of includes that lead to str_replace.h,
which is what needs config.h).

Fixes #6271

### DIFF
--- a/client/client_msgs.h
+++ b/client/client_msgs.h
@@ -45,7 +45,7 @@ struct MESSAGE_DESC {
 struct MESSAGE_DESCS {
     std::deque<MESSAGE_DESC*> msgs;
     void insert(PROJ_AM *p, int priority, int now, char* msg);
-    void write(int seqno, class MIOFILE&, bool translatable);
+    void write(int seqno, MIOFILE&, bool translatable);
     int highest_seqno();
     void cleanup();
 };
--- a/client/hostinfo_linux.cpp
+++ b/client/hostinfo_linux.cpp
@@ -33,6 +33,7 @@
 #endif
 
 #include "str_replace.h"
+#include "str_util.h"
 #include "util.h"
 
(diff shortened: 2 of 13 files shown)


AenBleidd:
I keep this open for a while to make a full test with building `libboinc` independently and linking to the test code.
Should take me a couple of days to do that.

davidpanderson:
BTW, the test is: compile an app that uses libboinc_api and libboinc
(and therefore includes api/boinc_api.h and e.g. lib/util.h)
without config.h anywhere.
I did this using samples/example_app/ (by renaming boinc/config.h)
on Win and Linux and it worked.

AenBleidd:
That's similar to what I plan to do, with the only different that in my case I want to have these builds separated, and the app should not have physical access to the 'config.h' file to be sure that there is no other hidden place.
And I want to make it another automated test to prevent such cases in the future.
Anyway, thank you for the fix, @davidpanderson 👍

PR #6306, opened 2025-05-06
[web] Ensure global $config is only ever written to by a single entity

brevilo:
Ensure `global $config` is only ever written to by a single entity (currently `util.inc`)

### DIFF
--- a/html/inc/prefs.inc
+++ b/html/inc/prefs.inc
@@ -288,7 +288,6 @@ global $venue_name;
 // in config.xml so the right default is set for minimum free disk space
 //
 function get_disk_space_config() {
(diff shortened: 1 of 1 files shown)


tristanolive:
Agreed, it would be cleaner if Drupal didn't depend on those globals--and as we've seen with `$config`, it could prevent future compatibility issues. 

As for the change to `get_disk_space_config()` in prefs.inc, that looks like a good move. I'm not sure the global from util.inc would necessarily even be initialized, so this prevents any confusion.

davidpanderson:
I'd like to get rid of global $config, and most other globals.
I can make a PR for this.

PR #6308, opened 2025-05-07
web: Remove some globals.  No functional change.

davidpanderson:
Remove globals $master_url, $config, $g_config,
    $recaptcha_public_key, and $recaptcha_private_key.
Add functions that return these.

There are still some globals but they're within one file. We can remove them later.

Fix small bug in batch display.


AenBleidd (approved):

tristanolive:
This works great to prevent any possibility of collisions in the global space with Drupal or other such integrations, thanks for this update!

PR #6311, opened 2025-05-08
client and docker_wrapper (Mac/podman): set env vars to use accessable dir

### LINKED ISSUE #6302: Enabling Podman on BOINC for Mac
By default, Podman stores its files and directories in hidden subdirectories under the user's home folder. But BOINC's user account-based sandbox security prevents BOINC from accessing the user's home folder, in order to protect the user's data. This prevents Podman from working under BOINC.

Fortunately, Podman checks two environment variables _XDG_CONFIG_HOME_ and _XDG_DATA_HOME_. These can be set to paths which Podman will use instead of the user's home directory. So the key to allowing BOINC to use Podman on the Mac is to set these two environment variables to values accessible by BOINC's special users _boinc_master_ and _boinc_project_.

However, environment variables normally have a very short life span and so must be set for every shell command or script, unless they can be set globally. 

I have identified 3 potential approaches for implementing these environment variables for BOINC on the Mac. All 3 include having the Mac BOINC installer create a subdirectory named _podman_ inside the BOINC data directory, with user _boinc_master_, group _boinc_project_ and permissions _drwxrwxr--_

**[1]** Add the following to the file _/etc/zprofile_:
  ```
XDG_CONFIG_HOME="/Library/Application Support/BOINC Data/podman"
 export XDG_CONFIG_HOME
 XDG_DATA_HOME ="/Library/Application Support/BOINC Data/podman"
 export XDG_DATA_HOME
```
**Pros:** 
 - Very simple to implement
 - Requires no changes to any BOINC client or library code or to any project code.

**Cons:**
 - Won't work if ant BOINC user has not set zsh as their default shell
 - Modifies a basic operating system parameter at a very low level
 - Affects users who use Podman outside of BOINC
   - Podman will no longer work for users who are not authorized to use BOINC
      - Example: a high school or college computer lab where students are learning to use virtualization, and BOINC runs on the machines when not in use by students.
    - If a user is running a Podman job outside BOINC when BOINC is installed, the job will fail because the data generated so far will suddenly be inaccessible to Podman.
    - Uninstalling BOINC must revert the /etc/zprofile file. The sudden change in Podman's working directory will again disrupt any non-BOINC Podman job in progress.

Because of all the cons, **I do not recommend this approach [1]**.

**[2]** preface every Podman command with:
   `env XDG_CONFIG_HOME="/Library/Application Support/BOINC Data/podman" XDG_DATA_HOME ="/Library/Application Support/BOINC Data/podman" `
For example:
  `env XDG_CONFIG_HOME="/Library/Application Support/BOINC Data/podman" XDG_DATA_HOME ="/Library/Application Support/BOINC Data/podman" podman --version`

This is quite easy to do if we can count on all podman commands issued by the BOINC client or by projects to use `DOCKER_CONN::command()`. The _DOCKER_CONN_ and related code can simply be updated to preface each command if the _DOCKER_TYPE_ is _podman_ and `__APPLE__` is defined.

**Pros:**
 - Requires no changes to any project code, as long as all podman commands go through `DOCKER_CONN::command()`, though existing projects will need to be relinked with the updated BOINC libraries.
 - Does not require any changes outside the BOINC code base.
 - Has no effect on any non_BOINC podman jobs which a user may run.
 - Could be used on platforms other than Mac if desired, to keep all BOINC data together and separate form user's data.

**Cons:**
 - Fails if any BOINC or project code directly issues Podman commands without going through `DOCKER_CONN::command()`. This potentially makes this approach somewhat fragile.

**[3]** Expand users _boinc_master_ and _boinc_project_
Currently the pseudo-users _boinc_master_ and _boinc_project_ do not have home directories or associated shells. Under this approach, the BOINC installer will create subdirectories _bm_home_ and _bp_home_ under the _BOINC Data_ folder and add them to the database of users as the home directories for _boinc_master_ and boinc_project. It also sets _zsh_ as their default shells and adds a file _.zprofile_ in their new home directories. The file _.zprofile_ contains the following:
```
XDG_CONFIG_HOME="/Library/Application Support/BOINC Data/podman"
export XDG_CONFIG_HOME
XDG_DATA_HOME ="/Library/Application Support/BOINC Data/podman"
export XDG_DATA_HOME
```
This adds the two environment variables globally only to users _boinc_master_ and _boinc_project_, so they will be in the environment of any code executed by BOINC.

**Pros:**
 - Does not require any code changes outside the BOINC installer, uninstaller and the _Mac_SA_Secure.sh_ script.
 - Has no effect on non_BOINC podman jobs which a user may run.
 - Works even for podman commands issued without using `DOCKER_CONN::command()`.

**Cons:**
 - Causes _boinc_master_ and _boinc_project_ to appear in the list of users in the _Users & Groups_ pane of the _System Settings_ control panel, which may be puzzling to users. (They were previously hidden.)
 - MacOS may add unneeded standard directories and files under the new home directories bm_hme and bp_home, taking up disk storage (though probably no more than about 1GB.)
 _ Requires extensive and thorough testing, as this is a hack and might  have unforeseen negative consequences.

**Conclusion:**

I have tested approaches [1] and [3] and confirmed that the basic concept of [2] (prefacing each `podman` command with the `env` statement) works.

My preferred approach is [2], but only if we can be confident that no projects will directly issue Podman commands without going through `DOCKER_CONN::command()`.

### PR DESCRIPTION
**Pull request by davidpanderson**
Podman on Mac was failing because it tried to store data in .config/ in the user's home dir.
Podman runs as boinc_master (in the client)
and boinc_projects (in the wrapper)
and neither account has access to this dir.

To fix this, we'll have the installer create an accessable dir 'podman' in the BOINC data dir.
This change prepends an 'env' directive to Podman commands on Mac, telling it to use this dir.

Fixes #6302

### DIFF
--- a/client/check_security.cpp
+++ b/client/check_security.cpp
@@ -374,6 +374,26 @@ int use_sandbox, int isManager, char* path_to_error, int len
             return retval;
     }
 
+    snprintf(full_path, sizeof(full_path), "%s/%s", dir_path, PODMAN_DIR);
+    retval = stat(full_path, &sbuf);
+    if (! retval) {                 // Client can create podman directory if it does not yet exist.
+       if (use_sandbox) {
+            if (sbuf.st_gid != boinc_project_gid)
+                return -1071;
+
+            if ((sbuf.st_mode & 0777) != 0770)
+                return -1072;
+        }
+
+        if (sbuf.st_uid != boinc_master_uid)
+            return -1073;
+
+        // Step through slot directories
+        retval = CheckNestedDirectories(full_path, 1, use_sandbox, isManager, false, path_to_error, len);
+        if (retval)
+            return retval;
+    }
+
     snprintf(full_path, sizeof(full_path),
         "%s/%s", dir_path, GUI_RPC_PASSWD_FILE
     );
@@ -595,6 +615,7 @@ static int CheckNestedDirectories(
         if (!S_ISLNK(sbuf.st_mode)) {   // The system ignores ownership & permissions of symbolic links
             if (depth > 1)  {
                 // files and subdirectories created by projects may have owner boinc_master or boinc_project
+                // TODO: Is this is true of subdirectories created by Podman tasks?
                 if ( (sbuf.st_uid != boinc_master_uid) && (sbuf.st_uid != boinc_project_uid) ) {
                     if (!isSlotDir) {
                             retval = -1202;
--- a/client/check_security.cpp
+++ b/client/check_security.cpp
@@ -15,7 +15,8 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
 
-// check_security.C
+// Functions to check ownership and permissions of files and dirs
+// in the BOINC data directory on Mac OS
 
 //#define DEBUG_CHECK_SECURITY
 
--- a/client/file_names.h
+++ b/client/file_names.h
@@ -89,6 +89,7 @@ extern void send_log_after(const char* filename, double t, MIOFILE& mf);
 #define LOOKUP_WEBSITE_FILENAME     "lookup_website.html"
 #define MASTER_BASE                 "master_"
 #define NOTICES_DIR                 "notices"
+#define PODMAN_DIR                  "podman"
 #define PROJECTS_DIR                "projects"
 #define REMOTEHOST_FILE_NAME        "remote_hosts.cfg"
 #define SCHED_OP_REPLY_BASE         "sched_reply_"
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1266,12 +1266,12 @@ bool HOST_INFO::get_docker_version_aux(DOCKER_TYPE type){
     if (type == PODMAN) {
         snprintf(cmd, sizeof(cmd),
             "%s podman machine init 2>/dev/null",
-            docker_cmd_prefix()
+            docker_cmd_prefix(type)
         );
         system(cmd);
         snprintf(cmd, sizeof(cmd),
             "%s podman machine start 2>/dev/null",
-            docker_cmd_prefix()
+            docker_cmd_prefix(type)
         );
         system(cmd);
     }
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1289,7 +1289,8 @@ bool HOST_INFO::get_docker_version_aux(DOCKER_TYPE type){
 }
 
 bool HOST_INFO::get_docker_version(){
-#ifdef __linux__
+    bool check_podman = true;
+    #ifdef __linux__
     // podman doesn't work on Linux with remote FS
     bool remote;
     int retval = is_filesystem_remote(".", remote);
@@ -1318,8 +1319,10 @@ bool HOST_INFO::get_docker_version(){
     }
 #endif
 
-    if (get_docker_version_aux(PODMAN)) {
-        return true;
+    if (check_podman) {
+        if (get_docker_version_aux(PODMAN)) {
+            return true;
+        }
     }
 
     if (get_docker_version_aux(DOCKER)) {
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -1234,14 +1234,35 @@ int HOST_INFO::get_virtualbox_version() {
     return 0;
 }
 
-// check if docker/podman is installed on this host
-// populate docker_version on success
+// check if docker/podman is installed and functional on this host.
+// if so, populate docker_version and return true
 //
 bool HOST_INFO::get_docker_version_aux(DOCKER_TYPE type){
-    bool ret = false;
-    char cmd[1024];
+    char cmd[1024], buf[256];
+
+    snprintf(cmd, sizeof(cmd), "%s %s --version 2>/dev/null",
+        docker_cmd_prefix(type),
+        docker_cli_prog(type)
+    );
+    FILE* f = popen(cmd, "r");
+    if (!f) return false;
+    // normally the version is on the first line,
+    // but it's on the 2nd line if using podman-docker
+    //
+    bool found = false;
+    while (fgets(buf, sizeof(buf), f)) {
+        string version;
+        if (get_docker_version_string(type, buf, version)) {
+            safe_strcpy(docker_version, version.c_str());
+            found = true;
+            break;
+        }
+    }
+    pclose(f);
+    if (!found) return false;
 
 #ifdef __APPLE__
+    // download (if not there) and start QEMU VM
     if (type == PODMAN) {
         snprintf(cmd, sizeof(cmd),
             "%s podman machine init 2>/dev/null",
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

**Comment by CharlieFenton** [comment]
I see a number of omissions. I will check in changes to fix those, as well as additional changes needed to the Mac installer.

**Comment by CharlieFenton** [comment]
I've checked i n my changes. @davidpanderson please review them. I'll do another code review after you tell me you're finished with your changes.

Here are some comments on my changes:

[1] I removed from HOST_INFO::get_docker_version_aux(DOCKER_TYPE type)
```
#ifdef __APPLE__
    if (type == PODMAN) {
        system("podman machine init 2>/dev/null");
        system("podman machine start 2>/dev/null");
    }
#endif

```
I don't think it is needed. If it is, please insert the set_docker_cmd_prefix(DOCKER_TYPE type) at the start of each command.

[2] If docker_cli_prog is used only for building commands, then set_docker_cmd_prefix() could be merged into it.

[3] I added the creation of the podman directory to HOST_INFO::get_docker_version(). Should it go into file_nams.cpp (along with make_project_dir and make_slot_dir, or somewhere else?

[4] I'm not sure whether to give other users and groups read / write permission for files and subdirectories under the podman directory and whether to allow them to traverse the podman directory's subdirectories

**Comment by CharlieFenton** [comment]
@AenBleidd @davidpanderson Do you understand why Scrutinizer and codecov/patch failed? Are these things I should worry about?

**Comment by AenBleidd** [comment]
@CharlieFenton,
Scrutinizer and codecov fails could be completely ignored.
I need once to configure them properly to not fail in such cases.

**Comment by CharlieFenton** [comment]
@davidpanderson To test this, in Xcode select target _Build_All_, then in the Xcode menu _Product -> Build For -> Profiling_.

You can then either run the shell script _mac_installer / release_boinc.sh_ .(See the comments at the top of _release_boinc.sh_ for instructions), or run the script mac_build/Mac_SA_Secure.sh.

**Comment by CharlieFenton** [comment]
> You can then either run the shell script _mac_installer / release_boinc.sh_ .(See the comments at the top of _release_boinc.sh_ for instructions)

And then run the installer this creates at _../BOINC_Installer/New_Release_x_y_z/boinc_x.y.z1_macOSX_universal_ where x y and z are the version number x.y.z

Note that you will only be able to run it on the Mac where you built it. You can't run a version downloaded from the Internet unless you submit it to Apple for notarizing. (You can get  it notarized using the script _mac_installer/notarize_boinc.sh_.)

**Comment by davidpanderson** [comment]
I made some small fixes and restored the 'machine init' stuff, which I think is needed.

It looks like there's code to create podman/ in both the client and in the installer.
Could it be just in the client?

Should we give it a more generic name than podman/?
It's conceivable that we'll have other uses for a directory that's
writable by boinc_master and boinc_project.
Wouldn't this have solved the virtualbox issue?

**Comment by CharlieFenton** [comment]
> I made some small fixes and restored the 'machine init' stuff, which I think is needed.

That code  is in `HOST_INFO::get_docker_version_aux(DOCKER_TYPE type)`. In my earlier tests with setting the environment variables, I was able to get the Podman version by just calling `podman --version` without using either `podman machine init` or `podman machine start`.

My understanding of Podman is very limited, but my impression from reading the documentation is that those 2 `machine` commands are used for starting the virtual machine, which then is used to run the containers from the various project tasks. I suspect  the virtual machine takes memory and significant CPU cycles, so I don't think we want to do that unless BOINC is actually going to run a task using Podman, and then it could  be done either by the client or by the _docker_wrapper_. 

> It looks like there's code to create podman/ in both the client and in the installer.
Could it be just in the client?

The code in _clientgui/mac/SetupSecurity.cpp_ is there only for completeness. It is there so that rerunning the installer fixes any ownership and permissions problems hat might have developed for any reason. It doesn't create the directory. It does a complete job of checking and fixing the ownerships and permissions of all BOINC's files and directories on the Mac, which has been very useful over the years when a user has encountered a problem in those areas. That is why the error messages triggered by the extensive checks in _client/check_security.cpp_ say "Please reinstall BOINC."

The Mac_SA_Secure.sh script is just a utility to quickly fix ownerships and permissions of BOINC's files and directories on the Mac as an alternate to a full reinstall. Since it is a shell script requiring use of the command-line interface, it's not something most users will be comfortable using. It doesn't create the directory. 

> Should we give it a more generic name than podman/?
It's conceivable that we'll have other uses for a directory that's writable by boinc_master and boinc_project.
Wouldn't this have solved the virtualbox issue?

Yes, that could have worked. I'll leave it up to you, but I think it's better to keep these sorts of things separate. Mixing files and directories from VurtualBox or some other entity in the same directory as Podman files and directories could confuse troubleshooting, and could have unintended consequences if two different uses of that directory have file or directory names that clash.  It would have been quite easy to have the client create the _virtualbox_ directory at the top level like I did for the podman directory. It will be pretty straightforward to create any new directories in the same way as I've done for Podman. @AenBleidd do you have an opinion on this question?

**Comment by CharlieFenton** [comment]
Regarding the _podman machine init_ and `podman machine start` commands: If they are needed to start the virtual machne in preparation for running containers, would it not make more sense to do that after confirming that podman is actually installed via the `podman --version` command, since my earlier tests had indicated they aren't needed before the `podman --version` command?

**Comment by davidpanderson** [comment]
The 'machine init' stuff happens only once; it completes the installation of podman
by downloading a Qemu VM which is where podman runs containers.
Having the client do this eliminates a manual step for the user.

On subsequent runs of the client, the 'machine init/start' commands are no-ops;
they exit immediately.

Fine w/ me to keep it as podman/.

**Comment by davidpanderson** [comment]
Doesn't really matter.  The machine init/start is needed to use podman.

**Comment by CharlieFenton** [comment]
Regarding the `podman machine init` and _podman machine start_ commands, even if we want to start the virtual machine before actually testing whether we have any podman tasks to run, would it still not make more sense to do that after confirming that Podman is actually installed via the `podman --version` command, since my earlier testing had shown that the `podman --version` command works without them.

**Comment by davidpanderson** [comment]
Does podman --version require the podman/ dir and the env vars?
If not we could create them only if podman is actually there.

I moved the machine init after the podman --version

**Comment by CharlieFenton** [comment]
> Does podman --version require the podman/ dir and the env vars?

It does require the _XDG_CONFIG_HOME_ environment variable when run from the client. Since that has to point somewhere accessible by user _boinc_master_, It would be best to create the  _podman/_ directory first and use that rather than using some other directory (like _/Users/Shared/_).

Here is the transcript from my earlier testing"

```
charliefenton@M1-iMac ~ % cd /Library/Application\ Support/BOINC\ Data
charliefenton@M1-iMac BOINC Data % mkdir Podman
charliefenton@M1-iMac BOINC Data % sudo chown boinc_master:boinc_project ./Podman
Password:
charliefenton@M1-iMac ~ % sudo -s -u boinc_master
Password:
boinc_master@M1-iMac BOINC Data % env "XDG_CONFIG_HOME=/Library/Application Support/BOINC Data/Podman" podman --version
podman version 5.4.2
boinc_master@M1-iMac BOINC Data % exit
charliefenton@M1-iMac BOINC Data % sudo -s -u boinc_project
boinc_project@M1-iMac BOINC Data % env "XDG_CONFIG_HOME=/Library/Application Support/BOINC Data/Podman" podman --version
podman version 5.4.2
```

davidpanderson:
In the current code, podman/ is created first,
and podman --version is run with the env prefix.
So it should work.

AenBleidd:
@CharlieFenton, @davidpanderson,
while 'podman' as a name of the directory is fine, however we have some confusion here already:
BUDA is BOINC Universal `Docker` application, but we are suggesting to use `podman`.
This is ok, however it might be better (since we don't stop users from using Docker on MacOs) to have the directory with the name wither `docker` or `container`.
But this is more recommendation.
Anyway this is internal, it should be only separate from `virtualbox`.

davidpanderson:
The dir should not be called 'buda', because there can be apps that use Docker/Podman
(and docker_wrapper) that are not BUDA.

davidpanderson:
This PR looks OK to me.  Vitalii, should we merge it?

AenBleidd (approved):

AenBleidd:
Looks good to me.

PR #6322, opened 2025-05-13
add 'Final_Install-Info.plist' to the set-client-version.py script

AenBleidd:


CharlieFenton:
This PR is Redundant. The Xcode project builds a utility _SetVersion_ from _ClientGUI/Mac/Setversion.cpp_ which updates the version numbers in all 12 plist files, based on the value of `BOINC_VERSION_STRING` in _version.h_. This has been present for many years.

CharlieFenton:
> The Xcode project builds a utility SetVersion 

... and runs it every time it builds BOINC. This guarantees that all the components of BOINC always have the same version number as provided by _version.h_, automatically.

AenBleidd:
Hm, in this case we have an inconsistency in the version management, since the version in the source tree will be different from the version that you get during the build process.
But ok, this is something we can think about and fix later. I'll revert this PR.
@CharlieFenton, thank you for the information.

PR #6329, opened 2025-05-18
Fix print statements in database module

wargoblin:
**Description of the Change**
Replaced the `print` statement with the `print()` function and added `from __future__ import print_function` to ensure compatibility with Python 2.

**Alternate Designs**
No alternate designs were necessary, as replacing `print` with `print()` is a straightforward solution to the problem.

**Release Notes**
Replaced deprecated print syntax in `py/Boinc/database.py` to ensure compatibility with Python 3 (and Python 2 if needed).


AenBleidd:
@wargoblin, thank you for the fix

PR #6337, opened 2025-05-20
Fix bug in 'Tasks for computer' view

brevilo:
**Description of the Change**
Fixes a regression in https://github.com/BOINC/boinc/commit/3691eb7385d29480cdeae843e7cd6c81cad9ee5d. There's no `$av` object anywhere and it's the `$plan_class` value we're interested in, so just use it as is. This probably never worked 🤷‍♂️


AenBleidd:
@tristanolive, @brevilo, any update on this?

brevilo:
@tristanolive 🛎️

PR #6338, opened 2025-05-20
Web, scheduler: support keyword prefs on BOINC projects

### LINKED ISSUE #6326: Keyword preferences in BOINC web code
Keywords (science area and location) are used in Science United to select projects.
We need to add them to BOINC Central (for users who attach directly);
this may as well be part of the standard BOINC code.

davidpanderson:
Science United lets users specify keyword prefs.
These are used to decide what projects to assign the user. They are also conveyed (in scheduler requests)
to projects that have jobs with keywords.
The scheduler uses this info to decide what jobs to send.

This PR extends this by
- letting non-SU users of such projects (e.g. BOINC Central) specify keyword prefs on the project web site
- assigning keywords to BUDA jobs: the union of the keywords of the BUDA app (usually science areas) and the "yes" keywords of the job submitter (usually location)

Note to self: avoid strtok(); it changes the string. Use stringstream and getline() instead.
I spent several hours chasing a bug caused by this.

Fixes #6326


AenBleidd:
@davidpanderson, please check and fix build errors (most of them are about FCGI, but logs are too big to be copied here)

PR #6339, opened 2025-05-20
client: make resource usages take effect after reread app_config.xml

### LINKED ISSUE #6334: resource usage changes to app_config.xml do not affect running tasks or downloaded tasks
### Describe the bug

In previous versions of BOINC, when you made a resource usage change to app_config.xml (using <cpu_usage> and <gpu_usage> tags under the <gpu_versions> element) for a given project, and either clicked "Read config files" in the Manager or executed "boinccmd --read_cc_config", the changes would apply to all tasks present on the system, either in behavior only (not updating the task status label) or in both the behavior in the status label. 

Now, when you do the same thing, changes are not applied real time for resource usage. they only affect newly downloaded tasks and will not come into play until the new tasks begin running. it seems that BOINC is now strictly following the "label" rather than what is trying to be changed. 

a recycle of the BOINC client WILL pick up the new change. but this was not necessary in previous BOINC versions. also other kinds of parameters in the app_config like <project_max_concurrent> still work properly and will take immediate effect on all tasks in the queue. it seems this regression only affects the resource usage parameters. 

### Steps to reproduce

1. download GPU tasks to a host. 
2. edit the app_config.xml file to change the resource usage such that it changes how many tasks are running per GPU <gpu_usage>0.5</gpu_usage> for example
3. click Read config files in Manager
4. observe in the Event log that the client found the app_config.xml file
5. observe that the change did not change how many tasks are running on each GPU for both the tasks in progress and tasks already downloaded. 

### Expected behavior

applied changes to app_config.xml resource usage should take effect on tasks in progress and tasks downloaded real-time. not only with a client recycle, as has been the case in previous client versions. 

### Screenshots

In this screenshot, the beginning state was with <gpu_usage>0.33</gpu_usage> so that 3x GPUGRID tasks were running on one GPU (the only GPU in the system). The app_config was changed to <gpu_uage>0.5</gpu_usage> to change the behavior to 2x tasks per GPU. "Read config files" was clicked. no change happened in the behavior of running tasks. you can see that a newly downloaded task has the new correct label. I expected that one task should go into "waiting to run" state and immediately move to two tasks running. 

Image

### System information

Ubuntu 24.04.2 LTS
BOINC client 8.3.0 (built from the current master) 

### Additional context

Users expect resource changes to app_config to take immediate effect without a client recycle.

davidpanderson:
Resource usage is now stored in RESULT,
and is copied either from the workunit (for BUDA-type jobs) or from the app version (for other jobs).

This is done at startup; we also need to do it after rereading app_config.xml since the app version resource usage may have changed.

Fixes #6334

### DIFF
--- a/client/app_config.cpp
+++ b/client/app_config.cpp
@@ -95,6 +95,11 @@ int APP_CONFIGS::config_app_versions(PROJECT* p, bool show_warnings) {
             );
         }
     }
+
+    // copy new app version resource usage to results
+    //
+    gstate.init_result_resource_usage();
+
     if (showed_notice) return ERR_XML_PARSE;
     return 0;
 }
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -707,14 +707,7 @@ int CLIENT_STATE::init() {
     // must go after check_app_config() and parse_state_file()
     // and after the above app version stuff
     //
-    for (RESULT* rp: results) {
-        rp->init_resource_usage();
-        if (rp->resource_usage.missing_coproc) {
-            msg_printf(rp->project, MSG_INFO,
-                "Missing coprocessor for task %s", rp->name
-            );
-        }
-    }
+    init_result_resource_usage();
 
     // this needs to go after parse_state_file() because
     // GPU exclusions refer to projects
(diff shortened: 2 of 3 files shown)


AenBleidd:
@davidpanderson, please fix build error:
```
app_config.obj : error LNK2019: unresolved external symbol "public: void __cdecl CLIENT_STATE::init_result_resource_usage(void)" (?init_result_resource_usage@CLIENT_STATE@@QEAAXXZ) referenced in function "public: int __cdecl APP_CONFIGS::config_app_versions(struct PROJECT *,bool)" (?config_app_versions@APP_CONFIGS@@QEAAHPEAUPROJECT@@_N@Z) [D:\a\boinc\boinc\win_build\sim.vcxproj]
```

davidpanderson:
fixed

AenBleidd (approved):

Issue #6332, opened 2025-05-18
Less CPU usage by Boinc client when running tasks change


hucker75:
### Describe the problem

I often see Boinc client using a whole CPU core for quite some time, during which I cannot see what Boinc is doing.  This is especially annoying when monitoring 10 computers using Boinctasks, as if one does not respond, Boinctasks waits for it before displaying anything else (I have contacted them about it but they cannot fix it).  Why is Boinc client using so much processor?  It happens most often if it's deciding which task to run next.  Not sure why this needs so much computation to make a simple decision.  On a fast CPU like a Ryzen 9 3900XT with 24 threads, it's not much of a problem.  But with older 4 core CPUs it is.  The wait can be a whole minute!

AenBleidd:
Please specify OS and release version.
It's hard to believe BOINC client takes one minute to do the task scheduling.
However, it would be nice to investigate this.
But in order to do that I need more information, so please attach detailed logs.
Thank you in advance.

hucker75:
@AenBleidd What log options should I turn on to get useful information?

Normally I can't see what Boinc is doing with the local boinc manager or boinctasks on another machine when it's busy.  I have since reduced the number of cores for projects down to 2 of 4 to prevent this.  But when I checked it's now showing what's going on in boinc manager, but not in boinctasks.  This is happening on both my old 4 core machines.

AenBleidd:
@hucker75, these flags should be fine

![Image](https://github.com/user-attachments/assets/99af767f-d250-47f1-85b0-b36a52f88b5f)

Another question: how have you measured that the BOINC client took the whole core for a minute or so?
Please keep in mind if you checked this on Windows Task Manager, you should look at the 'Details' tab, since on the 'Processes' tab it groups all the processes that have the common parent into one entity, and shows sum of the CPU and RAM usage for **all** those processes, not for the parent one only.

I think, you issue not with the BOINC client itself, but with some of the running applications that might overload system and thus make BOINC client to respond too slow.
But this will be shown in the logs.

hucker75:
@davidpanderson  At the moment about 200 tasks are queued (a day's work).  They're all the same project, Private GFN Server.

[...]

PR #6340, opened 2025-05-21
Add a config option <no_disk_usage/> that skips disk usage in sched requests

davidpanderson:
This is an experiment to see if calculating disk usage is causing client performance problems.

Related to #6332

### DIFF
--- a/lib/cc_config.cpp
+++ b/lib/cc_config.cpp
@@ -243,6 +243,7 @@ void CC_CONFIG::defaults() {
     max_tasks_reported = 0;
     ncpus = -1;
     no_alt_platform = false;
+    no_disk_usage = false;
     no_gpus = false;
     no_info_fetch = false;
     no_opencl = false;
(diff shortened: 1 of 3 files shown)


lib/cc_config.cpp
```diff
@@ -416,6 +417,7 @@ int CC_CONFIG::parse_options(XML_PARSER& xp) {
         if (xp.parse_int("max_tasks_reported", max_tasks_reported)) continue;
         if (xp.parse_int("ncpus", ncpus)) continue;
         if (xp.parse_bool("no_alt_platform", no_alt_platform)) continue;
+        if (xp.parse_bool("no_disk_usage", no_disk_usage)) continue;
```
BrianNixon, line 420: Does this need to go in `parse_options_client()` as well?

davidpanderson: Oops!! right you are.

PR #6362, opened 2025-06-01
Change size of email and user name wxTextCtrl to default

### LINKED ISSUE #5785: Linux Manager:  'Add project' wizard text entry
**Describe the bug**
The text entry box for the user's email address on page 2 is not high enough - letters with descenders (g, q, y) are truncated and hard to proof-read.

Both versions are affected - new users creating an account, and returning members adding a new machine.

**System Information**
 - OS: Linux Mint 22
 - BOINC Version: 8.0.4

**Additional context**
This problem doesn't occur on Windows - the email address is fully legible. Please check for consequential changes in other OSs when making changes.

Vulpine05:
Having a fixed size makes characters not completely visibile in some builds.

Fixes #5785

**Description of the Change**
Changed the size of the wxTextCtrl from a fixed value to wxDefaultSize

**Alternate Designs**
N/A

**Release Notes**
[Manager] Fixed bug where some characters were not completely visible when entering email or username in add project wizard.

### DIFF
--- a/clientgui/AccountInfoPage.cpp
+++ b/clientgui/AccountInfoPage.cpp
@@ -159,7 +159,7 @@ void CAccountInfoPage::CreateControls()
     itemFlexGridSizer64->Add(m_pAccountEmailAddressStaticCtrl, 0, wxALIGN_RIGHT|wxALIGN_CENTER_VERTICAL|wxALL, 5);
 
     m_pAccountEmailAddressCtrl = new wxTextCtrl;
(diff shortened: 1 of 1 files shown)


AenBleidd (approved):

CharlieFenton:
On the Mac, this PR does not change the height of the email address text field (which was not a problem before), but reduces the width of both the email address  and password text fields.
Here is the dialog on Mac version 8.2.4:
Screenshot_8_2_4

Here is the dialog form GIT Master after this PR was merged:
Screenshot_PR6362

Vulpine05:
Thank you for the update, @CharlieFenton.  Obviously that was not my intention.  I could lengthen the field, I just have to be careful with making sure it isn't too long for users using lower resolution screens.

PR #6372, opened 2025-06-06
docs: update INSTALL with server build info

wargoblin:
Updated the INSTALL guide with instructions for configuring and building the server, and added pointers to scripts in the 3rdParty directory for downloading dependencies

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -16,3 +16,20 @@ you can build and install BOINC libraries using [vcpkg](https://github.com/Micro
     ./vcpkg install boinc
 
 The BOINC port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
+
+To build the BOINC server components use the normal configure and make steps
+without `--disable-server`:
+
+    ./_autosetup
+    ./configure
+    make
+
+More detailed information about setting up a server can be found on the
+BOINC wiki (<https://boinc.berkeley.edu/trac/wiki/ServerIntro>).
+
+If you prefer downloading the dependencies yourself instead of using vcpkg,
+have a look at the helper scripts in the `3rdParty` directory.  They fetch and
+build the required libraries on Linux and macOS:
+
+    ./3rdParty/buildLinuxDependencies.sh
+    ./3rdParty/buildMacDependencies.sh


INSTALL
```diff
@@ -16,3 +16,20 @@ you can build and install BOINC libraries using [vcpkg](https://github.com/Micro
...
+
+To build the BOINC server components use the normal configure and make steps
+without `--disable-server`:
+
+    ./_autosetup
+    ./configure
+    make
+
+More detailed information about setting up a server can be found on the
+BOINC wiki (<https://boinc.berkeley.edu/trac/wiki/ServerIntro>).
```
AenBleidd, line 28: `https://boinc.berkeley.edu/trac/wiki/` (here and above) is outdated, better to use the corresponding pages in GitHub wiki: https://github.com/BOINC/boinc/wiki

INSTALL
```diff
@@ -16,3 +16,20 @@ you can build and install BOINC libraries using [vcpkg](https://github.com/Micro
     ./vcpkg install boinc
 
 The BOINC port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
+
+To build the BOINC server components use the normal configure and make steps
+without `--disable-server`:
```
AenBleidd, line 21: This is confusing. It's better to specify which flags need to be set, not vice versa

INSTALL
```diff
@@ -16,3 +16,20 @@ you can build and install BOINC libraries using [vcpkg](https://github.com/Micro
...
+    make
+
+More detailed information about setting up a server can be found on the
+BOINC wiki (<https://boinc.berkeley.edu/trac/wiki/ServerIntro>).
+
+If you prefer downloading the dependencies yourself instead of using vcpkg,
+have a look at the helper scripts in the `3rdParty` directory.  They fetch and
+build the required libraries on Linux and macOS:
+
+    ./3rdParty/buildLinuxDependencies.sh
```
AenBleidd, line 34: This is needed to build the client/manager, not the server

INSTALL
```diff
@@ -16,3 +16,20 @@ you can build and install BOINC libraries using [vcpkg](https://github.com/Micro
...
+    ./3rdParty/buildMacDependencies.sh
```
AenBleidd, line 35: we don't support building server of MacOS

davidpanderson:
This page is confusing.  It shouldn't exist.
There are 7 separate cases:
1) building the BOINC client and manager
  a) on Linux
  b) on Win with VS
  c) on Mac with XCode
2) building an application linked with the BOINC API
  a) on Linux
  b) on Win with VS
  c) on Mac with XCode
3) building the server software (on Linux)

We should (on the wiki) give a cookbook for each of these.

AenBleidd:
Then we should use this page as a enter-point with the links to all the other relevant wiki pages, including the information how to install BOINC client and Manager on any supported platform without building it from sources.

PR #6393, opened 2025-06-16
pass args in WSL wrapper

### LINKED ISSUE #6387: pass args in WSL wrapper
### Describe the bug

... like in wrapper or docker_wrapper

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### System information

_No response_

### Additional context

_No response_

ahorek:
Fixes #6387

**Description of the Change**
wsl_wrapper allows passing custom arguments to the app
```wsl_wrapper.exe --pass_thru "--threads 10"```

This is great for static arguments like controlling the number of threads, but applications often use arguments to specify what the WU should do (for example, prime ranges to scan)
```wsl_wrapper.exe --threads 10```
currently fails on an unknown option

This change parses the wrapper args like ```--os_name_regexp``` and passes all other arguments as app arguments, so that the wrapper can be used directly for some boinc projects that won't work correctly without them

example
```wsl_wrapper.exe --os_name_regexp Ubuntu --threads 10```
launches
```./main --threads 10```

the ```--pass_thru``` option is kept for compatibility and potentially conflicting flags like ```--verbose```, but in most cases it shouldn't be necessary

**Alternate Designs**
For BOINC, the wrapper is just an app, so the arguments would need to be separated
```
<app_version>
    <cmdline>--threads 10</cmdline> // + additional app args
    <wsl_wrapper>1</wsl_wrapper>
    <wrapper_args>--main_prog application --verbose</wrapper_args>
...
```
and I don’t see a need to complicate the app_info's API if the wrapper is capable of handling it

**Release Notes**
N/A

### DIFF
--- a/samples/wsl_wrapper/wsl_wrapper.cpp
+++ b/samples/wsl_wrapper/wsl_wrapper.cpp
@@ -57,6 +57,7 @@
 
 #include <cstdio>
 #include <string>
+#include <vector>
 
 #include "boinc_win.h"
 #include "util.h"
@@ -65,6 +66,8 @@
 #include "app_ipc.h"
 
 using std::string;
+using std::vector;
+using std::ostringstream;
 
 WSL_CMD app_wc;
 WSL_CMD ctl_wc;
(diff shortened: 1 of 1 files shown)


davidpanderson:
I just submitted a PR (#6394) that does the same thing as this one.  Let's use #6394.

PR #6396, opened 2025-06-18
Mac: Fix path to Podman app under Mac OS 26

CharlieFenton:
Fixes Podman detection and execution under Mac OS 26 Beta.

Under Mac OS 26, BOINC client does not inherit all default paths from user;s environment, so client must specify full path to the podman app: _/opt/podman/bin/podman_.

I will be creating one or more additional PRs for the BOINC screensaver and possibly other issues under Mac OS 26, but this PR is now ready to merge.


davidpanderson:
Charlie: Copilot pointed out an inconsistency.  Which path is right?

CharlieFenton:
The code is correct; my original description in the PR was wrong. The correct full path to the podman app is _/opt/podman/bin/podman_.

PR #6422, opened 2025-07-09
client (Win): don't recommend docker on Win 10 Home

davidpanderson:


AenBleidd:
I'll keep this open since I want to test this first.
I found a little bit confusing information.
Before I read that WSL is not available on Win10 Home, but according to this [page]( https://learn.microsoft.com/en-us/windows/wsl/faq#is-wsl-2-available-on-windows-10-home-and-windows-11-home-) it is available.
I'll check this before the weekend and come with an answer

AenBleidd:
This is not required. WSL2 is available on Windows 10 Home edition (tested on the release 22H2 build 19045.4291).

PR #6464, opened 2025-07-30
Mac podman fixes2

CharlieFenton:
Continue getting Podman working properly on Macintosh.

* Adds the missing prefix (to set needed environment variables) to `podman machine init` and `podman machine start` calls.
* Adds _HOME_ environment variable to prefix used in `podman` calls
* Moved _podman_ directory out of _BOINC Data_ directory and renames it to _BOINC podman_.
* Adjusted code to deal with the fact that Podman always writes some files as the logged-in user not boinc_master or boinc_project

I spent a great deal of time trying to get Podman to write all its files as user _boinc_master_ / _boinc_project_, experimenting with every combination of environment variables and optional arguments in the podman documentation, but was unable to do so. It always wrote at least some of its files owned by the logged in user who was running BOINC.

The `system()` and `popen()` APIs create a shell and run the passed command under that shell, but that shell always runs as the logged-in user even if the `system()` or `popen()` call was called from a process running as _boinc_master_ / _boinc_project_. One can run a command as the calling user by using `fork()` and `execvp()` via the `run_program()` function. But Podman fails with permission errors trying to write and / or read some of its necessary files even when Podman itself runs as _boinc_master_  or  _boinc_project_. 

So I decided it is best to run Podman as the logged-in user. This is the only way I was able to get `podman machine init` and `podman machine start` to succeed. I believe that it is safe (from a security standpoint) to run BOINC podman tasks as the logged-in user because podman runs them in a container.

But because BOINC otherwise uses account-based sandboxing on the mac for security, it was problematic to have the podman directory used by BOINC inside the _/Library/Application Support/BOINC Data_ directory, so I moved it out of that directory and renamed it thus: _/Library/Application Support/BOINC podman_. I changed the name to _BOINC podman_ to make it clear that this is a directory belonging to BOINC. (This means that the client no longer has permission to create the podman directory, but the BOINC installer for MacOS can because it has root privileges, so I added the code there.

It is worth noting that Podman apparently has an issue with computers that are set up with multiple users who can each login separately. If a Podman VM is created by one user, then if a different user logs in and attempts to use that same VM it will fail because Podman doesn't have permission to access the files it created under the first user.  But I haven't tested this, so I'm not certain of it. It wouldn't surprise me if even trying to create a second, separate VM under a different user via `podman machine init` and `podman machine start` might fail because Podman keeps some of its files in shared directories like _/tmp_ and _/var_ (which I have found no way to change), but will fail because they were created with a  different owner. This is probably true  on all platforms, not just MacOS.

All this implies that, in a system where multiple users have separate logins, and each can run BOINC, podman jobs will fail if started under one user and resumed by another. This is different from non-podman tasks which work correctly when handed off from one user to another (at least on macOS) because of running as _boinc_project_ rather than as the logged-in user.

### DIFF
--- a/client/check_security.cpp
+++ b/client/check_security.cpp
@@ -383,24 +383,29 @@ int use_sandbox, int isManager, char* path_to_error, int len
             return retval;
     }
 
-    snprintf(full_path, sizeof(full_path), "%s/%s", dir_path, PODMAN_DIR);
+    // "BOINC Podman" Directory is in "/Library/Application/Support/"
+    // directory, alongside "BOINC Data" directory
+    snprintf(full_path, sizeof(full_path), "%s/../%s", dir_path, PODMAN_DIR);
     retval = stat(full_path, &sbuf);
     if (! retval) {                 // Client can create podman directory if it does not yet exist.
        if (use_sandbox) {
             if (sbuf.st_gid != boinc_project_gid)
                 return -1071;
 
-            if ((sbuf.st_mode & 0777) != 0770)
+            if ((sbuf.st_mode & 0777) != 0777)
                 return -1072;
         }
 
         if (sbuf.st_uid != boinc_master_uid)
             return -1073;
 
-        // Step through slot directories
+#if 0   // We allow podman to write directories and files with different ownerships.
+        // This is safe because Pdoman runs everything in containers
+        // Step through subdirectories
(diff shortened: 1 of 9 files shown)


CharlieFenton:
After all that, it looks like I finally found a way to have Podman run entirely as user _boinc_master_ or  _boinc_project_ and to write all its files as that user. So I am closing this PR and will soon open a new one.

davidpanderson:
Yay!

PR #6465, opened 2025-08-01
Mac podman fixes3

CharlieFenton:
* Runs Podman as user boinc_project via a new utility _Run_Podman_
* Mac installer gets the path to podman executable and makes it available for all podman calls
* Adds HOME environment variable to prefix used in podman calls
* Moved podman directory out of BOINC Data directory and renames it to BOINC podman.

Podman always creates its files with owners by the logged in user who runs Podman. This is not what we want for BOINC. This PR creates a new utility _Run_Podman_ which simulates a login by user _boinc_project_ and makes all Podman calls as that user, so Podman creates all its files owned by user _boinc_project_.

_Run_Podman_ sets several Podman environment variables so Podman creates all its files in the _BOINC Podman_ directory.

This means that:
* Processing should work properly even if BOINC begins a Podman task under one user, and that user logs out and BOINC resumes the task under a different user.
* BOINC has its own Podman virtual machine and containers. So a user can do other Podman work while BOINC is running Podman task  without any interference between them.

Podman will create all its files and subdirectories owned by user and group _boinc_project_. Different files and subdirectories created by Podman have different permissions (some are readable by user and group, others only by user, etc), so we must not modify ownership or permissions on them. 

Because BOINC otherwise uses account-based sandboxing on the mac for security, it was problematic to have the podman directory used by BOINC inside the _/Library/Application Support/BOINC Data_ directory, so I moved it out of that directory and renamed it thus: _/Library/Application Support/BOINC podman_. I changed the name to _BOINC podman_ to make it clear that this is a directory belonging to BOINC. (This means that the client no longer has permission to create the podman directory, but the BOINC installer for MacOS can because it has root privileges, so I added the code there.

The _Run_Podman_ utilty takes care of adding the prefix (with environment variables) to the podman calls, so that code is no longer needed in the client.

It is worth noting that once a Podman VM is started, it continues to run even if you exit BOINC. I suggest adding a `podman machine stop` call when the client exits. Even moe efficiency might be achieved by starting the VM only when running a Norman project task and stopping it when there is none.

Note also that the first time` podman machine init` is called, it downloads a very large file from the Internet (4GB after the zip file is inflated.) The client is unresponsive during that time, so the Manager can't connect to the client until the download finishes and the zip file is inflated. Users might assume something is wrong due to that delay, especially those with a slower Internet connection.

### DIFF
--- a/client/Run_Podman.cpp
+++ b/client/Run_Podman.cpp
@@ -1,0 +1,125 @@
+// This file is part of BOINC.
+// http://boinc.berkeley.edu
+// Copyright (C) 2025 University of California
+//
+// BOINC is free software; you can redistribute it and/or modify it
+// under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation,
+// either version 3 of the License, or (at your option) any later version.
+//
+// BOINC is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
+
+// Run_Podman.cpp
+//
+// This must be called setuid root.
+// The BOINC installer for Macintoah sets this executable's permissions
+// to include the setuid bit and its owner / user to root.
+//
+
+#include <unistd.h>
+#include <stdio.h>
+#include <cstring>
+#include <sys/wait.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include<stdlib.h>
+#include <errno.h>
+
+#define VERBOSE false
+
+int main(int argc, char *argv[])
+{
+    char *args[128];
+    int argsCount = 0;
+    int argvCount;
+    char buf [2048];
+    char Podman_Path[1024];
+    FILE *f;
+    int status = 0;
+
+setuid(0);  // May not actually be needed
+
+#if VERBOSE
+    FILE *debug_file = fopen("/Users/Shared/Run_Podman-debug.txt", "a");
+    fprintf(debug_file, "uid=%d, euid=%d\n", getuid(), geteuid());
+#endif
+
+    // While the official Podman installer puts the porman executable at
+    // "/opt/podman/bin/podman", other installation methods (e.g. brew) might not
+#if 1
+    // The Mac installer wrote path to podman executable in Path_to_podman.txt
+    Podman_Path[0] = 0;
+    f = fopen("/Library/Application Support/BOINC Data/Path_to_podman.txt", "r");
+    fgets(Podman_Path, sizeof(Podman_Path), f);
+    if (Podman_Path[0] == 0) {
+        // If we couldn't get it rom that file, use the default
+        strlcpy(Podman_Path, "/opt/podman/bin/podman", sizeof(Podman_Path));
+    }
+    fclose(f);
+#else
+    // Get the path to the podman executable dynamically
+    char                    s[2048];
+    // Mac executables get a very limited PATH environment vatiable, so we must get the
+    // PATH variable used by Terminal and search ther of the path to podman
+    f = popen("a=`/usr/libexec/path_helper`;b=${a%\\\"*}\\\";env ${b} which podman", "r");
+    s[0] = '\0';
+    if (f) {
+        fgets(s, sizeof(s), f);
+        pclose(f);
+        char * p=strstr(s, "\n");
+        if (p) *p='\0'; // Remove the newline character
+        fprintf(debug_file, "popen #2 returned podman path = \"%s\"\n", s);
+        fclose(f);
+    }
+#endif
+
+    // Podman always writes its files owned by the logged in user and
+    // mosly in that user's home directory. To get around this, we
+    // simulate a login by user boinc_project and set envirtonment
+    // variables for Podman to use our BOINC podman" directory instead
+    args[argsCount++] = "su";
+    args[argsCount++] = "-l";
+    args[argsCount++] = "boinc_project";    // Create Podman VM using boinc_project so projects can access it
+    args[argsCount++] = "-c";
+    strlcpy(buf, "env XDG_CONFIG_HOME=\"/Library/Application Support/BOINC podman\" XDG_DATA_HOME=\"/Library/Application Support/BOINC podman\" HOME=\"/Library/Application Support/BOINC podman\" /opt/podman/bin/podman", sizeof(buf));
+
+    argvCount = 1;   // arguments to be passed to Podman
+    while (argv[argvCount]) {
+        strlcat(buf, " ", sizeof(buf));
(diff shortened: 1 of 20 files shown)


CharlieFenton:
@AenBleidd @davidpanderson I don't understand the error on Windows:
> 2025-08-01T15:21:26.8890986Z          D:\a\boinc\boinc\clientgui\AdvancedFrame.cpp(1569,25): error C2512: 'wxTimerEvent': no appropriate default constructor available [D:

The Linux builds have the same error. 

This code has not changed in years, and I did not make any changes to this file in this PR. Also, it compiled successfully on my Mac. The only possibility I can think of is that perhaps the CI is using a new version of wxWidgets that has a different API for wxTimerEvent.

In any case, None of my changes have anything to do with these errors. I am changing this PR to "ready for review." Please merge it.

AenBleidd:
Yes, this is the issue with the new release of wxWidgets.
Btw, that variable was added by you significantly recently.
I haven't look too deep but looks like this variable isn't ever used.
Could you please double check why it's there and if it's needed at all if you have time?

CharlieFenton:
@AenBleidd I have no idea why they are there (there are two: line 1569 and line 1999. I don't remember adding them and they clearly serve no purpose. I'll remove them now and add them to this PR.

CharlieFenton:
@AenBleidd I'm guessing that perhaps they were part of a change I tried and decided to remove because it didn't work, but that i accidentally left thosedeclarations when i removed the other code. But i really don't remember anything about them.

CharlieFenton:
@AenBleidd The Windows error is:
> 2025-08-01T16:19:46.0609579Z          LINK : fatal error LNK1181: cannot open input file 'wxbase32u.lib' [D:\a\boinc\boinc\win_build\boincmgr.vcxproj]

Do you know why this is happening? I son't think it has anything to do with this PR; does it?

The Windows Manager build has 1131 warning(s), and the word "error" appears 47 times, which makes it very difficult to find the actual error message. And so many warnings hides any warnings that may actually help diagnose problems. It would be nice if something could be done about that..

CharlieFenton:
@AenBleidd @davidpanderson Please merge this.

mac_installer/PostInstall.cpp
```diff
@@ -241,6 +241,24 @@ int main(int argc, char *argv[])
             return DeleteReceipt();
     }
 
+    // While the official Podman installer puts the porman executable at
```
Vulpine05: Typo on "podman".

mac_installer/release_boinc.sh
```diff
@@ -63,6 +63,8 @@
 ## Updated 5/17/23 to add comments about notarize_BOINC.sh script
 ## Updated 2/12/25 to add support for Fix_BOINC_Users utility
 ## Updated 7/22/25 to add MacOS 26 support, no Finish_Install embedded in Postinstall.
+## Updaed 7/29/25 to add "BOINC podman" directory
```
Vulpine05: "updated" typo

mac_installer/release_boinc.sh
```diff
@@ -63,6 +63,8 @@
...
+## Updaed 7/31/25 to add "Run_Podman" utility
```
Vulpine05: "updated" typo

Vulpine05 (commented):
I found a couple of typos in the comments.

AenBleidd:
I will fix the Windows build before merge

CharlieFenton:
@AenBleidd If Copilot doesn't recommend any more changes, please merge this

CharlieFenton:
> I will fix the Windows build before merg

OK, but unless you make that fix in this branch, this CI for this PR will still show those errors

client/Run_Podman.cpp
```diff
@@ -0,0 +1,130 @@
+// This file is part of BOINC.
+// http://boinc.berkeley.edu
```
Vulpine05: Please change to https.

client/Run_Podman.cpp
```diff
@@ -0,0 +1,130 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
+
+// Run_Podman.cpp
+//
+// This must be called setuid root.
+// The BOINC installer for Macintoah sets this executable's permissions
```
Vulpine05: Typo: "Macintosh".

client/Run_Podman.cpp
```diff
@@ -0,0 +1,130 @@
...
+// as published by the Free Software Foundation,
+// either version 3 of the License, or (at your option) any later version.
+//
+// BOINC is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with BOINC.  If not, see <http://www.gnu.org/licenses/>.
```
Vulpine05: Please change to https.

client/Run_Podman.cpp
```diff
@@ -0,0 +1,130 @@
...
+    args[argsCount++] = "boinc_project";    // Create Podman VM using boinc_project so projects can access it
+    args[argsCount++] = "-c";
+    snprintf(buf, sizeof(buf), "env XDG_CONFIG_HOME=\"/Library/Application Support/BOINC podman\" XDG_DATA_HOME=\"/Library/Application Support/BOINC podman\" HOME=\"/Library/Application Support/BOINC podman\" %s", Podman_Path);
+
+    argvCount = 1;   // arguments to be passed to Podman
+    while (argv[argvCount]) {
+        strlcat(buf, " ", sizeof(buf));
+        strlcat(buf, argv[argvCount++], sizeof(buf));
+    }
+//    strlcat(buf, "\'", sizeof(buf));
```
Vulpine05: Is this line needed?  If it is being kept in for debugging or testing, could you add a comment explaining what it is for?  It helps people like me that aren't familiar with the purpose of the code.

Vulpine05 (commented):
I found a few more minor nit picks.

CharlieFenton:
@AenBleidd Please merge this NOW! Please don't hold this up for minor typos in comments or for unrelated issues in Windows/

We need to get this merged so @davidpanderson can use it to debug the server side. My test show that either the wrapper is never getting called by the project code, or it needs to be  updated on the server side, neither of which I can do. When I built a wrapper with the new code and  tried simply substituting it for the original in the BOINC projects directory nothing changed; so @davidpanderson must look into that and he can't easily do this until this PR has been merged.

AenBleidd:
@CharlieFenton, I'll merge it today. Sorry, I was unavailable during the last few days.

AenBleidd (approved):

davidpanderson:
Are you able to submit BUDA jobs on the test project?

If you upload a new docker_wrapper to the server I'll deploy it in the test project.

I'm away from my Mac until 15 Aug so my debugging capabilities are limited.

PR #6473, opened 2025-08-06
update NVidia flops calculation for Blackwell

### LINKED ISSUE #6472: update coproc.cpp nvidia Flops calculation for 50-series GPUs (and others)
### Describe the bug

sorry i don't know how to create a PR in github. see additional comments for suggested code changes

similar to PR#4031 : [(https://github.com/BOINC/boinc/pull/4031)]

50-series GPUs have a CC of 12.0, and I believe the flops calculation is hitting the 8.0 code path in lib/coproc.cpp which leads to the incorrect flops calculation for some GPUs with CC major version > 8. since they all have minor version of "0", they incorrectly use cores_per_proc value of "64" instead of the real value. 

Hopper Datacenter (H100/H200) = CC 9.0 : cores_per_proc = 128
Blackwell Datacenter (B200) = CC 10.0 : cores_per_proc = 128 
Blackwell GeForce (50-series) = CC 12.0 : cores_per_proc = 128

### Steps to reproduce

start BOINC client with Nvidia 50-series GPU, observe that flops calculation is roughly half what it should be based on expected values. 

ex, a 5070 reports ~15.9 TFLOP when it should be closer to 31 TFLOP

### Expected behavior

BOINC client calculates the correct FLOPs estimate for newer nvidia GPUs and 50-series GPUs

### Screenshots

_No response_

### System information

should affect all clients. Linux/Windows and all older BOINC versions. 

### Additional context

this code block should be added to lib/coproc.cpp around line 597. 

        case 9:    // for cc9.0 (Hopper Datacenter H100/H200)
            flops_per_clock = 2;
            cores_per_proc = 128;
            break;
        case 10:    // for cc10.0 (Blackwell Datacenter B200)
            flops_per_clock = 2;
            cores_per_proc = 128;
            break;
        case 12:    // for cc12.0 (Blackwell GeForce 50x0)
            flops_per_clock = 2;
            cores_per_proc = 128;
            break;

ahorek:
Fixes #6472

**Description of the Change**
With CUDA 12.0, Blackwell is treated under the same default 8.0 special case as the A100 Tensor Core datacenter GPU.

https://developer.nvidia.com/cuda-gpus

**Alternate Designs**
Instead of relying on a lookup table, we could measure it through benchmarking (like CPU apps), though that would be a more significant change.

**Release Notes**
Fixed estimated flops calculation for NVidia Blackwell

### DIFF
--- a/lib/coproc.cpp
+++ b/lib/coproc.cpp
@@ -582,8 +582,7 @@ void COPROC_NVIDIA::set_peak_flops() {
             flops_per_clock = 2;
             cores_per_proc = 64;
             break;
-        case 8:    // for cc8.0 (A100) and cc8.6 (GeForce RTX 30x0 - GA102 and above)
-        default:
+        case 8:    // for cc8.0 (A100) and cc8.6 (GeForce RTX 30x0 - GA102)
             flops_per_clock = 2;
             switch (minor) {
             case 0:    // special for A100 Tensor Core datacenter GPU
@@ -593,6 +592,10 @@ void COPROC_NVIDIA::set_peak_flops() {
                 cores_per_proc = 128;
                 break;
             }
+            break;
+        default:   // for cc9.0-12.0 (and above) (Hopper Datacenter H100/H200, Blackwell Datacenter B200, Blackwell GeForce 50x0)
+            flops_per_clock = 2;
+            cores_per_proc = 128;
             break;
         }
 


IanSteveC (approved):
looks good

davidpanderson:
Looks good, thanks.

PR #6494, opened 2025-08-13
client: remove Docker containers named hello-world; we probably created them

davidpanderson:


TeeVeeEss:
The created names of the hello-world containers are random picked, like this example in my containers.json:
```json
{
  "id": "124dc77c41b6246f4cd0eca38a5f83d361e8bc1f80dcd4dd7f0de9969ca265d4",
  "names": [
    "reverent_haslett"
  ],
  "image": "74cc54e27dc41bb10dc4b2226072d469509f2f22f1a3ce74f4a59661a1d44602",
  "layer": "1abc6bec1d1a2ea977dbbb62f628d1637e69dc056f29d576a71be4c8503ebe99",
  "metadata": "{\"image-name\":\"docker.io/library/hello-world:latest\",\"image-id\":\"74cc54e27dc41bb10dc4b2226072d469509f2f22f1a3ce74f4a59661a1d44602\",\"name\":\"reverent_haslett\",\"created-at\":1751120539}",
  "created": "2025-06-28T14:22:19.190258588Z",
  "flags": {
    "MountLabel": "",
    "ProcessLabel": ""
  }
}
```
So maybe check the metadata?

AenBleidd:
This is not needed. Now all the created 'hello-world' containers will be automatically removed.
Old created containers could be removed manually if needed, this must not be a part of the client.

TeeVeeEss:
Imho I disagree. This PR is probably not good enough, but -more generally speaking- : what is created by the client, should be deleted by the client.

AenBleidd:
And this will be done with the new released version.
However we will not do any clean up of leftovers made by older clients.
That is why this PR was rejected.

TeeVeeEss:
I agree to disagree :) Is the de-installation of the client taking care of removing the boinc user and it's files?

AenBleidd:
No, and this is done by intention.
Again: new version of BOINC will do this cleanup. For the corrent version we will check how many users are really affected and what is the real number of leftovers.
There might be the case that there is no issue at all.

TeeVeeEss:
Ok, you can ping me for testing the cleanup. There are ~20 hello-world containers on my host now, running version `8.2.4 x86_64-pc-linux-gnu` , dating from 1751120221 to 1758666100 (jun 26 - sep 24 2025)

PR #6497, opened 2025-08-16
Mac screensaver: fix annoying white flash when dismissing screensaver

CharlieFenton:
As explained [here](https://github.com/BOINC/boinc/pull/6482#issue-3307191169), the MacOS does not allow other windows to cover the screensaver, so we can no longer just launch project graphics apps on top of the MacOS screensaver. Instead, we use _IOSurfaceBuffers_ to display the graphics from the project graphics app in the screensaver's window.

We don't render the graphics in the project graphics app's own window, because that would waste processor resources since that window is obscured by the screensaver's window. As a result, the project graphics app's window, which is the size of the full screen, remains completely white.

When the screensaver is dismissed, a fraction of a second may pass before the graphics app exits. This may briefly reveal the graphics app's white fullscreen window, causing a disturbing bright flash. The BOINC screensaver code uses a small helper app _gfx_cleanup_ to cover the white window with its own black fullscreen window, and to ensure that the graphics app exits before the black covering window is removed. It also kills the graphics app if necessary.

As I made additional changes over time to BOINC's Mac screensaver logic to address new limitations imposed by MacOS updates, the code in _gfx_cleanup_ became inadequate. This PR updates that logic and its associated code.

In addition, this PR makes a small efficiency improvement to the screensaver bridging utility SS_GFX_Bridge introduced in PR #6482.

### DIFF
--- a/clientscr/sharedGraphicsController.h
+++ b/clientscr/sharedGraphicsController.h
@@ -48,7 +48,6 @@ static NSTimer * myTimer = NULL;
 @interface SharedGraphicsController : NSObject <NSMachPortDelegate>
 {
     bool isFromGFXApp;
-    bool runningSharedGraphics;
     char portNameToLookUp[128];
 
 	NSMachPort *serverPort;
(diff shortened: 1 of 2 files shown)


CharlieFenton:
Please merge

PR #6503, opened 2025-08-20
Mac: Fix Podman support for MacOS versions earlier than MacOS 26

CharlieFenton:
I don't understand why, but `su -l boinc_project -c` worked under MacOS 26 but not MacOS 15.6, but `su -m boinc_project -c` works under both for running Podman commands.


CharlieFenton:
This change does not fix the problem running podman on the Mac and is superseded by #6508.

PR #6517, opened 2025-08-24
client (Mac): count the VM used by Podman as BOINC CPU time

davidpanderson:


CharlieFenton:
@davidpanderson Do you need to do the equivalent for Podman on Windows and Linux, and for Docker on all platforms? Or have you already done all that?

davidpanderson:
Already done.

PR #6519, opened 2025-08-25
Update all_projects_list.xml

LezheGao:
**Description of the Change**
Changed the description of PrimeGrid and SiDock@Home

**Release Notes**
Updated the description of PrimeGrid and SiDock@Home


AenBleidd:
@LezheGao, thank you for the pr.
However, I have to close it because we have a different mechanism to update it.

PR #6529, opened 2025-08-28
Client, docker_wrapper (Mac): debug BUDA

davidpanderson:
There were several issues
(note: on the Mac, the data directory is
"/Library/Application Support/BOINC Data")

* Podman doesn't accept relative paths (like .) on Mac.
* When you pass it an arg with spaces (like the above path)
    it's not enough to quote it.
    You also have to escape the spaces.
    This applies to paths and also to cmdline args like '--nsecs 20'
* By default, you can only mount directories under /Users.
    Not, for example, under /Library.
    To do that, you need to initialize Podman's VM using
    machine init -v /Library:/Library

Also:
* Make the client's app_test feature work on Unix.
    App files need to be executable.


CharlieFenton:
Please do NOT merge this. I cannot accept giving podman access to the entire /Library/ directory.

CharlieFenton:
Regarding my comment [here](https://github.com/BOINC/boinc/pull/6529#pullrequestreview-3173311779), after spending quite a bit of time trying to avoid that, I have  been unable to find a better solution and so withdraw my objection.

CharlieFenton (approved):

PR #6535, opened 2025-09-01
 Remove Info-ZIP library and use libzip instead

AenBleidd:


CharlieFenton:
The Mac provides lib a a standard available library. It is what the Mac client uses. BOIN projects for the Mac can also use the standard lib. Is there any reason for us to have our own code dealing with zip and unzip on the Mac?

AenBleidd:
@CharlieFenton, I'm not sure I got your question correctly: `zip` and `unzip` functionality are used by sample applications (e.g. wrapper, docker_wrapper). I would prefer to not have it in a separate library with just two functions.

CharlieFenton:
@AenBleidd Sorry, I now noticed that my earlier comment had a typo and was missing the letter "z" (auto-correct keeps removing the "z" when I type). The standard Zip library provided by MacOS and used by the Mac client is libz and is invoked by the linker flag -lz when building the client. 

But I am now rather confused. The Mac client is built with the linker flag -lz which brings in the library libz, but Xcode associated the calls in the client such as gzopen() with the header file zlib.h, which says:
> zlib.h -- interface of the 'zlib' general purpose compression library
  version 1.2.12, March 11th, 2022 

So the Mac client is linking with libz while using the lib API, even though the Mac does not supply zlib.

On the other hand, boinc_zip.cpp calls zip_open(), which is not in zlib.h but is listed in the [libzip documentation](https://libzip.org/documentation/zip_open.html) .

It would be nice if we don't have to build our own libzip in addition to the libz already provided as part of macOS.

Please note that the Mac build of the client manager don't link with any BOINC libraries, but instead directly build and link to all the needed files from the lib/ and api/ directories. I did it this way to simplify debugging under the debugger.

AenBleidd:
@CharlieFenton,

first of all, thank you for your time you spent while reviewing this PR.

'libz' is just a compression library. It can compress one file but cannot manage multiple files inside one ZIP archive.
In order to not implement this functionality before we have used 'Info-Zip' that I replaced with 'libzip' (that is more supported and don't need to be agreed inside our repo).

Regarding the Mac build, I didn't get exactly what are you talking about.
Linking with the static library is a normal way of managing the application dependencies on all platforms. I don't know what are the improvements you get during the debugger, but I definitely know that in case of every change inside libboinc file that is also included into your Mac build, you will have to build it at least twice: one time for the libboinc, and the second time for every project this file is included into. So from my point of view this just makes the build time bigger without any actual benefits.
Also, please keep in mind: it might be that your approach gave you benefits a decade ago, but tools (compiler, linker, debugger, etc) are significantly improved since then, so your existing approach might not give you any benefits anymore.
But this is another topic.

Going back to the original question: with my explanation of the purpose of the 'libzip' library: are we on the same page or do we have some open questions still?

Thank you in advance.

CharlieFenton:
Please ignore my earlier comments. I committed a few changes to this branch so I can now build locally using both the Xcode GUI and the BuildMacBOINC.sh script. I will need to update the build instructions in _mac_build/HowToBuildBOINC_XCode.rtf_ and _mac_build/HowToBuildBOINC_XCode.pdf_ before we merge. I'll try to get to that in the next couple of days.

For future reference, I avoid doing `make install` on local builds of the dependent libraries to keep my Mac as close to standard as possible to avoid being misled when looking at user's problem reports. So the `make install` in _buildlibzip.sh_ should be caked only for CI builds; I made that change.

CharlieFenton:
I may also add a `-nolibs` option to the _mac_build/BuildMacBOINC.sh_ script so people can build the installer themselves without needing _cmake_. The _lib/boinc_zip.cpp_ file has `#include <zip.h>` which is part  of _libzip_ and has `#include <zipconf.h>` which doesn't exist until libzip is built. So building _libboinc_ now fails unless you first build _libzip_ and that requires that you first install cmake.

AenBleidd:
@CharlieFenton, thank you very much for that!
I fixed a little typo in one of your scripts.
Will wait for the updated build instructions.

CharlieFenton:
@AenBleidd I apologize, but I'm afraid I have been having serious second thoughts about this PR. When I told you I was OK installing CMake to build _libzip_, the files related to zip were in a separate part of the source tree and were not part of the main BOINC builds. The main Xcode project for BOINC, _mac_build/boinc.xcodeproj_,  builds the client and Manager, screensaver, installer and uninstaller, the default screensaver graphics _boincscr_ and the 4 BOINC libraries _libboinc_, _libboinc_api_, _libboinc_grapics2_ _libboinc_opencl_ as well as _libjpeg_. I believe the same was true of the MS Windows builds using _boinc.sln_, though it has been a long time since I had a Windows box to do this on so my memory may be faulty.

The zip sources, including _boinc_zip.cpp_ and _boinc_zip.h_, were part of a separate Xcode project _zip/boinc_zip.xcodeproj_, which built the _libboinc_zip_ library. So I expected that CMake would be needed only to build that separate zip project.

I was surprised when I found you had moved _boinc_zip.cpp_ and _boinc_zip.h_ into _libboinc_ because this means that one cannot build the entire _boinc.xcodeproj_ (or run the  _mac_build/BuildMacBOINC.sh_ script with the default -all option) without installing CMake for the reason I wrote [here](https://github.com/BOINC/boinc/pull/6535#issuecomment-3333770881). I think adding this complexity is a bad decision, given that few people will need to use the functions provided by _boinc_zip.cpp_.

I have put quite a bit of effort into making it easy for people to build the client and installer themselves, since some feel more comfortable running software they have built. Requiring a person to install CMake on a Mac adds an unnecessary additional level of complexity. (It is not enough to installing CMake on the Mac. One then needs to type a command into the Terminal application with _sudo_ to enable invoking CMake from the command line. Without that, the _buildlibzip.sh_ script doesn't work.)

This leads me to wonder why this is needed at all on the Mac. You wrote that libzip is preferable to zlib because:
> 'libz' is just a compression library. It can compress one file but cannot manage multiple files inside one ZIP archive.
In order to not implement this functionality before we have used 'Info-Zip' that I replaced with 'libzip' 

While I'm skeptical about zlib being unable to handle multiple files, the fact is that the BOINC client on Mac has been using zlib for decades to compress and uncompress project files without any problem. So I don't understand what part of BOINC needs a capability not provided by zlib.

My proposal is to modify _boinc_zip.cpp_ and _boinc_zip.h_ so they are compatible with zlib on the Mac (perhaps with `#idfef __APPLE__` guards), eliminating the need for libzip on the Mac, and thus eliminating the need for CMake on the Mac. If that is not practical, then I would like to see _boinc_zip.cpp_ and _boinc_zip.h_ moved out of _libboinc_ and back into the separate _libboinc_zip_ library with its own Xcode project.

Thank you for considering this.

CharlieFenton:
> While I'm skeptical about zlib being unable to handle multiple files,

As I have looked into this, I see you are correct about that, though this can be circumvented by first tarring the files. But I doubt any projects have used this capability of tipping multiple files together, so I'm not sure we need to support it.

> zip and unzip functionality are used by sample applications (e.g. wrapper, docker_wrapper). 

I don't see any mention of zip or unzip in _docker_wrapper_. The use of `boinc_zip()` in docker_wrapper is for single files, not multiple files compressed together. The only other file in  our source that uses `boinc_zip()` is _samples/example_app/slide_show.cpp_ which does show how to download multiple image files as a single zip archive. But I am not aware of any actual project that compresses multiple files in a single zip archive. The Mac client has successfully downloaded the slide shows (for Simple View) for all projects I have run, using zlib.

AenBleidd:
@CharlieFenton, thank you for the detailed review.

> Requiring a person to install CMake on a Mac adds an unnecessary additional level of complexity.

Let's be realistic: I doubt anybody is still building BOINC for Windows or MacOS manually nowadays except you and David.
Also, cmake is a standard now. The only reason we don't ust it to build BOINC itself is because we don't have resources to move to it to unify our Android, Linux, MacOS, and Windows builds. And installing cmake is not a big deal: it's just one tool that is comparably small.

> I think adding this complexity is a bad decision

Basically, I moved these two files to the `libboinc` to remove the complexity of having an additional library with just two functions. We already have too complext source code, and we need to try to make it more maintainable even if that means to change something.

>  But I am not aware of any actual project that compresses multiple files in a single zip archive.

Yes, you are right, there is no usage of `boinc_zip` in docker_wrapper, that's my mistake.
However we do use it in `wrapper` but we never check the content (at the list of resulting files) after we unzip the archive, so you might just never mentioned this.

As a conclusion: I don't see cmake as a big issue, but if you do and insist on moving libboinc_zip library back - I will do that.
Just let me know.

Thank you in advance.

CharlieFenton:
>  do and insist on moving libboinc_zip library back - I will do that.

I wanted to think about this overnight before deciding, but my conclusion is that yes, please remove boinc_zip.cpp and boinc_zip.h from _/lib_ and from _libboinc_ and again make them part of a separate _libboinc_zip_ library in a separate _/zip_ directory, as before.

Once you have done this, I will create a new _zip/boinc_zip.xcodeproj_ Xcode project , to build the libboinc_zip library on the Mac, which will expect to link with _libzip_.

One of the reasons I made this decision is that, as you have it now, every app that links with _libboinc_ will need to be modified to also link with _libzip_, though very few actually use that functionality. This will create extra work for many projects, as their current builds will fail unless they build _libzip_ and make that change to their build (makefile, vcxproj or xcodeproj).

CharlieFenton:
installing CMake on the Mac was not fully intuitive. The download provides CMake as a GUI application with instructions to copy it into the /Applications directory. But an additional step is needed to make it usable from the command line, which is needed for the _buildlibzip.sh_ script to work, and I had to do a Google search to figure out how to do that.

I will add a comment to the _buildlibzip.sh_ script  explaining the necessary steps.

CharlieFenton:
> Once you have done this, I will create a new _zip/boinc_zip.xcodeproj_ Xcode project , to build the _libboinc_zip_ library on the Mac, which will expect to link with _libzip_.

We will also need to revert the main Xcode project _mac_build/boinc.xcodeproj_ and _setupForBOINC.sh_ script to the ones currently in Github master. I'll check to see if there are any others. The existing _mac_build/HowToBuildBOINC_XCode.rtf_ and _mac_build/HowToBuildBOINC_XCode.pdf_ will then not need any changes.

@AenBleidd I apologize for creating extra work for you, but I strongly feel this is the right thing to do.

AenBleidd:
@CharlieFenton, that is fine, and you are probably right.
I was also reflecting on your response today and came to the conclusion that you suggestion is better and more reliable.
I made the required changes.
I haven't reverted back your changes (ok, at least not all of them) in case you want to reuse those scripts that left.
I hope this is read for review, but I'll keep an eye on the CI checks: it might be that something went wrong with the merge again.
Anyway, I'll ping you again when this PR is ready for your review again.
Thank you.

AenBleidd:
@CharlieFenton, this PR is now ready for your review.
Please let me know if you need me do make any additional changes.
Thank you in advance.

CharlieFenton:
@AenBleidd I've been trying to get the slide_show sample to build. I have modified _samples/example_app/Makefile_mac_ and _samples/example_app/Makefile_mac2_ to build successfully as long as it links to both _libzip_ and _zlib_. However, if I move the _-lz_  from the linking instructions at _slide_show_x86_64:_ and _slide_show_arm64:_ then I get these errors:
Undefined symbols for architecture x86_64:
  "_crc32", referenced from:
      __zip_cdir_write in libzip.a[x86_64][10](zip_dirent.c.o)
      __zip_write in libzip.a[x86_64][51](zip_io_util.c.o)
      __zip_pkware_encrypt in libzip.a[x86_64][57](zip_pkware.c.o)
      __zip_pkware_encrypt in libzip.a[x86_64][57](zip_pkware.c.o)
      __zip_pkware_encrypt in libzip.a[x86_64][57](zip_pkware.c.o)
      __zip_pkware_encrypt in libzip.a[x86_64][57](zip_pkware.c.o)
      __zip_pkware_decrypt in libzip.a[x86_64][57](zip_pkware.c.o)
      __zip_pkware_decrypt in libzip.a[x86_64][57](zip_pkware.c.o)
      __zip_pkware_decrypt in libzip.a[x86_64][57](zip_pkware.c.o)
      __zip_pkware_decrypt in libzip.a[x86_64][57](zip_pkware.c.o)
      ...
  "_deflate", referenced from:
      _process in libzip.a[x86_64][5](zip_algorithm_deflate.c.o)
  "_deflateEnd", referenced from:
      _end in libzip.a[x86_64][5](zip_algorithm_deflate.c.o)
  "_deflateInit2_", referenced from:
      _start in libzip.a[x86_64][5](zip_algorithm_deflate.c.o)
  "_inflate", referenced from:
      _process in libzip.a[x86_64][5](zip_algorithm_deflate.c.o)
  "_inflateEnd", referenced from:
      _end in libzip.a[x86_64][5](zip_algorithm_deflate.c.o)
  "_inflateInit2_", referenced from:
      _start in libzip.a[x86_64][5](zip_algorithm_deflate.c.o)
ld: symbol(s) not found for architecture x86_64

Do you have any ideas why this is not working without zlib?

AenBleidd:
@CharlieFenton,
'libzip' is a library to manage zip archives that uses DEFLATE compression algorithm from 'zlib' (technically speaking, it's one of the algorithms that could be used by libzip but in BOINC we're are interested in DEFLATE only).
So, we need zlib (-lz) together with libzip (-lzip).
I hope this answers your question.

CharlieFenton:
> However, if I move the -lz from the linking instructions at slide_show_x86_64: and slide_show_arm64: then I get these errors:

The same happens if I remove the -lz from the slide show target in _/samples/mac_build/UpperCase2.xcodeproj_.

AenBleidd:
@CharlieFenton, please read my answer above

CharlieFenton:
> However, if I move the _-lz_ from the linking instructions at _slide_show_x86_64: and slide_show_arm64_: then I get these errors:

And also the same happens if I remove the _-lz_ from _samples/wrappr/Makefile_mac_

AenBleidd:
@CharlieFenton, yes this is correct since every componen that builds with  libzip required zlib.
libzip is not a replacement for zlib, they are working together.
Please don't try to overengineer something that is already working.

CharlieFenton:
> libzip is not a replacement for zlib, they are working together.

Thank you. I did not understand that, I thought libzip was designed to stand alone.

I should be able to finish up my fixes later tonight.

CharlieFenton:
@AenBleidd I have finished my changes for the Mac. Thank you for your patience. Although the CI builds succeeded, the local Mac builds are failing because the paths to the dependent libraries are different on the CI builds.

AenBleidd:
@CharlieFenton,
you basically highlighted the issue we have here: there are several ways of building the same.
We should try to avoid that and keep only one solution.
I did this for Windows and Android. For Linux the work in is progress, for MacOS I currently have no experience and no time to invest in it, but from the other hand this is not so urgent to do this now.

> Although the CI builds succeeded, the local Mac builds are failing because the paths to the dependent libraries are different on the CI builds.

So are there any open building issues left or this PR is ready to be merged?
Thank you in advance for the clarification.

PR #6557, opened 2025-09-19
client (Mac): initialize Podman VM in a separate process

AenBleidd:


AenBleidd:
@davidpanderson, could you please review conflict here?
Thank you in advance.

davidpanderson:
Done

AenBleidd:
@CharlieFenton, please review

AenBleidd:
@CharlieFenton, so is it something that need to be fixed/added in this PR or it is ready to be merged?

CharlieFenton:
> so is it something that need to be fixed/added in this PR or it is ready to be merged?

I think the delay from `podman machine start` (10 seconds on a fast arm64 computer) is long enough that it should be included in this PR.

> So similar logic probably needs to be added for that, though the two Podman commands could be combined into one call of cmd() by putting a semicolon between them.

While combining them  into one call is a little bit simpler, it would be better to do the 2 calls separately because we are considering having the client stop the VM when not in use and starting it again each time it is needed.

CharlieFenton:
@AenBleidd I have some ideas I want to look at. Please don't make any more changes yet.

CharlieFenton:
My idea is to do both the `podman machine init `followed by `podman machine start` in a separate thread rather than a separate process. I think this will simplify the code. I should have time to test this within a few days.

davidpanderson:
The current code does both init and start.
It does this in a separate process, which is probably simpler than using a thread
(which would end up doing the work in a process anyway)

CharlieFenton:
From my reading of the code in this PR, it only does the init in a separate process not the start. A separate thread can init, wait for that and then start before setting a shared bool to signify it is ready.

Adding a thread is easy; i’ve done it in other parts of BOINC for Mac. Remember only the Mac needs to do the start and init.

I’ll create an alternate PR so you can see what I mean and then decide which to use.

davidpanderson:
This PR currently has

        snprintf(cmd, sizeof(cmd),
            "%s machine init; %s machine start",
            docker_cli_prog(type),
            docker_cli_prog(type)
        );

CharlieFenton:
> "%s machine init; %s machine start"

Ah, OK I didn't notice that before. So I'm fine with doing the way you have. But I just tested it on my Mac and it works but there are some issues:

During init and start, the Manager Tasks tab says "Suspended - unknown reason" instead of "Podman initializing"

Testing for success may not tell you what you want to know because `podman machine start` fails if the VM is already running, as mentioned [here](https://docs.podman.io/en/stable/markdown/podman-machine-start.1.html). Once the VM is running, if you quit and relaunch BOINC, then stderrdae.txt shows _Error: unable to start "podman-machine-default": already running_. Despite that, the client still reports _Podman VM initialized_, because _Run_Podman_ currently always returns a zero, even though it also calls `waitpid`. 

I just did some testing (adding temporary printf statements to _Run_Podman_) and discovered that `podman machine init` and `podman machine start` each return a zero exit code when the VM did not already exist, but `podman machine init` returns exit code 125 (0x7D) if the VM already existed, and `podman machine start` returns exit code 125 (0x7D) if the VM was already running.

I don't know if you might want to modify _Run_Podman_ to return the `status` value from its call to `waitpid`, or exit with that `status` value if not 0x7D. But it's unclear what if any exit code `podman machine start` will return for other errors, though I would be a bit cautious since the Podman documentation doesn't say anything about error codes or exit codes.

CharlieFenton:
@davidpanderson To minimize BOINC's impact on volunteers' use of their computers, it would be good to stop the VM whenever BOINC is suspended for user activity and resume it after the specified period of inactivity. But this is not practical because `podman machine start` takes too long. I propose that, at least on APPLE, we instead  call `kill(VMPid, SIGSTOP)` and `kill(VMPid, SIGCONT)` where _VMPid_ is the PID of the process _com.apple.Virtualization.VirtualMachine_. Since you already have code that finds that process in `boinc_related_cpu_time()`, it would take almost no additional code or CPU time to get its PID.

If you think this is a good idea, would you add that code to this PR?

One possible issue with pausing the VM immediately might be the delay before a project task (in this case the docker_wrapper ?) obeys the client's  request to suspend or exit, but I don't know if that would actually cause problems or if any problems could be dealt with in the docker-wrapper.

CharlieFenton:
> I don't know if you might want to modify Run_Podman to return the status value from its call to waitpid, or exit with that status value if not 0x7D. But it's unclear what if any exit code podman machine start will return for other errors, though I would be a bit cautious since the Podman documentation doesn't say anything about error codes or exit codes.

Another approach might be to see if _com.apple.Virtualization.VirtualMachine_ is running. And as I wrote [here](https://github.com/BOINC/boinc/pull/6557#issuecomment-3317490859):
> Since you already have code that finds that process in `boinc_related_cpu_time()`, it would take almost no additional code or CPU time to get its PID.

CharlieFenton:
@davidpanderson Do you have any reaction to m comments [here](https://github.com/BOINC/boinc/pull/6557#issuecomment-3315971461), [here](https://github.com/BOINC/boinc/pull/6557#issuecomment-3317490859)  and [here](https://github.com/BOINC/boinc/pull/6557#issuecomment-3317507830)? Do you plan to act  on them?

It would be good to finish this so we can release a new version for the Mac. I have other changes for the mac I would like to get released.

davidpanderson:
I fixed the "Unknown reason" issue.

I don't think we should stop and start the VM at this point.
It's one more thing that can go wrong, and I don't think users will know or care about the CPU load.
I think we should most as fast as possible to use Apple's new native container system,
which presumably won't have this issue.

Machine init/start in a separate process should work AFAIK.

Unless there's something else let's do a Mac client test release.

CharlieFenton:
> Unless there's something else let's do a Mac client test release.

@davidpanderson @AenBleidd There is just one more thing I've been thinking about. Should we give our VM a different name than the default, to help avoid conflicts for users (now or in the future) who also use Podman or Docker for non-BOINC work? If we wait and decide to change the VM name in the future, the transition will be much messier than if we do it now before it gets wide use in BOINC.

Creating and using a VM with a non-default name is not difficult; it just requires adding another argument to many of the Docker / Podman commands. What do you think?

CharlieFenton:
> I think we should most as fast as possible to use Apple's new native container system, which presumably won't have this issue.

I think that is already available under the new MacOS 26 which was released on September 15. [According to Apple](https://github.com/apple/containerization):
> Containerization is a Swift package for running Linux containers on macOS.

It runs only on ARM Macs, so I presume it will run only ARM Linux apps. Swift is Apple's proprietary programming language; most APIs to date have had both Swift and Objective-C versions. But it appears that this container system is available only for Swift code. 

I do not have any experience in, or significant knowledge of, programming in Swift and have absolutely no desire to learn. I also have no interest in learning to use this new containerization system. So I'm afraid you'll have to find someone else to do that.

CharlieFenton:
I forgot to mention that the Apple GitHub page I referenced in [my previous comment](https://github.com/BOINC/boinc/pull/6557#issuecomment-3343649942) has a link in the upper right corner of the page that leads to Apple's documentation for this containerization package.

AenBleidd:
@CharlieFenton,

> But it appears that this container system is available only for Swift code.

Are you sure about that?
From what I read, this is a package similar to docker/podman.
So for us it doesn't matter what language it was written with, we can take it as a black box and use available cli parameters to use it, since they should have 100% support of the `Dockerfile` format.

AenBleidd:
> Should we give our VM a different name than the default, to help avoid conflicts for users (now or in the future) who also use Podman or Docker for non-BOINC work?

This is really interesting. How much overhead it will make to run two copies of the VM (one run by the user, another one run by the BOINC)? I'd like to have this type of isolation if it doesn't add any sigificant complexity.

davidpanderson:
Given our limited resources, we should probably focus on using Apple's native containers
(which run each container in a separate lightweight VM)

davidpanderson:
It's unclear from Apple's page whether this is an experimental thing for developers,
which you have to install and configure,
or whether at some point it will be built into the OS.

davidpanderson:
My impression is that it offers a CLI interface (like Docker and Podman)
so it doesn't matter what language they implemented it in

CharlieFenton:
Back on June 11, @AenBleidd wrote about the (at that time upcoming) Apple containerization package with a [link](https://github.com/apple/container) to sample code and a tutorial from Apple. It requires an Apple Silicon (arm64) Mac running MacOS 26, which was released 2 weeks ago. So it will work only for BOINC users who have a Mac fulfilling those requirements, which will be most of the Mac users sooner or later. But it won't work for Mac users who have repurposed their old Intel Macs to run BOINC.

> It's unclear from Apple's page whether this is an experimental thing for developers, which you have to install and configure, or whether at some point it will be built into the OS.

The capability is built into the OS, but we will need to add code written in Swift to implement it. It is my understanding that you can link Swift code with C++ and Objective-C code, but it takes a bit of effort to make it compatible. I recommend one of you get an arm64 Mac and download and build the sample code and study the tutorials in the link I reference above.

That page links to a [technical overview ](https://github.com/apple/container/blob/main/docs/technical-overview.md) that seems to say that a separate VM will be run for each container. Does that mean that i f we have 4 tasks running there will be 4 VMs? This seems wasteful of resource compared to Podman which shares a VM among containers.

At this point, we don't know if it will work for our needs, so the first step will be to get familiar with it. I am not willing to do this, especially not now. I feel my life has been consumed with BOINC stuff for the last several moths, and I need to take an extended break. I'm willing to do small maintenance tasks like building a release, but no new development.

CharlieFenton:
> Given our limited resources, we should probably focus on using Apple's native containers (which run each container in a separate lightweight VM).

A the point, we don't even know whether Apple's containerization is suitable for our needs. Meanwhile, the small amount of time to give our VM  unique name instead of using the default is a small step to help avoid potential issues in the future, and can probably be done much more easily before the rollout than as a retrofit later. If you do this, I am willing to take the time to test the revised code on the Mac. Let's get this right the first time.

AenBleidd:
@CharlieFenton, according to my understanding (I might be wrong), you don't need to mess with the Swift code unless you want to interact with the containerization technology directly.
But if we take the same way we do with docker/podman, then we just need to ask users to install [this package](https://github.com/apple/container) that is basically the same as docker/podman (and even shared the same cli interfact (here I again might be wrong, but on the quick glance the commands are the same)).
But this required deeper investigation, of course

davidpanderson:
We don't understand Podman VM names.  The docs talk about 'the Podman VM', like there's only one of them.  What if the user has created a Podman VM, and we create one with a different name?

None of the Podman commands like create and run take a VM name argument,
so it doesn't make sense to have multiple Podman VMs.

Also, I'll be that the number of BOINC Mac users who use Podman for something else is zero.

CharlieFenton:
@AenBleidd You may very well be correct. I don't know.

@davidpanderson @AenBleidd But bear in mind that the current version of Apple's containerization package is 0.41. So this may still be very much under development. Do you rally want to make BOINC depend on a technology that is not even version 1.0 yet? 

> But if we take the same way we do with docker/podman, then we just need to ask users to install [this package](https://github.com/apple/container)

There is also a separate ["containerization" page](https://github.com/apple/containerization) with other code. I am unclear about the difference, though perhaps this page provides a wrapper for the code on [this "container" page](https://github.com/apple/container). It's not completely clear to me whether basic "container " software provides a Linux kernel, though the [tutorial](https://github.com/apple/container/blob/main/docs/tutorial.md) seems to imply that it does offer a default VM or you can use your own. So it appears that, like Podman, the Apple's basic "container" software requires downloading a VM after first installing Apple's "container" software.

(It is unfortunate that Apple has decided to name  its containerization software "container" which makes it awkward to discuss it without confusing references to that software with references to the actual containers it supports.)

In any case, this will require some significant study and playing with the software before deciding whether it is suitable for BOINC's use, and before deciding the best way to implement for BOINC. I repeat my earlier recommendation that one of you needs to obtain an Apple Silicon (arm64) Mac running MacOS 26 and try out this software. But it will not be me.

@davidpanderson You can [buy](https://www.apple.com/shop/buy-mac/mac-mini) a new Mac Mini with the latest M4 processor for $599 and may be able to get credit for trading in your old one, though I recommend paying the extra $200 to get the larger SSD.

CharlieFenton:
> We don't understand Podman VM names. The docs talk about 'the Podman VM', like there's only one of them. What if the user has created a Podman VM, and we create one with a different name?

According to the Podman [documentation](https://docs.podman.io/en/stable/markdown/podman-machine-init.1.html): 
> The default machine name is podman-machine-default. If a machine name is not specified as an argument, then the new machine will be named podman-machine-default.

You can specify a different name than _podman-machine-default_ for the default VM with the _name_ arguments in `podman machine init [options] [name]` and `podman machine start [name]`. [This page](https://docs.podman.io/en/stable/markdown/podman-machine-start.1.html) says:
> Only one Podman managed VM can be active at a time. If a VM is already running, podman machine start returns an error.

To switch active VMs, you stop one VM and start the other. All commands which don't have a way to specify the VM name apply to the currently active VM.

 Hopefully, any user knowledgeable enough to be running Podman jobs outside of BOINC will understand the need to stop and start the VMs as needed. Further down the line, we can add code to optionally do that within BOINC (perhaps via a cc_config.xml setting.) But if we don't change our VM name now, it will be much harder to retrofit it for existing BOINC installations later.

CharlieFenton:
As far as I have been able to determine, the only Podman commands which take a machine name as an argument are `podman machine xxx`. So  only 2 lines in `HOST_INFO::get_docker_version_aux()` in _client_hostinfo_unix.cpp_ need to be adjusted to do this. It is a trivial change indeed.  I suggest we name our VM _BOINC-VM_.

As long as no user is also using Podman for non-BOINC work, the VM does not need to be stopped and restarted. So there is currently no other change needed to the code or to the usage instructions from hat we already have.

davidpanderson:
The question: what does it accomplish to use a different name, and how does it affect existing non-BOINC Podman usage

CharlieFenton:
I ran some tests and found that it is possible for BOINC to have a Podman VM with the default name _podman-machine-default_ at the same time as my user has a separate Podman VM with the default name _podman-machine-default_. They can run different jobs at the same time apparently without interfering. I had Podman run th hello-world sample app while BOINC was running Podman tasks from the test server and there was no problem doing both simultaneously.

I believe this works because we have configured Podman for BOINC to use non-standard directories for all its files (the _BOINC podman_ directory) and we run Podman as user _boinc_project_.

So I have concluded there is no need to give BOINC's VM a non-standard name.

However, the test you currently have for determining whether the VM stated successfully doesn't work a I explained [here](https://github.com/BOINC/boinc/pull/6557#issuecomment-3315971461). I'll fix that.

AenBleidd:
@CharlieFenton, @davidpanderson, here's a short but quite good inroductionary video into apple containers: https://youtu.be/CYd9MC3L11o?si=HftStTr5KJ-tghBx

CharlieFenton:
I have updated the test for successful initialization and start of the VM, but I am seeing some strange behavior in the Manager, which may be unrelated to the podman changes. I need to investigate further. Please do not merge yet.

I have also seen some inconsistent errors from the test server tasks, which may or may not be related to running a separate VM in another user. I may want to investigate that further as well.

CharlieFenton:
The strange behavior the BOINC Manager apparently is due to a problem with the latest Xcode26. I rebuilt with Xcode16 and all is fine again.

> I have also seen some inconsistent errors from the test server tasks, which may or may not be related to running a separate VM in another user. I may want to investigate that further as well.

i have now run more tests, running the hello-world roman example app constantly in a loop while running jobs from the BOINC test server and everything worked correctly. So I am now quite confident that there is no need to give BOINC's VM a non-standard name.

As far as i am concerned, this PR is ready to merge and release. Let's get this, along with other improvements I have made to the mac client, into alpha testing!

AenBleidd:
@CharlieFenton, thank you for your hard work.
I'll check this week what do we have for this release and will send you a separate email.
We do have some other improvements ready for other platforms, so it might be reasonable to make a release for Linux and Windows (and probably Android) also.
In any case, I'll send you am email hopefully today.

CharlieFenton:
@davidpanderson Please look at these error messages from run [112669](https://boinc.berkeley.edu/test/result.php?resultid=112669). Noe that despite these messages, the task completed and verified successfully.

This message occurs 21 times:
> Error: unable to get list of containers: unable to look up container {{.MemUsage}}: no container with name or ID "{{.MemUsage}}" found: no such container

This message occurs 4 times:
> Error: "exited" is not running, can't pause: container state improper

This error occurs 3 times:
> Error: "ad03738243b43e8668e0b03e8f275517b0fc39894f1cdab6f356aed16cf6d090" is not paused, can't unpause: container state improper

CharlieFenton:
> @davidpanderson Please look at these error messages from run [112669](https://boinc.berkeley.edu/test/result.php?resultid=112669). Noe that despite these messages, the task completed and verified successfully.

I believe fixing these would only affect the _docker_wrapper_ or BUDA job submission and so should not delay client release.

davidpanderson:
Fixed in #6582

PR #6558, opened 2025-09-19
web (job submission): make output file download more secure

AenBleidd:
In the scheme for downloading output files, we were passing a path in a URL.
We checked for '..' but there may be other vulnerabilities.
Instead, pass the result ID and file index, which are ints.
Don't pass paths in URLs.


AenBleidd:
@davidpanderson, please check if any of the Copilot comments is applicable

davidpanderson:
Done

PR #6560, opened 2025-09-19
web (job submission): 'only own computers' fixes

davidpanderson:
- It's implemented by passing --target_user to create_work in the job submission script. This was being done for autodock, but not BUDA.

- It's enforced in the scheduler, but only if <enable_assignment/> is set in config.xml. If the user tries to set 'only my computers' and this is not enabled, show an error page.

Also: don't use <li> without enclosing </ul>;
it produces ugly flush-left layout.

### DIFF
--- a/html/user/buda_submit.php
+++ b/html/user/buda_submit.php
@@ -213,7 +213,7 @@ function stage_input_files($batch_dir, $batch_desc, $batch_id) {
 // Use --stdin, where each job is described by a line
 //
 function create_jobs(
-    $app, $app_desc, $batch_desc, $batch_id, $batch_dir_name,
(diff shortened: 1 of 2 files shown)


html/user/buda_submit.php
```diff
@@ -252,6 +252,9 @@ function create_jobs(
     if ($keywords) {
         $cmd .= " --keywords '$keywords'";
     }
+    if ($user->seti_id) {
```
AenBleidd, line 255: is `seti_id` a correct value?
you check for it but later pass just `id`

davidpanderson: We use user.seti_id as a boolean flag for 'use only my computers'.

PR #6568, opened 2025-09-25
Server: fix DB purge for batch-oriented apps

davidpanderson:
The BOINC back-end architecture was originally designed
for stream-oriented apps, in which jobs are processed, assimilated,
file-deleted, and then purged from the DB.

Recently we've added support for batch-oriented apps,
for which we don't want to purge the WU and result records
until the batch has been retired.

Fix: add an option --retired_wus to db_purge.
If set, it purges only WUs and results in retired batches.
(If not set, it purges only WUs that are not in a batch;
so if your project has both stream- and batch-oriented apps,
run one of each).

Also: fix the logic for retire_batch().
It shouldn't set WUs to FILE_DELETE_DONE.
WUs may be in progress, and we still need to file-delete them.
Instead, mark unsent WUs as WU_ERROR_CANCELLED so we don't send them.

Also: add a script ops/delete_batches.php which deletes
retired batches with no WUs (i.e. they've all been purged).
This improves the efficiency of db_purge.
You don't need to do it very often - once a week is enough.

Also: aborting a batch sets its WU error masks to CANCELLED,
in which case the WUs won't go through file-deletion
and their input files will be orphaned.
Fix: when aborting a job, set the WU file delete state to READY.
Similar for unsent jobs when you retire a batch.

### DIFF
--- a/db/boinc_db_types.h
+++ b/db/boinc_db_types.h
@@ -407,17 +407,22 @@ struct HOST_DELETED {
     void clear();
 };
 
-// values for file_delete state
+// values for
+// WORKUNIT::file_delete_state
+// RESULT::file_delete_state
+// TRANSITIONER_ITEM::res_file_delete_state
 // see html/inc/common_defs.inc
+//
 #define FILE_DELETE_INIT        0
 #define FILE_DELETE_READY       1
-    // set to this value only when we believe all files are uploaded
+    // WU is assimilated and all results are OVER,
+    // so we don't need output files anymore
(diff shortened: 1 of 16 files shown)


AenBleidd:
@davidpanderson, ignore failing CIs, there are some issues on the docker side: https://www.dockerstatus.com/
I'll restart affected tasks when the incident is resolved.
Sorry for the incinvenience.

AenBleidd (approved):

Issue #5305, opened 2023-07-07
Battery charge level 0.0% < threshold 31.0%


JoseAngel-MAD:
**Describe the bug**
I get an error message every time a task is executed and the screen is off:

battery charge level 0.0% < threshold 31.0%`
Suspending computation - battery low

**Steps to reproduce**
1. Set a minimum battery threshold.
2. Execute a task.
3. Turn off the screen.

**Expected behaviour**
The app won't stop computation as long as the battery is not lower than the battery threshold.

**System Information**
Samsung Galaxy A52s 5G - Snapdragon 778G - One UI 5.1 (Android 13)
Boinc 7.22.2

**Additional Context**

I had been using version 7.18 of Boinc for several months and the bug was already present sometimes. I decided to update to the newest version, hoping that the issue would be resolved, but now it's even worse than before because the bug is constant.
Now I get some strange messages on the app interface such as: 'The app will continue with the computation when the battery reaches 31% (now 56%)'

Here are the logs from the client and the GUI:

CLIENT LOG

5 jul 2023 9:35:13||(battery charge level 0.0% < threshold 31.0%

5 jul 2023 9:35:13||Suspending computation - battery low

5 jul 2023 9:35:13||Checking presence of 57 project files

5 jul 2023 9:35:13||Setting up GUI RPC socket

5 jul 2023 9:35:13|World Community Grid|URL http://www.worldcommunitygrid.org/; Computer ID 8714668; resource share 0

5 jul 2023 9:35:13|Rosetta@home|URL https://boinc.bakerlab.org/rosetta/; Computer ID 6257239; resource share 100

5 jul 2023 9:35:13|BOINC Central|URL https://boinc.berkeley.edu/central/; Computer ID 6283; resource share 100

5 jul 2023 9:35:13||Checking active tasks

5 jul 2023 9:35:13||Setting up project and slot directories

5 jul 2023 9:35:13||-  (to change preferences, visit a project web site or select Preferences in the Manager)

5 jul 2023 9:35:13||-  max disk usage: 62.66 GB

5 jul 2023 9:35:13||-  Store up to an additional 0.50 days of work

5 jul 2023 9:35:13||-  Store at least 0.10 days of work

5 jul 2023 9:35:13||-     Suspend if no input in last 60.000000 minutes

5 jul 2023 9:35:13||-     max memory usage: 5.24 GB

5 jul 2023 9:35:13||-     suspend if non-BOINC CPU load exceeds 100%

[...]

PR #6606, opened 2025-10-17
client (Android): fix battery check logic

davidpanderson:
We were doing battery charge/temp checks on startup, before getting an RPC from the GUI giving us battery info.

Also: check for user-active suspend before battery checks. We had moved it after on Android, not sure why.

Also: put all DEVICE_STATUS stuff in #ifdef ANDROID


AenBleidd:
Hopefully this will fix #5305

dchmelik:
still happening on 8.2.8

PR #6657, opened 2025-11-11
Menu items sync on skin change

### LINKED ISSUE #6101: [Manager] [Simple VIew] When changing skins, menu text does not change with updated skin
### Describe the bug

Under simple view, when you change the skin (for example, Charity Engine), the text for the app header changes to the skin.  The same also happens when you open other windows such as "about....", "Event Log", "Computing Preferences", etc.

However, under all the menus, the text does not update.  It shows the text of the skin that initially was loaded when the Manager was opened.  This only will be updated if the Manager is closed, then reloaded.

### Steps to reproduce

1. Open Boinc Manager, switch to simple view.
2. Change a skin.
3. Open a menu.  The skin that was used when opening the Manager should be referenced.

### Expected behavior

The menu should be updated to reflect the skin chosen.

### Screenshots

Image

### System information

- OS:  Ubuntu 24.04, but also Windows 10.  For some reason, skins are only available if I build a pre-release version for Windows.
- BOINC version:  8.0.2
- wxWidgets Version:  3.2.5 (Ubuntu) and 3.2.6 (Windows 10 building a debug version)

### Additional context

_No response_

AufarZakiev:
Fixes #6101

**Description of the Change**
On skin change, menus are recreated

image

**Alternate Designs**
More granular update could be implemented, but it feels like overengineering

**Release Notes**
Client: Fix menu items not synced with a chosen skin

### DIFF
--- a/clientgui/sg_BoincSimpleFrame.cpp
+++ b/clientgui/sg_BoincSimpleFrame.cpp
@@ -747,6 +747,7 @@ void CSimpleFrame::OnReloadSkin(CFrameEvent& WXUNUSED(event)) {
     wxASSERT(pSkinAdvanced);
 
     m_pBackgroundPanel->ReskinInterface();
+    CreateMenus();
     SetTitle(pSkinAdvanced->GetApplicationName());
     SetIcon(pSkinAdvanced->GetApplicationIcon()->GetIcon(wxDefaultSize));
 


AenBleidd:
@AufarZakiev, thank you for the fix!

PR #6663, opened 2025-11-12
Fix deadlock on WSL removal after installing Docker

### LINKED ISSUE #6296: BOINC.exe gets stuck during startup if WSL is turned off after installing docker
### Describe the bug

BOINC Client won't start properly if WSL is turned off after Docker was installed.

### Steps to reproduce

1. Install BOINC
2. Install Docker (From BOINC docker install instructions):
3. Enable WSL and Virtual Machine Platform, Reboot (Note: Virtual Machine Platform is not in the instructions but is required)
4. Install wsl.2.4.13.0.x64.msi
5. Run boinc-buda-runner.wsl
6. Run "wsl.exe -d boinc-buda-runner" in Windows terminal
7. Run BOINC Manager (WSL/Docker will be listed in log)
8. Close manager and client
9. Disable WSL and Virtual Machine Platform (In Windows features), Reboot
10. Run BOINC Manager. It will be stuck trying to connect to localhost.
11. Open Task Manager
12. Right click on BOINC.exe and select "Analyze wait chain"
13. It will report that BOINC.exe is in deadlock
14. If WSL is installed again, BOINC will start

### Expected behavior

_No response_

### Screenshots

_No response_

### System information

Windows 10 x64 (Latest)
Tested with BOINC 8.0.2 and later

### Additional context

_No response_

AufarZakiev:
Fixes #6296

**Description of the Change**
Prevents BOINC from stucking if WSL was before and then was disabled. BOINC works fine w/o WSL, so it may ignore WSL/distros absence and work as usual. Basically, return 0 instead of -1.

Let me if there could be side effects. Tested on Win 11

**Alternate Designs**
.

**Release Notes**
Fix deadlock on WSL removal after installing Docker

### DIFF
--- a/client/hostinfo_wsl.cpp
+++ b/client/hostinfo_wsl.cpp
@@ -21,6 +21,7 @@
 #include "boinc_win.h"
 #include "win_util.h"
 
+#include "error_numbers.h"
 #include "str_replace.h"
 #include "client_state.h"
 #include "client_msgs.h"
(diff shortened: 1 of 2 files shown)


AenBleidd:
@AufarZakiev, thank you for the fix!

PR #6666, opened 2025-11-13
Add Dark Mode support for Windows

### LINKED ISSUE #5274: [Manager] Add dark mode support on Windows
Current 3.2 version of wxWidgets has no dark mode support for Windows OS.
However, this feature is implemented and currently merged to git master.
Need to verify next release of wxWidgets and, if dark mode for Windows is released, enable it in the code and test BOINC Manager.

AufarZakiev:
Fixes #5274

**Description of the Change**
1. Bumps WxWidgets version, we need higher than 3.3.1
2. Forces common elements to use OS theme
3. Retrieves dark theme flag and carefully repaints
 - Charts (Statistics, Disk panels)
 - Progress bars (to make progress values easily visible)
 - Notices list
4. Progress column alignment changed to `center` to follow progress column alignment in other tabs

Tested on Win 11.

**Technical implementation**
- Modified MyEvtHandler::OnPaint() to skip PostDrawProgressBarEvent() in dark mode
- Implemented two-stage NM_CUSTOMDRAW handling:
  - CDDS_ITEMPREPAINT: Request post-paint notification from ListView
  - CDDS_ITEMPOSTPAINT: Draw progress bars directly on back buffer HDC
- Used raw Win32 GDI for rendering to avoid wxWidgets double-buffering issues

**Alternate Designs**
Proper progress bars rendering requires to introduce hardly-maintainable code (see `DrawItemProgressBar` method). To be honest, it is a legacy code from the moment of creation. 

**Screenshots:**

image
image
image

**Release Notes**
Add Dark Mode support for Windows

**Update history**
- Initial PR
- Fixed progress bars disappearing on hover in Dark Mode


AufarZakiev:
Nice PR number for such a change, btw

AenBleidd:
@AufarZakiev,

> Note: wxWidgets' IsDark() returns false on Win 11 now even if dark mode is enabled. Therefore, registry reading mechanism was introduced as an additional check

For Windows use `IsSystemDark()` function instead:
https://docs.wxwidgets.org/latest/classwx_system_appearance.html#ac6c70858bb3e4c35a0b1af42997b6c42

AufarZakiev:
`AreAppsDark()` fits better, doesn't it?
At least, apps on my desktop react on this value and ignore `IsSystemDark()`

AenBleidd:
> `AreAppsDark()` fits better, doesn't it?

Yes, it looks like this is really better. Good catch!

3rdParty/vcpkg_ports/configs/msbuild/vcpkg.json
```diff
@@ -11,6 +11,7 @@
     "rappture",
     {
       "name": "wxwidgets",
+      "version>=": "3.3.1",
```
AenBleidd: This is not needed as we always build the latest version of wxWidgets for every platform.

clientgui/common/wxPieCtrl.cpp
```diff
@@ -50,18 +50,11 @@ wxPieCtrl::wxPieCtrl(wxWindow * parent, wxWindowID id, wxPoint pos,
 		wxSize sz, long style, wxString name)
 		:wxWindow(parent, id, pos, sz, style, name)
 {
-    bool isDarkMode = wxGetApp().GetIsDarkMode();
-
-    if (isDarkMode) SetBackgroundColour(*wxBLACK);
+  ApplyTheme(wxGetApp().GetIsDarkMode());
```
AenBleidd: We use 4 spaces for tabulation

clientgui/common/wxPieCtrl.h
```diff
@@ -139,6 +139,8 @@ using wxWindow::SetTransparent;
 	/// Sets the colour used for displaying legend box background
 	void SetLegendBackColour(wxColour colour);
 
+  void ApplyTheme(bool isDarkMode);
```
AenBleidd: We use 4 spaces for tabulation

clientgui/BOINCGUIApp.cpp
```diff
@@ -84,6 +84,40 @@ typedef unsigned int (*RsvgErrorQuark) (void);
 typedef void (*RsvgHandleGetPixbuf) (void*);
 #endif
 
+#if SUPPORTDARKMODE && defined(__WXMSW__) && wxCHECK_VERSION(3,3,0)
```
AenBleidd: Remove check version. We don't support building with different versions

clientgui/BOINCGUIApp.cpp
```diff
@@ -84,6 +84,40 @@ typedef unsigned int (*RsvgErrorQuark) (void);
...
+// wxWidgets' IsDark() returns false on Win 11 now even if dark mode is enabled
+// so we need to additional function to check the registry for the dark mode setting
+bool IsWindowsInDarkMode() {
```
AenBleidd: The whole function should be replaced with just the `AreAppsDark()` call as you suggested.

clientgui/BOINCGUIApp.cpp
```diff
@@ -108,6 +142,13 @@ bool CBOINCGUIApp::OnInit() {
 #if SUPPORTDARKMODE
     wxSystemAppearance appearance = wxSystemSettings::GetAppearance();
     m_isDarkMode = appearance.IsDark();
+// Dark theme only makes sense on Win if WxWidgets has version >3.3
+#if defined(__WXMSW__) && wxCHECK_VERSION(3,3,0)
```
AenBleidd: Remove version check

clientgui/BOINCGUIApp.cpp
```diff
@@ -108,6 +142,13 @@ bool CBOINCGUIApp::OnInit() {
...
+    // wxWidgets' IsDark() returns false on Win 11 now even if dark mode is enabled
+    // so we need to additionally check the registry for the dark mode setting
+    m_isDarkMode = m_isDarkMode || IsWindowsInDarkMode();
+    MSWEnableDarkMode(wxApp::DarkMode_Auto);
```
AenBleidd, line 114: Are you sure this is needed? I do remember when I tested the dark mode, I didn't have this enabled

clientgui/BOINCGUIApp.h
```diff
@@ -34,14 +34,16 @@
 #define BOINC_ADVANCEDGUI                   1
 #define BOINC_SIMPLEGUI                     2
 
+#include <wx/version.h>
```
AenBleidd: remove this as we don't support older wxwidgets versions

clientgui/BOINCGUIApp.h
```diff
@@ -34,14 +34,16 @@
...
+
 // NOTE: MacOS automatically adjusts all standard OS-drawn UI items
 // in Dark Mode, so BOINC must not modify their colors for Dark Mode.
 // For MacOS, we adjust Dark Mode colors only for our custom UI items.
 // If you implement Dark Mode support for another OS which requires
 // BOINC to adjust standard UI items for Dark Mode, be sure to guard
 // those changes so they do not affect the Mac implementation.
 //
-#if (defined(__WXMAC__) || defined(__WXGTK__))
+#if (defined(__WXMAC__) || defined(__WXGTK__) || (defined(__WXMSW__) && wxCHECK_VERSION(3,3,0)))
```
AenBleidd: remove version check

clientgui/NoticeListCtrl.cpp
```diff
@@ -34,6 +34,48 @@
 #include "MainDocument.h"
 #include "NoticeListCtrl.h"
 
+namespace {
```
AenBleidd: Don't use namespace here.

clientgui/NoticeListCtrl.cpp
```diff
@@ -34,6 +34,48 @@
...
+
+#if wxUSE_WEBVIEW
```
AenBleidd: could you please show using some screenshots why this is needed?
code we currently have for dark and light modes displays everything correctly

AenBleidd (changes requested):

clientgui/common/wxPieCtrl.h
AufarZakiev: Ah, i see the trick now! I spent half an our trying to understand why it looks different in my IDE

CharlieFenton:
@AufarZakiev @AenBleidd Please make absolutely sure that **all** your changes are guarded by `#ifdef _WIN32` to prevent creating problems for Mac and Linux.

CharlieFenton:
> @AufarZakiev @AenBleidd Please make absolutely sure that all your changes are guarded by #ifdef _WIN32 to prevent creating problems for Mac and Linux.

or  `#ifdef __WXMSW__`

AenBleidd:
@CharlieFenton, of course we will, no worries. This is too far from being finished, and it's too early to talk about this now, but this is definitely something I will keep in mind while doing a review of the code.

clientgui/BOINCGUIApp.cpp
AufarZakiev: Did it according to docs. Yes, without it just ignores OS settings

clientgui/NoticeListCtrl.cpp
AufarZakiev: Yes, this part only changes code structure to remove duplication and mixing `if` with `#if`. Original code also looks for Webview support and acts accordingly

AenBleidd: Yes, this is true but we don't embed so much styles as you do. Why all these additionas you implemented are required?

clientgui/common/wxPieCtrl.h
AufarZakiev: image

Well, there is a mix of tabs (denotes as arrows) and spaces (denoted as points) in different files. I can fix it but it feels like it deserves another PR with some linting

AenBleidd: No, this is not required at the moment.
The biggest issue with this class is that it was copied a couple of decades ago from the original wxWidgets code and later modified.
Now we have an outdated and self-patched version of this class (and some others, they could be easily identified by the copyright header) and at the same time updated and well-supported official implementation by wxWidgets team.
At one moment we need to check what was manually implemented in this class, completely remove this class from the codebase to use the one provided in the wxWidgets library and implement required changes in one of our classed to keep the existing functionality.
If this is something that looks interesting to you (but I want to warn you in advance sine this will require a lot of research) - please let me know, and I'll help you with it.
For now keep the tabulations are they are, and don't fix those lines you don't specifically touch.

clientgui/NoticeListCtrl.cpp
AufarZakiev: Reverted this part. Too hard too follow the idea, that's right

AufarZakiev:
@AenBleidd , do you have a chance to review PR?

AenBleidd:
I had no chance to test this on devices I have.
Hopefully will be done during this weekend.

AenBleidd:
@AufarZakiev, we have a progress bar that has a self-written function to draw itself.
It blinks when in dark mode (see the video).
This is something that need to be fixed, because otherwise very annoying.
I have tested this on Win11 and Win10 (video was recorded on Win11, but on Win10 blinking is even more visible).
After this is fixed, I will retest it including Linux.
https://github.com/user-attachments/assets/0e415283-8840-48e5-8f48-4a841a5ed6be

clientgui/ViewStatistics.cpp
CharlieFenton, line 1: Why are you changing all these colors? The contents of the graphs do not need to be changed for dark mode. In my opinion, it is wrong to change them. Furthermore, you have not guarded your changes by `#ifdef _WIN32` or `#ifdef __WXMSW__` to prevent creating problems for Mac and Linux. This file already had all the changes needed for dark mode. Here is how the light and dark mode appear on the Mac with the existing code:

Statistics tab light mode

Statistics tab dark mode

clientgui/BOINCListCtrl.cpp
CharlieFenton, line 1: You are making wholesale changes affecting all platforms, including Mac and Linux. This is unacceptable. You **must** take care not to change anything affecting the Mac by guarding your changes with `#ifdef _WIN32 `or  `#ifdef __WXMSW__`

CharlieFenton, line 1: Please note also that on MS Windows, BOINC uses the native list control, but on Mac and Linux it uses wxWidgets' _wxGenericListCtrl_ built from wxWidgets source code.

clientgui/ViewStatistics.cpp
AenBleidd: @CharlieFenton, maybe we still can change these colors to be less bright when in Dark mode?
Personally I think with this change they look (at least on Windows) way more consistent than on macOS:
image

CharlieFenton: Thank you for the screenshot. Now that I've seen it, I feel even more certain that the changes in this PR are wrong. It looks terrible and it is very difficult to see the grid lines. You can do what you want on Windows, but please do not change the appearance on the Mac. The BOINC dark mode has been available on the Mac for quite some time, and people have been happy with it as is.

AenBleidd: It would be better to have the same look and feel on all platforms. Maybe we can adjust the intensity of the grid lines to make them more visible while keeping more dark background?
@davidpanderson, what do you think?

davidpanderson:
I agree that, in dark mode, the background of the graphs should be fairly dark
(but if this is a problem on the Mac, the light background is OK)

CharlieFenton:
I think it is worthwhile to take a step back and look at what Microsoft and Apple say about Dark Mode.

[This article](https://www.microsoft.com/en-us/windows/learning-center/when-to-use-dark-mode)  from Microsoft says:

> Light mode is typically the standard setting for most computers, but lots of users are making the switch to Dark mode for a variety of reasons. Here are some benefits to turning on Dark mode:
**Aesthetic preference**: A common reason why people prefer Dark mode is because it aligns with their aesthetic better than Light mode. If you prefer darker colors over lighter ones, Dark mode will likely fit your style.
**Easier on the eyes**: Dark mode is often credited for reducing eyestrain, which is a syndrome that can occur if you look at a bright screen for too long. Turning on Dark mode may also make the letters easier to read.
**Better for darker locations**: A bright screen in a dark room can be harsh on the eyes and could potentially bother those around you. Using Dark mode in a dimly lit airplane cabin or a shared bedroom will help reduce the brightness of your screen to make it stand out less.
**Preserves battery**: Does Dark mode save battery life? You bet!

I've seen some websites say _Reduce eyestrain_  rather than _Easier on the eyes_. I think that is primarily about reducing the glare that comes from black letters on a white background. It's pretty clear to me that Dark mode doesn't necessarily involve inverting all colors. If you scroll down a bit in [this section of this microsoft article](https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/ui/apply-windows-themes#know-when-dark-mode-is-enabled) you can see that a color quite similar to our graph background is unchanged in dark mode.

When I asked Google for _Microsoft dark mode programming guidelines_ it replied:
> Microsoft's dark mode guidelines focus on using dark grays for backgrounds, lighter surfaces for importance (elevation), avoiding pure white/saturated colors, ensuring accessibility (contrast, focus indicators)

and

> **Elevation**: Use lighter shades of gray for more important surfaces (like dialogs) and darker grays for less important ones to create depth.
**Avoid Pure White**: Use dark grays for backgrounds and avoid pure white (#FFFFFF) for text/elements to reduce eye strain in dark mode.
**Don't Invert**: Don't simply invert colors; design specifically for dark mode for optimal results.
**Accessibility**: Ensure focus indicators (like blue outlines) remain visible. 

Another important consideration I see mentioned in the above articles and various others is _accessibility_. Contrast among elements must be high enough to help people with impaired vision understand the display. That means the grid lines must be more prominent than the screenshot from this PR. I feel the pale blue background we have in Light Mode also works well in Dark Mode and allows the grid lines and data to be easily visible without causing eye strain or glare.

I am willing to consider using the colors that we settle on for Windows in the statistics tab on the Mac if they are pleasing (_aesthetic preference_) and satisfy all the other considerations discussed above. But I feel the current proposal does not meet those criteria.

Finally, please remember that BOINC has supported Dark Mode on the Mac since June 2023 and the current implementation has been very well received.

AenBleidd:
> Finally, please remember that BOINC has supported Dark Mode on the Mac since June 2023 and the current implementation has been very well received.

That doesn't mean that it's perfect and cannot be made better ;)

> If you scroll down a bit in [this section of this microsoft article](https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/ui/apply-windows-themes#know-when-dark-mode-is-enabled) you can see that a color quite similar to our graph background is unchanged in dark mode.

True, but the color itself is a little bit more soft and darker in comparison with the color we have in BOINC Manager.

But I do agree. I do not want to say that the set of colors in this PR is the final.
I'm open for a discussion, and we need to find the solution that will be satisfiable for all of us.
Personally I don't like graphs background in macOS version because I find it too bright, however on the real macOS screen they might look differently (we all remember that display calibration of macOS devices is one of the best, and other displays usually show slightly different colors.

This needs to be further discussed to find better colors for the next try.

AenBleidd:
@CharlieFenton, some examples of graphics from GitHub I personally like:
image
image
I'm not pushing to use them as a guideline, however I like them because they don't have big areas that are too bright or too dark in comparison with the other surrounding areas.

Again, need to be further discussed.

AenBleidd:
One more example of nice graphs: https://grafana.kiska.pw/d/boinc/boinc?orgId=1&var-project=BOINC%20Central&from=now-24h&to=now

AufarZakiev:
FYI: need more time to deal with progress bars blinking. Still a mystery why it happens and only in the dark mode.

CharlieFenton:
> progress bars blinking

This is only a wild guess, but might it be associated with your addition of the _remainderColor_, _textColor_, _remainderBrush_ and / or _remainderPen_? This doesn't happen on the Mac, but as I [wrote before](https://github.com/BOINC/boinc/pull/6666/files#r2596289385):
> on MS Windows, BOINC uses the native list control, but on Mac and Linux it uses wxWidgets' wxGenericListCtrl built from wxWidgets source code.

You can test whether it blinks when using the generic list control by temporarily changing lines 25-29 in _BOINCListCtrl.h_.

AufarZakiev:
> This is only a wild guess, but might it be associated with your addition of the remainderColor, textColor, remainderBrush and or remainderPen? 

No, reverting these changes do not prevent blinking. 

>You can test whether it blinks when using the generic list control by temporarily changing lines 25-29 in BOINCListCtrl.h.

Generic list is not available in Win, I guess. At least, I cannot build project enabling it.

AufarZakiev:
WxWidgets' developers [directly say](https://github.com/wxWidgets/wxWidgets/issues/26001#issuecomment-3592520142) that 
> In dark mode, the whole list is custom draw

So it is a little hacky from the very beginning. Now they are fixing numerous bugs in dark mode. For example, master branch already has a fix for proper selected row highlight. But our problem is deeper, since BOINC uses a little hacky "post-draw" method to draw progress bars. Using master branch of WxWidgets does not fix it

**Option 1.** I can dig even deeper to find root cause in WxWidgets' dark mode implementation. After that, root cause maybe fixed either in WxWidgets or I will come up with another progress bars drawing method specifically for Win dark mode if it possible. 
**Option 2.** Progress bars could be drawn entirely differently, without post-draw events handling. This could be a good option if progress bars render code is legacy. A good moment to rewrite it. 
**Option 3.** Give up and wait for WxWidgets to be fixed internally. 

I think, we should try Option 1. @AenBleidd , do you agree?

AenBleidd:
@AufarZakiev, yeah, I agree. Option 1 is preferred.
Thank you for your detailed research.

AufarZakiev:
@AenBleidd , fix is finally there. I updated the initial PR message
The root issue and fix was found by Claude Opus 4.6. It works but the code is hardly maintainable (by humans, ofc).

I asked for very comprehensive comments, they are really good. Here is a main problem description, it is also in the code itself.
<details>
<summary> WxWidgets' Dark mode is a nightmare </summary>

```
// WHY THIS IS NEEDED:
//
// BOINC's list views (Tasks, Transfers, Projects) show progress bars
// inside a wxListCtrl column. These aren't native progress bar controls --
// they are custom-drawn rectangles with text on top, painted by
// DrawProgressBars() above.
//
// In light mode, the rendering flow works like this:
//
//   1. Windows sends WM_PAINT to the ListView.
//   2. wxWidgets' MyEvtHandler::OnPaint() fires (we pushed it onto the
//      event handler chain in the constructor). It calls event.Skip()
//      to let the native ListView paint itself, then posts a custom
//      wxEVT_DRAW_PROGRESSBAR event via PostDrawProgressBarEvent().
//   3. That posted event is processed after WM_PAINT completes, calling
//      DrawProgressBars(), which uses wxClientDC to draw directly onto
//      the screen surface -- on top of whatever the ListView just painted.
//
// This works fine in light mode because the native Win32 ListView paints
// directly to the screen, so our wxClientDC drawing persists until the
// next WM_PAINT.
//
// In dark mode, wxWidgets enables full owner-draw rendering for the
// ListView via MSWEnableDarkMode(). This changes the painting pipeline:
//
//   1. wxWidgets intercepts NM_CUSTOMDRAW notifications from the ListView.
//   2. For each item, wxWidgets' HandleItemPaint() calls FillRect() to
//      paint the entire row background, draws the item text, and returns
//      CDRF_SKIPDEFAULT to suppress the native theme rendering.
//   3. Critically, all of this drawing happens on a BACK BUFFER HDC --
//      the ListView uses double-buffering (LVS_EX_DOUBLEBUFFER) to
//      eliminate flicker. The back buffer is blitted to the screen only
//      after the entire NM_CUSTOMDRAW cycle completes.
//
// This breaks the old approach: our wxClientDC draws onto the screen
// surface AFTER WM_PAINT, but the back buffer doesn't contain our
// progress bars. The next time the ListView needs to repaint (e.g. on
// hover, which triggers hot-tracking animation via the Explorer theme),
// it blits its back buffer to screen -- erasing our progress bars.
// On hover this happens rapidly, making progress bars flicker or vanish.
```

</details>

I personally do not like adding basically a legacy code (nobody will ever touch it, really). However, I see no other way till the UI is coupled to WxWidgets.

AufarZakiev:
I wonder is it too optimistic to research for a modern multi-platform UI ? For more web-like experience, as it is mentioned in roadmap.
Such huge changes is a place where the AI shines.

AenBleidd:
@AufarZakiev, thank you for the commit. I'll try to review it when I have time.
Regarding the new framework, at the moment our Manager is small (memory and resources), and it rarely touched.
New web-based GUI would be interesting to see, but it will definitely be more huge, and I'm not sure it will bring any significant advantage. Also, we do not currently have resources for its development and testing.

CharlieFenton:
@AufarZakiev Your changes to CBOINCListCtrl::DrawProgressBars() are modifying code that is used by Mac and Linux. This is absolutely unacceptable! I don't have the time to constantly review your code. Please do not let this happen again.

CharlieFenton:
> @AufarZakiev Your changes to CBOINCListCtrl::DrawProgressBars() are modifying code that is used by Mac and Linux.

@AufarZakiev I only looked at your changes long enough to find one instance of unguarded changes which may break Mac and Linux. Please make sure **all** your changes are guarded by #ifdef and affect only Wiindows. Never replace existing code.

davidpanderson:
@AufarZakiev is trying to help.  Let's be encouraging and constructive.

CharlieFenton:
@AufarZakiev One correction / clarification: please guard your code with `#if USE_NATIVE_LISTCONTROL` rather than `#ifdef _WIN32` or  `#ifdef __WXMSW__`. The Native list control mostly works on the Mac (and probably on Linux, though I don't have an easy way to test that), but is inefficient. If wxWodgets improves it enough in the future, we may want to switch to using the native meson in the future. Using `#if USE_NATIVE_LISTCONTROL` leaves that option open.

AufarZakiev:
> *Disclaimer: This comment was drafted with the help of Claude Code to maintain clear English and save everyone's time reading it. The content and opinions are mine.*

Hi @CharlieFenton,

Thank you for your vigilance in protecting the Mac and Linux builds. I want to address your concerns directly

---

### Changes that are strictly Windows-only

The vast majority of this PR is guarded by platform checks and has **zero effect** on Mac or Linux:

- **`BOINCGUIApp.cpp`** — `AreAppsDark()` and `MSWEnableDarkMode()` are inside `#if defined(__WXMSW__)` (these are MSW-specific wxWidgets APIs, not list control related)
- **`BOINCListCtrl.cpp/.h`** — All new code (`MSWOnNotify`, `HandleDarkModeCustomDraw`, `DrawItemProgressBar`, the `PostDrawProgressBarEvent` skip in dark mode) is inside `#if USE_NATIVE_LISTCONTROL`, following your recommendation to use this guard rather than `#ifdef __WXMSW__` so the option remains open for future use on other platforms
- **`BOINCGUIApp.h`** — Adds `__WXMSW__` to `SUPPORTDARKMODE`; Mac and GTK were already included, so no change for them
- **`NoticeListCtrl.cpp`** — Removes a Windows-specific `IsSystemDark()` fallback that was always `false` on Mac/Linux anyway; no behavioral change on other platforms

---

### Changes that do affect all platforms

There are a few small cross-platform changes I'd like to discuss openly:

**1. `ViewWork.cpp` — Progress column alignment**

Progress column header alignment changed from `wxLIST_FORMAT_RIGHT` to `wxLIST_FORMAT_CENTRE`. This is a minor cosmetic tweak affecting the column header text. Would you be OK with this, or would you prefer I guard it?

**2. `ViewResources.cpp` — Bug fix + refactor**

Fixed a copy-paste bug: `m_pieCtrlTotal->SetLabelColour(...)` was being called instead of `m_pieCtrlBOINC->SetLabelColour(...)` when setting up the BOINC disk usage pie chart. This meant the BOINC pie chart's label color was never explicitly set. Also refactored `isDarkMode` from a local variable to a `m_isDarkTheme` member. The dark-mode color values themselves are unchanged.

**3. `wxPieCtrl.cpp/.h` — Refactor only**

Extracted theme initialization into an `ApplyTheme()` method. Same logic, same values — purely a refactor to make it callable after construction.

---

### Question on Statistics graph colors (`ViewStatistics.cpp`)

The color scheme is now split into separate `if (isDarkMode) / else` blocks. The **light-mode values are identical** to what was there before, so Mac/Linux in light mode are unaffected.

For **dark mode**, the colors are adjusted (e.g., darker axis background, brighter graph lines for better contrast on dark backgrounds). I understand your concern about changing the Mac dark-mode appearance. I will be changing these colors on Windows regardless — would you prefer I guard the dark-mode palette with `#ifdef __WXMSW__` so Mac keeps its current colors?

---

### A note on workflow

I notice you've assigned yourself as a reviewer and assignee on this PR. I appreciate the attention, but this PR is still a work in progress — I have not requested a review yet. You've mentioned several times that you don't have time to constantly review my code, and I completely understand that. To save us both time and frustration, I'd suggest removing the self-assignment for now and waiting until I explicitly request your review. That way you won't be reviewing intermediate states that are still changing, and when I do ask for your review, I'll be able to confirm that all platform guards are in place and the PR is ready for your attention.

---

Looking forward to your feedback.

AufarZakiev:
I will also deeply appreciate your feedback on a new BOINC manager called Fresco. I developed it as an experiment and it is pretty good. There is an announcement: https://github.com/BOINC/boinc/discussions/6866

I want to express my admiration on how good manager and client are separated. It was a real pleasure to know how easy is to make a replacement for a manager without touching any of client's code. Magnificent!

CharlieFenton:
> Thank you for your vigilance in protecting the Mac and Linux builds. I want to address your concerns directly

@AufarZakiev Thank you of the detailed response and for your commitment to exercising care not to break the Mac and Linux functionality. As you suggest, i will wait until you let us know your changes aer ready for review before doing anything further.

But I do want to clarify that [this comment](https://github.com/BOINC/boinc/pull/6666#issuecomment-3902861230) was in response to your changes to `CBOINCListCtrl::DrawProgressBars()`. Please correct me if I am mistaken, but your changes throughout this routine are not guarded, specifically your changes to the original lines 527 and, 609-610.

While these do seem to preserve the progress bars' appearance on the Mac in light mode, they substantially change it in dark mode. I don't know what the standards are for dark mode on Windows, but on the Mac dark mode does not mean changing everything to shades of gray, as your changes did with the progress bars. I spent some time selecting a progress bar color for dark mode on the Mac and Linux and I prefer that not be changed on these platforms.

> Progress column header alignment changed from wxLIST_FORMAT_RIGHT to wxLIST_FORMAT_CENTRE.

I don't have a strong feeling either way. On the one hand, you have made it consistent with the header for that column in the Projects and Transfers tabs; on the other hand, it is inconsistent (in all three tabs) with all the other headers which are either right justified or left justified.

AufarZakiev:
Yes, you are right, `DrawProgressBars()` affects Mac colors.

Ok, summary is:
- Do not change colors for Mac at all (both progress bars and charts)
- Change Progress bar alignment to `wxLIST_FORMAT_CENTRE` to keep it at least consistent with other progress columns

Other changes are refactoring / bugfixes / Windows-only. Is it OK with you, @CharlieFenton ?

CharlieFenton:
> Is it OK with you, @CharlieFenton ?

Yes, I think so, but I will again review everything when you say you are ready for me to do that. Thank you.

AufarZakiev:
@CharlieFenton @AenBleidd you are welcome to review

clientgui/BOINCGUIApp.h
```diff
@@ -41,7 +41,7 @@
 // BOINC to adjust standard UI items for Dark Mode, be sure to guard
 // those changes so they do not affect the Mac implementation.
 //
-#if (defined(__WXMAC__) || defined(__WXGTK__))
+#if (defined(__WXMAC__) || defined(__WXGTK__) || defined(__WXMSW__))
```
AenBleidd: `SUPPORTDARKMODE` can be completely removed since now we have support for all desktop OSs, so having it is just a redundancy since it will always be set to `true`

clientgui/BOINCListCtrl.h
```diff
@@ -104,9 +104,15 @@ class CBOINCListCtrl : public LISTCTRL_BASE {
 #if USE_NATIVE_LISTCONTROL
 public:
    void                     PostDrawProgressBarEvent();
+   // Windows notification handler override (delegates to HandleDarkModeCustomDraw in dark mode)
+   bool                     MSWOnNotify(int idCtrl, WXLPARAM lParam, WXLPARAM *result) override;
```
AenBleidd, line 104: I would move this (and two new functions below) function from `USE_NATIVE_LISTCONTROL` to `__WXMSW__` since this is a Windows specific code, and even if `USE_NATIVE_LISTCONTROL` is currently set for Windows builds only, this is something that we might reconsider in the future, and having Windows-specific code guarded with Windows-specific macro is more reliable.

AenBleidd (changes requested):
Sorry for the long delay, I just got time to review and test this properly.
@CharlieFenton, please take a look at this PR when you have time.
I have tested in on Windows and Linux, and while there are some minor issues left, they are unrelated to this particular PR, and could be fixed separately (I will handle that since some of the issues are specific to Linux, and some other specific to Windows).
Everything else looks good to me.
@AufarZakiev, please wait for @CharlieFenton review, and apply the suggestions when you have time.

CharlieFenton:
I won’t be able to look at this until some time next week.

clientgui/BOINCListCtrl.h
CharlieFenton: From @CharlieFenton: Please do not make this change. I may decide to switch to the native list control on the Mac at some future time. I have not done so in the past because it was much less efficient on the Mac than the generic version. This is the reason I did not  just use _WXMSW__. (There is no native list control for Linux.)

clientgui/BOINCListCtrl.cpp
```diff
@@ -524,8 +524,29 @@ void CBOINCListCtrl::DrawProgressBars()
     int n = (int)m_iRowsNeedingProgressBars.GetCount();
     if (n <= 0) return;
 
-    wxColour progressColor = isDarkMode ? wxColour(0, 64, 128) : wxColour(192, 217, 217);
+    wxColour progressColor;
+    wxColour remainderColor;
+    wxColour textColor;
```
CharlieFenton: This generates a _variable set but not used_ warning since line 636 is commented out. Please delete or comment out all references to tis variable to eliminate the warning.

clientgui/NoticeListCtrl.cpp
```diff
@@ -159,12 +155,7 @@ void CNoticeListCtrl::SetItemCount(int newCount) {
...
     m_itemCount = newCount;
-    bool isWindowsDarkMode = false;
-#ifdef __WXMSW__
-    const wxSystemAppearance appearance = wxSystemSettings::GetAppearance();
-    isWindowsDarkMode = appearance.IsSystemDark();
-#endif
-    if (wxGetApp().GetIsDarkMode() || isWindowsDarkMode){
+    if (wxGetApp().GetIsDarkMode()){
 #if wxUSE_WEBVIEW
         m_noticesBody =  wxT("<html><style>body{background-color:black;color:white;}</style><head></head><body><font face=helvetica>");
```
CharlieFenton: I noticed that links are hard to see in dark mode. (This is not your fault; it was true of my version on the Mac even before your changes.) I would appreciate it if you could change this line to:
`m_noticesBody =  wxT("<html><style>body{background-color:black;color:white;a:link {color:#0080FF;}}</style><head></head><body><font face=helvetica>");`

clientgui/NoticeListCtrl.cpp
```diff
@@ -101,12 +101,8 @@ bool CNoticeListCtrl::Create( wxWindow* parent ) {
...
-    bool isWindowsDarkMode = false;
-#ifdef __WXMSW__
-    const wxSystemAppearance appearance = wxSystemSettings::GetAppearance();
-    isWindowsDarkMode = appearance.IsSystemDark();
-#endif
-    if (wxGetApp().GetIsDarkMode() || isWindowsDarkMode){
+
+    if (wxGetApp().GetIsDarkMode()){
 #if wxUSE_WEBVIEW
         m_noticesBody = wxT("<html><style>body{background-color:black;color:white;}</style><head></head><body></body></html>");
```
CharlieFenton: I noticed that links are hard to see in dark mode. (This is not your fault; it was true of my version on the Mac even before your changes.) I would appreciate it if you could change this line to:
`m_noticesBody = wxT("<html><style>body{background-color:black;color:white;a:link {color:#0080FF;}}</style><head></head><body></body></html>");`

CharlieFenton (changes requested):
I requested 3 minor changes today:
* BOINCListCtrl.cpp line 529 (also affecting lines 535, 539, 544.)
* BOINCNoticeCtrl .cpp line 106
* BOINCNoticeCtrl .cpp line 169.

Otherwise it looks good to me.
Thank you for your contribution and for your efforts to remain compatible with the existing mac and Linux code.

CharlieFenton:
@AufarZakiev I noticed one more issue. (Again, this was already an issue with my code, not something you changed.)
In the Event Log, the zebra stripes (alternate row background colors) don't work in Dark Mode, at least not on the Mac. Also, error messages in the Event Log appear in red text in Light Mode but not in Dark Mode. They should also appear red in Dark Mode. If you could figure out how to fix this, I would find that very helpful. But if this is more than you are comfortable taking on, I certainly understand.

One way to get red text in the Event Log is to put unrecognized tags in the _cc_config.xml_ file in the BOINC Data directory.

CharlieFenton:
> the zebra stripes (alternate row background colors) don't work in Dark Mode,

Never mind. I figured it out and I will submit a PR to fix this after this PR has been merged.

CharlieFenton:
@AufarZakiev I just realized that I have write access to this branch in the repository. I will make the changes I requested and also the fix for the Event Log in Dark Mode. 

I just noticed that if I set `USE_NATIVE_LISTCONTROL` to 1 for the Mac, it now has compilation errors. I will fix those by adding `#ifdef __WXMSW__` in the appropriate places.

CharlieFenton (approved):
@AufarZakiev @AenBleidd I have completed the changes I requested. I have now approved this PR. Please check that I  have not broken anything for Dark Mode on MS Windows everything still works as you expect.

Thank you again for your thoughtful and thorough contribution to BOINC.

CharlieFenton:
Oops not quite done. I need to fix something.

CharlieFenton:
OK, Fixed now. I had missed one instance of `#if SUPPORTDARKMODE`. @AufarZakiev @AenBleidd Please review and make sure it is working properly on MS Windows. 

@AenBleidd If it looks OK to both of you, then please merge.

clientgui/BOINCGUIApp.h
CharlieFenton: Done

clientgui/BOINCListCtrl.h
CharlieFenton: See my comment above

clientgui/BOINCListCtrl.cpp
CharlieFenton: Fixed

clientgui/NoticeListCtrl.cpp
CharlieFenton: Fixed

clientgui/NoticeListCtrl.cpp
CharlieFenton: Fixed

clientgui/BOINCGUIApp.cpp
```diff
@@ -173,13 +177,15 @@ bool CBOINCGUIApp::OnInit() {
     wxSystemOptions::SetOption(wxT("msw.staticbox.optimized-paint"), 0);
 #endif
 #ifdef __WXMAC__
+#if ! USE_NATIVE_LISTCONTROL
```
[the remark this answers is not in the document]
CharlieFenton: Fixed

AenBleidd:
Looks good to me.
Thank you @AufarZakiev and @CharlieFenton for your hard work on this PR!

AenBleidd (approved):

PR #6680, opened 2025-11-18
client: improve 'non-BOINC total CPU usage' measurement

davidpanderson:
We compute this as (total CPU usage) - (BOINC-related total CPU usage)
where the latter is the sum of several sets of processes:
- BOINC tasks
- low-priority processes
- Vbox processes, i.e. VboxSVC.exe
- (Win/Mac) the VM processes that Docker/Podman containers run in

We compute this every 10 seconds.

This works fine until a process exits.
Suppose e.g. a low-priority process with X CPU seconds exits.
Then BOINC-related total CPU usage will drop by X,
so non-BOINC total CPU usage will (incorrectly) increase by X,
and the client will possibly (incorrectly) suspend computing for 20 seconds.

Solution: compute the CPU usage of each of the above 4 types separately.
If any of them decreases (e.g. because a process has exited)
don't compute non-BOINC CPU usage this time.

This isn't 100% accurate.
For example, if a low-priority process runs and exits
entirely in a 10-second period,
it will incorrectly be counted as non-BOINC CPU time.

### DIFF
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -594,7 +594,7 @@ void ACTIVE_TASK_SET::get_memory_usage() {
                         total_cpu_time_now, brc, nbrc, delta_nbrc, delta_t
                     );
                 }
-			}
+            }
         }
         prev_nbrc = nbrc;
     } else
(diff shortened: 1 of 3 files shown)


AenBleidd:
I'll let @CharlieFenton test this before merging.

PR #6705, opened 2025-11-26
client, docker_wrapper (Win): run Podman commands as root

davidpanderson:
Also: minor cleanup of WSL-related code

---
## Summary by cubic
On Windows, run Podman commands as root inside WSL to avoid permission errors and ensure consistent Docker/Podman behavior. Also pass WSL_DISTRO objects instead of raw strings in WSL helpers.

- **Bug Fixes**
  - Run Podman as root via wsl.exe; keep default user for boinc-buda-runner versions < 4.
  - Use BOINC_WSL_DISTRO_NAME in standalone wrappers to avoid mismatched distro names.

- **Refactors**
  - DOCKER_CONN::init and WSL_CMD methods now accept WSL_DISTRO.
  - Updated call sites (hostinfo_wsl.cpp, cs_apps.cpp, wrappers); minor headers/signatures adjusted.

<sup>Written for commit cde53e31bbc94627c302d1980fde4b9ba6045fc1. Summary will update automatically on new commits.</sup>

### DIFF
--- a/samples/docker_wrapper/docker_wrapper.cpp
+++ b/samples/docker_wrapper/docker_wrapper.cpp
@@ -724,11 +724,12 @@ int wsl_init() {
         distro.docker_type = PODMAN;
         distro.boinc_buda_runner_version = 4;
         dp = &distro;
-    }
-    dp = aid.host_info.wsl_distros.find_docker();
-    if (!dp) {
-        fprintf(stderr, "wsl_init(): no usable WSL distro\n");
-        return -1;
+    } else {
+        dp = aid.host_info.wsl_distros.find_docker();
+        if (!dp) {
+            fprintf(stderr, "wsl_init(): no usable WSL distro\n");
+            return -1;
+        }
     }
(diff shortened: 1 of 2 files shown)


samples/docker_wrapper/docker_wrapper.cpp
```diff
@@ -718,21 +718,22 @@ double get_fraction_done() {
 // find a WSL distro with Docker and set up a command link to it
 //
 int wsl_init() {
-    string distro_name;
+    WSL_DISTRO* distro;
```
AenBleidd: Do we really need to have a pointer here?
Can't we just allocate an object on stack instead?
This will avoid a memory leak.

samples/wsl_wrapper/wsl_wrapper.cpp
```diff
@@ -268,20 +268,22 @@ int main(int argc, char** argv) {
         }
     }
 
-    string distro_name;
+    WSL_DISTRO *distro;
```
AenBleidd: Same here. I'd not use pointer.

AenBleidd (approved):

PR #6714, opened 2025-12-01
client, Android: clarify message

davidpanderson:
If we suspend because of battery charge, we don't resume for at least 5 minutes. If that's why we're suspended, say so.

---
## Summary by cubic
Clarified Android suspend messaging for low battery and overheating by adding separate “battery charge wait” and “battery heat wait” reasons with the remaining wait time. Avoids confusing “charging” or “overheated” messages during the 5-minute backoff.

- **Bug Fixes**
  - Return SUSPEND_REASON_BATTERY_CHARGE_WAIT and SUSPEND_REASON_BATTERY_HEAT_WAIT during backoff; log seconds until resume.
  - Use SUSPEND_REASON_BATTERY_CHARGING only when below the min battery percentage, and SUSPEND_REASON_BATTERY_OVERHEATED only when actually overheated.

- **Refactors**
  - Removed obsolete testbase script.

<sup>Written for commit 8e8909b94eaaa6ba27b8212e03527c867c49185b. Summary will update automatically on new commits.</sup>

### DIFF
--- a/client/cs_prefs.cpp
+++ b/client/cs_prefs.cpp
@@ -283,7 +283,6 @@ int CLIENT_STATE::check_suspend_processing() {
         // If suspend because of hot battery, don't resume for at least 5 min
         // (crude hysteresis)
         //
-        static double battery_heat_resume_time=0;
         if (device_status.battery_state == BATTERY_STATE_OVERHEATED) {
             battery_heat_resume_time = now + ANDROID_BATTERY_BACKOFF;
             return SUSPEND_REASON_BATTERY_OVERHEATED;


client/cs_prefs.cpp
```diff
@@ -381,6 +379,14 @@ void CLIENT_STATE::show_suspend_tasks_message(int reason) {
                 );
             }
             break;
+        case SUSPEND_REASON_BATTERY_CHARGE_WAIT:
+            if (log_flags.task) {
+                msg_printf(NULL, MSG_INFO,
+                    "(battery charge wait: %.0f sec",
+                    battery_charge_resume_time - now
```
AenBleidd, line 384: @davidpanderson, please fix build error:
```
cs_prefs.cpp:386:21: error: use of undeclared identifier 'battery_charge_resume_time'
  CXX      boinc_client-cs_scheduler.o
                    battery_charge_resume_time - now
                    ^
```

PR #6734, opened 2025-12-07
client: improve Docker-related messages

davidpanderson:
all platforms: if no Docker, suggest they install it
Win:
    if Docker but not our WSL distro, suggest they use ours
    if they have our distro but not current, suggest upgrade

Do this both at startup and after biweekly get-version RPC

Also: if a notice has a link, write it to log as well

---
## Summary by cubic
Improve Docker and WSL guidance in the client to help users install Docker or the BOINC WSL distro when needed. Messages appear at startup and after the biweekly get-version check, and linked notices are now written to the log.

- **New Features**
  - Centralized Docker/WSL notices via show_docker_messages at startup and after get-version RPC.
  - Non-Android: if Docker is missing, show a notice with the correct OS install link.
  - Windows: if Docker is present but not using the BOINC WSL distro, recommend installing it; if the distro is outdated, suggest upgrading with a link.
  - Log linked notice URLs to the console.
  - Manager notices: larger link text and “more info…” label.

<sup>Written for commit d4bd11796b3c696a5e3d1471f68204c7b7aabcc4. Summary will update automatically on new commits.</sup>

### DIFF
--- a/client/client_state.cpp
+++ b/client/client_state.cpp
@@ -2539,7 +2539,7 @@ void show_docker_messages() {
             }
         } else {
             msg_printf_notice(0, true, url,
-                "Docker is present but not using the BOINC WSL distro.  Some proect apps may not function properly. We recommend that you install the BOINC WSL distro."
+                "Docker is present but not using the BOINC WSL distro.  Some project apps may not function properly. We recommend that you install the BOINC WSL distro."
             );
         }
 #endif


AenBleidd:
@davidpanderson, please fix typos and build error on Android:
```
ld: error: undefined symbol: show_docker_messages()
>>> referenced by current_version.cpp
>>>               boinc_client-current_version.o:(GET_CURRENT_VERSION_OP::handle_reply(int))
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [Makefile:823: boinc_client] Error 1
```

PR #6736, opened 2025-12-08
Client: use 'abort' rather than 'quit' when appropriate

davidpanderson:
The client has two ways to tell a job to exit:
'quit' if we plan to restart the job later;
'abort' if we're not going to restart it.

We were incorrectly using quit instead of abort
when detaching or resetting a project.

This doesn't make much difference in general, but for Docker jobs it does: abort should remove the container and image, quit should not.

Also: rename ACTIVE_TASK::request_exit() to request_quit() to clarify the distinction.

---
## Summary by cubic
Switch task exit behavior to use abort for detach/reset and quit for client exit, so non-restarting jobs are properly cleaned up. This fixes Docker jobs by removing containers/images on abort.

- **Bug Fixes**
  - Detach/reset now send abort; client exit sends quit.
  - Docker jobs clean up containers and images on abort.

- **Refactors**
  - Renamed ACTIVE_TASK::request_exit() to request_quit().
  - request_tasks_exit and exit_tasks now take a will_restart flag to choose quit vs abort.
  - Updated logs and comments for clarity.

<sup>Written for commit aaded253ce01bcccfb9ebc8a0ee01acff4f6bdd6. Summary will update automatically on new commits.</sup>

### DIFF
--- a/client/sim_util.cpp
+++ b/client/sim_util.cpp
@@ -114,8 +114,8 @@ int ACTIVE_TASK::suspend() {
     return 0;
 }
 
-int ACTIVE_TASK::request_exit() {
-    set_task_state(PROCESS_UNINITIALIZED, "request_exit");
+int ACTIVE_TASK::request_quit() {
+    set_task_state(PROCESS_UNINITIALIZED, "request_quit");
     return 0;
 }
 


AenBleidd:
@davidpanderson, please fix build errors on Windows:
```
       (ClCompile target) -> 
         D:\a\boinc\boinc\client\sim_util.cpp(117,18): error C2039: 'request_exit': is not a member of 'ACTIVE_TASK' [D:\a\boinc\boinc\win_build\sim.vcxproj]
         D:\a\boinc\boinc\client\sim_util.cpp(118,5): error C3861: 'set_task_state': identifier not found [D:\a\boinc\boinc\win_build\sim.vcxproj]
```

PR #6757, opened 2025-12-31
Client: fix Docker issues

davidpanderson:
- Mac: we suspend computing while Podman init process is running.
    If (for some reason) the process never exits, we never compute.
    Solution:
    Don't suspend while Podman init is running; let non-Docker jobs run.
    Add bool HOST_INFO::podman_inited.
    If not set and we're using Podman:
        - Don't include Docker info in sched requests
        - Don't run Docker jobs

- All platforms: if Docker/Podman not installed, and client has a Docker job,
    don't try to run it.  Show message telling user to install Podman.

---
## Summary by cubic
Fix Docker job handling to avoid client stalls and bad scheduler info. Non-Docker tasks continue while Podman initializes on macOS; Docker tasks are deferred when Docker/Podman isn’t available, with a clear user message.

- **Bug Fixes**
  - macOS: don’t suspend while Podman init runs; defer Docker jobs until the Podman VM is running; omit Docker info from scheduler requests until ready.
  - All platforms: if Docker/Podman isn’t installed, skip Docker jobs and alert the user with an install link.
  - Minor: add a helper to detect Docker plan classes; fix resource cleanup in Podman VM check.

<sup>Written for commit f41d39f1bba63ab7b2582433b207c0d5ff2dd81d. Summary will update on new commits.</sup>

### DIFF
--- a/client/cpu_sched.cpp
+++ b/client/cpu_sched.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
-// http://boinc.berkeley.edu
-// Copyright (C) 2008 University of California
+// https://boinc.berkeley.edu
+// Copyright (C) 2025 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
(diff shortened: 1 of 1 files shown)


AenBleidd:
@davidpanderson, please fix this build error:
```
D:\a\boinc\boinc\client\cpu_sched.cpp(169,42): error C2039: 'docker_version': is not a member of 'HOST_INFO' [D:\a\boinc\boinc\win_build\sim.vcxproj]
```

AenBleidd (approved):

PR #6822, opened 2026-01-29
lib: use modern for() syntax

davidpanderson:
no functional change

---
## Summary by cubic
Replaced index-based loops with range-based for loops across lib and added missed const qualifiers to improve readability and avoid copies. No functional changes.

- **Refactors**
  - Converted loops to range-based for with const references; delete loops iterate pointers directly.
  - Marked EXCLUDE_GPU::write() and GUI_URL::print() as const and updated call sites; added missed const in print helpers.
  - Kept explicit counters only where needed (e.g., file_N in HTTP form posts).

<sup>Written for commit a87f7fe99eea7aaba0a68a4bc5dd9d31669b55b9. Summary will update on new commits.</sup>

### DIFF
--- a/lib/remote_submit.cpp
+++ b/lib/remote_submit.cpp
@@ -447,7 +447,7 @@ int submit_jobs_params(
         job_params.delay_bound
     );
     string request = buf;
-    for (JOB &job: jobs) {
+    for (const JOB &job: jobs) {
         request += "<job>\n";
         snprintf(buf, sizeof(buf), "  <name>%s</name>\n", job.job_name);
         request += buf;


CharlieFenton:
@davidpanderson Why are these changes necessary?

I really wish you wouldn't do this and stick to the old tried and true syntaxes. I've never heard of this "modern" syntax. It may not be supported by all (especially) older compilers. I'm not familiar or comfortable with many of the newer forms you've been using lately. I'm barely comfortable with use of vectors in for loops and such and I suspect I am not the only one.

If you want the largest number of volunteers (including me) to be able to contribute to BOINC development, using such arcane syntax is not the way to accomplish that.

AenBleidd:
@CharlieFenton, this is a feature of C++11 standard (15 years on the market already, so it isn't that 'new'). Also, it's much easier to use it, and much safer (meaning it's harder to misuse it) and to be completely honest, if we want to have more volunteers to help us with the project, we need to have our codebase more or less modern.
The good stuff, you don't need to use it if you don't want, and I can handle that for you.

davidpanderson:
Range-based for has significant advantages, which I discuss here:

https://github.com/BOINC/boinc/wiki/Coding-style#range-based-for

PR #6823, opened 2026-01-29
client: use modern for() syntax (hopefully no functional change)

davidpanderson:
TODO:
- use iterators for loops that delete things
- finish up remaining for()s

---
## Summary by cubic
Refactor the client to use range-based for loops for clearer, safer iteration, and fix a startup ordering bug so GPUs are detected before projects initialize.

- **Refactors**
  - Replaced index-based loops with range-based for across client code (tasks, scheduler, file xfers, RPC, prefs, simulation).
  - Used references in loops to avoid unnecessary copies where safe.
  - Added const qualifiers to function signatures (e.g., FILE_REF::write, OLD_RESULT::print, and related).
  - Follow-ups: use iterators for loops that erase elements; finish remaining legacy for()s.

- **Bug Fixes**
  - Ensure GPU detection runs before project initialization so projects see the correct GPU count at startup.
  - Make submit_buda work again; default variant is now intel_cpu.

<sup>Written for commit 3560afed1378b907b1acf032ef73611353957ba8. Summary will update on new commits.</sup>

### DIFF
--- a/client/acct_mgr.cpp
+++ b/client/acct_mgr.cpp
@@ -649,7 +649,7 @@ void ACCT_MGR_OP::handle_reply(int http_op_retval) {
 
         // process projects
         //
-        for (AM_ACCOUNT& acct: accounts) {
+        for (const AM_ACCOUNT& acct: accounts) {
             pp = gstate.lookup_project(acct.url.c_str());
             if (pp) {
                 if (acct.detach) {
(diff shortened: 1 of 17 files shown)


CharlieFenton:
@davidpanderson Why are these changes necessary?

I really wish you wouldn't do this and stick to the old tried and true syntaxes. I've never heard of this "modern" syntax. It may not be supported by all (especially) older compilers. I'm not familiar or comfortable with many of the newer forms you've been using lately. I'm barely comfortable with use of vectors in for loops and such and I suspect I am not the only one.

If you want the largest number of volunteers (including me) to be able to contribute to BOINC development, using such arcane syntax is not the way to accomplish that.

AenBleidd:
@CharlieFenton, already replied to you here: https://github.com/BOINC/boinc/pull/6822#issuecomment-3817892557

PR #6242, opened 2025-04-11
client (Unix): add support for idle detection in a separate process

davidpanderson:
The design is described here:
https://github.com/BOINC/boinc/wiki/Linux-idle-detection

PR #6865, opened 2026-02-17
[Draft] Linux idle helper RPC probe + cc_config knobs (fixes #1187 path)

### LINKED ISSUE #1187: boinc_client does not notice (or does not care) that "computer is in use"

nicepopo86-lang:
## Summary
Phase-1 client-side integration for Linux idle helper RPC (RFC spike for #1187):

- adds optional UNIX-socket probe in `HOST_INFO::user_idle_time()` Linux path
  - request: `GET_IDLE_SECS\n`
  - response parse: `OK <seconds> <source>`
  - short socket timeout and safe fallback to existing logic when unavailable
- keeps existing XScreenSaver (`xss_idle()`) fallback path unchanged
- adds minimal `cc_config.xml` knobs:
  - `<enable_idle_helper_rpc>` (default `0`)
  - `<idle_helper_socket>` (default `/run/user/%d/boinc/idle-helper.sock`)
  - `<idle_helper_timeout_ms>` (default `50`)
- wires parsing/writing/showing of new config fields in both config parsers

## Notes
- Defaults preserve current behavior unless helper RPC is explicitly enabled.
- Could not run a full compile in this environment because bootstrap dependency `m4` is missing (`./_autosetup` fails before build).

Refs: #1187

---
## Summary by cubic
Adds an optional Linux idle helper RPC probe over a UNIX socket with new cc_config knobs, keeping current idle detection as the default. Supports #1187 by querying a helper and safely falling back when unavailable.

- **New Features**
  - Linux: probe idle via UNIX socket (default /run/user/%d/boinc/idle-helper.sock) with short timeout and safe fallback.
  - New cc_config options: enable_idle_helper_rpc (false), idle_helper_socket, idle_helper_timeout_ms (50). Parsed, written, and shown in logs.
  - Idle time now uses min(existing, helper result). XScreenSaver path unchanged; debug logs behind idle_detection_debug.

- **Migration**
  - Off by default. To enable, set <enable_idle_helper_rpc>true</enable_idle_helper_rpc> in cc_config.xml; optionally set <idle_helper_socket> and <idle_helper_timeout_ms>.

<sup>Written for commit d3d5319414a50a37332306cb5fa194f83864057f. Summary will update on new commits.</sup>

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -42,7 +42,6 @@
 #undef Always
 #include <dirent.h> //for opening /tmp/.X11-unix/
   // (There is a DirScanner class in BOINC, but it doesn't do what we want)
-#include "log_flags.h" // idle_detection_debug flag for verbose output
 #endif
 
 #include <cstdio>
@@ -114,6 +113,14 @@
 #include "client_types.h"
 #include "client_msgs.h"
 #include "hostinfo.h"
+
+#include "log_flags.h"
+
+#if LINUX_LIKE_SYSTEM
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <sys/time.h>
+#endif
 
 using std::string;
 using std::min;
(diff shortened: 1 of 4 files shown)


AenBleidd:
See #6242

PR #6869, opened 2026-02-18
client: restore get_screensaver_tasks() RPC, which I mistakenly removed

davidpanderson:
## Summary by cubic
Restores the get_screensaver_tasks GUI/Manager RPC with correct XML, so the screensaver can fetch scheduled tasks and the suspend reason. Also tidies Docker/Podman CLI detection on macOS.

- **New Features**
  - Added get_screensaver_tasks RPC, exposed in gui_rpc_server and Manager AsyncRPC; returns suspend_reason and results for CPU_SCHED_SCHEDULED tasks.

<sup>Written for commit ed7fafa83901ecef232db487c372fedeef210ab8. Summary will update on new commits.</sup>

### DIFF
--- a/client/gui_rpc_server_ops.cpp
+++ b/client/gui_rpc_server_ops.cpp
@@ -647,7 +647,7 @@ static void handle_get_screensaver_tasks(GUI_RPC_CONN& grc) {
     );
     for (i=0; i<gstate.active_tasks.active_tasks.size(); i++) {
         atp = gstate.active_tasks.active_tasks[i];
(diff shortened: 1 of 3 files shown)


CharlieFenton:
@davidpanderson While I realize that _AsyncRPC.cpp_ and _AsyncRPC.h_ are not used by the screensaver, I would prefer that you restore the code for the `get_screensaver_tasks()` RPC there as well for the sake of completeness.

davidpanderson:
I restored the RPC in the Manager API

PR #6460, opened 2025-07-24
MacOS 26 Screensaver fixes

CharlieFenton:
This PR updates 2 aspects of BOINC screensaver logic for MacOS 25:
 - The BOINC screensaver coordinator runs properly under the legacy screensaver engine
 -  MacOS 14 and MacOS 15 broke the ability select a screensaver programmatically, so the BOINC installer cold no longer offer the option to set BOINC as your screensaver. MacOS 26 restores that capability, but using a different API. Changes in this PR implement that capability, again offering the user the option to set BOINC as their screensaver.

This code is backward compatible with earlier versions of MacOS, so it can be merged into all new releases of BOINC.

PR #6985, opened 2026-04-12
Mac: in command-line installs don't ask about setting screensaver

CharlieFenton:
The BOINC installer for the Mac is normally interactive, asking the user about options including whether to allow non-administrative users to manage BOINC and whether to set BOINC as the user's screensaver.

Command-line installs are intended primarily for situations where the operator is not sitting at the target computer, such as deploying BOINC on multiple computers in a school computer lab. It might also be used for alternate installation methods, such as homebrew. In these cases, interactive  operation is not desirable.

PR #6460 added a dialog "Do you wish to set BOINC as your screensaver?" to the BOINC installer for Macintosh but did not include code to suppress that dialog for command-line installs. This PR fixes that.

---
## Summary by cubic
Make Mac command-line installs fully non‑interactive by suppressing the screensaver prompt and signaling finish-install with `-c`. Screensaver detection/setting now only runs on macOS < 14 (Sonoma); GUI installs are unchanged.

- **Bug Fixes**
  - Detect `COMMAND_LINE_INSTALL=1` in Postinstall and `BOINC_Finish_Install` to skip the "Set BOINC as your screensaver?" dialog.
  - Pass `-c` to the finish-install LaunchAgent during command-line installs to ensure no prompts.
  - Restrict screensaver checks and setting to macOS versions earlier than 14.

<sup>Written for commit 6f4229bfbe35b3bdf3e3e4f690af76279aca4dea. Summary will update on new commits.</sup>

### DIFF
--- a/mac_installer/PostInstall.cpp
+++ b/mac_installer/PostInstall.cpp
@@ -24,7 +24,8 @@
 //  [1] allow non-administrative users to run the BOINC Manager
 //      (asked only if this Mac has any non-administrative users)
 //  [2] set BOINC as the screensaver for all users who can run BOINC
-//      (asked only if BOINC screensaver is not already set for them)
+//      (only for MacOS < 14.0 Sonoma and only if BOINC screensaver is
+//      not already set for them.)
 //
 // The installer can also be run from the command line.  This is useful
 //  for installation on remote Macs.  However, there is no way to respond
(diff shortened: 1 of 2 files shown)


AenBleidd (approved):

CharlieFenton:
@AenBleidd This PR is ready to be merged

PR #6964, opened 2026-04-08
Mac: Fix broken links in ReadMe files

CharlieFenton:
Mac: Fix broken links in ReadMe files for BOINC and World Community Grid installers
Note: We no longer support branded Mac installers for Progress Thru Processors, Charity Engine or Grid Republic

---
## Summary by cubic
Fixes broken BOINC wiki links in the Mac installer ReadMes by pointing to the GitHub wiki, and applies minor formatting fixes. Branded Mac installers for Progress Thru Processors, Charity Engine, and Grid Republic are no longer supported.

- **Bug Fixes**
  - Updated links in `mac_installer/ReadMe.rtf` and `mac_installer/wcgrid-ReadMe.rtf` to `https://github.com/BOINC/boinc/wiki/Tools-for-MacOS` and `https://github.com/BOINC/boinc/wiki/GPU-computing`; NVIDIA CUDA driver archive link unchanged.
  - Cleaned up RTF formatting and punctuation in `wcgrid-ReadMe.rtf` per CUBIC feedback.

<sup>Written for commit 9515da1231f54eabcdfb265e630acadebd246780. Summary will update on new commits.</sup>

PR #7000, opened 2026-04-15
web: fix outdated wiki links

davidpanderson:
## Summary by cubic
Updated user-facing help links to the BOINC GitHub Wiki to replace outdated `boinc.berkeley.edu/wiki` URLs. This fixes broken links on the download and welcome pages.

- **Bug Fixes**
  - Replaced `boinc.berkeley.edu/wiki` with `github.com/BOINC/boinc/wiki` in `download_software.php` and `welcome.php`.
  - Updated pages: Account managers, BOINC Help, BOINC Manager, Choosing and joining projects.

<sup>Written for commit 563234bdf03f52f8e5a62ca219846534b9ae8450. Summary will update on new commits.</sup>

### DIFF
--- a/html/user/download_software.php
+++ b/html/user/download_software.php
@@ -169,7 +169,7 @@ function direct_to_boinc() {
                 to connect BOINC to your %1 account.
                 See <a href=%2>detailed instructions</a>.",
                 PROJECT,
-                'https://boinc.berkeley.edu/wiki/Account_managers'
+                'https://github.com/BOINC/boinc/wiki/Account_managers'
             )
         );
     } else {
--- a/html/user/welcome.php
+++ b/html/user/welcome.php
@@ -25,7 +25,7 @@ echo tra("
     If not, %2get help here%3.
     ",
     "<img src=https://boinc.berkeley.edu/logo/boinc32.bmp>",
-    "<a href=https://boinc.berkeley.edu/wiki/BOINC_Help>",
+    '<a href=https://github.com/BOINC/boinc/wiki/BOINC-Help>',
     "</a>"
 );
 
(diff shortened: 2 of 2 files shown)


AenBleidd (approved):

CharlieFenton:
There are many pages  on the wiki that are very hard to find. When I wanted to update these two links in the Mac installer ReadMe file, it was quite a challenge to find them: https://github.com/BOINC/boinc/wiki/Tools-for-MacOS and https://github.com/BOINC/boinc/wiki/GPU-computing . None of the categories on the home wiki page list these and they really don't fit any of those categories. They might fit under a category like _User Documentation_. Of the 581 pages on the BOINC wiki, only a small percentage can be found through those links.

While you can search for a page, you need to know the page title. That didn't work for me because GitHub ones have hyphens where the original wiki pages had underscores. The same is true for the ability to find a page alphabetically.

Back when the documentation was under Trac, there were pages https://boinc.berkeley.edu/trac/wiki/TitleIndex and https://boinc.berkeley.edu/wiki/Special:Allpages which I found very helpful in finding the documentation. One of those (I think Alleges) had a hierarchical list of every page. Can we implement something like those? Does GitHub offer anything comparable?

It would have been helpful if the old wiki URLs had been kept with links to their replacements, which would have eliminated this issue of fixing dead links piecemeal that this PR and my PR #6964 are addressing, but I don't know if that was possible..

AenBleidd:
@CharlieFenton, wiki is just another git repository, so you can just clone it locally:
```
git clone https://github.com/BOINC/boinc.wiki.git
```
There are no special pages here, because this is not a wiki engine, but just a collections of Markdown files.
We can create a special page with all files listed, but it needs to be manually edited to be up to date.
I don't think it worth doing that.

davidpanderson:
I do what @AenBleidd said: clone the wiki repo to my own computer (Linux in my case), where I can grep for things, and I do editing there.

davidpanderson:
Also, on the wiki page (right column) there's a 'Pages' link.  This shows a list of all the file names (currently 581 of them)

CharlieFenton:
On the wiki page https://github.com/BOINC/boinc/wiki/Volunteer The link [becoming an Alpha tester](https://github.com/BOINC/boinc/pull/AlphaInstructions) is broken

davidpanderson:
I fixed the link

PR #7002, opened 2026-04-15
support cross-compiling for windows_arm64

ahorek:
**Description of the Change**
fixes a few issues to allow cross-compiling libraries for --host=aarch64-w64-mingw32 --with-boinc-platform="windows_arm64"

https://github.com/mstorsjo/llvm-mingw/releases 

**Alternate Designs**
Visual Studio works well, but many existing projects rely on GCC / Clang, and porting them to Visual Studio isn’t always straightforward.

**Release Notes**
N/A

---
## Summary by cubic
Enable cross-compiling BOINC libraries for Windows ARM64 with MinGW/Clang (`--host=aarch64-w64-mingw32`, `--with-boinc-platform=windows_arm64`). CI builds x86_64 and ARM64 libs on Ubuntu using `mingw-w64`/`llvm-mingw`, verifies the `llvm-mingw` checksum, and tags artifacts/logs by arch.

- **New Features**
  - Add `OS_WINARM64` and detect `AARCH64_BUILD` in `lib/Makefile.mingw`.
  - Exclude `stackwalker_win.cpp` on ARM64 (Autotools and Makefile guards).
  - Keep `WINVER`/`_WIN32_WINNT` defines; drop legacy cross-build defines.
  - Guard `std::set_unexpected` and callstack dumps on `__aarch64__`.
  - Include `<stddef.h>` for `socklen_t`.

- **CI**
  - Matrix by `arch` (`x86_64`, `aarch64`) with per-arch toolchains (`mingw-w64`, `llvm-mingw`).
  - Download `llvm-mingw` 20260421, verify SHA-256, add to `PATH` for ARM64.
  - Separate build steps for libs by arch; skip `apps-mingw-vcpkg` and `libs-mingw-cmake` on ARM64.
  - Name jobs, artifacts, and logs with `arch`.

<sup>Written for commit 91ccb8d3631ae7bedab543c08d9aa1d0571b5122. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/mingw.yml
+++ b/.github/workflows/mingw.yml
@@ -35,11 +35,17 @@ env:
 
 jobs:
   build:
-    name: ${{ matrix.type }}-build
+    name: ${{ matrix.type }}-${{ matrix.arch }}-build
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        arch: [x86_64, aarch64]
         type: [libs-mingw, apps-mingw-vcpkg, libs-mingw-cmake]
+        exclude:
+          - arch: aarch64
+            type: apps-mingw-vcpkg
+          - arch: aarch64
+            type: libs-mingw-cmake
       fail-fast: false
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
@@ -57,11 +63,29 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y mingw-w64 mingw-w64-tools mingw-w64-x86-64-dev mingw-w64-i686-dev tar curl zip unzip p7zip-full
+          sudo apt-get install -y tar curl zip unzip p7zip-full
 
-      - name: Make libs with mingw
-        if: success() && matrix.type == 'libs-mingw'
+      - name: Install x86_64 mingw toolchain
+        if: success() && matrix.arch == 'x86_64'
+        run: |
+          sudo apt-get install -y mingw-w64 mingw-w64-tools mingw-w64-x86-64-dev mingw-w64-i686-dev
+
+      - name: Install aarch64 llvm-mingw toolchain
+        if: success() && matrix.arch == 'aarch64'
+        run: |
+          wget -q https://github.com/mstorsjo/llvm-mingw/releases/download/20260421/llvm-mingw-20260421-msvcrt-ubuntu-22.04-x86_64.tar.xz
+          echo "18488138efd8d621930fa85c86c3e3eb9967c0bfa253e1f6f73c5212f7180f31 llvm-mingw-20260421-msvcrt-ubuntu-22.04-x86_64.tar.xz" | sha256sum -c -
+          tar -Jxf llvm-mingw-20260421-msvcrt-ubuntu-22.04-x86_64.tar.xz
+          echo "$GITHUB_WORKSPACE/llvm-mingw-20260421-msvcrt-ubuntu-22.04-x86_64/bin" >> $GITHUB_PATH
+
+      - name: Build libs (x86_64)
+        if: matrix.type == 'libs-mingw' && matrix.arch == 'x86_64'
         run: cd lib && MINGW=x86_64-w64-mingw32 make -f Makefile.mingw
+
+      - name: Build libs (aarch64)
+        if: success() && matrix.type == 'libs-mingw' && matrix.arch == 'aarch64'
+        run: cd lib && MINGW=aarch64-w64-mingw32 make -f Makefile.mingw
+
       - name: Install Powershell for vcpkg
         if: success() && matrix.type == 'apps-mingw-vcpkg'
         run: |
@@ -101,7 +125,7 @@ jobs:
         if: ${{ failure() }}
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
         with:
-          name: mingw_logs_${{ matrix.type }}_${{ github.event.pull_request.head.sha }}
+          name: mingw_logs_${{ matrix.type }}_${{ matrix.arch }}_${{ github.event.pull_request.head.sha }}
           path: deploy/logs.7z
 
       - name: Prepare artifacts for deploy
@@ -112,5 +136,5 @@ jobs:
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
         if: ${{! contains(matrix.type, 'libs')}}
         with:
-          name: win_${{ matrix.type }}_${{ github.event.pull_request.head.sha }}
+          name: win_${{ matrix.type }}_${{ matrix.arch }}_${{ github.event.pull_request.head.sha }}
           path: deploy/win_${{ matrix.type }}.7z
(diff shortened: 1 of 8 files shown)


AenBleidd:
Please add a CI test for this here: https://github.com/BOINC/boinc/blob/master/.github/workflows/mingw.yml

.github/workflows/mingw-aarch64.yml
```diff
@@ -0,0 +1,119 @@
+# This file is part of BOINC.
```
AenBleidd: Please reuse the existing mingw.yml file to avoid duplications and remove complexity

lib/Makefile.mingw.aarch64
```diff
@@ -0,0 +1,230 @@
+# This file is part of BOINC.
```
AenBleidd: Please reuse the existing Makefile.mingw file to avoid duplications and remove complexity

AenBleidd (changes requested):

.github/workflows/mingw.yml
```diff
@@ -57,12 +63,31 @@ jobs:
...
+          sudo apt-get install -y tar curl zip unzip p7zip-full
 
-      - name: Make libs with mingw
-        if: success() && matrix.type == 'libs-mingw'
+      - name: Install x86_64 mingw toolchain
+        if: success() && matrix.arch == 'x86_64'
+        run: |
+          sudo apt-get install -y mingw-w64 mingw-w64-tools mingw-w64-x86-64-dev mingw-w64-i686-dev
+
+      - name: Install aarch64 llvm-mingw toolchain
```
AenBleidd, line 73: Is there any system package that provides the same executables? I prefer to not stich to  a particular version, expecially from the GitHub repo that doesn't look like something more or less official (neither gcc not clang).

.github/workflows/mingw.yml
```diff
@@ -35,11 +35,17 @@ env:
 
 jobs:
   build:
-    name: ${{ matrix.type }}-build
+    name: ${{ matrix.arch }}-${{ matrix.type }}-build
```
AenBleidd: We usually use TYPE-ARCH, not vice versa

lib/Makefile.mingw
```diff
@@ -130,11 +137,17 @@ OPTFLAGS = -O3
 # various cpp & gcc flags (for both C and C++ mode)
 # set NOCYGWIN=-mno-cygwin to build non-Cygwin Windows libs under Cygwin
 # -D_WIN32_WINDOWS=0x0410 sets to use Win98 API
-WINVERFLAGS = -D_WINDOWS -D_WIN32 -DWIN32 -DWINVER=0x0500 -D_WIN32_WINNT=0x0500 -D_MT
 HAVEFLAGS = -DHAVE_STRCASECMP
-CCXXFLAGS = $(INCS) $(DEBUG) --include $(BOINC_SRC)/version.h -DEINSTEINATHOME_CROSS_BUILD -DMINGW_WIN32 \
+WINVERFLAGS = -D_WINDOWS -D_MT
+ifndef AARCH64_BUILD
+	WINVERFLAGS += -D_WIN32 -DWIN32 -DWINVER=0x0500 -D_WIN32_WINNT=0x0500
```
AenBleidd: Why removing these flags?
`_WIN32` and `WIN32` are valid for x86, x64 and arm64 architectures, and with these macroses we identify that we build for Windows, not *nix.
`WINVER` and `_WIN32_WINNT` define compatibility level. It's ok to put them to something newer for arm64, but I'd not just remove them completely.

lib/Makefile.mingw
```diff
@@ -130,11 +137,17 @@ OPTFLAGS = -O3
...
+endif
+CCXXFLAGS = $(INCS) $(DEBUG) --include $(BOINC_SRC)/version.h -DEINSTEINATHOME_CROSS_BUILD \
 	$(HAVEFLAGS) $(WINVERFLAGS) -DBOINC \
 	-DNODB -D_CONSOLE -fexceptions $(OPTFLAGS) $(NOCYGWIN)
+ifndef AARCH64_BUILD
+	WINVERFLAGS += -DMINGW_WIN32
```
AenBleidd: `MINGW_WIN32` identifies the code that should be built only when building for Windows with MinGW toolkit. However, I don't see any single place it our code where it's being used, so I assume it's safe to remove it completely.

lib/Makefile.mingw
```diff
@@ -130,11 +137,17 @@ OPTFLAGS = -O3
...
 # set NOCYGWIN=-mno-cygwin to build non-Cygwin Windows libs under Cygwin
 # -D_WIN32_WINDOWS=0x0410 sets to use Win98 API
-WINVERFLAGS = -D_WINDOWS -D_WIN32 -DWIN32 -DWINVER=0x0500 -D_WIN32_WINNT=0x0500 -D_MT
 HAVEFLAGS = -DHAVE_STRCASECMP
-CCXXFLAGS = $(INCS) $(DEBUG) --include $(BOINC_SRC)/version.h -DEINSTEINATHOME_CROSS_BUILD -DMINGW_WIN32 \
+WINVERFLAGS = -D_WINDOWS -D_MT
+ifndef AARCH64_BUILD
+	WINVERFLAGS += -D_WIN32 -DWIN32 -DWINVER=0x0500 -D_WIN32_WINNT=0x0500
+endif
+CCXXFLAGS = $(INCS) $(DEBUG) --include $(BOINC_SRC)/version.h -DEINSTEINATHOME_CROSS_BUILD \
```
AenBleidd: `-DEINSTEINATHOME_CROSS_BUILD` is not used in the code, and it's anyway a custom code, so  it's better to remove it completely.

AenBleidd (changes requested):

.github/workflows/mingw.yml
ahorek: reversed

.github/workflows/mingw.yml
ahorek: Understand, but unfortunately, there’s no system package like mingw-w64-aarch64-dev available. This toolchain is recommended on the official website https://www.mingw-w64.org/downloads/ (LLVM-MinGW)

lib/diagnostics.cpp
```diff
@@ -386,7 +386,9 @@ int diagnostics_init(
     //_set_abort_behavior(NULL, _WRITE_ABORT_MSG);
 #ifdef __MINGW32__
     std::set_terminate(boinc_term_func);
-    std::set_unexpected(boinc_term_func);
+    #ifndef __aarch64__
+      std::set_unexpected(boinc_term_func);
```
ahorek, line 390: It's been deprecated since c++11 https://runebook.dev/en/docs/cpp/error/set_unexpected

It might be worth revisiting.

lib/Makefile.mingw
ahorek: removed

lib/Makefile.mingw
ahorek: removed

lib/Makefile.mingw
ahorek: Ok, the issue was with the EaH apps, they define
```
#ifdef __MINGW32__
  #include "exchndl.h"
#endif
```
but it’s something that should be fixed in the app’s code, not by removing the flag.

reverted

AenBleidd (approved):

AenBleidd:
Looks fine to me.
@ahorek, thank you for your contribution!

PR #7007, opened 2026-04-16
Mac fix recovery mode

### LINKED ISSUE #6970: After every OS update, it boots into Recovery Mode
### Describe the bug

If BOINC is installed, any macOS update or upgrade will boot into Recover mode with the following error:

Your computer started up in Recovery because a failure occurred during installation.
An error occurred migrating user data during an install.
Reinstall macOS to resolve the issue.

BOINC versions 8.2.8 and 8.2.9. Maybe earlier too. Happens on all machines, Intel or Apple Silicon. Tahoe, Sequoia, maybe earlier.

### Steps to reproduce

Install BOINC.
Do a macOS update or upgrade.

### Expected behavior

macOS updates happen as normal, boot to normal login screen.

### Screenshots

Image

### System information

macOS
Boinc 8.2.8 and 8.2.9

### Additional context

_No response_

CharlieFenton:
Fixes #6970

As part of implementing Podman support, some versions of BOINC changed the user shell setting for users _boinc_master_ and _boinc_project_ from _/usr/bin/false_ to _/bin/zsh_ (PR #6465, #6508, #6520), but this created a user directory for each user (_/Users/boinc_master_ and /_Users/boinc_project_)  and also caused MacOS system updates to drop into Recovery Mode (Issue #6970.) To fix the Recovery Mode issue we are reverting to a shell of _/usr/bin/false_ (which tells the system that this user has no shell.)

The _Run_Podman_ utility runs Podman commands as user _boinc_project_ to allow projects to access it. It had been relying on _boinc_project_'s user shell. Since this PR eliminates user _boinc_project_'s shell, this PR modifies _Run_Podman_ to instead use _root_'s user shell.

---
## Summary by cubic
Restore “no shell” for macOS BOINC service users to prevent system updates entering Recovery Mode, and adjust Podman runner to work without a user shell. Fixes #6970 while keeping project access to Podman.

- **Bug Fixes**
  - Set shells for `boinc_master` and `boinc_project` back to `/usr/bin/false` (was `/bin/zsh`) to avoid creating home dirs and triggering Recovery Mode.
  - Update `Run_Podman` to use `su -m boinc_project -c ...` so it runs with root’s shell (`/bin/sh`), since `boinc_project` has no shell.

<sup>Written for commit 62bea248c1d1ea564d735bc095cfe66d642406cd. Summary will update on new commits.</sup>

### DIFF
--- a/client/Run_Podman.cpp
+++ b/client/Run_Podman.cpp
@@ -86,7 +86,7 @@ int main(int argc, char *argv[])
     // Podman always writes its files owned by the logged in user and
     // mostly in that user's home directory. To get around this, we
     // simulate a login by user boinc_project and set environment
-    // variables for Podman to use our BOINC podman" directory instead
+    // variables for Podman to use our "BOINC podman" directory instead
     strlcpy(buf, "env XDG_CONFIG_HOME=\"/Library/Application Support/BOINC podman\" XDG_DATA_HOME=\"/Library/Application Support/BOINC podman\" HOME=\"/Library/Application Support/BOINC podman\" ", sizeof(buf));
 
     argvCount = 1;   // arguments to be passed to Podman
@@ -105,7 +105,7 @@ int main(int argc, char *argv[])
         system(buf);
     } else {
         args[argsCount++] = "su";
-        args[argsCount++] = "-l";
+        args[argsCount++] = "-m";   // boinc_project doesn't have a shell, so use root's shell /bin/sh
         args[argsCount++] = "boinc_project";    // Create Podman VM using boinc_project so projects can access it
         args[argsCount++] = "-c";
         args[argsCount++] = buf;
--- a/clientgui/mac/SetupSecurity.cpp
+++ b/clientgui/mac/SetupSecurity.cpp
@@ -954,9 +954,14 @@ static OSStatus CreateUserAndGroup(char * user_name, char * group_name) {
     }           // if (! userExists)
 
 setGroupForUser:
-    // Older versions set shell to /usr/bin/false so do this even if the user exists
-    // Something like "dscl . -create /users/boinc_master shell /bin/zsh"
-    err = DoSudoPosixSpawn(dsclPath, ".", "-create", buf2, "shell", "/bin/zsh", NULL);
+    // As part of implementing Podman support, some versions of BOINC changed the
+    // shell setting from /usr/bin/false to /bin/zsh (PR #6465, #6508, #6520), but
+    // this created a user directory (/Users/boinc_master or /Users/boinc_project)
+    // and also caused MacOS system updates to drop into Recovery Mode (Issue #6970.)
+    // To fix the Recovery Mode issue we are reverting to a shell of /usr/bin/false.
+    // Since some versions set shell to /bin/zsh, do this even if the user exists.
+    // Something like "dscl . -create /users/boinc_master shell /usr/bin/false"
+    err = DoSudoPosixSpawn(dsclPath, ".", "-create", buf2, "shell", "/usr/bin/false", NULL);
     if (err)
         return err;
 


AenBleidd (approved):

CharlieFenton:
A footnote for future reference: setting the user shell for users _boinc_master_ and _boinc_project_ to _/bin/zsh_ does not cause MacOS updates to drop into Recovery Mode if we set their home directory to _/Users/boinc_master_ and _/Users/boinc_project_ instead of to _/var/empty_ (which means no home directory.) Apparently the combination of having a user shell but no home directory was causing the problem.

But (at least for now) I prefer the method I used in this PR (eliminating the user shell) because the home directories might unnecessarily consume significant disk space. No home directory seems to be OK as long as there is also no user shell.

jameslbrown:
@CharlieFenton I installed 8.2.10 dev version, then did an OS update. I was put back into Recovery Mode. :-( 

Should I have run the uninstaller first, then installed 8.2.10?

CharlieFenton:
@jameslbrown: well, that  is very disappointing. Yes, please try running the uninstaller and then rerunning the installer and see if that makes a difference.

CharlieFenton:
@jameslbrown Please try [this](https://boinc.berkeley.edu/dl/TestInstallerForJames.zip) test build of BOINC.

jameslbrown:
That worked! No more Recovery mode. Well done!

I did uninstall before running the installer, not sure if this helped or not.

After the OS update, BOINC opened and presented me with the Add Project dialog. That was a bit unnerving. I clicked cancel. Quit BOINC and re-opened it and eventually it connected and all my projects/tasks were there. Maybe the OS hadn't finished all its setup after the update.

jameslbrown:
Just did another machine, from macOS 26.4 to 26.4.1 and it worked flawlessly. This time I didn't do the uninstall first.

jameslbrown:
@CharlieFenton should I use the version above (8.3.0) or the just-announced 8.2.11?

AenBleidd:
@jameslbrown, please use 8.2.11

CharlieFenton:
Yes please use 8.2.11

CharlieFenton:
Also , please test without first uninstalling 8.2.9 if possible. If you can also test with first uninstalling that would be a bonus.

jameslbrown:
Just tested on another computer. Installed 8.2.11 (over 8.2.9). Did NOT uninstall first. 

macOS update worked with no issues.

jameslbrown:
If I copy "BOINC Installer.app" from ~/Downloads/boinc_8.2.11_macOSX_universal to /Applications/ it will not run from there.

That is how we deployed it previously.

But the 8.3.0 one from @CharlieFenton does.

CharlieFenton:
@jameslbrown if you did not run the installer with the 8.3.0 I previously sent you but only copied the application, that would explain why it didn’t fix the recovery Mode issue for you but iit did for me. All the changes to fix this in both that one and 8.2.11 are in the installer, not in the application. (There were also some changes in the application to fix other issues.)

While you might usually get away with just copying the application, it’s not safe to do that. In fact, the last few versions of BOINC for the Mac moved the actual application elsewhere and put a symbolic link (similar to an alias but not exactly the same) in the /Applications folder as part of a different fix. So please always run the installer.  

If you want to do it remotely without user interaction, follow the instructions for command-line installs [here](https://github.com/BOINC/boinc/wiki/Tools-for-MacOS).

jameslbrown:
Hi @CharlieFenton I wasn't clear. I was talking about running the Installer app from within /Applications/, not the "BOINC Manager.app".

CharlieFenton:
I’ve never tried running the installer from the /Applications folder but I can’t think of any reason that shouldn’t work. I’m out of town at the moment but I can try it when I return home.

Where did you run the 8.2.11 installer from that did work?

jameslbrown:
It was in my Downloads folder, ~/Downloads/boinc_8.2.11_macOSX_universal/

CharlieFenton:
> It was in my Downloads folder

That’s where I usually run it, though sometimes elsewhere during preliminary testing. I’ve not tried it in the /Applications folder though I can’t think of any reason why it wouldn’t work.

CharlieFenton:
> If I copy "BOINC Installer.app" from ~/Downloads/boinc_8.2.11_macOSX_universal to /Applications/ it will not run from there.


@jameslbrown Thanks for finding this bug. I have now fixed it in PR #7048

PR #7013, opened 2026-04-18
[mac] provide a path to the system root certificates file

AenBleidd:
## Summary by cubic
On macOS, set cURL’s CA bundle to `/etc/ssl/cert.pem` when it exists to prevent HTTPS certificate verification errors. Adds a file-existence check and sets `CURLOPT_CAINFO` during request setup in `client/http_curl.cpp`.

<sup>Written for commit cf15a9f0a90a25ff23be541133decd11aa04515d. Summary will update on new commits.</sup>

### DIFF
--- a/client/http_curl.cpp
+++ b/client/http_curl.cpp
@@ -490,6 +490,13 @@ int HTTP_OP::libcurl_exec(
         curl_easy_setopt(curlEasy, CURLOPT_CAINFO, CA_BUNDLE_FILENAME);
     }
 #endif
+#ifdef __APPLE__
+    // It looks like for the Mac we need to specify the path
+    // to the system certificates file if it exists
+    if (boinc_file_exists("/etc/ssl/cert.pem")) {
+        curl_easy_setopt(curlEasy, CURLOPT_CAINFO, "/etc/ssl/cert.pem");
+    }
+#endif
 
     // set the user agent as this boinc client & version
     //


AenBleidd:
@CharlieFenton,
It looks like that we need to provide a path to the system root certificates file on MacOS since the Secure Transport is deprecated and we cannot rely on it anymore.
Please review and test this.
In particular, you can use Asteroid@home project since it's the project we always have issue with when running on Mac.

CharlieFenton (commented):
I tested this change in 2 different ways:

- with both curl_easy_setopt(curlEasy, CURLOPT_CAINFO, CA_BUNDLE_FILENAME); and curl_easy_setopt(curlEasy, CURLOPT_CAINFO, "/etc/ssl/cert.pem");
- with only curl_easy_setopt(curlEasy, CURLOPT_CAINFO, "/etc/ssl/cert.pem");

In both cases connections with projects other than Asteroids@home worked, but I got the same error as reported by @gomes for Asteroids@home:

> 9-Apr-2026 00:49:07 [---] [http] [ID#2] Info:  Host asteroidsathome.net:443 was resolved.
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:  IPv6: (none)
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:  IPv4: 195.113.29.201
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:    Trying 195.113.29.201:443...
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:  ALPN: curl offers http/1.1
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:  TLSv1.3 (OUT), TLS handshake, Client hello (1):
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:   CAfile: /etc/ssl/cert.pem
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:   CApath: none
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:  TLSv1.3 (IN), TLS handshake, Server hello (2):
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:  TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:  TLSv1.3 (IN), TLS handshake, Certificate (11):
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:  TLSv1.3 (OUT), TLS alert, unknown CA (560):
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:  SSL certificate problem: unable to get local issuer certificate
19-Apr-2026 00:49:07 [---] [http] [ID#2] Info:  closing connection #0
19-Apr-2026 00:49:07 [---] [http] HTTP error: SSL peer certificate or SSH remote key was not OK

(Note: the line "19-Apr-2026 00:49:07 [---] [http] [ID#2] Info: CAfile: /etc/ssl/cert.pem" appeared only when I tested with only `only curl_easy_setopt(curlEasy, CURLOPT_CAINFO, "/etc/ssl/cert.pem");`)

This is in addition to the problem identified by CoPilot:

> On macOS this unconditionally overrides any CURLOPT_CAINFO that may have been set just above (e.g., a configured/custom CA bundle via CA_BUNDLE_FILENAME). If the intent is to provide a fallback, this should only set /etc/ssl/cert.pem when no CA bundle has already been configured; otherwise it can silently break deployments that rely on a custom CA store.

In summary, this does not fix the problem with Asteroids@home. Since that appears to be the only project with this problem, I strongly doubt that the problem is with BOINC.

PR #7042, opened 2026-04-27
Update the type of the rss field to be long for Linux

saiarcot895:
Fixes #

**Description of the Change**

The `proc_pid_stat` manpage on my system (Ubuntu 25.10) has the type of this field as `%ld`. Update the type used here to match that printf formatting.

This also fixes a bug where applications using more than 4GB of RSS memory were shown and handled incorrectly due to integer overflow that happened during multiplication.

**Alternate Designs**

**Release Notes**

---
## Summary by cubic
Update `rss` in `PROC_STAT` from int to long and parse with `%ld` to match `proc_pid_stat`, fixing overflow and incorrect RSS reporting for processes over 4 GB.

- **Bug Fixes**
  - Changed struct field and format spec in `lib/procinfo_unix.cpp`.
  - Aligns with the manpage and prevents integer overflow when converting pages to bytes.

<sup>Written for commit 60a556ce6bce01e3bba724794f2cfd51b6b3813e. Summary will update on new commits.</sup>

### DIFF
--- a/lib/procinfo_unix.cpp
+++ b/lib/procinfo_unix.cpp
@@ -89,7 +89,7 @@ struct PROC_STAT {
     int itrealvalue;
     unsigned long starttime;
     unsigned long vsize;
-    int rss;
+    long rss;
     unsigned long rlim;
     unsigned long startcode;
     unsigned long endcode;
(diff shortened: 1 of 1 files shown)


AenBleidd (approved):

AenBleidd:
@saiarcot895, thank you for your contribution!

PR #7064, opened 2026-05-09
Client: avoid XScreenSaver probing on Wayland

sharkxuanbee:
## Summary
- Detect Wayland in daemon/non-session contexts by checking active Wayland sockets and Xwayland processes in addition to session environment variables.
- Re-evaluate Wayland detection before legacy XScreenSaver idle probing instead of caching the initial result, so clients started before a desktop session do not later probe Xwayland displays.
- Skip `XOpenDisplay()` XScreenSaver probing on Wayland/Xwayland to avoid blocking the client main thread and starving GUI RPC handling.

## Root Cause
On Ubuntu 26.04 Wayland systems, `boinc-client` can run without `WAYLAND_DISPLAY` or `XDG_SESSION_TYPE` in its environment. The existing `detect_wayland()` then returns false, and the legacy idle path enumerates `/tmp/.X11-unix/*` and calls `XOpenDisplay()` for those displays. On the reproduced system, opening the Xwayland display blocks in `unix_wait_for_peer`, leaving local GUI RPC requests unanswered; `boinccmd` then reports authorization failures and raw `<auth1/>` requests time out.

## Repro
- Start an isolated client with `--allow_multiple_clients --skip_cpu_benchmarks --gui_rpc_port 31427 --no_gpus`.
- Run `boinccmd --host localhost:31427 --passwd boinc123 --get_state`.
- Observe `Authorization failure: -102`; raw `<auth1/>` to `127.0.0.1:31427` times out.
- `strace`/`wchan` show the client stuck while connecting to `/tmp/.X11-unix/X1`.

## Verification
- `git diff --check`
- `g++ -std=c++17 -fsyntax-only ... -DHAVE_XSS=0 client/hostinfo_unix.cpp` using a temporary generated `config.h` and curl stub
- `g++ -std=c++17 -fsyntax-only ... -DHAVE_XSS=1 client/hostinfo_unix.cpp` using a temporary generated `config.h` and curl stub

`./_autosetup` could not run in this environment because `m4`/`gm4` is not installed.

---
## Summary by cubic
Prevent GUI RPC deadlocks on Wayland by skipping XScreenSaver probing and improving Wayland detection. Detection now scans sockets/processes and caches the result for idle probing.

- **Bug Fixes**
  - Detect Wayland by scanning `XDG_RUNTIME_DIR` and `/run/user/*` for `wayland-*` sockets and by checking `/proc/*/comm` for `Xwayland`.
  - Cache Wayland detection for XScreenSaver idle probing.
  - Skip `XOpenDisplay()` probing on Wayland/Xwayland to prevent main-thread blocking and stalled GUI RPC.

<sup>Written for commit a1c9bbcd79f484047890dbc47fef4059c6d4dbf8. Summary will update on new commits.</sup>

### DIFF
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -2488,7 +2488,10 @@ https://github.com/jamescowens/idle_detect");
 
 #if HAVE_XSS
 #ifndef ANDROID
-    if (!detect_wayland()) {
+    // Detect Wayland once. The display server type is not expected to
+    // change while this client process is running.
+    static bool wayland_detected = detect_wayland();
+    if (!wayland_detected) {
         //printf("Using XScreenSaver API for idle detection\n");
         idle_time = min(idle_time, xss_idle());
     }
--- a/client/hostinfo_unix.cpp
+++ b/client/hostinfo_unix.cpp
@@ -40,9 +40,12 @@
 // prevents naming collision between X.h define of Always and boinc's
 // lib/prefs.h definition in an enum.
 #undef Always
-#include <dirent.h> //for opening /tmp/.X11-unix/
-  // (There is a DirScanner class in BOINC, but it doesn't do what we want)
 #include "log_flags.h" // idle_detection_debug flag for verbose output
+#endif
+
+#if LINUX_LIKE_SYSTEM
+#include <dirent.h> //for opening /tmp/.X11-unix/ and /run/user/
+  // (There is a DirScanner class in BOINC, but it doesn't do what we want)
 #endif
 
 #include <cstdio>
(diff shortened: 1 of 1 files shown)


qixuancao:
This is related to Wayland, but it appears distinct from #7052.\n\nPR #7052 fixed a BOINC Manager crash on Wayland and touched packaging/build-related files, while this PR focuses on a client-side GUI RPC deadlock in `client/hostinfo_unix.cpp`. In the reproduced case, `boinc-client` runs without Wayland session env vars, falls back to legacy XScreenSaver/X11 probing, then blocks while probing Xwayland. That leaves local GUI RPC requests unanswered, which shows up as Manager/`boinccmd` auth timeouts or `Authorization failure: -102`.\n\nSo the symptoms are both Wayland-related, but the affected component and root cause appear different.

client/hostinfo_unix.cpp
```diff
@@ -2389,8 +2488,7 @@ utility that works on Wayland and X11: \
 
 #if HAVE_XSS
 #ifndef ANDROID
-    static bool wayland_detected = detect_wayland();
-    if (!wayland_detected) {
+    if (!detect_wayland()) {
```
AenBleidd: We need to detect this only once on startup. I doubt system during runtime will change its environment from X11 to Wayland.
@sharkxuanbee, why do you need to check this every second?
Is there any valid scenario for this?

AenBleidd (changes requested):

client/hostinfo_unix.cpp
sharkxuanbee: Agreed. I updated the patch to cache Wayland detection and run it only once before the XScreenSaver path.

The enhanced fallback detection is still kept, but it no longer scans /run/user or /proc on every idle poll.

AenBleidd (approved):

AenBleidd:
@sharkxuanbee, thank you for your contribution!

PR #6088, opened 2025-02-16
[WIP] Mac: After MacOS update, repair primary groups for users boinc_master, boinc_project

CharlieFenton:
Fixes #5588 #5910 #5975 #6044

**Description of the Change**
Every MacOS update changes the primary group of user boinc_master from group boinc_master to staff (20), and changes the primary group of user boinc_project from group boinc_project to staff (20). This PR adds code to fix them when needed. It also adds corresponding code to the command-line utility Mac_SA_Secure.sh.

**Alternate Designs**
The incorrect primary groups caused problems with use of GPUs which were difficult to diagnose. I implemented #5502 to prevent BOINC from being run with the wrong primary groups by generating permissions error -1301 telling the user to reinstall BOINC. This PR is in response to complaints from users about  needing to perform this extra step.

**Release Notes**
Automatically fix permissions error -1301 on Macs after updating the operating system.

PR #7074, opened 2026-05-12
Mac: Update PR #6088 for MacOS 26.5

CharlieFenton:
Under MacOS 26.5, getpwnam() returns the old primary group ID unless we first flush the Directory Services cache, even if the value was changed by a MacOS system updated. This PR adds the code to flush the cache before checking the primary group of boinc_master & boinc_project. It also flushes the cache after repairing them.

**Alternate Designs**
We could instead use `FILE *f = popen "dscl . -read /Users/boinc_master PrimaryGroupID" ` to check the primary group of boinc_master & boinc_project, but that would require parsing the response, which is of the form `PrimaryGroupID: 505`. This method is simpler and more robust.

**Release Notes**
Updates fixing directory permissions on MacOS updates for MacOS 26.5

---
## Summary by cubic
Flush the macOS Directory Services cache on macOS 26.5 to avoid stale primary group data when verifying and repairing BOINC users. Ensures correct groups for `boinc_master` and `boinc_project` after system updates.

- **Bug Fixes**
  - Run `/usr/bin/dscacheutil -flushcache` before `getpwnam()` checks in `client/check_security.cpp`.
  - Run the same cache flush after updating primary groups in `clientgui/mac/MacFixUserGroups.cpp`.

<sup>Written for commit d0e9a178f228cae7433387b17933d7bd3e6842fe. Summary will update on new commits.</sup>

### DIFF
--- a/client/check_security.cpp
+++ b/client/check_security.cpp
@@ -555,6 +555,8 @@ static int check_boinc_users_primarygroupIds(int useFakeProjectUserAndGroup, int
     passwd              *pw;
     group               *grp;
 
+    callPosixSpawn("/usr/bin/dscacheutil -flushcache");
+
     if ((!useFakeProjectUserAndGroup) || isMacInstaller) {
        // Require absolute owner and group boinc_master:boinc_master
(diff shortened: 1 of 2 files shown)


CharlieFenton:
This PR is ready to merge

Issue #3770, opened 2020-05-26
[GDPR / privacy] Don't ping Google without consent


brevilo:
Can we please not ["ping" Google](https://github.com/BOINC/boinc/blob/master/client/current_version.cpp#L43) just to see if the Internet is available? The BOINC client currently shares my IP with Google without my consent or any any privacy policy mentioning that. That's arguably a GDPR violation as IPs are personal data.

How about just using `boinc.berkeley.edu` instead...?

Thanks

AenBleidd:
It's less possible that `google.com` could be unavailable

brevilo:
Probably right, but that doesn't solve the issue.

AenBleidd:
@brevilo, you're tight, it doesn't. How about 8.8.8.8? It's a public DNS by Google. As far as I know it's widely used to detect connection issues.
Also we can put somewhere  an option to turn on/off pinging any address but with warning that if this option is off we will not be able to detect any connections issues properly.

brevilo:
1. The test doesn't use a real `ping` (ICMP), it uses an HTTP/GET if I'm not mistaken. So a DNS ping would require a code change, part from just the URL update. That's probably not worth it.
2. There should be only a few times where this test actually necessary. Those few times Berkeley is probably available. Their uptimes can't be _that_ bad, can they?
3. Using a Google DNS instead of their homepage doesn't really improve anyhthing WRT this issue, does it? You could probably use Cloudflare instead (1.1.1.1) which is better in terms of privacy (no IP or query logging) but it's still an external service that needs a mention in a privacy policy.

Ageless93:
There's of course cc_config.xml options available that could be added by default:
_<suppress_net_info>0|1</suppress_net_info>
If enabled, don't send this host's IP address and domain name to servers. Otherwise, this information is sent to, and stored on, servers. It is visible to you (but not other users) via the web._
_<dont_contact_ref_site>0|1</dont_contact_ref_site>
    To determine if a physical network connection exists, the client occasionally contacts a highly-available web site (google.com). If this flag is set, this behavior is suppressed. This flag also suppresses a periodic fetch of a project list from boinc.berkeley.edu._

And I thought there was another one with which you can set which site to ping, as WCG uses that to ping IBM.com instead of Google.

RichardHaselgrove:
Yes, that's
```
<network_test_url>URL</network_test_url>
The URL to use when checking network connectivity; default is http://www.google.com/.
```
[...]

PR #7084, opened 2026-05-20
Client, Manager, API, docker_wrapper: support network access from apps

### LINKED ISSUE #6399: If can't build Docker image because disconnected, handle correctly
### Describe the bug

If a Docker job starts while host is disconnected, image creation will fail.
Detect this particular type of failure, and do a temporary exit.

Ideally there should be a variant of temporary exit that says:
retry when we have a network connection

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### System information

_No response_

### Additional context

_No response_

davidpanderson:
The goal of this PR is to allow apps that do network communication
to work properly in the presence of network outages and suspension.
With these changes:

- If a job that needs network (such as a docker job) runs while
the network is disconnected, it figures this out, tells the client,
and the user is shown a notice suggesting that they reconnect.

- If such a job needs network and network access is suspended by the user
(via time-of-day prefs or the 'suspend network' command)
it sees this, tells the client that it needs to communicate,
and the user is shown a notice suggesting that they unsuspend.

This requires some logic in apps; this PR adds this to docker_wrapper.

This is described here:
https://github.com/BOINC/boinc/wiki/Apps-that-do-network-communication

Note: in ~2004 there was an unsuccessful effort to run Folding@home
as a BOINC app, and we added logic where the client is responsible
for knowing whether a connection exists, and it notifies the app.
This PR removes this logic; in the new design it's up to the app
to detect the absence of a connection.

Remove an API where the app tells the client how much data it transferred.
This info isn't used, and in the case of Docker the app doesn't know
how much data was transferred.

---
## Summary by cubic
Apps can now tell the client when they’re waiting for Internet, and the UI shows clear notices and a “Waiting for network” task state. `docker_wrapper` detects outages, retries image builds until connectivity returns, and surfaces better errors.

- **New Features**
  - API: added `boinc_waiting_for_network(bool)`; app status always includes `<want_network>`; removed `boinc_need_network()`, `boinc_network_poll()`, `boinc_network_done()`, `boinc_network_usage()`.
  - Client/Manager: shows notices when tasks need Internet or network is suspended; tracks connection vs. suspension notices separately and clears them on reconnection or when no tasks want network; GUI shows “Waiting for network”; GUI RPC exposes `want_network`; don’t strip tags in notices; removed unused bytes sent/received tracking.
  - `docker_wrapper`: waits for a heartbeat before checking suspend state; detects disconnect/suspension, sets `want_network`, and retries image builds until network returns; checks connectivity via ping to `berkeley.edu`; honors exit/abort while waiting; prints network-related logs only with verbose flags; shows command output on errors; uses `--retry 0` for Podman builds; on Windows, copies the command string before `CreateProcess`.
  - Work fetch: piggyback requests no longer skip higher‑priority projects that previously returned no jobs.
  - Lib: `run_command()` returns non‑zero on command failures and (on Windows) still returns command output; added `network_connected()`.

- **Migration**
  - Replace removed APIs with `boinc_waiting_for_network(bool)`.
  - Apps should detect connectivity and call `boinc_waiting_for_network(true/false)`; include `<want_network>` in status updates.

<sup>Written for commit cbe058ee89fdbbeae0608461447ceb00d4399e1a. Summary will update on new commits. <a href="https://cubic.dev/pr/BOINC/boinc/pull/7084?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/lib/network.cpp
+++ b/lib/network.cpp
@@ -368,12 +368,16 @@ int boinc_get_port(bool is_remote, int& port) {
 // So use ping instead.
 // But guess what?  The Win version is different from Unix
 //
-bool network_connected() {
+bool network_connected(const char *hostname) {
+    char cmd[256];
+    snprintf(cmd, sizeof(cmd), "ping %s %s",
+        hostname,
 #ifdef _WIN32
-    const char* cmd = "ping google.com -n 1";
+        "-n 1"
 #else
-    const char* cmd = "ping google.com -c 1";
+        "-c 1"
 #endif
+    );
     vector<string> out;
     int retval = run_command(cmd, out);
     // ping exits nonzero on failure
--- a/lib/network.cpp
+++ b/lib/network.cpp
@@ -360,8 +360,15 @@ int boinc_get_port(bool is_remote, int& port) {
 // is the given host reachable?
 // (used w/ google.com to check for network connection)
 //
+// We used to do this directly, by gethostbyname() and connect().
+// But on Windows, gethostbyname() caches negative results,
+// so once 'google.com' fails (due to network disconnection)
+// it keeps failing after reconnection!  WTF??
+//
+// So use ping instead.
+// But guess what?  The Win version is different from Unix
+//
 int network_connected() {
-#if 1
 #ifdef _WIN32
     const char* cmd = "ping google.com -n 1";
 #else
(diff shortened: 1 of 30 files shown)


lib/network.cpp
```diff
@@ -351,3 +356,29 @@ int boinc_get_port(bool is_remote, int& port) {
...
+// But on Windows, gethostbyname() caches negative results,
+// so once 'google.com' fails (due to network disconnection)
+// it keeps failing after reconnection!  WTF??
+//
+// So use ping instead.
+// But guess what?  The Win version is different from Unix
+//
+int network_connected() {
+#ifdef _WIN32
+    const char* cmd = "ping google.com -n 1";
```
AenBleidd: @davidpanderson, can we have reference site configurable via `cc_config.xml`?
I personally don't care but some might not want to send any statistics to any corporate website (even via ping request).
It's ok to fallback to `google.com` if there is no site configured, but it would be nice to give users a choice.
See [this ticket](#3770) for reference.

lib/network.cpp
```diff
@@ -351,3 +356,29 @@ int boinc_get_port(bool is_remote, int& port) {
...
+#else
+    const char* cmd = "ping google.com -c 1";
+#endif
+    vector<string> out;
+    int retval = run_command(cmd, out);
+    if (retval) {
+        fprintf(stderr, "%s failed: %d\n", cmd, retval);
+    }
+    // ping exits nonzero on failure
+    return 0;
```
AenBleidd: @davidpanderson, please validate this logic. I believe you should return `retval` here

AenBleidd (changes requested):

lib/network.cpp
davidpanderson: Yup.  Cubic caught that too.

lib/network.cpp
davidpanderson: I added this (using the test_network_url that we already have in nvc_config.xml)

AenBleidd (approved):

PR #7108, opened 2026-06-03
client: enforce swap usage limit pref

davidpanderson:
Previously we had the pref but ignored it.
Enforcing it is tricky because to a free a process's swap space,
you have to kill it, losing CPU time since last checkpoint.
Details are here: https://github.com/BOINC/boinc/wiki/Memory-management

---
## Summary by cubic
Enforces the swap usage limit (`vm_max_used_frac`) with scheduler logic that kills and revives tasks based on swap pressure, and tightens RAM handling. This reduces out-of-swap and OOM errors, and shows clearer UI states for memory vs swap limits.

- New Features
  - Enforce swap limit via `swap_limit_check()`:
    - When over limit, kill running tasks ordered by least uncheckpointed CPU time; revive killed tasks in deadline order as swap frees up.
    - Skip tasks whose estimated swap would exceed `swap_left`; trigger reschedule when total WSS, total swap, or any task’s WSS exceed current limits.
  - BOINC Manager shows “Waiting for swap space”; startup logs the effective swap cap.

- Refactors
  - Replace `too_large` with `wss_too_large` and add `swap_too_large` across client and GUI RPC; keep XML compatibility by still emitting `<too_large/>`.
  - RAM policy: if WSS > available RAM, suspend/wait; if WSS > max allowed RAM, abort.
  - Run-list enforces both RAM (`ram_left`) and swap (`swap_left`); preempt WSS with suspend, swap with quit.

<sup>Written for commit 12389396eb32009fcfe7132def063eec9df4dd8e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/BOINC/boinc/pull/7108?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -428,6 +428,10 @@ void ACTIVE_TASK_SET::get_memory_usage() {
         if (pi.swap_size > atp->peak_swap_size) {
             atp->peak_swap_size = pi.swap_size;
         }
+        boinc_total.working_set_size += pi.working_set_size;
+        boinc_total.working_set_size_smoothed += pi.working_set_size_smoothed;
+        boinc_total.swap_size += pi.swap_size;
+        boinc_total.page_fault_rate += pi.page_fault_rate;
 
         if (!first) {
             int pf = pi.page_fault_count - last_page_fault_count;
@@ -444,10 +448,6 @@ void ACTIVE_TASK_SET::get_memory_usage() {
                     pi.user_time,
                     pi.kernel_time
                 );
-                boinc_total.working_set_size += pi.working_set_size;
-                boinc_total.working_set_size_smoothed += pi.working_set_size_smoothed;
-                boinc_total.swap_size += pi.swap_size;
-                boinc_total.page_fault_rate += pi.page_fault_rate;
             }
         }
     }
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -151,6 +151,7 @@ ACTIVE_TASK::ACTIVE_TASK() {
     sporadic_ca_state = CA_NONE;
     sporadic_ac_state = AC_NONE;
     sporadic_ignore_until = 0;
+    swap_kill_time = 0;
 }
 
 bool ACTIVE_TASK::process_exists() {
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -128,7 +128,8 @@ ACTIVE_TASK::ACTIVE_TASK() {
     max_mem_usage = 0;
     have_trickle_down = false;
     send_upload_file_status = false;
-    too_large = false;
+    wss_too_large = false;
+    swap_too_large = false;
     needs_shmem = false;
     want_network = 0;
     abort_time = 0;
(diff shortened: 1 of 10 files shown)


AenBleidd (approved):

CharlieFenton:
I think this logic has some problems, at least one of which is serious.

If I understand this code correctly, this is supposed to examine the swap space used by each BOINC process to determine whether BOINC is using more than the permitted amount of swap space. The Mac (_procinfo_mac.cpp_) and, I believe, Linux (_procinfo_unix.cpp_) set _procinfo.swap_size_ to the virtual size (_vsz_ from the `PS` command.) But from everything I've read, that virtual size value shown on these operating systems is the **maximum** amount of total memory (real memory + swap space) made **available** to that process by the OS, not the amount of swap space currently used. Since a process rarely uses all of that maximum, it does not even indicate the amount of real memory + swap space currently used by the process. (At least on the Mac, the OS kills any process which tries to exceed its allocated virtual memory space, though it appears that a process can request additional space.)

To complicate matters, I have seen that the virtual size of a process can change over time, perhaps because the process has requested more space.

The good news is that I think I have finally found a reasonably efficient way to determine the actual amount of swap space currently used by a process on the Mac. 

Another possible issue is the determination of the total amount of swap space  on the system. The Mac uses all free space on the boot volume as swap space, and lines 1637-1638 of hosting_unix.cpp tries to determine that:
```
    statfs(".", &fs_info);
    m_swap = (double)fs_info.f_bsize * (double)fs_info.f_bfree;
```

But this assumes that the current working directory is on the boot volume. While this is usually true, a user can move the BOINC Data directory to a different volume as described [here](https://github.com/BOINC/boinc/wiki/Tools-for-MacOS#moving-boinc-manager-or-boinc-data-folder-to-a-different-drive). This should be easy to fix.

davidpanderson:
What are the problems with the logic?
I know that virtual size is not the same as swap space used;
but it's an upper bound on it

CharlieFenton:
As I wrote, the Mac (_procinfo_mac.cpp_) and, I believe, Linux (_procinfo_unix.cpp_) set procinfo.swap_size to the virtual size (_vsz_ from the `PS` command.)

The virtual size of tasks will usually be significantly greater than the amount of swap space they ever actually use. If we use the total of the virtual sizes of all active tasks rather than their actual swap space used, _CLIENT_STATE::swap_limit_check()_ is likely to kill tasks more often than necessary. In other words, using virtual size here instead of swap space actually used causes the test to be much too sensitive.

As an example, the Einstein tasks I've been running have a virtual size well over 40GB, but I've never seen them use even 10GB of swap space, and rarely over 5GB. 

The same is true in _ACTIVE_TASK_SET::get_memory_usage()_:
`        boinc_total.swap_size += pi.swap_size;`
   and
```
    if (boinc_total.swap_size
        > (gstate.global_prefs.vm_max_used_frac)*gstate.host_info.m_swap
    ) {
        gstate.request_schedule_cpus("Swap limit exceeded");
    }
```

PR #7110, opened 2026-06-05
Upgrade MacOS projects to use c++17

AenBleidd:
## Summary by cubic
Upgrade macOS Xcode projects (boinc, zip, UpperCase2, vboxwrapper) to C++17 by setting `CLANG_CXX_LANGUAGE_STANDARD` to `c++17` in Debug/Release where needed. This enables modern C++ features, standardizes builds, and requires Xcode/Clang with C++17 support.

<sup>Written for commit cbee39c3bbe16323d56c11fa71c44487a2ff8801. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/BOINC/boinc/pull/7110?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mac_build/boinc.xcodeproj/project.pbxproj
+++ b/mac_build/boinc.xcodeproj/project.pbxproj
@@ -5161,7 +5161,6 @@
 		DD9843E209920F220090855B /* Deployment */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				CLANG_CXX_LANGUAGE_STANDARD = "c++17";
 				CREATE_INFOPLIST_SECTION_IN_BINARY = YES;
 				INFOPLIST_FILE = "BoincCmd-Info.plist";
 				OTHER_LDFLAGS = "-lboinc";
@@ -5367,7 +5366,6 @@
 		DD9E2376091CBDAE0048316E /* Development */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				CLANG_CXX_LANGUAGE_STANDARD = "c++17";
 				CREATE_INFOPLIST_SECTION_IN_BINARY = YES;
 				INFOPLIST_FILE = "BoincCmd-Info.plist";
 				OTHER_LDFLAGS = "-lboinc";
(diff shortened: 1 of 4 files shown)


AenBleidd:
@CharlieFenton, please review and merge if it looks ok to you.

CharlieFenton:
According to Google, the default C++ language on the current version of Xcode is GNU++20. There are also options for C++20, C++23 and GNU++23, in addition to C++17 and GNU++17.

 BOINC is set to use the default C language, which is GNU23. There is also an options for C17, GNU17, C23 and GNU++23. 

Does it make more sense to just use the compile defaults?

In any case, it is generally not recommended to edit the Xcode project file directly. I can do it for you using the Xcode GUI.

AenBleidd:
@CharlieFenton, this is the very minimal change, and I doubt it required XCode GUI to make it.
However, if you feel there is a better way to use C++17 or later in XCode - please feel free to do that.

CharlieFenton:
The script mac_build/buildwxMac.sh specifically sets building wxWidgets to use C11 and C++11. Since this works and changing it might cause problems with wxWidgets. I don't think it should cause any incompatibilities with BOINC Manager even if BOINC Manager is built with newer versions of C and C++. I am inclined to leave it alone. What do you think?

AenBleidd:
@CharlieFenton, I would leave dependencies with the flags they have now, and do the change to out code only.

CharlieFenton:
Should we use C++17 or the compiler default (GNU++20) or one of the other options?

> According to Google, the default C++ language on the current version of Xcode is GNU++20. There are also options for C++20, C++23 and GNU++23, in addition to C++17 and GNU++17.

AenBleidd:
I would keep the compiler default.
It's important for the libboinc to be compatible with c++11, but this we usually check with Linux and some other ci pipelines.
So please remove any hardcoded c++ standards in BOINC Projects while keeping 3parry dependencies' scripts as-is.

CharlieFenton:
@AenBleidd I fixed some things that weren't quite correct. Please merge when CI checks complete.

CharlieFenton (approved):

PR #7112, opened 2026-06-05
prefer openssl certstore on macos

### LINKED ISSUE #6987: [macOS] SSL Handshake Timeout on Tahoe 26.4.1 (M4) despite native certificate integration
### Describe the problem

Yes. On a fresh installation of BOINC 8.2.9 on macOS Tahoe 26.4.1 (M4 Mac Mini), the client fails to attach to projects (specifically tested with Asteroids@home) due to SSL handshake timeouts.

The client log shows a 23-second hang at get_project_config.php followed by a communication failure. This occurs even though the system trust store is fully updated.

### Describe the solution you'd like

The BOINC client should be updated to handle the stricter TLS/SSL handshake protocols in macOS Tahoe.

If the current implementation relies on a static ca-bundle.crt or a flawed symlink to the system keychain, it needs to be replaced with a robust call to the native Security Framework to ensure modern hardware can connect to legacy project servers without manual terminal intervention.

### Additional context

This is a critical blocker for "non-technical users" on 2026-gen Mac hardware.
Logs show:
Fetching configuration file from https://asteroidsathome.net/boinc/get_project_config.php
[---] Exiting (after 23s timeout)

ahorek:
Fixes #6987

**Description of the Change**
It appears that /etc/ssl/cert.pem are outdated even on recent Macos versions which is causing trouble with recognizing new certificates.

**Alternate Designs**

**Release Notes**
Macos now prefers openssl certificates
/opt/homebrew/etc/ca-certificates/cert.pem
/usr/local/etc/ca-certificates/cert.pem
rather than the default
/etc/ssl/cert.pem

the certificate store can be overridden using a local
cert.pem

---
## Summary by cubic
Prefer OpenSSL CA bundles on macOS to fix TLS failures caused by outdated `/etc/ssl/cert.pem`. The client now picks a better CA file automatically to recognize newer certificates.

- **Bug Fixes**
  - On macOS, set `CURLOPT_CAINFO` using the first existing path in this order: `cert.pem` (local), `/opt/homebrew/etc/ca-certificates/cert.pem`, `/usr/local/etc/ca-certificates/cert.pem`; otherwise fall back to `/etc/ssl/cert.pem`.
  - No changes for Windows or non-macOS builds.

<sup>Written for commit 4e97f29ae051ece27191ff3764bfcf555a9b768b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/BOINC/boinc/pull/7112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/client/http_curl.cpp
+++ b/client/http_curl.cpp
@@ -478,7 +478,7 @@ int HTTP_OP::libcurl_exec(
     curl_easy_setopt(curlEasy, CURLOPT_SSL_VERIFYPEER, 1L);
     //curl_easy_setopt(curlEasy, CURLOPT_SSL_VERIFYPEER, FALSE);
 
-    // MSW now uses schannel and Mac now uses Secure Transport
+    // MSW now uses schannel
     // so neither uses ca-bundle.crt
 #if (!defined(_WIN32) && !defined(__APPLE__))
     // if the above is nonzero, you need the following:
@@ -489,6 +489,19 @@ int HTTP_OP::libcurl_exec(
         //
         curl_easy_setopt(curlEasy, CURLOPT_CAINFO, CA_BUNDLE_FILENAME);
     }
+#endif
+
+#ifdef __APPLE__
+    // local override, call this only if a local copy of cert.pem exists
+    if (boinc_file_exists("cert.pem")) {
+        curl_easy_setopt(curlEasy, CURLOPT_CAINFO, "cert.pem");
+    // prefer openssl versions
+    } else if (boinc_file_exists("/opt/homebrew/etc/ca-certificates/cert.pem")) {
+        curl_easy_setopt(curlEasy, CURLOPT_CAINFO, "/opt/homebrew/etc/ca-certificates/cert.pem");
+    } else if (boinc_file_exists("/usr/local/etc/ca-certificates/cert.pem")) {
+        curl_easy_setopt(curlEasy, CURLOPT_CAINFO, "/usr/local/etc/ca-certificates/cert.pem");
+    }
+    // else the default "/etc/ssl/cert.pem"
 #endif
 
     // set the user agent as this boinc client & version


AenBleidd:
@CharlieFenton, please take a look at this. I would still prefer using SecTrust, but this might be an alternative solution.

CharlieFenton:
None of these exist on my Mac:
"/Library/Application Support/BOINC Data/cert.pem"
"/opt/homebrew/etc/ca-certificates/cert.pem"
"/usr/local/etc/ca-certificates/cert.pem"
so your code will fall back on "/etc/ssl/cert.pem".

Does this actually do anything?

CharlieFenton:
@AenBleidd : I just noticed [this](https://eissing.org/icing/posts/curl-apple-sectrust/):
> Today we merged a new feature in curl: support for Apple SecTrust. This will become available in curl 8.17.0

I made the test build for Jerome using:
> cURL libraries: libcurl/8.16.0 OpenSSL/3.0.0 zlib/1.2.12 c-ares/1.28.1

So I just checked my curl build log from yesterday and it says:
> configure: WARNING: unrecognized options: --with-apple-sectrust

I will make a new test build for Jerome to try in the next day or 2 using a newer version of curl (probably 8.20.0) and --with-apple-sectrust.

CharlieFenton:
@AenBleidd:
> I will make a new test build for Jerome to try in the next day or 2 using a newer version of curl (probably 8.20.0) and --with-apple-sectrust.

Should I also include the changes from this P, or should we try it first using only -with-apple-sectrust?

ahorek:
@CharlieFenton 
Please try updating curl 8.7.0+ and enable -with-apple-sectrust first. If this works, it would be a preferable option, but I don't have an Apple device to test it myself.

Apparently, the command-line version of curl wasn't built with Apple SecTrust support, even though curl 8.7.1 supports it.
curl 8.7.1 (x86_64-apple-darwin25.0) libcurl/8.7.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.12 nghttp2/1.68.0

> None of these exist on my Mac:
"/Library/Application Support/BOINC Data/cert.pem"
"/opt/homebrew/etc/ca-certificates/cert.pem"
"/usr/local/etc/ca-certificates/cert.pem"
so your code will fall back on "/etc/ssl/cert.pem".

Does this actually do anything?

It only works with the Homebrew version, which includes a more up-to-date certificate bundle. So it's not a solution for those who don't use it.

ahorek:
@CharlieFenton, could you open a PR for the Apple SecTrust change? If it works, that would be the preferred approach.

CharlieFenton:
@ahorek Please download and test [this release candidate](https://boinc.berkeley.edu/dl/boinc_8.2.15_macOSX_universal.zip).

PR #7123, opened 2026-06-10
client: don't try enforce swap limit on Mac

davidpanderson:
The way we get virtual size on the Mac (ps vsz) returns huge numbers, like 35GB for a small program.
This produces spurious out-of-swap events.
In any case, MacOS doesn't have a dedicated swap area, so there's no real limit anyway.

---
## Summary by cubic
Disable swap limit enforcement and reporting on macOS to prevent false “out-of-swap” events from inflated `ps` virtual size. On other platforms, swap is only enforced and shown when it can be measured.

- **Bug Fixes**
  - Added `is_swap_defined()`; always false on macOS, true only if `m_swap > 0` elsewhere.
  - Gated swap checks in memory accounting and CPU scheduler (swap_limit_check, enforce_run_list).
  - Only print virtual/swap info and swap-related prefs when swap is defined; otherwise show “Memory: X physical”.

<sup>Written for commit a748ca5150ef65d05a4449c1a3791a509944395a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/BOINC/boinc/pull/7123?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


AenBleidd:
Windows build fail is expected. I'll fix it in a separate PR.

AenBleidd (approved):

PR #7165, opened 2026-07-03
[ci] add fmtlib

AenBleidd:
## Summary by cubic
Add `fmt` as a required dependency across Linux, macOS, and Windows. CI installs it on all arches, Autotools enforces it, and macOS builds a universal static `libfmt.a` with updated build scripts and docs.

- **Dependencies**
  - Added `fmt` to all `vcpkg.json` configs (apps/libs/client/msbuild/android/mingw/manager) and the tests `boinc` port.
  - Linux CI: install `libfmt-dev` for x64, i386, arm64, and armhf in both `linux.yml` and `linux-release.yml` (including non-`vcpkg` arm paths).
  - Autotools: detect `fmt` via `pkg-config` or fallback header/link check; clear error if missing.
  - macOS: new `mac_build/buildfmt.sh` builds static `libfmt.a` for x86_64/arm64 (10.13 min); integrated into `buildMacDependencies.sh`, `dependencyNames.sh`, and `setupForBOINC.sh`; docs updated to include `fmt`; minor fixes in `buildlibzip.sh`; use `wget -O` for downloads.
  - Link `fmt` only where used (`boinc`, `boincapi`, `boinc_zip`); tests add `find_dependency(fmt)`; Windows resolves `fmt` via `vcpkg`; fixed `vboxwrapper.vcxproj` additional library paths.

- **Migration**
  - Linux: install `libfmt-dev` (Debian/Ubuntu) or `libfmt-devel` (Fedora), or use `vcpkg`.
  - macOS: install CMake command-line tools; `buildMacDependencies.sh`/`setupForBOINC.sh` build `fmt` automatically.
  - Windows: `vcpkg` pulls `fmt`; clean rebuild if linker errors appear.

<sup>Written for commit bc28b5852aaec33dbf02df4336ec56c333225710. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/BOINC/boinc/pull/7165?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


CharlieFenton:
@AenBleidd I have fixed a few issues in the Mac build scripts and updated the Mac build instructions. The local build seems to work correctly on my Mac. If it works correctly for the CI build, then I think it is ready from my point of view.

AenBleidd:
@CharlieFenton, thank you for your fixes.
I will wait for a new release of fmtlib that contains fixes for x86 builds, and then we will be reado for the merge.

CharlieFenton:
@AenBleidd : 2 questions.

[1] Do you expect to use libfmt mainly in the Manager? If so, I can update the Xcode project to link it to the Manager and add the headers to the header search path. Should i do that?

[2] The Mac Manager uses a precompiled prefix header _clientgui/mac/MacGUI.pch_ which includes all the wxWidgets headers, _config.h_ and _stdlib.h_ plus a few other items. If we expect to use libfmt widely (like we now use stdlib) I could add the `#include fmt/format.h` to this file. But that only makes sense if there is a corresponding way to do it for Windows and Linux builds. If those require adding the `#include fmt/format.h` to the individual .cpp files, then nothing  is to be gained from putting it in the precompiled header for the Mac.
Should I add the `#include fmt/format.h` to _clientgui/mac/MacGUI.pch_?

AenBleidd:
@CharlieFenton,

> Do you expect to use libfmt mainly in the Manager?

I would like to use it everywhere since it's more comfortable and more powerfull with the better and simpler syntax.
But I definitely don't want to start refactoring just to switch to this library.
We were talking with David recently about this, and his idea was to start using this library in the docker_wrapper first (since this is the component we are extensively working on during the last sveral months).
So I would not add fmtlib globally and instead do that when time comes.
Of course, if you want to do that in advance - feel free to do that.
However, I believe, doint this now will make the whole building process a little bit longer because it will be trying to link the library that is not used in aone or another prohect (and eventually will not do that since there is no any symbols imported from that library) .

> Should I add the #include fmt/format.h to clientgui/mac/MacGUI.pch?

The answer is basically the same: I would not do that unless we start using it in more that one file.
For Windows we do use precompiled headers (they are not unified at the moment, but this is something I fix later, since currently it works and is not a big time loss during the compilation phase).
For Linux we do not using precompiled headers at all, but this is something I plan to fix.

So just to summarize: I would not include it into the list of precompiled headers and would not include this library by default until we start using it actively.

davidpanderson:
In general, .h files should be included only in files that directly use them

AenBleidd:
@davidpanderson, yes, but if any external header file is heavily used across several files - having it precompiled will significantly lower compilation time.
This is expecially applicable to those libraries that use heavy templates (e.g. nlohmann-json, fmtlib, etc).

PR #7191, opened 2026-07-27
client: fix swap limit stopping jobs from running

### LINKED ISSUE #7132: Waiting for swap space limits running tasks on Linux
### Describe the bug

The default swap space on Linux for all my computers is 8GB, therefore with a limit of 100 % I can only run 1 tasks for ATLAS as the VM is configured to run with ~ 4GB each.  I would need to overide the linux defaults to set swap space = system memory to run a full complment of tasks.

Also the running task is marked as waiting for swap space in the UI

### Steps to reproduce

1. Install 8.2.15
2. Set app_config for LHC Atlas to  <cmdline>--memory_size_mb 4096</cmdline>
3. Start BOINC
4. Observer only 1 tasks running

### Expected behavior

Not sure, seems like there should be some limit based on physical memory

### Screenshots

Image

### System information

Ubuntu 26.04
BOINC 8.2.15

### Additional context

_No response_

### LINKED ISSUE #7146: Don't used more than xx.xx of swap space
### Describe the bug

I upgraded my Android phone to the latest BOINC version 8.2.15 and now can only run 2 of 3 tasks concurrently on World Community Grid or Einstein.

Might be related to issue 7132
https://github.com/BOINC/boinc/issues/7132

### Steps to reproduce

Download, install & run latest version, try to run 3 concurrent tasks as specified.

### Expected behavior

Eliminate the check of swap file size on Android since there is no way for a user to change it without jailbreak phone.

### Screenshots

_No response_

### System information

OS: Android 3.18.120-perf-gfb8faa3398e9 (Android 9)
Version: BOINC for Android 8.2.15
Memory	2.79 GB
Swap space	0.5 GB
Total disk space	24.6 GB
Free Disk Space	5.28 GB

### Additional context

6/21/2026 11:02:49	Reading preferences override file	
6/21/2026 11:02:49	Computing preferences:	
6/21/2026 11:02:49	-  When computer is in use	
6/21/2026 11:02:49	-     'In use' means mouse/keyboard input in last 3.00 minutes	
6/21/2026 11:02:49	-     don't compute	
6/21/2026 11:02:49	-     max CPUs used: 3	
6/21/2026 11:02:49	-     Use at most 100% of the CPU time	
6/21/2026 11:02:49	-     suspend if non-BOINC CPU load exceeds 100%	
6/21/2026 11:02:49	-     max memory usage: 2.79 GB	
6/21/2026 11:02:49	-  When computer is not in use	
6/21/2026 11:02:49	-     max CPUs used: 3	
6/21/2026 11:02:49	-     Use at most 100% of the CPU time	
6/21/2026 11:02:49	-     suspend if non-BOINC CPU load exceeds 100%	
6/21/2026 11:02:49	-     max memory usage: 2.79 GB	

Never saw this before upgrading BOINC Android Version 8.2.15 >>>> 6/21/2026 11:02:49	-  Don't used more than 0.50 GB swap space (100% of 0.50 GB)	 <<<<

6/21/2026 11:02:49	-  Leave apps in memory if not running	
6/21/2026 11:02:49	-  Store at least 0.10 days of work	
6/21/2026 11:02:49	-  Store up to an additional 0.00 days of work	
6/21/2026 11:02:49	-  max disk usage: 5.01 GB	
6/21/2026 11:02:49	-  (to change preferences, visit a project web site or select 'Options / Computing preferences...' in the Manager)

KNineLat:
Fixes #7132 and #7146.

In both cases the swap limit stops jobs from running even though nothing is
swapping.

On Linux the client estimates an unstarted job's swap usage from the app
version's max RSS, which is the wrong quantity. It charges a job for memory it
may never swap out, so with 8 GB of swap and a 4 GB app you get two jobs no
matter what. I now track max swap usage separately. When nothing is swapping
that's zero, so it imposes no limit.
On Android swap is zram, which the RAM limit already covers and which you can't
resize without root. I excluded it there, the same way it already is on macOS.

Three other things turned up while I was in there. `procinfo_app()` counts a
process twice when an app reports a helper PID that is also a descendant of its
main process, the swap-kill scheduler can stall the client permanently, and
`defaults()` sets `vm_max_used_frac` twice. Details are in the commit messages.

Each commit stands alone and builds. I tested the Android and desktop clients.
I couldn't exercise the swap-kill and revive paths, since they need real swap
pressure with `vm_max_used_frac` below 100%.

I think this is worth releasing sooner than usual.
The client simply runs fewer tasks than it should.

Happy to change anything or drop parts of it. I'd rather see the problem fixed
than these particular patches merged, and if the approach is wrong, I hope it is still useful.

LLMs usage: I used Claude (opus 4.6 and deepseek v4 preview) for the commit messages and to clean up my English.

---
## Summary by cubic
Fixes swap-limit logic that stalled jobs even when nothing was swapping by using real swap usage, keeping enforcement out of scheduling, and disabling swap limits on Android. Also fixes process accounting overcounts and makes swap-kill/revive reliable so jobs always make progress. Fixes #7132 and #7146; addresses #7191.

- **Bug Fixes**
  - Estimate unstarted jobs’ swap from `APP_VERSION::max_swap_usage` (from live running tasks), not RSS; use it in scheduling.
  - Keep swap-limit handling out of scheduling: `swap_limit_check()` only kills/revives; always build/enforce the run list; skip non-fitting revivals and, if nothing fits, revive one with a one-pass exemption (cleared on all exits, including GPU-usable early return).
  - Avoid pointless kills: don’t kill zero-swap tasks and don’t kill the last task that holds swap; exempt these from the swap check.
  - Never refuse zero-swap jobs even if swap_left is negative.
  - Set/clear `swap_kill_time` consistently and hand enforcement quits to the revive loop; drop stale swap figures when a process has exited.
  - Disable swap enforcement on Android (zram) by treating swap as undefined (same as macOS).
  - Correct process accounting: skip already-scanned PIDs when `other_pid` overlaps children and in child aggregation; run the graphics-executable check for every process.
  - Remove dead assignment in `GLOBAL_PREFS::defaults()` so `vm_max_used_frac` stays 1.0.

<sup>Written for commit 57a7548f30be90a1c897a20aefa4ca708e60cfed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/BOINC/boinc/pull/7191?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/client/app.cpp
+++ b/client/app.cpp
@@ -152,6 +152,7 @@ ACTIVE_TASK::ACTIVE_TASK() {
     sporadic_ac_state = AC_NONE;
     sporadic_ignore_until = 0;
     swap_kill_time = 0;
+    ignore_swap_limit = false;
 }
 
 bool ACTIVE_TASK::process_exists() {
--- a/client/app.h
+++ b/client/app.h
@@ -175,7 +175,8 @@ struct ACTIVE_TASK {
         // last time this task was killed to free swap space
     bool ignore_swap_limit;
         // run this job even if it's over the swap limit;
-        // set when nothing else can run.
+        // set when it's the only one holding swap, or when nothing
+        // else can run.
         // enforce_run_list() clears it on every exit path
 
     APP_CLIENT_SHM app_client_shm;
--- a/client/app.h
+++ b/client/app.h
@@ -173,6 +173,10 @@ struct ACTIVE_TASK {
     double last_deadline_miss_time;
     double swap_kill_time;
         // last time this task was killed to free swap space
+    bool ignore_swap_limit;
+        // run this job even if it's over the swap limit;
+        // set when nothing else can run.
+        // enforce_run_list() clears it on every exit path
 
     APP_CLIENT_SHM app_client_shm;
         // core/app shared mem segment
--- a/client/client_state.h
+++ b/client/client_state.h
@@ -323,7 +323,7 @@ struct CLIENT_STATE {
     void make_run_list(vector<RESULT*>&);
     bool enforce_run_list(vector<RESULT*>&);
     void append_unfinished_time_slice(vector<RESULT*>&);
-    bool swap_limit_check();
+    void swap_limit_check();
 
     double runnable_resource_share(int);
     void adjust_rec();
--- a/client/client_types.cpp
+++ b/client/client_types.cpp
@@ -873,6 +873,7 @@ void APP_VERSION::init() {
     graphics_exec_path[0] = 0;
     graphics_exec_file[0] = 0;
     max_rss = 0;
+    max_swap_usage = 0;
     is_vbox_app = false;
     is_docker_app = false;
     is_wrapper = false;
--- a/client/client_types.h
+++ b/client/client_types.h
@@ -372,6 +372,10 @@ struct APP_VERSION {
         // to use this much RAM,
         // so that we don't run a long sequence of jobs,
         // each of which turns out not to fit in available RAM
+    double max_swap_usage;
+        // same, for swap space.
+        // Zero unless this app version's jobs are actually swapping,
+        // so it imposes no limit on a system under no memory pressure
     bool is_vbox_app;
         // set if plan class includes "vbox"
     bool is_docker_app;
--- a/client/cpu_sched.cpp
+++ b/client/cpu_sched.cpp
@@ -794,11 +794,9 @@ bool CLIENT_STATE::schedule_cpus() {
     //
     tasks_throttled = false;
 
-    if (!swap_limit_check()) {
-        make_run_list(run_list);
-        return enforce_run_list(run_list);
-    }
-    return true;
+    swap_limit_check();
+    make_run_list(run_list);
+    return enforce_run_list(run_list);
 }
 
 // sort by increasing cpu_time - checkpoint_cpu_time;
@@ -825,22 +823,23 @@ static bool compare_swap_revive(void *p1, void *p2) {
     return atp1 < atp2;
 }
 
-// if swap limit has been reached, we use different scheduling logic.
-// Return true if this is the case.
+// Enforce the swap limit: kill jobs if we're over it, revive
+// swap-killed ones if we're under it.  Scheduling itself is
+// make_run_list()'s and enforce_run_list()'s business either way.
 //
-bool CLIENT_STATE::swap_limit_check() {
+void CLIENT_STATE::swap_limit_check() {
     if (!is_swap_defined()) {
-        return false;
+        return;
     }
-
-    vector<RESULT*> run_list;
 
     // compute swap usage of running tasks
     //
     double swap_usage = 0;
+    int n_executing = 0;
     for (ACTIVE_TASK *atp: active_tasks.active_tasks) {
         if (atp->task_state() == PROCESS_EXECUTING) {
             swap_usage += atp->procinfo.swap_usage;
+            n_executing++;
         }
     }
 
@@ -877,13 +876,6 @@ bool CLIENT_STATE::swap_limit_check() {
                 break;
             }
         }
-        // runnable list is the rest of the tasks
-        for (ACTIVE_TASK *atp: active_tasks.active_tasks) {
-            if (atp->task_state() == PROCESS_EXECUTING) {
-                run_list.push_back(atp->result);
-            }
-        }
-        enforce_run_list(run_list);
     } else {
         // here currently running tasks fit in swap
         //
@@ -894,9 +886,9 @@ bool CLIENT_STATE::swap_limit_check() {
                 swap_kill_tasks.push_back(atp);
             }
         }
-        // if no swap-killed tasks, use normal scheduling
+        // if no swap-killed tasks, there's nothing to revive
         if (swap_kill_tasks.empty()) {
-            return false;
+            return;
         }
 
         if (log_flags.cpu_sched_debug) {
@@ -905,23 +897,20 @@ bool CLIENT_STATE::swap_limit_check() {
             );
         }
 
-        // initial run list is running tasks
-        //
-        for (ACTIVE_TASK *atp: active_tasks.active_tasks) {
-            if (atp->task_state() == PROCESS_EXECUTING) {
-                run_list.push_back(atp->result);
-            }
-        }
         // see if we can run swap-killed tasks
         std::sort(
             swap_kill_tasks.begin(), swap_kill_tasks.end(),
             compare_swap_revive
         );
-        // run as many as will fit
+        // revive as many as will fit.
+        // The list is in deadline order; skip one that doesn't fit
+        // rather than stopping, so it can't veto the smaller jobs
+        // behind it
         //
+        int n_revived = 0;
         for (ACTIVE_TASK *atp: swap_kill_tasks) {
             if (swap_usage + atp->procinfo.swap_usage > swap_limit) {
-                break;
+                continue;
             }
             if (log_flags.cpu_sched_debug) {
                 msg_printf(atp->result->project, MSG_INFO,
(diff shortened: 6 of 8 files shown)


AenBleidd:

KNineLat:
Bear with me for a moment before reviewing, the cubic comment is a real bug i believe. Fixing it now.

KNineLat:
I believe we are good to go now that 'cubic ai' has no further issues. Sadly, this PR grew much larger than I would have liked.

davidpanderson:
In master, on Linux (including Android) swap usage is taken from the VmSwap field of /proc/pid/status.
It's not derived from RSS or virtual size.
As far as I can tell, this is correct.

Please set the mem_usage_debug flag in cc_config.xml.
This shows mem usage info, including swap usage, in the stdout log.
If this is showing the wrong value (compared with /proc/pid/status), let me know.

I'm also not convinced that there's a problem with the logic for restarting jobs
that were killed because of swap usage.
The logic is conservative to avoid a situation where jobs are repeatedly killed.

KNineLat:
Thanks for looking at it. I'll be working long hours the next few days so it'll take me a bit, I'll ping when I'm ready.

I want to put together a Podman example that shows the scheduling misbehaving, so the problem is easy to see without guessing. I'm also going to drop c26bc48 and do a small rework of the rest, so expect a force push. That way cubic can take a look too.

KNineLat:
The VmSwap reading is unchanged, and the `mem_usage` lines agree with
`/proc/pid/status`.

One commit does touch the measurement, though not that part of it. `procinfo_app()`
counts a process twice when an app reports an `<other_pid>` that is also a child of
its main process: `add_child_totals()` walks it, and then the outer loop adds the
same pid again because it never checked `p.scanned`. CPU time, RSS, virtual size
and swap usage all come out doubled for that process. That one stands on its own.

What the PR is about: when a job has no running process there is no `VmSwap` to
read, so `enforce_run_list()` falls back to this:

```cpp
    } else {
        erss = rp->avp->max_rss;
        eswap = rp->avp->max_rss;  // best guess
    }
```

and that is where an RSS figure ends up being compared against swap space. You
asked for `mem_usage_debug`, so here it is on current master: four jobs of one app
version, 800 MB resident each, swap limit 480 MB, and nothing on the host swapping
at all.

```
23:22:02 [mem_usage] test_result_0: virtual size 0.78 GB, RSS 0.78 GB, swap usage 0.00 GB
23:22:02 [mem_usage] test_result_1: virtual size 0.78 GB, RSS 0.78 GB, swap usage 0.00 GB
23:22:02 [mem_usage] BOINC totals: RSS 1.57 GB, swap usage 0.00 GB
23:22:03 [mem_usage] enforce: available RAM 28.15 GB, swap space 0.47 GB
23:22:03 [cpu_sched_debug] skipping test_result_2: swap usage (0.69 GB) exceeds swap left (0.47 GB)
23:22:03 [cpu_sched_debug] skipping test_result_3: swap usage (0.69 GB) exceeds swap left (0.47 GB)
```

The `mem_usage` lines are correct, and that is the point: swap usage really is
0.00 GB. The 0.69 GB the last two jobs are refused on is `rss_smoothed` of the
running job, arriving through `max_rss`; they never start, with a CPU free the
whole time. That is what #7132 and #7146 look like from the user's side.

Tracking `APP_VERSION::max_swap_usage` next to `max_rss`, by the same rule, covers
it. When nothing is swapping it is zero, so it imposes no limit at all; swap usage
is a property of memory pressure rather than of the job.

I'm not arguing that a swap-killed job should come back sooner. What I'd like to
separate out is that `swap_limit_check()` returns true and `schedule_cpus()` then
skips the scheduler entirely for that pass:

```cpp
    if (!swap_limit_check()) {
        make_run_list(run_list);
        return enforce_run_list(run_list);
    }
    return true;
```

so the run list becomes whatever `swap_limit_check()` built, and that only ever
contains jobs which are already running or swap-killed. Jobs of a different app,
using no swap at all, are not considered. Two jobs holding 0.39 GB of real
`VmSwap` each against a 0.53 GB limit, plus two jobs of a second app that never
swap:

```
23:23:49 [cpu_sched_debug] swap usage limit exceeded: 0.78 GB > 0.53 GB
23:23:49 [cpu_sched_debug] killing test_result_0
23:23:49 [cpu_sched_debug] scheduling test_result_1
23:23:49 [cpu_sched_debug] skipping test_result_0: swap usage (0.39 GB) exceeds swap left (0.14 GB)
23:23:49 [cpu_sched_debug] using only 1.00 out of 2 CPUs
...
23:23:59 [cpu_sched_debug] checking for runnable swap-killed tasks
23:23:59 [cpu_sched_debug] using only 1.00 out of 2 CPUs
```

The killed job not coming back is the conservatism you describe, and I'll leave it
alone; its `procinfo.swap_usage` is frozen at 0.39 GB because `get_memory_usage()`
skips `PROCESS_UNINITIALIZED` tasks, so the figure can only be an over-estimate.
But `make_run_list()` is not called again either, so the second app never starts
and a CPU sits idle for the rest of the run. I don't think it needs to hold back
jobs that have nothing to do with swap.

With the limit below what one job holds, the kill loop can't stop early, so master
kills that job too:

```
23:24:58 [cpu_sched_debug] swap usage limit exceeded: 0.78 GB > 0.28 GB
23:24:58 [cpu_sched_debug] killing test_result_0
23:24:58 [cpu_sched_debug] killing test_result_1
...
23:25:08 [cpu_sched_debug] checking for runnable swap-killed tasks
23:25:08 [cpu_sched_debug] using only 0.00 out of 2 CPUs
```

After that nothing is executing, and every later pass takes the revive branch,
where each job's frozen figure (0.39 GB) exceeds the whole limit (0.28 GB) and
vetoes everything behind it; `make_run_list()` is never called again. The second
app's jobs were never considered, for the rest of the run.

There is no setting of this that is simply right: the swap usage of a job that
isn't running can't be measured, so you're guessing either way. Comparing RSS
against swap space is the wrong quantity rather than a cautious estimate of the
right one; counting a process twice is wrong at any setting; and stopping
unrelated jobs from being scheduled at all isn't caution about the killed job,
it's collateral (Missile).

`swap_limit_check()` no longer returns anything and no longer builds a run list. It
kills when we're over the limit and revives what fits when we're under it, and
scheduling goes back to `make_run_list()` and `enforce_run_list()`, which already
apply the swap test to each job's own figure. The kill loop skips jobs that hold no
swap, since killing one frees nothing and costs it its checkpoint, and it leaves
the last job that does hold swap running: if we're still over the limit by then,
that job exceeds it alone and killing it brings us nothing. Running one job over
the limit seemed better than running none. (The Android special case from the
earlier version of this branch is gone; it wasn't a proper fix for this.)

One issue I know about. A killed job whose frozen figure lands between what's left
and the whole limit still won't be revived while other jobs are using swap, even
though the dead process holds none. I only zero that figure when it exceeds the
whole limit and the process is gone, because zeroing it unconditionally is the
restart loop you're describing.

The logs come from a "harness" that runs the real client with the scheduler exactly
as built; nothing in it models scheduling. `swap_worker` calls
`madvise(MADV_PAGEOUT)`, so `VmSwap` is a known number and the runs repeat exactly.
It replaces `client/app_test.cpp` and one line of `app_start.cpp`.

https://github.com/KNineLat/boinc-scheduler_star

`./compare.sh --prebuilt` runs both clients over every scenario:

```
                                          master    patched
e1  unstarted jobs charged RSS as swap
    idle CPU passes                         1          0

e2  scheduling after swap kill (real VmSwap)
    jobs killed                             1          1
    make_run_list() after kill              1          6
    idle CPU passes                         5          5

e3  second app (no swap) alongside swap kill
    jobs killed                             1          1
    second app scheduled                    0          5
    idle CPU passes                         5          0

e4  one job exceeds limit alone
    jobs killed                             2          1
    second app scheduled                    0          9
    idle CPU passes                         9          0
```

In e2 the killed job doesn't come back in either build; what changes is that `make_run_list()` runs again, so
other jobs aren't locked out. In e3 and e4 the second app holds zero swap and starts immediately on
the patched build.

I have force-pushed the changes and would be glad to hear your thoughts.

KNineLat:
On another note, I haven't been able to test the Android app on this branch yet, but I will provide an update here once I do. This is, of course, assuming you have no further objections to the commit.

KNineLat:
It took me a while to get to this point. 
Hope the explanation isn't too long. I wanted to share my thought process. Previously, I focused too much on pointless AI reports and ended up with the same behavior as the master branch but worse.
@davidpanderson

davidpanderson:
I'm closing this PR because it combines too many things, and I'm not convinced that any of them are problems.  On the platforms I test on (Win and Linux) I'm not seeing any issues.  Swap usage is taken from the OS, not estimated from RSS.

Which is not to say there are no problems.  But we need to address them one at a time, using the standard protocol:
- create an issue describing the problem and a repro scenario.  Include logs (with mem_usage_debug set) that show the problem.
- discussion (in the issue)
- someone makes a PR

PR #7195, opened 2026-07-28
Android: link 64-bit client binaries with 16 KB page alignment

### LINKED ISSUE #5737: [Android] Set correct page size when building native binaries (might be a hard requirement for NDK r28)
Ref: https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md#Page-sizes

### LINKED ISSUE #7194: Android: client binaries are 4 KB aligned and fail to load on 16 KB page-size devices
### Describe the bug

While emulator-testing #7193 I hit this: on Android system images with a 16 KB page size (the google_apis_ps16k images), the BOINC app UI runs but the native client never starts. The Monitor logs `getPidForProcessName(): /data/user/0/edu.berkeley.boinc/client/boinc not found in ps output!` and retries `connect failure: IO` forever, and the app sits on the splash screen. The same APK works fine on a standard 4 KB image (Android 14, API 34).

The cause is ELF alignment. The client binaries shipped in the app assets are linked with 4 KB load segment alignment: I checked the program headers of the current CI-built binaries and both x86_64 and arm64-v8a have PT_LOAD alignment 0x1000. A 16 KB kernel refuses to load such executables, which from the app's side looks like the client silently never starting.

This matters beyond emulators. Google requires 16 KB page-size support on Play for newer target SDKs, and new arm64 devices are shipping 16 KB kernels. On those devices BOINC will install and show its UI, but it can never compute.

### Steps to reproduce

1. Create an emulator from a ps16k system image (for example google_apis_ps16k, x86_64)
2. Install any current BOINC Android build and launch it
3. The app stays on the splash screen; logcat shows the Monitor connect retries and the client process never appears in ps

### Expected behavior

The native client starts on 16 KB page-size devices the same way it does on 4 KB ones.

### Proposed fix

Link the Android client binaries with a 16 KB max page size by adding `-Wl,-z,max-page-size=16384` to the NDK link flags in the Android build scripts. Binaries aligned to 16 KB still load fine on 4 KB kernels, so this is backward compatible. The Android apps and wrapper builds will eventually want the same flag.

I have a fix in progress and will open a PR once I've rebuilt the client and verified both the alignment and the behavior on a ps16k image.

pmclSF:
Fixes #7194

**Description of the Change**

Adds `-Wl,-z,max-page-size=16384` to the link flags of the arm64 and x86_64 client builds, so the client binaries load on Android devices with 16 KB page-size kernels. Today they are linked with 4 KB ELF segment alignment (PT_LOAD align 0x1000 in the current CI binaries), which a 16 KB kernel refuses to load: the app UI runs but the native client silently never starts. Details and reproduction are in #7194.

16 KB aligned binaries remain fully compatible with 4 KB kernels, so nothing changes for existing devices. The 32-bit ABIs are unchanged since 16 KB page-size kernels only exist for 64-bit.

**Verification (all done)**

- [x] The CI artifacts from this PR's own Android run have PT_LOAD alignment 0x4000: verified by parsing the ELF program headers of the x86_64 and arm64-v8a client binaries from the run artifacts.
- [x] Runtime test on a 16 KB kernel (Android 15 google_apis_ps16k x86_64 emulator, getconf PAGE_SIZE confirms 16384): the current 4 KB aligned client reproducibly fails to start there (the Monitor loops on "not found in ps output"), and the same app repacked with this PR's CI-built client starts normally, the UI connects, and it keeps running with no errors.
- [x] Control on a standard 4 KB Android 14 (API 34) x86_64 emulator: the 16 KB aligned client starts and runs exactly like before, so no regression on existing devices.

**Alternate Designs**

Building separate 16 KB variants of the client was not considered seriously since a single 16 K aligned binary works on both page sizes. Adding the flag to the 32-bit ABIs too was considered and skipped as meaningless there, though it would be harmless.

**Release Notes**

Android: BOINC now works on devices with 16 KB memory page kernels.

### DIFF
--- a/android/build_boinc_arm64.sh
+++ b/android/build_boinc_arm64.sh
@@ -39,7 +39,7 @@ export CXX=aarch64-linux-android21-clang++
 export LD=aarch64-linux-android-ld
 export CFLAGS="--sysroot=$TCSYSROOT -DANDROID -DANDROID_64 -DDECLARE_TIMEZONE -Wall -I$TCINCLUDES/include -O3 -fomit-frame-pointer -fPIE -D__ANDROID_API__=21"
 export CXXFLAGS="--sysroot=$TCSYSROOT -DANDROID -DANDROID_64 -Wall -I$TCINCLUDES/include -funroll-loops -fexceptions -O3 -fomit-frame-pointer -fPIE -D__ANDROID_API__=21"
-export LDFLAGS="$CONFIG_LDFLAGS -llog -fPIE -pie -latomic -static-libstdc++"
+export LDFLAGS="$CONFIG_LDFLAGS -llog -fPIE -pie -latomic -static-libstdc++ -Wl,-z,max-page-size=16384"
 export GDB_CFLAGS="--sysroot=$TCSYSROOT -Wall -g -I$TCINCLUDES/include"
 
 MAKE_FLAGS=""
--- a/android/build_boinc_x86_64.sh
+++ b/android/build_boinc_x86_64.sh
@@ -39,7 +39,7 @@ export CXX=x86_64-linux-android21-clang++
 export LD=x86_64-linux-android-ld
 export CFLAGS="--sysroot=$TCSYSROOT -DANDROID -DANDROID_64 -DDECLARE_TIMEZONE -Wall -I$TCINCLUDES/include -O3 -fomit-frame-pointer -fPIE -D__ANDROID_API__=21"
 export CXXFLAGS="--sysroot=$TCSYSROOT -DANDROID -DANDROID_64 -Wall -I$TCINCLUDES/include -funroll-loops -fexceptions -O3 -fomit-frame-pointer -fPIE -D__ANDROID_API__=21"
-export LDFLAGS="$CONFIG_LDFLAGS -llog -fPIE -pie -latomic -static-libstdc++"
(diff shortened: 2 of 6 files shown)


AenBleidd:
@pmclSF, I allowed this once, but I kindly ask you to follow the AI assistant usage guide, in particular: AI cannot be a co-author of the PR, it might only assist a developer, and all the respobsibility lays on the developer, not an AI tool.
Please fix the commit message.
Thank you in advance.

AenBleidd (changes requested):
Please fix the commit message

pmclSF:
Thanks for the pointer. Fixed: the commit message now uses the `Assisted-by:` trailer per the AI assistants usage policy instead of a co-author line.

AenBleidd (approved):

AenBleidd:
Besides client (the changed scripts), we also build examples and libraries (see the corresponding scripts in the same directory).
They need to be changed too.
Thank you in advance.

pmclSF:
Useful feedback! The 64-bit example and libraries scripts now get the same flag, so all six scripts (client, example, libraries for arm64 and x86_64) link with 16 KB alignment.

AenBleidd (approved):

AenBleidd:
Looks good to me.
@pmclSF, thank you for your contribution!

PR #7198, opened 2026-07-31
client: fixes to NVIDIA GPU detection

### LINKED ISSUE #7196: Client: NVIDIA detection uses the deprecated cuDeviceComputeCapability API and aborts if it's missing
### Describe the bug

The client reads GPU compute capability through cuDeviceComputeCapability (client/gpu_nvidia.cpp:419), which Nvidia deprecated years ago. Two problems with that:

- If the symbol is ever missing from the driver library, detection doesn't degrade, it aborts. The warning path at client/gpu_nvidia.cpp:352 gives up on all Nvidia GPUs, so the day Nvidia drops the deprecated symbol, clients lose GPU computing entirely instead of just falling back.
- The call's return value isn't checked, so a failed read silently leaves whatever was in the struct.

The modern replacement is cuDeviceGetAttribute with CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR/MINOR, and the client already binds cuDeviceGetAttribute for other attributes.

### Steps to reproduce

Not reproducible with current drivers (they still export the symbol). The abort path is at client/gpu_nvidia.cpp:352: if the lookup fails, detection jumps to the exit instead of continuing without the deprecated call.

### Expected behavior

Compute capability is read via cuDeviceGetAttribute, the deprecated call stays only as a fallback for old drivers, and a missing legacy symbol no longer disables all GPU detection.

### LINKED ISSUE #7197: Client: hosts with OpenCL-only NVIDIA detection report compute capability 0 to the scheduler
### Describe the bug

When CUDA detection fails but OpenCL succeeds, correlate_opencl() backfills memory, clock and name from the OpenCL properties but not cuda_prop.major/minor (client/gpu_opencl.cpp:817). The scheduler request then carries `<major>0</major>`, so every server side compute capability check (min_nvidia_compcap and max_nvidia_compcap in plan_class_spec.cpp, the legacy cuda_check) sees 0 for these hosts.

The OpenCL reported values are already detected (nv_compute_capability_major/minor via the cl_nv_device_attribute_query extension) and are even used for peak flops estimation in COPROC_NVIDIA::set_peak_flops() (lib/coproc.cpp:564). They just never get copied into the fields that are serialized as `<major>/<minor>`.

### Expected behavior

On OpenCL-only hosts, cuda_prop.major/minor are backfilled from the OpenCL compute capability values, with CUDA values winning whenever present, matching the precedence set_peak_flops() already uses.

davidpanderson:
1) The cuDeviceComputeCapability() function in the CUDA library was deprecated some time ago (2019?) and may be removed in the future. Currently CUDA GPU detection fails if this function is absent. Instead, use cuDeviceGetAttribute() instead.

Fixes #7196

2) If we find an NVIDIA GPU via OpenCL and not CUDA, set its compute capability in the CUDA_DEVICE_PROP; that's what is used by the server plan-class logic.

Fixes #7197

---
## Summary by cubic
Fixes NVIDIA GPU detection to work without the deprecated CUDA API and for OpenCL-only NVIDIA GPUs. Adds error checks and removes unused `#ifdef` code to make detection robust and ensure correct plan-class selection.

- **Bug Fixes**
  - CUDA: Fallback to `cuDeviceGetAttribute()` (compute capability major/minor) when `cuDeviceComputeCapability()` is missing; add error checks and warnings for CUDA attribute/memory queries.
  - OpenCL-only NVIDIA: Populate `CUDA_DEVICE_PROP` (compute capability, memory, clock rate, name) from OpenCL so scheduler logic works.

- **Refactors**
  - Removed unused `#ifdef` code paths in GPU detection.

<sup>Written for commit 8acdd541104db4d7a5a2a5698797a8ad4bc6b956. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/BOINC/boinc/pull/7198?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/client/gpu_nvidia.cpp
+++ b/client/gpu_nvidia.cpp
@@ -1,6 +1,6 @@
 // This file is part of BOINC.
-// http://boinc.berkeley.edu
-// Copyright (C) 2012 University of California
+// https://boinc.berkeley.edu
+// Copyright (C) 2026 University of California
 //
 // BOINC is free software; you can redistribute it and/or modify it
 // under the terms of the GNU Lesser General Public License
@@ -19,10 +19,10 @@
 
 #ifdef _WIN32
 #include "boinc_win.h"
-/* get annotation macros from sal.h */
-/* define the ones that don't exist */
+// get annotation macros from sal.h
+// define the ones that don't exist
 #include "sal.h"
-/* These are just an annotations.  They don't do anything */
+// These are just an annotations.  They don't do anything
 #ifndef __success
 #define __success(x)
 #endif
@@ -165,6 +165,8 @@ int nvidia_compare(COPROC_NVIDIA& c1, COPROC_NVIDIA& c2, bool loose) {
     return 0;
 }
 
+// NOTE: WE SHOULD BE GETTING THESE FROM AN NVIDIA .h FILE!
+//
 enum CUdevice_attribute_enum {
     CU_DEVICE_ATTRIBUTE_MAX_THREADS_PER_BLOCK = 1,
     CU_DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_X = 2,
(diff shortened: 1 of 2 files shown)


AenBleidd (approved):

client/gpu_nvidia.cpp
```diff
@@ -349,10 +355,12 @@ void* cudalib = NULL;
         gpu_warning(warnings, "cuDeviceTotalMem() missing from NVIDIA library");
         goto leave;
     }
+#if 0
```
AenBleidd: @davidpanderson, since we are checking this function later in this file, I suppose we can drop this block completely to keep the code cleaner.

AenBleidd (changes requested):

AenBleidd (approved):

PR #7214, opened 2026-08-12
Update ProjectProcessingPage.cpp

### LINKED ISSUE #6104: [Manager] [Linux] When using wizards, clicking on back button causes unexpected pages
### Describe the bug

If I build BOINC Manager from the Master and go to the "Add project..." or "Use account manager...", the back button behaves unexpectedly.  It seems to cycle through different pages, related to the project/AM wizard or not.  For example, I was getting Proxy Configuration after clicking the back button a couple of times.

### Steps to reproduce

1. Build Master.
2. Run BOINC Manager.
3. Click on "Add project..."
4. Select a project (any project), click next.
5. Click the back button.  Sometimes the wizard stays on the same page, so click it again and you should see something that wasn't the previous page.
6. Continue to click back, it cycles through various pages.  Back never gets disabled.

### Expected behavior

The back button to behave normally.

### Screenshots

Image

### System information

- OS:  Ubuntu 24.04.2
- BOINC:  Master built from [ecc3330c0d1f3a8b79e90133b6e6b87f1cd63a3d](https://github.com/BOINC/boinc/tree/ecc3330c0d1f3a8b79e90133b6e6b87f1cd63a3d)
- wxWidgets 3.2.4

### Additional context

This does not happen in the official 8.0.2 release for Linux, nor was I able to reproduce this in Windows (either building from Master or from the official release).

I wonder if the bug is related to this compile warning:

```
ProjectProcessingPage.h:47:100: warning: cast between incompatible pointer to member types from ‘void (CProjectProcessingPage::*)(CProjectProcessingPageEvent&)’ to ‘wxEventFunction’ {aka ‘void (wxEvtHandler::*)(wxEvent&)’} [-Wcast-function-type]
   47 |     DECLARE_EVENT_TABLE_ENTRY(wxEVT_PROJECTPROCESSING_STATECHANGE, -1, -1, (wxObjectEventFunction) (wxEventFunction) &fn, NULL),
/usr/include/wx-3.2/wx/event.h:87:73: note: in definition of macro ‘wxDECLARE_EVENT_TABLE_ENTRY’
   87 |     wxEventTableEntry(type, winid, idLast, wxNewEventTableFunctor(type, fn), obj)
      |                                                                         ^~
ProjectProcessingPage.h:47:5: note: in expansion of macro ‘DECLARE_EVENT_TABLE_ENTRY’
   47 |     DECLARE_EVENT_TABLE_ENTRY(wxEVT_PROJECTPROCESSING_STATECHANGE, -1, -1, (wxObjectEventFunction) (wxEventFunction) &fn, NULL),
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~
ProjectProcessingPage.cpp:74:5: note: in expansion of macro ‘EVT_PROJECTPROCESSING_STATECHANGE’
   74 |     EVT_PROJECTPROCESSING_STATECHANGE( CProjectProcessingPage::OnStateChange )
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

There are several more like this.  Seeing an incompatible pointer makes me think that may be why different pages show up.

LezheGao:
Fixes #6104 

**Description of the Change**
The root cause of was an unsafe cast of a member function pointer in the static event table:
```cpp
EVT_PROJECTPROCESSING_STATECHANGE( CProjectProcessingPage::OnStateChange )
```
This macro forced a cast from `void (CProjectProcessingPage::*)(CProjectProcessingPageEvent&)` to `wxEventFunction` (i.e., `void (wxEvtHandler::*)(wxEvent&)`), which is a cast between incompatible pointer types. Under GCC with `-Wcast-function-type`, this results in undefined behavior – the `this` pointer is misadjusted when the event handler is called, causing the state machine (`m_iCurrentState`, flags, and button enable/disable logic) to read/write garbage memory. Consequently, the wizard’s internal page history becomes corrupted, leading to the looping back behaviour and the back button never being greyed out.
To fix this, the static event table entry for `wxEVT_PROJECTPROCESSING_STATECHANGE` has been removed and replaced with a type‑safe dynamic binding using `Bind()` in the `Create()` method. `Bind()` uses templates to correctly adjust the `this` pointer, eliminating the UB and the compiler warning. The other static entries (`EVT_WIZARDEX_PAGE_CHANGED` and `EVT_WIZARDEX_CANCEL`) are unchanged because they use compatible function signatures.
After this change, the wizard’s state transitions work correctly, the back button follows the expected page history, and it becomes disabled when appropriate.

Assisted-by: (not using an agent): DeepSeek-V4

**Alternate Designs**
Keeping the static event table but modify the macro to use `wxDECLARE_EVENT_TABLE_ENTRY` with a proper functor that performs a safe cast. This would require changes in the header file and deeper understanding of wxWidgets’ internal event mechanism, making it more error‑prone and less maintainable.

**Release Notes**
Fixed erratic "Back" button behaviour in the project attach wizard (Linux builds from master).

---
## Summary by cubic
Fixes #6104 by replacing an unsafe static event handler with a type-safe `Bind()` for the state-change event. The project attach wizard now advances correctly and disables the Back button when expected.

- **Bug Fixes**
  - Replaced static `EVT_PROJECTPROCESSING_STATECHANGE` entry with `Bind(wxEVT_PROJECTPROCESSING_STATECHANGE, &CProjectProcessingPage::OnStateChange, this)` in `Create()`.
  - Left `EVT_WIZARDEX_PAGE_CHANGED` and `EVT_WIZARDEX_CANCEL` unchanged.

<a href="https://cubic.dev/pr/BOINC/boinc/pull/7214?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clientgui/ProjectProcessingPage.cpp
+++ b/clientgui/ProjectProcessingPage.cpp
@@ -67,8 +67,6 @@ IMPLEMENT_DYNAMIC_CLASS( CProjectProcessingPage, wxWizardPageEx )
 
 /*!
  * CProjectProcessingPage event table definition
- * NOTE: The EVT_PROJECTPROCESSING_STATECHANGE entry has been removed
- * and replaced with a Bind() call in Create() to avoid unsafe pointer casts.
  */
 BEGIN_EVENT_TABLE( CProjectProcessingPage, wxWizardPageEx )
 
@@ -119,8 +117,6 @@ bool CProjectProcessingPage::Create( CBOINCBaseWizard* parent )
 ////@end CProjectProcessingPage creation
 
     // Bind the custom state-change event to its handler.
-    // This is type-safe and avoids the compiler warning about cast between
-    // incompatible member function pointer types.
     Bind(wxEVT_PROJECTPROCESSING_STATECHANGE,
          &CProjectProcessingPage::OnStateChange, this);
 
--- a/clientgui/ProjectProcessingPage.cpp
+++ b/clientgui/ProjectProcessingPage.cpp
@@ -67,16 +67,14 @@ IMPLEMENT_DYNAMIC_CLASS( CProjectProcessingPage, wxWizardPageEx )
 
 /*!
  * CProjectProcessingPage event table definition
+ * NOTE: The EVT_PROJECTPROCESSING_STATECHANGE entry has been removed
+ * and replaced with a Bind() call in Create() to avoid unsafe pointer casts.
  */
-
 BEGIN_EVENT_TABLE( CProjectProcessingPage, wxWizardPageEx )
-
-    EVT_PROJECTPROCESSING_STATECHANGE( CProjectProcessingPage::OnStateChange )
 
 ////@begin CProjectProcessingPage event table entries
     EVT_WIZARDEX_PAGE_CHANGED( -1, CProjectProcessingPage::OnPageChanged )
     EVT_WIZARDEX_CANCEL( -1, CProjectProcessingPage::OnCancel )
-
 ////@end CProjectProcessingPage event table entries
 
 END_EVENT_TABLE()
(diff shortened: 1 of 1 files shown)


AenBleidd:
#7262 contains a proper fix for the issue mentioned.

PR #7216, opened 2026-08-12
Update sg_BoincSimpleFrame.cpp

### LINKED ISSUE #6103: [Manager] [Simple View] Unable to permanently revert back to default skin
### Describe the bug

If the user had selected a non-default skin in the past, they cannot switch back to the default skin permanently.  When Manager is closed and opened up again, the skin that was previously chosen would be the last one used.

### Steps to reproduce

Example 1:
1. Change to simple view.  It is assumed the skin is default
2. Change the skin to a non-default skin.
3. Close Manager.
4. Open Manager again.  The skin that was last selected should be displayed.
5. Change the skin to default and close Manager.
6.  Open the Manager, and the previous (non-default) skin should display.

Example 2:
1. Repeat steps 1-4 above.
2. Change to another non-default skin.  For example, go from CharityEngine to Progress through Processors.  Do NOT close Manger.
3. After choosing PTP skin,  switch to the default skin.
4. Close Manager.
5. Open the Manager, and the PTP skin will be displayed.

### Expected behavior

When the default skin is selected, it should stay applied after the Manager is closed.

### Screenshots

_No response_

### System information

- OS:  Ubuntu 24.02.2
- BOINC:  8.0.2
- wxWidgets 3.2.5

### Additional context

I was not able to reproduce this with a local build on Windows 10.

LezheGao:
Fixes #6103

**Description of the Change**
When a user selects the "Default" skin via the View → Skin menu, the GUI reloads the default appearance, but the selection is never saved to the configuration file. On next startup, `RestoreState()` reads the previously saved (non‑default) skin name and re‑applies that skin, effectively discarding the user's choice.

The fix adds the same persistence logic already present in `OnSelectSkin()` for non‑default skins to `OnSelectDefaultSkin()`:
- `wxGetApp().SaveState()` – saves the current skin name via `CSkinManager`
- `wxConfigBase::Get(FALSE)->Flush()` – ensures the config is written to disk immediately

This makes the behaviour consistent across all skin selections and ensures the default skin is remembered across sessions.

Assisted-by: (not using an agent): DeepSeek-V4

**Alternate Designs**
- A more extensive refactoring could unify the persistence logic into a helper method (e.g., `SaveSelectedSkin()`), but that would touch more files and add unnecessary churn for a two‑line fix.

The chosen approach is minimal, safe, and mirrors the existing working implementation for non‑default skins, reducing the risk of regression.

**Release Notes**
Fixed an issue where selecting the default skin did not persist after restarting BOINC Manager.

---
## Summary by cubic
Persists selecting the “Default” skin so BOINC Manager restores it after restart; previously it reapplied the last saved non‑default skin. Also guards config flushing to avoid a null dereference. Fixes #6103.

**Changes**
- In `OnSelectDefaultSkin()`: explicitly call `SetSelectedSkin()` with the default name, reload the skin, then call `wxGetApp().SaveState()` and `wxConfigBase::Get(FALSE)->Flush()` only if a config exists.
- In `OnSelectSkin()`: guard the `Flush()` call behind a null check on `wxConfigBase::Get(FALSE)`.

<sup>Written for commit 74514e784d5e22854e005beddb565579bdf8e778. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/BOINC/boinc/pull/7216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clientgui/sg_BoincSimpleFrame.cpp
+++ b/clientgui/sg_BoincSimpleFrame.cpp
@@ -586,7 +586,10 @@ void CSimpleFrame::OnSelectDefaultSkin( wxCommandEvent& WXUNUSED(event) ) {
     pSkinManager->ReloadSkin(pSkinManager->GetDefaultSkinName());
 
     wxGetApp().SaveState();
-    wxConfigBase::Get(FALSE)->Flush();
+    wxConfigBase* pConfig = wxConfigBase::Get(FALSE);
+    if (pConfig) {
+        pConfig->Flush();
+    }
 }
 
 
--- a/clientgui/sg_BoincSimpleFrame.cpp
+++ b/clientgui/sg_BoincSimpleFrame.cpp
@@ -584,6 +584,9 @@ void CSimpleFrame::OnSelectDefaultSkin( wxCommandEvent& WXUNUSED(event) ) {
     // The "Default" skin menu item is localized, but
     // the name of the default skin is not localized
     pSkinManager->ReloadSkin(pSkinManager->GetDefaultSkinName());
+
+    wxGetApp().SaveState();
+    wxConfigBase::Get(FALSE)->Flush();
 }
 
 
--- a/clientgui/sg_BoincSimpleFrame.cpp
+++ b/clientgui/sg_BoincSimpleFrame.cpp
@@ -581,8 +581,8 @@ void CSimpleFrame::OnSelectDefaultSkin( wxCommandEvent& WXUNUSED(event) ) {
     wxASSERT(pSkinManager);
     wxASSERT(wxDynamicCast(pSkinManager, CSkinManager));
 
-    // The "Default" skin menu item is localized, but
-    // the name of the default skin is not localized
+    pSkinManager->SetSelectedSkin(pSkinManager->GetDefaultSkinName());
+
     pSkinManager->ReloadSkin(pSkinManager->GetDefaultSkinName());
 
     wxGetApp().SaveState();
--- a/clientgui/sg_BoincSimpleFrame.cpp
+++ b/clientgui/sg_BoincSimpleFrame.cpp
@@ -627,7 +627,10 @@ void CSimpleFrame::OnSelectSkin( wxCommandEvent& event ){
     pSkinManager->ReloadSkin(newSkinName);
 
     wxGetApp().SaveState();
-    wxConfigBase::Get(FALSE)->Flush();
+    wxConfigBase* pConfig = wxConfigBase::Get(FALSE);
+    if (pConfig) {
+        pConfig->Flush();
+    }
 }
 
 


CharlieFenton:
This is unnecessary. The change to default skin is remembered on exit from BOINC Manager and restored when BOINC Manager is relaunched. It works correctly in BOINC 8.2.15 on my Mac.

The default skin is already saved when CBOINCGUIApp::OnExit() calls CBOINCGUIApp::SaveState():
```
    if (m_pSkinManager) {
        m_pConfig->Write(wxT("Skin"), m_pSkinManager->GetSelectedSkin());
    }

```

AenBleidd:
@CharlieFenton, however, we still have the issue desribed.
I believe this is just the fix that is incorrect.
@LezheGao, it looks like you are using some kind of AI.
Have you followed our [AI policy](https://github.com/BOINC/boinc/wiki/Contributing-to-BOINC-with-AI-assistants)?

CharlieFenton:
> we still have the issue described.

I don't think that issue exists. I cannot reproduce this issue on the Mac, though I have no way to test it on Linux. It is unclear in the description of #6103 whether by "Close Manager" and "Open Manager again" he means closing and reopening the window or exiting the Manager and relaunching it. But I tested both interpretations following his steps and the default skin was preserved.

CharlieFenton:
In issue #6103 @Vulpine05 wrote:
> I was not able to reproduce this with a local build on Windows 10.

If the issue exists on Linux but not on Mac or Windows, it seems to imply that CBOINCGUIApp::SaveState() is not being called when the user exits the Manager (assuming he means exiting the Manager and relaunching it.) If so, that is a much bigger issue.

But if he means closing and reopening the Manager window, then calling CBOINCGUIApp::SaveState() then calling CBOINCGUIApp::SaveState() won't fix it because the value is read only in CBOINCGUIApp::OnInit() when the Manager is first launched.

LezheGao:
> Have you followed our [AI policy](https://github.com/BOINC/boinc/wiki/Contributing-to-BOINC-with-AI-assistants)?

Fixed

davidpanderson:
I'm not sure this applies here, but:
config changes should be changed when the user makes them,
not when the manager exits.

LezheGao:
> @CharlieFenton, however, we still have the issue desribed. I believe this is just the fix that is incorrect.

Hopefully [74514e7](https://github.com/BOINC/boinc/pull/7216/commits/74514e784d5e22854e005beddb565579bdf8e778) will work.

CharlieFenton:
@LezheGao There should be no harm in making the change in [74514e7](https://github.com/BOINC/boinc/pull/7216/commits/74514e784d5e22854e005beddb565579bdf8e778), and seems like the right thing to do based on [this comment](https://github.com/BOINC/boinc/pull/7216#issuecomment-5268208017).

And as I wrote [here](https://github.com/BOINC/boinc/issues/6103#issuecomment-5391490605), it would be good to know why the code in _CBOINCGUIApp::SaveState()_ is apparently not being called when BOINC Manager exits on Linux. It should be called from _CBOINCGUIApp::OnExit()_.

PR #7255, opened 2026-08-26
Increase buffers in case of long oses and machines

LocutusOfBorg:
This is an old Debian patch, I guess there were some bugs related to old os versions being more than 256 chars

### DIFF
--- a/html/ops/db_update.php
+++ b/html/ops/db_update.php
@@ -1332,6 +1332,15 @@ function update_8_23_2026() {
     ");
 }
 
+function update_8_26_2026() {
+    do_query("alter table host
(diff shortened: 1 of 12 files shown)


AenBleidd:
@davidpanderson, please review this.

html/ops/db_update.php
```diff
@@ -1332,6 +1332,15 @@ function update_8_23_2026() {
     ");
 }
 
+function update_8_2_2026() {
```
AenBleidd: This function has the wrong name, should be 'update_8_26_2026'

davidpanderson:
I don't know of an OS name or version that's 80 chars, much less 256.
So this isn't needed.

PR #7256, opened 2026-08-26
Increase some buffers based on MAXPATHLEN

LocutusOfBorg:
Better use MAXPATHLEN instead of hardcoding 1024

### DIFF
--- a/client/file_names.cpp
+++ b/client/file_names.cpp
@@ -76,7 +76,7 @@ void get_pathname(FILE_INFO* fip, char* path, int len) {
         if (fip->is_auto_update_file) {
             boinc_version_dir(*p, gstate.auto_update.version, buf);
         } else {
-            strcpy(buf, p->project_dir());
+            safe_strcpy(buf, p->project_dir());
         }
 #else
         safe_strcpy(buf, p->project_dir());
@@ -88,28 +88,28 @@ void get_pathname(FILE_INFO* fip, char* path, int len) {
 }
 
 void get_sched_request_filename(PROJECT& project, char* buf, int len) {
-    char url[1024];
+    char url[MAXPATHLEN];
 
     escape_project_url(project.master_url, url);
     snprintf(buf, len, "%s%s.xml", SCHED_OP_REQUEST_BASE, url);
 }
 
 void get_sched_reply_filename(PROJECT& project, char* buf, int len) {
-    char url[1024];
+    char url[MAXPATHLEN];
 
     escape_project_url(project.master_url, url);
     snprintf(buf, len, "%s%s.xml", SCHED_OP_REPLY_BASE, url);
 }
 
 void get_master_filename(PROJECT& project, char* buf, int len) {
-    char url[1024];
+    char url[MAXPATHLEN];
 
     escape_project_url(project.master_url, url);
     snprintf(buf, len, "%s%s.xml", MASTER_BASE, url);
 }
 
 void job_log_filename(PROJECT& project, char* buf, int len) {
-    char url[1024];
+    char url[MAXPATHLEN];
 
     escape_project_url(project.master_url, url);
     snprintf(buf, len, "%s%s.txt", JOB_LOG_BASE, url);
@@ -217,7 +217,7 @@ int make_slot_dir(int slot) {
 // delete unused stuff in the slots/ directory
 //
 void delete_old_slot_dirs() {
-    char filename[1024], path[MAXPATHLEN];
+    char filename[MAXPATHLEN], path[MAXPATHLEN];
     DIRREF dirp;
     int retval;
 


client/file_names.cpp
```diff
@@ -66,53 +67,53 @@ int make_soft_link(PROJECT* project, char* link_path, char* rel_file_path) {
 //
 void get_pathname(FILE_INFO* fip, char* path, int len) {
     PROJECT* p = fip->project;
-    char buf[MAXPATHLEN];
+    std::vector<char> buf(std::max(len, MAXPATHLEN));
```
AenBleidd: This is a local buffer, so why to use std::vector here?
The same is applicable to the places below

LocutusOfBorg: using char buf[std::max(len, MAXPATHLEN))]; was failing on windows pipeline...

LocutusOfBorg: I'm not sure how to better fix this without having to use std::vector

AenBleidd: But what is the issue you are trying to fix? I don't see any, and having the original code is safe enough.

AenBleidd: I took a deeper look at this function: `MAXPATHLEN` here is fine enough, and changing it to any other value will not actually take any effect, since `boinc_version_dir()` function doesn't check the size of the input buffer,
`safe_strcpy` checks the size, and the `len` parameters down below affects data copy **from** buf, and thus this change does basically nothing.

client/file_names.cpp
```diff
@@ -66,53 +67,53 @@ int make_soft_link(PROJECT* project, char* link_path, char* rel_file_path) {
...
     // an associated PROJECT.
     //
     if (p) {
 #ifdef ENABLE_AUTO_UPDATE
         if (fip->is_auto_update_file) {
-            boinc_version_dir(*p, gstate.auto_update.version, buf);
+            boinc_version_dir(*p, gstate.auto_update.version, buf.data());
         } else {
-            strcpy(buf, p->project_dir());
+            strcpy(buf.data(), p->project_dir());
```
AenBleidd: `strcpy` should be replaced with `safe_strcpy` here

client/file_names.cpp
```diff
@@ -66,53 +67,53 @@ int make_soft_link(PROJECT* project, char* link_path, char* rel_file_path) {
...
         }
 #else
-        safe_strcpy(buf, p->project_dir());
+        safe_strcpy(buf.data(), p->project_dir());
```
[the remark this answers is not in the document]
AenBleidd: This review comment is completely valid: passing `buf.data()` will break the functionality and will allow copying of the 7 bytes only.

client/file_names.cpp
```diff
@@ -66,53 +67,53 @@ int make_soft_link(PROJECT* project, char* link_path, char* rel_file_path) {
...
-        snprintf(path, len, "%s/%s", buf, fip->name);
+        snprintf(path, len, "%s/%s", buf.data(), fip->name);
     } else {
         strlcpy(path, fip->name, len);
     }
 }
 
 void get_sched_request_filename(PROJECT& project, char* buf, int len) {
-    char url[1024];
+    std::vector<char> url(std::max(len, MAXPATHLEN));
```
AenBleidd: Logically we don't need to change anything here: I don't believe there will be a valid URL address longer than 1023 symbols.

client/file_names.cpp
```diff
@@ -66,53 +67,53 @@ int make_soft_link(PROJECT* project, char* link_path, char* rel_file_path) {
...
 
-    escape_project_url(project.master_url, url);
-    snprintf(buf, len, "%s%s.xml", SCHED_OP_REQUEST_BASE, url);
+    escape_project_url(project.master_url, url.data());
+    snprintf(buf, len, "%s%s.xml", SCHED_OP_REQUEST_BASE, url.data());
 }
 
 void get_sched_reply_filename(PROJECT& project, char* buf, int len) {
-    char url[1024];
+    std::vector<char> url(std::max(len, MAXPATHLEN));
```
AenBleidd: The same applies here

client/file_names.cpp
```diff
@@ -66,53 +67,53 @@ int make_soft_link(PROJECT* project, char* link_path, char* rel_file_path) {
...
 
-    escape_project_url(project.master_url, url);
-    snprintf(buf, len, "%s%s.xml", SCHED_OP_REPLY_BASE, url);
+    escape_project_url(project.master_url, url.data());
+    snprintf(buf, len, "%s%s.xml", SCHED_OP_REPLY_BASE, url.data());
 }
 
 void get_master_filename(PROJECT& project, char* buf, int len) {
-    char url[1024];
+    std::vector<char> url(std::max(len, MAXPATHLEN));
```
AenBleidd: ... and here

client/file_names.cpp
```diff
@@ -66,53 +67,53 @@ int make_soft_link(PROJECT* project, char* link_path, char* rel_file_path) {
...
 
-    escape_project_url(project.master_url, url);
-    snprintf(buf, len, "%s%s.xml", MASTER_BASE, url);
+    escape_project_url(project.master_url, url.data());
+    snprintf(buf, len, "%s%s.xml", MASTER_BASE, url.data());
 }
 
 void job_log_filename(PROJECT& project, char* buf, int len) {
-    char url[1024];
+    std::vector<char> url(std::max(len, MAXPATHLEN));
```
AenBleidd: ... and here

client/file_names.cpp
```diff
@@ -217,7 +218,7 @@ int make_slot_dir(int slot) {
 // delete unused stuff in the slots/ directory
 //
 void delete_old_slot_dirs() {
-    char filename[1024], path[MAXPATHLEN];
+    char filename[MAXPATHLEN], path[MAXPATHLEN];
```
AenBleidd, line 220: I hardly believe there will be a filename that long, but this is a fine change.

AenBleidd (changes requested):

LocutusOfBorg:
I think I addressed the changes, for the url, I hope MAXPATHLEN is a better lenght, what do you think?

AenBleidd:
@davidpanderson, please review.
It looks ok to me though.

AenBleidd (approved):

davidpanderson:
MAXPATHLEN is for paths, not URLs.

We could have a different constant for max URL length,
but this is not needed because project URLs will never exceed 256 chars

LocutusOfBorg:
@AenBleidd would you mind committing at least the change
            safe_strcpy(buf, p->project_dir()); ?

AenBleidd:
@LocutusOfBorg, please open a separate PR. Thank you.

PR #7258, opened 2026-08-26
Give a better error message in case of open failure

LocutusOfBorg:
This helps debugging failures


AenBleidd:

AenBleidd (approved):

davidpanderson:
Functions at this level should not print anything; they return error codes and the caller can print something if it wants

PR #7285, opened 2026-09-06
Add missing Exception.php require in send_email()

### LINKED ISSUE #7277: send_email() crashes with an uncaught fatal error on any real SMTP failure (missing PHPMailer\PHPMailer\Exception require)
\`html/inc/email.inc\`'s \`send_email()\` requires \`PHPMailer.php\` and \`SMTP.php\`, but never \`Exception.php\`, even though all three ship together in the vendored PHPMailer library:

\`\`\`php
if (file_exists("../inc/PHPMailer/src/PHPMailer.php") && file_exists("../inc/PHPMailer/src/SMTP.php")) {
    require_once("../inc/PHPMailer/src/PHPMailer.php");
    require_once("../inc/PHPMailer/src/SMTP.php");
}
\`\`\`

This is invisible on the happy path — \`PHPMailer\PHPMailer\Exception\` is only touched when \`smtpConnect()\`/\`smtpSend()\`/\`postSend()\` hit a real connection or auth failure and try to \`throw new Exception(...)\`. Since that class was never loaded, PHP can't construct it, so instead of a catchable exception that PHPMailer's own code turns into a clean \`$mail->Send()\` returning \`false\`, you get an uncaught fatal \`Error: Class "PHPMailer\PHPMailer\Exception" not found\`, crashing the entire request with a raw stack trace on any project running with \`display_errors\` on.

Reproduced live by pointing a test project's SMTP credentials at an invalid password.

**Fix:**
\`\`\`php
if (file_exists("../inc/PHPMailer/src/PHPMailer.php") && file_exists("../inc/PHPMailer/src/SMTP.php") && file_exists("../inc/PHPMailer/src/Exception.php")) {
    require_once("../inc/PHPMailer/src/Exception.php");
    require_once("../inc/PHPMailer/src/PHPMailer.php");
    require_once("../inc/PHPMailer/src/SMTP.php");
}
\`\`\`

Happy to open a PR.

---

FraSanga:
Closes #7277.

`send_email()` requires `PHPMailer.php` and `SMTP.php` but never `Exception.php`, even though all three are part of the vendored PHPMailer library. This is invisible on the happy path — `PHPMailer\PHPMailer\Exception` is only touched when `smtpConnect()`/`smtpSend()`/`postSend()` hit a real connection or auth failure and try to throw it. Since the class was never loaded, PHP can't construct it, so instead of a catchable exception (which PHPMailer's own code turns into a clean `Send()` returning `false`), you get an uncaught fatal `Error` that crashes the whole request.

Tested against a live deployment with a deliberately-broken SMTP password: before this change, a failed send crashes with `Class "PHPMailer\PHPMailer\Exception" not found`; after, `send_email()` returns `false` as intended.

---

---
## Summary by cubic
`send_email()` no longer crashes with an uncaught `Class "PHPMailer\PHPMailer\Exception" not found` error on real SMTP connection or auth failures; it now requires `Exception.php` alongside `PHPMailer.php` and `SMTP.php` and returns `false` as intended. SMTP error details are now written to the error log via `error_log()` instead of being echoed to end users. Closes #7277.

<sup>Written for commit 6217599ee5b2d16a33b4a7d5a7578952ed1bcf27. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/BOINC/boinc/pull/7285?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/html/inc/email.inc
+++ b/html/inc/email.inc
@@ -64,7 +64,7 @@ function send_email($user, $subject, $body, $body_html=null, $email_addr=null) {
             $mail->Body = $body;
         }
         if (!$mail->Send()) {
-            echo htmlspecialchars($mail->ErrorInfo, ENT_QUOTES, 'UTF-8');
+            error_log("send_email() failed: ".$mail->ErrorInfo);
             return false;
         } else {
             return true;
--- a/html/inc/email.inc
+++ b/html/inc/email.inc
@@ -64,7 +64,7 @@ function send_email($user, $subject, $body, $body_html=null, $email_addr=null) {
             $mail->Body = $body;
         }
         if (!$mail->Send()) {
-            echo $mail->ErrorInfo;
+            echo htmlspecialchars($mail->ErrorInfo, ENT_QUOTES, 'UTF-8');
             return false;
         } else {
             return true;
--- a/html/inc/email.inc
+++ b/html/inc/email.inc
@@ -43,7 +43,8 @@ function send_email($user, $subject, $body, $body_html=null, $email_addr=null) {
         pclose($pipe);
         return true;
     } else if (function_exists("make_php_mailer")) {
-        if (file_exists("../inc/PHPMailer/src/PHPMailer.php") && file_exists("../inc/PHPMailer/src/SMTP.php")) {
+        if (file_exists("../inc/PHPMailer/src/PHPMailer.php") && file_exists("../inc/PHPMailer/src/SMTP.php") && file_exists("../inc/PHPMailer/src/Exception.php")) {
+            require_once("../inc/PHPMailer/src/Exception.php");
             require_once("../inc/PHPMailer/src/PHPMailer.php");
             require_once("../inc/PHPMailer/src/SMTP.php");
         } else if (file_exists("../inc/phpmailer/class.phpmailer.php")) {


FraSanga:
Added a second commit (`cf80d87`) addressing something found while reviewing #7287: `echo $mail->ErrorInfo;` here was unescaped, and several callers (see #7286, #7287) now render this text into an actual HTML page on a failed send — raw `<`/`>`/`&` in an SMTP error could break markup or inject HTML. Escaping it once here, at the source, covers every caller instead of requiring each one to remember to do it individually.

Note: `send_email()` is also called from a couple of CLI/cron scripts (`update_uotd.php`, `notify.php`) where this text goes into a plain-text log file rather than a web page — with this change, a rare SMTP error containing those characters would show up HTML-entity-encoded there instead of literal. Minor readability quirk in that one context, not a concern there since it's not rendered as HTML; happy to reconsider if that tradeoff isn't wanted.

---

FraSanga:
@davidpanderson @AenBleidd — before going further on the related PRs (#7286, #7287), I'd like a maintainer decision on the approach taken here.

The original fix (`c33d4cb`) just adds the missing `Exception.php` require to stop the fatal error. A review comment then raised a separate concern: on a failed send, `send_email()` was echoing `$mail->ErrorInfo` (raw PHPMailer/SMTP error text) directly into the page — visible to whatever end user triggered the send. `cf80d87` escaped that output, but a follow-up comment pointed out escaping isn't really the fix: SMTP error text can include internal hostnames, credentials-adjacent config details, or server responses that shouldn't reach an end user at all, escaped or not.

`6217599` changes this to `error_log()` the error and return `false` without echoing anything, so callers decide what (if anything) to show the user. This also removes the earlier `htmlspecialchars()` call entirely, since there's no longer any error text being output here.

This is a behavior change beyond the original bug report, so I wanted to flag it explicitly rather than assume it's the right call. Does this approach look right, or would you prefer `send_email()` keep returning the error text (e.g. via `$mail->ErrorInfo` staying readable by the caller) so a project can choose to surface a sanitized version itself?

---

*Assisted by AI (Claude Sonnet 5) per the BOINC AI Assistants Usage Policy.*

PR #7286, opened 2026-09-06
Check send_email()'s return value in 4 account-flow pages

### LINKED ISSUE #7278: validate_email_addr.php, delete_account_request.php, edit_email_action.php, team_founder_transfer_action.php don't check send_email()'s return value
\`send_email()\` returns \`false\` on failure, and several pages that depend on it ignore that return value and unconditionally show a success message:

- \`html/user/validate_email_addr.php\`'s \`send_validate_email()\` — always shows "An email has been sent to you," even on a failed send.
- \`html/user/delete_account_request.php\`'s \`delete_account_request_action()\` — always shows "Confirmation Email Sent." This one's the most consequential: the confirmation link is the only way to delete an account, so a silent failure here blocks the feature with zero indication anything went wrong.
- \`html/user/edit_email_action.php\` — the address change itself succeeds (correctly), but the page never says whether the two notice emails (to the old and new address) failed.
- \`html/user/team_founder_transfer_action.php\` — both the \`initiate_transfer\` and \`decline\` branches claim the other party was notified "by email" regardless of whether the send actually worked.

**Suggested fix:** check \`send_email()\`'s return value at each call site and show an honest failure message. The underlying action (DB update, PM, etc.) should stay non-blocking either way — only the message needs to be accurate.

Happy to open a PR with all four.

---

FraSanga:
Closes #7278.

`send_email()` returns `false` on failure, but these pages ignore that return value and unconditionally show a success message:

- `validate_email_addr.php`'s `send_validate_email()` — always shows "An email has been sent to you."
- `delete_account_request.php`'s `delete_account_request_action()` — always shows "Confirmation Email Sent." This one's the most consequential: the confirmation link is the only way to delete an account, so a silent failure blocked the feature with zero indication anything went wrong.
- `edit_email_action.php` — the address change itself succeeds (correctly, unchanged here), but the page never said whether the two notice emails (old + new address) failed.
- `team_founder_transfer_action.php` — both the `initiate_transfer` and `decline` branches claimed the other party was notified "by email" regardless of whether the send actually worked.

Each fix checks the return value and shows an honest message on failure; the underlying action (DB update, PM, etc.) stays non-blocking either way — only the message changes.

`validate_email_addr.php`/`delete_account_request.php` also had `page_head()` called only after the send, so a failure's raw `echo $mail->ErrorInfo` (from inside `send_email()`) landed before any `<head>`/stylesheet output and broke the page's CSS. Moved `page_head()` before the send in both.

Tested against a live deployment with SMTP deliberately broken: each page now shows the correct failure message instead of a false success, with CSS intact.

---

---
## Summary by cubic
Fixes four account-flow pages that claimed an email was sent even when `send_email()` failed.

- `delete_account_request.php` is the most consequential: the confirmation email is the only way to delete an account, so a silent failure blocked the feature with no indication.
- `validate_email_addr.php` and `delete_account_request.php` now call `page_head()` before sending, so a failed send no longer prints raw error output before the stylesheet and breaks page CSS.
- `edit_email_action.php` and `team_founder_transfer_action.php` now report when the notice emails failed to send; the underlying action (email address change, transfer request/decline) stays non-blocking.

<sup>Written for commit 434204e6b38186c1a3cc7a6f11178960714fec65. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/BOINC/boinc/pull/7286?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/html/user/delete_account_request.php
+++ b/html/user/delete_account_request.php
@@ -57,10 +57,14 @@ function delete_account_request_form($user) {
 function delete_account_request_action($user) {
     $passwd = post_str("passwd");
     check_passwd_ui($user, $passwd);
-    send_confirm_delete_email($user);
+    page_head(tra("Delete account request"));
+    $success = send_confirm_delete_email($user);
 
-    page_head(tra("Confirmation Email Sent"));
-    echo "<p>".tra("The email to confirm your request to delete your account has been sent.")."</p>";
+    if ($success) {
+        echo "<p>".tra("The email to confirm your request to delete your account has been sent.")."</p>";
+    } else {
+        echo "<p>".tra("Can't send confirmation email to %1. Please try again later.", $user->email_addr)."</p>";
+    }
     page_tail();
 }
 
--- a/html/user/edit_email_action.php
+++ b/html/user/edit_email_action.php
@@ -77,7 +77,10 @@ if (!is_valid_email_syntax($email_addr)) {
                 if (defined("SHOW_NONVALIDATED_EMAIL_ADDR")) {
                     echo "<p>".tra("Please %1 validate this email address %2.", "<a href=validate_email_addr.php>", "</a>")."\n";
                 }
-                send_changed_email($user);
+                $email_success = send_changed_email($user);
+                if (!$email_success) {
+                    echo "<p>".tra("Your email address was changed, but we were unable to send the confirmation email to %1 or %2. Please note this change yourself.", $user->email_addr, $user->previous_email_addr)."</p>";
+                }
             } else {
                 echo tra("We can't update your email address due to a database problem.  Please try again later.");
             }
--- a/html/user/team_founder_transfer_action.php
+++ b/html/user/team_founder_transfer_action.php
@@ -105,9 +105,15 @@ case "initiate_transfer":
         // whose founder email is invalid
         //
         $team->update("ping_user=$user->id, ping_time=$now");
-        echo "<p>".tra("The current founder has been notified of your request by email and private message.<br /><br />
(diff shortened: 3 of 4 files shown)


html/user/delete_account_request.php
```diff
@@ -57,10 +57,14 @@ function delete_account_request_form($user) {
 function delete_account_request_action($user) {
     $passwd = post_str("passwd");
     check_passwd_ui($user, $passwd);
-    send_confirm_delete_email($user);
+    page_head(tra("Delete account request"));
+    $success = send_confirm_delete_email($user);
```
FraSanga, line 61: Checked this — `page_head()` (`html/inc/util.inc`) has its own idempotency guard:

```php
global $did_page_head;
if ($did_page_head) {
    return;
}
$did_page_head = true;
```

So if `send_confirm_delete_email()` hits `error_page()` internally (failed `create_token()`), that function's own `page_head()` call is a no-op since `$did_page_head` is already true from our earlier call — no double header, no nested/invalid HTML. `echo $msg; page_tail(); exit();` then run against the page we already started. The only visible effect is the page keeps our "Delete account request" title instead of "Unable to handle request" for that specific rare failure, which is harmless.

No output-buffering needed for this path specifically since `page_head()` is safe to call twice by design — happy to add a comment noting this if useful, but didn't want to change the actual logic without confirming the concern first.

PR #7287, opened 2026-09-06
Don't hide moderation-email failures inside a 302 redirect body

### LINKED ISSUE #7279: forum_moderate_thread_action.php/forum_moderate_post_action.php redirect immediately after a possibly-failed send_email(), silently swallowing any failure
Both files call their respective notification-email function and then immediately \`header('Location: ...')\`, with nothing printed first:

\`\`\`php
send_thread_moderation_email($forum, $thread, $reason, $action_name, $explanation);
header('Location: forum_thread.php?id='.$thread->id);
\`\`\`

Since nothing is output before the redirect, \`header()\` always succeeds — but if the send failed, \`send_email()\`'s own \`echo $mail->ErrorInfo\` still produced output, which ends up as the **body of the 302 response**. Browsers never render a redirect's body, they just follow \`Location:\`, so the moderator sees a completely normal-looking redirect with zero indication the notification email failed — not even the ugly raw error the other flows show, nothing at all. Confirmed by capturing the actual POST request and replaying it with \`curl -i\` (no \`-L\`): the raw SMTP error text is genuinely present in the 302 response body, just never shown by any browser.

Separately, in \`forum_moderate_thread_action.php\`, the "Reason" field's own label says "Mailed if nonempty," but the code sends the notification unconditionally regardless of whether a reason was provided (falling back to the literal string \`"None given"\`):

\`\`\`php
$reason = post_str('reason', true);
if (!$reason) $reason = "None given";
send_thread_moderation_email($forum, $thread, $reason, $action_name, $explanation);
\`\`\`

**Suggested fix:** buffer the send (\`ob_start()\`/\`ob_get_clean()\`), then redirect cleanly on success or render a real page with the captured output on failure. For the label mismatch, only attempt the send when \`$reason\` is actually non-empty, matching what the label already promises.

Happy to open a PR with both.

---

FraSanga:
Closes #7279.

Both files called their respective notification-email function and then immediately `header('Location: ...')`, with nothing printed first:

```php
send_thread_moderation_email($forum, $thread, $reason, $action_name, $explanation);
header('Location: forum_thread.php?id='.$thread->id);
```

Since nothing is output before the redirect, `header()` always succeeds — but if the send failed, `send_email()`'s own `echo $mail->ErrorInfo` still produced output, which ends up as the body of the 302 response. Browsers never render a redirect's body, they just follow `Location:`, so the moderator sees a completely normal-looking redirect with zero indication the notification email failed. Confirmed by capturing the actual POST request and replaying it with `curl -i` (no `-L`): the raw SMTP error text is genuinely present in the 302 response body, just never shown by any browser.

Fix: buffer the send with `ob_start()`/`ob_get_clean()`, then redirect cleanly on success or render a real page with the captured output on failure. The moderation action itself already succeeded before this point either way and is never rolled back. Any dynamic text rendered on the failure page (`$email_output`, `$action_name`) is HTML-escaped.

**Edit:** this PR originally also changed `forum_moderate_thread_action.php` to only send the notification when a reason was given, based on a misreading of the "Reason (Mailed if nonempty)" label as gating the whole send. It doesn't — that label describes `$explanation` (built separately, from the same field) conditionally including the reason *text*; the notification itself should always go out. That change has been reverted (see the commit history and review thread below) — the notification is unconditional again, matching original behavior.

Tested against a live deployment with SMTP deliberately broken: before, a failed send's error was invisible inside the redirect; after, a real "Moderation notice" page shows the failure and the raw error text, CSS intact.

---

*Assisted by AI (Claude Sonnet 5) per the BOINC AI Assistants Usage Policy — see the commits' `Assisted-by:` trailers.*

### DIFF
--- a/html/user/forum_moderate_post_action.php
+++ b/html/user/forum_moderate_post_action.php
@@ -168,7 +168,7 @@ if ($email_success) {
     header('Location: forum_thread.php?id='.$thread->id);
 } else {
     page_head(tra("Moderation notice"));
-    echo $email_output;
+    echo htmlspecialchars($email_output, ENT_QUOTES, 'UTF-8');
     echo "<p>".tra("The post was successfully %1, but the notification email failed to send.", $action_name)."</p>";
     echo "<a href='forum_thread.php?id=$thread->id'>".tra("Return to thread")."</a>";
     page_tail();
--- a/html/user/forum_moderate_post_action.php
+++ b/html/user/forum_moderate_post_action.php
@@ -160,8 +160,18 @@ if (!$result) {
     error_page("Action failed: possible database problem");
 }
 
-send_moderation_email($forum, $post, $thread, $explanation, $action_name);
+ob_start();
+$email_success = send_moderation_email($forum, $post, $thread, $explanation, $action_name);
+$email_output = ob_get_clean();
 
-header('Location: forum_thread.php?id='.$thread->id);
+if ($email_success) {
+    header('Location: forum_thread.php?id='.$thread->id);
+} else {
+    page_head(tra("Moderation notice"));
+    echo $email_output;
+    echo "<p>".tra("The post was successfully %1, but the notification email failed to send.", $action_name)."</p>";
+    echo "<a href='forum_thread.php?id=$thread->id'>".tra("Return to thread")."</a>";
+    page_tail();
+}
 
 ?>
--- a/html/user/forum_moderate_post_action.php
+++ b/html/user/forum_moderate_post_action.php
@@ -166,10 +166,11 @@ $email_output = ob_get_clean();
 
 if ($email_success) {
     header('Location: forum_thread.php?id='.$thread->id);
+    exit;
 } else {
     page_head(tra("Moderation notice"));
-    echo htmlspecialchars($email_output, ENT_QUOTES, 'UTF-8');
-    echo "<p>".tra("The post was successfully %1, but the notification email failed to send.", $action_name)."</p>";
+    echo htmlspecialchars($email_output, ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8');
+    echo "<p>".tra("The post was successfully %1, but the notification email failed to send.", htmlspecialchars($action_name, ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8'))."</p>";
     echo "<a href='forum_thread.php?id=$thread->id'>".tra("Return to thread")."</a>";
     page_tail();
 }
--- a/html/user/forum_moderate_thread_action.php
+++ b/html/user/forum_moderate_thread_action.php
@@ -117,10 +117,23 @@ if (!$result) {
 }
 
 $reason = post_str('reason', true);
-if (!$reason) $reason = "None given";
-send_thread_moderation_email(
-    $forum, $thread, $reason, $action_name, $explanation
-);
-header('Location: forum_thread.php?id='.$thread->id);
+$email_success = true;
+$email_output = '';
+if ($reason) {
+    ob_start();
+    $email_success = send_thread_moderation_email(
+        $forum, $thread, $reason, $action_name, $explanation
+    );
+    $email_output = ob_get_clean();
+}
+if ($email_success) {
+    header('Location: forum_thread.php?id='.$thread->id);
+} else {
+    page_head(tra("Moderation notice"));
+    echo $email_output;
+    echo "<p>".tra("The thread was successfully %1, but the notification email to the thread owner and/or forum moderators failed to send.", $action_name)."</p>";
+    echo "<a href='forum_thread.php?id=$thread->id'>".tra("Return to thread")."</a>";
+    page_tail();
+}
 
 ?>
--- a/html/user/forum_moderate_thread_action.php
+++ b/html/user/forum_moderate_thread_action.php
@@ -117,15 +117,12 @@ if (!$result) {
 }
 
 $reason = post_str('reason', true);
-$email_success = true;
-$email_output = '';
-if (is_string($reason) && trim($reason) !== '') {
-    ob_start();
-    $email_success = send_thread_moderation_email(
-        $forum, $thread, $reason, $action_name, $explanation
-    );
-    $email_output = ob_get_clean();
-}
+if (!$reason) $reason = "None given";
+ob_start();
+$email_success = send_thread_moderation_email(
+    $forum, $thread, $reason, $action_name, $explanation
+);
+$email_output = ob_get_clean();
 if ($email_success) {
     header('Location: forum_thread.php?id='.$thread->id);
 } else {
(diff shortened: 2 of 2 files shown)


html/user/forum_moderate_post_action.php
```diff
@@ -160,8 +160,18 @@ function hide_explanation() {
...
+ob_start();
+$email_success = send_moderation_email($forum, $post, $thread, $explanation, $action_name);
+$email_output = ob_get_clean();
 
-header('Location: forum_thread.php?id='.$thread->id);
+if ($email_success) {
+    header('Location: forum_thread.php?id='.$thread->id);
+} else {
+    page_head(tra("Moderation notice"));
+    echo htmlspecialchars($email_output, ENT_QUOTES, 'UTF-8');
```
FraSanga: Fixed in 334702a339beac7ff4c985bb849716c32fc98959 — wrapped in `htmlspecialchars($email_output, ENT_QUOTES, 'UTF-8')`.

html/user/forum_moderate_thread_action.php
```diff
@@ -117,10 +117,23 @@
...
+    $email_success = send_thread_moderation_email(
+        $forum, $thread, $reason, $action_name, $explanation
+    );
+    $email_output = ob_get_clean();
+}
+if ($email_success) {
+    header('Location: forum_thread.php?id='.$thread->id);
+} else {
+    page_head(tra("Moderation notice"));
+    echo htmlspecialchars($email_output, ENT_QUOTES, 'UTF-8');
```
FraSanga: Fixed in 334702a339beac7ff4c985bb849716c32fc98959 — wrapped in `htmlspecialchars($email_output, ENT_QUOTES, 'UTF-8')`.

html/user/forum_moderate_thread_action.php
```diff
@@ -117,10 +117,23 @@
...
 
 $reason = post_str('reason', true);
-if (!$reason) $reason = "None given";
-send_thread_moderation_email(
-    $forum, $thread, $reason, $action_name, $explanation
-);
-header('Location: forum_thread.php?id='.$thread->id);
+$email_success = true;
+$email_output = '';
+if ($reason !== null && trim($reason) !== '') {
```
FraSanga: Fixed in 334702a339beac7ff4c985bb849716c32fc98959 — changed to `$reason !== null && trim($reason) !== ''`, which also addresses the whitespace-only-reason point.

html/user/forum_moderate_thread_action.php
```diff
@@ -117,10 +117,23 @@
...
+        $forum, $thread, $reason, $action_name, $explanation
+    );
+    $email_output = ob_get_clean();
+}
+if ($email_success) {
+    header('Location: forum_thread.php?id='.$thread->id);
+} else {
+    page_head(tra("Moderation notice"));
+    echo htmlspecialchars($email_output, ENT_QUOTES, 'UTF-8');
+    echo "<p>".tra("The thread was successfully %1, but the notification email to the thread owner and/or forum moderators failed to send.", htmlspecialchars($action_name, ENT_QUOTES, 'UTF-8'))."</p>";
```
FraSanga: Good catch — fixed in 334702a339beac7ff4c985bb849716c32fc98959. This was a real gap: the "title" action builds $action_name from raw `post_str('newtitle')` with no escaping, and this new failure branch is the first place in this file that ever renders $action_name into an HTML page (stock code only ever put it into a plain-text email body). Wrapped in `htmlspecialchars($action_name, ENT_QUOTES, 'UTF-8')`, matching your suggestion. Checked forum_moderate_post_action.php's $action_name values too — those are all static strings ("hidden"/"unhidden"/"moved to another thread"), no equivalent issue there.

html/user/forum_moderate_thread_action.php
```diff
@@ -117,10 +117,23 @@
...
 
 $reason = post_str('reason', true);
-if (!$reason) $reason = "None given";
-send_thread_moderation_email(
-    $forum, $thread, $reason, $action_name, $explanation
-);
-header('Location: forum_thread.php?id='.$thread->id);
+$email_success = true;
+$email_output = '';
+if (is_string($reason) && trim($reason) !== '') {
```
FraSanga: Good catch, valid — confirmed `post_str()` (`html/inc/util.inc`) does no type checking, so `reason[]=x` in the POST body makes $_POST['reason'] an array and `trim()` on it throws. Fixed in 9406be784ce9f501788dd176bbd2181613b468db: switched to `is_string($reason) && trim($reason) !== ''`, which also subsumes the earlier null/empty-string check.

html/user/forum_moderate_thread_action.php
```diff
@@ -117,10 +117,23 @@
...
-header('Location: forum_thread.php?id='.$thread->id);
+$email_success = true;
+$email_output = '';
+if ($reason !== null && trim($reason) !== '') {
+    ob_start();
+    $email_success = send_thread_moderation_email(
+        $forum, $thread, $reason, $action_name, $explanation
+    );
+    $email_output = ob_get_clean();
+}
```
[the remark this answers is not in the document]
FraSanga: Confirmed, you were right — traced through `send_thread_moderation_email()` and $message/$reason (the 3rd param) is never actually referenced in the function body; $explanation (built separately in the caller from the same field) already correctly handles the nonempty-reason-text case. Reverted the send-gating change in 7af0ebb483e1f3e3c30c7692be3768e5dcc234f5 — the notification is unconditional again, matching original behavior. Thanks for catching this.

html/user/forum_moderate_thread_action.php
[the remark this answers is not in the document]
FraSanga: Confirmed, valid — traced through `send_thread_moderation_email()` and $message/$reason (the 3rd param) is never actually referenced in its body; $explanation (built separately in the caller from the same field) already correctly handles the nonempty-reason-text case. Reverted the send-gating change in 7af0ebb483e1f3e3c30c7692be3768e5dcc234f5 — the notification is unconditional again, matching original behavior.

FraSanga:
Pushed a follow-up commit (c68c1e6) addressing the remaining review feedback:
- `exit;` after both successful redirects, so no code path can ever emit output after a `Location:` header.
- `$action_name` is now escaped in `forum_moderate_post_action.php` too, matching the thread-action file.
- `ENT_SUBSTITUTE` added alongside `ENT_QUOTES` on every `htmlspecialchars()` call here, so invalid UTF-8 in an SMTP error can't silently collapse the output to an empty string.

On the info-disclosure point about showing raw SMTP/email-helper output to moderators at all (also raised on #7285): agreed this shouldn't happen. The fix belongs in `send_email()` itself (#7285, currently in review) — logging the error server-side instead of returning/echoing it. Once that lands, the `ob_start()`/`ob_get_clean()` buffering in both files here becomes dead code and I'll remove it in a follow-up, but I'm holding off on that removal until #7285 actually merges so this PR stays independently mergeable in the meantime.

---

